PR #3, opened 2014-07-07
7.0 travis coverage

yvaucher:


pedrobaeza:
:+1:

bwrsandman:
Don't forget to add README.md with badges

yvaucher:
Will need to add settable options on run test for that running_env parameter

bwrsandman:
> Will need to add settable options on run test for that running_env parameter

What do you mean? What is running_env?

yvaucher:
See the thrace:

> File "/home/travis/build/OCA/server-tools/server_environment/serv_config.py", line 40, in <module>
> 
> "The parameter 'running_env' has not be set neither in base config "
> 
> Exception: The parameter 'running_env' has not be set neither in base config file option -c or in openerprc.
> 
> We strongly recommend against using the rc file but instead use an explicit config file with this content:
> 
> [options]
> 
> running_env = dev

This is used in server_environement module to define dev/test/qa/prod instance

yvaucher:
yay it passed the test with server_environment

PR #2, opened 2014-07-06
add module configuration_helper

bealdav:
validated on https://code.launchpad.net/~akretion-team/server-env-tools/module-configuration-helper-dbl/+merge/220392

PR #7, opened 2014-07-08
[ADD] module tree_view_record_id

bealdav:


legalsylvain:
Hi David, 
:question:
I think that there is a little problem. You propose for merge both modules 'tree_view_record_id' and 'configuration_helper' (allready merge proposed in #2 ).
Maybe a little rebase, or something like that could fix the problem.

Regards.

bealdav:
caramba, encore raté,

In fact #2 was one error because proposed on master instead of 7.0.
My git skills are so insufficient.
I don't know if I must define a new branch per PR or not ?

bealdav:
PR cancellation

PR #8, opened 2014-07-09
[ADD] new module 'module_parent_dependencies'

legalsylvain:
Add a module 'module_parent_dependencies' that allows admin user to see the modules that depend of a target modules. (direct and undirect dependencies)
Flake8: OK;
Coverage: 100%;

(Sample:  https://github.com/legalsylvain/server-tools/blob/a9761578835226ab8df897b1ba397d8e6e85987d/module_parent_dependencies/static/src/img/screenshots/dependencies_list.jpg)


module_parent_dependencies/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+    'version': '0.1',
+    'summary': 'Display the parent dependencies of a module',
+    'category': 'Tools',
+    'description': """
+Display the parent dependencies of a module
+===========================================
+
+Functionality:
+--------------
+    * This module display to the admin user, the possibility to see the list"""
```
rvalyi: displays

but I would reformulate:
This module enables the admin user to see the list of modules dependencies of a given module, at the full depth of the dependency tree.

module_parent_dependencies/tests/test_module_parent_dependencies.py
```diff
@@ -0,0 +1,65 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Database Integrity module for OpenERP
```
rvalyi: bad copy pase?

module_parent_dependencies/view/view.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ********************************************************************** -->
+<!--   odule - Parent Dependencies module for Odoo                          -->
```
rvalyi: odule -> module

rvalyi:
see my previous inline comments, would approve otherwise. Cool that you made full test coverage.

module_parent_dependencies/__openerp__.py
pedrobaeza: Not exactly: anyone that has configuration permission (and thus see Installed modules menu entry), can see this addition, so I would put:

This module allows to see the list of modules dependencies of a given module, at the full depth of the dependency tree.

module_parent_dependencies/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+    """ of the modules that depends (directly or indirectly) on a module.
+
+Copyright, Authors and Licence:
+-------------------------------
+    * Copyright: 2014, GRAP: Groupement Régional Alimentaire de Proximité;
+    * Author: Sylvain LE GAL (https://twitter.com/legalsylvain);
+    * Licence: AGPL-3 (http://www.gnu.org/licenses/);""",
```
pedrobaeza: There is an specific field on ir.module.module for license. Do you see this necessary?

legalsylvain: I put it because: 
- this field is not displayed in Odoo apps, and it's important information ; Sample: https://www.odoo.com/apps/7.0/web_widget_float_formula/
- it is possible to add a link, to help new people to discover the licences;

Regards.

pedrobaeza: Now, we are including the LICENSE file at the root of each repository, so I think it's redundant.

module_parent_dependencies/model/ir_module_module.py
```diff
@@ -0,0 +1,72 @@
...
+class module(Model):
+    _inherit = 'ir.module.module'
+
+    # Field function Section
+    def _get_all_parent_ids(
+            self, cr, uid, ids, field_name, arg, context=None):
+        res = self._get_direct_parent_ids(
+            cr, uid, ids, field_name, arg, context=context)
+        for id in ids:
+            parent_ids = list(res[id])
```
hbrunn, line 36: why is this necessary?

legalsylvain, line 36: Hi Holger, 
thanks for your review. I'm sorry but I don't understand your question. What don't you find unnecessary ? l35 & 36 ?
If yes, the objective is just to remove duplicated ids. res[id] give the parents, and parent_ids is the concatenation of all the direct parents, but unique, to avoid to call _get_all_parent_ids (l37) with the same args.
I can see that there is maybe a shorter recursive function to do the same things.

Regards.

hbrunn, line 36: I asked about the list() call in 36. From what I understand, you can't get anything else than an iterable there. And from what you write, I think you want this to be a set()

legalsylvain, line 36: Hi Holger, 

Sorry for the delays.
That line is used to realize a copy of res[id]

```
parent_ids = list(res[id]) 
```

If I just write:

```
 parent_ids = res[id]
```

it will break the for loop, because the loop add new item in res[id] and so in parent_ids.
I'm not Python expert, maybe there is something more clear or more elegant.

To test that point:  
- install 'module_parent_dependencies' and some  module with a lot of dependencies like 'point_of_sale';
- select 'base' module to see dependencies;

Result:
- with : parent_ids = list(res[id]) --> OK.
- with : parent_ids = res[id] --> KeyError on line 40. ir_module_module.py

Regards.

legalsylvain:
@pedrobaeza @rvalyi Are you OK with the changes I did ? (licences, error in copy past of header, etc...) 

Regards.

rvalyi:
:+1:  for me.
sorry for the delay; I had lost it from the radar.

module_parent_dependencies/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import model
```
pedrobaeza: `from . import model`

module_parent_dependencies/__openerp__.py
```diff
@@ -0,0 +1,56 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Parent Dependencies of Modules',
+    'version': '0.1',
+    'summary': """allows to see the list of modules dependencies of a given"""
+    """ module, at the full depth of the dependency tree""",
```
pedrobaeza: Why multiple """? This is intended for a multi-line text. You do this if you are treating with single (') or double quotes (")

module_parent_dependencies/model/ir_module_module.py
pedrobaeza, line 36: The question here is that you are overwritting the value each time on the loop. I don't know exactly what do you want to do, but this are the possibilities:
- You want to add all values:

```
parent_ids = []
for id in ids:
    parent_ids.append(res[id])
```
- You only want first value:

```
parent_ids = list(res[ids[0]])
```
- You want last value:

```
parent_ids = list(res[ids[-1]])
```

module_parent_dependencies/model/ir_module_module.py
```diff
@@ -0,0 +1,72 @@
...
+            imd_ids = imd_obj.search(
+                cr, uid, [('name', '=', imm.name)], context=context)
+            tmp = imd_obj.read(
+                cr, uid, imd_ids, ['module_id'], context=context)
+            imm_ids = [x['module_id'][0] for x in tmp]
+            # Select only non uninstalled module
+            imm_ids = self.search(
+                cr, uid, [
+                    ('id', 'in', imm_ids),
+                    ('state', 'not in', ['uninstalled', 'uninstallable'])],
```
pedrobaeza, line 58: Why don't you want to show uninstalled parents on the dependency graph?

legalsylvain, line 58: Hi, 

I realize that module because some times, I want to uninstall a module X, and I want to know clearly which modules depend on X directly or indirectly and so which module will be uninstalled.

I think that seing all modules that depend on X in your all official and custom repositories will be mostly unreadable. Or, it's another functionality.

module_parent_dependencies/model/ir_module_module.py
legalsylvain, line 36: I'm sorry, but I really don't understand what is the problem with that line.

I just want to copy all value of res[id] in parent_ids.
Your three possibilities will do other result. (I don't want to append values each time).

Regards.

module_parent_dependencies/__openerp__.py
legalsylvain: Changed. Thanks for the remark.

module_parent_dependencies/__init__.py
legalsylvain: Done. Thanks.

module_parent_dependencies/model/ir_module_module.py
pedrobaeza, line 58: I think it can be then interesting to let the use choose if they want all modules or only install ones

legalsylvain, line 58: I can not let the user choose. It's a field.function and the aims of this field is to show all the installed modules that depend on a given module. (directly or indirectly).
Showing ALL the modules that depend on a given module (installed or not) is just an other thing.
I just don't need that information but feel free to propose a PR with that feature. 

Regards.

module_parent_dependencies/model/ir_module_module.py
pedrobaeza, line 36: Sorry, I made a mistake copying/pasting. Now my comment is correct. Please read it again carefully to try to understand what I'm saying.

module_parent_dependencies/model/ir_module_module.py
pedrobaeza, line 58: OK, I'll abstain on this point, but you should warn in the module description about this behaviour.

OSguard, line 58: i agree, change the description that only install modules will be listed. I think this are the important one.
Also Update the label and include "installed" in it.
The use case i see if you "click upgrade" in a module OpenERP upgrades all parents. So this module will helps very well to understand what will happen if you upgrade a module an which other modules will be triggered.

max3903:
The pot file is missing.

:-1: 

I am not in favor of having header in the xml file, but I would accept this PR if you want to keep it.

legalsylvain:
Hi @max3903. 
Thanks for the review.
I'll add the pot file;
About the header in the xml file, are there some reasons you can explain ?

module_parent_dependencies/model/ir_module_module.py
legalsylvain, line 58: Hi @OSguard & @pedrobaeza,
I did the suggedted changes. Hope that all is ok for you.

Regards.

pedrobaeza:
:+1: for me.

legalsylvain:
FYI, Travis is failing for pep8 bug on other module;
@max3903 : Are you now ok ?
@ all, @max3903 : About the XML header, can you explain or send me a link about why is it bad to include licence & copyright in xml files. I did'nt see any kind of Best Practice that said that in OCA / Odoo website.

legalsylvain:
About pep8, you can review and merge #71  to fix the problem.
Regards.

max3903:
@legalsylvain: Yes I am ok now. Thank you. :+1: 

Regarding the header, XML files in Odoo usually contains data and I don't think data can be copyrighted. I would not argue on a legal standpoint, as I am not expert, but you will agree with me there is nothing either in the best practices of Odoo or OCA encouraging the use of header in XML files. 

It is also something that it is not very common and widespread in Odoo and OCA modules, so why should we make an exception here ?

foutoucour:
This module is interesting.

Could you fixed the flake8 warnings?

Regards
J

veloutin:
This looks interesting. The build seems to fail flake8 for unrelated modules.
@legalsylvain Regarding the copyright in XML files, it is rather unusual. In fact, I have a hard time
finding any other XML file in different OCA repositories which have one. 

:+1:  for the module.

OSguard:
@legalsylvain can you rebase if this is the problem with flake as #71 is already merged

@max3903 is right xml is the same as html and css it is data so any open source license can protect this, so you need something like creative commons. So can you remove it for now and we can make a general discussion on that

legalsylvain:
@OSguard : merged from upstream done; Travis & Coverage are OK.
@OSguard, @max3903, @veloutin  : I Changed xml header.

Thanks for your review.

OSguard:
:+1: so as all are fix it can be merged

PR #9, opened 2014-07-11
[ADD] tree_view_record_id module

bealdav:


tree_view_record_id/orm.py
```diff
@@ -0,0 +1,57 @@
...
+fields_view_get_orginal = orm.BaseModel.fields_view_get
+
+
+def fields_view_get(self, cr, uid, view_id=None, view_type='form', context=None,
+                    toolbar=False, submenu=False):
+    res = fields_view_get_orginal(
+        self, cr, uid, view_id=view_id, view_type=view_type, context=context,
+        toolbar=toolbar, submenu=submenu)
+    if view_type == 'tree':
+        compatible_tree = False
```
rvalyi: wouldn't it it be better to put straight:

```
compatible_tree = res.get('field_parent', True) is False
```

so without the test later and potential double assignation?

tree_view_record_id/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+#
+###############################################################################
+
+{
+    'name': 'Tree View Record Id',
+    'version': '0.1',
+    'category': 'Other modules',
+    'sequence': 10,
+    'author': 'Akretion',
+    'summary': "Add id field to tree views",
```
rvalyi: Add -> Adds

tree_view_record_id/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+    'description': """
+Add Id field to all non arborescent tree views.
```
rvalyi: Add -> Adds

rvalyi:
Hello @bealdav,

see previous inline comments. Also I'm not sure if it's better to have this as a monkey patch like you did or as a mixin we could apply on the objects we want the feature selectively. I have no opinion here, just mentioning the alternative. I would approve if you fix my previous points. Thanks!

bealdav:
Hi Raphaël, 
Thanks for comments. I fixed the code as you recommended
That works fine on local base, but travis is failed.
It seems travis details doesn't indicate that it is tree_view_record_id that break tests but I'm expert in CI backtrace analyse.
What do you think ?

Caramba ! encore raté

tree_view_record_id/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+    'website': 'http://www.akretion.com',
+    'depends': [
+        'base',
+    ],
+    'data': [
+    ],
+    'demo': [
+    ],
+    'installable': True,
+    'auto_install': True,
```
legalsylvain: :question: 
'auto_install': true + depends: 'base' = we have no choice to not install that module ?
ref: https://www.odoo.com/fr_FR/forum/help-1/question/auto-install-installable-active-in-openerp-py-file-19776

regards.

bealdav: Thanks to notice Sylvain

Caramba

rvalyi:
:+1: 
for the record Travis is failing because of an other module (nothing to do with your pull request). I approve your module now. Thanks!

legalsylvain:
:+1: Thanks!
(Code review, no test.)

hbrunn:
wouldn't it be less hassle to add the field on the client side in an extension of web.ListView.load_list? If you still want to do this on the client side, you can do the monkey patch in DummyModel's _register_hook, this way you know it only happens if the addon is installed

bealdav:
Hi Holger, 
Thanks to notice me _register_hook. I only saw one use in addons. You use it in one of your module ?
Is it possible to extend web.ListView with conditions on database structure (some tables may not have id field) in client side ? I don't know how to do it.
Thanks

bealdav:
Hi @hbrunn,
Is it possible to extend web.ListView with conditions on database structure ? 
Thanks

hbrunn:
You could call fields_get on the model and do something depending on the result. But I don't think that's necessary in this case. If you end up there, there should be an id columns. Simply try something like

```
load_list: function(data)
{
    if(!data.fields['id'])
    {
        data.fields['id'] = {'type': 'integer'};
        data.arch.children.unshift({'tag': 'field', 'attrs': {'name': 'id'}});
    }
    return this._super.apply(this, arguments);
}
```

As for _register_hook, here's an example: https://github.com/Therp/fulltextsearch/blob/7.0/fts_base/fts_proxy.py#L66

bealdav:
Hi @hbrunn

I've seen than the code about '_register_hook' use change a lot between those 2 branches
https://github.com/odoo/odoo/blob/8.0/addons/base_action_rule/base_action_rule.py#L155
https://github.com/odoo/odoo/blob/7.0/addons/base_action_rule/base_action_rule.py#L192

This hook is only use in one official addon.
I tested to use it but with no success and I have no more time.
My initial solution works and I suggest to add a comment to move to '_register_hook' on 8.0 version.

What do you think ?

hbrunn:
Yes, I'm okay with that. The last thing to do is fix the line lengths as indicated in
./tree_view_record_id/**openerp**.py:33:80: E501 line too long (118 > 79 characters)
./tree_view_record_id/**openerp**.py:35:80: E501 line too long (83 > 79 characters)
./tree_view_record_id/orm.py:39:80: E501 line too long (80 > 79 characters)
./tree_view_record_id/orm.py:56:80: E501 line too long (82 > 79 characters)

bealdav:
OK Holger, 

fixed, then 3 approves, 
which step to do before merge ?

Thanks a lot for indications.

hbrunn:
:+1:

hbrunn:
now say git rebase and squash all commits so that there's only [ADD] $yourmodule, then it's done

max3903:
:+1:

foutoucour:
:+1:

PR #16, opened 2014-07-17
[7.0] Tighten type check in domain

bwrsandman:
Fixes #15 and #24
Check that `domain[2]` is `str` because in some cases it's an `int` and `list`
operations don't work on `int`s.

PR #18, opened 2014-07-17
[master] Tighten type check in domain

### LINKED ISSUE #15: [Bug] mass_editing - Search in ir.model.fields not working
I setup the mass_editing for res.partner then i go to:
Settings -> Technical -> Database Structure -> Fields
and go to the Filter with ('model_id','=',61) where 61 is the id of res.partner in res.model. By typ "res.partner" and select "Partner" from the suggestion.

Then i get the following error:
File "[..]/mass_editing/mass_editing.py", line 34, in search
    model_domain += [('model_id', 'in', map(int, domain[2][1:-1].split(',')))]
TypeError: 'int' object has no attribute '**getitem**'

bwrsandman:
Fixes #15
Check that `domain[2]` is `str` because in some cases it's an `int` and `list`
operations don't work on `int`s.

### DIFF
--- a/mass_editing/models/ir_model_fields.py
+++ b/mass_editing/models/ir_model_fields.py
@@ -31,12 +31,14 @@ class IrModelFields(orm.Model):
             count=False):
         model_domain = []
         for domain in args:
-            if domain[0] == 'model_id' and domain[2]\
-                    and type(domain[2]) != list:
-                model_domain += [(
-                    'model_id', 'in', map(int, domain[2][1:-1].split(',')))]
+            if (len(domain) > 2 and domain[0] == 'model_id'
+                    and isinstance(domain[2], basestring)):
+                model_domain += [
+                    ('model_id', 'in', map(int, domain[2][1:-1].split(',')))
+                ]
             else:
                 model_domain.append(domain)
         return super(IrModelFields, self).search(
             cr, uid, model_domain, offset=offset, limit=limit, order=order,
-            context=context, count=count)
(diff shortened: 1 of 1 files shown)


OSguard:
i tested the patch and it solve the problem. Thanks

bwrsandman:
a `:+1:` would be apreciated :)

pedrobaeza:
Here you have ;)

:+1: (code review)

hbrunn:
:+1:

sbidoul:
:+1: (code review)

bwrsandman:
Rebased

mass_editing/models/ir_model_fields.py
```diff
@@ -32,7 +32,7 @@ def search(
         model_domain = []
         for domain in args:
             if domain[0] == 'model_id' and domain[2]\
```
gurneyalex: @bwrsandman I think you should also remove the `and domain[2]` which is useless (and confusing) after the change below, as an empty string will not cause a problem. You may want to change it to `and len(domain) == 3` which will avoid an `IndexError` when confronted with a bad formed domain, and which maybe was the original intent, but I would not do this as this will most certainly crash a couple of calls further down in the stack. 

-> needs fixing

bwrsandman:
@gurneyalex done

gurneyalex:
@bwrsandman OK, I'll merge this and port the merge to 7.0 too (instead of #16)

PR #26, opened 2014-07-22
[ADD] port disable_openerp_online

hbrunn:


disable_openerp_online/__openerp__.py
```diff
@@ -39,14 +39,10 @@
         'mail',
     ],
     "data": [
+        "views/disable_openerp_online.xml",
         'data/ir_ui_menu.xml',
         'data/ir_cron.xml',
     ],
-    "js": [
-        'static/src/js/disable_openerp_online.js',
```
ovnicraft, line 46: Ths files is removed but loaded from template, will keep working the module ? @hbrunn

ovnicraft, line 46: In 7.0 is annon bar is added by https://github.com/odoo/odoo/blob/7.0/addons/mail/static/src/xml/announcement.xml 
In 8.0 added by https://github.com/odoo/odoo/blob/8.0/addons/mail/static/src/xml/announcement.xml

hbrunn, line 46: ad 1: yes, the 8.0 way to add this kind of files via an xml template (see template with id 'assets_backend').
ad 2: right, and both use a javascript-function show_annoucement_bar to actually display that, this is overwritten with a noop in disable_openerp_online.js

disable_openerp_online/views/disable_openerp_online.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="assets_backend" name="disable_openerp_online assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/disable_openerp_online/static/src/js/disable_openerp_online.js"></script>
```
ovnicraft, line 6: Just to know: this JS will load if not defined on js dict on **openerp** ? @hbrunn

hbrunn, line 6: to convince yourself, do
`grep js addons/*/__openerp__.py`
in your odoo master branch

pedrobaeza:
Hi, Holger, thank you for the port! Some remarks:
- On module title, you should put `Remove odoo.com bindings` instead `openerp`.
- On _data/ir_ui_menu.xml_, instead of removing menus by ID, remove them making a search, like @bwrsandman has made on: https://github.com/OCA/server-tools/pull/27/files#diff-1d976f79901564a61ea766b940d403c1R4. This is intended for avoiding the warning when the second pass that Travis made tries to remove again the menu and it's not located. Current test doesn't made this, but we have to fix it according last maintainer-tools commits.

Regards.

pedrobaeza:
Tests in this branch are outdated, but we can fix it in other PR.

For this, :+1:

rvalyi:
:+1: LGTM sorry for the delay

pedrobaeza:
I proceed with the merge due to the 2 approvals and the past time.

PR #27, opened 2014-07-23
7.0 pep8

bwrsandman:
Forgot commit d007be5
Fix additional style issues
Use official travis

### DIFF
--- a/disable_openerp_online/data/ir_ui_menu.xml
+++ b/disable_openerp_online/data/ir_ui_menu.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
     <data>
-        <delete model="ir.ui.menu" id="base.module_mi" />
-        <delete model="ir.ui.menu" id="base.menu_module_updates" />
+        <delete model="ir.ui.menu" search="[('parent_id', '=', ref('base.menu_management'), ('name', '=', 'Apps'))]" />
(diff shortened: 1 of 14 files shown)


rvalyi:
:+1: LGTM

disable_openerp_online/data/ir_ui_menu.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
     <data>
-        <delete model="ir.ui.menu" id="base.module_mi" />
-        <delete model="ir.ui.menu" id="base.menu_module_updates" />
+        <delete model="ir.ui.menu" search="[('parent_id', '=', ref('base.menu_management'), ('name', '=', 'Apps'))]" />
```
pedrobaeza, line 4: Why do you change this? I prefer removal by id.

bwrsandman, line 4: Because it will cause an exception when init (not update) is run twice, which travis tests do.
I don't like leaving any code that can spontaneously crash.
This is not an easy problem to solve. Either we delete with a search or we tell users to never init twice AND change the Odoo test code.

pedrobaeza, line 4: Yeah, this is a good reason indeed. OK, let's change this then.

Regards.

pedrobaeza:
:+1:

lmignon:
:+1:  LGTM

PR #29, opened 2014-07-24
[7.0] Fix wrong parenthesis

bwrsandman:
Introduced in #27 (sorry)


pedrobaeza:
Uf, what a mistake. I overpassed it on the review.

:+1: Thanks for the patch.

PR #28, opened 2014-07-24
[ADD] This module initialize the session by looking for the field HTTP_R...

lmignon:
...EMOTE_USER in the HEADER of the HTTP request and trying to bind the given value to a user

PR #30, opened 2014-07-25
[ADD] Authentication via HTTP Remote User

lmignon:
This module initialize the session by looking for the field HTTP_REMOTE_USER in the HEADER of the HTTP request and trying to bind the given value to a user.
The PR follows a previous one (https://github.com/OCA/server-tools/pull/28) but all changes have been squashed into a single commit.


lmignon:
The module has been fixed / improved for V8.

PR #32, opened 2014-08-01
[ADD] Migrated <mass_editing> for Odoo

oihane:


pedrobaeza:
:+1:

mass_editing/__openerp__.py
```diff
@@ -23,27 +23,28 @@
...
-    The video explaining the features and how-to for OpenERP Version 6"""
-    """ is here http://t.co/wukYMx1A
-    The video explaining the features and how-to for OpenERP Version 7 is"""
-    """ here : http://www.youtube.com/watch?v=9BH0o74A748&feature=youtu.be
-    For more details/customization/feedback contact us on"""
-    """ contact@serpentcs.com.
+    The video explaining the features and how-to for OpenERP Version 6
+    is here http://t.co/wukYMx1A
+    The video explaining the features and how-to for OpenERP Version 7 is
+    here : http://www.youtube.com/watch?v=9BH0o74A748&feature=youtu.be
```
bwrsandman, line 38: Should we make a Version 8 video?

mass_editing/__openerp__.py
```diff
@@ -23,27 +23,28 @@
...
+    For more details/customization/feedback contact us on
+    contact@serpentcs.com.
     """,
     'depends': ['base'],
     'data': [
         "security/ir.model.access.csv",
-        'mass_editing_view.xml',
+        'views/mass_editing_view.xml',
```
bwrsandman, line 45: I think the idea of putting views in their own subfolder.

Don't you think it should be singular, like the `data/`, `demo/` folders in other modules, and better if you take out the `_view` part of the name?

mass_editing/models/mass_editing.py
```diff
@@ -20,11 +20,12 @@
...
 ##############################################################################
 
-from openerp.osv import orm, fields, osv
+from openerp import SUPERUSER_ID
+from openerp.osv import orm, fields
 from openerp.tools.translate import _
 
 
-class ir_model_fields(orm.Model):
+class IrModelFields(orm.Model):
```
bwrsandman: Do you think you can put one model per python file?

Such as:
- mass_object.py
- ir_model_fields.py

mass_editing/__openerp__.py
bwrsandman, line 45: It's a little confusing with `models/` as well, though I would keep that one, as `odoo.py scaffold` generates it, so I assume that's recommended.

anajuaristi:
:+1:  (code review)

mass_editing/__openerp__.py
pedrobaeza, line 38: If someone wants to do it, it would be good, but it's not mandatory for this PR, don't you think?

mass_editing/__openerp__.py
pedrobaeza, line 45: Odoo again doesn't have an homogeneus rule for naming this, so there are some folders on plural and some on singular. `data` and `demo` are plural themself, so I prefer to make `models` and `views` also plural. About `_view` suffix, it's not that bad, if you make a file search or you make specific files for menus.

mass_editing/models/mass_editing.py
pedrobaeza: I agree with this.

bwrsandman:
:+1:

max3903:
:+1:

PR #34, opened 2014-08-05
[ADD] Authentication via HTTP Remote User 8.0

lmignon:
The module allows integration with external security systems (Shibolleth, Tivoli access manager, ..) that can pass along authentication of a user via Remote_User HTTP header field. In many cases, this is achieved via server like Apache HTTPD or nginx proxying Odoo.


sbidoul:
:+1: code review and tests

gurneyalex:
please resubmit on the 8.0 branch

Issue #35, opened 2014-08-07
Log viewer in Odoo interface


bwrsandman:
Is there a module out there which can show the output of Odoo logs inside a Odoo view?

If not, I feel this would a very important module to make for Odoo admins who do not have access to host filesystems.

There are cases where Cronjobs fail silently (no error message in the interface) and it would be great to be able to see the logs and even to filter them by level.

This is an example in gitlab:

![gitlablog](https://cloud.githubusercontent.com/assets/1013356/3843309/335d5fc2-1e3f-11e4-9008-b5c798419f3f.png)

Implementation hints:
- http://opensourcehacker.com/2011/02/23/temporarily-capturing-python-logging-output-to-a-string-buffer/
- https://docs.python.org/2/library/collections.html#collections.deque
- https://docs.python.org/2/library/logging.handlers.html

max3903:
:+1:

pedrobaeza:
Very good idea.

Can this be a part of:
https://code.launchpad.net/~camptocamp/server-env-tools/7.0-monitoring/+merge/215138 ?

StefanRijnhart:
That would absolutely be a great asset. Looks like a good basis too for a module that I've been thinking about which would be able report any occuring exceptions to an external channel (mail or xmlrpc message)

bwrsandman:
@gurneyalex could you port the MP on https://code.launchpad.net/~camptocamp/server-env-tools/7.0-monitoring/+merge/215138

gurneyalex:
@bwrsandman yep, will do. Thanks for reminding me of this :smile_cat:

gurneyalex:
@bwrsandman see #39

bwrsandman:
In v8, one can enable `--log-db=DBNAME` and logs will be in the interface under Technical Features

PR #36, opened 2014-08-08
WIP - Forward port of remote_logs

bwrsandman:
Implementation by @xrg of #35 taken from http://git.hellug.gr/?p=xrg/openobject-addons;a=commit;h=fc321ca2ee1304

Code is quite old and probably needs to be updated

### DIFF
--- a/remote_logs/__init__.py
+++ b/remote_logs/__init__.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#    
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2011 OpenERP SA. (www.openerp.com)
+#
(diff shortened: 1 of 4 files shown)


rvalyi:
@bwrsandman I assume you are aware that in v8 logs go into the database too. What do you think about that and how will it relates (or not) with this module? Thanks.

bwrsandman:
I wasn't aware of this.

hbrunn:
Two questions: Is the name actually appropriate? To me, the name implies logging to a remote location, wouldn't something involving http be more appropriate?
Logging can include sensitive information, so I think users should be made aware that this path should be blocked by their proxy or something similar.

pedrobaeza:
This one is not going to be worked anymore, so I close.

PR #38, opened 2014-08-13
7.0: fix flake8 errors

gurneyalex:
Fix flake8 errors to get a clean build:
- mostly line wrapping / reindenting
- in super_calendar, extracted a method from an already too large method to gain 2 indentation levels and some clarity
- added local copy of images for super_calendar documentation

On my way, fixed a couple of calls to logging.logger methods which did the param substitutions using % rather than delegating this to the logging handler.


import_odbc/import_odbc.py
```diff
@@ -180,12 +222,19 @@ def import_run(self, cr, uid, ids=None, context=None):
...
+            _logger.log(level,
+                        'Imported %s , %d rows, %d errors, %d warnings.',
+                        model_name,
+                        log['last_record_count'],
+                        log['last_error_count'],
+                        log['last_warn_count'])
             # Write run log, either if the table import is active or inactive
             if log['last_log']:
-                log['last_log'].insert(0, 'LEVEL|== Line ==    |== Relationship ==|== Message ==')
+                log['last_log'].insert(0,
```
pedrobaeza: You can use PEP8 indentation alternative to avoid too much lines:

```
log['last_log'].insert(
    0,'LEVEL|== Line == |== Relationship ==|== Message ==')
```

pedrobaeza:
PEP8 fixed. I prefer not to put too much lines align with the first open (, but move all arguments to next line with one indentation level. In this way, you get higher width to put more arguments in the same line, but this only a taste of preference.

:+1:

nbessi:
:+1:

mail_environment/env_mail.py
```diff
@@ -89,10 +91,12 @@ def _get_smtp_conf(self, cursor, uid, ids, name, args, context=None):
             string='smtp_encryption',
             type="char",
             multi='outgoing_mail_config',
-            help="Choose the connection encryption scheme:\n"
-                 "- none: SMTP sessions are done in cleartext.\n"
-                 "- starttls: TLS encryption is requested at start of SMTP session (Recommended)\n"
-                 "- ssl: SMTP sessions are encrypted with SSL/TLS through a dedicated port (default: 465)",
+            help=("Choose the connection encryption scheme:\n"
```
guewen: AFAIK parenthesis is useless here (already inside parenthesis)

gurneyalex:
@pedrobaeza I personally dislike this way of indenting code (newline after opening paren), and use it only as a last resort... so indeed a matter of taste... :wink:

guewen:
:+1:

bwrsandman:
:+1:

PR #39, opened 2014-08-13
7.0 monitoring

gurneyalex:
port MP from https://code.launchpad.net/~camptocamp/server-env-tools/7.0-monitoring/+merge/215138 as requested as part of #35


gurneyalex:
@pedrobaeza you had comments as part of the Launchpad MP. Some of them are done. 

as for the others: 
- "L685: the 'info' field is either "rpc call" or "cron job". Wouldn't a selection field be more efficient here?" -> I don't think so, as selection fields are not implemented with a foreign table, so you end up with a string in db + a check on the validity of the value
- "L756: option to disable Server Process log_measure is not available yet, correct?" -> correct
- "L771: I worry about this extra search to translate model_name into model_id. Why not store model_name directly?" -> storing an integer instead of a string is a lot more efficient, and it allows using the translations later on, and making joins. 
- "L778: I see that 'status' is kept empty; is this intended to store WARNING and ERROR results? That would be an excellent addition." -> it was originally thought for storing the HTTP status of the xmlrpc query. I'm not sure how to get this, so I'm leaving empty for now. 
- "L1146: for Process logs, would it be possible to add an option to group by day? (currently group by month, which is insufficient...)" -> I'm not sure how to do this. Contributions / help welcome.

gurneyalex:
travis is red because of flake8 issues in other addons. #38 fixes these.

pedrobaeza:
Thanks for the changes, @gurneyalex. For now, is enough. This is a great improvement from having nothing :wink: Afterwards, I will make another PR trying to improve current module.

:+1: 

Regards.

hbrunn:
:+1:

server_monitoring/server_monitoring.py
```diff
@@ -0,0 +1,447 @@
+# -*- coding: utf-8 -*-
```
bwrsandman: This script is far too long, could you please split it per class/model.

server_monitoring/server_monitoring_view.xml
```diff
@@ -0,0 +1,342 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bwrsandman: Same as with the python script, can you split this file by model, the larger parts, at least.

server_monitoring/server_monitoring.py
```diff
@@ -0,0 +1,447 @@
...
+TODO: check if we can log pooler state (whatever that is...)
+"""
+
+import logging
+import gc
+from operator import itemgetter
+import types
+import os
+import threading
+import resource
```
legalsylvain: Hi, 
I have some bad log, installing your module.

```
Ambiguous import: the OpenERP module `resource` is shadowed by another
module (available at /usr/lib/python2.7/lib-dynload/resource.so).
To import it, use `import openerp.addons.<module>.`.
```

server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,59 @@
...
+        'server_monitoring_view.xml',
+        'server_monitoring_data.xml',
+        'security/ir.model.access.csv',
+        ],
+    'test': [
+        'tests/test_monitor_process.yml',
+        'tests/test_monitor_database.yml',
+        ],
+    'installable': True,
+    'application': False,
```
legalsylvain: (Minor comment)
L57: 'True' is the default value, is it necessary ?
L58: 'False' is the default value, is it necessary ?

legalsylvain:
Hi, 
Just finish to test this module and it's very usefull. Thanks a lot !
:+1: (except the import problem)
(Functional review)

hbrunn:
I'll label this 'needs fixing' for Sandy's points

dreispt:
@gurneyalex did you figure out anything about those SERIALIZATION_FAILURE errors I found before?

server_monitoring/server_monitoring.py
gurneyalex: There's nothing much I can do about this. What I need here is the stdlib resource module, not openerp.addons.resource. This warning is caused by Odoo playing tricks with sys.path to allow shortcut imports, as well as importing from 'odoo' or 'openerp' indifferently.

gurneyalex:
@dreispt I'm a bit under the water atm and did not get a chance to work on this addon. I'll give this a careful look Real Soon[tm].

From what I remember, these errors where logged but the actions would be performed anyway and retried. But I need to double check.

server_monitoring/__openerp__.py
dreispt: L57 is better left as is, since it allows a find/replace to disable modules.
L58 is unnecessary.

pedrobaeza:
@gurneyalex, can you check last comments and I'll merge?

gurneyalex:
@dreispt @pedrobaeza I've updated the addon, split the large files in smaller ones, added a README.rst, squashed the commits and fixed the CI. 

Regarding the serialization failures, I think this should be solved by installing `base_concurrency` (in OCA/server-tools). If you feel it's worth it, I can add a dependency.

server_monitoring/README.rst
```diff
@@ -0,0 +1,88 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+server monitoring
+==============
```
pedrobaeza: This should be of the same length

pedrobaeza: You can add the same line on top to make the title more important (see https://github.com/OCA/maintainer-tools/pull/109)

server_monitoring/README.rst
```diff
@@ -0,0 +1,88 @@
...
+=====
+
+Just install the addon in your instance. You will be able to access the logs
+under Reporting -> Server Monitoring.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza: Remove these lines

server_monitoring/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Usage
+=====
+
+Just install the addon in your instance. You will be able to access the logs
+under Reporting -> Server Monitoring.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
pedrobaeza: Replace {repo_id}/{branch} with the proper information

server_monitoring/README.rst
```diff
@@ -0,0 +1,88 @@
...
+* enhance group by for process logs. 
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/{project_repo}/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/{project_repo}/issues/new?body=module:%20{module_name}%0Aversion:%20{version}%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: replace placeholders for their values

server_monitoring/README.rst
```diff
@@ -0,0 +1,88 @@
...
+* check if we can / wish to log session id
+* check if we can log pooler state (whatever that is...)
+* check if we can log HTTP status
+* enhance group by for process logs. 
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/{project_repo}/issues>`_.
```
pedrobaeza: Replace placeholder by server-tools

gurneyalex:
expecting 3 :green_heart: CI checks anytime now

gurneyalex:
I'll port this addon to 8.0

server_monitoring/README.rst
```diff
@@ -0,0 +1,89 @@
...
+=====
+
+Just install the addon in your instance. You will be able to access the logs
+under Reporting -> Server Monitoring.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/7.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza: Remove these lines, as they are guidelines for the template

server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,63 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Server Monitoring',
+    'version': '0.1',
+    'category': 'Connector',
```
pedrobaeza: Why connector category?

pedrobaeza:
Little remark, but :+1: functional and technical

max3903:
:+1:

moylop260:
:+1:

PR #42, opened 2014-08-22
7.0 Add email_template_dateutil

veloutin:
This module adds a format_date filter in the Jinja2 environment for mail templates, allowing you to do
${object.invoice_date|format_date("%m/%d/%Y")} to have the date in the format you want, and converted into the user timezone.

### DIFF
--- a/email_template_dateutil/__init__.py
+++ b/email_template_dateutil/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2014 Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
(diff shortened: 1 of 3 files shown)


hbrunn:
That's a great module! But I'd suggest to have the format as optional parameter and default it to the email's language's date format.

max3903:
I agree with hbrunn, otherwise :+1:

veloutin:
@hbrunn, I've implemented your suggestion. 

Also, this no longer calculates an offset for dates, since the time zone offsets are only meaningful for datetimes. Any negative offset for dates would result in the previous day.

hbrunn:
:+1:  (travis errors are unrelated)

pedrobaeza:
I think you shouldn't remove timezone offset, because it's also meaningful in dates: if you are in a border zone (for example UTC+9), you'll get a different day when you compose a mail at central hours if you don't make the offset.

veloutin:
@pedrobaeza We have the opposite problem when setting an offset on a date. In all the timezones west of UTC, any offset on a date will be the previous day.

I don't believe having an offset on a date (without time) makes any sense. PostgreSQL does not have a date with time zone type either. If we instead had fields that show the date of a datetime with time zone, then it would make sense.

veloutin:
Travis failures still unrelated.

pedrobaeza:
:+1:

PR #44, opened 2014-08-26
Add sentry integration module

bwrsandman:
Module for enabling raven initializing to report to sentry.

Based on feedback from my presentation at odoo days 2014

Related project: https://github.com/gisce/openerp-sentry
## Todo
- [x] Add git module version overwrite to raven
- [x] Overwrite the except_orm message pop-up with something a little more friendly to the user
- [x] Don't log `SessionExpiredException`
- [x] Don't log `TransactionRollbackError` ?
- [x] Do a better job at sanitizing the passwords (openerp/netsvc.py in dispatch_rpc )
  - Maybe exclude dispatch_rpc from the traceback
- [ ] ~~Enable tests with sentry server in travis~~

### DIFF
--- a/sentry_logger/__openerp__.py
+++ b/sentry_logger/__openerp__.py
@@ -33,6 +33,12 @@ Integration with Sentry Error reporting engine.
 Insert sentry DSN to ~/.openerp_serverrc with value:
     sentry_dsn = sync+<Your Sentry DSN>
 
+
+Optional Dependencies
+---------------------
+
+* bzrlib (for revno reporting on module repositories from LP)
+
 Contributors
 ------------
 * Sandy Carter (sandy.carter@savoirfairelinux.com)
@@ -49,7 +55,11 @@ Contributors
     # any module necessary for this one to work correctly
     'depends': ['web'],
     'external_dependencies': {
-        'python': ['raven', 'raven_sanitize_openerp', 'bzrlib', 'git'],
+        'python': [
+            'raven',
+            'raven_sanitize_openerp',
+            'git',
+        ],
     },
     'data': [
     ],
--- a/sentry_logger/__openerp__.py
+++ b/sentry_logger/__openerp__.py
@@ -39,6 +39,12 @@ Optional Dependencies
 
 * bzrlib (for revno reporting on module repositories from LP)
 
+Notes
+-----
+
+Module may slow down spawning of workers (~1s) due to the access of openerp
+configuration
+
 Contributors
 ------------
 * Sandy Carter (sandy.carter@savoirfairelinux.com)
--- a/sentry_logger/__openerp__.py
+++ b/sentry_logger/__openerp__.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2010 - 2014 Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': "sentry",
+
+    'summary': "Sentry integration",
+    'description': """
+Sentry
+======
+
+Integration with Sentry Error reporting engine.
+
+Insert sentry DSN to ~/.openerp_serverrc with value:
+    sentry_dsn = sync+<Your Sentry DSN>
+
+Contributors
+------------
+* Sandy Carter (sandy.carter@savoirfairelinux.com)
+""",
+
+    'author': "Savoir-faire Linux",
+    'website': "http://www.savoirfairelinux.com",
+
+    # Categories can be used to filter modules in modules listing
+    # Check <odoo>/addons/base/module/module_data.xml of the full list
+    'category': 'Extra Tools',
+    'version': '1.0',
+
+    # any module necessary for this one to work correctly
+    'depends': ['base'],
+    'external_dependencies': {
+        'python': ['raven', 'raven_sanitize_openerp', 'bzrlib', 'git'],
+    },
+    'data': [
+    ],
+
+    'demo': [
+    ],
+
+    'tests': [
+    ],
+}
--- a/sentry_logger/__openerp__.py
+++ b/sentry_logger/__openerp__.py
@@ -21,7 +21,7 @@
 ##############################################################################
 
 {
-    'name': "sentry",
+    'name': "Sentry Logger",
 
     'summary': "Sentry integration",
     'description': """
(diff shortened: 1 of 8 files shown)


sentry/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+from openerp.tools import config
+from odoo_sentry_client import BzrClient
```
nbessi: from .odoo_sentry_client maybe better

sentry/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+from odoo_sentry_handler import OdooSentryHandler
```
nbessi: from .odoo_sentry_handler

sentry/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+import cgitb
+
+
+root_logger = logging.root
+
+processors = (
+    'raven.processors.SanitizePasswordsProcessor',
+    'raven_sanitize_openerp.OpenerpPasswordsProcessor'
+)
+if config.get(u'sentry_dsn'):
```
nbessi: Is this code executed even if addons is not installed ? if yes it needs fixing

sentry/__openerp__.py
```diff
@@ -0,0 +1,62 @@
...
+
+    'summary': "Sentry integration",
+    'description': """
+Sentry
+======
+
+Integration with Sentry Error reporting engine.
+
+Insert sentry DSN to ~/.openerp_serverrc with value:
+    sentry_dsn = sync+<Your Sentry DSN>
```
nbessi: This is a little too short at least information what to see and what to do and what will the addons exactly do.
For what I understand it will redirect Odoo exception to configured Sentry server. I will also add sample of parameter in an Odoo configuration file

sentry/odoo_sentry_client.py
```diff
@@ -0,0 +1,104 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+import logging
+import platform
+import os
+import sys
+from raven import Client
+from bzrlib.branch import Branch
```
nbessi: Please protect the python import with except log block. It may break server start-up if dependencies are not installed on host.

sentry/odoo_sentry_client.py
```diff
@@ -0,0 +1,104 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class BzrClient(Client):
+    """Subclass raven.Client to be able to report Bazaar revno as Module
+    versions"""
+
+    def __init__(self, dsn=None, **options):
+        # Send the following library versions, include all eggs
```
nbessi: I did not understand the docstring

sentry/odoo_sentry_client.py
```diff
@@ -0,0 +1,104 @@
...
+        self.bzr_revs = {}
+        super(BzrClient, self).__init__(
+            dsn=dsn, include_paths=include_paths, tags=tags, **options)
+        self.set_rev_version(config.get(u'root_path') + u"/..", )
+        # Create and test message for Sentry
+        self.captureMessage(u'Sentry Tracking Activated!')
+
+    def set_rev_version(self, path):
+        """Given path, get source and revno, careful not to raise any
+        exceptions"""
```
nbessi: I would add because it is used in error management

nbessi:
Nice addition. It would be nice to add a command line argument too.
Some comment in line

bwrsandman:
Thanks for the review @nbessi, I will get to it eventually :tired_face:

nbessi:
@bwrsandman I did not dare to talk about unit test :sweat:

gurneyalex:
tests are red because of missing external dependency, unhandled in the imports. This must be taken care of because the sole presence of the addon in the repository without the `sentry` python module will cause a crash at instance startup.

nbessi:
@gurneyalex this was allready pointed as a need fixing point in sentry/odoo_sentry_client.py line 27,28

bwrsandman:
@nbessi, @gurneyalex, Idealy, I would like to be able to overwrite the error message displayed by the client.

Instead of displaying a traceback to the user, it would be a lot better to show a message saying that an error occured and was sent to sentry.

Do you have an idea how I can do this?

sentry/odoo_sentry_client.py
legalsylvain: Is it necessary to try / except with that statement in __ openerp __ file ?

```
'external_dependencies': {
'python': ['raven', 'raven_sanitize_openerp', 'bzrlib'],
},
```

bwrsandman: The reasoning behind that was to never have sentry cause exceptions.

In the event that you have the module installed, but for some reason or another, one of the python dependencies disappears or causes an error, we don't want sentry itself to start causing crashes.

legalsylvain: How, I think that it is necessary to change the travis file project, to install extra lib, that could not be present on Travis System.
Like in my PR: https://github.com/OCA/server-tools/pull/50/files#diff-0

This should fix Travis: https://travis-ci.org/OCA/server-tools/jobs/33630905#L310

Don't you think ?
Regards.

legalsylvain:
@bwrsandman, maybe that feature should be integrated to another module like:
'web_display_friendly_message' in https://github.com/OCA/web
It' not just a need of your module I think.
What do you think ?
I could work on that if you want but in a while.

Do you need help for other think ? I'm very interested by your module and Sentry system, so if I can help for something.

sentry/__init__.py
bwrsandman: Sorry I don't understand your objection here

legalsylvain:
Hi @bwrsandman, 
Thanks for all the changes. (Friendly messages to users, ...) 
I planned to review this PR with @arthru, realizing complete tests with sentry service installed in one or two weeks.
I'll let you know.
Regards.

sentry_logger/__openerp__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Categories can be used to filter modules in modules listing
+    # Check <odoo>/addons/base/module/module_data.xml of the full list
+    'category': 'Extra Tools',
+    'version': '1.0',
+
+    # any module necessary for this one to work correctly
+    'depends': ['web'],
+    'external_dependencies': {
+        'python': ['raven', 'raven_sanitize_openerp', 'bzrlib', 'git'],
```
kryskool: Hi

Please removed "bzrlib" and "git" dependancies on **openerp**.py, because i only use git, and convert bzr repo with fast-export and fast-import, and i cannot install this module. The protection on the odoo_sentry_client.py file is adequacy to check if bzrlib and git is installed

Regards,

bwrsandman: The code should work with and without git, bzr python libraries, so I see nothing wrong with having these be optional dependencies.

bwrsandman: I set bzrlib as optional, but let git in, since this repo is git and we would like to always have sentry_logger report the version of itself.

bwrsandman:
Raven no longer logs passwords. Note that I had trouble with python-raven 5.0.0, using 5.1.1 seems to work.

legalsylvain:
Could you rebase ? Travis fails for fixed pylint errors.

Regards.

bwrsandman:
@legalsylvain not rebased, but tests restarted

legalsylvain:
- Code Reviewed.
- Tests realized with sentry server.
  :+1: 
  All is OK. Thanks a lot @bwrsandman.

arthru:
:+1:

Reviewed, tested, works like a charm !

bwrsandman:
Small note and something to test: it impacts performance of spawning workers.

Since each worker fetches the config, it can take up to a second more for workers to spawn.

legalsylvain:
Can you add this note in the openerp file ?

bwrsandman:
Added

pedrobaeza:
:+1:

bwrsandman:
Thank you for the review, @legalsylvain  and @arthru. I hope this module serves you as well as it has for me!

PR #48, opened 2014-09-09
Port of server_environement

nbessi:
- Port of server_enviroment
- Renaming of server_enviroment_files to server_enviroment_files_sample
- Support of git in system info


gurneyalex:
LGTM :+1:

server_environment/serv_config.py
```diff
@@ -123,18 +128,58 @@ def __init__(self, pool, cr):
...
+            self._conf_defaults[col_name] = value
+
+    def _get_base_cols(self):
+        """ Compute base fields"""
+        res = {}
+        for col, item in system_base_config.options.items():
+            key = self._format_key('openerp', col)
+            # fld = fields.Char(string=col, readonly=True)
+            # fld.name = key
+            # fld.model_name = self._model
```
yvaucher: code to remove if deprecated

yvaucher:
appart from the commented lines :+1:

server_environment/tests/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import test_server_environment
+checks = [test_server_environment]
```
lepistone, line 22: Are you confident creating a new configuration won't break an existing database? If not, we can change that to 'fast_suite'

nbessi, line 22: ?? I did not get your point @lepistone I do not create any new configuration I use the sample one. Please have a look in travis file

lepistone, line 22: I admit I don't know what's going on in detail, but I see in `test_default` a `model.create()`. Since checks should be safe at any time in the lifetime of a database, I was wondering whether fast_suite would be safer (it's executed only when you create the database).

Still, maybe I'm worrying too much, and we're in an uncommitted transaction anyway.

Thanks

lepistone:
My comment above is not a blocker for me. thanks :+1:

PR #50, opened 2014-09-11
[7.0] 'module_repository'  new. Informations of Modules Repositories

legalsylvain:
Hi all,

I propose this modules to see faster wich repositories are used on a Odoo Database (url, name, revision), if there is local (ie = unstaged / uncommited) modifications, etc.
See screenshots to see covered features.

Thanks for your review.

**Remarks:**
_Flake8_ : OK;
_Coverage_: ~75 - 90 % (depending of your version controls during the tests)

To get informations about source version, I used 'git' and 'bzrlib' librairies. I'm not very confortable with that libraries. If some reviewers know better that librairies please propose improved code.

```
def _read_bazaar_information():
def _read_git_information():
```


legalsylvain:
It seems that Travis is pretty lost with my extra-dependencies of 'git' and 'bzrlib' librairies.
https://travis-ci.org/OCA/server-tools/jobs/34999409#L294

```
except_orm: ('Error', u'Unable to install module "module_repository" 
because an external dependency is not met: No module named git')
```

However, I wrote in __ openerp __.py: 

```
'external_dependencies': {
    'python': ['git', 'bzrlib'],
},
```

@bwrsandman: Did you ever had this problem ? How can I fix it ?

Regards.

bwrsandman:
Git is not a python library. Bzr is in python so it works there.
This should work, though:

```
'external_dependencies': {
    'python': ['bzrlib'],
    'bin': ['git'],
},

```

legalsylvain:
Hi @bwrsandman,
Thanks for your answer !

It doesn't work. In fact I guess 'git' is a python librairy. To install it, i just ran:

```
sudo apt-get install python-git
```

After that, I have a folder '/usr/lib/pymodules/python2.7/git/' with the python code and the following import works: 

```
import git
```

I tried: 

```
'bin': ['git'],
```

and 

```
'bin': ['python-git'],
```

But both failed.

Do you have any idea ?

Regards.

bwrsandman:
If you want to enforce a python library use:

```
'python': ['some_python_lib'],
```

If you want to enforce some binary in UNIX `$PATH` use

```
'bin': ['an_executable_in_PATH']
```

It seems to me you're trying to make sure git and bzr are installed. For example, you're calling them with `subprocess.Popen()` Those are two executables that you call like this:

``` bash
$ which git
/usr/bin/git
$ which git
/usr/bin/git
```

They should therefore go in `bin`:

```
'external_dependencies': {
    'bin': ['git', 'bzr'],
},
```

If they happen to be python libraries that youre calling by importing the libraries in python with `import gitlib, bzrlib`, then they are python dependencies and you should use:

```
'external_dependencies': {
    'bin': ['python-git', 'bzrlib'],
},
```

gurneyalex:
@bwrsandman since when does the tools used in travis configuration process the external_dependencies entry of `__openerp__.py`? I thought these had to be added manually in `.travis.yml`

dreispt:
Yes, @gurneyalex is right. TravisCI does not know about `external_dependencies`.

bwrsandman:
Of course not... 

You need to install them in the pre-install section with either apt-get or pip for bin and python external_dependencies respectively.

Sorry for the confusion

legalsylvain:
OK! 
Basically: 
1/ I use python lib (import git, import bzrlib) but I install them by sudo apt-get install python-git python-bzrlib (because there is no pip install)
- I put python dependencies in _ _openerp_ _ file; ('git', 'bzrlib')
- I put apt-get install in travis file; ('python-git' 'python-bzrlib')

2/ I realize an extra call by Popen to have bzr status;
- I put bin dependency in _ _openerp_ _ file; ('bzr')
- I put apt-get install in travis file; ('bzr')

And It finally passed on Travis.
@gurneyalex @bwrsandman @dreispt : Thanks for your remarks.

bwrsandman:
> - I put python dependencies in _ openerp _ file; ('git', 'bzrlib')

``` python
['python-git', 'python-bzrlib']
```

legalsylvain:
Maybe I'm wrong but I don't think so. The librairies names are 'git' and 'bzrlib'. (although i use apt-get install python-xxx to install them)
- If I write 'python': ['python-git', 'python-bzrlib'] --> ERROR
- If I write 'bin': ['python-git', 'python-bzrlib'] --> ERROR
- If I write 'python': ['git', 'bzrlib'] --> SUCCEED

bwrsandman:
Install them via `sudo pip install git bzrlib`, then it is consistent.

legalsylvain:
Yes I tried, but, I have this result. Did I miss something ? Sorry if my question is not relevant. 

```
>> sudo pip install git
Downloading/unpacking git
  Could not find any downloads that satisfy the requirement git
No distributions at all found for git
>> sudo pip install bzrlib
Downloading/unpacking bzrlib
  Could not find any downloads that satisfy the requirement bzrlib
No distributions at all found for bzrlib
```

legalsylvain:
Hi @bwrsandman, 
Thanks to @arthru, I finally changed Travis file using pip install GitPython bzr.
Can you say if it's now ok for you ? Thanks.
Regards.

.travis.yml
```diff
@@ -14,7 +14,8 @@ install:
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly ${VERSION}
-  - sudo pip install python-ldap
+  - sudo pip install python-ldap GitPython bzr
+  - sudo apt-get install bzr
```
bwrsandman, line 19: :+1:

bwrsandman:
:+1: LGTM, no test

.travis.yml
legalsylvain, line 19: Glad to see that you're agree and thanks for the help.

bwrsandman:
The name of the package is `GitPython`, the module is called `git`

``` bash
$ cd /tmp
$ virtualenv2 env
$ source env/bin/activate
$ pip install GitPython
$ python -c "import git"
$ python -c "import imp; print imp.find_module('GitPython')"
  File "<string>", line 1, in <module>
ImportError: No module named GitPython
$ python -c "import imp; print imp.find_module('git')"
(None, '/tmp/env/lib/python2.7/site-packages/git', ('', '', 5))
```

gurneyalex:
needs rebase

PR #51, opened 2014-09-11
FIX onchange method name on mass editting module

jjscarafia:
on views method is called by "onchange_model_id", on mass_object.py was called "onchange_model" and now renamed to "onchange_model_id"


rvalyi:
nice catch, indeed: https://github.com/ingadhoc/server-tools/blob/8.0/mass_editing/views/mass_editing_view.xml#L18

:+1:

pedrobaeza:
Thanks for the fix!

:+1:

guewen:
:+1:

PR #53, opened 2014-09-13
OpenERP / Odoo Module which help you to set the database for Test or Development mode.

bizzappdev:
## Features:

```
* Set-up Test or Development environment at the level of database.
* Provides unique mode-bar for notifying either database is in Develop mode or Test mode.
* Mail restriction for outgoing mails.
* Set default test/develop environment for the system With changing TEST and DEVELOP Variable to True
```

**Also for more detail documentation you can check READ.md**


oerp_develop_test_mode/controllers/main.py
```diff
@@ -0,0 +1,117 @@
...
+        if not db:
+            dbs = db_list(req)
+            db = dbs and dbs[0] or False
+        return config.get(mode, {}).get(db, False)
+
+
+class cp_Binary(Binary):
+    _cp_path = "/web/binary"
+
+    @openerpweb.httprequest
```
nhomar, line 92: Hello.

If you move this logic at model side, this module will be almost functional in production, writting the right models, like the ir.ui.view you can achieve this behavior (I think).

Write the controllers for technical logic can introduce several bugs in the client behavior (due to the young of the framwork) and the framework in v8.0 changed "again" at this point.

You should mix this with some tricky js helpers (but JS is always more secure I think).

nhomar:
Extraordinary work I think, this PR will be pulled at some moment or we are waiting for the 0.26% less?

I can merge it and fix the details I think.

for me it is a +1 (not functionally tested yet, and with some TODO -commented inline-)

regards.

oerp_develop_test_mode/controllers/main.py
bizzappdev, line 92: Hello Mr. @nhomar  .
I this a logic for getting Image of Logo to display above the Menu.
This is from core OpenERP. I have just inherited it and has changed the logic to fetch test/development image when it is in appropriate mode.

pedrobaeza:
Why are you closing this?

We don't want to discourage you from contributing with the comments in the other PR, but let you know some concerns about the module.

PR #54, opened 2014-09-13
Super Admin users as member of all group by default

bizzappdev:
## Features:

```
  * Include Super Admin user to all Group.
  * Make Installed module as first menu in settings menu.
```


super_admin/base.py
```diff
@@ -0,0 +1,63 @@
...
+                    if kind == 'selection':
+                        self.write(cr, SUPERUSER_ID, [SUPERUSER_ID], {
+                            name_selection_groups(map(int, gs)): gs[-1].id})
+                    else:
+                        for g in gs:
+                            self.write(
+                                cr, SUPERUSER_ID,
+                                [SUPERUSER_ID],
+                                {name_boolean_group(g.id): True})
+                except:
```
hbrunn, line 34: That looks scary. What kind of error do you expect and why is it okay to ignore it?

super_admin/base.py
```diff
@@ -0,0 +1,63 @@
...
+                    continue
+        except:
```
hbrunn, line 36: same as above

super_admin/base.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class module(osv.osv):
+    _inherit = "ir.module.module"
+
+    def button_immediate_install(self, cr, uid, ids, context=None):
+        res = super(module, self).button_immediate_install(cr, uid, ids,
+                                                           context=context)
+
+        for app, kind, gs in self.pool.get(
```
hbrunn: Given that you do the same thing here than in line 19ff, make a function from this

super_admin/base_view.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hbrunn, line 1: All of those are convenient in my opinion, but how do they relate to the module's purpose?

super_admin/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+##############################################################################
+{
+    'name': 'Super Admin',
+    'version': '1.0',
+    'category': '',
+    "sequence": 14,
+    'complexity': "easy",
+    'category': 'Hidden',
+    'description': """
+        Mark All Super Admin users as member of all group by default
```
hbrunn: This is a quite misleading description. From this, I expected all members of base.group_system to be made member of all groups. This would enhance the module's value enormously I think and isn't very hard to implement.

hbrunn:
And as a general remark, I think there's agreement in the community that OCA modules shouldn't have a company logo as their icon

super_admin/__openerp__.py
bizzappdev: Sorry It is just a Superadmin . uid =1 not all members of base.group_system
changing desc.

super_admin/base.py
bizzappdev, line 34: It is getting error when trying to access Group from share module. 
As from share module it is adding "Share" fields. and before returning. function it will not have cr.commit. so it is having problem to add share group for super admin

super_admin/base_view.xml
bizzappdev, line 1: Because We this module is mostly for the developers who is normally developing the modules and they want to go to settings menu every often. so it is annoying for have App menu at first  position. and as a developer Most time we do not have to do any thing in apps menu we are concern about installed module menu .
That is why changed the position of menu.

super_admin/base.py
bizzappdev: Converted to function

super_admin/__openerp__.py
bizzappdev: changed desc.

bizzappdev:
Image file has been removed.

hbrunn:
now travis complains about something else:
./super_admin/base.py:15:1: E303 too many blank lines (3)

Best to get flake8 locally and run it yourself, that simplifies everything

bizzappdev:
fixed pep8

dreispt:
Can you explain the rationale for this module?
Maybe I'm missing something, by I find that assign all groups to the admin is not desirable, and could even break some UI tweaks that rely on groups.
Also for this reason I find troubling the `'auto_install': True`.

pedrobaeza:
Indeed, there are some groups like portal or public that restricts the access to certain records.

bizzappdev:
Hello !
@dreispt Rational use of this module is it will set all group to admin users. so while developing we do not have to go to user access management and manually give rights to every one. Also some times we do not know about some fields which are already there but hidden because of not have group assigned and we are struggling or adding new fields for the same reason.
@dreispt and @pedrobaeza 
As far As UI Tweaks is concern this is at level of development database So even better to know in advance that even what we are developing in future if we will enable group in production there will be a UI Problem or so ..
Also As far as Restriction and access concern . OpenERP/Odoo record rules are working like .
If we have 2 different group assign to same user and both group has its Record rule for one object/model then it will have OR Condition. So we have always other group assigned to admin user which have proper access rights.

**By The Way For Super admin user There is no restriction for access rights or record rules. So it is not a topic to concern about.**

eLBati:
Hello @bizzappdev then I suggest to put these remarks in the module
description.

dreispt:
I see your case. I'm not sure it's the best solution for the problem, but I grant that different persons can prefer different solutions.

Now, the `auto_install` should absolutely be removed.

And I'm also concerned that the module effect is not undoable: if you install it and regret it, you get stuck with the changes made. I suggest you a solution for that: create a special group "All Permissions", assigned by default to the Admin user, and use it to assign all existing groups. With this a module uninstall would rollback the changes.

kryskool:
Hi

For me **auto_install** must be set to False

All my clients have multi-company installn and user admin must be only used on Administration Parts, because admin user bypass check rules [see method check() on ir.model.access](https://github.com/OCA/OCB/blob/7.0/openerp/addons/base/ir/ir_model.py#L687)

Regards,

foutoucour:
:-1: 
There is not relation between the rights of the admin and the layout of the menu of settings.
I can't accept double blind try/except pass. That would be a bomb that can't be debugged. Find the error that can be raised and informed through the logger what's happening.

bizzappdev:
@foutoucour  it is not in-between user rights and menu . it is regarding the Gourps and the users. 
Lets have an example Share module have a field "share" added from share module in the res.groups. 
while my code is executed . cr.commit is not done so share field is added to the groups but in database field is not create. 
So while reading the res.groups it is having a problem. just for the share module .
this is the main reason for putting those code in Try and except.

super_admin/base.py
bwrsandman, line 34: You shouldn't have unspecified catches, please add the error you're guarding against.

pedrobaeza:
I don't think we should include this module like I say on https://github.com/OCA/server-tools/pull/54#issuecomment-55708346.

:-1:

bizzappdev:
@pedrobaeza this is my comments to your comment on another PR
Of Course I am disappointed Not because of negative comments but Because they way people are commenting out of the box problem . Like 
1. There is no relation between Menus and user group rights. 
2. Super admin will have a problem because of the restriction from the group.

My point is When you are giving your opinion you have to be sure or be proper. 
Like Someone suggested to have All Permission Group that was really nice idea. But again I am stuck . Whether to go ahead with that idea and implement it or not. 
No one is there to direct the things are going on.Every one is giving their opinion that is good but There should be some one/people to deiced or go in which direction.

Like some one raise a question that "There is not relation between the rights of the admin and the layout of the menu of settings." And any one who is just using openerp can understand that there is a relation between user rights and menu layout. 

For the rights problem question. https://github.com/odoo/odoo/blob/7.0/openerp/osv/orm.py#L3987 it is explicitly written that if uid == SUPERUSER_ID: return.

My point is if commenting at-least negative you have to be sure or more explained. 

Also another major problem is that it is taking much more time . Like https://github.com/OCA/server-tools/pull/53 it is there since almost 2 month . within this time period odoo might have new release ;)  I can understand that everyone is contributing in free time. and always not possible to have free time. but for those who are managing at-least they can escalate to someone else we have big community try to include more people.
Community is always to work to gather. Not like working individually. 
Also I would like to add when something is taking much more time then it is a human mentality that they will loss interest in that.

pedrobaeza:
I regret hear that, @bizzappdev, but I don't see here reasons for your discouragement, because the debate in this PR follows the rules of an usual peer review of an open-source community: people pose doubts, questions or possible problems, and the commiter defends current code or rectifies it depending on the discussion.

I think that reasoning like mine that some groups restricts permissions instead of giving them are well funded. If I'm wrong, you can explain me why, and I will be very happy to rectify my position.

Other reviewer (or me in some occassions - more as the review hour progresses, hehe) can say a nonsense, or even stay in it although you argument the opposite, but at the end, the important is that enough reviewers give their approval.

I hope you still be interested on contributing, because you have made interesting module through this time, and don't be too upset by these little rules whose function is only to assure the quality and maintainability of the contributed code, which is good for everyone. Think in this as a free code audit.

Regards

dreispt:
@bizzappdev One of the reasons this is taking two much time is because some reviewer concerns are yet to be addressed.
One major blocker is the `auto_install=True` mentioned since 16 Sep and by more than one reviewer. I don't see how this could be merged without this issue being addressed.

bizzappdev:
Okay ... @pedrobaeza  For the access rights problem. I have check from my side I am not getting any access permission error .  I can totally understand the approval and the strictness to keep code and modules are has to be proper.

@dreispt  yes you are right I have to respond to auto_install=True . But my question for taking more time is not related to this PR it was related to #53

pedrobaeza:
@bizzappdev, good to know that there is no problem with my commented issue.

About auto_install, please remove it and you will get my approval.

PR #55, opened 2014-09-14
[FIX] renamed 'email_template' m2o field in 'email_template_id' in sched...

bealdav:
in scheduler_error_mailer module

### DIFF
--- a/scheduler_error_mailer/__openerp__.py
+++ b/scheduler_error_mailer/__openerp__.py
@@ -1,11 +1,10 @@
 # -*- encoding: utf-8 -*-
 ##############################################################################
 #
-#    Scheduler Error Mailer module for OpenERP
 #    Copyright (C) 2012-2013 Akretion (http://www.akretion.com/)
-#    @author: Sébastien Beau <sebastien.beau@akretion.com>
-#    @author David Beal <bealdavid@gmail.com>
-#    @author Alexis de Lattre <alexis.delattre@akretion.com>
+#    @author: Sébastien Beau
+#    @author Alexis de Lattre
+#    @author David Beal
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Why renaming it? Nowadays, I'm preferring notation without '_id', because most of the time you use dot notation for accesing subfields, and it gets weird with that name. For example, _order.partner.name_ instead of _order_id.partner_id.name_. It's also more weird if you do _order_id.id_, because there are two 'id' very close each other. The only use I can see is for distinguishing relation field (many2one, many2many, etc) from other fields, but having a good field name, you would know.

dreispt:
I disagree with Perdo: IMO the `_id`  and `_ids` are useful conventions.
But changing a field name will break existing installations so it should be avoided.
It would be best to do that when porting to v8.

bealdav:
Hi Pedro, 
99% of m2o fields are my_field_id.
I don't know what name I prefer but I can't do 5000 PR to rename all m2o without id, don't you think ?
I prefer consistency even if could have an idea about convention choice. 
When I see onename_id it's easy to me to think 'foreign key' and foreign keys implies to have a particular attention to them in coding.
Without 'id' I think char or text if no string 'date' in name
Thanks

nbessi:
Hello, 

Thanks for the patch.

I agree with @dreispt 

For me there is a kind of tacit convention _id means m2o _ids mean o2n m2m .
But anyway, changing a column name, if it does not fix something should not be accepted until module migration as we do not want to break other devs.

:-1:

bealdav:
Hello, 
Is it Ok if I switch my PR to 8.0 ?

pedrobaeza:
I'm not talking of renaming all of the existing fields, but to keep the non '_id' fields, and taking into account for future developments.

dreispt:
@bealdav Renaming fields is acceptable when porting to 8.0

pedrobaeza:
...but providing the corresponding migration script.

bealdav:
Is 'oldname' attribute is accepted in field definition ? Included in this proposal.
Framework provide this feature, it works and field will never renamed, isn't it ?
Thanks

pedrobaeza:
OK, I think 'oldname' it's handled by standard upgrade machinery, but I abstain, as I say about my preferences.

hbrunn:
`oldname` is fine to use for that purpose, but as the previous posters wrote, renaming a column without necessity is a very bad idea (think of customer modules depending on this one). Please do that when porting to 8.0. Meanwhile, I close this PR.

bealdav:
Ok, 
You're right Holger. As I said, I'll switch my PR to 8.0

PR #56, opened 2014-09-15
7.0 add search enhanced operators

OSguard:
Allows =ilike operator to advanced search option.
Use % as a placeholder. 
Example: "Zip matches 1%" gives all zip starting with 1

### DIFF
--- a/search_enhanced_operators/__openerp__.py
+++ b/search_enhanced_operators/__openerp__.py
@@ -30,7 +30,7 @@
     'license':  'AGPL-3',
     "description": """
     Allows =ilike operator to advanced search option.
-    Use % as a placeholder. 
+    Use % as a placeholder.
     Example: "Zip matches 1%" gives all zip starting with 1
     """,
     'data': [
--- a/search_enhanced_operators/__openerp__.py
+++ b/search_enhanced_operators/__openerp__.py
@@ -26,10 +26,12 @@
     "depends":  ["web"],
     'author':   'initOS GmbH & Co. KG',
     "category": "",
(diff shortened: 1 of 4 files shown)


hbrunn:
very useful, thanks! But I would put this module rather into https://github.com/OCA/web than here.

Also, even though I admit doing the same from time to time, words like 'enhanced' are imho a poor choice for a name, better describe the specific enhancement, like advanced_search_wildcards or something like that?

It might be worth mentioning in the module's description that this also allows case insensitive exact searches, so 'John' matches the input 'john', but not 'Johnson'.

OSguard:
The reason to add it here, is that is a simular improvement like "mass_editing" and combine they are really greate. But i can also move to the web repository, it depends on 'web'.

So for the new name i think: web_advanced_search_wildcard would be fit.

bealdav:
Maybe I missed something but without this module, in products
if I do 'bou%e' in search box
I get 'bougie' and 'bouge'

The same in partner with zip

It works ok with no module.

What is the difference ?

bealdav:
Search box has an item named 'advanced search'. Is it only for this search or classic search too ?

OSguard:
bou%e will be transformed to %bou%e%

if you search 3% in zip you get all data with 3 inside zip (so %3% will the search), but i want to search only witch starts with 3.

And yes it only adds an option for the text dropdown in the "advance search" of the search box, classic is not affected.

bealdav:
thanks for these clarifications

OSguard:
PR moved https://github.com/OCA/web/pull/38

PR #57, opened 2014-09-15
8 email tpl field fix

bealdav:
Renamed 'email_template' m2o field in 'email_template_id' in scheduler_error_mailer module

Refused for v7, but ok for v8

https://github.com/OCA/server-tools/pull/55

Thanks for your review

### DIFF
--- a/scheduler_error_mailer/ir_cron.py
+++ b/scheduler_error_mailer/ir_cron.py
@@ -35,9 +35,9 @@ class ir_cron(orm.Model):
     _inherit = "ir.cron"
 
     _columns = {
-        'email_template': fields.many2one(
-            'email.template',
+        'email_template_id': fields.many2one('email.template',
             'Error E-mail Template',
+            oldname="email_template",
             help="Select the email template that will be "
                  "sent when this scheduler fails."),
     }
@@ -50,7 +50,7 @@ class ir_cron(orm.Model):
 
         my_cron = self.browse(cr, uid, job_id)
 
-        if my_cron.email_template:
+        if my_cron.email_template_id:
             # we put the job_exception in context to be able to print it inside
             # the email template
             context = {
--- a/scheduler_error_mailer/ir_cron.py
+++ b/scheduler_error_mailer/ir_cron.py
@@ -35,7 +35,8 @@ class ir_cron(orm.Model):
     _inherit = "ir.cron"
 
     _columns = {
-        'email_template_id': fields.many2one('email.template',
+        'email_template_id': fields.many2one(
+            'email.template',
             'Error E-mail Template',
             oldname="email_template",
             help="Select the email template that will be "
--- a/scheduler_error_mailer/ir_cron.xml
+++ b/scheduler_error_mailer/ir_cron.xml
@@ -14,7 +14,7 @@
     <field name="inherit_id" ref="base.ir_cron_view"/>
     <field name="arch" type="xml">
         <field name="doall" position="after">
-            <field name="email_template" />
+            <field name="email_template_id" />
         </field>
     </field>
 </record>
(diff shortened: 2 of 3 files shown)


dreispt:
LGTM :+1:

sylvainc:
LGTM

pedrobaeza:
I'm against the renaming, as I said previously, and it's make more sense without the _id.

bealdav:
@pedrobaeza, as I understand here 
https://github.com/OCA/server-tools/pull/55#issuecomment-55564682
you are ok if others are ok for 'oldname' to replace script migration in this case
Thank you.

pedrobaeza:
I always forget that trick!! Indeed, no migration needed (I have to burn by fire this), but I don't like the renaming, as you know. I'm going to abstain again removing my last comment.

dreispt:
@pedrobaeza The fact is that the "_id" suffix is the common practice. so I don't think we should object. And the migration is assured by the `oldname` attribute.

pedrobaeza:
These are my facts:
- This has been the common practise in the past, but not necessarily now
- 'Common practise' doesn't mean 'best practise', and the redundant `_id` seems odd and not needed, as @rvalyi stated in one mailing list post.
- Let's have this discussion for new models, but keep this one as is.

That's why I say we shouldn't accept this PR.

Regards.

rvalyi:
@pedrobaeza the naming convention chosen by OpenERP SA was extremely unfortunate... But I'm convinced that trying to reverse it now in a non coordinated/planned manner while it's now enforced at 90% in the code would IMHO just increase the mess quite a bit.

So I think it's important we raise awareness this convention sucks and we eventually debate it in the future. I would say, with these kinds of convention, Odoo will never become an attractive web platform for instance. But IMHO there are may be 100 critical design issues that will block it before that one. So I would say let's focus on the other issue we can more easily fix, keep that one in mind and eventually put in on the table we if come later to the conclusion that it's part of the last miles to become a decent framework.

IMHO it's on the contrary very likely that Odoo sticks trapped with legacy features, monolithic framework design and will never become an attractive generic framework. My take on this is: make it acceptable where we can and wrap it in more modern frameworks that care much about the design, for instance behind a JSON API, like in Ooor, use it like a database and move on instead of trying to fix the design against the own will of the official maintainers.

dreispt:
@pedrobaeza The PR does not have any technical issues, does not go against any style guide, and  goes toward common practice in existing codebase. I can understand your opinion on how the common practice should be changed, but IMHO that's not enough argument to reject this PR.

pedrobaeza:
As I stated, I'm abstaining in this case, although I want to tell you my reasons.

achaussende:
OK for me :+1:

sebastienbeau:
Ok for me too :+1:

bealdav:
Noboby to refuse, please merge it

dreispt:
@bealdav Has merge conflicts, needs a rebase.

bealdav:
@dreispt I have resolved conflicts, thanks if you can merge

yvaucher:
@bealdav could you do rebase instead of those multiple merges ?

From my point of view, it would keep this PR cleaner to know what will be added and it keeps the history cleaner as well.

Otherwise :+1:

bealdav:
I have spent a lot of time battling against git and rebase. These dirty merge are the result. Is there a way to merge as it or do i have to spend more time.
I'll do better work next times, surely ;-)

hbrunn:
It seems to me quite a bit went wrong here, I can't even really identify which is the commit meant to be committed here. Could you have another look on this?

lepistone:
For the record, I agree with the opinions being expressed here by @pedrobaeza and @rvalyi: I hate to talk about stuff I dislike as "best practices", "community conventions" and so on. On the other hand, those are not worth the fight at least for now, and this specific one has little impact (wrt monolithic design and testing, for example).

No objection on this PR then.

bealdav:
@hbrunn fixed

bealdav:
It's ok on travis, OK for merge, thanks for the review

PR #58, opened 2014-09-16
Port cron_run_manually

yajo:
The code is much simplified using the new api. It worked in all of my tests.


pedrobaeza:
:+1:

cron_run_manually/ir_cron.py
```diff
@@ -0,0 +1,77 @@
...
+        # Hack the UID
+        old_uid = self.env.uid
+        self.env.uid = self.user_id
+
+        # Execute the cron job
+        try:
+            method(*args)
+        finally:
+            # Revert UID to original
+            self.env.uid = old_uid
```
dreispt: Are you sure this is the proper way? I have the feeling you can create and use a new env, instead of modifying the current one.

yajo: Humm I have done additional tests and does not seem to work anyway. Changing UID with the new api seems to be undocumented.

Any ideas?

pedrobaeza: According this: https://github.com/nbessi/odoo_new_api_guideline/blob/master/source/environment.rst#changing-user, you have to do:

```
self.sudo(user_id)
```

yajo:
Thanks @pedrobaeza, permissions issue fixed.

cron_run_manually/ir_cron.py
```diff
@@ -0,0 +1,68 @@
...
+    def run_manually(self):
+        """Run a job from the cron form view."""
+
+        if self.env.uid != SUPERUSER_ID and (not self.active or
+                                             not self.numbercall):
+            raise exceptions.AccessError(
+                _('Only the admin user is allowed to '
+                  'execute inactive cron jobs manually'))
+
+        _logger.warn(self.__dict__)
```
nbessi: Why log self properties ?

yajo: Yikes, that was for debugging, I'll remove it. Thanks.

nbessi:
Tests are missing, for the rest it is ok for me. Can you add a fast_suite test please.

yajo:
@nbessi No problem. Can you give me the docs for that, please? Thanks.

nbessi:
@Yajo https://doc.odoo.com/trunk/server/05_test_framework/

dreispt:
@nbessi Unfortunately, that link is not very helpful.
@yajo My suggestion is to look at a core module for tests/*.yml files, and create a yaml test for this module based on them.

yajo:
I have spent 2 days messing with bugs that I think are in the test framework itself. I am submitting tests that should work, but don't. If you can point me in the right direction, I'll be glad to fix them.

yajo:
The error is [when creating a fake user](https://travis-ci.org/OCA/server-tools/jobs/35619814#L560), but I copied it from [the account module](https://github.com/odoo/odoo/blob/master/addons/account/test/account_test_users.yml#L4), I don't understand the failure.

yajo:
Sorry, I have to close this to reorganize git tree. I will reopen it soon.

PR #59, opened 2014-09-17
add base_mail_bcc module

OSguard:
Allows to configure a list of email addresses to which a blind carbon
copy (BCC) of all emails is sent.

The BCC email addresses can be configured by the system parameter
mail.always_bcc_to as a comma-separated list of email addresses.

### DIFF
--- a/base_mail_bcc/ir_mail_server.py
+++ b/base_mail_bcc/ir_mail_server.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2010-2013 OpenERP s.a. (<http://openerp.com>).
+#    Copyright (C) 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+#    Author Thomas Rehn <thomas.rehn at initos.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import osv
+from email.Utils import COMMASPACE
+
+
+class IrMailServer(osv.Model):
+    _inherit = "ir.mail_server"
+
+    def send_email(self, cr, uid, message, mail_server_id=None,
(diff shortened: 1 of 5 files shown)


dreispt:
I'm curious about a use case for this.
It would be nice for that to be added to the `__openerp__.py` description.

OSguard:
the reason is simple collecting all emails sented by openerp in a separate mailbox

base_mail_bcc/ir_mail_server.py
```diff
@@ -0,0 +1,57 @@
...
+        config_email_bcc = ir_config_parameter.get_param(cr, uid,
+                                                         "mail.always_bcc_to")
+
+        if config_email_bcc:
+            config_email_bcc = config_email_bcc.encode('ascii')
+            if message['Bcc']:
+                message['Bcc'] += COMMASPACE +\
+                    COMMASPACE.join(config_email_bcc.split(','))
+            else:
+                message['Bcc'] = COMMASPACE.join(config_email_bcc.split(','))
```
sbidoul: According to [1] this adds a Bcc header to the message, instead of replacing it. This may not be what you want to do.

[1] https://docs.python.org/2/library/email.message.html#email.message.Message.__setitem__

dreispt:
Bcc is watching you :wink:

base_mail_bcc/ir_mail_server.py
OSguard: i do not understand this comment? can someone explain me what is wrong?

eLBati: @OSguard , @sbidoul means

> Note that this does not overwrite or delete any existing header with the same name. If you want to ensure that the new header is the only one present in the message with field name name, delete the field first, e.g.:
> 
> ```
> del msg['subject']
> msg['subject'] = 'Python roolz!'
> ```

max3903:
I thought email stuff were to be hosted in http://github.com/OCA/social

tremlin:
v8 Migration in https://github.com/OCA/social/pull/56

OSguard:
:+1: 

@max3903 social has no 7.0 branch. So we keep this here and add BCC to social for 8.0 as @tremlin did already

pedrobaeza:
@codingforfun are you going to continue with this?

tremlin:
@pedrobaeza Is there still something to do here? I am currently addressing the comments in https://github.com/OCA/social/pull/56

dreispt:
@OSguard closing since it seems that this feature is being hosted at OCA/social.

codingforfun:
@dreispt 
As @OSguard pointed out, https://github.com/OCA/social has no 7.0 branch. 
So what to do? Should we start one for this single module? 

There are no open comments on this PR so seams like this was just ready to merge before it got closed.

pedrobaeza:
I'm afraid it's more like no one reviewed it in the past. In other PR for the same feature in 8.0 @rafaelbn has questioned the usefulness of this module, and I have to agree on that. What do you think about this topic? Reference: https://github.com/OCA/social/pull/56

dreispt:
@codingforfun  @OSguard having a module in a repo for 8.0 and for another for 7.0 does not seem very wise.
We can open the 7.0 branch there, but is it relevant today, when we're at v10?

PR #60, opened 2014-09-17
[ADD] base_field_serialized

hbrunn:
This is meant to fix https://github.com/odoo/odoo/issues/1263 (or a part of it).

Note that fields.py is not necessary right now, but probably this code will be removed in some subsequent cleanup and then it's good to have it already in place.

So base_field_sparse should be very similar to this I guess, but I don't have time currently to do that myself.


lmignon:
:+1:  Great! LGTM

base_field_serialized/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import fields
+import base_field_serialized
```
dreispt: from . import

base_field_serialized/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Serialized fields",
+    "version": "1.0",
+    "author": "Therp BV",
+    "license": "AGPL-3",
+    "complexity": "normal",
```
dreispt: Isn't `complexity` deprecated?

dreispt:
Minor comments, but looks great! :+1:

hbrunn:
Thanks, done so

guewen:
:+1:

diggy128:
Is this module somewhere in the repo? Because I cannot find it in it.

(from another repo that matched this code)
Also there is a note in fields.py 
> Note: only plain components allowed.

I have a module that depends heavily on serialized many2many fields.
What exactly does the note above mean?

Thanks in advance.

hbrunn:
@diggy128 we removed this again because core reintroduced serialized fields. The comment basically means that you can only stick JSON serializable data into a serialized field.
What you want is sparse fields, you can take this module (or https://github.com/OCA/server-tools/tree/8.0/field_rrule or https://github.com/OCA/server-tools/tree/8.0/field_char_transformed which are slightly more advanced technically) as an example on how to implement sparse fields on top of serialized fields. This will be a lot of work to do right, I think a good way to implement this is to monkey patch fields to have a slot `sparse` following a similar syntax as the `related` attribute which will be evaluated in `init` to generate `compute` and `inverse` functions to write to the serialized field backing up the sparse field. Then Odoo's machinery takes care of the rest.

PR #62, opened 2014-09-18
Cron run manually

yajo:
Reopening https://github.com/OCA/server-tools/pull/58. Tests will fail. The error is [when creating a fake user](https://travis-ci.org/OCA/server-tools/jobs/35619814#L560), but I copied it from [the account module](https://github.com/odoo/odoo/blob/master/addons/account/test/account_test_users.yml#L4), I don't understand the failure.

In my manual tests, the module works so if you can help me patch the tests, I will be glad to do it. Otherwise, I can remove it and it can be merged without the test.

Thanks and sorry for the noise.


nbessi:
@Yajo It seems that the mandatory field  notify_email is not set on res.partner
Have a look at addons/mail/data/mail_demo.xml

yajo:
@nbessi But one question: When you run tests, automatically all demo data gets installed? Also, I am creating a `res.users`, not a `res.partner`. How do I set a default value for a model that I'm not using?

yajo:
Today, the tests started to work for me with the last commit. I make them in an empty database with the latest upstream nightly package. Is it possible that a fix landed there and Travis did not notice?

pedrobaeza:
I force Travis to rebuild, because it doesn't detect automatically new commits. Let's see the result.

Regards.

dreispt:
This will stay red until #64 is merged.

yajo:
https://github.com/OCA/server-tools/pull/64 is merged, can you trigger travis again please?

pedrobaeza:
@Yajo, you neeed to rebase the PR first. See this for more information:

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

yajo:
Thanks, but again Travis fails where local empty database works.

As you can see, Travis fails when [creating the `res.partner`](https://travis-ci.org/OCA/server-tools/jobs/36570909#L544) for [the `res.users` that is created in the test](https://github.com/OCA/server-tools/pull/62/files#diff-3b48bd4ae4f7c2bd375dde83d9aa6fdaR2). This is an automatic procedure that [is often used in upstream tests](https://github.com/odoo/odoo/blob/b74d830eb299c6d672c58f3b53410dde37608a5f/addons/project_issue/test/issue_users.yml#L4), and it works as is.

I execute all my tests in Docker to ensure I use always empty environments, but of course I can be wrong. Maybe I rebased bad? Do you want me to tell you how to reproduce my tests?

Thanks.

PR #64, opened 2014-09-19
Fix pylint on server_environment

dreispt:

PR #65, opened 2014-09-19
Fix ExternalId error on "disable_openerp_online" module update

dreispt:
Error found here: https://travis-ci.org/dreispt/server-tools/jobs/35732685

### DIFF
--- a/disable_openerp_online/__openerp__.py
+++ b/disable_openerp_online/__openerp__.py
@@ -28,7 +28,7 @@ This module deactivates all bindings to openerp.com that
 come with the standard code:
 
 * update notifier code is deactivated and the function is overwritten
-* apps and updates menu items in settings are removed
+* apps and updates menu items in settings are hidden inside Tools\Parameters
 * help and account menu items in user menu are removed
 * prevent lookup of OPW for current database uuid and resulting
   'unsupported' warning
@@ -40,15 +40,11 @@ come with the standard code:
     ],
     "data": [
         "views/disable_openerp_online.xml",
-        'data/ir_ui_menu.xml',
+        'views/ir_ui_menu.xml',
         'data/ir_cron.xml',
     ],
     "qweb": [
         'static/src/xml/base.xml',
     ],
-    "auto_install": False,
     "installable": True,
-    "external_dependencies": {
-        'python': [],
-    },
 }
--- a/disable_openerp_online/data/ir_ui_menu.xml
+++ b/disable_openerp_online/data/ir_ui_menu.xml
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<openerp>
-    <data>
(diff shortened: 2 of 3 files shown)


dreispt:
Red because of #64.

rvalyi:
:+1:  LGTM

dreispt:
If this can be reviewed and merged before https://github.com/OCA/maintainer-quality-tools/pull/99 it would spare us from a red TravisCI on server-tools.

hbrunn:
I'm fine with this specific changes, but i think it creates the wrong precedent: As of https://github.com/odoo/odoo/blob/master/openerp/tools/convert.py#L294 we'll get a warning every time we try to delete something not existing any more. In effect that means OCA modules can't ever delete anything. Do we really want that? I think we should rather have the tests ignore this case.

pedrobaeza:
I also think that we are changing the meaning of the module: disable_openerp_online, not hiding them in other menu.

dreispt:
@pedrobaeza I don't believe this affects the module's effectiveness.

@hbrunn The error found in the test log is different from that, its: `ValueError: External ID not found in the system: base.menu_module_updates`. Ignoring this particular message could hide actual issues on modules.

I believe that this is caused because the test script first install the module, and only in a second run in tests it. The actual test triggers again the deletes and fails because the deletion target is gone by now.
I tried adding a `noupdate="1"` flag to the XML but it has no effect. I'm not sure if it's because of the `<delete>` tag or because of the `-i` option in the test run.

Could this be a design issue with the test script? If so, how could it be improved?

hbrunn:
@dreispt Ultimately, I think we have to move away from a line based approach and look at the complete error message (that would be 460 through 481 in your log). Then we see that in this case, it's within a warning context which is why it seems safe to ignore it.

I'll prepare a PR for maintainer-quality-tools that allows this, then we can treat warnings emitting this error differently than errors/exceptions.

dreispt:
I see. You could "navigate back" until you find a INFO/WARNING/ERROR "parent" line and have the rules work on both line and parent line text.

hbrunn:
I made my proposal here: https://github.com/OCA/maintainer-quality-tools/pull/102

guewen:
I think the line that makes travis red should not:

```
----------
2014-10-02 09:35:18,759 23425 WARNING openerp_test openerp.modules.loading: The model base.field.serialized.test.model has no access rules, consider adding one. E.g. access_base_field_serialized_test_model,access_base_field_serialized_test_model,model_base_field_serialized_test_model,,1,1,1,1
----------
Found 1 lines with errors
```

Do you know why this error is there? I don't remember seeing it in other builds.

max3903:
:+1:

bwrsandman:
@guewen Missing access rights Warnings cause travis to fail. This is intended behaviour.

yvaucher:
Was travis restarted it's not red ? I don't see that access right error. So unless I'm missing something :+1:

guewen:
:+1:

PR #46, opened 2014-09-02
Port users ldap mail

dreispt:

PR #45, opened 2014-09-02
Port users ldap populate

dreispt:

PR #66, opened 2014-09-22
Port Users ldap populate

yajo:


dreispt:
That's was already done and is waiting review at #45.
You might also like to review #46.

yajo:
Yikes, true, I didn't notice. Thanks.

PR #68, opened 2014-09-25
Fix pylint tests

nbessi:


mail_environment/env_mail.py
```diff
@@ -140,7 +140,9 @@ def _get_incom_conf(self, cursor, uid, ids, name, args, context=None):
             res[fetchmail.id] = config_vals
         return res
 
-    def _type_search(self, cr, uid, obj, name, args, context={}):
+    def _type_search(self, cr, uid, obj, name, args, context=None):
+        if context is None:
+            context = self.pool['res.users'].context_get(cr, uid)
```
hbrunn, line 145: This changes the semantics of passing None as context. Shouldn't you better use an empty dictionary then?

nbessi, line 145: Please see https://lists.launchpad.net/openerp-community/msg06816.html

hbrunn, line 145: I saw that mail, but was not aware that we adopted this as a standard. Anyway, if you ie want to force no language, you can still use context={} in the function call, so I'm fine with that

hbrunn:
:+1:
would you mind to also fixing
server_environment/serv_config.py:59:38: E261 at least two spaces before inline comment
in this PR, then we're green again faster?

lepistone:
@hbrunn: fixed.
@nbessi  :+1: thanks

PR #69, opened 2014-10-01
Fix and update travis

bwrsandman:
Remove .coveragerc because new tests creates its own


pedrobaeza:
:+1:

rvalyi:
:+1: LGTM

dreispt:
TravisCI is red because of #65. That should really be sorted out.

PR #81, opened 2014-11-13
Cron run manually

yajo:
Trying again https://github.com/OCA/server-tools/pull/62.


hbrunn:
It seems you have to update the test for v8

yajo:
Travis says:

> [2014-11-13 11:57:03,347 23744 ERROR openerp_test openerp.sql_db: bad query: INSERT INTO "res_partner" [...]](https://travis-ci.org/OCA/server-tools/jobs/40876245#L265)
> [IntegrityError: null value in column "notify_email" violates not-null constraint](https://travis-ci.org/OCA/server-tools/jobs/40876245#L269)

So I think [this is the offending line](https://github.com/grupoesoc/server-tools/blob/cron_run_manually/cron_run_manually/tests/correct_uid.yml#L2), but I never coded anything about `notify_email` at all.

[Many official tests follow the same structure](https://github.com/odoo/odoo/search?l=yaml&q=%22!record+{model%3A+res.users%2C%22&type=Code&utf8=%E2%9C%93), and in my machine the test passes, so I think it must be some bug in the Travis build itself.

If I'm wrong, can anyone please help me find the error? Thanks.

cron_run_manually/ir_cron.py
```diff
@@ -0,0 +1,73 @@
...
+            # until the end of the transaction
+            self.env.cr.execute(
+                """SELECT *
+                   FROM ir_cron
+                   WHERE id=%s
+                   FOR UPDATE NOWAIT""",
+                (self.id,),
+                log_exceptions=False)
+
+        except Exception as e:
```
guewen: why not `except psycopg2.OperationalError as e:`?

hbrunn:
It's not exactly a bug in either your code or the travis test run, but a combination of unfortunate circumstances: We get the mail module installed in the database because some module in this branch depends on it. This adds the field including the not-null constraint. But run_cron_manually is loaded before the mail module, so the default it installs is not in effect, resulting in the error we see.

Your database either doesn't have the mail module installed or by coincidence it is loaded before run_cron_manually, then everything's fine, too. The official modules you mention all depend on the mail module directly or indirectly, so they'll never have this problem.

How to fix this? You could depend on mail and be done, but I don't see a good technical reason for that. It should suffice to pass `default_notify_email: 'none'` in the record's context as in https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/test/01_order_to_payment.yml#L110

yajo:
@hbrunn Thank you very much for this insight on the issue. Seems like testing the module `point_of_sale` is broken, so I guess it's not a good example. I had to put the context outside a string to make it work. Otherwise it was printing:

```
2014-12-09 17:07:14,935 21 ERROR demo openerp.tools.yaml_import: dictionary update sequence element #0 has length 1; 2 is required
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 894, in process
    self._process_node(node)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 905, in _process_node
    self.process_record(node)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 292, in process_record
    view_id = record.view
  File "/usr/lib/python2.7/site-packages/openerp/api.py", line 237, in wrapper
  File "/usr/lib/python2.7/site-packages/openerp/addons/base/ir/ir_model.py", line 1007, in _update
    context = dict(context, install_mode=True)
ValueError: dictionary update sequence element #0 has length 1; 2 is required
2014-12-09 17:07:14,960 21 ERROR demo openerp.tests: module cron_run_manually: an exception occurred in a test
  File "/usr/lib/python2.7/site-packages/openerp/modules/loading.py", line 64, in load_test
    _load_data(cr, module_name, idref, mode, 'test')
  File "/usr/lib/python2.7/site-packages/openerp/modules/loading.py", line 118, in _load_data
    tools.convert_file(cr, module_name, filename, idref, mode, noupdate, kind, report)
  File "/usr/lib/python2.7/site-packages/openerp/tools/convert.py", line 899, in convert_file
    convert_yaml_import(cr, module, fp, kind, idref, mode, noupdate, report)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 958, in yaml_import
    yaml_interpreter.process(yaml_string)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 894, in process
    self._process_node(node)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 905, in _process_node
    self.process_record(node)
  File "/usr/lib/python2.7/site-packages/openerp/tools/yaml_import.py", line 292, in process_record
    view_id = record.view
  File "/usr/lib/python2.7/site-packages/openerp/api.py", line 237, in wrapper
  File "/usr/lib/python2.7/site-packages/openerp/addons/base/ir/ir_model.py", line 1007, in _update
    context = dict(context, install_mode=True)
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

Anyway, let's hope the last commit fixes the build.

hbrunn:
ah, of course, as the field is not in the orm, the default won't do anything. Just depend on mail, then it will work

yajo:
I see this is still labeled as _needs fixing_. Anything else to do?

pedrobaeza:
I don't think so.

:+1:

hbrunn:
me neither. (The label has to be removed by hand, and Pedro already did) :+1:

PR #84, opened 2014-11-17
[7.0] [add] fields validator

eLBati:
This module allows to set a regular expresion as field validator.
When the regular expresion is set, write and create operations on the involved
field are blocked, if the regular expression is not satisfied.
See demo and test data for an example with partner email.


pedrobaeza:
Hi, @eLBati, thanks for the module. I think a better name for it is base_field_validator

Entering in the functionality, I feel that this kind of things must be done client-side (in JS widgets), not server-side. In this direction, there is this PR https://github.com/OCA/web/pull/34 and collides in functionality with this one. We should decide in which direction go.

eLBati:
Hi @pedrobaeza thanks for pointing it out (I don't follow https://github.com/OCA/web enough :pensive:)
It is fine for me to add `web_fields_masks` only.
We are going to test it.

eLBati:
Not sure if `web_fields_masks` supports regex (https://github.com/OCA/web/pull/34#issuecomment-63775884). It seems it does not

eLBati:
Just to notify that this needs review (can't edit labels)

pedrobaeza:
Don't you think it's better to add (or check) regex support on the other PR. This is to follow with my thought about having this kind of checks client-side.

lmignon:
@pedrobaeza IMO, the client side would stay without business logic... (and field validation is part of the business logic).  If I understand well the PR OCA/web#34, the validation doesn't prevent to save the value. In this PR, the validation is enforced even on the server side. I've the feeling that this PR covers another UC. What do you think?

pedrobaeza:
I think this is not business logic, but basic validations. Check for example HTML forms: this kind of validations (mandatory fields, numeric values, etc) are made always client side.

nhomar:
I think exactly as @lmignon It is a good feature, if you go throught a customized clients over web-servides the OCA/web#34 will be jumped.

+1 to have it in both sides (we need both).

Client side:

To avoid the POST that consume time and make a uncommit of the changes.

Server side:

To be consistent thoruhgt web-services.

On my side a total +1 "Conceptually Speaking" to this PR.

lmignon:
For sure the validation on the client will leverage the User experience and avoid unnecessary calls to the server. But it's the responsibility of the server to enforce the data consistency whatever  they are modified (by a request from the browser or from web-services or even internal calls). Conceptually the concept of validator sounds good. One step further would be to leverage the fields_view_get from BaseModel to add the validator definition in the field description send to the JS client and implement the validation also on the client side.... (I've a dream...)

pedrobaeza:
Yeah, we'll have to put it as wishlist.

@eLBati, please consider following remarks then and I'll approve it:
- Change name to _base_field_validator_
- Add a know issue / roadmap section in description about the lack of client side check.

eLBati:
Thanks for the reviews
@pedrobaeza done

pedrobaeza:
:+1: Thanks

yvaucher:
commits from hurrinico are not well defined, it is certainly missing email in author

Can you amend them?

base_field_validator/__openerp__.py
```diff
@@ -0,0 +1,90 @@
...
+{
+    'name': "Fields Validator",
+    'version': '0.1',
+    'category': 'Tools',
+    'summary': "Validate fields using regular expressions",
+    'description': """
+Fields Validator
+================
+
+This module allows to set a regular expresion as field validator.
```
tafaRU: s/expresion/expression

base_field_validator/__openerp__.py
```diff
@@ -0,0 +1,90 @@
...
+        'ir_model_view.xml',
+        'security/ir.model.access.csv',
+        ],
+    "demo": [
+        'ir_model_demo.xml',
+        ],
+    'test': [
+        'test/validator.yml',
+    ],
+    "active": False,
```
tafaRU: ```active``` is deprecated and is an alias for ```auto_install``` set False as default. So you can remove it.

base_field_validator/__openerp__.py
eLBati: Done, thanks

yvaucher:
:+1:

eLBati:
I think this can be merged

base_field_validator/ir_model.py
```diff
@@ -0,0 +1,111 @@
...
+            new_id = old_create(cr, uid, vals, context=context, **kwargs)
+            return new_id
+
+        return wrapper
+
+    def _wrap_write(self, old_write, model):
+        def wrapper(cr, uid, ids, vals, context=None, **kwargs):
+            self.check_vals(cr, uid, vals, model, context=context)
+            old_write(cr, uid, ids, vals, context=context, **kwargs)
+            return True
```
hbrunn: I think you should return whatever `old_write` returns

base_field_validator/ir_model.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return wrapper
+
+    def _register_hook(self, cr, ids=None):
+        """ Wrap the methods `create` and `write` of the model
+        """
+        if ids is None:
+            ids = self.search(cr, SUPERUSER_ID, [])
```
hbrunn: why not search for `[('validator_line_ids', '!=', False)]` here?

base_field_validator/ir_model.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _wrap_write(self, old_write, model):
+        def wrapper(cr, uid, ids, vals, context=None, **kwargs):
+            self.check_vals(cr, uid, vals, model, context=context)
+            old_write(cr, uid, ids, vals, context=context, **kwargs)
+            return True
+
+        return wrapper
+
+    def _register_hook(self, cr, ids=None):
```
hbrunn, line 64: call super

hbrunn:
@eLBati shouldn't you call `_wrap*` on `ir.model.validator.line#create` if it's the first validator line for this model? Otherwise, the user has to restart the server (or trigger `_register_hook` by updating some modules) for changes to be in effect

base_field_validator/ir_model.py
```diff
@@ -0,0 +1,111 @@
...
+                        model_obj.create, model)
+                    model_obj.write = self._wrap_write(model_obj.write, model)
+                    model_obj.field_validator_checked = True
+                    updated = True
+        return updated
+
+    def create(self, cr, uid, vals, context=None):
+        res_id = super(IrModel, self).create(
+            cr, uid, vals, context=context)
+        if self._register_hook(cr, [res_id]):
```
hbrunn: don't rely on `_register_hook`'s result, this can be messed up by some other module's override. Move the relevant code to its own function, and call it both in `_register_hook` and here

hbrunn:
forget about https://github.com/OCA/server-tools/pull/84#issuecomment-136956525, I see you do this in `ir.model#write` and `create`. Still, I think the code would be simplified a lot if you do it in `ir.model.validator.line#create`

eLBati:
@hbrunn thanks, I made the proposed changes.
Now I try to find out why travis failed

eLBati:
Ok, my bad. Now travis is green :green_heart:

eLBati:
@OCA/tools-maintainers please set this 'needs review'

StefanRijnhart:
As on the 8.0 PR, better not to change the signature of _register_hook

guewen:
Can we merge this PR?

gurneyalex:
:+1:

PR #85, opened 2014-11-19
[7.0] [ADD] 'document extract from database'

legalsylvain:
Hi, 

I propose a module developped by Julius Network Solution, I got here http://bazaar.launchpad.net/~julius-network-solutions/julius-openobject-addons/7.0/files.

I did some refactoring (flake8, Pylint, adding tests)

Coverage: 93%.

Note: Travis fails for Pylint bugs in other modules.

Thanks for your review.


document_extract_from_database/model/ir_attachment.py
```diff
@@ -0,0 +1,47 @@
...
+            }
+            self.write(cr, uid, id, vals, context=context)
+        return True
+
+    def write_again(self, cr, uid, ids, context=None):
+        if context is None:
+            context = {}
+        for document_id in ids:
+            self._write_again(cr, uid, document_id, context=context)
+        return True
```
clonedagain: What is this for?

document_extract_from_database/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+    "website": "http://julius.fr",
+    "category": "Tools",
+    "depends": [
+        "document",
+    ],
+    "description": """
+Button to extract documents from your database
+==============================================
+If you want to extract the document stored in your database this module is for
+you.
```
clonedagain: Can you please rephrase and elaborate on what this module actually does? The meaning of "extract" is not clear.

document_extract_from_database/data/ir_values.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+
+        <record model="ir.values" id="document_multiple_action">
+            <field name="model_id" ref="base.model_ir_attachment" />
+            <field name="name">Document Extraction</field>
+            <field name="key2">client_action_multi</field>
+            <field name="value" eval="'ir.actions.act_window,' + str(ref('action_view_document_multiple_action_form'))" />
```
clonedagain: I think you can write %(xml ref)d instead of injecting python here.

document_extract_from_database/__openerp__.py
legalsylvain: Thanks for you review. 
I'm not very fluent in english. Are you agree with that: 
- Module Name: document_extract_from_database --> document_export_from_db_to_fs
- Name: Extract Documents from database --> Export Existing Documents from Database to File System; 
- Description replaced by: Add a Button to export existing documents from database to File System

document_extract_from_database/model/ir_attachment.py
legalsylvain: On which line(s) is your question ?

document_extract_from_database/data/ir_values.xml
legalsylvain: Thanks. Done.

document_extract_from_database/model/ir_attachment.py
clonedagain: Why is there write_again and _write_again ?

document_extract_from_database/data/ir_values.xml
legalsylvain: I redo the changes, because it didn't work. 
The concat of 'ir.actions.act_window, + the xml_id is needed.

document_extract_from_database/tests/test_document_extract_from_database.py
```diff
@@ -0,0 +1,78 @@
...
+        })
+
+        # Run Write Again Functionnality
+        self.dma_obj.write_again(cr, uid, 0, {'active_ids': [ia_id]})
+
+        data_after = self.ia_obj.browse(cr, uid, ia_id).datas
+
+        # Test if datas have not changed
+        # FIXME: After rewriting document, there is \n in 'datas' field
+        # The file is still readable.
```
legalsylvain: Pretty weird behaviour. If some people want to investigate...
Update: it's normal and it occurs when File system is used. This is not due to this module.

legalsylvain:
@clonedagain : I did the changes you proposed. Can you check again and say if all is OK ?

Regards.

clonedagain:
Looks fine but I'm still not sure I fully understand what problem it is supposed to solve.
Is it some kind of migration help for those who start a database using db storage and the switch to FS storage?

legalsylvain:
Exactly.
My company uses Odoo since 2 years. We have a DB of 1GO and never set 'FS storage'. So this module allow after setting FS storage, to put all the documents in file system easily.
(So the size of my DB is now 180 MO)

Even if it's not usefull for all odoo user, I think that most of people that begin to use Odoo do not configure correctly file storage (as me) and after a long time, discover that there is a problem.

Maybe this module will not be as usefull in V8 as in V7, because the default behaviour is not the same.
Ref: http://anybox.fr/blog/postgresql-large-object-storage-for-odoo

Regards.

clonedagain:
OK I think I see. I've done more or less the same on our DB, using an ad-hoc XMLRPC/python script inspired by various sources.
So with this module, when you change your storage, you go to the list of attachments, select some or all of them, and call the action.
Then 2 more remarks:
- I expect the process will work equally well to reload the FS attachements to BLOBS if you revert to database storage. It would be nice to reflect this versatility in the description.
- maybe a wizard would be nice to have to start the action on all attachments at once (but I'll not block the PR for it since it's a seldom needed maintenance operation)

Other than that :+1: for me

guewen:
:+1: 

thanks

PR #86, opened 2014-11-20
[REF] [7.0] pylint;

legalsylvain:


tree_view_record_id/orm.py
```diff
@@ -12,12 +12,9 @@
...
 
-""" procedural code is executed even if the module is not installed
-
-    TRY TO SWITCH to _register_hook() in the model to avoid
-    execution when not installed in V 8.0 version
-
-"""
+# procedural code is executed even if the module is not installed
+#    TRY TO SWITCH to _register_hook() in the model to avoid
+#    execution when not installed in V 8.0 version
```
guewen: It is valid to have a docstring in a python module, but it must be before python code, so before the imports.

fetchmail_attach_from_folder/match_algorithm/email_domain.py
```diff
@@ -20,10 +20,10 @@
 #
 ##############################################################################
 
-from email_exact import email_exact
+from . import email_exact
 
 
-class email_domain(email_exact):
+class email_domain(email_exact.email_exact):
```
guewen: Why not?

```
from .email_exact import email_exact
```

Anyway both work, you don't need to change.

tafaRU:
@legalsylvain I think you should use the raw notation as suggested that is r prefix.

legalsylvain:
Can you propose me the exact code, I don't know regex very well.
Thanks.

tafaRU:
@legalsylvain,  raw string notation (r"text") keeps regular expressions sane....
In your case:

```
attribute_match = re.search(
              r'([a-zA-Z_]+)=\%s', conf['ldap_filter'])
```

tree_view_record_id/orm.py
legalsylvain: Done.

fetchmail_attach_from_folder/match_algorithm/email_domain.py
legalsylvain: Thanks, I didn't know.

legalsylvain:
@tafaRU: Thanks.
Done.

guewen:
:+1:

pedrobaeza:
:+1:

tafaRU:
Once it's green, :+1:

Issue #91, opened 2014-12-08
Propose auth module to generate randomly user's password


legalsylvain:
Hi all, 

I recently developped an 'auth' module for my company (auth_generate_password). The features are: 
- Users can not change them password;
- Password are generated randomly and an email is send to them;
- an action is available to reinitialize many users passwords;

Maybe it can be usefull for other people in the community. If yes, please :+1:, and I'll share it.

If you want to check the code...
https://github.com/grap/odoo-addons-grap/blob/7.0/auth_generate_password/__openerp__.py

Regards.

hbrunn:
That's certainly useful! In a review though, I would ask you to do the mailing via a template and have the password length configurable via an ir_config_parameter.

legalsylvain:
Thanks for your answer. I did the PR #92 .
Regards.

PR #92, opened 2014-12-09
[ADD] 'auth_generate_password' module to generate randomly user's password

legalsylvain:
Propose to the community a simple module to:
- block users that can not change them password;
- generate password, and send to the users;

The type of password is defined in two parameters:
- password_size (default : 6);
- password_chars (default : string.ascii_letters + string.digits)
- flake 8 / pylint : OK;
- coverage of the module : ~ 95 %
- [X] change the module icon;
- [X] make this module compatible if password is encryted;


legalsylvain:
#91 
@hbrunn , I change to use email.template. It works pretty fine but I set in the template "object.password". If the password is encrypted by another module, it will not work. Do you know how transmit a defined value to a template ?

UPDATE: 
- Fixed, using ctx.get to have the password in template;

legalsylvain:
@all : Travis fails if 'mail_environment' is installed. ( so it doesn't fail in UNITEST mode);
The reason is that in my test.py file, I try to send mail and 'mail_environment' drops columns (making columns fields.function).
https://travis-ci.org/OCA/server-tools/jobs/43467649#L397

How can I solve the problem ?

UPDATE: 
- Problem fixed, deleting in the test, the ir_mail_server object.

hbrunn:
:+1:  (and sorry for not replying last week)

auth_generate_password/data/email_template.xml
```diff
@@ -0,0 +1,45 @@
...
+            <field name="body_html"><![CDATA[
+Your OpenERP credentials has been changed:
+<ul>
+<li>Login : ${object.login}</li>
+<li>New Password : ${ctx.get('generated_password')}</li>
+</ul>
+Please,
+<ul>
+<li>remember this password and delete this email;</li>
+<li>Communicate the password to your team, if you are many people to use the same credentials;</li>
```
pedrobaeza: s/if you are to use/if there are many people that are using

auth_generate_password/i18n/fr.po
```diff
@@ -0,0 +1,72 @@
...
+msgid "\n"
+"Your OpenERP credentials has been changed:\n"
+"<ul>\n"
+"<li>Login : ${object.login}</li>\n"
+"<li>New Password : ${object.password}</li>\n"
+"</ul>\n"
+"Please,\n"
+"<ul>\n"
+"<li>remember this password and delete this email;</li>\n"
+"<li>Communicate the password to your team, if you are many people to use the same credentials;</li>\n"
```
pedrobaeza: Change it accordingly

auth_generate_password/model/res_users.py
```diff
@@ -0,0 +1,65 @@
...
+#
+##############################################################################
+
+# This line is noqa flaged, to allow users to use string functions written
+# in ir_config_parameter
+import string # flake8: noqa
+import random
+
+from openerp.osv.orm import Model
+from openerp.osv.osv import except_osv
```
pedrobaeza: You can use orm.except_orm instead

pedrobaeza:
:+1:

legalsylvain:
2 :+1:, but the runbot is failing.
I don't know how to identify the problem. (it is the first time I see the runbot interface) 
Somebody can help me ?

auth_generate_password/model/res_users.py
```diff
@@ -0,0 +1,64 @@
...
+    def generate_password(self, cr, uid, ids, context=None):
+        if not context:
+            context = {}
+        mm_obj = self.pool['mail.mail']
+        icp_obj = self.pool['ir.config_parameter']
+        imd_obj = self.pool['ir.model.data']
+        et_obj = self.pool['email.template']
+        password_size = eval(icp_obj.get_param(
+            cr, uid, 'auth_generate_password.password_size'))
+        password_chars = eval(icp_obj.get_param(
```
bealdav, line 49: Maybe you may

```
try:
   int(cp_obj.get_param(
        cr, uid, 'auth_generate_password.password_size'))
except:
    raise except_orm(_("error"), _("Only digit chars authorized"))
```

bealdav:
Except use of eval, it is ok for me

legalsylvain:
Thanks, but it only works for password_size not for password_chars.
I'll do the change.

legalsylvain:
@pedrobaeza, @hbrunn, @bealdav : runbot is now green. (after a simple rebase).
This PR can be merged if you're OK.

kind regards.

dreispt:
There is also a PR to prevent users from changing passwords: https://github.com/OCA/server-tools/pull/249
I think that people could want random passwords but then let people change them.
Wouldn't it be best to have the two features in two separate modules?

auth_generate_password/data/ir_config_parameter.yml
```diff
@@ -0,0 +1,29 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+- !record {model: ir.config_parameter, id: password_size}:
+    key: auth_generate_password.password_size
+    value: 6
```
dreispt: IMO 6 is quite a bad default. I would have 12, or at least 10.
Also, I guess this ought to be created with `noupdate="1"`...

legalsylvain:
@dreispt : 

> There is also a PR to prevent users from changing passwords: #249
> I think that people could want random passwords but then let people change them.
> Wouldn't it be best to have the two features in two separate modules?

Yes, you're right. When I'll port this module, I'll remove this feature from this module. (my module is a V7, and you're pointing a V8 PR) ;

auth_generate_password/data/ir_config_parameter.yml
legalsylvain: do you know how to set noupdate in a yml file ?
6 -> 12 : agree.

dreispt:
I'm still concerned that parameters get overwritten on a module upgrade.

legalsylvain:
@dreispt : 

> I'm still concerned that parameters get overwritten on a module upgrade.

I can figure out your PoV, but it depends how do you apply settings on your database.
1/ you realize changes by UI:
-> if you don't set noupdate, a module upgrade will break the setting; (what you say)

2/ you realize changes by a custom module:
-> if you set noupdate, this will not work;

So, there is no perfect settings.

My personal case (2), I write a 'custom_module' and in this module I put all my specific configuration, and so I can recover a complete demo environment, if I want.
With what you propose, my kind of implementation will not work.

What do you think ?

But yes, I could understand that all ir.config_parameter should be noupdatable... 
I don't know, if there are some conventions.

dreispt:
@legalsylvain AFAIK mopdules should expect parameters to be modified by power users through the GUI. So 1/ is case to support, and Odoo has the tooling for that. You will need to convert the YML to XML.

hbrunn:
or write a yaml stanza that modifies the generated ir.model.data record (I never use yaml, no idea if there's a better way to do it)

Concerning update/noupdate: You can have both of it. Set it to noupdate in this module, depending modules can do this:

``` xml
<data noupdate="1">
    <function model="ir.model.data" name="write">
        <function model="ir.model.data" name="search">
            <value eval="[('module', '=', 'auth_generate_password'), ('name', '=', 'password_size')]" />
        </function>
        <value eval="{'noupdate': False}" />
    </function>
</data>
<data>
    <!-- change the noupdate record /-->
</data>
<data noupdate="1">
    <function model="ir.model.data" name="write">
        <function model="ir.model.data" name="search">
            <value eval="[('module', '=', 'auth_generate_password'), ('name', '=', 'password_size')]" />
        </function>
        <value eval="{'noupdate': True}" />
    </function>
</data>
```

This pattern allows us to change noupdate records on module installation, but doesn't mess with the records on subsequent updates

legalsylvain:
@hbrunn : Thanks a lot for this trick ! I didn't know. Very usefull.
@dreispt : Thanks for your review. I changed the code.

dreispt:
All good, but it would be safer to add to the module description a "Roadmap" section explaining that the restrict password changing feature should be removed when porting to 8.0.

legalsylvain:
@dreispt : done.

dreispt:
Thanks! :+1:

PR #93, opened 2014-12-10
Fix server_environment module imports

### LINKED ISSUE #89: ImportError: No module named server_environment_files
I stumbled upon this when trying the scaffold command: https://github.com/odoo/odoo/issues/4034

Looking at the `server_environment` module, is has a dependency on a `server_environment_files` module that doesn't exist: https://github.com/OCA/server-tools/blob/8.0/server_environment/__openerp__.py#L25

Is this doomed to error or am I missing something?

dreispt:
Quick fix to make `./odoo.py help` work without errors when the
server_environment_files modules does not exist.
Fixes #89.

### DIFF
--- a/server_environment/serv_config.py
+++ b/server_environment/serv_config.py
@@ -29,29 +29,35 @@ from openerp.tools.config import config as system_base_config
 
 from .system_info import get_server_environment
 
-from openerp.addons import server_environment_files
-_dir = os.path.dirname(server_environment_files.__file__)
+try:
+    from openerp.addons import server_environment_files
+except ImportError:
+    _dir = False
+else:
+    _dir = os.path.dirname(server_environment_files.__file__)
 
-# Same dict as RawConfigParser._boolean_states
-_boolean_states = {'1': True, 'yes': True, 'true': True, 'on': True,
-                   '0': False, 'no': False, 'false': False, 'off': False}
+if _dir:
+    # Same dict as RawConfigParser._boolean_states
+    _boolean_states = {
+        '1': True, 'yes': True, 'true': True, 'on': True,
+        '0': False, 'no': False, 'false': False, 'off': False}
 
-if not system_base_config.get('running_env', False):
-    raise Exception(
-        "The parameter 'running_env' has not be set neither in base config "
-        "file option -c or in openerprc.\n"
-        "We strongly recommend against using the rc file but instead use an "
-        "explicit config file with this content:\n"
-        "[options]\nrunning_env = dev"
-    )
+    if not system_base_config.get('running_env', False):
+        raise Exception(
+            "The parameter 'running_env' has not be set neither in "
+            "base config file option -c or in openerprc.\n"
+            "We strongly recommend against using the rc file but "
+            "instead use an explicit config file with this content:\n"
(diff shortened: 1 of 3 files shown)


lepistone:
would `get_module_resource` be useful here?

dreispt:
Build fails when trying to run tests for server_environment.
I'm clueless on how this is supposed to work.
At least I have Odoo commands working in my local environment.

dreispt:
@nbessi Do you think you could spare a few minutes to have a look at this?

nbessi:
@dreispt   it may come form the fact that  you removed the dependence on `server_environment_files`.  So there is nothing that prevent `server_environnement` to be loaded before `server_environnement_files`.

As far as I can see the setup still seems to be ok.

```
- ln -s server_environment_files_sample ./server_environment_files
```

Regards

Nicolas

dreispt:
@nbessi Thanks for the pointers. I tried that, and when I try `./odoo.py help` now I get:

```
  File "./odoo.py", line 160, in <module>
    main()
  File "./odoo.py", line 157, in main
    openerp.cli.main()
  File "/home/dr/work/odoo/openerp/cli/__init__.py", line 60, in main
    __import__(m)
  File "/home/dr/work/odoo/openerp/modules/module.py", line 77, in load_module
    mod = imp.load_module('openerp.addons.' + module_part, f, path, descr)
  File "/home/dr/work/oca/server-tools/server_environment/__init__.py", line 21, in <module>
    from .serv_config import serv_config, setboolean
  File "/home/dr/work/oca/server-tools/server_environment/serv_config.py", line 41, in <module>
    "The parameter 'running_env' has not be set neither in base config "
Exception: The parameter 'running_env' has not be set neither in base config file option -c or in openerprc.
We strongly recommend against using the rc file but instead use an explicit config file with this content:
[options]
running_env = dev
```

server_environment/serv_config.py
```diff
@@ -29,29 +29,35 @@
...
+    _dir = False
+else:
+    _dir = os.path.dirname(server_environment_files.__file__)
+
+if _dir:
+    # Same dict as RawConfigParser._boolean_states
+    _boolean_states = {
+        '1': True, 'yes': True, 'true': True, 'on': True,
+        '0': False, 'no': False, 'false': False, 'off': False}
+
```
nbessi, line 44: @dreispt We have to add a way to check if module is installed here.
There is not need to raise an exception if module is not installed. 

By default we wanted to ensure that setup is correct if module was installed and OpenERP should not boot if it is the case.

An option is to move the check later but it is more error prone.

hbrunn:
Can't you move all of that to server.config._register_hook? This fail fail early enough if there is no configuration, but won't interfere if the module is not installed

dreispt:
@hbrunn it causes errors in v8 even if the module is not installed - it's enough to be in the addons path.. Just try `./odoo.py help` and you will get a traceback.

hbrunn:
Yes, that's why I propose to move that code to _register_hook

StefanRijnhart:
+1 for putting this in _register_hook if possible.

dreispt:
Maybe the authors can provide a better fix and we close this?

PR #94, opened 2014-12-10
[ADD] Add module base_recursive_model

YannickB:
Hello world,

Here is a module I developed for the community verticalisation (https://github.com/OCA/vertical-community) but since it's not specific to communities I think it's best to place it here.

This module manage two concept relative to recursivity in Odoo : 
-The parent_id link, which is used in many modules even in official one, but each time the concept is reimplemented. I think it's best to create an abstract class as I did here.
-Configuration inheritance, this class allow to inherit some configuration from parent object or specified object. It's kinda complicated to describe it here, I suggest you check the modules project_assignment and vote on https://github.com/OCA/vertical-community. I'll PR project_assignment today to help with the functional reviewing of this part.


base_recursive_model/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+    'description': """
+Base recursive model
+====================
+
+Create two abstract model which can be used to manage recursive relations
+-------------------------------------------------------------------------
+    * Easily create recursive model (instead of having to create parent_id
+        each time, you now just have to inherit the abstract model
+    * Define configuration fields which can then be inherited and overridden
+        by children and children models
```
yvaucher: Here you can kill 2 birds with one stone.

Instead of putting a `README.md` plus a description, now in 8.0 you can remove description from `__openerp__.py` and put everything in a README file.

When odoo see no description in `__openerp__.py` it look up for README file and will use it.

Actually we are creating `README.rst` and put description in it which will be readable on Github and on Odoo.

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+            'base.config.inherit.line.del', 'res_id',
+            domain=lambda self: [('model', '=', self._name)],
+            auto_join=True,
+            string='Configuration line deleted'
+        ),
+    }
+
+    def _prepare_config(self, cr, uid, id, record, vals={}, context=None):
+        # This function can be overridden by inheriting model
+        # to specify the field contains in the configuration
```
yvaucher: Please use """ for docstring

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+        res = {}
+        for key, value in vals.iteritems():
+            res[key] = value
```
yvaucher: Here you certainly want a res = vals.copy()

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+    _columns = {
+        'base_config_inherit_line_del_ids': fields.one2many(
+            'base.config.inherit.line.del', 'res_id',
+            domain=lambda self: [('model', '=', self._name)],
+            auto_join=True,
+            string='Configuration line deleted'
+        ),
+    }
+
+    def _prepare_config(self, cr, uid, id, record, vals={}, context=None):
```
yvaucher: Please don't use mutable in method signature
use vals=None as for context

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+    ):
+        # This function can be overriden in inheriting model
+        # to trigger the compute of children models
+        return True
+
+    def _update_stored_config(self, cr, uid, ids, context=None):
+        # When called, this function will purge all stored configuration
+        # and recompute them, by recover it from parent and checking
+        # if the configuration isn't override in the current record.
+        config_obj = self.pool.get(self._base_config_inherit_model)
```
yvaucher: instead of self.pool.get(), you can use self.pool[]

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+            self, cr, uid, ids, context=None
+    ):
+        # This function can be overriden in inheriting model
+        # to trigger the compute of children models
+        return True
+
+    def _update_stored_config(self, cr, uid, ids, context=None):
+        # When called, this function will purge all stored configuration
+        # and recompute them, by recover it from parent and checking
+        # if the configuration isn't override in the current record.
```
yvaucher: This method is a bit long could you split it ?

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+                for config_line_del in config_del_obj.browse(
+                        cr, uid, parent_config_del_ids, context=context
+                ):
+                    config_line_dels[config_line_del['key']] = {
+                        'model': self._name,
+                        'res_id': record.id,
+                        'key': config_line_del['key']
+                    }
+                    if config_line_del['key'] in config_lines:
+                        del config_lines[key]
```
yvaucher: this `key` field seems strange to me as it would be the last one of loop on 106.
Is it an indentation error from line 108 ?

Or you simply want to

```
del config_lines[config_line_del['key']]
```

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,245 @@
...
+                    config_lines[key] = self._prepare_config(
+                        cr, uid, record.id, config_line, context=context
+                    )
+
+            for config_line in getattr(record, self._base_config_inherit_o2m):
+                key = getattr(config_line, self._base_config_inherit_key).id
+                if config_line.action == 'add':
+                    config_lines[key] = self._prepare_config(
+                        cr, uid, record.id, config_line,
+                        vals={}, context=context
```
yvaucher: You should need to pass vals={}
vals default value should be None and in method add

```
if vals is None:
    vals = {}
```

base_recursive_model/base_recursive_model.py
```diff
@@ -0,0 +1,98 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import fields, orm, osv, osv
```
yvaucher: If you see double it is not because of :beers: 

osv twice there

base_recursive_model/base_recursive_model.py
```diff
@@ -0,0 +1,98 @@
...
+            res.append((record['id'], name))
+        return res
+
+    def name_search(
+            self, cr, uid, name, args=None, operator='ilike',
+            context=None, limit=100
+    ):
+        # Search all category which contain name
+        # in their own name or in their parents
+        if not args:
```
yvaucher: check if args is None
if it's already [] we don't want to reassign it again

base_recursive_model/base_recursive_model.py
```diff
@@ -0,0 +1,98 @@
...
+            args = []
+        if not context:
```
yvaucher: context is None

base_recursive_model/base_recursive_model.py
```diff
@@ -0,0 +1,98 @@
...
+        'complete_name': fields.function(
+            _name_get_fnc, type="char", string='Full Name', store=True
+        ),
+        'parent_left': fields.integer('Left parent', select=True),
+        'parent_right': fields.integer('Right parent', select=True),
+    }
+    _constraints = [
+        (
+            osv.osv._check_recursion,
+            'Error ! You can not create recursive categories.', ['parent_id']
```
yvaucher: Error! without space

Space after punctuation marks is only for french

base_recursive_model/i18n/de_DE.po
```diff
@@ -0,0 +1,112 @@
...
+msgstr "Erstellt von"
+
+#. module: base_recursive_model
+#: field:base.config.inherit.line.del,create_date:0
+msgid "Created on"
+msgstr "Erstellt am"
+
+#. module: base_recursive_model
+#: constraint:base.recursive.model:0
+msgid "Error ! You can not create recursive categories."
```
yvaucher: msgid to fix

base_recursive_model/i18n/fr.po
```diff
@@ -0,0 +1,113 @@
...
+msgstr "Créé par"
+
+#. module: base_recursive_model
+#: field:base.config.inherit.line.del,create_date:0
+msgid "Created on"
+msgstr "Créé le"
+
+#. module: base_recursive_model
+#: constraint:base.recursive.model:0
+msgid "Error ! You can not create recursive categories."
```
yvaucher: msgid to fix

base_recursive_model/i18n/base_recursive_model.pot
```diff
@@ -0,0 +1,112 @@
...
+msgstr ""
+
+#. module: base_recursive_model
+#: field:base.config.inherit.line.del,create_date:0
+msgid "Created on"
+msgstr ""
+
+#. module: base_recursive_model
+#: constraint:base.recursive.model:0
+msgid "Error ! You can not create recursive categories."
```
yvaucher: msgid to fix

YannickB:
Thank you Yannick for your complete review, I think I have corrected all your remarks.

bwrsandman:
Hi @YannickB.
This looks very interesting indeed.

Would you mind adding some unittests to the module?

YannickB:
Hello, since base_recursive_model only contain abstract class I'm not sure it's a good idea to place unit test here. I made unit test in the module which use it, for example project_assignment https://github.com/OCA/project-service/pull/47

dreispt:
I think it's reasonable to add a test module, that also showcases it's usage.

YannickB:
Well I'm not so sure since the module project_assignment is quite simple and self-explanatory, I don't think we need another one. I also have to say, I know I'll not have time to make this test module before a long time and this will put the PR on hold for too long.

I know base_recursive_model is quite difficult to understand so it's best to try project_assignment to fully understand it's purpose. You can check http://demo.wezer.org demo/demo and go to project and task objects to understand what I mean by inherited configuration.

dreispt:
OK. My concern is that, in principle, we should aim for 100% code test coverage.
Thinking again, Abstract Models can be tested using Models created by tests. I think there is an example of that in this repo.

YannickB:
Ok I agree with you then, we shall improve this later but not in the initial PR I think.

YannickB:
Just came back from holiday, can you tell me where I can find such example?

I'm sorry, I am really not comfortable with tests on abstract model, if someone can help writing the test I'd be really grateful, I can't continue the integration of the community verticalisation until this module is merge here. I bet everything on the fact that the unit test on project_assignment would be enough.

dreispt:
Try looking at this example: https://github.com/OCA/server-tools/blob/8.0/base_field_serialized/tests/test_serialized.py

YannickB:
Finally the test script is ready, the module shall be ready for merge

base_recursive_model/README.rst
```diff
@@ -0,0 +1,10 @@
+Base recursive model
+====================
+
+Create two abstract model which can be used to manage recursive relations
+-------------------------------------------------------------------------
+    * Easily create recursive model (instead of having to create parent_id
+        each time, you now just have to inherit the abstract model
+    * Define configuration fields which can then be inherited and overridden
+        by children and children models. You have use example with modules
+        vote and project_assignment.
```
bwrsandman, line 10: Could you expand on this with the help of the [OCA README template](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)?

base_recursive_model/base_config_inherit.py
```diff
@@ -0,0 +1,291 @@
...
+        Purge all computed configuration in the given record
+        """
+        config_obj = self.pool[self._base_config_inherit_model]
+        config_del_obj = self.pool['base.config.inherit.line.del']
+
+        # Purge all existing configuration
+        config_stored_ids = config_obj.search(
+            cr, uid,
+            [
+                ('model', '=', self._name), ('res_id', 'in', ids),
```
bwrsandman, line 95: Not huge, but it would be easier to read if these two were on their own lines

base_recursive_model/base_recursive_model.py
```diff
@@ -0,0 +1,102 @@
...
+    _columns = {
+        'complete_name': fields.function(
+            _name_get_fnc, type="char", string='Full Name', store=True
+        ),
+        'parent_left': fields.integer('Left parent', select=True),
+        'parent_right': fields.integer('Right parent', select=True),
+    }
+    _constraints = [
+        (
+            osv.osv._check_recursion,
```
bwrsandman, line 95: `osv.osv` is deprecated, how about using `orm.Model._check_recursion`?

pedrobaeza:
@YannickB, is this still in use? If so, can you correct comments?

YannickB:
Hello, not really so let's close the PR for now. I'll reopen it if needed in the future.

pedrobaeza:
OK, thanks for answering.

PR #98, opened 2014-12-17
Skip init if raven is not available on the system

bwrsandman:
I've been informed that modules and repos using server tools are getting errors related to raven when the module is not installed.

This is a quick fix which skips everything when raven is not available.

If someone has a solution to have this code only run when sentry is installed, please let me know so I can integrate it.

### DIFF
--- a/sentry_logger/__init__.py
+++ b/sentry_logger/__init__.py
@@ -26,39 +26,39 @@ import cgitb
 from openerp.tools import config
 from openerp.addons.web.controllers.main import Session
 
-from .odoo_sentry_client import OdooClient
-from .odoo_sentry_handler import OdooSentryHandler
+try:
+    from .odoo_sentry_client import OdooClient
+    from .odoo_sentry_handler import OdooSentryHandler
 
+    root_logger = logging.root
 
-root_logger = logging.root
-
-processors = (
-    'raven.processors.SanitizePasswordsProcessor',
-    'raven_sanitize_openerp.OpenerpPasswordsProcessor'
-)
-if config.get(u'sentry_dsn'):
-    cgitb.enable()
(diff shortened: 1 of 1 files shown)


dreispt:
This is similar to https://github.com/OCA/server-tools/issues/89

lepistone:
In general the problem is that odoo imports (in the python sense) the modules that are not installed. So if we don't have python dependencies of modules that are not installed, they crash.

For that reason, I agree we want to catch and ignore import errors (otherwise, for an **installed** module, it's not bad to get an `ImportError`).

This can potentially affect many people, and I'm for merging that quickly even if it's not ideal. :+1: 

Thanks!

joaoalf:
I agree with @lepistone about quick merge.

:+1:

bwrsandman:
This isn't an ideal solution, but it's a quick one.

foutoucour:
:+1:

bwrsandman:
@foutoucour, @joaoalf please make sure this patch fixes your problem. I haven't tested this.

joaoalf:
@bwrsandman 

I just tested and it works.
:+1:

bwrsandman:
Okay, please merge!

max3903:
:+1:

foutoucour:
A bit late, but tested and it works :)

On Wed, Dec 17, 2014 at 11:53 AM, max3903 notifications@github.com wrote:

> [image: :+1:]
> 
> ## 
> 
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/server-tools/pull/98#issuecomment-67353356.

## 
- Jordi Riera, Connecting Softwares and people.
  +1 438 930 3718
  www.linkedin.com/in/jordirieracg/ http://ca.linkedin.com/in/jordirieracg/

PR #99, opened 2014-12-17
[ADD] [7.0] 'sentry_logger' Possibility to customize logging level

legalsylvain:
Possibility to customize log level;

### DIFF
--- a/sentry_logger/__init__.py
+++ b/sentry_logger/__init__.py
@@ -26,6 +26,8 @@ import cgitb
 from openerp.tools import config
 from openerp.addons.web.controllers.main import Session
 
+_DEFAULT_LOGGING_LEVEL = logging.ERROR
+
 try:
     from .odoo_sentry_client import OdooClient
     from .odoo_sentry_handler import OdooSentryHandler
@@ -40,12 +42,17 @@ try:
         cgitb.enable()
         # Get DSN info from config file or ~/.openerp_serverrc (recommended)
         dsn = config.get('sentry_dsn')
+        level = config.get('sentry_logging_level')
+        if not level:
+            level = _DEFAULT_LOGGING_LEVEL
+        else:
+            level = eval(level)
         # Create Client
         client = OdooClient(
             dsn=dsn,
             processors=processors,
         )
-        handler = OdooSentryHandler(client, level=logging.ERROR)
+        handler = OdooSentryHandler(client, level=level)
         root_logger.addHandler(handler)
     else:
         root_logger.warn(u"Sentry DSN not defined in config file")
--- a/sentry_logger/__init__.py
+++ b/sentry_logger/__init__.py
@@ -42,11 +42,10 @@ try:
         cgitb.enable()
         # Get DSN info from config file or ~/.openerp_serverrc (recommended)
         dsn = config.get('sentry_dsn')
-        level = config.get('sentry_logging_level')
-        if not level:
+        try:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I suppose that this is a WIP, doesn't it?

legalsylvain:
@pedrobaeza, no finally I will realize Two PR because, the second customization is pretty hard to do.
So ready for review if you want !

CC : @bwrsandman, @arthru

bwrsandman:
You're going to get flooded with logs if a log level of warning.

Technically, it looks good.

sentry_logger/__init__.py
```diff
@@ -40,12 +42,17 @@
         cgitb.enable()
         # Get DSN info from config file or ~/.openerp_serverrc (recommended)
         dsn = config.get('sentry_dsn')
+        level = config.get('sentry_logging_level')
+        if not level:
+            level = _DEFAULT_LOGGING_LEVEL
+        else:
+            level = eval(level)
```
arthru: Avoid to use eval. I would rather do something like :

``` python
if hasattr(logging, level):
    level = getattr(logging, level)
```

See http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

sentry_logger/__openerp__.py
```diff
@@ -27,12 +27,18 @@
...
+Settings:
+---------
+You have to add extra parameters in your odoo config file (~/.openerp_serverrc)
+    * mandatory parameters:
+        * Insert sentry DSN with value:
+          sentry_dsn = sync+<Your Sentry DSN>
+    * optional parameters:
+        * Define the level of log sent to sentry:
+          sentry_logging_level = logging.WARNING
+          (by default logging.ERROR)
```
arthru: According to my previous comment, remove "logging." from that two last lines)

sentry_logger/__init__.py
legalsylvain: This will return a string. 
I need a logging.xxx to initialize my handler : https://github.com/OCA/server-tools/pull/99/files#diff-f7ec9d9cf4dfd5c128a322da6ee5ac11R55

bwrsandman: Simplest form:

``` python
try:
    level = getattr(logging, config.get('sentry_logging_level'))
except (AttributeError, TypeError):
    level = _DEFAULT_LOGGING_LEVEL
```

legalsylvain:
@bwrsandman : 
1/ In my production server there is not so many warnings. And some of them are interesting like :

```
WARNING DATABASE openerp.modules.loading: 
    Model hr.expense.line is declared but cannot be loaded! 
    (Perhaps a module was partially removed or renamed)
```

or

```
WARNING DATABASE openerp.modules.module:
    module custom_module: module not found
```

For other irrelevant warning, so 'mute' / 'ignore' functions in Sentry could be used; for example : 

```
WARNING DATABASE openerp.addons.base.ir.ir_translation:
    module custom_module: no translation for language fr
```

2/ Thanks for the part of code you proposed !

pedrobaeza:
:+1:  Thanks.

Put it in kamikaze mode with DEBUG level, hehe

legalsylvain:
@pedrobaeza. ;-) 
In fact this is an easy and good way to know if your sentry server react correctly when it receives a lot of logs.

bwrsandman:
:+1:
Your contribution is appreciated!

guewen:
:+1:

PR #100, opened 2014-12-19
Add defer_parent_store_computation module

bwrsandman:
Module adds option to import form which allows importing with the
'defer_parent_store_computation' key in context which speeds up imports
tremendously.

As you can see in this graph, trying to import 500 account.accounts not only takes a lot of time, but the performance gets worse after a few accounts have been added. The deferred method (red line) is much much quicker.

importing accounts three methods

Huge credit goes to @pedrobaeza who taught me this trick.

The results can be re-created with these lines with a large `data`:

``` python
# Regular call
account_model.load(cr, uid, fields, data, context=context)

# Chunks method
def chunks(l, n):
    """ Yield successive n-sized chunks from l.
    """
    for i in xrange(0, len(l), n):
        yield l[i:i+n]

for chunk in chunks(data, 100):
    account_model.load(cr, uid, fields, chunk, context=context)


# Deferred parent store method
context['defer_parent_store_computation'] = True
account_model.load(cr, uid, fields, data, context=context)
account_model._parent_store_compute(cr)
```

### DIFF
--- a/defer_parent_store_computation_import/__init__.py
+++ b/defer_parent_store_computation_import/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2010 - 2014 Savoir-faire Linux
+#    (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import ir_import
--- a/defer_parent_store_computation_import/__openerp__.py
+++ b/defer_parent_store_computation_import/__openerp__.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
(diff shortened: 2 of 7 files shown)


foutoucour:
:+1: 
You might want to include the plot in the module to illustrate the case.

bwrsandman:
Where? Not sure how to add a graph to the module description.

max3903:
you can add screenshots ;)

dreispt:
@bwrsandman like this: https://github.com/odoo/odoo/tree/8.0/addons/crm/static/description

dreispt:
@bwrsandman  PR # 100 --> Achievement Unlocked :smile:

bwrsandman:
@dreispt Thanks! Do you know if this works in 7.0?
There is no static/description in the 7.0 crm.

bwrsandman:
> @bwrsandman PR # 100 --> Achievement Unlocked :smile:

:laughing:

dreispt:
Maybe not in 7.0, but there you can place screenshots in an `images/`directory. Example: https://github.com/odoo/odoo/tree/7.0/addons/hr/images

bwrsandman:
@dreispt I can do that, but they don't seem to appear in the description:
http://35462-7-0-10984a.runbot6.odoo.com/?db=35462-7-0-10984a-all&ts=1419003773989#id=114&view_type=form&model=ir.module.module&menu_id=59&action=38

pedrobaeza:
I think this fits better as an OCB/odoo patch in base_import module, because:
- It can be automatically determined if _parent_store variable is set.
- It doesn't interfere in any process.
- It doesn't have any side effects.

bwrsandman:
More graphs:

importing 23000 accounts

foutoucour:
You really need to find out how to add them to the module :)

pedrobaeza:
Hi, @bwrsandman, what do you think about my comment?

bwrsandman:
I think there may be other cases where you may want to do deferred method.

An OCB path sounds great, where is it?
An odoo patch isn't guarenteed to be released, especially not for 7.0.

The module adds functionality which is missing. Until it's officially supported, this perf enhansement is available to the community.

pedrobaeza:
There isn't still any OCB patch about this. Would you mind?

hbrunn:
I think this is perfectly fine as a module, :+1:

guewen:
:+1:

PR #103, opened 2014-12-27
Corrected the icon.jpg location and made as per v8 standard

JayVora-SerpentCS:


pedrobaeza:
Thanks :+1:

dreispt:
Isn't there a policy for no partner branding on OCA modules?
To comply the icon should be removed.

pedrobaeza:
Oh, I didn't see the icon was the Serpent logo. Daniel is right, you cannot put a brand image to the modules.

yvaucher:
So this is a :-1:

JayVora-SerpentCS:
Noted, Apologies. May be, we should put oca icon!

PR #109, opened 2015-01-02
initial german translations

rschnapka:


auth_admin_passkey/i18n/de.po
```diff
@@ -0,0 +1,123 @@
...
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+"Language: de\n"
+"X-Generator: Lokalize 1.5\n"
+
+#. module: auth_admin_passkey
+#: code:addons/auth_admin_passkey/model/res_users.py:66
+#, python-format
+msgid "<pre>User with login '%s' has the same password as you.</pre>"
+msgstr ""
```
pedrobaeza: Why this line feed?

cron_run_manually/i18n/de.po
```diff
@@ -0,0 +1,51 @@
...
+#: code:addons/cron_run_manually/model/ir_cron.py:52
+#, python-format
+msgid "Only the admin user is allowed to execute inactive cron jobs manually"
+msgstr "Nur der Administrator darf inaktive Cron-Aufgaben manuell ausführen"
+
+#. module: cron_run_manually
+#: model:_description:0
+#: model:ir.model,name:cron_run_manually.model_ir_cron
+msgid "ir.cron"
+msgstr ""
```
pedrobaeza: Translate this, although it is also `ir.cron`

auth_admin_passkey/i18n/de.po
pedrobaeza: It is still there.

pedrobaeza:
Strange the problem you're having with Lokalize...

:+1:

rschnapka:
maybe it's just me :-) as long as it gets fixed...

yvaucher:
:+1:

rschnapka:
1. Is this runbot-problem still due to my not approved license agreement?
2. po-editor problem ist not restricted to 'Lokalize' - 'PO Edit' also breaks the lines with excess linefeeds when they are longer than ~80 chars...

pedrobaeza:
No, runbot it's in beta state, and thus it breaks frequently.

I see the problem with localization tools. I always edit in raw mode, but I admit that they are more confortable. I remember that in past, these line feeds are parsed and introduces in texts (because the PO parser is a custom one). That's why I tell you to change, but you can try to see if this behaviour remains.

PR #111, opened 2015-01-12
add module mail_compose_message_default_attachment

kmatthes:
This module easily allows to add attachment files when composing an email.

### DIFF
--- a/mail_compose_message_default_attachments/__openerp__.py
+++ b/mail_compose_message_default_attachments/__openerp__.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#   Copyright (C) 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU Affero General Public License as
+#   published by the Free Software Foundation, either version 3 of the
+#   License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I see the module, but I don't see where you define the attachment to use. Is this a kind of patch?

kmatthes:
Yeah it's a patch of mail.compose.message to add attachment_ids (given as default_attachment_ids in context) to email.

pedrobaeza:
You should propose it then as a patch in OCB and Odoo.

mail_compose_message_default_attachments/mail_compose_message_default_attachments.py
```diff
@@ -0,0 +1,36 @@
...
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+from openerp.osv import orm
+
+
+class mail_compose_message(orm.Model):
+    """Patch mail.compose.message to add attachment_ids (given as \
```
yvaucher: With `"""` you don't need `\`

mail_compose_message_default_attachments/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{
+    'name': 'Compose Message Default Attachments',
+    'version': '1.0',
+    'category': 'Patch',
+    'description': "Patch mail.compose.message to add attachment_ids (given \
```
yvaucher: Please use `"""` for description string

OSguard:
@kmatthes can you fix to comments

@pedrobaeza having it as module we can make it requiere for other module need this function. Having this as patch (if it will merged) we need to add a message to description 'you need this patch that this module works' but maybe we can add a patch to odoo/master

pedrobaeza:
@OSguard, the idea is to get this into upstream and OCB branches, so you don't need to patch nothing. Maybe this won't get to both, but if you don't try, you won't know it. If not, then you can continue with this, and call the module with _fix at the end to notice what is this.

pedrobaeza:
Closing as no answer received

PR #115, opened 2015-01-15
Add module base_export_email

laetitia-gangloff:
Hello,

I propose a module that send by email an export data.
To get this, I add a new server action type (Export data by email) and a set of fields related to this type (filter, template, fields to export).

Exemple of use : in  a cron for a periodic export of some data to see the evolution.

### DIFF
--- a/base_export_email/__init__.py
+++ b/base_export_email/__init__.py
@@ -3,14 +3,6 @@
 #
 #    Authors: Laetitia Gangloff
 #    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
-#    All Rights Reserved
-#
-#    WARNING: This program as such is intended to be used by professional
-#    programmers who take the whole responsibility of assessing all potential
-#    consequences resulting from its eventual inadequacies and bugs.
-#    End users who are looking for a ready-to-use solution with commercial
-#    guarantees and support are strongly advised to contact a Free Software
-#    Service Company.
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
--- a/base_export_email/__init__.py
+++ b/base_export_email/__init__.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Laetitia Gangloff
+#    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
+#    All Rights Reserved
+#
+#    WARNING: This program as such is intended to be used by professional
+#    programmers who take the whole responsibility of assessing all potential
+#    consequences resulting from its eventual inadequacies and bugs.
+#    End users who are looking for a ready-to-use solution with commercial
+#    guarantees and support are strongly advised to contact a Free Software
+#    Service Company.
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import ir_actions
--- a/base_export_email/__openerp__.py
+++ b/base_export_email/__openerp__.py
@@ -22,7 +22,7 @@
 {
     "name": "Base export email addon",
     "version": "0.1",
-    "author": "ACSONE SA/NV",
+    "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "category": "Other",
     "website": "http://www.acsone.eu",
     "depends": ['email_template',
--- a/base_export_email/__openerp__.py
+++ b/base_export_email/__openerp__.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Laetitia Gangloff
+#    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
+#    All Rights Reserved
+#
+#    WARNING: This program as such is intended to be used by professional
+#    programmers who take the whole responsibility of assessing all potential
+#    consequences resulting from its eventual inadequacies and bugs.
+#    End users who are looking for a ready-to-use solution with commercial
+#    guarantees and support are strongly advised to contact a Free Software
+#    Service Company.
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Base export email addon",
+    "version": "0.1",
+    "author": "ACSONE SA/NV",
+    "category": "Other",
+    "website": "http://www.acsone.eu",
+    "depends": ['email_template',
+                ],
+    "description": """
+
+Base export email addon
+=========================
+
+- Add server action type to send by email a data export
+
+""",
+    "data": ["ir_actions.xml"],
+    "demo": [],
+    "test": [],
+    "active": False,
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": False,
+    "application": False,
+}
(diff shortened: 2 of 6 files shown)


base_export_email/__openerp__.py
```diff
@@ -0,0 +1,54 @@
...
+    "website": "http://www.acsone.eu",
+    "depends": ['email_template',
+                ],
+    "description": """
+
+Base export email addon
+=========================
+
+- Add server action type to send by email a data export
+
```
pedrobaeza, line 45: Please document here the example put in the PR description, and name each of the values to use.

pedrobaeza:
Please provide a base template in the module for sending the mails.

laetitia-gangloff:
But mail template depends of the objet in the server action.

Maybe can I provided demo data to show the example ?

base_export_email/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+#    Copyright (c) 2015 Acsone SA/NV (http://www.acsone.eu)
+#    All Rights Reserved
+#
+#    WARNING: This program as such is intended to be used by professional
+#    programmers who take the whole responsibility of assessing all potential
+#    consequences resulting from its eventual inadequacies and bugs.
+#    End users who are looking for a ready-to-use solution with commercial
+#    guarantees and support are strongly advised to contact a Free Software
+#    Service Company.
+#
```
sbidoul, line 6: Use standard header, remove All Rights Reserved

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,157 @@
...
+                    'ir.filters', string='Filter', ondelete='restrict'),
+                'email_template_id': fields.many2one(
+                    'email.template', string='Template', ondelete='restrict'),
+                'saved_export_id': fields.many2one(
+                    'ir.exports', string='Saved export', ondelete='restrict'),
+                'fields_to_export': fields.text(
+                    'Fields to export',
+                    help="The list of fields to export. \
+                          The format is ['name', 'seller_ids/product_name']")
+                }
```
sbidoul: The relationship between Saved export and Fields to export may not be obvious for the user.

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,157 @@
...
+            model = self.pool['ir.model'].browse(cr, uid, model_id,
+                                                 context=context)
+            data.update({'model': model.model})
+        return {'value': data}
+
+    def _search_data(self, cr, uid, action, context=None):
+        """
+         search data to export
+        """
+        obj_pool = self.pool[action.model]
```
sbidoul, line 92: Please elaborate the docstring, or remove it. A docstring that does not say anything more than the method name is useless.

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,157 @@
...
+            action.filter_id.context) or context
+        return obj_pool.search(cr, uid, domain,
+                               offset=0, limit=False, order=False,
+                               context=ctx)
+
+    def _export_data(self, cr, uid, obj_ids, action, context=None):
+        """
+        export data
+        """
+        obj_pool = self.pool[action.model]
```
sbidoul, line 105: idem

base_export_email/ir_actions.py
sbidoul: Can the export format CSV/XLS be selected?

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,157 @@
...
+        mail_compose = self.pool['mail.compose.message']
+        values = mail_compose.onchange_template_id(
+            cr, uid, 0, action.email_template_id, 'comment',
+            action.model, 0, context=context)['value']
+        values['partner_ids'] = [
+            (4, partner_id) for partner_id in values.pop('partner_ids',
+                                                         [])
+        ]
+        export = main.CSVExport()
+        filename = export.filename(action.model)
```
sbidoul: Hooking on the web controller for a backend module does not look right. Is there not a better way?

pedrobaeza:
You can provide a demo/base mail template that it's selected by default with noupdate, and people can customize this template or duplicate to take it as base for making a more ellaborated template.

lmignon:
It would be nice to put the module description in a README.rst file. Apart from that, LGTM :+1:

base_export_email/ir_actions_view.xml
```diff
@@ -0,0 +1,44 @@
...
+                                                          ('state','!=','dummy'),
+                                                          ('state','!=','sms'),
+                                                          ('state','!=','code'),
+                                                          ('state','!=','loop'),
+                                                          ('state','!=','trigger'),
+                                                          ('state','!=','object_copy'),
+                                                          ('state','!=','client_action'),
+                                                          ('state','!=','email'),
+                                                          ('state','!=','sms'),
+                                                          ('state','!=','other')]}</attribute>
```
sbidoul, line 26: This long list look fragile. What happens if other modules extend the list?

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,162 @@
...
+        return mail_compose.send_mail(cr, uid, [compose_id], context=context)
+
+    def run(self, cr, uid, ids, context=None):
+        """
+        If the state of an action is export_email,
+        export data related to the configuration and send the result by email
+        """
+        for action in self.browse(cr, uid, ids, context):
+            if action.state == 'export_email':
+                ids.remove(action.id)
```
sbidoul: Out of curiosity, what is the purpose of this remove?

base_export_email/ir_actions_view.xml
laetitia-gangloff, line 26: This long list comes from the standard.
I need it just to add "('state', '!=', 'export_email')"

A better solution is maybe to rewrite it with "('state', '=', ...)" but I tried to stay as to the standard.

base_export_email/ir_actions.py
laetitia-gangloff: In my opinion it is not usefull to evaluate the expression.
The super, do this evaluation as first step of the for and then dispatch call following the action.state.
I think, it is not usefull to send id of action in state 'export_email' because the case is already processed, so I remove if from the list.

sbidoul: Anyway, I think it's cleaner to create a new run_action_export_mail(_multi) instead of overriding run().

base_export_email/ir_actions.py
```diff
@@ -0,0 +1,162 @@
...
+    _inherit = "ir.actions.server"
+
+    def __init__(self, pool, cr):
+        super(actions_server, self).__init__(pool, cr)
+        # add state 'export_email'
+        self._columns['state'].selection.append(('export_email',
+                                                 _('Export data by email')))
+        return
+
+    _columns = {'model': fields.related('model_id', 'model',
```
sbidoul: nitpicking on style: it's nicer to have a newline after the opening {

base_export_email/ir_actions.py
sbidoul: Please ignore my previous comment about run_action_export_email... that's for 8.0 only.

base_export_email/ir_actions_view.xml
pedrobaeza, line 26: For better compatibility, you can overwrite fields_view_get and do something similar than this:
https://github.com/odoomrp/odoomrp-wip/blob/8.0/sale_pricelist_rules/models/sale.py#L150

guewen:
The overwrite of fields_view_get as proposed by @pedrobaeza could be nice.
Still nice module, LGTM :+1:

laetitia-gangloff:
Thank you !

I tried the overwrite proposed by @pedrobaeza but for the moment it doesn't work :(

I have not debug all the chain, but there is a problem somewhere.

I add the following method and remove the view part concerning the field srcmodel_id.

``` python
    def fields_view_get(self, cr, uid, view_id=None, view_type='form',
                        context=None, toolbar=False, submenu=False):
        res = super(actions_server, self).fields_view_get(
            cr, uid, view_id=view_id, view_type=view_type, context=context,
            toolbar=toolbar, submenu=submenu)
        if view_type == 'form':
            doc = etree.XML(res['arch'])
            nodes = doc.xpath("//field[@name='srcmodel_id']")
            for node in nodes:
                #==============================================================
                # attrs = doc.attrib.get(
                #     "attrs",
                #     "'required':[('state', '!=', 'export_email')]"
                # ).replace("'required':[",
                #           "'required':[('state', '!=', 'export_email'),")
                #==============================================================
                node.set("attrs",
                         "{'required':[('state', '!=', 'export_email')]}")
            res['arch'] = etree.tostring(doc)
        return res
```

I  think the problem is maybe somewhere in the inheritance.

hbrunn:
@laetitia-gangloff the problem here is that at this stage, you neet to write the modifiers attribute, not attrs: https://github.com/OCA/OCB/blob/7.0/openerp/osv/orm.py#L107-L140.

Looks good to me otherwise

gurneyalex:

laetitia-gangloff:
@gurneyalex done, thanks!

dreispt:
Is this good to merge?

qtheuret:
Thanks for this module. :+1:

base_export_email/__openerp__.py
```diff
@@ -0,0 +1,56 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Base export email addon",
+    "version": "0.1",
+    "author": "ACSONE SA/NV",
```
gurneyalex: Can you add "Odoo Community Association (OCA)" to the authors list?

laetitia-gangloff:
@gurneyalex authors list updated.

gurneyalex:
:+1:

PR #116, opened 2015-01-15
Add option to see original error in odoo

bwrsandman:
Sometimes you want to be able to see the error in the traceback in odoo even if sentry is connected.
This adds a collapsible box to see the original error with traceback when sentry is enabled.

capture d ecran de 2015-01-15 15 20 22
capture d ecran de 2015-01-15 15 20 10


bwrsandman:
Not an expert in css, so if anyone can suggest an improvement in my css, I would appreciate it.

foutoucour:
:+1:

pedrobaeza:
Me neither an expert, so for me :+1:

yvaucher:
:+1:

PR #117, opened 2015-01-18
[ADD] cron_run_manually: Tool for allowing to run manually a programmed task

pedrobaeza:


hbrunn:
:+1:  (verified that travis errors are unrelated)

lmignon:
:+1:

cron_run_manually/model/ir_cron.py
```diff
@@ -0,0 +1,82 @@
...
+    _inherit = 'ir.cron'
+
+    def run_manually(self, cr, uid, ids, context=None):
+        """
+        Run a job from the cron form view.
+
+        Cut and paste of code snippets from addons/base/ir/ir_cron.py
+        """
+        logger = logging.getLogger('cron_run_manually')
+        cr.execute(
```
foutoucour: why do you have to do sql query here? Can't you get the model instead?

StefanRijnhart:
Checked any changes compared to 7.0: just a single import was changed and Spanish translations were added. Thanks for the port! :+1:

pedrobaeza:
I merge then. Thanks for the reviews.

PR #119, opened 2015-01-22
[ADD] Module 'auditlog' - A substitute to the deprecated 'audittrail' module for Odoo 8

sebalix:
`auditlog` is a rewrite of the `audittrail` module to work with the Odoo 8 framework.

At this time, it logs create, write and delete operations.

What is missing:
- log read operations (encountered some problems with old/new API of the `read()` method, I had not enough time to look deeper) ;
- log workflows (is it really needed?) ;
- log only operations triggered by some users (currently it logs all users) => easy ;
- equivalent of the audittrail wizard to query logs => easy;

What improvements could be made in the future, making logs more readable:
- group logs by user HTTP query (thanks to werkzeug) ;
- group user HTTP queries by user session ;

About the performance, not really tested, but as a rule of thumb with a user operation (wizard) which triggers about 80/90 creates and write on 5 different data models, without logs it takes around 6-7 seconds, and with log activated on all of them, around 18 seconds.


sebalix:
NB: the build fails on the `base_field_serialized_test_model` module

auditlog/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': "Audit Log",
+    'description': """
```
dreispt: Could you put the description in a README.rst instead?

auditlog/models/rule.py
```diff
@@ -0,0 +1,398 @@
...
+        """Instanciate an unlink method that log its calls."""
+        @api.multi
+        def unlink(self, **kwargs):
+            rule_model = self.env['auditlog.rule']
+            rule_model.sudo().create_logs(
+                self.env.uid, self._name, self.ids, 'unlink')
+            return unlink.origin(self, **kwargs)
+        return unlink
+
+    def create_logs(self, uid, res_model, res_ids, method,
```
rvalyi, line 284: create_logs is more than 70 lines: it's a bit large... Isn't there any useful method you could extract to make overriding easier? (may be not but it's worth thinking about it).

auditlog/__openerp__.py
sebalix: Done, I added a copyright section as seen in another module.

dreispt:
Please also include the `pot` file.

auditlog/models/rule.py
sebalix, line 284: Done.

rvalyi, line 284: looks good. Thanks.

auditlog/models/log.py
```diff
@@ -0,0 +1,60 @@
...
+    log_id = fields.Many2one(
+        'auditlog.log', string=u"Log", ondelete='cascade')
+    old_value = fields.Text(u"Old Value")
+    new_value = fields.Text(u"New Value")
+    old_value_text = fields.Text(u"Old value Text")
+    new_value_text = fields.Text(u"New value Text")
+    field_name = fields.Char(u"Technical name", size=64)
+    field_description = fields.Char(u"Description", size=64)
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
dreispt: You can remove these vim lines.

auditlog/models/rule.py
```diff
@@ -0,0 +1,432 @@
...
+    user_ids = fields.Many2many(
+        'res.users',
+        'audittail_rules_users',
+        'user_id', 'rule_id',
+        string=u"Users",
+        help=u"if  User is not added then it will applicable for all users")
+    log_read = fields.Boolean(
+        u"Log Reads",
+        help=(u"Select this if you want to keep track of read/open on any "
+              u"record of the model of this rule"))
```
dreispt, line 78: `readonly=True` since it's not supported yet?

sebalix, line 78: Well, I have set the field invisible on the view while it's not implemented.

auditlog/models/log.py
sebalix: Done

auditlog/views/auditlog_view.xml
```diff
@@ -0,0 +1,191 @@
...
+            </field>
+        </record>
+
+        <record model="ir.actions.act_window" id="action_auditlog_rule_tree">
+            <field name="name">Rules</field>
+            <field name="res_model">auditlog.rule</field>
+            <field name="type">ir.actions.act_window</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">tree,form</field>
+            <field name="context">{'search_default_draft': 1}</field>
```
dreispt: This search default does nothing, since there is no field or filter named `draft`. My advice is to remove it and use just `{}`.

auditlog/views/auditlog_view.xml
```diff
@@ -0,0 +1,191 @@
...
+        </record>
+
+        <record id="view_auditlog_rule_search" model="ir.ui.view">
+            <field name="name">auditlog.rule.search</field>
+            <field name="model">auditlog.rule</field>
+            <field name="arch" type="xml">
+                <search string="Rules">
+                    <field name="name"/>
+                    <filter domain="[('state','=','draft')]" string="Draft"/>
+                    <filter domain="[('state','=','subscribed')]" string="Subscribed"/>
```
dreispt: It would be nice to add `name=` to the filters.

dreispt:
Build error is caused by `base_field_serialized`.
Minor suggestions but it's a :+1:

auditlog/models/rule.py
dreispt, line 78: Fine.

auditlog/views/auditlog_view.xml
sebalix: Fixed.

auditlog/views/auditlog_view.xml
sebalix: Added.

auditlog/models/log.py
```diff
@@ -0,0 +1,58 @@
...
+
+    name = fields.Char("Resource Name", size=64)
+    model_id = fields.Many2one(
+        'ir.model', string=u"Model")
+    res_id = fields.Integer(u"Resource ID")
+    user_id = fields.Many2one(
+        'res.users', string=u"User")
+    method = fields.Char(u"Method", size=64)
+    timestamp = fields.Datetime(
+        u"Date", default=lambda *a: time.strftime('%Y-%m-%d %H:%M:%S'))
```
hbrunn: is this any different from create_date that you get for free?

sebalix: `create_date` could do the trick indeed. Some fields were copied from the original `audittrail` module, I removed some but yes, this one could be removed too.

sebalix: `timestamp` replaced by `create_date`

hbrunn:
Nice, thanks! I proposed some small additions here: https://github.com/osiell/server-tools/pull/1

auditlog/models/rule.py
```diff
@@ -0,0 +1,433 @@
...
+    """Calculate the difference between two dictionaries as:
+    (1) items added
+    (2) items removed
+    (3) keys same in both but changed values
+    (4) keys same in both and unchanged values
+    """
+    def __init__(self, current_dict, past_dict):
+        self.current_dict, self.past_dict = current_dict, past_dict
+        self.set_current = set(current_dict.keys())
+        self.set_past = set(past_dict.keys())
```
gurneyalex: remove calls to `keys()`: no need to create an extra list here.

gurneyalex:
Please add some automated tests.

gurneyalex:
I'd like to see something that

gurneyalex:
Sorry I wanted a "cancel" button but I hit "close and comment"...

hbrunn:
@gurneyalex I picked up your suggestions in https://github.com/osiell/server-tools/pull/4

hbrunn:
:+1:

StefanRijnhart:
Thanks @sebalix. Did you file a CLA to the OCA? See http://odoo-community.org/page/cla

auditlog/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+from . import models
+
+
+def pre_init_hook(cr):
+    cr.execute("SELECT 1 FROM pg_class WHERE relname = 'audittrail_rule'")
+    if cr.fetchall():
+        migrate_from_audittrail(cr)
+
+
+def migrate_from_audittrail(cr):
```
StefanRijnhart, line 31: Excellent to migrate existing audittrail rules. You should claim this feature in the README.rst.

auditlog/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+    'depends': [
+        'base',
+    ],
+    'data': [
+        'security/ir.model.access.csv',
+        'views/auditlog_view.xml',
+    ],
+    'application': True,
+    'installable': True,
+    'active': False,
```
StefanRijnhart: Module model has no field 'active'. Please remove to prevent problems in the future.

auditlog/__init__.py
sebalix, line 31: Thanks to @hbrunn for this.
I will add a note in the README.

sebalix:
@StefanRijnhart : yes we should have an entity CLA signed by @ericflaux (CEO ABF OSIELL)

auditlog/models/log.py
```diff
@@ -0,0 +1,54 @@
...
+    line_ids = fields.One2many(
+        'auditlog.log.line', 'log_id', string=u"Fields updated")
+
+
+class auditlog_log_line(models.Model):
+    _name = 'auditlog.log.line'
+    _description = "Auditlog - Log details (fields updated)"
+
+    field_id = fields.Many2one(
+        'ir.model.fields', string=u"Field", required=True)
```
StefanRijnhart: Would you consider making this ondelete="cascade"? I'm thinking of migrations between major versions, in which the datamodel can change drastically sometimes.

auditlog/models/log.py
```diff
@@ -0,0 +1,54 @@
...
+    log_id = fields.Many2one(
+        'auditlog.log', string=u"Log", ondelete='cascade')
+    old_value = fields.Text(u"Old Value")
+    new_value = fields.Text(u"New Value")
+    old_value_text = fields.Text(u"Old value Text")
+    new_value_text = fields.Text(u"New value Text")
+    field_name = fields.Char(u"Technical name", size=64)
+    field_description = fields.Char(u"Description", size=64)
```
StefanRijnhart: Why duplicate this data on every log? Given that the field is required you could make this a function field for display purposes (same for name)

auditlog/models/rule.py
```diff
@@ -0,0 +1,446 @@
...
+        for rule in self:
+            # Remove the shortcut to view logs
+            act_window = act_window_model.search(
+                [('name', '=', 'View Log'),
+                 ('res_model', '=', 'auditlog.log'),
+                 ('src_model', '=', rule.model_id.model)])
+            if act_window:
+                value = 'ir.actions.act_window,%s' % act_window.id
+                act_window.unlink()
+            if value:
```
StefanRijnhart: Strange flow, fixed when you remove this condition and leave the indentation the same. Allows you to remove the declaration of 'value' in line 427 too.

StefanRijnhart:
Some nags, but great work overall! I'll check out the CLA situation further.

StefanRijnhart:
The OCA board just decided that the ECLA will cover any work by people who are a contact person of the organization with the CLA. So I added you as a contact of the Osiell organization in the OCA Odoo instance.

sebalix:
An error is raised after that some data are cached to improve performance on further logs creation, just added a test to raise it:

``` python
OperationalError: Unable to use a closed cursor
```

@hbrunn have you an idea about it?

hbrunn:
@sebalix I think that's because I unnecessarily put the whole browse record into the cache. Change that to only cache ids and strings, then it should work

sebalix:
I have nothing more to add currently, all seems to work now.

StefanRijnhart:
Tests reveal a bug:
If I have a delete or write log on res.partner, I cannot unlink contacts anymore from the partner form. It gives me "MissingError. One of the documents you are trying to access has been deleted, please try again after refreshing. "

One question: A create log entry lists fields with empty values. Can/should these be filtered out?

sebalix:
I think your partner is also a res.users on which logs are linked (`user_id` M2O on `auditlog.log`). I have no problem to delete a partner which had logs related to it.
Should we delete on cascade such logs when we remove a user account, or maintain a weak link between logs and users?

Regarding the create log behaviour, it was easier to build the log like that, and I thought it could be useful to have the complete changes (including computed field values). But it could be improve later indeed.

StefanRijnhart:
To clarify, I have a problem removing a partner through the embedded
contacts form on the main partner. No users involved. To reproduce:
Create a partner. In this partner's form view: create a contact. Save
the partner. Edit the parter form view. Remove the partner's contact.
Try and save the partner.

sebalix:
Okay! I reproduced the error, I will check that.

ruifranco:
The question might be a candid one but: where/how can I get this module so that I can test it. A whole line of work is pending on getting the "audittrail" back :|

StefanRijnhart:
@ruifranco "git clone https://github.com/osiell/server-tools -b 8.0". You will get a 'server-tools' directory with the auditlog module in there. Link, copy or move this auditlog module into your addons-path. Update the module list in Odoo (need 'Technical features' checkbox on your admin user), go to installed modules menu item, remove the filter on 'installed' modules, search for the module and install.

ruifranco:
Thanks for the link Stefan

sebalix:
It's fixed (lint check fails on `server_environment` module). As a workaround the deleted resources will have a 'DELETED' text value in logs. It would need changes to get the actual values before the unlink operation, but it's better than nothing.

StefanRijnhart:
Thank you for fixing this issue perfectly! :+1:

StefanRijnhart:
Counting 3 thumbs up and you also made @gurneyalex happy with automated tests so proceeding with the merge.

sebalix:
Thank you!

PR #121, opened 2015-01-27
[ADD] port fetchmail_attach_from_folder

hbrunn:
# Email gateway - folders

Adds the possibility to attach emails from a certain IMAP folder to objects,
ie partners. Matching is done via several algorithms, ie email address, email
address's domain or the original Odoo algorithm.

This gives a simple possibility to archive emails in Odoo without a mail
client integration.
# Configuration

In your fetchmail configuration, you'll find a new list field `Folders to 
monitor`. Add your folders here in IMAP notation (usually something like
`INBOX.your_folder_name.your_subfolder_name`), choose a model to attach mails
to and a matching algorithm to use.
## Exact mailaddress

Fill in a field to search for the email address in `Field (model)`. For
partners, this would be `email`. Also fill in the header field from the email
to look at in `Field (email)`. If you want to match incoming mails from your
customers, this would be `from`. You can also list header fields, so to match
partners receiving this email, you might fill in `to,cc,bcc`.
## Domain of email addresses

Match the domain of the email address(es) found in `Field (email)`. This would
attach a mail to `test1@example.com` to a record with `Field (model)` set to
`test2@example.com`. Given that this is a fuzzy match, you probably want to
check `Use 1st match`, because otherwise nothing happens if multiple possible
matches are found.
## Odoo standard

This is stricly speaking no matching algorithm, but calls the model's standard
action on new incoming mail, which is usually creating a new record.
# Usage

A widespread configuration is to have a shared mailbox with several folders,
i.e. one where users drop mails they want to attach to partners. Let this
folder be called `From partners`. Then create a folder configuration for your
server with path `"INBOX.From partners"` (note the quotes because of the space,
this is server dependent). Choose model `Partners`, set `Field (model)` to
`email` and `Field (email)` to `from`. In `Domain`, you could fill in
`[('customer', '=', True)]` to be sure to only match customer records.

Now when your users drop mails into this folder, they will be fetched by Odoo
and attached to the partner in question. After some testing, you might want to
check `Delete matches` in your folder configuration so that this folder doesn't grow indefinitely.


fetchmail_attach_from_folder/wizard/attach_mail_manually.py
```diff
@@ -87,7 +80,7 @@ def attach_mails(self, cr, uid, ids, context=None):
                 connection.select(this.folder_id.path)
                 result, msgdata = connection.fetch(mail.msgid, '(RFC822)')
                 if result != 'OK':
-                    logger.error('Could not fetch %s in %s on %s' % (
+                    logging.error('Could not fetch %s in %s on %s' % (
```
gurneyalex: don't use % when calling logging methods.

fetchmail_attach_from_folder/wizard/attach_mail_manually.py
```diff
@@ -54,15 +49,13 @@ def default_get(self, cr, uid, fields_list, context=None):
                 None,
                 'FLAGGED' if folder.flag_nonmatching else 'UNDELETED')
             if result != 'OK':
-                logger.error('Could not search mailbox %s on %s' % (
+                logging.error('Could not search mailbox %s on %s' % (
```
gurneyalex: don't use % when calling logging methods

fetchmail_attach_from_folder/wizard/attach_mail_manually.py
```diff
@@ -54,15 +49,13 @@ def default_get(self, cr, uid, fields_list, context=None):
...
                     folder.path, folder.server_id.name))
                 continue
-            attach_mail_manually_mail._columns['object_id'].selection = [
-                (folder.model_id.model, folder.model_id.name)]
             for msgid in msgids[0].split():
                 result, msgdata = connection.fetch(msgid, '(RFC822)')
                 if result != 'OK':
-                    logger.error('Could not fetch %s in %s on %s' % (
+                    logging.error('Could not fetch %s in %s on %s' % (
```
gurneyalex: don't use % on logging methods

fetchmail_attach_from_folder/model/fetchmail_server.py
```diff
@@ -75,111 +68,109 @@ def fetch_mail(self, cr, uid, ids, context=None):
...
+    def handle_folder(self, connection, folder):
         '''Return ids of objects matched'''
 
         matched_object_ids = []
 
-        for this in self.browse(cr, uid, ids, context=context):
-            logger.info('start checking for emails in %s server %s',
-                        folder.path, this.name)
+        for this in self:
+            logging.info(
```
gurneyalex: throughout this module you are using the `logging` module shortcut methods, which by default use the root logger. This is unusual in Odoo addons and can make log filtering configuration more difficult (AFAIR if I change the logging level of the root logger, this has an impact on all its descendents).

I think you should use a named logger. If you don't want to use the `_logger = logging.getLogger(__name__)` idiom, fine, but you still should not use the root logger.

hbrunn:
@gurneyalex thanks for pointing this out! I also added tests, you should like that

StefanRijnhart:
Great work, thanks. :+1:

pedrobaeza:
Another great addition from your part!

:+1:

Issue #125, opened 2015-02-05
server-environment: spurious messages from lsb_release are printed to the terminal


lepistone:
On startup, the module `server_environment` collects about the system and allows the user to display it.

Every time the message `No LSB modules are available.` is printed out on my ubuntu system.

This could be used by the module to show infomation to the user, but it makes no sense to print out that string. This gets especially annoying when running automated tests, as the message will be shown among the tests results.

PR #126, opened 2015-02-05
capture stderr as well as stdout when diagnosing

lepistone:
This fixes OCA/server-tools#125 because in some Ubuntu systems
lsb_release outputs to stderr as well as stdout. This message will be
printed to console, not logged nor shown among server infomation. That
way the message is lost and only pollutes unit tests output.

### DIFF
--- a/server_environment/system_info.py
+++ b/server_environment/system_info.py
@@ -30,7 +30,8 @@ from openerp.tools.config import config
 
 def _get_output(cmd):
     bindir = config['root_path']
-    p = subprocess.Popen(cmd, shell=True, cwd=bindir, stdout=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why not subprocess.STDERR?

lepistone:
@pedrobaeza I gave a look at the doc at https://docs.python.org/2/library/subprocess.html and `subprocess.STDERR` it is not a choice you can make.

In the line after, with `communicate()[0]` we return stdout. If we did `stderr=PIPE` then stderr would go to `communicate()[1]` but then we wouldn't use it, so it would be discarded.

What I did is to return stderr together with stdout so that is displayed by the module. The other option would be to throw it away with `2> /dev/null`.

pedrobaeza:
OK, thanks for the explanation

:+1:

yvaucher:
:+1:

nbessi:
:+1:

PR #127, opened 2015-02-06
[ADD] 2 modules that allow you to manage your dns entries in DNSPod with odoo-connector

liulixia1990:
The DNS Connector module is a base framework that should allow integration with other providers.
Added 1 line in the .gitignore file to ignore .DS_Store files.


OSguard:
i suggest to move this to https://github.com/OCA/connector-interfaces as it depends on the connector framework

liulixia1990:
thanks very much,I have already moved this to https://github.com/OCA/connector-interfaces, and this is my pull request url: https://github.com/OCA/connector-interfaces/pull/12, please have a check.

PR #129, opened 2015-02-11
[IMP] Drop table with CASCADE option

clonedagain:
It should be relativly safe to use this option, as it should normally only drop foreign keys and views.

### DIFF
--- a/database_cleanup/model/purge_tables.py
+++ b/database_cleanup/model/purge_tables.py
@@ -37,45 +37,13 @@ class CleanupPurgeLineTable(orm.TransientModel):
         Unlink tables upon manual confirmation.
         """
         lines = self.browse(cr, uid, ids, context=context)
-        tables = [line.name for line in lines]
         for line in lines:
             if line.purged:
                 continue
 
-            # Retrieve constraints on the tables to be dropped
-            # This query is referenced in numerous places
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not sure if this is safe. @StefanRijnhart should say because he was the original developer and he will know the reason to put that block.

StefanRijnhart:
I fear a loss of consistency. The current code means to check this consistency by only allowing the removal of foreign keys between the tables that are going to be dropped, and not outside this table space. In other words: if the current method fails, then you probably have a live column in your database that is affected by the table drop and you want that to fail.

hbrunn:
One thing I can imagine where the current code fails is with views depending on the table. Or is this also a wanted failure?

StefanRijnhart:
Not a wanted feature, no. If this is the case, maybe it can be resolved by scheme introspection as well?

clonedagain:
OK I'll withdraw this proposal then, maybe the foreign keys that were left in my test databse are there for a reason, maybe a fault in the migration procedure or faulty module uninstalling.

clonedagain:
After more investigation, I'm now pretty sure the blocking foreign keys are artefacts of the previous migrations, ie. o2m fields changed into m2m.
So I'll manually delete the faulty keys.

PR #130, opened 2015-02-16
fix pep8 W503

gurneyalex:


lepistone:
:+1:

lepistone:
why "in progress" @nhomar ?

database_cleanup/model/purge_tables.py
```diff
@@ -110,9 +110,9 @@ def find(self, cr, uid, context=None):
             known_tables += [
                 column._sql_names(model_pool)[0]
                 for column in model_pool._columns.values()
-                if column._type == 'many2many'
+                if column._type == 'many2many' and
                 # unstored function fields of type m2m don't have _rel
-                and hasattr(column, '_rel')
+                hasattr(column, '_rel')
```
yvaucher: I think this is less readable

yvaucher: What about:

```
known_tables += [
    column._sql_names(model_pool)[0]
    for column in model_pool._columns.values()
    if (column._type == 'many2many' and
        # unstored function fields of type m2m don't have _rel
        hasattr(column, '_rel'))
]
```

lepistone:
@yvaucher for me it's :+1: both ways.

gurneyalex:
@yvaucher do you like it better this way?

yvaucher:
@gurneyalex Thanks for the change :+1:

yvaucher:
I fast track it

PR #135, opened 2015-03-02
[IMP] : run cron from tree view

AdilHoumadi:


pedrobaeza:
:+1:

hbrunn:
I suggest doing the view override only for groups that are supposed to run this, like `group_system` (you might add the same to the form view)

pedrobaeza:
Isn't it supposed that you didn't see that tree if you belong to that group?

hbrunn:
right, forget about it. :+1:

AdilHoumadi:
OCA runbot is not happy, is there any reason ? 

Thanks.

hbrunn:
we can ignore it for now

dreispt:
:+1:

PR #136, opened 2015-03-03
fix travis (flake8)

gurneyalex:


gurneyalex:
I had to use # noqa once because I could see no cleaner way to suppress the "import not at the top of the file" warning given the module structure.

lepistone:
No need for `#noqa` @gurneyalex: you can just put that line at the top and leave everything as is.

gurneyalex:
@lepistone are you sure this does not break anything? It will change the order of the elements in the list...

gurneyalex:
travis is green -> can be merged IMO

lepistone:
:+1: thanks

PR #144, opened 2015-03-26
Fix SQL error when an user logs in twice during a transaction

mdietrichc2c:
This PR is to add a new module, `base_login_date_improvement`.

When an user launches a long transaction in OpenERP with PostgreSQL >= 9.3, and opens a second window/tap to keep working, the method `res.users.login()` is called again, and modifies the `res.users` row; this creates a concurrency condition on the first transaction (as PostgreSQL is configured to be `REPEATABLE READ`), and raises an exception:

```
could not serialize access due to concurrent update
SQL statement "SELECT 1 FROM ONLY "public"."res_users" x WHERE "id" OPERATOR(pg_catalog.=) $1 FOR KEY SHARE OF x"
```

The module creates a new table, `res_users_login` and re-defines the field `login_date` as a function. The issue only appears in 9.3 and above, because the login method can write in `NO KEY UPDATE` instead of failing silentily like before.

Warning: the module is partially incompatible with `gamification`and `auth_openid`, since those 2 modules read the `login_date` in `res.users` using SQL. They will not crash, but not work appropriately.


mdietrichc2c:
I added the flag `nodrop=True` to `login_date` ; with this flag, the column will not be dropped from the table `res_users`, but will not be used anymore (which means updating the module via CLI or on Travis will work correctly).

mdietrichc2c:
Travis failed due to the Github DDoS...

vrenaville:
:+1: thanks for this fix

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    # Re-defining the login function in order to use the new table
+    def login(self, db, login, password):
+        if not password:
+            return False
+        user_id = False
+        cr = pooler.get_db(db).cursor()
+        try:
+            cr.autocommit(True)
```
gurneyalex: why are you using autocommit?

base_login_date_improvement/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+"re-logging" by opening a new tab changes the current res.user row,
+which creates concurrency issues with PostgreSQL in the first transaction.
+
+This creates a new table and a function field to avoid this.
+""",
+ "website": "http://camptocamp.com",
+ "depends": ['base'],
+ "data": ['security/ir.model.access.csv',
+          ],
+ "auto_install": False,
```
gurneyalex: do we want this auto_install: True?

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,109 @@
...
+                    cr.execute("SELECT login_dt "
+                               "FROM res_users_login "
+                               "WHERE user_id=%%s "
+                               "FOR %s NOWAIT" % update_clause,
+                               (user_id,), log_exceptions=False)
+                    # create login line if not existing
+                    result = cr.fetchone()
+                    if not result:
+                        cr.execute("INSERT INTO res_users_login "
+                                   "(user_id) VALUES (%s)", (user_id,))
```
gurneyalex: I was concerned at first with the race condition here. It is covered by the inner try...except block, although I feel catching Exception is a bit wide...

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,109 @@
...
+                    cr.execute("UPDATE res_users_login "
+                               "SET login_dt = now() AT TIME ZONE 'UTC' "
+                               "WHERE user_id=%s", (user_id,))
+                except Exception:
+                    _logger.debug("Failed to update last_login "
```
gurneyalex: can't have this logged at debug level. we want at least warning, IMO, because it is a sign that something is wrong.

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,109 @@
...
+                                  "for db:%s login:%s",
+                                  db, login, exc_info=True)
+        except openerp.exceptions.AccessDenied:
```
gurneyalex: since this can only be thrown by check_credentials, I'd move that try...except block just around that call.

gurneyalex:
for gamification and auth_openid: maybe we can add cron to periodically synchronize the dates on the res_users table ?

base_login_date_improvement/__openerp__.py
legalsylvain: I think we have to let the choice to the admin to install or not this module. I think that 'auto_install' is required in OCA modules only for glue module.
My 2 cents.

mdietrichc2c:
@gurneyalex thanks for the feedback.
Most of the code is from the `login` method in the `base` module, (the autocommit, the debug logging, the exception handling). I'll try to rework it.
As for the cron, given that the issue is caused by updating data in the res_users table, adding it would make the module useless, as a transaction called just before the cron update will still fail with the SQL error described on top.

mdietrichc2c:
@gurneyalex I did the following changes:
- no autocommit,
- better try/catches,
- log is now a warning.

gurneyalex:
@mdietrichc2c sorry for the bad wording of my remarks... I admit I did not check the original login method. Staying as close as possible to that method is a good enough reason for me. 

The point of the cron is that maybe you can synchronize everything once a day, which in most setups will not happen while another cron is running (because crons are generally not run in parallel).

base_login_date_improvement/__openerp__.py
gurneyalex: maybe, yes...

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,122 @@
...
+                                   "SET login_dt = now() AT TIME ZONE 'UTC' "
+                                   "WHERE user_id=%s", (user_id,))
+                        cr.commit()
+                    except psycopg2.OperationalError:
+                        _logger.warning("Failed to update last_login "
+                                        "for db:%s login:%s",
+                                        db, login, exc_info=True)
+                        cr.rollback()
+        except Exception:
+            pass
```
gurneyalex: you still need at least a logging statement here.

mdietrichc2c:
@gurneyalex good point: I added an (inactive) cron with a simple SQL request in order to update the original row, and an error log for instead of `pass`.

gurneyalex:
:+1:

base_login_date_improvement/res_users.py
```diff
@@ -0,0 +1,137 @@
...
+                except openerp.exceptions.AccessDenied:
+                    _logger.info("Login failed for db:%s login:%s", db, login)
+                    user_id = False
+
+                if user_id:
+                    try:
+                        update_clause = ('NO KEY UPDATE'
+                                         if cr._cnx.server_version >= 90300
+                                         else 'UPDATE')
+                        cr.execute("SELECT login_dt "
```
gracinet: maybe SELECT FOR UPDATE would be more robust (can prevent some race conditions).  http://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE. As far as I understand it from various readings, atomically inserting or updating according to what exists is a non trivial issue in PG.

For better concurrency a history table (with several entries per user) could maybe be even better, depending on workload, as long as one does not add an index on login_date to make latest entry retrieval faster.

gracinet:
+1
I'm generally in strong approval on this enhancement. 
I would call the module simply "base-login-concurrency" to leave room for further improvements not related to this latest login date. 
For instance, I noticed an update of ir_config_parameters on admin auth that should be investigated.

Ideally, that should even go straight to the core, (which would solve issues with gaming and openid) but I understand that it can be complicated policy-wise.

It turns out that we are currently strongly bitten by such concurrency issues, although quick experiments I conducted with the Odoo 8 on PG >= 9.3 showed that indeed this NO KEY UPDATE mitigates the breaking of _other_ transactions, but we have recent experience that it's a performance bottleneck anyway (got a prod server practically going down a week ago, crawling under load because of this).
And Mathieu's description actually displays some other cases.
Also note that login is triggered by any browser page reload, which is also what a typical user might do in case of lengthy request, worsening and spreading the performance issue.

On a pure technical side, using a separate table is by far the simplest solution, and that's the way it should have been done from the ground up. As a rule of thumb, one should never update frequently a table that's heavily referenced.

base_login_date_improvement/res_users.py
mdietrichc2c: I'm not sure FOR UPDATE would be more robust, since 2 transactions doing a SELECT FOR NO KEY UPDATE will result in the same behavior (unable to lock for the 2nd one). On the other hand, no table uses res_users_login as a foreign key, so to simplify, I will remove the condition and use FOR UPDATE. 
Also, I do not think a more thorught UPSERT is needed too; the behavior in Odoo is to "fail silently" if the lock was not available, and any concurrency condition would only add one second to the login date.

I'll rework the mecanism to use the SELECT FOR UPDATE first, then UPDATE or INSERT if something was found or not, and use either the lock or the unicity constraint on the user_id to fail and rollback if the row was created/updated concurrently.

mdietrichc2c:
@gracinet thanks for the response. 
For renaming the module, I think I would rather rename it "base_concurrency" to handle all issues with concurrency: I didn't have the ir_config_parameter one, as no long-running transaction that I know of uses it, but I had a few as-of-yet-unreproductible ones on tables like product_product or stock_picking. 

PS: we also noticed here too a drop in performance in PSQL >= 9.3, especially with a CSV import.

mdietrichc2c:
Forgot to comment: the module is now renamed `base_concurrency` , and the UPSERT is now more readable. It also uses `SELECT FOR UPDATE` on the `res_users_login`table.

sebalix:
_For instance, I noticed an update of ir_config_parameters on admin auth that should be investigated._
I think it's when the admin logout, it updates the `web.base.url` parameter with the URL used for the current session, unless you set the `web.base.url.freeze` parameter to avoid this update (which is really annoying when you use SSH tunnels to access customer instances).

mdietrichc2c:
@sebalix yes, that's probably it; however, apart from removing this update altogether (which I am in favor of as well) or just adding `web.base.url.freeze`, I don't think much can be accomplished to avoid concurrency conditions.

jgrandguillaume:
Hi,

Thanks for the contribs ! Would be nice to have the OCA format in the description, otherwise :+1: 

Regards,

Joël

lmignon:
:+1:  Thank you for the contribution

PR #145, opened 2015-03-26
Initial commit of encryption_tools

thomas-willdooit:
This module is useful as a starting point for anything that required some form of public key encryption. As an example, it has been used to securely store CC numbers so that front line worker can enter them, but only the manager can access them when generating reports to be sent off to a processing service.


pedrobaeza:
Is this going to be continued?

gurneyalex:
I'm closing the PR as inactive. Just reopen it if you want to work on the module.

PR #147, opened 2015-03-30
Mail footer notified partners

JonathanNEMRY:
This module adds into the footer the partners' name notified by this email

### DIFF
--- a/mail_footer_notified_partners/README.rst
+++ b/mail_footer_notified_partners/README.rst
@@ -4,7 +4,7 @@
 Mail Footer Notified Partners
 =============================
 
-This module adds into the footer the partner's name notified by this email
+This module adds the list of notified partners in the footer of notification e-mails sent by Odoo.
 
 Usage
 =====
--- a/mail_footer_notified_partners/README.rst
+++ b/mail_footer_notified_partners/README.rst
@@ -1,0 +1,35 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Mail Footer Notified Partners
+=============================
+
+This module adds into the footer the partner's name notified by this email
+
+Usage
+=====
+
+For further information, please visit:
+
+  * https://www.odoo.com/forum/help-1
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jonathan Nemry <jonathan.nemry@acsone.eu>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/mail_footer_notified_partners/__openerp__.py
+++ b/mail_footer_notified_partners/__openerp__.py
@@ -25,8 +25,9 @@
 {
     'name': "Mail Footer Notified Partners",
     'summary': """
-        This module adds into the footer the partner's name notified by this
-        email""",
+        This module adds the list of notified partners in the footer of
+        notification e-mails sent by Odoo.
+        """,
     'author': "ACSONE SA/NV,Odoo Community Association (OCA)",
     'website': "http://acsone.eu",
     'category': 'Mail',
--- a/mail_footer_notified_partners/__openerp__.py
+++ b/mail_footer_notified_partners/__openerp__.py
@@ -31,7 +31,7 @@
     'author': "ACSONE SA/NV,Odoo Community Association (OCA)",
     'website': "http://acsone.eu",
     'category': 'Mail',
-    'version': '0.1',
+    'version': '8.0.1.0.0',
     'license': 'AGPL-3',
     'depends': [
         'base',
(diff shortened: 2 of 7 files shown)


Cedric-Pigeon:
@gurneyalex I confirm that @JonathanNEMRY is my colleague and part of Acsone and therefore have an Enterprise CLA signed.

gurneyalex:
@Cedric-Pigeon OCA database updated. Thanks!

Cedric-Pigeon:
@gurneyalex  Thanks

Cedric-Pigeon:
:+1:  Code and tests

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -0,0 +1,75 @@
...
+#     along with mail_footer_notified_partners.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, api
+from openerp import tools
+from openerp.tools.translate import _
+
+ADDITIONAL_FOOTER = _('This message was also sent to')
```
gurneyalex: Could you move the colon from line 55 inside the translation string? 

this will make things easier e.g. for the french translation where colons are preceeded by a space.

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -0,0 +1,75 @@
...
+        partner_ids = self.env.context.get('partners_to_notify')
+        if footer and partner_ids:
+            partners = self.env['res.partner'].browse(partner_ids)
+            partners_name = [
+                partner.name for partner in partners if
+                partner.notify_email != 'none'
+            ]
+            additional_footer =\
+                '<br /><small>%s: %s</small><br />' %\
+                (ADDITIONAL_FOOTER, ', '.join(partners_name))
```
gurneyalex: please use unicode strings. 

Also should there not be some kind of HTML sanitizing of the partner names / emails?

mail_footer_notified_partners/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#     along with mail_footer_notified_partners.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "Mail Footer Notified Partners",
+    'summary': """
+        This module adds into the footer the partner's name notified by this
+        email""",
+    'author': "ACSONE SA/NV",
```
gurneyalex: could you please add "Odoo Community Association (OCA)" to the author entry (by appending it to the string, separated by a comma)? This allows some visibility of OCA addons on apps.odoo.com.

gurneyalex:
@JonathanNEMRY thanks for this useful addon. I've made a review, and there are a couple of things to update, but nothing serious. :smile_cat:

JonathanNEMRY:
@gurneyalex 
Thanks for the review!! I think it is OK now.
About the HTML sanitizing I'm not sure it is required (I've just made this like Odoo into their footer signature). But feel free to suggest anything you think better :)

JonathanNEMRY:
@gurneyalex: Is the label `needs fixing` still required?

mail_footer_notified_partners/models/mail_followers.py
```diff
@@ -0,0 +1,75 @@
...
+#     along with mail_footer_notified_partners.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, api
+from openerp import tools
+from openerp.tools.translate import _
+
+ADDITIONAL_FOOTER = _('This message was also sent to: ')
```
hbrunn: this will give you the wrong translation because the translation is not executed in the user's context but evaluated on load. Better put this into the code

hbrunn:
:+1:

mail_footer_notified_partners/README.rst
```diff
@@ -0,0 +1,35 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Mail Footer Notified Partners
+=============================
+
+This module adds into the footer the partner's name notified by this email
```
sbidoul: This module adds the list of notified partners in the footer of notification e-mails sent by Odoo.

sbidoul:
@JonathanNEMRY while testing I got this message

```
Sent by ACSONE SA/NV using Odoo 
This message was also sent to: Bidoul, Stéphane
about Expense XYZ
```

while this would have been nicer:

```
Sent by ACSONE SA/NV using Odoo about Expense XYZ
This message was also sent to: Bidoul, Stéphane
```

is it feasible?

JonathanNEMRY:
@sbidoul it is due to the `_get_partner_access_link`  from `Portal`  (https://github.com/odoo/odoo/blob/8.0/addons/portal/mail_mail.py)
To set `about Model Record` before the custom footer it should have a second module that depends on `mail_footer_partner_notifed` and `portal` in order to control the inheritance chain. And the current module should be refactored to use the same `_get_partner_access_link` method (`instead of get_signature_footer`)

jgrandguillaume:
:+1:

JonathanNEMRY:
@sbidoul An idea should be to add the notifed partners at the beginning of the footer.
I think the problem could be easily solve with this workaround. If you agree with this solution I can implement it before the merge of this PR

sbidoul:
@JonathanNEMRY yes I still believe the text as shown in https://github.com/OCA/server-tools/pull/147#issuecomment-95490738 is not very nice to read. So putting the notified partner before "Sent by Odoo" could be a good solution.

sbidoul:
With the last commit 81776f7, the notified partners come before the signature block.

IMHO, it's looking better.

:+1:

sbidoul:
superseded by https://github.com/OCA/social/pull/37

PR #148, opened 2015-04-01
[IMP] Module 'auditlog' - Performing logs on 'read' operations

sebalix:
- Some bugfixes to log inherited fields and dummy fields such as 'in_group_X' in 'res.users' model
-  Unit tests updated

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -219,9 +219,11 @@ class auditlog_rule(models.Model):
         def read(self, *args, **kwargs):
             result = read.origin(self, *args, **kwargs)
             # Sometimes the result is not a list but a dictionary
-            if not isinstance(result, list):
-                result = [result]
-            read_values = dict((d['id'], d) for d in result)
+            # Also, we can not modify the current result as it will break calls
+            result2 = result
+            if not isinstance(result2, list):
+                result2 = [result]
+            read_values = dict((d['id'], d) for d in result2)
             # Old API
             if args and isinstance(args[0], sql_db.Cursor):
                 cr, uid, ids = args[0], args[1], args[2]
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -219,10 +219,9 @@ class auditlog_rule(models.Model):
         def read(self, *args, **kwargs):
             result = read.origin(self, *args, **kwargs)
             # Sometimes the result is not a list but a dictionary
-            try:
-                read_values = dict((d['id'], d) for d in result)
-            except TypeError:
-                read_values = dict((d['id'], d) for d in [result])
+            if not isinstance(result, list):
+                result = [result]
+            read_values = dict((d['id'], d) for d in result)
             # Old API
             if args and isinstance(args[0], sql_db.Cursor):
                 cr, uid, ids = args[0], args[1], args[2]
@@ -318,8 +317,7 @@ class auditlog_rule(models.Model):
             # - search the field in the current model and those it inherits
             field_model = self.env['ir.model.fields']
             all_model_ids = [model.id]
-            all_model_ids.extend(
-                inherited.id for inherited in model.inherited_model_ids)
+            all_model_ids.extend(model.inherited_model_ids.ids)
             field = field_model.search(
                 [('model_id', 'in', all_model_ids), ('name', '=', field_name)])
             # The field can be a dummy one, like 'in_group_X' on 'res.users'
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 
-from openerp import models, fields, api, modules, _, SUPERUSER_ID
+from openerp import models, fields, api, modules, _, SUPERUSER_ID, sql_db
 
 FIELDS_BLACKLIST = [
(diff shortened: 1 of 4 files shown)


auditlog/models/rule.py
```diff
@@ -215,41 +215,30 @@ def create(self, vals, **kwargs):
...
-        #         self.env.context, auditlog_method_intercepted=True)
-        #     result = read.origin(
-        #         self.with_context(context), fields, load, **kwargs)
-        #     print "RESULT", result
-        #     return result
 
         def read(self, *args, **kwargs):
             result = read.origin(self, *args, **kwargs)
+            # Sometimes the result is not a list but a dictionary
+            try:
```
pedrobaeza: Isn't better to test if it's a list, and if not, encapsulate it in one instead of try/except?

sebalix: Is there any drawback to test the type directly? We can assume that there are always int or list types passed in parameter.

As performance are important here, I opt for a `try/except`: lists of IDs are the most common (no check = no performance loss) while an ID alone is the exception, and we fall in the `except` in this case. A test implies performance loss on all `read`, I thought it was not desirable.

I was thinking I could drop some `isinstance()` at some places too.

pedrobaeza: Well, I don't have measures, but a `isinstance` call is so performance penalty?

sebalix: I've done some tests and you are right, `isinstance` seems fast enough. I will change that (and fix the missing required field in the test).

sebalix:
I am not able to find out the cause of this failure. I turn around since yesterday without success. It happens only with a `log_read: True` on `res.users`. No problem with `ir.rule` or any other data models. I keep looking.

EDIT: I think it's now fixed

auditlog/models/rule.py
```diff
@@ -294,9 +284,13 @@ def create_logs(self, uid, res_model, res_ids, method,
         log_model = self.env['auditlog.log']
         for res_id in res_ids:
             model_model = self.env[res_model]
-            res_name = model_model.browse(res_id).name_get()
+            # Avoid recursivity with the 'read' method called by 'name_get()'
+            res_name = "%s,%s" % (res_model, res_id)
+            if method is not 'read':
+                name = model_model.browse(res_id).name_get()
+                res_name = name and name[0] and name[0][1] or res_name
```
sebalix: I am not very comfortable with this piece of code. We need to qualify the log with a good name, but a call to `name_get` + `read` log enabled  = falling in recursivity.

sebalix:
I can not reproduce the issue locally. Commands used to run tests:

``` sh
$ createdb -E unicode auditlog_dev
$ ./odoo/server/odoo.py -c odoo.conf -d auditlog_dev -i auditlog --test-enable
```

hbrunn:
@sebalix this is a weirdness of the travis script. It always installs the mail module, and mail adds a not null constraint in res_partner, that's why your test fails (a user is a partner)

That's why I used res.groups in my test, I propose you do the same (or use whatever other model that is not involved with res.partner)

auditlog/models/rule.py
hbrunn: a possibility would be to put some key into the context that causes read above not to do any logging. And given you use the new api here, you'd only have to check that in the v8 branch

auditlog/models/rule.py
```diff
@@ -307,23 +301,67 @@ def create_logs(self, uid, res_model, res_ids, method,
...
-                [('model_id', '=', model.id), ('name', '=', field_name)])
-            field_data = field.read(load='_classic_write')[0]
-            cache[model.model][field_name] = field_data
+                [('model_id', 'in', all_model_ids), ('name', '=', field_name)])
+            # The field can be a dummy one, like 'in_group_X' on 'res.users'
+            # As such we can't log it (field_id is required to create a log)
+            if not field:
+                cache[model.model][field_name] = False
+            else:
+                field_data = field.read(load='_classic_write')[0]
```
sebalix, line 349: Another problem too about calls to `read` on `ir.model.fields` if audit `read` rules are defined on it (did not tested, but it should fail I think) : check if there is a `read.origin` attribute and call this one to avoid the recursivity.

auditlog/models/rule.py
sebalix: Yes this is an idea!

sebalix:
@hbrunn Thanks for the explanations. In fact I used `res.users` to test logs on a model which has inherited fields (as mentioned in the doctstring of the unit test). Maybe is there another data model available in the `base` module to test that, otherwise we will do without it.

dreispt:
@hbrunn wasn't aware the `mail` [installation](https://github.com/dreispt/maintainer-quality-tools/blob/master/travis/test_server#L167) was causing trouble.
It is just used to setup a basic database template to then use for each module individual tests. 
AFAICT it could perfectly be replaced by a saner choice (`base`?).

hbrunn:
@dreispt oh, I thought there was some well thought of reason for that, so I never dared to propose to change this... In https://github.com/OCA/OCB/blob/8.0/.travis.yml, I also had problems with this and override it with base, that works fine

sebalix:
I'm done with this PR. Anyone to review/merge it?

hbrunn:
:+1:

yvaucher:
:+1:

PR #151, opened 2015-04-16
Disable logging while in shell

### LINKED ISSUE #150: Annoying logs while using the shell
I get interrupted while writing in the shell by the following logs:
openerp.addons.hw_escpos.controllers.main: ESC/POS Device Disconnected: Printer Not Found
openerp.addons.hw_escpos.controllers.main: disconnected : Printer Not Found

dreispt:
Fixes #150

### DIFF
--- a/shell/cli/shell.py
+++ b/shell/cli/shell.py
@@ -18,6 +18,7 @@
 
 from __future__ import print_function
 import code
+import logging
 import os
 import signal
 import sys
@@ -62,6 +63,7 @@ class Shell(Command):
                       ' to get one.')
             for i in sorted(local_vars):
                 print('%s: %s' % (i, local_vars[i]))
+            logging.disable(logging.CRITICAL)
             Console(locals=local_vars).interact()
 
     def shell(self, dbname):


foutoucour:
Hey @dreispt 

Looks great,
Just could you explain why you need to disable the logging during shell? Is it to avoid spam?

Otherwise, :+1:

pedrobaeza:
:+1:

lmignon:
:+1:

dreispt:
@foutoucour It can happen to have processes running in the server writing log messages, and these are annoying for our shell session. I personally didn't find this, but #150 reported `hw_escpos` to do that, and I remember having that issue at some point with long polling messages.

lepistone:
I arrived too late, but I am still curious about this one. Normally for local development I set logging to WARN, and my `pdb` sessions are clean, not polluted by logs. Usually I want to see Warnings, and fix them. Even more, if I type something in the debugger that causes an Error, I'd like to see it.

I don't know about the `hw_escpos` module, but if it generates a lot of spurious Warning messages, couldn't we fix it in the module itself, or downgrade to a lower logging level?

Thanks!

dreispt:
@lepistone It's a core module: you could do a PR for that, but...

lepistone:
Yes, I understand. Still, if the module logs at warning of error level for situations we can't fix, we might do a PR anyway, since it looks like a bug to me. (the long polling messages, on the other hand, are INFO, which is fine by me).

PR #153, opened 2015-04-20
[FIX] Super Calendar, generate calendar records

ismaelcj:
Se estaba creando sólo una entrada en el calendario porque se hacía un return antes de terminar la iteración que las genera.

It was creating only one entry in the calendar because a 'return' is made before the end of the iteration that generates them.

### DIFF
--- a/super_calendar/super_calendar.py
+++ b/super_calendar/super_calendar.py
@@ -67,7 +67,7 @@ class super_calendar_configurator(orm.Model):
                     context=context)
 
                 for current_record_id in current_record_ids:
-                    vaues = self._generate_record_from_line(cr, uid,
+                    values = self._generate_record_from_line(cr, uid,
                                                             configurator,
                                                             line,
                                                             current_record_id,
--- a/super_calendar/super_calendar.py
+++ b/super_calendar/super_calendar.py
@@ -120,7 +120,7 @@ class super_calendar_configurator(orm.Model):
                 parse_dict = {'o': record}
                 mytemplate = Template(line.description_code)
                 name = mytemplate.render(**parse_dict)
-            super_calendar_values = {
+            return {
                 'name': name,
                 'model_description': line.description,
                 'date_start': record[line.date_start_field_id.name],
@@ -135,7 +135,7 @@ class super_calendar_configurator(orm.Model):
                 'res_id': line.name.model + ',' + str(record['id']),
                 'model_id': line.name.id,
             }
-        return super_calendar_values
+        return {}
 
 
 class super_calendar_configurator_line(orm.Model):
--- a/super_calendar/super_calendar.py
+++ b/super_calendar/super_calendar.py
@@ -3,6 +3,8 @@
 #
 #    Copyright (C) 2012 Agile Business Group sagl (<http://www.agilebg.com>)
 #    Copyright (C) 2012 Domsense srl (<http://www.domsense.com>)
+#    Copyright (C) 2015 FactorLibre (<http://www.factorlibre.com>)
+#                       Ismael Calvo <ismael.calvo@factorlibre.com>
 #
 #    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


super_calendar/super_calendar.py
```diff
@@ -128,7 +128,9 @@ def _generate_record_from_line(self, cr, uid, configurator, line, context):
                     'res_id': line.name.model + ',' + str(record['id']),
                     'model_id': line.name.id,
                 }
-                return super_calendar_values
+                super_calendar_pool.create(cr, uid, super_calendar_values,
```
pedrobaeza: Don't make the create call in the same method. Return the dictionary and make the create where it is (although you remove the too early return. This is to easy inheritance if someone wants to extend the method adding values to the dictionary. It is even mode desirable to make a method that works for only one line and returns a dictionary.

ismaelcj:
The problem is the _generate_record_from_line function should generate a dictionary for each line we want to create. So I think it has to iterate on generate_calendar_records function. I changed the paremetres in _generate_record_from_line function, I know it can be a problem for those who had inherited it.

super_calendar/super_calendar.py
```diff
@@ -57,78 +59,83 @@ def generate_calendar_records(self, cr, uid, ids, context=None):
...
+                duration = (date_stop - date_start).total_seconds() / 3600
+            elif line.duration_field_id:
+                duration = record[line.duration_field_id.name]
+            if line.description_type != 'code':
+                name = record[line.description_field_id.name]
+            else:
+                parse_dict = {'o': record}
+                mytemplate = Template(line.description_code)
+                name = mytemplate.render(**parse_dict)
+            super_calendar_values = {
```
pedrobaeza: Put return here directly

super_calendar/super_calendar.py
```diff
@@ -57,78 +59,83 @@ def generate_calendar_records(self, cr, uid, ids, context=None):
...
+                    line.user_field_id and
+                    record[line.user_field_id.name] and
+                    record[line.user_field_id.name].id or
+                    False
+                ),
+                'configurator_id': configurator.id,
+                'res_id': line.name.model + ',' + str(record['id']),
+                'model_id': line.name.id,
+            }
+        return super_calendar_values
```
pedrobaeza: Put `return {]` to avoid undefined variable in some cases.

gurneyalex:
@ismaelcj could you add a test, or describe a scenario where the bug you are fixing happens?

ismaelcj:
Hi @gurneyalex ,

I do this:
- Create a Super Calendar configurator with Phonecall as model (needs crm module installed). Going to _Super Calendar/Configuration/Configurators_
- Create some phone calls. Going to _Sales/Phone Calls/Logged Calls_
- Push the 'Generate Calendar' button. Going to _Super Calendar/Configuration/Configurators_
- Go to calendar view. It was created just one event instead of one for each phonecall record. Going to _Super Calendar/Calendar/Calendar_

super_calendar/super_calendar.py
```diff
@@ -57,78 +59,83 @@ def generate_calendar_records(self, cr, uid, ids, context=None):
...
+                    uid,
+                    line.domain and safe_eval(line.domain) or [],
+                    context=context)
+
+                for current_record_id in current_record_ids:
+                    values = self._generate_record_from_line(cr, uid,
+                                                            configurator,
+                                                            line,
+                                                            current_record_id,
+                                                            context)
```
gurneyalex: can you fix the indentation on the 4 lines above?

gurneyalex:
I'm :+1: once the indentation issue above is fixed (provided this produces a green build)

eLBati:
I would not change `_generate_record_from_line` signature: if other modules are using/overriding it, they will stop working.
Is it possible to reach what you want without changing method signature?

gurneyalex:
no action taken on the PR since 18 month, I'm closing as inactive. Please reopen if needed.

PR #154, opened 2015-04-22
[ADD] base_import_restrict

alejandrosantana:
Add a module called "import_csv_restrict" that allows to set group-based permissions to each user for allowing/restricting to import data from CSV files.


pedrobaeza:
Call it _base_import_restrict_ to follow naming conventions.

alejandrosantana:
@pedrobaeza Done with the rename.
I also modified the code and now there is no flickering problem I had before.

base_import_restrict/static/src/js/import.js
```diff
@@ -0,0 +1,57 @@
+openerp.base_import_restrict = function(instance) {
```
alejandrosantana: Should I add at the top some licensing summary text like the following?:

``` javascript
// License, author and contributors information in:
// __openerp__.py file at the root folder of this module.
```

pedrobaeza: As you prefer...

eLBati:
I would keep the reference to 'csv' in module name, so that the context is immediately clear .
What do you think?

gurneyalex:
:+1:

alejandrosantana:
@gurneyalex @eLBati @pedrobaeza Is there any naming convention to check, besides the 'singular' policy in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.rst ?
I guess two usual options would be:
- autodescriptive name (in the fashion @eLBati suggests with reference to CSV.
- main base module + some hint of the extended functionality, as I understood @pedrobaeza suggested: base_import + _restrict = base_import_restrict

I am not sure about it, as both have pros and cons, but I would go for the second form, as seems easy for developers to set our minds in a specific module environment. For the rest there is the module _full name_, _summary_ or _full description_.

What do you say?

pedrobaeza:
@eLBati mentions about csv to not confuse with other import features that have been develop under OCA's umbrella.

alejandrosantana:
I see. 
In that case, _base_import_csv_restrict_ or _base_import_csv_optional_ would fit?

pedrobaeza:
First one seems good

alejandrosantana:
OK, if no one disagrees, I change it to _base_import_csv_restrict_ then.

alejandrosantana:
I have cleaned the commits and changed module name to "base_import_csv_restrict".

guewen:
LGTM
Suggestion to improve: add the check on the group in the `load` method so the import is restricted server-side.

alejandrosantana:
@guewen Do you refer to the `do` method in base_import/models.py? I see no `load` method there.

guewen:
> @guewen Do you refer to the do method in base_import/models.py? I see no load method there.

I speak about: https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L1023

alejandrosantana:
@guewen I am trying to override the method you say, but I have a bit of a problem:
`class BaseModel(object):` has `_name = None` so... how do I inherit it in my module?
I am not getting it to work, so I cannot override the method to also check the group there.
Any hint?

pedrobaeza:
You have to monkey-patching the class. You can see something similar here: https://github.com/OCA/crm/pull/38/files#diff-df883f43bfa35a2a786daa20cc741e74R100

gurneyalex:
@alejandrosantana can you rebase your PR: this should fix the runbot status.

alejandrosantana:
@pedrobaeza Thanks a lot, I got it working thanks to the given examples.

@gurneyalex I just reseted locally to match OCA repo, added my module and did a forced push, so it should be OK.

alejandrosantana:
@guewen Already modified the load() method to block import there.

alejandrosantana:
@gurneyalex Fixed, rebased and green (as requested in general in Twitter).

moylop260:
Can you add a unitests?

base_import_csv_optional/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+    '''
+    res = {}
+    current_user = self.env['res.users'].browse(self.env.uid)
+    allowed_group = 'base_import_csv_optional.group_import_csv'
+    if current_user and current_user.has_group(allowed_group):
+        res = base_load(self, fields=fields, data=data)
+    else:
+        msg = _('User (ID: %s) is not allowed to import data '
+                'in model %s.') % (self.env.uid, self._name)
+        _logger.error(msg)
```
guewen, line 27: Log messages shouldn't be translated

alejandrosantana, line 27: Oh, I see it. People reading logs can be from anywhere. I will fix it.

alejandrosantana:
@moylop260 I have not write unittests before but I will take a look. It is about time!

gurneyalex:
:+1: (again)

pedrobaeza:
The README should be improved according latest template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst, but I'm not sure if it's too late to ask for that (this PR is too old).

espo-tony:
I started a PR for migrating this functionality to v10, you can watch it here: #694 
In the discussion it came up that maybe could be better to change the name to _base_import_security_group_ in order to follow the OCA name conventions. Can I suggest to reflect this change on this PR too?

base_import_csv_optional/models/models.py
```diff
@@ -0,0 +1,36 @@
...
+@api.model
+def load_import_optional(self, fields=None, data=None):
+    '''Overriding this method we only allow its execution
+    if current user belongs to the group allowed for CSV data import.
+    An exception is raised otherwise, and also log the import attempt.
+    '''
+    res = {}
+    current_user = self.env['res.users'].browse(self.env.uid)
+    allowed_group = 'base_import_csv_optional.group_import_csv'
+    if current_user and current_user.has_group(allowed_group):
```
espo-tony, line 22: It would be safer to check for the existence of the group referenced with the external ID _base_import_csv_optional.group_import_csv_ before to use it. I suggest that when this group is not found (and therefore no user can reasonably belong to it) this method should just return the parent logic, which means to allow any user to run the import.

espo-tony (changes requested):

StefanRijnhart:
I'm continuing this PR here: https://github.com/OCA/server-tools/pull/743. @espo-tony I took your suggestion.

PR #155, opened 2015-04-23
[ADD] qweb_usertime

veloutin:
# QWeb Usertime Tag

This module adds a new tag renderer to QWeb, "usertime", which allows adding
the current time in the timezone of the user. It can be used as::

```
<t t-usertime="%Y-%m-%d %H:%M:%S" />
```


qweb_usertime/static/description/icon.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
yvaucher: I don't think we want the generic OCA icon svg in all modules.

yvaucher:
Can you create a README.rst ?

guewen:
Any way to make it respect the language's default format instead of hardcoding it?

veloutin:
@guewen currently there is no default format hard coded, the user simply need to provide the format they want. This is similar behaviour to using time.strftime, which is the current alternative for getting the current time.

I see res.lang provides date_format and time_format, I suppose I could use those to make up a default format, based on the context for language.

guewen:
> currently there is no default format hard coded, the user simply need to provide the format they want. This is similar behaviour to using time.strftime, which is the current alternative for getting the current time.

So the user has to hardcode it in its view, it is what I mean.

> I see res.lang provides date_format and time_format, I suppose I could use those to make up a default format, based on the context for language.

That's the point, it's what is used in the usual date and datetime widgets.

yvaucher:
Would be perfect to have format from context lang

anthony-muschang:
:+1: Work well, tested

yvaucher:
:+1: Works fine thanks

PR #157, opened 2015-04-27
[FIX][cron_run_manually] active_test cannot be propagated to the method to execute

Olivier-LAURENT:
`ir.cron` model uses a special way to handle the `active` flag allowing to `search()` inactive documents whereas it is usually not the case. This feature is assumed by setting the `active_test` key to `False` in the context of the `ir.cron` window action.

This mechanism should only concern the `ir.cron` model itself, not models on which jobs will be executed.

The PR simply removes the `active_test` key from the context before launching the job.

### DIFF
--- a/cron_run_manually/ir_cron.py
+++ b/cron_run_manually/ir_cron.py
@@ -60,9 +60,14 @@ class Cron(models.Model):
 
         _logger.info('Job `%s` triggered from form', self.name)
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

guewen:
:+1:

lmignon:
:+1:

PR #158, opened 2015-04-28
[IMP] hide unnecessary buttons in wizard

hbrunn:
I propose to rewrite all wizard actions to server actions, because this way we can present the user a saved record. The way it is currently, purging one model is only possible after saving, which is confusing. Also, we have the possibility to hide all unnecessary buttons (save/delete/attachments).


pedrobaeza:
But pressing a button or an action doesn't save it automatically?

hbrunn:
this is true for buttons on the wizard form, but this wizard generates a list of models that can be purged separately. And the buttons on the x2many field are only usable after saving

pedrobaeza:
OK for me then :+1:

anthony-muschang:
Tested :+1:

PR #159, opened 2015-04-29
[IMP] deal with modules whose models can't be loaded

hbrunn:


database_cleanup/model/purge_modules.py
```diff
@@ -23,6 +23,42 @@
...
 from openerp.modules.module import get_module_path
 from openerp.tools.translate import _
+from openerp.addons.base.ir.ir_model import MODULE_UNINSTALL_FLAG
+
+
+class IrModelConstraint(orm.Model):
+    _inherit = 'ir.model.constraint'
+
+    def _module_data_uninstall(self, cr, uid, ids, context=None):
+        '''this function crashes for constraints on undefined models'''
```
pedrobaeza: Put `"""` for doc.

database_cleanup/model/purge_modules.py
```diff
@@ -23,6 +23,42 @@
...
+                this.unlink()
+        return super(IrModelConstraint, self)._module_data_uninstall(
+            cr, uid, ids, context=context)
+
+
+class IrModelData(orm.Model):
+    _inherit = 'ir.model.data'
+
+    def _module_data_uninstall(self, cr, uid, modules_to_remove, context=None):
+        '''this function crashes for xmlids on undefined models or fields
```
pedrobaeza: Put `"""` for doc.

pedrobaeza:
I understand that this functionality is apart from the module itself, and it allow to uninstall a module. Isn't it?

hbrunn:
@pedrobaeza yes, the module is meant for the leftovers of a homebrew migration (or any other case that puts you database in a weird state concerning installed modules)

pedrobaeza:
Yeah, but I mean if this is interesting to go on OCB.

hbrunn:
I wouldn't want to propose this to the standard code. This module can seriously mess up your database, so better keep the radical removal code there

pedrobaeza:
Yeah, I know that the module is an advanced tool, but I mean the content of this PR, that only reacts in a module uninstall, and avoid the crash when there are missing things.

If you don't think it's safe, I'm not going to insist more, so :+1:

database_cleanup/model/purge_modules.py
```diff
@@ -23,6 +23,42 @@
...
+    def _module_data_uninstall(self, cr, uid, modules_to_remove, context=None):
+        """this function crashes for xmlids on undefined models or fields
+        referring to undefined models"""
+        ids = self.search(cr, uid, [('module', 'in', modules_to_remove)])
+        for this in self.browse(cr, uid, ids, context=context):
+            if this.model == 'ir.model.fields':
+                field = self.pool[this.model].browse(
+                    cr, uid, [this.res_id],
+                    context=dict(
+                        context or {}, **{MODULE_UNINSTALL_FLAG: True}))[0]
```
gurneyalex: seriously?
I'd rather have something like:

```
if context is None:
    context = {}
context_uninstall = context.copy()
context_uninstall[MODULE_UNINSTALL_FLAG] = True
```

at the beginning of the method. 

Side note: if you browse with `this.res_id` rather than `[this.res_id]` you can get rid of `[0]`

gurneyalex:
globally :+1: 

I feel the suggested change would improve readability.

hbrunn:
@gurneyalex thanks, done (even though i like `dict(context, my_extra_key=value)` as a pattern but in this case it indeed becomes a monstrosity)

StefanRijnhart:
:+1:

PR #160, opened 2015-04-30
[IMP] Add 'save_translation_file' module

sylvainc:
This module allows to save translation files automatically from a button in modules list/form, instead of launching manually the configuration "Export Translation" wizard.


save_translation_file/save_translation_file.py
```diff
@@ -0,0 +1,56 @@
...
+            iso_code = get_iso_codes(lang.code)
+            filename = '%s.%s' % (iso_code, format_)
+            files.append((filename, lang.code))
+
+        for filename, lang in files:
+            path = os.path.join(i18n_path, filename)
+            with open(path, 'w') as buf:
+                tools.trans_export(lang, self.name, buf, format_, self.env.cr)
+
+        self.ensure_one()
```
pedrobaeza: Put it at the beginning of the method to avoid unnecessary computation.

save_translation_file/save_translation_file.py
```diff
@@ -0,0 +1,56 @@
...
+
+        files = [('%s.pot' % self.name, False)]
+        for lang in langs:
+            iso_code = get_iso_codes(lang.code)
+            filename = '%s.%s' % (iso_code, format_)
+            files.append((filename, lang.code))
+
+        for filename, lang in files:
+            path = os.path.join(i18n_path, filename)
+            with open(path, 'w') as buf:
```
pedrobaeza, line 36: You should present the dialog to save the file, because not everybody has access to the server (or prefer to make it server-side)

sebastienbeau, line 36: In fact, it's exactly what we want. This module is only usefull for developper. You can translate from  each screen using the anybox module, and then when you're module is translated you save it on the file store, launch a git diff and commit it ;)
@sylvainc Can you use the OCA template and add in the description what is the aim of the module? And specify explicitly that it's only for developper. Thanks

bealdav:
really useful module in combination with https://github.com/anybox/odoo-translation/tree/master/ui_translation
:+1:

guewen:
Is it work in progress or not? If not, please correct the PR's description.

bealdav:
I've just tested the module. The only bug was fixed by 'trans_export: pass module name as list' commit.
Thanks for this nice job which bring a better productivity for all of us.

chafique-delli:
@sylvainc, Very convenient for developers, good job. :+1:

hparfr:
:+1:

bealdav:
And now a test, nice :+1: please merge it @sebastienbeau

sylvainc:
Hello @pedrobaeza, @guewen, @sebastienbeau 

I have "rebased" the code of the module, it isn't a WIP anymore.
Thanks for the review.

sebastienbeau:
Everything is green so it's ok

PR #167, opened 2015-05-17
[ADD] Added auditlog_easy_config module.

alejandrosantana:
Module _auditlog_easy_config_ aims to make easier and faster the management od _auditlog_ module.
It enables to create and subscribe at once a rule for each model in the database, at any time.
Also allows to select some rules and subscribe, unsubscribe and change tracked actions.


alejandrosantana:
We will be dealing with @forvas CLA during this week.

pedrobaeza:
Why not integrating this options in the same module?

alejandrosantana:
Did not think on it! We created it as an extension for a customer and then we thought it could be useful. But yes, it can be integrated into original _auditlog_. We will do it this week, then.

alejandrosantana:
Close this PR. We will integrate new functionality into _auditlog_ module, as suggested.

PR #171, opened 2015-05-25
[ADD] secure_uninstall module

bealdav:
Proposition for https://github.com/odoo/odoo/issues/6835

Reviews appreciate


secure_uninstall/README.md
```diff
@@ -0,0 +1,5 @@
+Secure Uninstall
+================
+
+Ask admin password ('admin_passwd' key from config file)
+before to proceed to module uninstallation
```
gurneyalex, line 5: why not validate the supplied password using the admin password from the database?

bealdav, line 5: Admin db password might be known of the customer at a time or another of the project, perhaps for a short time for a ponctual action
Config file password will stay secret of all the users.
Could you have a look to https://github.com/OCA/server-tools/pull/171/files#diff-8600e5eba033357ad3d6fdc709908ccdR29, please. It works but i would be sure there is no bug

jholze:
:+1:  should be implemented in odoo by default :(

secure_uninstall/module.py
```diff
@@ -0,0 +1,81 @@
...
+from openerp.osv import orm, fields
+from openerp.tools.config import config
+
+
+class Module(orm.Model):
+    _inherit = 'ir.module.module'
+
+    def button_uninstall(self, cr, uid, ids, context=None):
+        if 'uninstall_authorized' in context:
+            del context['uninstall_authorized']
```
gurneyalex: I'm not sure this will work if the new API comes into play, with context being a frozen dict. You should make a copy of context before modifying it. 

Also, check for context being None, and set it to `{}`  before your test for the key

secure_uninstall/module.py
```diff
@@ -0,0 +1,81 @@
...
+                raise orm.except_orm(
+                    'Missing configuration key',
+                    "'admin_passwd' configuration key is not set in "
+                    "your Odoo server configuration file: "
+                    "please set it a value")
+            if elm.password != config_passwd:
+                raise orm.except_orm(
+                    "Password Error",
+                    "Issue\n_____\nProvided password '%s' doesn't match with "
+                    "'admin_passwd' comes from your "
```
gurneyalex: s/comes from your/found in the/

secure_uninstall/wizard_view.xml
```diff
@@ -0,0 +1,28 @@
...
+    <data>
+
+
+<record id="view_uninstall_wizard_form" model="ir.ui.view">
+    <field name="model">uninstall.check.wizard</field>
+    <field name="arch" type="xml">
+        <form string="Check" version="7.0">
+            <p class="oe_inline oe_grey">
+Uninstall module process remove all data managed by the module.
+            </p>
```
gurneyalex, line 14: maybe we should explain that "some columns can be dropped and there is no way to recover that except restoring a database backup". 

I know users don't read messages, but still... I do read them :smiley_cat:

gurneyalex:
a few changes needed in the code, some improvement suggested. 

Will you be porting this to 8.0?

secure_uninstall/module.py
gurneyalex: ok I had not checked the target branch. But sill the remarks apply: don't modify the context as it impacts the caller. Pass an altered copy of the context instead.

gurneyalex:
Runbot is red because of server_env_files. known issue, we will try try and fix this during the sprint.

bealdav:
Thanks for your review, I want to use it in v7 for now. Help welcome for porting v8.

secure_uninstall/README.md
legalsylvain, line 5: Agree with @bealdav. In many situations, admin password is known by the customer. (maybe not a good practice of course), but it's the reality. So, OK for me to use admin_passwd, as removing a module is a very rare and important action.

legalsylvain:
:+1:

LGTM, thanks. Nice feature. (Code review / No test)

rvalyi:
:+1:

Hello David, you convinced me that checking for the admin password was good enough to avoid having to hack into the ORM for preventing removing data on module uninstall. Code looks all right to me. Thanks for this useful module.

chafique-delli:
Thanks, David for this very useful module. :+1:

bealdav:
Thanks a lot

lmignon:
:+1:  (Code review / No test) As others say,a useful module!

ahilali:
good job  @bealdav  ,Thanks for this useful module

gurneyalex:
@bealdav rebasing your module will get the recent fixes on runbot (runbot does not merge so rebase is required).

secure_uninstall/README.md
gurneyalex, line 5: Elsewhere in the interface (database management screens) this is called 'Master Password'. I think this should be used here too to reduce confusion.

secure_uninstall/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+    'category': 'Base',
+    'summary': "Ask password to authorize uninstall",
+    'depends': [
+        'base',
+    ],
+    'description': """
+Secure Uninstall
+================
+
+Ask admin password ('admin_passwd' key from config file)
```
gurneyalex: same remark as above: "Master password"

secure_uninstall/module.py
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+
+class UninstallCheckWizard(orm.TransientModel):
+    _name = 'uninstall.check.wizard'
+
+    _columns = {
+        'password': fields.char(
+            string='Password', required=True,
+            help="'admin_passwd' value from Odoo configuration file")
```
gurneyalex: maybe mention "Master password" here?

gurneyalex:
:+1: (but I'd really like the "Master password" term to be used for consistency)

bealdav:
Thnaks for remarks, fix in 2 weeks

secure_uninstall/module.py
```diff
@@ -0,0 +1,84 @@
...
+from openerp.tools.config import config
+
+
+class Module(orm.Model):
+    _inherit = 'ir.module.module'
+
+    def button_uninstall(self, cr, uid, ids, context=None):
+        if 'uninstall_authorized' in context:
+            if not context:
+                context = {}
```
moylop260: Change to

``` python
    def button_uninstall(self, cr, uid, ids, context=None):
        if not context:
            context = {}
        if 'uninstall_authorized' in context:
```

secure_uninstall/__openerp__.py
```diff
@@ -0,0 +1,50 @@
+# coding: utf-8
```
moylop260, line 1: Use `# -*- coding: utf-8 -*-`

moylop260:
:+1: (with small comments)

secure_uninstall/__openerp__.py
bealdav, line 1: @moylop260 
in french
http://sametmax.com/un-header-dencoding-plus-simple-pour-python
in english

http://translate.google.com/translate?hl=fr&sl=auto&tl=en&u=http%3A%2F%2Fsametmax.com%2Fun-header-dencoding-plus-simple-pour-python%2F&sandbox=1

moylop260, line 1: You can generate a new pr to MT to change our current rule:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms

bealdav, line 1: ok thanks

jholze:
@moylop260 good job, very useful. Do you have the migration to v8 on your list ?

bealdav:
@jholze  no migration on my side, but you can make a PR (appreciate by a lot people I think).

moylop260:
Discuss in progress of my comment:
https://github.com/OCA/maintainer-tools/pull/124

This pr has 7 approvals then merge...

PR #173, opened 2015-06-05
[8.0] New module 'auditlog_async', asynchronous mode for the 'auditlog' module

sebalix:
This module overrides the `auditlog` module to make asynchronous the log creation, resulting in greater performance and a better user experience.
It uses the `connector` framework to delay the log creation in a job queue.

### DIFF
--- a/auditlog_async/README.rst
+++ b/auditlog_async/README.rst
@@ -1,0 +1,43 @@
+Track user operation on data models (asynchronous mode)
+=======================================================
+
+This module overrides the ``auditlog`` module to make asynchronous the
+log creation, resulting in greater performance and a better user experience.
+
+Usage
+=====
+
+This module uses the ``connector`` framework to delay the log creation in a
+job queue.
+
+To take advantage of this module, you have to:
+
+ * run `Odoo` in multiprocess mode (``workers`` > 0)
+ * start `Odoo` with the ``ODOO_CONNECTOR_CHANNELS`` environment variable
+   defined and ``--load=web,connector``
+
+For further information, please visit:
+
+ * http://odoo-connector.com/guides/jobrunner.html#how-to-use-it
+ * https://www.odoo.com/forum/help-1
+
+Credits
+=======
+
+Contributors
+------------
+
+* Sebastien Alix <sebastien.alix@osiell.com>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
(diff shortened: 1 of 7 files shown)


sebalix:
The `connector` can not be installed on the runbot. Do I forget something?

hbrunn:
I think @gurneyalex is the good person to ask

auditlog_async/README.rst
```diff
@@ -0,0 +1,43 @@
...
+=====
+
+This module uses the ``connector`` framework to delay the log creation in a
+job queue.
+
+To take advantage of this module, you have to:
+
+ * run `Odoo` in multiprocess mode (``workers`` > 0)
+ * start `Odoo` with the ``ODOO_CONNECTOR_CHANNELS`` environment variable
+   defined and ``--load=web,connector``
```
sbidoul, line 17: Perhaps add a link the connector documentation: http://odoo-connector.com/guides/jobrunner.html

sebalix, line 17: Yes, I put the link just below, but it's not obvious yes.

sbidoul:
Interesting idea. Have you done some timing comparisons? I'm curious to see how much the job creation is faster that a direct auditlog writing.

auditlog_async/README.rst
sbidoul, line 17: Right, I see it now!

sebalix:
It is not as fast as expected, but here some numbers about creating 100 partners on the server-side (average time from 4 attemps for each test), with one rule logging the `create` operation:

w/o Audit Log = ~5.5s
Audit Log (normal) = ~18.6s
Audit Log (async) = ~13.5s

This is about 28% faster than the normal behaviour of original module, but this is only about the `create` operation.

Now if we create 100 partners (always on the server-side) and update them right after with a rule set on `create` and `write` operations, there is almost no difference in normal or asynchronous mode (24-25s on my machine). The `read` call to fetch data before performing the `write` operation is what takes time, and it is outside the `create_logs()` method.

Right now, the `auditlog` module makes a snapshot of the data before the `write` operation, and another right after, then logs the difference between them. This allows to display the value of computed fields in the log, but it costs a huge time. Another approach would be to just log the `vals` parameter, we lost some information if computed fields are based on the updated fields, but this should greatly improves the performance as we could entirely delay the log creation in the job queue. Don't know which is better... ? Maybe an option, if it is relevant?

gurneyalex:
I'm a bit reluctant to add a dependency from server-tools to connector, and I'd rather have it the other way round... But I could be convinced otherwise.

@sebalix side remark, did you try base_concurrency to check if there is any speedup (both with and without your addon)?

sebalix:
@gurneyalex no I didn't.

I think this module is not stable enough (there is some drawbacks as written above), and I wonder if it worth it. In fact I was about to close this PR as we don't use this module in production, but maybe someone will be interested by the concept and make some benchmarks?
Still, auditlog is very nice and I would like to improve the performance with one way or another.

gurneyalex:
OK then. I've changed the labels to WIP. Feel free to close if you think it's better (or write a blog post about your attempt :smile_cat: )

gurneyalex:
I'm closing this PR because it's been inactive for a long time. Please reopen if work resumes.

sebalix:
@gurneyalex no problem, I will no longer work on this I think.

PR #175, opened 2015-06-08
ADD sql_export module

florian-dacosta:
This module is usefull when a user knows how to make sql request. He can extract data from the database easily.
He also can choose wich user is able to use the sql resquest he creates.


sql_export/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+##############################################################################
+
+{'name': 'SQL Export',
+ 'version': '0.1',
+ 'author': 'Akretion,Odoo Community Association (OCA)',
+ 'website': 'http://www.akretion.com',
+ 'license': 'AGPL-3',
+ 'category': 'Generic Modules/Others',
+ 'summary': 'Export data in csv file with SQL requests',
+ 'depends': ['mrp',
```
lmignon: Do you  depend on 'mrp'?  IMO this should be 'base'.

florian-dacosta: Of course! I'll change that now, it is a mistake.

sql_export/sql.py
```diff
@@ -0,0 +1,135 @@
...
+
+class SqlExport(Model):
+    _name = "sql.export"
+    _description = "SQL export"
+
+    def _check_query_allowed(self, cr, uid, ids, context=None):
+        for obj in self.browse(cr, uid, ids, context=context):
+            if any(word in obj.query.lower() for word in PROHIBITED_WORDS):
+                return False
+            if obj.query.strip().split(' ')[0].lower() != 'select':
```
lmignon: IMO the second `if` is too restrictive.  It's sometimes useful to build query using the `with` clause (http://www.postgresql.org/docs/9.1/static/queries-with.html). Moreover, if you want to test that the query starts with `select` you should write

``` python
            if not obj.query.lower().startswith('select'):
                return False
```

IMO, checking for prohibited works is enough.

sql_export/sql.py
```diff
@@ -0,0 +1,135 @@
...
+                     'prohibited actions (Delete, drop...)', ['query'])]
+
+    def export_sql_query(self, cr, uid, ids, context=None):
+        if not context:
+            context = {}
+        for obj in self.browse(cr, uid, ids, context=context):
+            now = time.strftime('%Y-%m-%d', time.localtime())
+            output = StringIO.StringIO()
+            query = "COPY (" + obj.query.strip() + \
+                    ") TO STDOUT WITH CSV HEADER DELIMITER ';'"
```
lmignon: It should be nice to be able to specify at least the delimiter and the encoding into the wizard.

lmignon: Or maybe, you should add a new field 'copy_options'  with `CSV HEADER DELIMITER ';'`as default value and use this new field in your query. In this way, the options can be adapted to the user needs.

``` python
query = "COPY (" + obj.query.strip() + ")  TO STDOUT WITH " + obj.copy_options
```

sql_export/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+
+{'name': 'SQL Export',
+ 'version': '0.1',
+ 'author': 'Akretion,Odoo Community Association (OCA)',
+ 'website': 'http://www.akretion.com',
+ 'license': 'AGPL-3',
+ 'category': 'Generic Modules/Others',
+ 'summary': 'Export data in csv file with SQL requests',
+ 'depends': ['base',
+             ],
```
lmignon, line 35: As your PR is for 7.0 you must provide a description.

sql_export/README.rst
```diff
@@ -0,0 +1,33 @@
...
+    :alt: License: AGPL-3
+
+SQL Export
+==========
+
+Allow to export data in csv files FROM sql requests.
+There are some restrictions in the sql sql request, for example, it has to
+begin by "SELECT"
+A new menu named Export is created.
+
```
lmignon, line 11: The 'bug tracker' section is missing. (see https://github.com/OCA/maintainer-tools/tree/master/template/module#bug-tracker)

sql_export/sql.py
florian-dacosta: Thanks for your comment!
You are right, checking first word is a bit restrictive.
But I am not sure checking for prohibited words is enought.
I could forget one or two important since I am not a huge expert in sql
Also, I did not include some word like update or create in this list.
So someone could update any tables, which we do not want!

I chose not to put create because of the field create_date that we may want to extract.
Also update, is easily used in fields name. (like in prestashop or magento connector)
That is why I restrict on the word SELECT.

Of course, I am open to idea to make it better.

lmignon: You should use a regular expression to search for the whole word and include `update` and `create` in your list of prohibited words.

lmignon:
@florian-dacosta Thank you for this PR: nice functionality for advanced user only. Your module works fine (tested in my env). Some additional remarks.
- If you don't provides an icon, use the one provided by OCA (https://github.com/OCA/maintainer-tools/tree/master/template/module/static/description)
- As it's a new functionality, you should also provide a unittest for your module.
- You should rename your file `sql.py` to `sql_export.py`  since it only provides 1 model (and `sql_view.xml` -> `sql_export_view.xml`).

lmi

sql_export/sql.py
```diff
@@ -0,0 +1,135 @@
...
+    def _check_query_allowed(self, cr, uid, ids, context=None):
+        for obj in self.browse(cr, uid, ids, context=context):
+            if any(word in obj.query.lower() for word in PROHIBITED_WORDS):
+                return False
+            if obj.query.strip().split(' ')[0].lower() != 'select':
+                return False
+        return True
+
+    def _get_editor_group(self, cr, uid, *args):
+        gr_obj = self.pool.get('res.groups')
```
lmignon: you should query `ir.model.data` with the xmlid

``` python
def _get_editor_group(self, cr, uid, *args):
    model_obj = self.pool['ir.model.data']
    return model_obj.get_object_reference(
        cr, uid, 'sql_export', 'group_sql_request_editor')[1]
```

sql_export/sql.py
```diff
@@ -0,0 +1,135 @@
...
+
+    _constraints = [(_check_query_allowed,
+                    'The query you want make is not allowed : '
+                     'prohibited actions (Delete, drop...)', ['query'])]
+
+    def export_sql_query(self, cr, uid, ids, context=None):
+        if not context:
+            context = {}
+        for obj in self.browse(cr, uid, ids, context=context):
+            now = time.strftime('%Y-%m-%d', time.localtime())
```
lmignon: You should include the time into the file name and use the datetime returned by `fields.datetime.context_timestamp` (https://github.com/odoo/odoo/blob/7.0/openerp/osv/fields.py#L337)  to get the dt converted to the client's timezone.

sql_export/security/sql_export_security.xml
```diff
@@ -0,0 +1,21 @@
...
+    </record>
+
+    <record model="ir.rule" id="sql_export_restric_access_user_or_group">
+        <field name="name" >SQL Export users and groups rules</field>
+        <field name="model_id" ref="model_sql_export"/>
+        <field eval="1" name="perm_read"/>
+        <field eval="0" name="perm_create"/>
+        <field eval="0" name="perm_write"/>
+        <field eval="0" name="perm_unlink"/>
+        <field name="domain_force">['|', ('user_ids','=',user.id), ('group_ids','=', [x.id for x in user.groups_id])]</field>
```
lmignon:
``` python
('group_ids','in', [x.id for x in user.groups_id])]
```

florian-dacosta:
@lmignon Thanks for your feedback.
It should be ok now!

sql_export/README.rst
```diff
@@ -0,0 +1,41 @@
...
+No update, deletion or creation are possible.
+A new menu named Export is created.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/maintainer-tools/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/maintainer-tools/issues/new?body=module:%20sql_export%0Aversion:%200.1%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
lmignon: the url is https://github.com/OCA/server-tools/issues :smirk: 
and version is the one of the branch 8.0

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_sql_query(self):
+        test = self.sql_model.export_sql_query(
+            self.cr, self.uid, [self.query_id])
+        wizard = self.registry('sql.file.wizard').browse(
+            self.cr, self.uid, test['res_id'])
+        export = base64.b64decode(wizard.file)
+        self.assertEqual(export.split(';')[0], 'name')
+        self.assertTrue(len(export.split(';')) > 6)
+
+    def setUp(self):
```
lmignon, line 23: Common usage is to put the setUp method on top of the test class

sql_export/sql_export.py
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    _defaults = {
+        'group_ids': _get_editor_group,
+        'copy_options': "CSV HEADER DELIMITER ';'",
+    }
+
+    _constraints = [(_check_query_allowed,
+                    'The query you want make is not allowed : '
+                     'prohibited actions (Delete, drop...)', ['query'])]
```
lmignon: you should put the complete list of prohibited words in your error message to help the user to fix the problem

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_sql_query(self):
+        test = self.sql_model.export_sql_query(
+            self.cr, self.uid, [self.query_id])
+        wizard = self.registry('sql.file.wizard').browse(
+            self.cr, self.uid, test['res_id'])
+        export = base64.b64decode(wizard.file)
+        self.assertEqual(export.split(';')[0], 'name')
+        self.assertTrue(len(export.split(';')) > 6)
+
```
lmignon: I would be nice to have a test to check that prohibited words are detected within this kind of queries

``` python
t1 = """create
"""
t2 =""" create """

t3 = """ col_create """

t4 = """ create_col """

tx = ....
```

lmignon:
@florian-dacosta Thank you, good job!  A few comments remain but the module looks fine.

florian-dacosta:
@lmignon No problem, I took your last comments into consideration

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,79 @@
...
+            'name': 'test delete3',
+            'query': """SELCT id FROM sale_order"""
+        }
+        ok_query = {
+            'name': 'test ok',
+            'query': "SELECT create_date FROM sale_order"
+        }
+        with self.assertRaises(except_orm):
+            self.sql_model.create(self.cr, self.uid, prohibited_query1)
+            self.sql_model.create(self.cr, self.uid, prohibited_query2)
```
lmignon: You should repeat the 'with' statement before each call to `self.sql_model.create`
Currently, the assertion is true once one call fails. 

``` python
def test_prohibited_queries_creation(self):
    prohibited_queries = [
        "upDaTe sale_order SET partner_id = 1 WHERE id = 1",
        "DELETE FROM sale_order;",
        "  DELETE FROM sale_order   ;",
        """DELETE
         FROM
         sale_order
         """,
    ]
    for query in prohibited_queries:
        with self.assertRaises(except_orm):
            self.sql_model.create(self.cr, self.uid, {'name': 'test_prohibited',
                                                      'query': query})
    wrong_syntax_query = {
        'name': 'test delete3',
        'query': """SELCT id FROM sale_order"""
    }
    with self.assertRaises(except_orm):
        self.sql_model.create(self.cr, self.uid, wrong_syntax_query)

    ok_query = {
       'name': 'test ok',
        'query': "SELECT create_date FROM sale_order"
    }
    query_id = self.sql_model.create(self.cr, self.uid, ok_query)
    self.assertIsNotNone(query_id)
```

lmignon:
:+1:  once travis is green

sql_export/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from . import test_sql_query
```
lmignon, line 2: Your tests are never launched by odoo .
You need to put your test class into a checks list

``` python
checks = [
    test_sql_query
]
```

florian-dacosta, line 2: I did not know that! I did unit test the same way very recently for a module in v8, and odoo was launching the tests without any troubles.
Is it specific to the v7?

Thanks once again for your help!

lmignon, line 2: in v8, tests are now automatically launched  if they are written under the module 'tests' in a submodule with a the name starting with test_
But before V8, we need to add the mentioned lines to say to Odoo which tests must be launched

florian-dacosta, line 2: Well, I can't see if it ran the tests.
Strange, because in localhost, my odoo did run the tests without any problem.

laetitia-gangloff:
:+1: (functionnal use)
nice module ! thank you !

sql_export/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{'name': 'SQL Export',
+ 'version': '0.1',
+ 'author': 'Akretion,Odoo Community Association (OCA)',
+ 'website': 'http://www.akretion.com',
+ 'description': """
```
dreispt, line 26: <s>If you leave this description key here, it will override the README. Maybe it's best to remove it.</s>
Sorry - not true for v7.

sql_export/sql_export.py
```diff
@@ -0,0 +1,147 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv.orm import Model
+from openerp.osv import fields, orm
+from openerp.tools.translate import _
+from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
+import StringIO
```
dreispt, line 22: AFAIK standard library imports should go first

sql_export/sql_export.py
```diff
@@ -0,0 +1,147 @@
...
+
+    _defaults = {
+        'group_ids': _get_editor_group,
+        'copy_options': "CSV HEADER DELIMITER ';'",
+    }
+
+    _constraints = [(_check_query_allowed,
+                    'The query you want make is not allowed : prohibited '
+                     'actions (delete, drop, insert, alter, truncate, '
+                     'execute, create, update)', ['query'])]
```
dreispt: Nitpick: maybe use 

```
                'The query you want make is not allowed : prohibited '
                 'actions (%s)' % PROHIBITED_WORDS.join(', '),
                 ['query'])]
```

sql_export/sql_export_view.xml
```diff
@@ -0,0 +1,61 @@
...
+        <field name="name">SQL Export</field>
+        <field name="res_model">sql.export</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form</field>
+    </record>
+
+
+    <menuitem id="sql_export_menu" name="Sql Export" parent="menu_export" sequence="1"/>
+
+    <menuitem id="sql_export_menu_view" name="Sql Export" parent="sql_export_menu" action="sql_export_tree_action" sequence="1"/>
```
dreispt, line 52: IIMO we're polluting the top menu with yet another option. 
Can we fit it in an existing top menu?
Since creating the SQL is quite a reserved feature, I suggest having it under Reporting\Configuration (`base.menu_reporting_config`).

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,65 @@
...
+#
+#    This program is distnaributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from openerp.tests.common import TransactionCase
+import base64
```
dreispt, line 16: Stdlib imports should go first

sql_export/sql_export_view.xml
florian-dacosta, line 52: I agree it is annoying to add one more option on the top menu.
The problem is that this feature can be used by almost every user.
Yes, just one or two user should be able to create a new sql request, but the requests can be executed by users in sales, purchases, warehouse, etc...
For each request, you can add groups or users, that will determine who can read this request. That is why we need to put the menu somewhere every users can have access.

dreispt, line 52: AFAIK the Reporting menu is available for all, only it's option are limited per security group. Can you try that?

florian-dacosta, line 52: You are right, it should be ok in the reporting menu.
I'll just add a menu under Reporting/
I don't think it belongs to the configuration submenu

dreispt, line 52: Agreed. Use "SQL" in uppercase in the menu labels?

laetitia-gangloff:
I think it is possible to merge.

gurneyalex:
:+1:

PR #176, opened 2015-06-09
[IMP] define a form view for wizard lines

hbrunn:
On a large screen, the purge button is quite far from the record's name, making it quite inconvenient to purge exactly one line. With this, we can simply click on the name, verify we're opening the right object, and purge it.


pedrobaeza:
Maybe a better option is to move the button at the beginning (first column)

PR #178, opened 2015-06-11
[FIX] ie related fields don't have an ir.model.fields entry

hbrunn:
The same happens in upstream already for read logging: https://github.com/OCA/server-tools/blob/8.0/auditlog/models/rule.py#L361

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -389,9 +389,10 @@ class auditlog_rule(models.Model):
             if field_name in FIELDS_BLACKLIST:
                 continue
             field = self._get_field(log.model_id, field_name)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thanks

:+1:

yvaucher:
:+1:

pedrobaeza:
Holger, please put a comment indicating why the "if" clause to serve as memo and I'll merge this one to move on, because there is other people with the same problem.

hbrunn:
@pedrobaeza done, I also added the comment to the case where it's already checked

pedrobaeza:
Thanks, I merge

PR #179, opened 2015-06-11
[ADD] web_norobots module

RoelAdriaans:
When you have a Odoo installation that is publicly accessible Google and other robots will index your site. You don't always want that.

This module  disallows indexing for webcrawlers that stick to the standard robots.txt guidelines.


RoelAdriaans:
@oca-clabot I've send my CLA on 16-4-2015 .. 
Resending it cla again..

web_norobots/views/disable_robots.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="website.robots">
```
pedrobaeza: Isn't this inserted in any other Qweb view?

pedrobaeza:
Thanks for your contribution.

We always try to put singular in all module names, so can you please put _website_no_robot_ or maybe _website_no_crawler_ as name?

max3903:
CLA is OK

web_norobots/views/disable_robots.xml
RoelAdriaans: No, the only view where website.robots is defined is addons/website/views/website_templates.xml
It is used in addons/website/controllers/main.py to display the robots.txt file in that http.route.
There are no extensions of the website.robots view anywhere but there..

RoelAdriaans:
@pedrobaeza No problem, I think web_no_crawler is indeed a better name, robots might be too technical.

dreispt:
Should this be proposed for the [`website`](https://github.com/OCA/website) project instead?

RoelAdriaans:
@dreispt Well, @sebalix opened an issue on B-Informed/web_norobots#1 and proposed to add it to the server-tools.
I agree because it does not add anything visible for the user and is probably installed by the server manager. Probably better for a normal user not to install this because it might do unexpected things. :)

sebalix:
Indeed, because some Odoo instances are visible on search engine even if the `website` module is not installed, so the `web_` prefix seems better.
Regarding server-tools/ or web/ repository, any feedback? We already have `web_context_tunnel` in server-tools/, so I think this module has its place here too.
Otherwise  :+1:

web_no_crawler/README.rst
```diff
@@ -0,0 +1,53 @@
+Disable robots crawling for Odoo websites
+=========================================
+
+By default, /robots.txt on an Odoo installation with de website module installed will allow indexing by webcrawlers.
+
+This module will overwrite view that generates /robots.txt and will disablow indexing.
```
dreispt, line 6: s/disablow/disalow

dreispt:
I just noticed that the module requires `website`, and only makes sense when `website` is installed.
I would find it surprising that installing a server-tools module, that would activate `website`on the instance.
So, as is, I think it should go in  the `website` repo.

Possibly dumb question: does it makes sense / is it possible to have this feature without the `website` dependency?

sebalix:
This is not dumb, I am ;) Didn't see the dependency against `website`, you are right.
At first I thought it was possible for a search engine to index an Odoo instance (without `website` installed), apparently not.
@RoelAdriaans-B-informed sorry for misleading you, OCA/website seems the right place for your module (this repo did not exist when I opened the issue), and can you renamed it `website_no_crawler` as suggested by @pedrobaeza ?

RoelAdriaans:
Indeed, @sebalix , robots.txt is only available when the website module is installed.

I will to a pull request in website the website repo.

PR #185, opened 2015-06-16
[ADD] Module ir_action_mail_translate

ghost:


ir_action_mail_translate/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+    """,
+    'depends': [
+        'base',
+    ],
+    'data': [
+    ],
+    'demo': [],
+    'test': [],
+    'external_dependencies': {
+        'python': ['num2words'],
```
max3903: Do you really need num2words ?

ghost: No, it's a mistake

ghost:
@max3903 You may remove needs fixing label.

ir_action_mail_translate/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+    'author': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'category': 'other',
+    'license': 'AGPL-3',
+    'description': """
+Server Action Email Translate
+=============================
+
+This module allows to translate server action email messages in the
+recipient's language.
```
max3903: Can you change the description to:

```
By default, emails used in server action are sent in english, even if they are translated. 
This module allows to send those email messages in the partner's language or 
the admin user language if there is no partner yet (HR applicant for example).
```

ir_action_mail_translate/models/ir_actions_server.py
```diff
@@ -0,0 +1,84 @@
...
+                'pool': self.pool,
+                'time': time,
+                'cr': cr,
+                'context': dict(context),
+                'uid': uid,
+                'user': recipient,
+            }
+
+            try:
+                address = eval(str(action.email), cxt)
```
StefanRijnhart: I think you should use safe_eval from openerp.tools.safe_eval.
Do you need the string casting?

ir_action_mail_translate/models/ir_actions_server.py
```diff
@@ -0,0 +1,84 @@
...
+            except Exception:
```
StefanRijnhart, line 65: Can you catch a more specific exception? -- edit -- void as per the above.

StefanRijnhart:
Looks good, apart from the little nag!

ir_action_mail_translate/models/ir_actions_server.py
ghost: @StefanRijnhart I admit this part of the code is a bit odd. I copied the 4 lines from addons/base/ir/ir_actions.py line 624. I would keep it this way to ensure that in every case, the value of the variable address will be the same here as in addons/base/ir/ir_actions.py. What do you think ?

StefanRijnhart: OK. Just use safe_eval, because that is what ir_actions.py uses (see imports in that file).

ghost: Done

ghost:
@max3903 you may remove needs fixing label

StefanRijnhart:
Thanks! :+1:

dreispt:
Isn't this also a candidate for the fresh `OCA/social` repo?

ghost:
@StefanRijnhart @max3903 what do you think ? should I move the PR to OCA/social ? I think it would be more acurate than server-tools

StefanRijnhart:
Go ahead if you like!

ghost:
@dreispt Could you create the branch 7.0 in OCA/social please ?

PR #186, opened 2015-06-17
[MIG] Base Custom Attributes module

ghost:
@guewen @pedrobaeza @florian-dacosta
Moved https://github.com/OCA/product-attribute/pull/71 here


rvalyi:
:-1:
Hello @dufresnedavid, First thanks for the migration work. However, I'm voting against the merge in the current form.

The reason is you are loosing both the copyright detail and the history of the module. And as Akretion is the original author of the module, that's annoying for us. You may think it doesn't really matter, but with the kind of licensing mess imposed by Odoo SA on the core and trying to influence the whole ecosystem, this kind of topic becomes really sensible. Like are AGPL modules compatibles with proprietary modules even without depending? A grey area, but if somebody needs to get his rights defended that would be the copyrightholders, so that kind of topic is important...

So please use git subtree to extract the module from the product-attributes branch and the rebase it on the server-tools branch. See https://makingsoftware.wordpress.com/2013/02/16/using-git-subtrees-for-repository-separation/

Really sorry to annoy you with that when the migration work looks solid otherwise.

rvalyi:
As for the copyright details, please keep it with this form
https://github.com/OCA/product-attribute/blob/7.0/base_custom_attributes/custom_attributes.py#L5

ghost:
@rvalyi Ok, I agree with you

ghost:
@rvalyi did as you asked. The module base_custom_attributes was imported from a subtree. Also fixed the copyright issue.

PR #192, opened 2015-06-25
[ADD] Language path mixin

StefanRijnhart:
# Language path mixin

This is a technical module to restore the possibility of Odoo to print reports
in other languages than the user's language (for instance, the customer's
language in the case of a sale order).

Odoo 8.0 has lost that capability due to an unlucky combination of technical
aspects of the deprecated RML report functionality and the new API. While the
static content of the report is translated fine, any translatable fields will
still be rendered in the language of the user.

See https://github.com/odoo/odoo/issues/7301 for the original bug report.

This module provides a tool for developers to work around this bug in their
Python code.
# Configuration

After installation of the module, you can import the mixin class that it
provides and have any model inherit from it in your python class definition.

You can then assign your class a __language_path_ member to indicate where
to find the language into which its reports are to be translated. See the
following code example:

.. code::

```
class SaleOrder(models.Model):
    _name = 'sale.order'
    _inherit = ['sale.order', 'language.path.mixin']
    _language_path = 'partner_id.lang'
```

In RML reports for such a model, you can then iterate over the records in the
correct language using

```
[[ repeatIn(objects.with_language_path(), 'o') ]]
```


StefanRijnhart:
@nhomar why the _in progress_ label?

antespi:
@StefanRijnhart this addon applies to qweb reports too?

And in res.partner form view, for instance? We have a customer that wants to show fields in partner form in the partner language, not in the user one. I don't know if i can use context `{'lang': lang}` to do it.

StefanRijnhart:
@antespi AFAIK there is no problem with this in qweb reports. As for showing form views in the cusomer language, this has far reaching implications for which this module will not help you.

yvaucher:
@StefanRijnhart I think the in progress is added autamatically by @nhomar because of Waffle

language_path_mixin/README.rst
```diff
@@ -0,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Language path mixin
+===================
+This is a technical module to restore the possibility of Odoo to print reports
```
dreispt: Make it clearer the it targets RML reports?
s/print reports/print RML reports

dreispt:
:+1:

antespi:
:+1:

StefanRijnhart:
@dreispt yes, indeed.

StefanRijnhart:
@yvaucher i see. Of course, these tests should be fixed at some point

gurneyalex:
elegant solution. 

:+1:

lfreeke:
:+1:  (test)

StefanRijnhart:
@dreispt description updated.

PR #194, opened 2015-06-29
Update Tests to Correctly Get DB

gdgellatly:
Upstream replaced the constant DB with a function.  Just a little one but should fix failing test

### DIFF
--- a/auth_from_http_remote_user/tests/test_res_users.py
+++ b/auth_from_http_remote_user/tests/test_res_users.py
@@ -45,11 +45,13 @@ class test_res_users(common.TransactionCase):
 
     def test_login(self):
         res_users_obj = self.registry('res.users')
-        res = res_users_obj.authenticate(common.DB, 'admin', 'admin', None)
+        res = res_users_obj.authenticate(
+            common.get_db_name(), 'admin', 'admin', None)
         uid = res
         self.assertTrue(res, "Basic login must works as expected")
         token = "123456"
-        res = res_users_obj.authenticate(common.DB, 'admin', token, None)
+        res = res_users_obj.authenticate(
+            common.get_db_name(), 'admin', token, None)
         self.assertFalse(res)
         # mimic what the new controller do when it find a value in
         # the http header (HTTP_REMODE_USER)
@@ -60,10 +62,11 @@ class test_res_users(common.TransactionCase):
         with mock_cursor(self.cr):
             # We can verifies that the given (uid, token) is authorized for
             # the database
-            res_users_obj.check(common.DB, uid, token)
+            res_users_obj.check(common.get_db_name(), uid, token)
 
             # we are able to login with the new token
-            res = res_users_obj.authenticate(common.DB, 'admin', token, None)
+            res = res_users_obj.authenticate(
+                common.get_db_name(), 'admin', token, None)
             self.assertTrue(res)
 
     @unittest.skipIf(os.environ.get('TRAVIS'),


pedrobaeza:
Runbot is still failing due to that nasty trick implemented on Travis that rename/copy the template for "server_environment_files" module.

OSguard:
Is their a way to fix this?

pedrobaeza:
I think the way is to rename `server_environment_files_sample` to `server_environment_files`

gdgellatly:
Have just been looking at this.  Will update pr tomorrow or create new
one.

gdgellatly:
I've made a number of edits to make this pass, but not overly happy with it.  I see in the OCA runbot we have some extra functionality to specify extra build commands for a repo.  I think if we can do that with a combination of some of the changes that will be better.  @gurneyalex, what do you think?

max3903:
@gurneyalex is on vacation...

gurneyalex:
@pedrobaeza no we cannot rename it to server_environment_files because then we get addons path issues and server_environment_files is supposed to be user suppled. 

@gdgellatly using extra build commands on the runbot currently entails explicitely listing the addons to test which is not convenient as the list of available addons differ in the various branches (I'm thinking of https://github.com/OCA/runbot-addons/tree/8.0/runbot_build_instructions here). 

What I was thinking about is adding an environment variable to the runbot set up and add a small hook in the server_environment set up to act when the variable is set (and rename the sample module). This will also be useful to conditionnally remove some WARNING / ERROR log which cause the CI to be red, but that we want to see in production instances (cf https://github.com/OCA/partner-contact/pull/120)

gdgellatly:
OK I will revert this to the original bug fix.

gdgellatly:
@gurneyalex @pedrobaeza reverted to original bug fix

gdgellatly:
@gurneyalex I was thinking the same thing, and that was the way I tried to implement.  No real need for a hook, once the path is set it is all good.  But without access to the server I was trying to find a run option that was unique to runbot. All it needs for this module is just testing for variable and switching the _dir command to the sample module.  

I would do it as 

``` python
try:
    from openerp.addons import server_environment_files
except ImportError:
    if os.environ.get('runbot'):  # or maybe 'demo'
        from openerp.addons import server_environment_files_sample as server_environment_files
        system_base_config['running_env'] = system_base_config.get('running_env', 'dev')
    else:
        raise ...
_dir = os.path.dirname(server_environment_files.__file__)

```

gdgellatly:
@gurneyalex Sorry I was wrong because of the dependency, in my implementation I actually moved all the files to an example folder and renmaed but it is impractical.  I also tried using sed but that didn't work.

Renaming is simplest.  Or else an alternate **openerp**.py file, say **runbot**.py that gets read instead if present and puts in the sample dependency.

gdgellatly:
Actually we could just use the above code and remove the dependency on server_environment_file from these 2 modules.  Think about all the modules that use decimal_precision but don't depend on it.  It will still raise an error, in fact we could make it raise the same as though it was a dependency.  Provided it is documented that you need to install that module first or separately then it should be OK.

gurneyalex:

gdgellatly:
@gurneyalex rebased, not sure who benefits here ;) btw, not sure how to trigger a rebuild properly on a rebase so I just close and reopen.

max3903:
:+1:

pedrobaeza:
:+1:

PR #196, opened 2015-06-30
[7.0] new module: record_archiver

guewen:
New module _record_archiver_.
This module allows to configure a lifespan for different models. When this lifespan is reached, the records are archived (with some exceptions such as records not being `done` or `canceled`).
The default archiving strategy is to set the `active` flag to `False`.


gurneyalex:

guewen:
@gurneyalex done

yvaucher:
It has lost it's alliteration :sob: 

:+1: Thanks for the tests and cleaning

It needs a separation in travis tests with the module adding `notification_email_send`

```
IntegrityError: null value in column "notification_email_send" violates not-null constraint
```

record_archiver/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+performance.
+
+Records are deactivated based on their last activity (write_date).
+
+You can configure lifespan of each type of record in
+`Settings -> Configuration -> Records Archiver`
+
+Lifespan is defined per record per company.
+ """,
+ 'author': 'Camptocamp',
```
lmignon: OCA is not listed in the list of authors

record_archiver/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+{'name': 'Records Archiver',
+ 'version': '0.1',
+ 'description': """
+Create a cron job that deactivates old records in order to optimize
```
lmignon, line 26: It would be nice to format the description with the same rules as for the 8.0 branch

guewen:
Rebased.

@lmignon you are right, fixed

> It needs a separation in travis tests with the module adding notification_email_send
> 
> IntegrityError: null value in column "notification_email_send" violates not-null constraint

:anguished: it is in the `mail` addon.

yvaucher:
It is the same issue as described here:

https://github.com/OCA/server-tools/blob/8.0/auth_from_http_remote_user/tests/test_res_users.py#L72-L80

yvaucher:
@lmignon as you are reviewing this PR you can comment on this error in travis? You know what cause this?

guewen:
My last commit excludes this module from the one-by-one tests. This error is very weird though because it shows that modules are impacting each other in these `UNIT_TESTS` tests which should exactly not be the case.

yvaucher:
:+1:

lmignon:
@yvaucher I've tried a lot of way to turn around this bug but failed to find the right solution. That's why I've added the `@unittest.skipIf(os.environ.get('TRAVIS'))` decorator in https://github.com/OCA/server-tools/blob/8.0/auth_from_http_remote_user/tests/test_res_users.py#L72-L80

lmignon:
:+1:

yvaucher:
@lmignon Thanks ok here @guewen disable it only when all unit tests are launched.

gurneyalex:
merging

PR #198, opened 2015-07-07
[ADD] mail_compose_select_lang

pedrobaeza:
# Select language in mail compose window

This module allows to select the language for the mail content directly in
the mail compose window.

It also translates attachments that can be linked to the email template.
# Usage

By default, mail compose window will use corresponding language for showing the
contents, but you will be able to select another language from the dropdown
box "Force language" in the bottom right part.


hbrunn:
@pedrobaeza would you care to move your PR to https://github.com/OCA/social ?

pedrobaeza:
Done in https://github.com/OCA/social/pull/2

PR #199, opened 2015-07-08
[FIX] Error logging on create

kittiu:
For V8, I installed on my instance, create Rule to log for sale.order. And when create, got following error,

  File "/home/kittiu/workspace/odoo_nstda/nstda_msd/auditlog/models/rule.py", line 324, in create_logs
    self._create_log_line_on_create(log, diff.added(), new_values)
  File "/home/kittiu/workspace/odoo_nstda/nstda_msd/auditlog/models/rule.py", line 438, in _create_log_line_on_create
    log, field, new_values)
  File "/home/kittiu/workspace/odoo_nstda/nstda_msd/auditlog/models/rule.py", line 446, in _prepare_log_line_vals_on_create
    'field_id': field['id'],
TypeError: 'bool' object has no attribute '__getitem__'

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -434,9 +434,10 @@ class auditlog_rule(models.Model):
             if field_name in FIELDS_BLACKLIST:
                 continue
             field = self._get_field(log.model_id, field_name)
-            log_vals = self._prepare_log_line_vals_on_create(
-                log, field, new_values)
-            log_line_model.create(log_vals)
+            if field:
+                log_vals = self._prepare_log_line_vals_on_create(
+                    log, field, new_values)
+                log_line_model.create(log_vals)
 
     def _prepare_log_line_vals_on_create(self, log, field, new_values):
         """Prepare the dictionary of values used to create a log line on a


pedrobaeza:
Uhm, curious, did you see which is the name of the field that throws the error?

kittiu:
It is a computed field without store value. Here it is,

```
@api.one
def _get_available_credit(self):
    self.credit_avail = self.credit_limit - self.credit - self.confirmed_sales

credit_avail = fields.Float(string='Available Credit', compute='_get_available_credit')
```

I am guessing it is because of the new API?

pedrobaeza:
I don't know enought the internals of the module to say it, but maybe we should pre-filter these fields or fetch them correctly. @sebalix, can you say anything of this.

sebalix:
Hi,

Do you use a recent version of Odoo 8.0 or OCA/server-tools (or maybe you download the module on the apps store)?
I have tested a log creation on `create` operation, on a model containing a Float computed field like yours, and it is working fine, even when the computed field is based on another field which is filled in the `create()` call.

``` python
class MyModel(models.Model):
    _name = 'my.model'

    name = fields.Char(u"Name", required=True)

    @api.one
    def _compute_plop(self):
        self.plop = 42.0 + len(self.name)

    plop = fields.Float(string='Plop', compute='_compute_plop')
```

I used the last version of _OCA/server-tools_ and almost the last version of _OCB_ for this test, may be I should test on Odoo.

sebalix:
Oh, I think this is a duplicate of the PR #178

pedrobaeza:
Yeah, that's it! Let's close this one then and move the other PR.

pedrobaeza:
Patch merged, please check again

PR #204, opened 2015-07-15
[ADD] base_suspend_security

hbrunn:
# Suspend security

This module was written to  allow you to call code with some `uid` while being sure no security checks (`ir.model.access` and `ir.rule`) are done. In this way, it's the same as `sudo()`, but the crucial difference is that the code still runs with the original user id. This can be important for inherited code that calls workflow functions, subscribes the current user to some object, etc.

Usually, you'll be in in the situation to want something like this if you inherit from a module you can't or don't want to change, and call `super()`.
# Usage

- depend on this module
- call `yourmodel.suspend_security().function_to_run()`, just the same as you would use `sudo()`

For further information, please visit:
- https://www.odoo.com/forum/help-1
# Known issues / Roadmap
- the magic works by wrapping uid in a marker class, so if some code unwraps this in the calling tree, security checks will be reenabled


hbrunn:
@StefanRijnhart just pointed me to a blatant security problem here, will be fixed soon

hbrunn:
@StefanRijnhart the problem is fixed now

base_suspend_security/tests/test_base_suspend_security.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class TestBaseSuspendSecurity(TransactionCase):
+    def test_base_suspend_security(self):
+        # tests are called before register_hook
+        self.env['ir.rule']._register_hook()
+        user_id = self.env.ref('base.user_demo').id
+        other_company = self.env['res.company'].create({
+            'name': 'other company',
+            # without this, a partner is created and mail's contraint on
```
StefanRijnhart: con**s**traint

StefanRijnhart:
Beautiful! :+1:

sbidoul:
Wow. I had been toying with this idea (in my head only) since a long time. 

This is very useful and a with a clever implementation.

:+1: (code review, no test)

gurneyalex:
I suggest you rebase to get the recently merged runbot fixes.

lmignon:
Brilliant!! :+1:  (code review)

pedrobaeza:
Merging as there is enough thumbs up and in main branch all checks will be OK

PR #206, opened 2015-07-16
port base_concurrency to 8.0

charbeljc:


dreispt:
Can you rebase please?

dreispt:
@mdietrichc2c Could this be better solved upstream? What if the login date is written only when needed - the first time of the day?
Something like this: https://github.com/odoo/odoo/compare/8.0...dreispt:lazy-login-date?expand=1

<s>(It still needs some timezone handling to be accurate...)</s>
(EDIT: just fixed it so that UTC is used)

mdietrichc2c:
@dreispt I only did the original on 7.0, and on that, the stable policy dictated a separate module from the main branch, hence the code shown here. I do agree that for 8.0 or master, depending on the policy, it would be preferable to put a better fix in the `base` module: the more upstream, the better.

RE: your branch: I like the solution, but you may still have concurrency issues if a user logs twice/opens 2 tabs on a new day (especially since it uses UTC; that means users with UTC-9 in California or UTC+10 in Australia could still have the issue), or 2 long automated processes launched af 23:50 UTC/00:10 UTC. A separate table to avoid concurrent update on `res_users` remains for me the best solution.

charbeljc:
Module rebased, currently in use to load 115000 partners in 8 parallels jobs, without anymore concurrent updates killing my jobs and all CPUs working to the max and not so bad IO perfs without tuning. Hey this module saved my month, actually ! Thanks @mdietrichc2c !

dreispt:
@charbeljc You were getting "TransactionRollbackError: could not serialize access due to concurrent update" errors?

charbeljc:
@dreispt yes

gurneyalex:
:+1: 

@charbeljc please rebase.

charbeljc:
@gurneyalex rebased

mdietrichc2c:
:+1:

gurneyalex:
:+1:

sebalix:
Cool module, thank you! It saves my day.

PR #209, opened 2015-07-17
[IMP] allow to select lines to purge in a tree view

hbrunn:
When there are thousands of lines to review (data entries are a likely candidate for that) and you don't feel lucky enough to simply purge everything, reviewing is quite impossible in the one2many view. This PR adds a button to show the lines in a tree view. Then we can filter the way we want and only purge a selections. Feels much safer.

This supersedes #176


pedrobaeza:
:+1:

max3903:
:+1:

PR #213, opened 2015-07-20
Mail cleanup

mdietrichc2c:
This PR adds a new module, 'mail_cleanup', in order to add a step before fetching e-mails.

The process is to mark as read and (optionally) move messages older than 30 days, in order to avoid keeping un-routed messages as unread; the "expiration date" and the folder where the message should be moved are both set with environment variables.


mail_cleanup/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+
+This module depends on mail_environment in order to add "expiration dates"
+per server.
+
+Example of a configuration file:
+
+[incoming_mail.openerp_pop_mail1]
+cleanup_days = 30 (default value)
+cleanup_folder = NotParsed (optional)
+    """,
```
gurneyalex: can you move the description to README.rst (using the OCA template)?

mail_cleanup/__openerp__.py
```diff
@@ -0,0 +1,53 @@
...
+those unroutable e-mails.
+
+This module depends on mail_environment in order to add "expiration dates"
+per server.
+
+Example of a configuration file:
+
+[incoming_mail.openerp_pop_mail1]
+cleanup_days = 30 (default value)
+cleanup_folder = NotParsed (optional)
```
gurneyalex: use a comment for things in parenthesis.

mail_cleanup/mail_cleanup.py
```diff
@@ -0,0 +1,144 @@
...
+            _logger.info('start cleaning up emails on %s server %s',
+                         server.type, server.name)
+            context.update({'fetchmail_server_id': server.id,
+                            'server_type': server.type})
+            count, failed = 0, 0
+            imap_server = False
+            if server.type == 'imap' and server.cleanup_days > 0:
+                try:
+                    expiration_date = (datetime.now() +
+                                       relativedelta.relativedelta(
```
gurneyalex: I would extract this big block in a helper function. This will save you 12 columns which means fewer awkward line splits.

dreispt:
Somewhat on the border line, but since it's mail/messaging related I wonder if this wouldn't be better hosted at OCA/social?

mdietrichc2c:
@gurneyalex : the README.rst was added (and formatted), and a new helper function was created. I also cleanup up the calls to the `imaplib` library.

mail_cleanup/mail_cleanup.py
```diff
@@ -0,0 +1,135 @@
...
+            imap_server = False
+            if server.type == 'imap' and server.cleanup_days > 0:
+                try:
+                    self._cleanup_fetchmail_server(server)
+                except Exception:
+                    _logger.exception("General failure when trying to cleanup "
+                                      "mail from %s server %s.",
+                                      server.type, server.name)
+                finally:
+                    if imap_server:
```
gurneyalex: this is never True, with the refactoring you made.

mdietrichc2c:
@gurneyalex thanks; it's now fixed.
I also rebased the various fixes in a single commit.

mail_cleanup/mail_cleanup.py
```diff
@@ -0,0 +1,135 @@
...
+from osv import osv
+from dateutil import relativedelta
+from datetime import datetime
+
+from server_environment import serv_config
+
+_logger = logging.getLogger(__name__)
+
+
+class FetchmailServer(osv.osv):
```
gurneyalex: use orm.Model (osv.osv is deprecated)

mail_cleanup/mail_cleanup.py
```diff
@@ -0,0 +1,135 @@
...
+                           'cleanup_folder': False}
+            if serv_config.has_section(global_section_name):
+                config_vals.update(serv_config.items(global_section_name))
+
+            custom_section_name = '.'.join((global_section_name,
+                                            fetchmail.name))
+            if serv_config.has_section(custom_section_name):
+                config_vals.update(serv_config.items(custom_section_name))
+
+            if config_vals.get('cleanup_days'):
```
gurneyalex: as far as I can tell, this is always true (you set a default on line 46)

mail_cleanup/mail_cleanup.py
```diff
@@ -0,0 +1,135 @@
...
+                config_vals['cleanup_days'] = \
+                    int(config_vals['cleanup_days'])
+            if config_vals.get('cleanup_folder'):
```
gurneyalex: this is always true (key set on line 47)

and the block inside the test does nothing -> please clean it (or do something useful in there)

gurneyalex:
:+1:

mdietrichc2c:
New additions: a new variable to delete messages instead (or after) archiving them, and the cleanup is now optional (if you only want to delete messages).

vrenaville:
:+1:

mail_cleanup/__openerp__.py
```diff
@@ -0,0 +1,88 @@
...
+Mail cleanup
+===========
+
+This module allows to mark e-mails older than x days as read
+and optionally move them on IMAP servers, just before fetching them.
+
+Since the main "mail" module does not mark unroutable e-mails as read,
+this means that if junk mail arrives in the catch-all address without
+any default route, fetching newer e-mails will happen after re-parsing
+those unroutable e-mails.
```
yvaucher, line 41: Is it wanted to have 2 versions of this text with the one in README.rst?

mdietrichc2c, line 41: It was due to a previous comment asking for the README, before realizing this is for 7.0.
I'll remove the README.

yvaucher:
Thanks :+1:

gurneyalex:
@dreispt this feels a bit too borderline to me. I'm in favor of keep this in this repository as it is typically a backend related issue.

PR #214, opened 2015-07-22
Add module datetime_formatter.

yajo:
This module allows the developer to format any date/time combination in the most comfortable way for the user, just with a quick call to :meth:`~.datetime_formatter` under any context.


pedrobaeza:
Can you explain the advantages of this against `formatLang` or `from_string` and then datetime functions?

datetime_formatter/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+If you are a developer, to use this module, you need to:
+
+* Call anywhere in your code::
+
+    formatted_string = self.env["res.lang"].datetime_formatter(datetime_value)
+
+* If you use Qweb::
+
+    <t t-esc="env['res.lang'].datetime_formatter(datetime_value)"/>
```
yajo, line 30: @pedrobaeza This is the main andvantage. The method resides in a DB model, so it can be called from wherever the registry is available, as in Qweb views. You cannot import from other modules in Qweb unless you create a custom controller for that view. This is much simpler.

datetime_formatter/models.py
```diff
@@ -0,0 +1,60 @@
...
+
+        :param str separator:
+            Only used when :param:`template` is :const:`MODE_DATETIME`, as the
+            separator between the date and time parts.
+        """
+        # Get the correct lang
+        if not lang:
+            lang = (self.env.context.get("lang") or
+                    self.env.user.lang or
+                    self.search([("active", "=", True)], limit=1).code)
```
yajo: @pedrobaeza Also this is another advantage: automatic detection of best matched language.

pedrobaeza:
OK, I see

:+1: then

datetime_formatter/__openerp__.py
```diff
@@ -0,0 +1,12 @@
+# -*- encoding: utf-8 -*-
+
+{
+    "name": "Date & Time Formatter",
+    "summary": "Helper functions to give correct format to date[time] fields",
+    "version": "1.0.0",
+    "author": "Odoo Community Association (OCA)",
+    "category": "Tools",
+    "depends": [
+        "base",
```
tafaRU, line 16: It seems to me that it's redundant since 8.0, isn'it?

tafaRU:
@Yajo thanks a lot for the module!
Could you please tell me the reason to have icon in both format?

The rest LGTM :+1:

datetime_formatter/__openerp__.py
yajo, line 16: Seems a common practice in Odoo source code. It may be redundant, but the developer reading this manifest immediately knows that this module needs nothing but the bare Odoo server. It's an opinion, though.

tafaRU, line 16: Just see the same comment in other PRs, this is the reason. No problem to keep it :wink:

datetime_formatter/static/description/icon.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hbrunn: I don't think you need to include the unchanged default icon's source code here

datetime_formatter/models.py
hbrunn: you might want to check before if self.ids is truthy, then you were called on a browse record, and in this case, simply use self[:1][0]. This way, we can say things like `my_object.lang.datetime_formatter(mydate)` and get it in the object's languge

hbrunn: passing active=True is superfluous btw, this is added automatically by odoo

datetime_formatter/static/description/icon.svg
pedrobaeza: Yeah, I missed this part, but I agree

datetime_formatter/models.py
yajo: > passing active=True is superfluous btw, this is added automatically by odoo

Are you sure? Do you have any relevant docs/code?

yajo:
Icon removed, new rule prepended to lang match finder. I also added comments to those rules, so they are easier to understand.

yajo:
I'll write tests now.

yajo:
Test passing, bugs fixed, and now you can autoconvert float time fields too.

yajo:
What's remaining here?

rafaelbn:
cc @antespi @JavierIniesta @sergio-incaser @carlos-incaser

antespi:
:+1: Cool and easy to use, Thanks!

JavierIniesta:
Very useful :+1:

sebalix:
Indeed :+1:

datetime_formatter/__openerp__.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+{
+    "name": "Date & Time Formatter",
+    "summary": "Helper functions to give correct format to date[time] fields",
+    "version": "1.0.0",
```
lmignon: Can you use the right version scheme: `8.0.1.0.0`  https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

lmignon:
Indeed, a useful addon. :+1: once the version number is fixed.

carlosdauden:
:+1:

datetime_formatter/README.rst
```diff
@@ -0,0 +1,68 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza: Put latest conventions on README.rst file according https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

yajo:
Fixed, rebased, squashed.

datetime_formatter/__openerp__.py
```diff
@@ -0,0 +1,14 @@
...
+{
+    "name": "Date & Time Formatter",
+    "summary": "Helper functions to give correct format to date[time] fields",
+    "version": "8.0.1.0.0",
+    "author": "Odoo Community Association (OCA)",
+    "category": "Tools",
+    "depends": [
+        "base",
+    ],
+}
```
moylop260, line 18: installable key?

moylop260:
Could you fix next lint checks:

``` bash
************* Module datetime_formatter
datetime_formatter/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:36 No role entry for "meth" in module "docutils.parsers.rst.languages.en".|Trying "meth" as canonical role name.
datetime_formatter/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:36 Unknown interpreted text role "meth".
************* Module datetime_formatter.__openerp__
datetime_formatter/__openerp__.py:5: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
************* Module datetime_formatter.__openerp__
datetime_formatter/__openerp__.py:5: [C8102(manifest-required-key), ] Missing required key "installable" in manifest file
```

datetime_formatter/__openerp__.py
yajo, line 18: Defaults to True, not needed.

pedrobaeza, line 18: Not needed, but OCA flow requires it.

yajo:
I'm not used to Sphinx, let's try and if it does not work I'll just remove the `:meth:` stuff.

yajo:
I'm pretty sure I wrote it fine, but checks fail, so I'm removing `:meth:`.

gurneyalex:
@Yajo thanks!

PR #216, opened 2015-07-27
add time_working_day module 

chafique-delli:
The code of this module was extracted from the module 'openerp-stock-delivery-times/stock_delivery_times_working_days' in v7 and ported in v8 to a more generic module.


dreispt:
I suggest to name it `resource_calendar_extensions` so that it can also host other calculation extensions. 
Would it be best hosted under `OCA/hr`?

Question: the resource calendar does not already provides methods for this?

chafique-delli:
@dreispt 
Hi, thanks for your review.
For your first suggestion, the name 'resource_calendar_working_day' seems a good compromise.
But for the choice of repository, I will wait for the return of the other members.
The resource calendar does not provides methods to compute a date with working days.

dreispt:
@chafique-delli Have you tried [`schedule_days_get_date()`](https://github.com/odoo/odoo/blob/8.0/addons/resource/resource.py#L555)? I think it does what you want.

There is no similar function for calculations in hours , so I had to implemented a [solution](https://github.com/OCA/project-service/blob/8.0/project_sla/project_sla_control.py#L127) for `project_sla`, and it could be isolated in a reusable module such as this.

gurneyalex:
This looks nice. If you could add an automated test, it would be wonderful. You'll get bonus points for converting the README.rst file to the OCA template (see https://github.com/OCA/maintainer-tools/tree/master/template/module be careful there are some place holders to fill in)

:+1:

pedrobaeza:
Now that I have worked with resource module, I agree with @dreispt that the mentioned method already do the same, so :-1:

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+    @api.model
+    def _get_date(self, start_date, delay, resource_id=False):
+        """This method gives the first date after a delay from the start date
+            considering the working time attached to the company calendar.
+        """
+        if isinstance(start_date, str):
+            start_date = fields.Date.from_string(start_date)
+        if not self.id:
+            model = self._context['params']['model']
+            obj_id = self._context['params']['id']
```
guewen, line 38: `self.env.context` instead of accessing to the private `self._context`

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+            obj = self.env[model].browse(obj_id)
+            self = obj.company_id.calendar_id
+            if not self:
+                raise Warning(_('Error !'), _(
+                    'You need to define a calendar for the company !'))
```
guewen, line 43: The `Warning` exceptions does not take 2 args but 1.
You can remove the `!`, no need to be so exclamatory.

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+
+class ResourceCalendar(models.Model):
+    _inherit = 'resource.calendar'
+
+    @api.model
+    def _get_date(self, start_date, delay, resource_id=False):
+        """This method gives the first date after a delay from the start date
+            considering the working time attached to the company calendar.
+        """
+        if isinstance(start_date, str):
```
guewen, line 34: `basestring` is better than `str` since it includes `unicode` too.

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+            start_date = fields.Date.from_string(start_date)
+        if not self.id:
+            model = self._context['params']['model']
+            obj_id = self._context['params']['id']
+            obj = self.env[model].browse(obj_id)
+            self = obj.company_id.calendar_id
```
guewen, line 40: Could you use a `calendar` variable instead of re-attributing `self`?

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+        dt_leave = self.get_leave_intervals(
+            resource_id, start_datetime=None, end_datetime=None)
+        worked_days = (
+            [day['dayofweek'] for day in self.attendance_ids])
+        if delay < 0:
+            delta = -1
+        else:
+            delta = 1
+        while datetime.strftime(
+            start_date, DEFAULT_SERVER_DATE_FORMAT) in dt_leave or str(
```
guewen, line 53: could use `fields.Date.to_string()`

resource_calendar_working_day/resource.py
```diff
@@ -0,0 +1,63 @@
...
+                start_date.weekday()) not in worked_days:
+            start_date = start_date + timedelta(days=delta)
+        date = start_date
+        while delay:
+            date = date + timedelta(days=delta)
+            if datetime.strftime(
+                date, DEFAULT_SERVER_DATE_FORMAT) not in dt_leave and str(
```
guewen, line 60: could use `fields.Date.to_string()`

guewen:
@chafique-delli could you check if you could do with https://github.com/OCA/server-tools/pull/216#issuecomment-125199926 ?

gurneyalex:
closing the PR as inactive. Please reopen if you want to resume work on this.

PR #220, opened 2015-07-27
[8.0] add graylog2_handler module

katyukha:


graylog2_handler/.gitignore
```diff
@@ -0,0 +1,7 @@
+*.pyc
+*.swp
+*.idea/
+*~
+*.swo
+*.pyo
+.ropeproject/
```
dreispt: Remove this file

graylog2_handler/__init__.py
```diff
@@ -0,0 +1,30 @@
+import logging
+import graypy
```
dreispt: You should use a try block here

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "GrayLog2 Handler",
+    "version": "0.0.1",
+    "author": "Management and Accounting On-line",
+    "website": "",
+    "summary": "Provides ability to send log messages to graylog2 server",
+    "category": "Added functionality",
+    "depends": [
+    ],
+    "description": """
```
dreispt: Move the description to a README file

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "GrayLog2 Handler",
+    "version": "0.0.1",
+    "author": "Management and Accounting On-line",
+    "website": "",
```
dreispt: I suggest setting website to the server-tools GitHub URL

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+    "summary": "Provides ability to send log messages to graylog2 server",
+    "category": "Added functionality",
+    "depends": [
```
dreispt: Is no other dependencies, it's good practice to at least depend on `base`

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+    "demo": [
+    ],
+    "test": [
+    ],
+    "update_xml": [
+    ],
+    "active": True,
+}
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
dreispt: Remove vim lines

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+    """,
+    "init_xml": [
+    ],
+    "demo": [
+    ],
+    "test": [
+    ],
+    "update_xml": [
+    ],
+    "active": True,
```
dreispt: <s>Remove auto install; add installable: True</s>
I just noticed that the module does not need installation to work.
This can be surprising, so it should be noted on the README.

graylog2_handler/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+import openerp.tools as tools
+
+
+_logger = logging.getLogger(__name__)
+_logger.info("GELF_Handler module loading...")
+
+gelf_enabled = tools.config.get('gelf_enabled', False)
+_logger.info("GELF Enabled: %s", gelf_enabled)
```
dreispt: Since the module does not need explicit installation, and you might have it in your addons path and not want to used it, I would prefer it to be silent unless "activated". 
You can use debug level for the two log messages above.

katyukha:
Seems to be fixed

graylog2_handler/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+_logger = logging.getLogger(__name__)
+
+# Check if this module is enabled in config
+if tools.config.get('gelf_enabled', False):
+    try:
+        import graypy
+    except ImportError:
+        _logger.error("Package 'graypy' not installed")
+    else:
```
dreispt, line 12: Maybe a nitpick, but do we really need for the next code to be in an `else` block?

dreispt:
<s>A minor nitpick, but it's</s> :+1:

graylog2_handler/__init__.py
katyukha, line 12: if next code is moved out from `else` block, and `graypy` is not installed, then error will be raised on import of this module. But if `else` block is present, just an error message will be written in log. no errors will be raised.

dreispt, line 12: Of course, you're right!

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "GrayLog2 Handler",
+    "version": "0.0.1",
+    "author": "Management and Accounting On-line",
+    "website": "https://github.com/OCA/server-tools/",
+    "summary": "Provides ability to send log messages to graylog2 server",
+    "category": "Added functionality",
+    "depends": [
+        'base',
+    ],
+    "init_xml": [
```
gurneyalex: plese remove empty file lists in the manifest.

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "GrayLog2 Handler",
+    "version": "0.0.1",
+    "author": "Management and Accounting On-line",
```
gurneyalex: please add "Odoo Community Association (OCA)" to the author

graylog2_handler/README.rst
```diff
@@ -0,0 +1,17 @@
+GrayLog2 GELF log handler
+=========================
```
gurneyalex, line 7: can you update the README to follow the OCA template (https://github.com/OCA/maintainer-tools/tree/master/template/module)

katyukha, line 7: Updated

gurneyalex:
can you rebase?

katyukha:
Rebased.

katyukha:
Is there somebody interested in this MR or I may close it?
We have no plans to port this addon to 9.0+ versions.

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "GrayLog2 Handler",
+    "version": "8.0.0.0.1",
+    "author": ("Management and Accounting On-line, "
+               "Odoo Community Association (OCA)"),
+    "website": "https://github.com/OCA/server-tools/",
+    "summary": "Provides ability to send log messages to graylog2 server",
+    "category": "Added functionality",
```
dreispt: I suggest changing Category to ["Extra Tools"](https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/module/module_data.xml#L83)

graylog2_handler/__openerp__.py
```diff
@@ -0,0 +1,13 @@
...
+    "depends": [
+        'base',
+    ],
+    "active": True,
```
dreispt: This key name is deprecated, and afaik it is teh equivalent to "auto_install", which we don't want.
Probably you wanted to use `"installable": True`

dreispt (changes requested):

dreispt (approved):

PR #221, opened 2015-07-29
New module to add timeout for fetchmail servers

mdietrichc2c:
This PR is to add what was done in https://github.com/OCA/OCB/pull/292 in a specific module.

### DIFF
--- a/mail_timeout/__openerp__.py
+++ b/mail_timeout/__openerp__.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author: Matthieu Dietrich
+#    Copyright 2015 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


mail_timeout/__openerp__.py
```diff
@@ -0,0 +1,83 @@
...
+#
+##############################################################################
+
+{
+    'name': 'Mail timeout',
+    'version': '0.1',
+    'category': 'Tools',
+    'author': "Camptocamp,Odoo Community Association (OCA)",
+    'summary': 'Clean up mails regularly',
+    'description': """
```
lepistone, line 28: Not needed: we have a `README`

pedrobaeza, line 28: This is v7, so unfortunately it's needed

pedrobaeza:
:+1:

I insist that maybe this should be treated as a bug and put in OCB a fixed timeout as workaround

gurneyalex:
@pedrobaeza sorry I was not fully aware of the issue and the initial PR or of the problem. 

If this is considered a bug, then we can probably have a hard coded timeout in the code without adding a field, indeed.

mdietrichc2c:
@pedrobaeza @gurneyalex I re-re-did this in OCB/odoo: https://github.com/odoo/odoo/pull/7790 and https://github.com/OCA/OCB/pull/294

dreispt:
If a parameters is needed, you can add it to `ir.config_parameter` instead of having a hardcoded value.

dreispt:
Remember that for OCB a fix PR to upstream Odoo is required.

PR #228, opened 2015-08-17
[FIX] users_ldap_groups - typo on _defaults dict

yvaucher:
It does nothing as it is False by default. But still it's wrong.

### DIFF
--- a/users_ldap_groups/__openerp__.py
+++ b/users_ldap_groups/__openerp__.py
@@ -21,7 +21,7 @@
 
 {
     "name": "Groups assignment",
-    "version": "1.2",
+    "version": "1.2.1",
     "depends": ["auth_ldap"],
     "author": "Therp BV,Odoo Community Association (OCA)",
     "description": """
--- a/users_ldap_groups/users_ldap_groups.py
+++ b/users_ldap_groups/users_ldap_groups.py
@@ -80,7 +80,7 @@ class CompanyLDAP(orm.Model):
                  'groups are preserved.')
     }
 
-    _default = {
+    _defaults = {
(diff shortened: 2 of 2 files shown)


alejandrosantana:
:+1:

users_ldap_groups/__openerp__.py
```diff
@@ -21,7 +21,7 @@
 
 {
     "name": "Groups assignment",
-    "version": "1.2",
+    "version": "7.0.1.2.1",
```
pedrobaeza: I think 7.0 doesn't support preprending major version, so I think the module version as is will be 7.0.7.0.1.2.1

hbrunn: and traditionally, you change version only when you need a migration

yvaucher: @hbrunn This was discussed here: https://github.com/OCA/maintainer-tools/pull/100

@pedrobaeza I'm unsure about that but ok let's keep it quiet in 7.0

yvaucher: @hbrunn This typo could have side effects I prefer to bump the fix counter

pedrobaeza: OK, I have just checked on runbot and there's no problem to keep it like that:

seleccion_001

yvaucher: @pedrobaeza already changed. Thanks for the check

pedrobaeza: Yeah, well, nevermind. We have to automate this checking/batch change I think.

gurneyalex:
:+1:

PR #229, opened 2015-08-18
Add missing default oca icons

yvaucher:

PR #230, opened 2015-08-18
[FIX] unconfigured server_environment doesn't play

hbrunn:
nice with static files


yvaucher:
:+1: restore previous behavior prior to #229 I fast-track

yvaucher:
linking https://github.com/odoo/odoo/issues/7507

moylop260:
FYI [CONTRIBUTING.md#external-dependencies](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies)

yvaucher:
@moylop260 this is not due to python external dependencies.

It tries to load a dependency `server_environement_files` module which should be created based on `server_environement_files_sample` so you need to create it and if you don't install `server_environement_files` you shouldn't need it.

PR #231, opened 2015-08-18
[FIX] like #230, but for mail_environment

hbrunn:
@yvaucher sorry to bother you a second time, but we need the same for `mail_environment`


yvaucher:
:+1: also fast-tracked

@hbrunn no problem cosmetic is far less important than having our tests and instances working

yvaucher:
linking https://github.com/odoo/odoo/issues/7507

PR #232, opened 2015-08-20
[FIX] users_ldap_populate: Protect ldap library import

pedrobaeza:
To avoid errors on dependent builds

### DIFF
--- a/users_ldap_populate/model/users_ldap.py
+++ b/users_ldap_populate/model/users_ldap.py
@@ -20,9 +20,15 @@
 ##############################################################################
 
 import re
-from ldap.filter import filter_format
 from openerp.osv import orm
 import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from ldap.filter import filter_format
+except ImportError:
+    _logger.debug('Can not `from ldap.filter import filter_format`.')
 
 
 class CompanyLDAP(orm.Model):
--- a/users_ldap_populate/__openerp__.py
+++ b/users_ldap_populate/__openerp__.py
@@ -41,6 +41,9 @@ object you want to query.
     "depends": [
         'auth_ldap',
     ],
+    'external_dependencies': {
+        'python': ['ldap'],
+    },
     "data": [
         'view/users_ldap.xml',
         'view/populate_wizard.xml',


users_ldap_populate/model/users_ldap.py
```diff
@@ -20,10 +20,16 @@
...
-from ldap.filter import filter_format
 from openerp.osv import orm
 import logging
 
+_logger = logging.getLogger(__name__)
+
+try:
+    from ldap.filter import filter_format
+except ImportError:
+    _logger.debug('Can not `from ldap.filter import filter_format`.')
```
sbidoul, line 31: `_logger.warning` instead of `debug`

pedrobaeza, line 31: Being discussed in https://github.com/OCA/account-financial-tools/pull/281/files#r37519322

moylop260:
:+1:

moylop260:
If fine to me
But if you add external depends is better:  https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#__openerp__py

pedrobaeza:
As this module depends of auth_ldap, and that module already has that external_dependencies defined, is not explicitly needed.

nhomar:
:+1:

nhomar:
@pedrobaeza the problem of expect the external_dependency only in the parent module, when you click on "install" the screen will retreive an error mentioning the dependency, not the module itself, IMHO it is useful here, even if technically it is not necesary, but BTW you have my +1 we have a rule to follow, and the rule do not mention explicitally how manage the depednency itself.

moylop260:
@pedrobaeza 
Ok, I got it
Thanks for explanation
confirmed :+1:

lmignon:
@pedrobaeza  once a module do an explicit import of an external dependency you must declare the dependency into the manifest. It's required to avoid that if the decency is removed from the dependent module the current one will be broken.

moylop260:
@lmignon 
You are right.
However, IMHO you should create a PR to add this case in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#__openerp__py
And later we will need to fix all repo's

IMHO By this case the main goal is done!

lmignon:
@moylop260  the rule is explicit.

> If your module use extras dependencies of python or binaries you should add to __openerp__py file the section external_dependencies.

``` python
from ldap.filter import filter_format
```

The import means that this module use an extra dependency.

pedrobaeza:
As stated before in others PRs "Better explicit than implicit", so I have added the external_dependencies in the manifest.

nhomar:
comments applied, then +1 valid, I merge.

PR #234, opened 2015-08-20
Remove another static folder

yvaucher:


yvaucher:
I close this as there is already https://github.com/OCA/server-tools/pull/232

PR #239, opened 2015-08-28
[ADD] allow to clean up menus

hbrunn:
why in 7.0 api? Because then it's simple to do the same for <= 7.0

### DIFF
--- a/database_cleanup/model/purge_menus.py
+++ b/database_cleanup/model/purge_menus.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2015 Therp BV (<http://therp.nl>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


pedrobaeza:
:+1:

database_cleanup/model/purge_menus.py
```diff
@@ -0,0 +1,81 @@
...
+        'menu_id': fields.many2one('ir.ui.menu', 'Menu entry'),
+    }
+
+    def purge(self, cr, uid, ids, context=None):
+        self.pool['ir.ui.menu'].unlink(
+            cr, uid,
+            [this.menu_id.id for this in self.browse(cr, uid, ids,
+                                                     context=context)],
+            context=context)
+        return self.write(cr, uid, ids, {'purged': True})
```
ajite: You forgot to pass the context in this write method.

guewen:
Once @ajite remark is  fixed, :+1:

database_cleanup/model/purge_menus.py
hbrunn: @ajite thanks, fixed

lmignon:
:+1:

sysadminmatmoz:
Just a remark.
When a PR changes or adds views, like:
-        'view/purge_menus.xml',

it should also change the module version (like 'version': '0.1', -> '0.1.1'), since an odoo restart is not enough in this case and the module has to be upgraded.

pedrobaeza:
You're right. I forget to check that in my review...

PR #110, opened 2015-01-09
Add new module: materialized sql view

petrus-v:
Hi,

I'm Pierre Verkest from Anybox, I'm pretty new in OCA and not sure about
the workflow to add new modules. Anyway this is a module we developed
on [bitbucket](https://bitbucket.org/anybox/materialized_sql_view).

I'm glad to contribute and dive into OCA tools and thanks @legalsylvain who
ask us to push this `materialized_sql_view` module here.

If you want to know more about this module you can read materialized_sql_view
module description in `__openerp__.py`.

I had to separate tests in an other module as I would tested the module with a
real model but wouldn't see it on production servers.

I already prepare a [branch running under v7](https://github.com/petrus-v/server-tools/tree/7.0-materialized_sql_view)
I guess if it's like odoo workflow I haven't to make a PR on both branches (7.0 and 8.0) ?

I was surprised some tested pass in Travis, like this one:
 https://travis-ci.org/petrus-v/server-tools/builds/46459751 based on this
commit: https://github.com/petrus-v/server-tools/commit/ab614eddd7f11dc81251080960f1969e6454d396
which should failed.

regards,

PR #245, opened 2015-09-01
Add: sql_view - create SQL views on the database

guewen:
This addon allows to create SQL views on the database.  It also features
a simple CSV export of the views to check their result.


sql_view/models/sql_view.py
```diff
@@ -0,0 +1,174 @@
...
+    def _complete_from_sql_name(self, cr, uid, sql_name, context=None):
+        return SQL_VIEW_PREFIX + (sql_name or '')
+
+    def _compute_complete_sql_name(self, cr, uid, ids, name, args,
+                                   context=None):
+        res = {}
+        for sql_view in self.browse(cr, uid, ids, context=context):
+            res[sql_view.id] = self._complete_from_sql_name(cr, uid,
+                                                            sql_view.sql_name,
+                                                            context=context)
```
sebalix, line 48: I am not a fan of this args' indentation, as it forces to re-indent all the lines if we change the length of the first one and make the reading harder. I always put the arguments on the second line if there is not enough space, like this:

``` python
            res[sql_view.id] = self._complete_from_sql_name(
                cr, uid, sql_view.sql_name, context=context)
```

But it is just a matter of taste.

sql_view/wizards/sql_view_csv_preview.py
```diff
@@ -0,0 +1,83 @@
...
+        form = self.browse(cr, uid, ids[0], context=context)
+        sql_view = self.pool['sql.view'].browse(cr, uid, sql_view_ids[0],
+                                                context=context)
+        query = self._query(cr, uid, form, sql_view, context=context)
+        cr.execute(query)
+        headers = [desc[0] for desc in cr.description]
+        records = cr.fetchall()
+        filedata = StringIO()
+        try:
+            writer = unicodecsv.writer(filedata, encoding='utf-8')
```
sebalix, line 69: Is it possible to use the `csv_internal_sep` server option as the delimiter to be consistent with other exports?

sebalix:
Tested, working fine.
Great module to maintain SQL views :+1:

sql_view/wizards/sql_view_csv_preview.py
guewen, line 69: from a quick grep, this option does not seem to be used?
I can even find in a test (openerp/tests/addons/test_impex/tests/test_import.py)

> ```
>  [[ # FIXME: hardcoded comma, import uses config.csv_internal_sep
>            # resolution: remove configurable csv_internal_sep
> ```

sebalix, line 69: No problem then :) I thought this option was used, thanks for the tip.

legalsylvain:
I gave a quick overview of this PR. It seems that as you said, there are some common functionnalities with #110.
@guewen : What do you think about that point, avoiding duplicated features ?

guewen:
@legalsylvain about functionalities, they have nothing that they share. This module handles views, when #110 handle materialized views. At first sight, I thought that both could be mixed but then I noticed that from a code point of view they have very few things they could share, so I don't see much benefits to mix both modules.

max3903:
:+1:

PR #249, opened 2015-09-06
[ADD] Added module password_change_restrict.

alejandrosantana:
This module allow only admin and users within Access Rights to change passwords, disallowing the rest.


moylop260:
Is posible add unitest?

dreispt:
The module name should begin with `auth_`.

password_change_restrict/models/res_users.py
```diff
@@ -0,0 +1,48 @@
...
+# License and authorship information in:
+# __openerp__.py file at the root folder of this module.
+
+from openerp import api, exceptions, models, SUPERUSER_ID, _
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    SELF_WRITEABLE_FIELDS = ['signature', 'action_id', 'company_id', 'email',
```
hbrunn, line 11: don't do this, remove the field in `_register_hook` - otherwise we loose all changes from other modules in the inheritance chain

password_change_restrict/models/res_users.py
```diff
@@ -0,0 +1,48 @@
...
+                             'name', 'image', 'image_medium', 'image_small',
+                             'lang', 'tz']
+
+    def _can_change_password(self):
+        group_ok = 'base.group_erp_manager'
+        user = self.env['res.users'].browse(self.env.uid)
```
hbrunn, line 17: use `self.env.user` instead of rebrowsing `uid`

password_change_restrict/models/res_users.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.multi
+    def write(self, values):
+        '''If user is trying to change password and allowed,
+        remove the key from the values dictionary to avoid the change
+        and even raise an exception, as this should be considered a
+        security attack.
+        '''
+        if 'password' in values and not self._can_change_password():
+            del values['password']
```
hbrunn, line 45: if you raise anyways, no need to change `values`

alejandrosantana, line 45: Agree. I will remove it.

password_change_restrict/models/res_users.py
alejandrosantana, line 17: @hbrunn I am having a weird issue with that. I explain:
Given a user that does not belong to `group_ok = 'base.group_erp_manager'`:
Using `user = self.env.user`, `user.has_group(group_ok)` returns True (should not!)
But using `user = self.env['res.users'].browse(self.env.uid)`, `user.has_group(group_ok)` returns False (correct)

Any idea why?

password_change_restrict/models/res_users.py
alejandrosantana, line 11: In `_register_hook` or `__init__`? (I see other examples expanding SELF_WRITEABLE_FIELDS in `__init__`)

alejandrosantana:
@moylop260 As soon as I fixed @hbrunn recommendations, I will try to do tests.
@dreispt Agreed, I am changing it.
Regarding the other module, I think there are two functionalities there that should be separated. I could do a backport of this one to v7 if needed (but may take a delay) once this one is OK.

password_change_restrict/models/res_users.py
hbrunn, line 17: the function is decorated with `@tools.ormcache(skiparg=2)` which seems weird to me (I'd have expected `skiparg=1`), so I'd guess this is a caching issue.

I personally use `self.env.ref('group.xmlid') in self.env.user.groups_id`

password_change_restrict/models/res_users.py
hbrunn, line 11: that shouldn't matter too much in this case. Generally, `_register_hook` is safer because your model is fully loaded (relevant if you need to look up stuff in the db, patch fields, etc)

PR #250, opened 2015-09-08
[FIX] use a nonserialized cursor for updating user groups

hbrunn:
Doing this in a serialized transaction can cause deadlocks, and we don't need serialization here

### DIFF
--- a/auth_dynamic_groups/model/res_users.py
+++ b/auth_dynamic_groups/model/res_users.py
@@ -19,7 +19,6 @@
 #
 ##############################################################################
 from openerp.models import Model
-from openerp.modules.registry import RegistryManager
 from openerp import SUPERUSER_ID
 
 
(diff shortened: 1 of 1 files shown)


auth_dynamic_groups/model/res_users.py
```diff
@@ -35,10 +35,9 @@ def _login(self, db, login, password):
         return uid
 
     def update_dynamic_groups(self, uid, db):
-        pool = RegistryManager.get(db)
-        cr = pool._db.cursor()
-        user = pool.get('res.users').browse(cr, SUPERUSER_ID, uid)
-        groups_obj = pool.get('res.groups')
+        cr = self.pool._db.cursor(serialized=True)
```
moylop260: Serialized=True to avoid serialized?

hbrunn: @moylop260 thanks so much, I messed up the only thing that was to mess up here...

pedrobaeza:
Are you sure you don't need to serialize? How user get correct groups then?

hbrunn:
Using a nonserialized cursor doesn't always help, as the ORM also wants to update the user's `write_{uid,date}`, which blocks again, we do it in SQL

hbrunn:
@pedrobaeza the worst thing that can happen is to have not all dynamic groups, which will be fixed with the next login

moylop260:
Is posible add a test case that reproduce the problem?

hbrunn:
difficult, because it's a timing issue with multiple users involved

moylop260:
:+1:

StefanRijnhart:
:+1:

PR #257, opened 2015-09-23
Module base_module_doc_rst migrated from version 6.1

sysadminmatmoz:
I use it as a tool for the relationship graph mostly, but it could be handy for someone else too.

Fresh PR since I messed up my previous git repo.

Added Pedro's change requests in README (todo) and the fact, that it performs it's actions only on actually installed modules in description.

The image gets presented only for the modules that are really currently installed (not just present as installable) in the same database where you run the creation button. The RST is the sphinx RST (as defined in description) - **the output file is a gzip archive** containing index.rst, module.rst and relations graph.

relation_graph

and sphinx output after make html:
sphinx1


sysadminmatmoz:
I don't get the coveralls stats; if I look at only my module, the coverage increases.

pedrobaeza:
You are adding code without tests, that's why the coverage decreases.

:+1: adding the known issues I commented in the other PR

sysadminmatmoz:
My CLA is there for quite some time! Even from before I became an OCA member.

StefanRijnhart:
Please rebase instead of merging the target branch. If you rebase now, you should get rid of the merge commits. It's some of the commits included in your merges that triggered the CLA warning.

sysadminmatmoz:
Done. Thanks.

sysadminmatmoz:
Finally all the checks are ok :smile:

base_module_doc_rst/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Generate Docs of Modules",
+    "version": "1.0",
```
StefanRijnhart: Please apply the new versioning convention and change to 8.0.1.0.0

base_module_doc_rst/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+    "category": "Tools",
+    "summary": "Modules Technical Guides in RST and Relationship Graphs",
+    "website": "https://odoo-community.org/",
+    "author": "OpenERP SA,Odoo Community Association (OCA)",
+    "contributors": [
```
StefanRijnhart, line 29: I don't think this is a valid key here. Contributors belong in the README.rst where you already put them.

base_module_doc_rst/wizard/tech_rst_guide.py
```diff
@@ -0,0 +1,523 @@
...
+            "",
+            """
+            .. tip:: This module is part of the OpenERP software,
+            the leading Open Source
+            """,
+            """  enterprise management system. If you want to discover OpenERP,
+            check our
+            """,
+            """  `screencasts <http://openerp.tv>`_ or download """,
+            """  `OpenERP <http://openerp.com>`_ directly.""",
```
StefanRijnhart: Maybe remove this old promotion bit.

StefanRijnhart:
Just some nags but generally good, thanks!

sysadminmatmoz:
Ok, I cleaned it further.
Regarding the contributors part: only in the server-tools there are 4 other modules with a openerp.py file set like that, but there are even a lot more of them in the other OCA repositories, and I took them as an example for a module, that is committed by someone that is not the original author.

StefanRijnhart:
Thanks! :+1:

PR #258, opened 2015-09-24
[UPDATE] Add hr_employee_gravatar

Endika:
Is the same for base_user_gravatar but in this case extends to employees.
Normally the res_user and hr_employee is the same person.
In the case you not have instaled HR addon only use base_user_gravatar
But in the case you have HR use this.

note: please don't kill me, gravatar not pay me to implement.


hbrunn:
I think this belongs in https://github.com/OCA/hr

hbrunn:
You could have also used https://github.com/OCA/hr/tree/8.0/hr_employee_data_from_work_address which synchronizes the data anyways, but this has some side effects you don't necessarily want

hr_employee_gravatar/models/res_users.py
```diff
@@ -0,0 +1,18 @@
...
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    @api.one
+    def get_gravatar_image(self):
+        super(ResUsers, self).get_gravatar_image()
+        hr_employee_list = self.env['hr.employee'].search(
+            [('user_id', '=', self.id)])
+        for hr_employee in hr_employee_list:
+            hr_employee.write({'image': self.image})
```
hbrunn: don't loop to write the same value, write on the whole recordset at once

Endika:
@hbrunn i tested hr_employee_data_from_work_address ok, this addon is unnecessary, i close the PR,

thx

Endika:
@hbrunn I tested again the addon hr_employee_data_from_work_address and not run correctly, the addon get user data from uid https://github.com/OCA/hr/blob/8.0/hr_employee_data_from_work_address/models/hr_employee.py#L74
I need get data from specific res.users

hbrunn:
@Endika yes, then it makes sense to have an own module for this. But still, please move your PR to the hr project

Endika:
Move this PR to https://github.com/OCA/hr/pull/138

PR #262, opened 2015-09-26
8.0 add auth track and prevent brut force

legalsylvain:
Add a module to limit brut force attack on authentication.
Define an ir.config_parameter that mentions max attempts.
Once this maximum reached, the remote user will be blocked. (by IP).

See [readme file](https://github.com/grap/server-tools/blob/8.0_ADD_auth_track_and_prevent_brut_force/auth_track_and_prevent_brut_force/README.rst) for the whole description.

Thanks for your review.

screenshot_attempts_list


pedrobaeza:
Very interesting module. I think the name can be shorten to _auth_brute_force_, because we don't need to put the full title on the name :wink:

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+==============================================
+Authentication Tracking and Prevent Brut Force
```
pedrobaeza: s/Brut Force/Brut-force attacks

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+==============================================
+
+This module register each request done by users trying to authenticate into
```
pedrobaeza: s/register/registers

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+Odoo. If the authentication fails, a counter is increased for the given remote
+IP. After 10 attempts (or other quantity, depending on an ir.config_parameter),
```
pedrobaeza: Put this in Configuration section

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+Odoo will ban the remote IP and ignore new requests.
+
+Admin user have the possibility to unblock a banned IP.
+
+This module realizes a call to a web API (http://ip-api.com) to try to have
+extra informations about remote IP.
+
+Note and Warning
+----------------
+The ID used to identify a remote request is the IP provided in the request
```
pedrobaeza: Put this on Known issue section

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+    :alt: License
+
+==============================================
+Authentication Tracking and Prevent Brut Force
+==============================================
+
+This module register each request done by users trying to authenticate into
+Odoo. If the authentication fails, a counter is increased for the given remote
+IP. After 10 attempts (or other quantity, depending on an ir.config_parameter),
+Odoo will ban the remote IP and ignore new requests.
```
pedrobaeza: What Odoo returns in this case? 500, 404?

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+(key 'REMOTE_ADDR').
+Depending of server and / or user network configuration, the idenfication
+of the user can be wrong, and mainly in the following cases:
+
+* if the Odoo server is behind an Apache / NGinx proxy without redirection,
+  all the request will be have the value '127.0.0.1' for the REMOTE_ADDR key;
+* If some users are behind the same Internet Service Provider, if a user is
+  banned, all the other users will be banned to;
+
+Logging
```
pedrobaeza: Do you think a logging is needed when having an specific table for login attempts?

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+==============================================
+
+This module register each request done by users trying to authenticate into
+Odoo. If the authentication fails, a counter is increased for the given remote
+IP. After 10 attempts (or other quantity, depending on an ir.config_parameter),
+Odoo will ban the remote IP and ignore new requests.
+
+Admin user have the possibility to unblock a banned IP.
+
+This module realizes a call to a web API (http://ip-api.com) to try to have
```
pedrobaeza: Does that call delay the login? It should be then put on Known issues, or prepare an asynchronous call for that

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+==============================================
+Authentication Tracking and Prevent Brut Force
+==============================================
+
+This module register each request done by users trying to authenticate into
+Odoo. If the authentication fails, a counter is increased for the given remote
+IP. After 10 attempts (or other quantity, depending on an ir.config_parameter),
+Odoo will ban the remote IP and ignore new requests.
+
+Admin user have the possibility to unblock a banned IP.
```
pedrobaeza: This can go in Usage section

auth_track_and_prevent_brut_force/README.rst
```diff
@@ -0,0 +1,99 @@
...
+.. image:: /auth_track_and_prevent_brut_force/static/description/screenshot_attempts_list.png
+
+**Detail of a banned IP**
+
+.. image:: /auth_track_and_prevent_brut_force/static/description/screenshot_custom_ban.png
+
+
+Usage
+=====
+
```
pedrobaeza: You can complete usage section indicating which menu is used for seeing logs and so on

pedrobaeza:
Good icon! Do you have the SVG version? Can you remove blank background from it and let it transparent?

pedrobaeza:
Please put also icon reference on credits (ensuring it can be used - CC license?)

pedrobaeza:
Sorry for bothering you again. In last commit you have changed references to `auth_brut_force` instead of `auth_brute_force`

auth_track_and_prevent_brut_force/README.rst
legalsylvain: The call is done when the user is banned. So it doesn't delay the login...

legalsylvain:
hi @pedrobaeza ! Thanks for your review !
- About what odoo returns, I updated the description. This point is pretty important, the module don't change the behaviour. Odoo will simply indicates that the combination is bad (even if the attacker finally find the good combination of login / password)
- About logging, log can be used by a system that parse logs, or by sentry system, when table will not. (or not so easy). But I can understand that it is redundant.
- About the image, I have to fix it;

I changed the rest according to your other comments.

auth_brute_force/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Authentification - Brute-force Attack',
+    'version': '8.0.1.0.0',
+    'category': 'base',
+    'summary': "Authentication Tracking and Prevent Brute-force Attack",
```
elicoidal: Suggestion:
Tracks Authentication Attempts and Prevents Brute-force Attacks
or
Authentication Tracking and Brute-force Attacks Prevention
=> 2 verbs or 2 nouns. I prefer the first one.

pedrobaeza:
One minor remark:

```
access_res_authentication_attempt,access_res_authentication_attempt,model_res_authentication_attempt,,1,0,0,0 
```

The rest seems :+1: (code review)

legalsylvain:
Thanks for your review, @pedrobaeza. 
I added security file, giving read access to all users, and write / Delete access to "config / settting" members. (admin, by default).
Kind regards.

max3903:
:+1:

rafaelbn:
Hi @legalsylvain , I would like to ash about this module after reading README. Does it notify to any user in Odoo to take care about? (notify -> email). Thanks!

auth_brute_force/README.rst
```diff
@@ -0,0 +1,114 @@
...
+---------------------
+The ID used to identify a remote request is the IP provided in the request
+(key 'REMOTE_ADDR').
+Depending of server and / or user network configuration, the idenfication
+of the user can be wrong, and mainly in the following cases:
+
+* if the Odoo server is behind an Apache / NGinx proxy without redirection,
+  all the request will be have the value '127.0.0.1' for the REMOTE_ADDR key;
+* If some users are behind the same Internet Service Provider, if a user is
+  banned, all the other users will be banned too;
```
dreispt, line 31: But this is expected to be the usual case, isn't it?
I'm a little out of step here, be shouldn't we also look into `X-Forwarded-For`?

auth_brute_force/security/ir_model_access.yml
```diff
@@ -0,0 +1,28 @@
+# -*- encoding: utf-8 -*-
+- !record {model: ir.model.access, id: access_res_authentication_attempt_all}:
+    model_id: model_res_authentication_attempt
+    name: Authentication Attempt All Users
+    perm_read: true
+
```
alexis-via, line 6: I think it's a kind of privacy flaw to allow all users to read res.authentication.attempt. And it's not needed to give that right to all users.

legalsylvain: You're probably right. (the same for model_res_banned_remote I guess)
feel free to propose a PR, I'll review it quickly. (you can too replace by xml, as yml is now deprecated)

PR #265, opened 2015-09-29
[8.0] port server_monitoring

gurneyalex:


server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,63 @@
...
+
+{
+    'name': 'Server Monitoring',
+    'version': '0.1',
+    'category': 'Tools',
+    'depends': ['base',
+                ],
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'description': """
```
RoelAdriaans: Description can be removed because you have a README.rst file

server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,63 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Server Monitoring',
+    'version': '0.1',
+    'category': 'Tools',
+    'depends': ['base',
```
RoelAdriaans: No longer needed to have a depends on base in v8 modules

gurneyalex: fixed

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+Just install the addon in your instance. You will be able to access the logs
+under Reporting -> Server Monitoring.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/8.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
```
tafaRU: Please remove three above lines, see https://github.com/OCA/maintainer-tools/pull/140 as reference.

server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Server Monitoring',
+    'version': '8.0.0.1',
```
tafaRU: Please update version according to https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

lmignon: Is-it the right scheme for version umber? IMO it should be 8.0.1.0.0 https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

StefanRijnhart:
Tested succesfully. Great module! Some process info entries have no model set, with common methods such as fields_view_get or read. Is this expected?

StefanRijnhart:
Now, if I log out of the database with this module installed and log on to another database on the same instance I get 

```
server-tools/server_monitoring/models/server_monitor_process.py", line 109, in _call_kw
    monitor_obj.log_measure(request.cr, request.uid, model, method,
AttributeError: 'NoneType' object has no attribute 'log_measure'
```

and I can not use this database.

gurneyalex:
@StefanRijnhart probably something I missed when adding the json rpc support. I'll check. 

re: multiple db, I admit I never test it. I'll check it.

gurneyalex:
@guewen you mentionned you had a solution for the monkey patching to only happen when the module is actually installed. Can you tell me more about this?

sebalix:
@gurneyalex Maybe the `post_load` key in the manifest: https://github.com/odoo/odoo/blob/8.0/openerp/modules/module.py#L322
OCA's template module mentions it regarding monkey patching: https://github.com/OCA/maintainer-tools/blob/master/template/module/hooks.py#L49

guewen:
I didn't know `post_load`, it could be a clean solution!

The one I had in head was the `_register_hook` (https://github.com/OCA/server-tools/blob/b2251a7ed3a12533bcf304e2a456fc3be000024c/base_import_match/models/base_import.py#L70). But it needs to be done on a model (it would be an arbitrary model in your case) so less readable/understandable than the `post_load` solution (if `post_load` works as I expect :).

dreispt:
This PR is about to make it's first birthday! 😸

gurneyalex:
@dreispt it's much older than this... it had started as a MP on launchpad on v7...

server_monitoring/__openerp__.py
```diff
@@ -0,0 +1,45 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Server Monitoring',
+    'version': '8.0.1.0.0',
+    'category': 'Tools',
+    'depends': [],
```
hbrunn: no `base`?

hbrunn (approved):
only small things

hbrunn:
one word of caution about `post_load`: This is run once server wide, so your monkey patches can end up being installed because the module is installed in database A, but run for B and then it will break. Here, I think we need to be quite sure that the module is actually installed, so you should follow the pattern of using `_register_hook` while checking if you already did the monkey patching (`_register_hook` can be called more often than once). Auditlog does that extensively: https://github.com/OCA/server-tools/blob/8.0/auditlog/models/rule.py#L104
I'm curious if you ever measured the performance penalty of the module? And if this is non-negligible, would it be useful to be able to turn that off for a running server?
When this is merged, I'll want to marry it to dead_mans_switch so that we can pull all the data into our own odoo's dashboard if we want.

server_monitoring/__openerp__.py
RoelAdriaans: Base is no longer needed as a dependency in the newest odoo.

server_monitoring/models/server_monitor_process.py
```diff
@@ -116,6 +118,14 @@ def __init__(self, pool, cr):
 
     _name = 'server.monitor.process'
 
+    def _register_hook(self, cr):
+        global _monkey_patched
+        if _monkey_patched:
+            return
+        __monkey_patched = True
```
hbrunn: the extra underscore messes up stuff I guess

gurneyalex:
@hbrunn I've updated the module following your suggestion (and using a naive lock). There is some additional protection on the monkey patched version to do nothing  when the model is not in the registry. 

Regarding the perf. hit, I admit I have no measures, but there must be some since each RPC call triggers a write. This can be mitigated by:
- adding a configuration option to globally not log RPC access (or maybe later we could want to fine tune this on a per-model basis)
- I'm also thinking of a cheap optimization in which the logs would be kept in memory for some time, and only flushed after a number of RPC calls. This could mean loosing some logs when the instance is shut down but I can probably live with it.

hbrunn:
@gurneyalex thanks for your points, you just need to loose one underscore I think.

About the performance stuff, sounds good. Parameters are read through a registry wide cache anyways, so this won't impact performance itself when checking whether to log or lot. But let's not block on this and take it as a future extension. I'm especially looking forward to having a convenient way of comparing index usage across different DBs and customer profiles.

guewen:
> adding a configuration option to globally not log RPC access (or maybe later we could want to fine tune this on a per-model basis)

Fine-tuning might be done by overrides of the `log_measures` methods. A global switch might be useful, but not strictly needed to merge this pull request.

> I'm also thinking of a cheap optimization in which the logs would be kept in memory for some time, and only flushed after a number of RPC calls. This could mean loosing some logs when the instance is shut down but I can probably live with it.

I also losing some logs is not an issue, we usually want to measure the tendencies. Not needed too for a merge.

gurneyalex:
@hbrunn @guewen I just added a configuration setting for enabling RPC calls monitoring.

hbrunn:
I still approve ;-) But I removed the needs fixing part as the error on travis seems quite unrelated

moylop260:
FYI the [lint build](https://travis-ci.org/OCA/server-tools/jobs/161299646#L441-L442) show us:

``` bash
server_monitoring/models/server_monitor_process.py:91:19: F821 undefined name 'self'
server_monitoring/models/field_types.py:29: [C8104(class-camelcase), bigint] Use `CamelCase` "Bigint" in class name "bigint"...
```

hbrunn:
@moylop260 thanks, those need to be fixed of course. I meant those: https://travis-ci.org/OCA/server-tools/jobs/161299648#L1730

hbrunn (approved):
whatever is wrong with travis doesn't seem to be caused by this module

moylop260 (approved):

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
+server monitoring
+=================
+
+This module allows in-database logging of some statistics in order to monitor
+the health of an openerp instance.
```
elicoidal: s/openerp/Odoo

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Database indicators are logged (number of rows, table size, number of reads,
+number of updates...), with a cron running each week by default. This cron
+needs to be activated after the module is installed.
+
+Some process indicators are logged (cpu time, memory) together with information
```
elicoidal: s/cpu/CPU

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Two crons are provided to cleanup old logs from the database.
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* Settings -> Scheduled Actions and tune the cron tasks
```
elicoidal: What about this?
#. `Settings -> Scheduled Actions` and tune the cron tasks
#. `Settings -> Configuration -> General Settings`
#. Enable Monitor RPC Calls if you want information about RPC calls 
being made (causes some performance overhead).

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+   :target: https://runbot.odoo-community.org/runbot/149/8.0
+
+
+Known issues / Roadmap
+======================
+
+* Not tested under other platforms than Linux
+
+TODO / WISH list:
+* visualization
```
elicoidal: Use capitals for the first letter

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+* enhance group by for process logs. 
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/server-tools/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/server-tools/issues/new?body=module:%20server_monitoring%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
elicoidal: Please remove this line

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

server_monitoring/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: please use short headers

server_monitoring/data/ir_cron.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
elicoidal, line 2: You can add a copyright notice in the XML files as well

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,208 @@
...
+    lines_read_total = fields.Bigint('Tot lines read', readonly=True)
+    num_insert = fields.Bigint('Inserts', readonly=True)
+    num_update = fields.Bigint('Updates', readonly=True)
+    num_delete = fields.Bigint('Deletes', readonly=True)
+    measure_id = fields.Many2one('server.monitor.database',
+                                 string='Measure',
+                                 ondelete='cascade',
+                                 readonly=True)
+    timestamp = fields.Datetime(related='measure_id.name',
+                                string='timestamp',
```
elicoidal: Timestamp

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,208 @@
...
+    _name = 'server.monitor.model.table.activity.read'
+    name = fields.Char('Table name')
+    disk_reads = fields.Bigint('Disk reads (heap blocks)', readonly=True)
+    cache_reads = fields.Bigint('Cache reads', readonly=True)
+    total_reads = fields.Bigint('Total reads', readonly=True)
+    measure_id = fields.Many2one('server.monitor.database',
+                                 string='Measure',
+                                 ondelete='cascade', readonly=True)
+    timestamp = fields.Datetime(related='measure_id.name',
+                                string='timestamp',
```
elicoidal: Timestamp (several)

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,208 @@
...
+    _order = 'timestamp DESC, total_reads DESC'
+
+
+class ModelTableActivityUpdate(models.Model):
+    _name = 'server.monitor.model.table.activity.update'
+
+    name = fields.Char('Table name', readonly=True)
+    seq_scan = fields.Bigint('Seq scans', readonly=True)
+    idx_scan = fields.Bigint('Idx scans', readonly=True)
+    lines_read_total = fields.Bigint('Tot lines read', readonly=True)
```
elicoidal, line 70: Lines read
(to be consistent with the other fields)

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,263 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+"""
+Monitor openerp instance.
```
elicoidal: s/openerp/Odoo (Several)

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,263 @@
...
+        for obj in gc.get_objects():
+            if isinstance(obj, BLACKLIST):
+                continue
+            try:
+                cls = obj.__class__
+            except:
+                if isinstance(obj, types.ClassType):
+                    cls = types.ClassType
+                else:
+                    _logger.warning('unknown object type for %r (%s)',
```
elicoidal: Unkown

server_monitoring/tests/test_monitor_database.yml
```diff
@@ -0,0 +1,64 @@
+-
```
elicoidal, line 1: yml tests being deprecated, should we keep this file?

elicoidal (changes requested):
Nothing blocking except maybe the headers
Thanks for the contribution!

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,263 @@
...
+
+def _monkey_patch_object_proxy_execute():
+    orig_execute_cr = openerp.service.model.execute_cr
+
+    def execute_cr(cr, uid, obj, method, *args, **kw):
+        result = orig_execute_cr(cr, uid, obj, method, *args, **kw)
+        pool = pooler.get_pool(cr.dbname)
+        monitor_obj = pool.get('server.monitor.process')
+        monitor = pool['ir.config_parameter'].get_param(
+            'server_monitoring.monitor_rpc_calls', default=False
```
mtelahun, line 78: I ran into a problem here. You are missing cr and uid arguments to get_param()

mtelahun (changes requested):

server_monitoring/models/server_monitor_process.py
gurneyalex: fixed, thanks.

gurneyalex:
most of the changes requested by @elicoidal done (I just kept the yaml tests and will rewrite them when porting to odoo 9 or 10, I promise). I think this can be merged :smile_cat:

gurneyalex:
regarding the pylint error on travis, cf. https://github.com/OCA/maintainer-quality-tools/issues/374

regarding the build failure on OCB, I'm baffled. Could there be a weird interaction between this addon and database cleanup? @hbrunn I'd appreciate if you could share some insight...

astirpe:
I've tried this module on a database that is a copy of a db used in production.
All seems working well. I like this module.
:+1:

server_monitoring/i18n/server_monitoring.pot
```diff
@@ -0,0 +1,445 @@
+# Translation of OpenERP Server.
```
yajo, line 1: Delete pot file.

server_monitoring/models/field_types.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class OsvBigint(osv.fields.integer):
+    _type = 'int8'
+
+
+fields.Bigint = Bigint
+osv.fields.bigint = OsvBigint
+
+models.FIELDS_TO_PGTYPES[OsvBigint] = 'int8'
```
yajo, line 20: Please don't do this. You are polluting the global namespace, no matter if this addon is installed or not, and in all databases in any case.

Just declare the classes and import them from here if you need them, not from `fields`.

And BTW, I'd remove the `osv` version. We want no more old api wherever possible.

server_monitoring/models/res_config.py
```diff
@@ -0,0 +1,37 @@
...
+        else:
+            param = False
+        return {'monitor_rpc_calls': int(param)}
+
+    @api.multi
+    def set_default_monitor_rpc_calls(self):
+        for record in self:
+            value = int(record.monitor_rpc_calls) or ''
+            self.env['ir.config_parameter'].set_param(
+                'server_monitoring.monitor_rpc_calls', value
```
yajo, line 36: You can do this in a data file, easier.

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+_logger = logging.getLogger(__name__)
+
+
+class ModelRowCount(models.Model):
+    _name = 'server.monitor.model.row.count'
+
+    name = fields.Char('Table name', readonly=True)
+    count = fields.Bigint('Row count', readonly=True)
```
yajo, line 20: As said above, this should be in another namespace. OTOH, I think you could simply use `fields.Integer('Row count', readonly=True, type="bigint")` and should simply work. (Not sure eh)

The same applies for all these declarations, so I won't repeat it further down below.

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,193 @@
...
+        query = (
+            "SELECT nspname || '.' || relname AS name, "
+            "        pg_size_pretty(pg_total_relation_size(C.oid)) AS hsize, "
+            "        pg_total_relation_size(C.oid) AS size "
+            "FROM pg_class C LEFT JOIN pg_namespace N "
+            "              ON (N.oid = C.relnamespace) "
+            "WHERE nspname NOT IN ('pg_catalog', 'information_schema') "
+            "  AND C.relkind <> 'i' "
+            "  AND nspname !~ '^pg_toast' "
+            "ORDER BY pg_total_relation_size(C.oid) DESC"
```
yajo, line 112: Is Odoo user supposed to have special permissions to do these queries?

If so, please specify that in README and forbid installation with a check in a `pre_install_hook` with a human-readable `ValidationError` raised if there are no perms.

BTW you could return a [`Query`](https://github.com/odoo/odoo/blob/8.0/openerp/osv/query.py#L30) in a separate method for better inheritance. This applies to all literal queries found in addon too.

server_monitoring/models/server_monitor_database.py
```diff
@@ -0,0 +1,193 @@
...
+        'measure_id',
+        'Model table update activity',
+        default=_model_table_activity_update,
+        readonly=True)
+
+    _order = 'name DESC'
+
+    @api.model
+    def log_measure(self):
+        self.create({})
```
yajo, line 182: Do you need this? Can't you use `create` directly?

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+
+    _name = 'server.monitor.process'
+
+    def _register_hook(self, cr):
+        global _monkey_patched
+        if _monkey_patched:
+            return
+        _monkey_patched = True
+        _monkey_patch_controller_call_kw()
+        _monkey_patch_object_proxy_execute()
```
yajo, line 119: You should use [`_patch_method` (and possibly `_revert_method`)](https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L5211-L5247) instead, as they are the standards for monkey patching.

These methods should be found/returned inside the class to allow inheritance.

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+                                    'jsonrpc call',
+                                    False, False, context={})
+        return result
+
+    openerp.addons.web.controllers.main.DataSet._call_kw = _call_kw
+
+
+class ServerMonitorProcess(models.Model):
+    def __init__(self, pool, cr):
+        super(ServerMonitorProcess, self).__init__(pool, cr)
```
yajo, line 109: This is useless.

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+                         default=_default_pid,
+                         readonly=True,
+                         group_operator='count')
+    thread = fields.Char('Thread ID',
+                         default=_default_thread,
+                         readonly=True)
+    cpu_time = fields.Float(
+        'CPU time',
+        readonly=True,
+        default=_default_cpu_time,
```
yajo, line 196: Normally [you should define a lambda calling that method, to allow inheritance, as guidelines explain](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields). The same applies in many other places.

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+    sessionid = fields.Char('Session ID', readonly=True)
+    info = fields.Char('Information')
+    class_count_ids = fields.One2many(
+        'server.monitor.class.instance.count',
+        'measure_id',
+        'Class counts',
+        default=_class_count,
+        readonly=True)
+
+    _order = 'name DESC'
```
yajo, line 223: Put this one at the beginning of class declaration, next to `_name`, `_description` and those kind of vars.

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+            request.cr, request.uid,
+            'server_monitoring.monitor_rpc_calls', default=False
+        )
+        if monitor_obj is not None and bool(monitor):
+            monitor_obj.log_measure(request.cr, request.uid, model, method,
+                                    'jsonrpc call',
+                                    False, False, context={})
+        return result
+
+    openerp.addons.web.controllers.main.DataSet._call_kw = _call_kw
```
yajo, line 104: You don't need to monkey patch a controller. Just declare it as usually.

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+        'server.monitor.class.instance.count',
+        'measure_id',
+        'Class counts',
+        default=_class_count,
+        readonly=True)
+
+    _order = 'name DESC'
+
+    @api.model
+    def log_measure(self, model_name, method_name, info,
```
yajo, line 226: Are you sure you want this method as public, reachable by XMLRPC?

server_monitoring/models/server_monitor_process.py
```diff
@@ -0,0 +1,248 @@
...
+        model = model_obj.search([('name', '=', model_name)], limit=1)
+        values = {'model': model.id,
+                  'method': method_name,
+                  'info': info,
+                  }
+        record = self_ctx.sudo().create(values)
+        return record
+
+    @api.model
+    def cleanup(self, age):
```
yajo, line 241: Are you sure you want this method as public, reachable by XMLRPC?

server_monitoring/tests/test_monitor_database.yml
yajo: Definitely not, although not a blocker for me until they completely disappear.

server_monitoring/views/server_monitor_database.xml
```diff
@@ -0,0 +1,97 @@
...
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+<openerp>
+  <data>
+
+    <record model="ir.ui.view" id="server_monitor_database_tree_view">
+      <field name="name">Server Monitor Database form view</field>
+      <field name="model">server.monitor.database</field>
+      <field name="arch" type="xml">
+        <tree string="Server Monitor" version="7.0">
```
yajo, line 14: drop version, here and in every form view.

server_monitoring/views/server_monitor_database.xml
```diff
@@ -0,0 +1,97 @@
...
+                      <field name='num_insert'/>
+                      <field name='num_update'/>
+                      <field name='num_delete'/>
+                    </tree>
+                  </field>
+                </group>
+              </page>
+            </notebook>
+            <group>
+              <field name='info' string="Information"/>
```
yajo, line 77: String should go into the field itself, right? Here and in other places...

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
+server monitoring
```
yajo, line 6: Fix case please.

server_monitoring/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+Known issues / Roadmap
+======================
+
+* Not tested under other platforms than Linux
+
+TODO / WISH list:
+* Visualization
+* Export
```
yajo, line 50: Just curious, are we talking about Prometheus?

yajo (changes requested):
Good idea, indeed, but I wouldn't merge it with current design, honestly. Some changes needed.

server_monitoring/models/field_types.py
gurneyalex: The osv version is required in v8 and probably in v9 too, afair.

server_monitoring/README.rst
gurneyalex: cheap lookalike, maybe yes.

gurneyalex:
can't work on this. Closing the PR.

PR #266, opened 2015-09-29
[FIX] disable rollbacks in matching code

hbrunn:
OpenERP 6.1 does a rollback when a field validation fails (https://github.com/odoo/odoo/blob/6.1/openerp/osv/orm.py#L1473). If this happens during matching in a cronjob, things can get ugly because the rollback also rolls back the cronjob's lock, possibly leading to parallel execution of several fetchmail threads, resulting in duplicate mails. We don't want that.


NL66278:
:+1:  Nice trick to set cr.rollback to lamda: None. And later restore. LGTM.

pedrobaeza:
:+1:

moylop260:
Travis fails unrelated

Issue #268, opened 2015-10-01
Error, when human resources manager edit employer


Endika:
module: auditlog
version: 8.0

**Steps to reproduce**
- Created new rule: for hr.employee models and checked write, unlink and create.
- Administrator user subscribe in this rules.
- Other user with Manager in human resources update any employeer and save. And show the error.

**Current behavior**

This is log:

Traceback (most recent call last):
  File "/opt/odoo/odoo/openerp/http.py", line 537, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/openerp/http.py", line 574, in dispatch
    result = self._call_function(*_self.params)
  File "/opt/odoo/odoo/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/opt/odoo/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/opt/odoo/odoo/openerp/http.py", line 307, in checked_call
    return self.endpoint(_a, *_kw)
  File "/opt/odoo/odoo/openerp/http.py", line 803, in __call__
    return self.method(_args, *_kw)
  File "/opt/odoo/odoo/openerp/http.py", line 403, in response_wrap
    response = f(_args, *_kw)
  File "/opt/odoo/odoo/addons/web/controllers/main.py", line 948, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/odoo/addons/web/controllers/main.py", line 940, in _call_kw
    return checked_call(request.db, _args, *_kwargs)
  File "/opt/odoo/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/opt/odoo/odoo/addons/web/controllers/main.py", line 939, in checked_call
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/opt/odoo/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/opt/odoo/odoo/openerp/api.py", line 363, in old_api
    result = method(recs, _args, *_kwargs)
  **File "/opt/odoo/oca/auditlog/models/rule.py", line 275, in write
    (d['id'], d) for d in self.sudo().read(list(self._fields)))**
  File "/opt/odoo/odoo/openerp/api.py", line 239, in wrapper
[...]

PR #269, opened 2015-10-05
[FIX] Issue: https://github.com/OCA/server-tools/issues/268

Endika:
Issue: https://github.com/OCA/server-tools/issues/268
I create conflict field list, to delete before check with auditlog

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -213,9 +213,17 @@ class auditlog_rule(models.Model):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env['auditlog.rule']
             new_record = create.origin(self, vals, **kwargs)
+            conflict_fields = {}
+            conflict_fields_list = ['total_wage']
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Before implementing blacklists of fields, it would be interesting to see why the value of this function field raises an exception when cast to a float. Could you trace that some more? Ideally, that would give you an idea how to cast such a value so that you could still reflect it in the log.

Or is it maybe the case that such function fields should not trigger a log in this case? What is the rule type?

StefanRijnhart:
Did you look at the method that supplies the values for 'total_wage' in hr_payroll? This did indeed not return a float value, but a dictionary. Fixed two weeks ago in Odoo: https://github.com/odoo/odoo/commit/f8b56e4f40b49436abbda006a0e77adfbe2ce72a. Can you try again after pulling your Odoo branch?

Endika:
@StefanRijnhart thanks, with update last Fixed, works

PR #273, opened 2015-10-19
[8.0] Add base_model_metaclass_mixin

houzefa-abba:
This module helps providing methods that override base Odoo model ones.
It is implemented via a nifty meta-class.
Check out its README file for use examples.


houzefa-abba:
Hey,

This is the beginning of work done at the PyconFR 2015 to merge some XCG Odoo addons into OCA.

We are going to start with analytic management; that starts with this base meta-class module, which our other addons make use of.

base_model_metaclass_mixin/README.rst
```diff
@@ -0,0 +1,26 @@
+### README ###
```
hbrunn, line 1: use OCA's readme template here

base_model_metaclass_mixin/README.rst
```diff
@@ -0,0 +1,26 @@
...
+
+
+###  Oemetaslfor Odoo/OpenERP ###
+
+
+This module handles methods and MetaClass to modify OpenERP base methods.
+For Exemple, we override copy method to restrict the copy in Odoo/OpenERP.
```
hbrunn, line 8: exAmple

base_model_metaclass_mixin/__init__.py
```diff
@@ -0,0 +1,33 @@
+from openerp.osv import orm
+from openerp.osv import osv
+from openerp.tools.translate import _
+
+__all__ = ['BaseModelMetaclassMixin']
+
+
+def get_overrides():
+    overrides = {}
```
hbrunn, line 9: this will be global for all databases running on the server. Shouldn't this be per database (as in: Set up the data structure in `_register_hook` and keep it per registry)?

hbrunn (changes requested):

PR #274, opened 2015-10-19
[8.0] Add auth saml

houzefa-abba:
This addon lets users log into Odoo via an SAML2 provider.
Requires lasso http://lasso.entrouvert.org.


houzefa-abba:
Hey,

This is part of an effort during the Pycon FR 2015 to submit some XCG addons to the OCA.

gurneyalex:
I added a whitelist request for python-lasso (https://github.com/travis-ci/apt-package-whitelist/issues/1488). 

In the meantime, can you add a sudo: required in the .travis.yml config file?

gurneyalex:
runbot buld fixed

houzefa-abba:
Thanks  @gurneyalex - the temporary "sudo: required" has been added.

houzefa-abba:
The various checks now pass, except for test runs, which also require lasso.
It looks like lasso is installed fine by the runbot check but not during travis test checks.
Any idea? Thanks

gurneyalex:
On 20/10/2015 12:03, Zouzou wrote:

> The various checks now pass, except for test runs, which also require lasso.
> It looks like lasso is installed fine by the runbot check but not during
> travis test checks.
> Any idea? Thanks
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/server-tools/pull/274#issuecomment-149505415.

python-lasso was whitelisted on travis, so the sudo: required can
probably be dropped.

## 

Alexandre Fayolle
Chef de Projet
Tel : +33 4 58 48 20 30

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac Cedex
http://www.camptocamp.com

houzefa-abba:
Hey @gurneyalex,
Done the revert!
Running with python-lasso seems to be working fine; tests are now failing because of logging issues (perhaps installing auth_saml changes some login rules, though I'm a bit surprised - to be looked into...).

auth_saml/views/auth_saml.xml
```diff
@@ -0,0 +1,70 @@
...
+        <record model="ir.ui.view" id="view_saml_provider_form">
+            <field name="name">auth.saml.provider.form</field>
+            <field name="model">auth.saml.provider</field>
+            <field name="arch" type="xml">
+                <form string="arch" version="7.0">
+                    <sheet>
+                        <group>
+                            <field name="name" />
+                            <field name="enabled" />
+                            <field name="body" />
```
petrus-v, line 45: Hi,

Should we also allow to set css_class attribute from the config view?

auth_saml/models/res_users.py
```diff
@@ -0,0 +1,219 @@
...
+                raise
+
+    @api.multi
+    def write(self, vals):
+        """Override to clear out the user's password when setting an SAML user
+        ID (as they can't cohabit).
+        """
+
+        if vals and vals.get('saml_uid'):
+            if not self._allow_saml_and_password():
```
petrus-v: Should we avoid to unset password when admin user wants to change its password?

I don't remember in case of `auth_crypt` module is installed, when the password is crypted when changing the value from the UI, we may want to del `password_crypt` as well?

auth_saml/models/res_users.py
```diff
@@ -0,0 +1,219 @@
...
+    # This method is using the old v7 API because it is called BEFORE the login
+    def check_credentials(self, cr, uid, token):
+        """token can be a password if the user has used the normal form...
+        but we are more interested in the case when they are tokens
+        and the interesting code is inside the except clause
+        """
+        token_osv = self.pool.get('auth_saml.token')
+
+        try:
+            baseuser.check_credentials(
```
petrus-v: here you bypass the auth_crypt module, so we can't connect if `password_crypt` is set but not password. So admin user can only connect to odoo with a fresh `password` set from sql or set saml_uid from sql or xml file.

I've done a test using super(..).credentials(...) while starting to migrate on odoo9, I haven't get any issues, what was wrong with super?

auth_saml/models/res_users.py
```diff
@@ -0,0 +1,219 @@
...
+                token
+            )
+
+        except (
+            openerp.exceptions.AccessDenied,
+            passlib.exc.PasswordSizeError,
+        ):
+            # since normal auth did not succeed we now try to find if the user
+            # has an active token attached to his uid
+            res = token_osv.search(
```
petrus-v: While migrate to odoo9 I had a try using @api.model decorator

and this become something like:

```
token_osv.sudo().search(
    [
        ('user_id', '=', self.env.uid),
        ('saml_access_token', '=', token),
    ]
)
```

it looks working!

auth_saml/models/res_users.py
```diff
@@ -0,0 +1,219 @@
...
+    saml_provider_id = fields.Many2one(
+        'auth.saml.provider',
+        string='SAML Provider',
+    )
+    saml_uid = fields.Char(
+        'SAML User ID',
+        help="SAML Provider user_id",
+    )
+
+    @api.one
```
petrus-v, line 35: I think @api.one is deprecated

You should use @api.multi them `self.ensure_one()`

auth_saml/models/auth_saml.py
```diff
@@ -0,0 +1,62 @@
...
+        server.addProviderFromBuffer(
+            lasso.PROVIDER_ROLE_IDP,
+            self.idp_metadata
+        )
+        return lasso.Login(server)
+
+    @api.multi
+    def _get_auth_request(self, state):
+        """build an authentication request and give it back to our client
+        WARNING: this method cannot be used for multiple ids
```
petrus-v: Should we remove the warning comment and add `self.ensure_one()` in the code?

auth_saml/models/base_settings.py
```diff
@@ -0,0 +1,86 @@
...
+    def allow_saml_and_password(self):
+        """Read the allow_saml_uid_and_internal_password setting.
+        Use the admin account to bypass security restrictions.
+        """
+
+        config_obj = self.env['ir.config_parameter']
+        config_objs = config_obj.sudo().search(
+            [('key', '=', _SAML_UID_AND_PASS_SETTING)],
+            limit=1,
+        )
```
petrus-v, line 40: You could use `self.env['ir.config_parameter'].sudo().get_param(_SAML_UID_AND_PASS_SETTING, default=False)` instead?

houzefa-abba:
Hey @petrus-v, thanks for the review. Most changes you have suggested, and some more to get this to properly work on Odoo 8 & 9, have been applied.

Remaining travis failures (I am stuck on both):
- [`pylint-odoo` fails](https://travis-ci.org/OCA/server-tools/jobs/154469187) because it does not understand some RST directives used in the sphinx doc that has been added. Removing that doc would be a pity; is there a way to tell `pylint-odoo` to ignore some files? -> https://github.com/OCA/pylint-odoo/issues/38
- [Tests fail](https://travis-ci.org/OCA/server-tools/jobs/154469189) because the `database_cleanup` addon attempts to remove a table added by `auth_saml` and fails?

This seems to be getting quite close to success; thanks for any help on these 2 points!

gurneyalex:
@hbrunn it looks like there is still some issues related to `database_cleanup`. Your input on this is welcome.

yasii1991:
Hi,

I have been struggling with the working of this module. I am trying to use the shibboleth idp, I have entered the metadata from the idp and also have the sp metadata configured. Now when I click on the provider login link, I get an error from lasso as follows

  File "/usr/lib/python2.7/dist-packages/lasso.py", line 6409, in addProviderFromBuffer
    Error.raise_on_rc(rc)
  File "/usr/lib/python2.7/dist-packages/lasso.py", line 62, in raise_on_rc
    raise exception
ServerAddProviderFailedError: <lasso.ServerAddProviderFailedError(-202): Failed to add new provider.>


Please provide details regarding why this is happening as I am not able to debug further into the package. And also does lasso work with Shibboleth IDP. If it doesn't what's the other alternative I have?

houzefa-abba:
PR canceled for now; dev will continue on <https://bitbucket.org/xcg/auth_saml> for the moment.

PR #277, opened 2015-10-20
[ADD][8.0] auth_supplier: New Module.

sergio-teruel:

PR #280, opened 2015-10-28
Fix argument name that did not let users to sign up in runbot.

yajo:
See https://github.com/OCA/server-tools/pull/277#discussion_r42502847. This should fix it.

CC @sergio-incaser @rafaelbn.

### DIFF
--- a/auth_from_http_remote_user/res_users.py
+++ b/auth_from_http_remote_user/res_users.py
@@ -35,7 +35,7 @@ class res_users(orm.Model):
     }
 
     def copy(self, cr, uid, id, default=None, context=None):
-        default = default or {}
+        default = default = {} if default is None else default
         default['sso_key'] = False
         return super(res_users, self).copy(cr, uid, id, default, context)
 
--- a/auth_from_http_remote_user/res_users.py
+++ b/auth_from_http_remote_user/res_users.py
@@ -34,10 +34,10 @@ class res_users(orm.Model):
                                readonly=True),
     }
 
-    def copy(self, cr, uid, rid, default=None, context=None):
+    def copy(self, cr, uid, id, default=None, context=None):
         default = default or {}
         default['sso_key'] = False
-        return super(res_users, self).copy(cr, uid, rid, default, context)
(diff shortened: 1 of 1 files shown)


moylop260:
:+1:

auth_from_http_remote_user/res_users.py
```diff
@@ -34,10 +34,10 @@ class res_users(orm.Model):
                                readonly=True),
     }
 
-    def copy(self, cr, uid, rid, defaults=None, context=None):
-        defaults = defaults or {}
-        defaults['sso_key'] = False
-        return super(res_users, self).copy(cr, uid, rid, defaults, context)
+    def copy(self, cr, uid, rid, default=None, context=None):
```
moylop260: s/rid/id/g

auth_from_http_remote_user/res_users.py
```diff
@@ -34,10 +34,10 @@ class res_users(orm.Model):
...
+        default = default or {}
+        default['sso_key'] = False
+        return super(res_users, self).copy(cr, uid, rid, default, context)
```
moylop260: s/rid/id/g

rafaelbn:
:+1:

sebalix:
:+1:

rafaelbn:
@OSguard @max3903 this PR I think is ready to be merge after a PSC review. Thanks
cc @pedrobaeza

auth_from_http_remote_user/res_users.py
```diff
@@ -34,10 +34,10 @@ class res_users(orm.Model):
                                readonly=True),
     }
 
-    def copy(self, cr, uid, rid, defaults=None, context=None):
-        defaults = defaults or {}
-        defaults['sso_key'] = False
-        return super(res_users, self).copy(cr, uid, rid, defaults, context)
+    def copy(self, cr, uid, id, default=None, context=None):
+        default = default or {}
```
pedrobaeza: Use here `default = {} if default is None else default` to not reinstantiate variable in case of False evaluation of it.

yajo:
Fixed last comment, and quite a few :+1: no?

PR #287, opened 2015-11-02
[ADD] users_ldap_push

hbrunn:
# 
# Push users to LDAP

This module was written in order to use Odoo as a frontend for creating LDAP
entries by creating user records. Updates to the user record will be propagated
to the linked LDAP entry afterwards.

When users change their passwords, they will be updated in the LDAP directory
too.
# Configuration

On the LDAP parameters of your company, check _Create ldap entry_ in order to
activate this functionality. Be sure to configure a bind DN that has
appropriate permissions to create and modify entries.

Fill in the object classes newly created entries should contain, separated by
colons. Those classes will determine which mappings from Odoo fields to LDAP
attributes you need. This is highly dependent on your LDAP setup.

For a standard slapd setup, you might want to use object classes
`inetOrgPerson,shadowAccount` and the following mapping:

========== ============== ==
Odoo field LDAP attribute DN
========== ============== ==
Login      userid         X
Name       cn
Name       sn
========== ============== ==

Matching is done by the new field _ldap_entry_dn_, so after installing this
module, you'll probably want to set this field. The module will write it when
a user logs in via Odoo.
# Usage

When you create or update users, their corresponding LDAP entries will be
updated too.


moylop260:
:+1:  (without test)

users_ldap_push/models/res_company_ldap.py
```diff
@@ -0,0 +1,68 @@
...
+            result.setdefault(
+                'create_ldap_entry_field_mappings',
+                [
+                    (0, 0, {
+                        'field_id':
+                        self.env.ref('base.field_res_users_login').id,
+                        'attribute': 'userid',
+                        'use_for_dn': True,
+                    }),
+                ])
```
gurneyalex: is there a reason why you're not defining a dedicated default method and passing it as the `default` kwarg to the field?

users_ldap_push/models/res_users.py
```diff
@@ -0,0 +1,161 @@
...
+        return bool(self._get_ldap_configuration())
+
+    @api.multi
+    def _get_ldap_configuration(self):
+        self.ensure_one()
+        return self.sudo().company_id.ldaps.filtered('create_ldap_entry')[:1]
+
+    @api.multi
+    def _get_ldap_values(self, values):
+        self.ensure_one()
```
gurneyalex, line 58: maybe you could avoid the redundant call here, _get_ldap_configuration calls this on the next line.

users_ldap_push/models/res_users.py
```diff
@@ -0,0 +1,161 @@
...
+                continue
+            result[mapping.attribute] = [str(values[field_name])]
+        if result:
+            result['objectClass'] = conf.create_ldap_entry_objectclass\
+                .encode('utf-8').split(',')
+        return result
+
+    @api.multi
+    def _get_ldap_dn(self, values):
+        self.ensure_one()
```
gurneyalex, line 73: ditto

users_ldap_push/models/res_users.py
```diff
@@ -0,0 +1,161 @@
...
+                (ldap_configuration.ldap_binddn or '').encode('utf-8'),
+                (ldap_configuration.ldap_password or '').encode('utf-8'))
+
+            if not this.ldap_entry_dn:
+                dn = this._get_ldap_dn(values)
+                try:
+                    ldap_connection.add_s(
+                        dn,
+                        ldap.modlist.addModlist(ldap_values))
+                except ldap.LDAPError, e:
```
gurneyalex: I recommand using `except ldap.LDAPError as e` (or better `as exc`). I think this is the only valid syntax in python3, and the readability is better.

users_ldap_push/models/res_users.py
```diff
@@ -0,0 +1,161 @@
...
+                    ldap_connection.unbind_s()
+                    raise exceptions.Warning(_('Error'), e.message)
```
gurneyalex: would we want to log something, in case the action is triggered from an ir.cron for instance? or maybe exceptions.Warning are already logged?

users_ldap_push/models/res_users.py
```diff
@@ -0,0 +1,161 @@
...
+                assert ldap_entry, '%s not found!' % this.ldap_entry_dn
+                ldap_entry = ldap_entry[0][1]
+                try:
+                    ldap_connection.modify_s(
+                        dn,
+                        ldap.modlist.modifyModlist(ldap_entry, ldap_values))
+                except ldap.LDAPError, e:
+                    ldap_connection.unbind_s()
+                    raise exceptions.Warning(_('Error'), e.message)
+            ldap_connection.unbind_s()
```
gurneyalex: I wish we had python ldap3 and context managers :unamused: 

In the meantime, I think it would be easier to have 1 single call to `ldap_connection.unbind_s()` in the `finally` block of a `try` statement inserted just after the call to `bind_s` (~line 97). I'd factor out the whole body of the try block in a separate method for better readability.

users_ldap_push/models/res_users.py
hbrunn, line 58: are those calls expensive? I try to develop the habit to say `ensure_one` on every `@api.multi` function that makes use of assignments in order to see at one glace what kind of method it is

hbrunn:
@gurneyalex thanks for the review, most points fixed, the remaining points left to discussion

gurneyalex:
:+1:

users_ldap_push/models/res_users.py
StefanRijnhart, line 58: Agreed @hbrunn

users_ldap_push/models/res_company_ldap_field_mapping.py
```diff
@@ -0,0 +1,40 @@
...
+    _name = 'res.company.ldap.field_mapping'
+    _description = 'Mapping from Odoo fields to ldap attributes'
+
+    field_id = fields.Many2one(
+        'ir.model.fields', string='Odoo field', required=True,
+        domain=lambda self: self._field_id_domain())
+    attribute = fields.Char('LDAP attribute', required=True)
+    use_for_dn = fields.Boolean('DN')
+    ldap_id = fields.Many2one(
+        'res.company.ldap', string='LDAP comfiguration', required=True)
```
StefanRijnhart: Typo: _configuration_

users_ldap_push/README.rst
```diff
@@ -0,0 +1,80 @@
...
+Name       sn
+========== ============== ==
+
+Matching is done by the new field *ldap_entry_dn*, so after installing this
+module, you'll probably want to set this field. The module will write it when
+a user logs in via Odoo.
+
+Usage
+=====
+
```
StefanRijnhart, line 43: From reading the code, I gather that a checkbox in the user form is used to decide at the time of user creation if the user should be stored in LDAP. If this is the case, can you add this feature to this section?

hbrunn:
@StefanRijnhart thanks, all points taken

StefanRijnhart:
Thanks for the update :+1:

PR #292, opened 2015-11-09
[9.0][MIG]ldap_populate 

flotho:


users_ldap_populate/view/users_ldap.xml
```diff
@@ -6,7 +6,7 @@
             <field name="model">res.company</field>
             <field name="inherit_id" ref="auth_ldap.company_form_view"/>
             <field name="arch" type="xml">
-                <xpath expr="//form[@string='LDAP Configuration']" position="inside">
+                <xpath name="populate_ldap" expr="//field[@name='ldaps']/form" position="inside">
```
dreispt, line 9: I don't think you should use `name` in an xpath element.

gurneyalex, line 9: why? the name is not translatable (while the string is, which is way the original version was changed I believe)

gurneyalex:
:+1:

users_ldap_populate/view/users_ldap.xml
dreispt, line 9: My bad, `name` is perfectly OK.
I was just checking if you guys were paying attention :wink:

dreispt:
:+1:

max3903:
:+1:

pedrobaeza:
You haven't checked that there's a README.rst file and the description is there with the proper OCA's template. Please make another PR with these changes.

PR #296, opened 2015-11-20
[WIP] Extend Automated Action Rules and support Bot agents

dreispt:
This works but is maturing.
I would appreciate some early feedback on it from other people.
Please see the README for details.

Thanks!

### DIFF
--- a/base_action_agent/README.rst
+++ b/base_action_agent/README.rst
@@ -1,0 +1,188 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Action Rules Extensions
+=======================
+
+Summary of features provided:
+
+  * Rule Sets organize Automated Rules in logical sets
+  * Run Automated Actions using a specific user.
+  * Define Rule filters composed of several combined Python snippets
+
+
+Rule sets are capable to provide an alternative workflow system:
+
+  * A Workflow is defined using a Rule Set
+  * Nodes, or States, are represented by Facts
+  * Signals, or transition conditions, are represented by Facts using
+    changed / old vs new values
+  * Transitions are represented by Action Rules
+
+
+With the "Run As" feature we have basic support for Bots or Agents:
+
+  * Create a User for the Bot. Make sure it has the necessary security permissions.
+  * Create a new Ruleset and set the Bot as its User.
+  * Create the Automated Action to be ran by the Bot.
+  * Check the Bot's ruleset "Enabled" flag to active the Rules.
+
+Make your OdooBot the best worker in the company!
+
+
+Rule Sets
+---------
+
+As the number of Action Rules increases it may get confusing.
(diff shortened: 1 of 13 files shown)


base_action_agent/README.rst
```diff
@@ -0,0 +1,185 @@
...
+    equal to Ready:
+    ``self.stage_id.name=='New' and self.kanban_state == 'ready'``
+
+
+Known issues / Roadmap
+======================
+
+  * The Bot support can be greatly expanded.
+    It could be plugged to Artificial Intelligence techniques, starting with
+    rule based expert systems up to machine leaning techniques.
```
yvaucher: s/leaning/learning ? :smile:

base_action_agent/README.rst
```diff
@@ -0,0 +1,185 @@
...
+
+These are Python expressions, and their evaluation context has available:
+
+  * ``self``, ``obj``: is the record object, after the create/write operation
+  * ``env``, ``context``, and ``user``
+  * ``creating``, ``inserting``: True if triggered by a create operation
+  * ``writing``, ``updating``: True if triggered by a write operation
+  * ``old('<field>')``: gets the old value before the write operation
+  * ``new('<field>')``: gets the new value after the create/write operation
+  * ``changed('<field>')``: gets the new value, if changed or if inserting
```
rco-odoo: To me `changed` looks redundant with `new`, which is itself redundant with `self`.
Moreover, changing a field to a falsy value looks the same as not changing it, which I find confusing.
In the example below, `not changed('user_id')` actually means "responsible not changed **or changed to nobody**".

What about passing changes as a dictionary instead?
This could merge both `new` and `changed` in a less confusing API, IMHO.
Taking the example,
- `'user_id' not in changed` means "responsible not changed";
- `'user_id' in changed and not changed['user_id']` means "responsible changed to nobody".

We have incorporated similar ideas in version 9.0, but in a crappier form, though :confused:

hbrunn: :+1:

dreispt: @rco-odoo Thanks for the input - you have a point about the change to empty values.
You could check `new` and `old` for detecting that, but it would probably be cumbersome expressions.
The implementation of `changed`as a lambda id for efficiency: it is only executed if and when needed. Precomputing a dictionary would be unnecessarity expensive.

BTW, am I trying to reimplement something already in v9?

dreispt:
I'm happy with the design now.
Still a lot of small improvements possible (look to `TODO` in the code) but the main pillars are all there.

In the evaluated expressions ("Facts"), 
- `changed()` now accepts multiple arguments and returns a boolean.
- `changed_to()` function and the original `vals` dict were added
- `Warning` was removed since that is to be done by the Rule's Server Actions.

dreispt:
Suggestion for a rule engine to plug in: https://github.com/knipknap/SpiffWorkflow

PR #298, opened 2015-11-26
[FIX] Existing implementation was broken after API v8 upgrade commit

amsys:
Existing implementation was broken after API v8 upgrade commit, it magiically worked only during user creation for the first time:

https://github.com/OCA/server-tools/commit/5710085027d909b886c56f678322f5dfd27cf5d1

`self` was used instead of `this` on several places.

This commit aims to resolve this issue by moving misleading parts out of the get_or_create_user

Additionally, the logger is now global.

### DIFF
--- a/users_ldap_groups/users_ldap_groups.py
+++ b/users_ldap_groups/users_ldap_groups.py
@@ -19,11 +19,14 @@
 #
 ##############################################################################
 
+import logging
 from openerp import models
 from openerp import fields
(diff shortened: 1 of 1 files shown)


gurneyalex:
:+1:

moylop260:
:+1:

moylop260:
:+1:

sebalix:
:+1:

PR #300, opened 2015-11-26
[FIX] Allow to deactivate and restore quick creation without restarting

StefanRijnhart:
The implementation marks seen model objects with a special attribute to prevent unnecessary processing.  If a model was modified, _register_hook was called but the attribute prevented quick creation to be disabled. Also, re-enabling quick creation at run time was not supported.

### DIFF
--- a/base_optional_quick_create/model.py
+++ b/base_optional_quick_create/model.py
@@ -40,26 +40,34 @@ class IrModel(orm.Model):
             return wrapper
 
         for model in self.browse(cr, SUPERUSER_ID, ids):
-            if model.avoid_quick_create:
-                model_name = model.model
-                model_obj = self.pool.get(model_name)
-                if model_obj and not hasattr(model_obj, 'check_quick_create'):
-                    model_obj._patch_method('name_create', _wrap_name_create())
-                    model_obj.check_quick_create = True
+            model_name = model.model
+            model_obj = self.pool.get(model_name)
+            if model_obj and not hasattr(model_obj.name_create, 'origin'):
+                model_obj._patch_method('name_create', _wrap_name_create())
         return True
 
     def _register_hook(self, cr):
-        self._patch_quick_create(cr, self.search(cr, SUPERUSER_ID, []))
+        ids = self.search(
+            cr, SUPERUSER_ID, [('avoid_quick_create', '=', True)])
+        self._patch_quick_create(cr, ids)
         return super(IrModel, self)._register_hook(cr)
 
     def create(self, cr, uid, vals, context=None):
         res_id = super(IrModel, self).create(cr, uid, vals, context=context)
-        self._patch_quick_create(cr, [res_id])
+        if vals.get('avoid_quick_create'):
+            self._patch_quick_create(cr, [res_id])
         return res_id
 
     def write(self, cr, uid, ids, vals, context=None):
         if isinstance(ids, (int, long)):
             ids = [ids]
         res = super(IrModel, self).write(cr, uid, ids, vals, context=context)
-        self._patch_quick_create(cr, ids)
+        if 'avoid_quick_create' in vals:
+            if vals['avoid_quick_create']:
+                self._patch_quick_create(cr, ids)
+            else:
+                for model in self.browse(cr, uid, ids, context=context):
+                    model_obj = self.pool[model.model]
+                    if hasattr(model_obj.name_create, 'origin'):
+                        model_obj._revert_method('name_create')
         return res
--- a/base_optional_quick_create/README.rst
+++ b/base_optional_quick_create/README.rst
@@ -1,0 +1,75 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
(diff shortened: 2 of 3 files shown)


base_optional_quick_create/model.py
```diff
@@ -38,16 +39,42 @@ def wrapper(cr, uid, name, context=None):
...
+                continue
             if model.avoid_quick_create:
-                model_name = model.model
-                model_obj = self.pool.get(model_name)
-                if model_obj and not hasattr(model_obj, 'check_quick_create'):
+                if not hasattr(model_obj, 'name_create_orig'):
+                    logger.debug(
+                        'Replacing original name_create method on %s',
+                        model.model)
+                    model_obj.name_create_orig = model_obj.name_create
```
hbrunn: how about using `_patch_method`? https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L5268

base_optional_quick_create/model.py
```diff
@@ -38,16 +39,42 @@ def wrapper(cr, uid, name, context=None):
...
+                        model_obj.name_create, model.name)
+            else:
+                if hasattr(model_obj, 'name_create_orig'):
+                    logger.debug(
+                        'Restoring original name_create method on %s',
+                        model.model)
+                    model_obj.name_create = model_obj.name_create_orig
+                    delattr(model_obj, 'name_create_orig')
+            model_obj.check_quick_create = True
         return True
```
hbrunn, line 47: return super's result

base_optional_quick_create/model.py
```diff
@@ -38,16 +39,42 @@ def wrapper(cr, uid, name, context=None):
...
                     model_obj.name_create = self._wrap_name_create(
-                        model_obj.name_create, model_name)
-                    model_obj.check_quick_create = True
+                        model_obj.name_create, model.name)
+            else:
+                if hasattr(model_obj, 'name_create_orig'):
+                    logger.debug(
+                        'Restoring original name_create method on %s',
+                        model.model)
+                    model_obj.name_create = model_obj.name_create_orig
```
hbrunn: if you take my comment above, you can use `_revert_method` here: https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L5299

base_optional_quick_create/model.py
```diff
@@ -38,16 +39,42 @@ def wrapper(cr, uid, name, context=None):
         return wrapper
 
     def _register_hook(self, cr, ids=None):
```
hbrunn: I think it's dangerous to change `_register_hook`'s signature. Better refactor this into its own method and just call it here

pedrobaeza:
Why don't expand https://github.com/OCA/web/tree/8.0/web_m2x_options? In my opinion, this is more a web client issue than a base/server one. You can take the idea of creating ir.parameter per model like in this module: https://github.com/OCA/social/blame/9.0/mail_restrict_follower_selection/README.rst#L13

base_optional_quick_create/model.py
StefanRijnhart: That trick was taken from addons/base_action_rule/base_action_rule.py

StefanRijnhart: But actually I agree with you.

StefanRijnhart:
The advantage of this method is that you can set the option per model
instead of per form. It also allows the administrator user to configure
each model, instead of having to add options to all the fields in all
the views (which is how I remember that module works).

pedrobaeza:
Yeah, I know, but I'm talking about implementing this granularity in that module, and make it at web client level, where it should be (as is this web client who includes the quick create option, not the ORM). Obviously, this requires to be done in JS.

pedrobaeza:
I didn't see that this module already exists, but for me it is a mistake to be implemented this way.

StefanRijnhart:
@pedrobaeza maybe you are right but can we keep this module in its current form in the 8.0 branch for the sake of continuity?

pedrobaeza:
OK, but please put a warning about this in the Known issues for the v9 migration.

base_optional_quick_create/model.py
eLBati: @hbrunn @StefanRijnhart what problem do you see with adding `ids=None`?

StefanRijnhart: @eLBati let's say another module overrides _register_hook for the same class. Its signature does not contain the ids argument, because it is identical to the signature in core Odoo. Your module then calls _register_hook with the new ids argument. If the other module happens to come first, this will cause a crash.

eLBati: @StefanRijnhart @hbrunn I would not know how to modify `_register_hook` keeping its behaviour.
What do you think about describing this incompatibility in README?

StefanRijnhart: Just cooked up this (untested): https://github.com/OCA/server-tools/pull/340/files

eLBati:
@StefanRijnhart as https://github.com/OCA/server-tools/pull/352 is merged, what is the destiny of this?

StefanRijnhart:
@elBati I have rebased

sbidoul:
:+1: 

Can we merge this in 8.0 at least?

eLBati:
:+1: thanks

PR #301, opened 2015-11-26
[FIX] turn off prefetching when reading values

hbrunn:


StefanRijnhart:
Does this prevent fetching all the other regular fields from the records?

sebalix:
Hi,

What kind of error did you encounter? (just to understand the fix and the `prefetch_fields` context key)

hbrunn:
It prevents fetching fields not requested, in my case function fields. And this is not a fix for an error, but a fix for performace.
The key's explanation is here: https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L3238

sebalix:
In our case, we use it to log computed fields (say, fields B and C) based on the field A. This way we have a complete log when the field A is updated. Does this break this use case?
As you say, it can be very slow, and I have in mind another feature: choose the log policy on the rule:
- full log (as it is now, including the computed values, default)
- restricted log (to log just the fields which are updated - the `vals` -, avoiding the read operations before/after the CRUD query, speeding the log process)

hbrunn:
for the current model, it reads all fields anyways (we pass `_fields`), so your use case won't be impacted. But depending on the calling tree, he would prefetch way too many other fields in other models while doing the computation

sebalix:
You are right, I tested on the runbot, there is no impact. :+1:

StefanRijnhart:
:+1:

sebalix:
Hi,

5 days have passed, is there someone available to merge this please?

guewen:
:+1: 
Thanks!

PR #299, opened 2015-11-26
port user_ldap_populate to new API

gurneyalex:
and remove use of deprecated except_orm on the way

PR #311, opened 2015-12-13
Fix 9.0 version number

sbidoul:


sbidoul:
These errors should be detected by MQT.

pedrobaeza:
Why this doesn't have the description in the README.rst?

sbidoul:
@pedrobaeza no idea. I guess it was missed during the review https://github.com/OCA/server-tools/pull/299.

I'm not going to fix that part myself.

rvalyi:
what is the policy for LDAP populate: move to 9.0.1.2.0 as you suggested or reset to 9.0.1.0.0 ?
Not sure about this...

sbidoul:
Is there a policy for that? Keeping 9.0.1.2.0 makes sense to me as it expresses the module is not new.

pedrobaeza:
On previous migration strategies threads, we talked about resetting version number to 9.0.1.0.0, because it's the first version that arrives to that version, so I'm with @rvalyi

elicoidal:
In favor of resetting. It doesnot make sense to keep the progression reached in 8 and set it in 9 as if the module have had several migration step which is not the case

dreispt:
There is no guideline about resetting version number when upgrading to a major version.
IMO the common practice on software is to reset them, but not an issue if not.

:+1: for the trivial fix

sbidoul:
IMO, resetting encourage people to consider the series as independent. I think there should be more efforts in maintaining parity of fixes across series. But maybe it's not realistic...

pedrobaeza:
I prefer to reset, or numbers will be very big on Odoo version 12, 13, etc (if we get that number...). There's no gain indicating through version number that this module is also present in a previous "series". That's why we have the prefix, so 7.0.2.0.0 is the previous series version to 8.0.1.0.0.

sbidoul:
Reset done.

sbidoul:
> There's no gain indicating through version number that this module is also present in a previous "series"

I don't agree with that but okay.

pedrobaeza:
:+1:

yvaucher:
Usually what I have been doing is resetting to 9.0.1.0.0 when version is lower than 9.0.1 (ex 9.0.0.1.20)
Because, as it has been used for a whole version and being migrated we can consider it a major version.

Otherwise, about resetting, firstly I had the same idea as @sbidoul that keeping version gives more info. But now I think resetting version might be preferable as we don't want people to think 8.0.1.4.10 is equivalent to 9.0.1.4.10. As both evolved since x.0.1.2 differently.

yvaucher:
Thus :+1:

PR #312, opened 2015-12-15
[8.0][ADD] Module 'nsca_client' - Send passive alerts to a NSCA daemon

sebalix:
NSCA Client
===========

This is a technical module to send passive alerts to your favorite NSCA daemon (Nagios, Shinken...).
This module is based on the Odoo cron system and requires a NSCA client installed on the system to satisfy the ``/usr/sbin/send_nsca`` command.

Installation
============

To use this module, you need to install a NSCA client.

On Debian/Ubuntu:

    $ sudo apt-get install nsca-client

Configuration
=============

* Configure your server and a passive service in your monitoring tool
  (e.g service ``Odoo Mail Queue`` on host ``MY-SERVER``).

* Declare your NSCA server in the menu Configuration / Technical / NSCA Client / Servers

* Create NSCA checks in the menu Configuration / Technical / NSCA Client / Checks

* Code the methods which will be called by the NSCA checks.

Such methods must return a tuple ``(RC, MESSAGE)`` where ``RC`` is an integer,
and ``MESSAGE`` a unicode string. ``RC`` values and the corresponding status are:

- 0: OK
- 1: WARNING
- 2: CRITICAL
- 3: UNKNOWN

E.g:

```python

class MailMail(models.Model):
    _inherit = 'mail.mail'

    @api.model
    def nsca_check_mails(self):
        mails = self.search([('state', '=', 'exception')])
        if mails:
            return (1, u"%s mails not sent" % len(mails))
        return (0, u"OK")
```


sebalix:
If this module is accepted, it would require the installation of the `nsca-client` package on Travis.
We are using this module to be notified when some jobs of the `connector` framework fail.

pedrobaeza:
You can add it by yourself in the .travis.yml file.

sebalix:
I did not even look `.travis.yml`... thanks.

sebalix:
```
Installing APT Packages (BETA)
Disallowing packages: nsca-client
If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process
```

Package approval process here: https://github.com/travis-ci/apt-package-whitelist/issues/2085

nsca_client/README.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    $ sudo apt-get install nsca-client
+
+Then configure the NSCA client in ``/etc/send_nsca.cfg`` (password
+and encryption method).
+
+Configuration
+=============
+
+To configure this module, you need to:
```
elicoidal, line 26: Screenshots are welcome

nsca_client/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+    "summary": "Send passive alerts to monitor your Odoo application.",
+    "version": "8.0.1.0.0",
+    "category": "Tools",
+    "website": "http://osiell.com/",
+    "author": "ABF OSIELL, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "bin": ['/usr/sbin/send_nsca'],
```
elicoidal: use same format as other keys
"bin": [
'/usr/sbin/send_nsca'
],

nsca_client/demo/demo_data.xml
```diff
@@ -0,0 +1,21 @@
...
+<data>
+
+<record id="demo_nsca_server" model="nsca.server">
+    <field name="name">nagios.example.net</field>
+    <field name="port">5667</field>
+</record>
+
+<record id="demo_nsca_check_mails" model="nsca.check">
+    <field name="server_id" ref="demo_nsca_server"/>
+    <field name="service">ODOO MAIL QUEUE</field>
```
elicoidal: No capitals

nsca_client/models/nsca_check.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf-8 -*-
+# © 2015 ABF OSIELL <http://osiell.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+import shlex
+import subprocess
+
+from openerp import models, fields, api, _
```
elicoidal: Please sort all imports in alphabetical order following: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

nsca_client/models/nsca_server.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2015 ABF OSIELL <http://osiell.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields
```
elicoidal: sort order

nsca_client/models/nsca_server.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class NscaServer(models.Model):
```
elicoidal, line 11: isnot there any module that declare servers and that we could use here and in other modules?
It is a nice to have but I would rather create a server_setup module for that that we could reuse and expand (odoo instance, DNS server etc.)

nsca_client/__openerp__.py
StefanRijnhart: Looks like valid python to me. @elicoidal can you point to the (OCA) guideline that prescribes this?

elicoidal: @StefanRijnhart it is just a format problem not a technical one.
My point was suggesting that for all other keys the brakets are in one separate line and we should be consistent in this one.
No big deal

nsca_client/models/nsca_server.py
StefanRijnhart: @elicoidal I know it is in the guidelines but honestly, this kind of review is less useful just like the guys who went full on pep8 before it was covered by CI.

nsca_client/README.rst
```diff
@@ -0,0 +1,112 @@
...
+
+* Configure your server and a passive service in your moniroting tool
```
StefanRijnhart: typo _monitoring_

nsca_client/demo/demo_data.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data>
+
+<record id="demo_nsca_server" model="nsca.server">
```
StefanRijnhart: Would it make sense to add a default server as regular data, so that additional modules containing checks can use its XML id?

nsca_client/models/nsca_server.py
elicoidal: @StefanRijnhart It is fine by me not to spot them if it will be taken care somewhere else. I know that some work is ongoing for deeper formal check: is what you meant?

nsca_client/demo/demo_data.xml
sebalix: Good idea, I moved it from demo/ to data/.

nsca_client/README.rst
sebalix, line 26: Added

StefanRijnhart:
Thanks for the updates!

The external dependency breaks tests. You would need to exclude this module or replace the external dependency with a runtime check.

sebalix:
I think so. It's not a problem if the exception raised is in the background (cron task)? It will appears only in logs.

StefanRijnhart:
So be it. I expect the users of this module to have the habit of checking their logs when something does not seem to work ;-)

sebalix:
It's OK for me, ready for reviews.

StefanRijnhart:
:+1:, thanks for the followup!

sebalix:
Someone available to put the label "needs review" please?

StefanRijnhart:
Sorry, I seem to have lost my privileges on the OCA repos overnight.

sebalix:
@StefanRijnhart @elicoidal thank you!

elicoidal:
I think the core maintainer team has been removed (?)

sebalix:
I updated this module to:
- avoid to send a message to the NSCA server if the method called does not return (option "Allow void result"),
- configure the hostname of the node at the NSCA server level (and not in a system parameter anymore, as the hostname could be different from one monitoring server to another if several of them are configured - which is our case),
- auto-generate the configuration files used by the `send_nsca` command (stored in `{DATA_DIR}/nsca_client/send_nsca_{nsca_server.id}.cfg`), and configure the password and encryption method to use directly in the NSCA server form.

sebalix:
Another fix to make the configuration file path compatible with a multi-databases instance, e.g: `{DATA_DIR}/nsca_client/{DB_NAME}/send_nsca_{NSCA_SERVER_ID}.cfg`.

legalsylvain (approved):

PR #313, opened 2015-12-17
[8.0][base_custom_info] New module to have custom information associated to any model

yajo:
This module does nothing by itself, so don't lose time doing functional reviews. It is the base module that was sent in https://github.com/OCA/product-attribute/pull/120#issuecomment-165512293.

It just adds a base to make other modules able to add a custom attribute list in any model with 2 LOC.

@carlos-incaser @rafaelbn @pedrobaeza.


base_custom_info/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# © 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from . import models
```
pedrobaeza: Line at the end

base_custom_info/__openerp__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# © 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# © 2015 Antiun Ingeniería S.L. - Jairo Llopis
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': "Base Custom Info",
+    'summary': "Add custom field in models",
+    'category': 'Customize',
```
pedrobaeza: This category is not a standard one

base_custom_info/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+    ],
+    'data': [
+        'views/custom_info_template_view.xml',
+        'views/custom_info_property_view.xml',
+        'views/custom_info_value_view.xml',
+        'views/menu.xml',
+        'security/ir.model.access.csv',
+    ],
+    'author': 'Antiun Ingeniería S.L., '
+              'Incaser Informatica S.L., ',
```
pedrobaeza: Add OCA

base_custom_info/models/custom_info.py
```diff
@@ -0,0 +1,101 @@
...
+    _name = "custom.info.template"
+    _description = "Custom information template"
+    _sql_constraints = [
+        ("name_model",
+         "UNIQUE (name, model_id)",
+         "Another template with that name exists for that model."),
+    ]
+
+    name = fields.Char(translate=True)
+    model_id = fields.Many2one(comodel_name='ir.model', string='Model')
```
pedrobaeza: Is it required a hard link? Isn't enough with a char with the name of the model? This will ease the data inclusion.

base_custom_info/models/custom_info.py
```diff
@@ -0,0 +1,101 @@
...
+    _name = "custom.info.value"
+    _description = "Custom information value"
+    _rec_name = 'value'
+    _sql_constraints = [
+        ("property_model_res",
+         "UNIQUE (property_id, model, res_id)",
+         "Another property with that name exists for that resource."),
+    ]
+
+    model = fields.Char(index=True, required=True)
```
pedrobaeza: Why this field here and being not related?

base_custom_info/models/custom_info.py
```diff
@@ -0,0 +1,101 @@
...
+        ("name_template",
+         "UNIQUE (name, template_id)",
+         "Another property with that name exists for that template."),
+    ]
+
+    name = fields.Char(translate=True)
+    template_id = fields.Many2one(
+        comodel_name='custom.info.template',
+        string='Template')
+    info_value_ids = fields.One2many(
```
pedrobaeza, line 64: Why do you need this one2many? The idea was to have at this level the value as a text, not mimicking a selection field (or at least not only doing that). Maybe you should implement something like the ir.property: several fields for storing each of the possible values (float, char, ref...), and one field to indicate the type of the data.

base_custom_info/models/custom_info.py
```diff
@@ -0,0 +1,101 @@
...
+            for info_name in self.custom_info_template_id.info_ids:
+                if info_name not in info_list:
+                    self.custom_info_ids |= self.custom_info_ids.new({
+                        'model': self._name,
+                        'property_id': info_name.id,
+                    })
+
+    @api.multi
+    def unlink(self):
+        info_values = self.mapped('custom_info_ids')
```
pedrobaeza, line 121: You don't need to do this if you set `ondelete="cascade"`

base_custom_info/README.rst
```diff
@@ -0,0 +1,90 @@
...
+This module allows to create custom fields in models without altering model's
+structure.
+
+Installation
+============
+
+This module serves as a base for other modules that implement this behavior in
+concrete models.
+
+This means that you should **not** install this module by itself. If you found
```
elicoidal: This module is a technical dependency and is be installed in parallel to other modules.

base_custom_info/README.rst
```diff
@@ -0,0 +1,90 @@
...
+
+To manage their values, you need to:
+
+* Go to *Settings > Custom Info > Values*.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/135/8.0
+
+Development
```
elicoidal, line 48: this is a sub category of usage no?
usage single highlight

base_custom_info/models/custom_info.py
yajo: Hard link? If you mean the Many2one, It's the common way to relate models AFAIK.

base_custom_info/models/custom_info.py
yajo, line 121: I know, but that means creating a Many2one field in `custom.info.value`, and that means creating a table for each related module and complicating inheritance.

What we do here is exactly mimic the `ondelete="cascade"` to be able to [inherit models with just one line](https://github.com/OCA/product-attribute/pull/121/files#diff-d7166026a131685af9abf8415c733581R11). This is what the mail module does.

base_custom_info/README.rst
yajo, line 48: It's a different one. Developing on a module is not the same as using it.

base_custom_info/models/custom_info.py
yajo, line 64: It shows a tree of values using this property. Setting the data type complicates things, but I can add it to known issues since most times strings are enough.

yajo:
All fixed (what needed to be fixed).

yajo:
Well, now.

base_custom_info/models/custom_info.py
pedrobaeza, line 64: But that's not the idea behind the module. The idea is to have custom info, not custom "selections".

yajo, line 64: I think you are confusing the model. This one is about properties: Size, color, etc. Since the same property is often shared between different objects, properties have their own table here.

This One2many allows the administrator to see how many values are set in any model with this property, it is not intended for end users.

Actual values are stored in `custom.info.value`, which has a property_id field pointing to this model to choose the property, a model_id + res_id fields combination to know where it applies to, and a value field to store that property's value.

rafaelbn:
Please @sergio-incaser @carlos-incaser @pedrobaeza @antespi  could you review? Thanks

sergio-teruel:
When try enter into a product
Odoo Server Error
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 537, in _handle_exception
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 574, in dispatch
    result = self._call_function(*_self.params)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/service/model.py", line 118, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 307, in checked_call
    return self.endpoint(_a, *_kw)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 803, in __call__
    return self.method(_args, *_kw)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/http.py", line 403, in response_wrap
    response = f(_args, *_kw)
  File "/home/sergio/desarrollo/odoo/v80/odoo/addons/web/controllers/main.py", line 944, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/api.py", line 256, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/models.py", line 3145, in read
    result = BaseModel.read(records, fields, load=load)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/api.py", line 254, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/models.py", line 3180, in read
    self._read_from_database(stored, inherited)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/api.py", line 254, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/models.py", line 3367, in _read_from_database
    res2 = self._columns[f].get(cr, self._model, ids, f, user, context=context, values=result)
  File "/home/sergio/desarrollo/odoo/v80/odoo/openerp/osv/fields.py", line 776, in get
    domain = self._domain(obj) if callable(self._domain) else self._domain
  File "/home/sergio/desarrollo/odoo/v80/server-tools-yajo/base_custom_info/models/custom_info.py", line 80, in <lambda>
    self.env["ir.model"].search([("model", "=", self._name)]).id)],
AttributeError: 'product.template' object has no attribute 'env'

yajo:
Could you try now please?

rafaelbn:
Travis fails

yajo:
Should be fixed now.

rafaelbn:
Very nice! Ready!

@carlos-incaser @sergio-incaser @pedrobaeza.

carlosdauden:
The last two commits are not related witdh this module.
:+1:

yajo:
They're there to get Travis :green_heart:

rafaelbn:
:+1: thanks!

base_custom_info/models/custom_info.py
pedrobaeza, line 121: But I think you can still put `ondelete="cascade"` in https://github.com/OCA/server-tools/pull/313/files#diff-e7d857a6918e0167e0c4c4e40d3bb170R97 (`custom_info_ids` field in this model) and it will work (I'm trying now).

yajo, line 121: See the docs. Only Many2one have ondelete AFAIK.

pedrobaeza, line 121: Sorry, I meant on the many2one field of the relation.

yajo, line 121: But there's no m2o field on the other side.

pedrobaeza, line 121: OK, I see. There's no other solution then.

rafaelbn:
Hi, please, this PR is ready to be merged, could any manager take a look? thanks!

pedrobaeza:
@Yajo, you have merged the latest changes from 8.0 branch instead of rebasing. Please remove latest commit, and make a proper rebase, and then I'll merge.

yajo:
Is there any real difference?

pedrobaeza:
Yes, see the commits and the diffs. They are reversed

yajo:
Commits and diffs are OK, I don't know what do you mean...

When you rebase you have to solve conflicts one by one for each problematic commit, and it is a potentially dangerous operation that destroys history. With merge history goes always forward and you only need to solve conflicts once.

That's why I usually prefer merges unless for squashing or something like that: less work, less danger, and at the end you get _exactly the same result_, so IMHO it's OK as it is :wink:

pedrobaeza:
They are not correct, see https://github.com/Yajo/server-tools/commit/be0981ad0202131ac499c4fa144cfe9b1018010f.

It's not a question of preferences, it's a question of not messing up the commit history.

yajo:
I'll rebase. :v:

yajo:
Rebased

pedrobaeza:
Thanks.

:+1:

PR #316, opened 2015-12-24
[8.0][FIX][dead_mans_switch_client] Add demo system parameter and some fixes to README.

### LINKED ISSUE #314: [8.0][dead_mans_switch_client] Module crashes runbots
I'm seeing more and more runbots with :x: because of this module. [This seems the offending line](https://github.com/OCA/server-tools/blob/8.0/dead_mans_switch_client/models/dead_mans_switch_client.py#L54). Any clue on how to fix it?

Example runbot: https://runbot.odoo-community.org/runbot/build/3137787

CC @hbrunn.

yajo:
Fix #314.

### DIFF
--- a/dead_mans_switch_client/models/dead_mans_switch_client.py
+++ b/dead_mans_switch_client/models/dead_mans_switch_client.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2015 Therp BV <http://therp.nl>
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import json
 import logging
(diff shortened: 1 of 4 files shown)


yajo:
It seems the fix works. Runbot's issue is now https://github.com/odoo/odoo/issues/10159, but not #314.

cc @hbrunn

dead_mans_switch_client/models/dead_mans_switch_client.py
```diff
@@ -66,3 +67,19 @@ def alive(self):
...
+        name = "dead_mans_switch_client.url"
+        param = conf.get_param(name)
+
+        if not param:
+            url = "{}/dead_mans_switch/alive".format(
+                conf.get_param(
+                    "report.url",
+                    conf.get_param(
+                        "web.base.url",
+                        "http://localhost")))
```
pedrobaeza, line 84: Why don't take the value from web.base.url parameter and thus avoid an arbitrary URL that doesn't have to be the correct?

yajo, line 84: It's what I'm doing here. The `http://localhost` part is only for when `web.base.url` is not available.

pedrobaeza, line 84: Oh, sorry, very late review...

pedrobaeza:
:+1:

gurneyalex:
:+1:

hbrunn:
:+1:

PR #321, opened 2016-01-01
9.0 base technical features

StefanRijnhart:
Make technical features persistent by adding users to a new group.


base_technical_features/__openerp__.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# © 2016 Opener B.V.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Technical features group",
+    "summary": "Access to technical features without activating debug mode",
+    "version": "8.0.1.0.0",
```
RoelAdriaans: Shoudn't this be 9.0?

StefanRijnhart: Fixed, thanks!

parthivgls:
superb :+1:

ghost:
+1

elicoidal:
magnifique !
:+1:

StefanRijnhart:
What do you think, do we want this as a user preference (for users that are already in base.group_no_one of course)?

max3903:
:+1: for this pr and to have this as a user preference

StefanRijnhart:
Yes, I would prefer that as well.

StefanRijnhart:
It's now a user preference.

nhomar:
:+1:

moylop260:
Could you add `images` key in manifest file?

StefanRijnhart:
@moylop260 I was wondering about that message from the OCA pylint plugin. What is this 'images' key for? I could not find any references in the Odoo documentation or even code but maybe I missed something.

StefanRijnhart:
@moylop260 I think I found it! It's used in the Apps store to indicate the image that should be used in the kanban representation of the module. I don't think my ugly screenshot qualifies for that (also see l10n_es on the Apps site!). If I did not miss anything, maybe this key should not be required after all?

rafaelbn:
Thanks! and happy new year! :+1:

sebalix:
Nice!
Should we add the PO template or it's useless?

StefanRijnhart:
That will come from Transifex.

sebalix:
Okay. Tested, :+1:

pedrobaeza:
:+1:

PR #323, opened 2016-01-05
Adding Module base_copy_access

mikevhe18:


andhit-r:
I think WizardBaseCopyUserAccess should be **models.TransientModel**, not **models.Model**

mikevhe18:
@andhit-r Yes, you're right. I'd missed it. Thank You

mikevhe18:
Dear Community,
Can someone tell me how to pass the coverage/coveralls checks?
I'm a newbie here.
Thank You

sebalix:
Hi,

What is the purpose of the module? The description is not very clear to me (I encourage you to follow the [CONTRIBUTING](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md) guide, and the [template module](https://github.com/OCA/maintainer-tools/tree/master/template/module) (README.rst, etc).
I read the code, and it seems that the wizard copies the groups of one user on the current user. Excepted the `admin` user account, no user should be able to do this by default. If a user is able to copy any user rights, this is a security hole. Maybe I do not understand well the purpose of this, that's why I am asking :)

mikevhe18:
Hi,

The main purpose of this module is to help out admin to work easier. Sometimes users want to delegate their access right to another user for some reason. Not all user can use this feature , only admin can be.

sebalix:
Okay, I was misreading the code (we are using https://github.com/osiell/base_user_role for this kind of use case). Anyway, you should update your module to follow the CONTRIBUTING guidelines (new module in 8.0 should be in written in new API) and add a README.rst file, explaining how to use this module (select users in the list, then launch the action, etc).

dreispt:
Please rebase on latest server-tools 8.0: I can see commits regarding unrelated modules, such as audit_log.

andhit-r (changes requested):
Remove empty key on https://github.com/OCA/server-tools/pull/323/files#diff-378482bb041861896faabba9932bb193R12

andhit-r (changes requested):
Use models.TransientModel on https://github.com/OCA/server-tools/pull/323/files#diff-032362b48ae883d61fa810d7f9f83c57R10

andhit-r (changes requested):
Add self.ensure_one()

PR #324, opened 2016-01-05
Credit creators, using same name across modules to avoid split statistics

yajo:
Same as https://github.com/OCA/partner-contact/pull/235


pedrobaeza:
:+1: 

Fast-tracking...

PR #328, opened 2016-01-08
[FIX] datetime_formatter - Set the correct locale to translate months and days

sebalix:
This is a common issue with OpenERP/Odoo since a long time: the translation of months and days in non-english dates.
The PR is not a nice fix (a proper fix should be global, and not in this module especially), so it's OK if it is refused.
What do you think?

### DIFF
--- a/datetime_formatter/__openerp__.py
+++ b/datetime_formatter/__openerp__.py
@@ -5,9 +5,10 @@
 {
     "name": "Date & Time Formatter",
     "summary": "Helper functions to give correct format to date[time] fields",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
     "category": "Tools",
     "website": "https://grupoesoc.es",
+    "images": [],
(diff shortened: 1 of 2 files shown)


hbrunn:
I think your approach is okay. A module that monkey patches this into core would be nicer, but well.

Two questions:
- `resetlocale` sets the locale to the default locale. Wouldn't it be better to save the current locale and reset that? There might be scenarios where some other code set the locale to whatever else
- I _think_ LC_TIME wins against LC_ALL. Wouldn't it be better to set/reset that?

sebalix:
@hbrunn yep, LC_TIME is a better option, fixed. Travis error is related to `base_external_dbsource`.

hbrunn:
:+1: verified pylint errors are unrelated

hbrunn:
@sebalix please rebase for green builds

sebalix:
Rebased

PR #331, opened 2016-01-13
[DEL] obsolete base_field_serialized

hbrunn:
I wrote this module when fields.serialized was removed from core, but it was restored quite soon after I released this module. I think I made a PR to remove it back then, but can't find it any more, so here we go.

Here the relevant code in core:
https://github.com/OCA/OCB/blob/8.0/openerp/osv/fields.py#L1689
https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L1844


pedrobaeza:
:+1:

guewen:
:+1:

sebalix:
:+1:

PR #333, opened 2016-01-15
[PORT] Port base_suspend_security to 9.0

adrienpeiffer:


base_suspend_security/__openerp__.py
```diff
@@ -30,7 +30,7 @@
     "test": [
     ],
     "auto_install": False,
-    'installable': False,
+    'installable': True,
```
JonathanNEMRY: I think version number should be upgraded too?

gurneyalex:
@adrienpeiffer can you also bump the version number in **openerp**.py ?

lmignon:
@adrienpeiffer @gurneyalex @hbrunn IMO We should advertise that this module must be used with caution in a multi companies environment.  Indeed, when you suspend the security you also suspend record rules and in some cases I've observed that the value of some fields is not linked to the current company.

pedrobaeza:
And a workaround/solution is to include in the context the key `force_company` with the id of the company that must be treated.

lmignon:
@pedrobaeza I'm not sure it's a solution since the record rules are not run at all...

pedrobaeza:
Well, it's not a complete solution for example for many2many records (like the taxes in the product), but it works with property/company_dependent fields.

hbrunn:
@lmignon how is `sudo()`'s behavior here? I think we should do what this function does for consistency

lmignon:
@gurneyalex version bumped

hbrunn:
:+1:

sebalix:
:+1:

pedrobaeza:
:+1:

Issue #327, opened 2016-01-07
[9.0]users_ldap_populate not works


falinwa:
Hello,

I found bug when I tried to use users_ldap_populate module on version 9,
here is the error message:

  File "/opt/falinwa_branch/users_ldap_populate/model/users_ldap.py", line 85, in populate_wizard
    res_id = wizard_obj.create({'ldap_id': self.id}).id
  File "/opt/odoo/openerp/api.py", line 230, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/opt/falinwa_branch/users_ldap_populate/model/populate_wizard.py", line 44, in create
    vals['users_created'] = ldap.action_populate()
  File "/opt/odoo/openerp/api.py", line 230, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/opt/falinwa_branch/users_ldap_populate/model/users_ldap.py", line 55, in action_populate
    for conf in self.get_ldap_dicts(self.cr, self.ids):
AttributeError: 'res.company.ldap' object has no attribute 'cr'

I think, because you use the new API, you need to change from self.cr to self.env.cr

Thank You.

PR #334, opened 2016-01-20
9.0 users_ldap_populate fix 2 bugs

gurneyalex:
- leftovers from old api
- python-ldap does not like Unicode strings

### DIFF
--- a/users_ldap_populate/model/users_ldap.py
+++ b/users_ldap_populate/model/users_ldap.py
@@ -21,7 +21,7 @@
 
 import re
 
-from openerp import models, api
+from openerp import models, api, _
 from openerp.exceptions import UserError
 import logging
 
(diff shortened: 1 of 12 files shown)


users_ldap_populate/model/users_ldap.py
```diff
@@ -61,11 +61,11 @@ def action_populate(self):
                 login_attr = attribute_match.group(1)
             else:
                 raise UserError(
-                    "No login attribute found"
+                    "No login attribute found: "
```
damdam-s: No need for this error message to be translatable ?

sebalix:
Bug related: https://github.com/OCA/server-tools/issues/327

gurneyalex:
good catch @sebalix (I shamefully admit I did not check for an existing PR or issue...)

gurneyalex:
@damdam-s fixed

damdam-s:
:+1:

sebalix:
:+1:

PR #338, opened 2016-01-24
Upgrade Gravatar module to v9, make installable, and add tests

lasley:
Hi All!

This PR removes deprecated `api.one` in the Gravatar module, adds tests, and a few styling changes in the ReadMe.

Not sure how to get Transifex to pass. I tried signing up, but it's still complaining so I ignored it.

### DIFF
--- a/base_user_gravatar/tests/test_res_users.py
+++ b/base_user_gravatar/tests/test_res_users.py
@@ -1,23 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Author: Dave Lasley <dave@laslabs.com>
-#    Copyright: 2016-TODAY LasLabs, Inc. [https://laslabs.com]
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


danimaribeiro:
:+1: Tested on runbot.

base_user_gravatar/tests/test_res_users.py
```diff
@@ -0,0 +1,77 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Please use short headers like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

base_user_gravatar/tests/test_res_users.py
```diff
@@ -0,0 +1,77 @@
...
+        self.url = 'http://www.gravatar.com/avatar/{}?s=200'
+
+    def _test_record(self, ):
+        partner_id = self.env['res.partner'].create(self.partner_vals)
+        self.vals['partner_id'] = partner_id.id
+        return self.env['res.users'].create(self.vals)
+
+    @mock.patch('%s.urllib2' % MODULE_LOCATION)
+    def test_get_gravatar_base64_opens_correct_uri(self, mk, ):
+        ''' Test that gravatar is pinged for image '''
```
pedrobaeza: Use double quote (`"""`) for docstring

lasley:
Thanks for the review @danimaribeiro & @pedrobaeza - changes made per suggestions

pedrobaeza:
:+1:

sebalix:
:+1:

PR #343, opened 2016-01-28
8.0 avoid a crash in log_forwarded_for_ip (even when the module is not installed)

alexis-via:
I sometimes have this in my odoo logs:

```
2016-01-20 13:49:23,463 440 ERROR prod1 openerp.service.server: Worker (440) Exception occured, exiting...
  File "/home/odoo/erp/odoo/openerp/service/server.py", line 756, in run
    self.process_work()
  File "/home/odoo/erp/odoo/openerp/service/server.py", line 788, in process_work
    self.process_request(client, addr)
  File "/home/odoo/erp/odoo/openerp/service/server.py", line 779, in process_request
    self.server.process_request(client, addr)
  File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 649, in __init__
    self.handle()
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 200, in handle
    rv = BaseHTTPRequestHandler.handle(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 234, in handle_one_request
    elif self.parse_request():
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 253, in parse_request
    self.send_error(400, "Bad request version (%r)" % version)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 364, in send_error
    self.log_error("code %d, message %s", code, message)
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 256, in log_error
    self.log('error', *args)
  File "/usr/lib/python2.7/dist-packages/werkzeug/serving.py", line 262, in log
    _log(type, '%s - - [%s] %s\n' % (self.address_string(),
  File "/home/odoo/erp/server-tools/log_forwarded_for_ip/__init__.py", line 9, in address_string
    forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
AttributeError: 'WSGIRequestHandler' object has no attribute 'headers'
```

and the module  log_forwarded_for_ip is not installed.

So I propose the patch to protect against this kind of crash. Beware that I didn't test this patch.

### DIFF
--- a/log_forwarded_for_ip/__init__.py
+++ b/log_forwarded_for_ip/__init__.py
@@ -12,7 +12,7 @@ class log_forwarded_for_ip_installed(models.AbstractModel):
 address_string_original = WSGIRequestHandler.address_string
 
 def address_string(self):
-    if self.pool.get('log.forwarded.for.ip.installed'):
+    if self.env.get('log.forwarded.for.ip.installed'):
         # Code executed if the module log_forwarded_for_ip is installed
         if self.headers and isinstance(self.headers, dict):
             forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
--- a/log_forwarded_for_ip/__init__.py
+++ b/log_forwarded_for_ip/__init__.py
@@ -3,25 +3,13 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from werkzeug.serving import WSGIRequestHandler
-from openerp import models
 
-
-class log_forwarded_for_ip_installed(models.AbstractModel):
-    _name = 'log.forwarded.for.ip.installed'
-
-address_string_original = WSGIRequestHandler.address_string
 
 def address_string(self):
-    if self.env.get('log.forwarded.for.ip.installed'):
-        # Code executed if the module log_forwarded_for_ip is installed
-        if self.headers and isinstance(self.headers, dict):
-            forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
-            if forwarded_for and forwarded_for[0]:
-                return forwarded_for[0]
-        return self.client_address[0]
-    else:
-        # If the module is NOT installed, we execute the original code
-        address_string_original(self)
-    return True
+    if self.headers and isinstance(self.headers, dict):
+        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
+        if forwarded_for and forwarded_for[0]:
+            return forwarded_for[0]
+    return self.client_address[0]
 
 WSGIRequestHandler.address_string = address_string
--- a/log_forwarded_for_ip/__init__.py
+++ b/log_forwarded_for_ip/__init__.py
@@ -3,13 +3,25 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from werkzeug.serving import WSGIRequestHandler
+from openerp import models
 
+
+class log_forwarded_for_ip_installed(models.AbstractModel):
+    _name = 'log.forwarded.for.ip.installed'
+
+address_string_original = WSGIRequestHandler.address_string
 
 def address_string(self):
-    if self.headers and isinstance(self.headers, dict):
-        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
-        if forwarded_for and forwarded_for[0]:
-            return forwarded_for[0]
-    return self.client_address[0]
+    if self.pool.get('log.forwarded.for.ip.installed'):
+        # Code executed if the module log_forwarded_for_ip is installed
+        if self.headers and isinstance(self.headers, dict):
+            forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
+            if forwarded_for and forwarded_for[0]:
+                return forwarded_for[0]
+        return self.client_address[0]
+    else:
+        # If the module is NOT installed, we execute the original code
+        address_string_original(self)
+    return True
 
 WSGIRequestHandler.address_string = address_string
--- a/log_forwarded_for_ip/__init__.py
+++ b/log_forwarded_for_ip/__init__.py
@@ -6,10 +6,10 @@ from werkzeug.serving import WSGIRequestHandler
 
 
 def address_string(self):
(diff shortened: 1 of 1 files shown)


alexis-via:
@pablocm-aserti Could you tell us what you think about this patch ? Blame says you are the author of the code :)

StefanRijnhart:
Maybe the monkeypatch can be placed in a dummy (Abstract) model's _register_hook so that it is only applied on instances on which the module is installed?

alexis-via:
Oh yes, you are right, we should do that ! At Akretion, we call that the "Guewen trick" :)

StefanRijnhart:
I call it the Holger trick ;-)

alexis-via:
I implemented the "Guewen trick" for clean monkey patching. I didn't test when the module is installed (I don't use this module... I just don't want him to cause crashes on my Odoo :).

alexis-via:
@StefanRijnhart Arf ! We need to investigate to know who was the first to have this great idea !

StefanRijnhart:
Do you still need to push a commit?

log_forwarded_for_ip/__init__.py
```diff
@@ -3,13 +3,25 @@
...
+    _name = 'log.forwarded.for.ip.installed'
+
+address_string_original = WSGIRequestHandler.address_string
+
 def address_string(self):
-    forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
-    if forwarded_for and forwarded_for[0]:
-        return forwarded_for[0]
-    else:
+    if self.pool.get('log.forwarded.for.ip.installed'):
```
bealdav: You can replace it by

``` python
if module_name in self.env.registry._init_modules:
```

no more abstract model to use
Reference:

https://github.com/OCA/connector/issues/85
https://github.com/OCA/connector/blob/8.0/connector/connector.py#L64

log_forwarded_for_ip/__init__.py
```diff
@@ -3,13 +3,25 @@
...
+    _name = 'log.forwarded.for.ip.installed'
+
+address_string_original = WSGIRequestHandler.address_string
+
 def address_string(self):
-    forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
-    if forwarded_for and forwarded_for[0]:
-        return forwarded_for[0]
-    else:
+    if self.env.get('log.forwarded.for.ip.installed'):
```
bealdav: You can replace it by

``` python
if 'log_forwarded_for_ip' in self.env.registry._init_modules:
```

no more abstract model to use
Reference:

https://github.com/OCA/connector/issues/85
https://github.com/OCA/connector/blob/8.0/connector/connector.py#L64

alexis-via:
OK, I should test it more properly before commiting ; I'm just angry at spending time to debug a module that I don't use !!!

The problem here is that we monkey patch a method that is NOT in Odoo. So self is not the usual "self" of Odoo, so we can't do self.env !!!

pablocm-aserti:
Sorry for the faulty lines, the module has already been in production for several months in several instances and we haven't had an empty header case yet, even with the xmlrpc calls of oerplib.
The fix LGTM in its current state, I'll try to find a way to load it only if it is installed. Just give me some time, I'm not as lightning fast as you! :)

StefanRijnhart:
@alexis-via I don't understand the last commit msg 'Remove Guewen's trick, because we don't monkey patch an Odoo method !'. It's more about whether or not a monkeypatch is applied, and by doing this in _register_hook you make sure it is only applied on instances on which the module is actually installed. What is the reason you went the other way?

pablocm-aserti:
Thanks, I didn't know the "Guewen trick", it is great!
This code indeed loads the monkey patch only when the class is loaded because the module is installed.
Too bad there is not an _unregister_hook to undo this, because I can't find a way to restore the original function if the module is uninstalled. The only way is restarting the server.
Anyway, it is way better than the current solution.

/!\ self.headers is a mimetools.Message, not a dict

``` python
from mimetools import Message

def address_string(self):
    if self.headers and isinstance(self.headers, Message):
        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
        if forwarded_for and forwarded_for[0]:
            return forwarded_for[0]
    return self.client_address[0]


class log_forwarded_for_ip_installed(models.AbstractModel):
    _name = 'log.forwarded.for.ip.installed'

    def _register_hook(self, cr):
        WSGIRequestHandler.address_string = address_string
        return super(log_forwarded_for_ip_installed, self)._register_hook(cr)
```

pablocm-aserti:
Mmm... I'll give "uninstall_hook" a try

hbrunn:
@pablocm-aserti you should check if you did this already, because `_register_hook` can be called more often than once in the process' lifetime. Set some marker on `WSGIRequestHandler`, and only patch if it doesn't exist (use `hasattr`)

This shall henceforth be known as the Holger trick ;-)

For uninstalling, use an `uninstall_hook`: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#installation-hooks

pablocm-aserti:
Thanks @hbrunn !
I've a version with the marker and the uninstall_hook. The first works perfectly by the uninstall_hook seems to restore the function only in the WSGIRequestHandler of the current worker, and then you see a mix of source IPs and proxy IPs in the log. But I guess it's not a big problem, eventually they will start using the original function when they get relaunched, and at least it is better than before.
Shall I propose the PR?

hbrunn:
@pablocm-aserti yes, please. Also include `fixes #343` in your commit message, then this PR should be closed automatically when yours is merged. @mention people involved here to get it reviewed and merged fast.

You should be able to enumerate the current workers somehow in your uninstall hook, but I doubt it's worth the time

alexis-via:
So let's abandon this PR in favor of @pablocm-aserti 's PR https://github.com/OCA/server-tools/pull/344

PR #344, opened 2016-01-28
[FIX] Crash on some requests without header

pablocm-aserti:
[IMP] Only load the patch when the module is installed

Fixes #343

### DIFF
--- a/log_forwarded_for_ip/hooks.py
+++ b/log_forwarded_for_ip/hooks.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# © 2015 Aserti Global Solutions
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from werkzeug.serving import WSGIRequestHandler
+
+
+def restore_address_string(cr, reg):
+    if hasattr(WSGIRequestHandler, '_address_string_org'):
+        WSGIRequestHandler.address_string = \
+            WSGIRequestHandler._address_string_org
+        del WSGIRequestHandler._address_string_org
--- a/log_forwarded_for_ip/models/log_forwarded_for_ip_installed.py
+++ b/log_forwarded_for_ip/models/log_forwarded_for_ip_installed.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# © 2015 Aserti Global Solutions
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from werkzeug.serving import WSGIRequestHandler
+from mimetools import Message
+from openerp import models
+
+
+def address_string(self):
+    if self.headers and isinstance(self.headers, Message):
+        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
+        if forwarded_for and forwarded_for[0]:
+            return forwarded_for[0]
+    return self.client_address[0]
+
+
+class LogForwardedForIpInstalled(models.AbstractModel):
+    _name = 'log.forwarded.for.ip.installed'
+
+    def _register_hook(self, cr):
+        if not hasattr(WSGIRequestHandler, '_address_string_org'):
+            WSGIRequestHandler._address_string_org = \
+                WSGIRequestHandler.address_string
+            WSGIRequestHandler.address_string = address_string
+        return super(LogForwardedForIpInstalled, self)._register_hook(cr)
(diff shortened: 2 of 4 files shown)


pablocm-aserti:
Thanks @alexis-via @hbrunn @StefanRijnhart for the help, reviews are welcome

log_forwarded_for_ip/__init__.py
```diff
@@ -3,13 +3,24 @@
...
+    return self.client_address[0]
 
-WSGIRequestHandler.address_string = address_string
+
+class log_forwarded_for_ip_installed(models.AbstractModel):
+    _name = 'log.forwarded.for.ip.installed'
+
+    def _register_hook(self, cr):
+        if not hasattr(WSGIRequestHandler, '_odoo_str_replaced'):
+            WSGIRequestHandler._odoo_str_replaced = True
```
hbrunn: consider assigning the original to your marker, then you don't need the global original variable below

log_forwarded_for_ip/__init__.py
```diff
@@ -3,13 +3,24 @@
...
-        return self.client_address[0]
+    if self.headers and isinstance(self.headers, Message):
+        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
+        if forwarded_for and forwarded_for[0]:
+            return forwarded_for[0]
+    return self.client_address[0]
 
-WSGIRequestHandler.address_string = address_string
+
+class log_forwarded_for_ip_installed(models.AbstractModel):
```
hbrunn: even though it's tiny, please do that in `models/log_forwarded_for_ip_installed.py`

log_forwarded_for_ip/__init__.py
hbrunn: I use something like `_${orignal_name}_org` for my markers, this makes it more clear what it's about if you ever come across the memer in a debugger

pablocm-aserti:
Done and tested

log_forwarded_for_ip/models/log_forwarded_for_ip_installed.py
```diff
@@ -0,0 +1,26 @@
...
+
+def address_string(self):
+    if self.headers and isinstance(self.headers, Message):
+        forwarded_for = self.headers.get('X-Forwarded-For', '').split(',')
+        if forwarded_for and forwarded_for[0]:
+            return forwarded_for[0]
+    return self.client_address[0]
+
+
+class log_forwarded_for_ip_installed(models.AbstractModel):
```
StefanRijnhart: Travis hints at using camelcase, i.e. LogForwardedForIpInstalled for the class name. I agree with that.This won't make Travis completely happy, because it also demands an 'images' key in the manifest which is nonsense.

pedrobaeza: images key is only a warning, not mandatory, but it makes sense because new Odoo apps put the first image as background of the tile, so if we want to make branding, we need to make good modules with some "marketing" material. But as stated, this is only a recommendation, not a requirement.

StefanRijnhart: @pedrobaeza: commented here now https://github.com/OCA/maintainer-quality-tools/pull/298#issuecomment-176335149

StefanRijnhart:
Looking good, apart from the style nag! Thanks for picking this up so quickly.

alexis-via:
I'm not an expert of monkey-patching, but it seems better than before :) :+1:

log_forwarded_for_ip/hooks.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2015 Aserti Global Solutions
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from werkzeug.serving import WSGIRequestHandler
+
+
+address_string_original = WSGIRequestHandler.address_string
```
hbrunn: this is not needed now?

pablocm-aserti: changes in this file were missing from my previous commit, sorry
Fixed

hbrunn:
:+1:

hbrunn:
but thinking about it, isn't this module fixing problems you wouldn't have if you employed `proxy_mode' properly? I really don't know, so this is a genuine question

pablocm-aserti:
proxy_mode is needed in order to make it work properly (IIRC, otherwise you get sometimes redirected to wrong addresses) but does not show the source IP in the log. In fact, that is handled by werkzeug who seems oblivious to the mode you are running the server on. That is why it had to be monkeypatched

StefanRijnhart:
Thanks for the updates! :+1:

PR #345, opened 2016-01-30
[ADD] new module 'mass_sorting' to allow user to mass sort lines in any models;

legalsylvain:
Hi, 

I propose a module 'mass_sorting' that works like  'mass_editing' (same repository) to reorder any model that have a one2many field that have a sequence field into.

Once the configuration done, you can reorder massivily any items.

It can be usefull to reorder lines in purchase orders / sale orders, or moves in pickings, by custom criterias.
### Sample

Sale Order -> Order Lines -> order by product and quantity desc
1_mass_sort_config
### Button and wizard created

2_button

**Case 1: users are not allowed to select custom criterias**
3_mass_sort_wizard

**Case 2: users have the right to select custom criterias**
3_mass_sort_wizard_custom
### Before

4_before
### Result

5_after


OSguard:
We did somethink simular https://github.com/OCA/sale-workflow/pull/241

legalsylvain:
Hi @OSguard.

Yes I remember, I reviewed it !
The purpose of this module is to be generic and so avoid to have to develop "purchase_reorder", "sale_reorder", "picking_reorder". 
Do you think that my module can cover your use case ? If yes, we could consider to accept this PR. If not, I can improve this PR with features you miss.

What I see for the time being : 

### mass_sorting
- can be used on any model : Yes (if the related one2many model contains sequence field); 
- can sort by any fields : yes ;
- can sort by any quantity of fields : yes ;
- manage asc / desc : Yes ;
- Is out of the box : No, required admin settings ;

### purchase_order_reorder_lines_wizard
- can be used on any model : only on purchase.order;
- can sort by any fields : no, by 'name', 'quantity', 'description' ;
- can sort by any quantity of fields : no, limited to 2 ;
- Is out of the box : Yes ;

What do you think ? Could you complete or correct, if I miss something about your module ?

Kind regard.

OSguard:
okay, i try your module in runbot. I like a generic solution.

I see two things that are not 'perfect' in your solution
- You need to make a pre-defined configuration, by the admin, the user has no choice at all
- For multiple configuration you have multiple action, so no selection in the popup.

Our initial idea was that the user can sort the lines in the view, but this is only in the browser and not saved. But this was not easy to fix, so we get up to the wizard where we add the fields the user see in the view.

legalsylvain:
Thanks for your answer. 

> You need to make a pre-defined configuration, by the admin, the user has no choice at all

You're right. I'll think about it. In fact : 
- pre-defined configuration -> interesting for user that sort always by the same fields, because the module will save a lot of click ;
- "free" configuration -> interesting to let users do what he want, without calling admin user;

Maybe I can improve my module in this way : 
-> let user select / changes the fields, but let the pre-defined configuration to save click;

> For multiple configuration you have multiple action, so no selection in the popup.

I don't know how to improve that point... In the other way, multiple action are interesting if after you want to restrict action access to a defined group. This is the same design (and problem) as mass_editing module.

legalsylvain:
Hi @OSguard, 

I added possibility to users to change / select others fields. (see printscreen updated in my first comment).
Can you check if this feature suits you ?

Kind regards.

OSguard:
I test it on runbot and i like it. Maybe add some help text in the popup.

mass_sorting/views/mass_sort_wizard_view.xml
```diff
@@ -0,0 +1,34 @@
...
+    <record id="view_mass_sort_wizard_form" model="ir.ui.view">
+        <field name="model">mass.sort.wizard</field>
+        <field name="arch" type="xml">
+            <form string="Mass Sorting" version="7.0">
+                <group col="4">
+                    <field name="description" nolabel="1" colspan="4" attrs="{'invisible': [('allow_custom_setting', '=', True)]}"/>
+                    <field name="line_ids" colspan="4" nolabel="1" attrs="{'invisible': [('allow_custom_setting', '=', False)]}">
+                        <tree editable="bottom">
+                            <field name="sequence" widget="handle"/>
+                            <field name="field_id"/>
```
OSguard, line 21: maybe use: `widget="selection"` here

legalsylvain, line 21: Good idea, but there is a bug in Odoo core. When I set such widget, Odoo loose domain. So I didn't commited your suggestion.

OSguard, line 21: Also if you add the domain again in the view?

legalsylvain, line 21: Ow ! I don't understand why change a widget impact the domain !

Well, I tried, 

```
domain="[('model', '=', one2many_model)]"
```

-> NameError: name 'one2many_model' is not defined

```
domain="[('model', '=', parent.one2many_model)]"
```

-> NameError: name 'parent' is not defined

(and other like     domain="[('model', '=', 'parent.one2many_model')]" with quotes but no items are displayed).

Do you have an idea how to do that ?

OSguard, line 21: https://bugs.launchpad.net/openobject-server/+bug/695089/comments/4
sorry there is no way. Odoo should remove it or fix it.
Maybe better to use web_m2x_options from OCA/web to make the dropdown a little nicer ;)

max3903:
:+1:

legalsylvain:
Hi, 
No second review since a while. Could we imagine optimistic merge on that old branch (7.0). I use this module in production without any problem since one year, and I really think that this feature is really generic and usefull for community.

kind regards.

pedrobaeza:
OK, merging.

PR #346, opened 2016-01-30
[FIX] flake8 (import in bad place);

legalsylvain:
Fix some "E402 module level import not at top of file"

### DIFF
--- a/server_monitoring/models/server_monitor_process.py
+++ b/server_monitoring/models/server_monitor_process.py
@@ -36,6 +36,7 @@ cleanup cron (2 different for db and process monitoring)
   RPC request log
 
 """
+from __future__ import absolute_import
 
 import logging
 import gc
--- a/server_monitoring/models/server_monitor_process.py
+++ b/server_monitoring/models/server_monitor_process.py
@@ -53,13 +53,6 @@ from openerp import pooler
 from openerp import SUPERUSER_ID
 from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 
-# ugly hack to avoid a WARNING message when importing stdlib resource module
-_logger = logging.getLogger('openerp.modules.module')
-_saved_log_level = _logger.getEffectiveLevel()
-_logger.setLevel(logging.ERROR)
-
-_logger.setLevel(_saved_log_level)
-
 _logger = logging.getLogger(__name__)
 
 BLACKLIST = (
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

moylop260:
:+1:

legalsylvain:
Hi, thanks for your review. I don't understand why runbot fail for this PR. Anybody has an idea ?

pedrobaeza:
There are 2 warnings in runbot:

```
2016-01-30 13:12:23,194 9423 WARNING 3140686-346-751d47-all root: Sentry DSN not defined in config file
2016-01-30 13:12:24,315 9423 WARNING 3140686-346-751d47-all openerp.modules.module: 
Ambiguous import: the OpenERP module `resource` is shadowed by another
module (available at /srv/openerp/instances/openerp-oca-runbot/sandbox/lib/python2.7/lib-dynload/resource.x86_64-linux-gnu.so).
To import it, use `import openerp.addons.<module>.`.
```

legalsylvain:
Oh. I see. the flake8 fix I wrote breaks runbot because of a "ugly hack" : 
https://github.com/OCA/server-tools/pull/346/files#diff-3d591500c249b214c340e905d5ae6290R47

Do you have an idea how to fix that ?
CC : @gurneyalex 

regards.

gurneyalex:
@legalsylvain I just found PEP328... 

Can you add this before the first import:

```
from __future__ import absolute_import
```

Then the `import resource` statement should no longer cause a warning.

StefanRijnhart:
You can now remove the ugly hack ;-)

server_monitoring/models/server_monitor_process.py
```diff
@@ -60,6 +53,13 @@
 from openerp import SUPERUSER_ID
 from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 
+# ugly hack to avoid a WARNING message when importing stdlib resource module
+_logger = logging.getLogger('openerp.modules.module')
+_saved_log_level = _logger.getEffectiveLevel()
+_logger.setLevel(logging.ERROR)
+
+_logger.setLevel(_saved_log_level)
```
dreispt: This is not doing anything, isn't it?

gurneyalex: yep it should be removed.

dreispt:
:+1:

sebalix:
:+1:

PR #347, opened 2016-02-01
[ADD] letsencrypt

hbrunn:
# 
# Request SSL certificates from letsencrypt.org

This module was written to have your Odoo installation request SSL certificates
from https://letsencrypt.org automatically.
# Installation

After installation, this module generates a private key for your account at
letsencrypt.org automatically in `$data_dir/letsencrypt/account.key`. If you
want or need to use your own account key, replace the file.

For certificate requests to work, your site needs to be accessible via plain
HTTP, see below for configuration examples in case you force your clients to
the SSL version.

After installation, trigger the cronjob `Update letsencrypt certificates` and
watch your log for messages.
# Configuration

This addons requests a certificate for the domain named in the configuration
parameter `web.base.url` - if this comes back as `localhost` or the like,
the module doesn't request anything.

If you want your certificate to contain multiple alternative names, just add
them as configuration parameters `letsencrypt.altname.N` with `N` starting
from `0`. The amount of domains that can be added are subject to `rate
limiting <https://community.letsencrypt.org/t/rate-limits-for-lets-encrypt/6769>`_.

Note that all those domains must be publicly reachable on port 80 via HTTP, and
they must have an entry for `.well-known/acme-challenge` pointing to your odoo
instance.
# Usage

The module sets up a cronjob that requests and renews certificates automatically.

After the first run, you'll find a file called `domain.crt` in
`$datadir/letsencrypt`, configure your SSL proxy to use this file as certificate.
# In depth configuration

This module uses `openssl` to generate CSRs suitable to be submitted to
letsencrypt.org. In order to do this, it copies `/etc/ssl/openssl.cnf` to a
temporary and adapts it according to its needs (currently, that's just adding a
`[SAN]` section if necessary). If you want the module to use another configuration
template, set config parameter `letsencrypt.openssl.cnf`.

After refreshing the certificate, the module attempts to run the content of
`letsencrypt.reload_command`, which is by default `sudo service nginx reload`.
Change this to match your server's configuration.

You'll also need a matching sudo configuration, like::

```
your_odoo_user ALL = NOPASSWD: /usr/sbin/service nginx reload
```

Further, if you force users to https, you'll need something like::

```
if ($scheme = "http") {
    set $redirect_https 1;
}
if ($request_uri ~ ^/.well-known/acme-challenge/) {
    set $redirect_https 0;
}
if ($redirect_https) {
    rewrite ^   https://$server_name$request_uri? permanent;
}
```


rafaelbn:
@antespi @yajo take a look here. We talked about it. Thanks

hbrunn:
BTW: https://therp.nl and all our other SSL sites use this in production

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+        return domain_key
+
+    @api.model
+    def validate_domain(self, domain):
+        if domain in ['localhost', '127.0.0.1'] or\
+                domain.startswith('10.') or\
+                domain.startswith('172.16.') or\
+                domain.startswith('192.168.'):
+            raise exceptions.Warning(
+                _("Let's encrypt doesn't work with private addresses!"))
```
sebalix: This whitelist is just to prevent common mistakes? From where are these addresses taken from? Internally we often use `172.24.*` for instance.
Maybe add a whitelist as a constant, , or use something like https://pypi.python.org/pypi/ipgetter (if it can help here).

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+            _logger.info('generating rsa domain key for %s', domain)
+            self.call_cmdline([
+                'openssl', 'genrsa', '-out', domain_key,
+                str(DEFAULT_KEY_LENGTH),
+            ])
+        return domain_key
+
+    @api.model
+    def validate_domain(self, domain):
+        if domain in ['localhost', '127.0.0.1'] or\
```
Garamotte: The loopback IP address can be any of 127.*, not only 127.0.0.1

letsencrypt/models/letsencrypt.py
hbrunn: This comes from https://en.wikipedia.org/wiki/Private_network without looking closely at the actual ip range. Will have to enumerate 16-31 then if we want to keep string comparisons.

And yes, that's meant to save people from themselves, or better, to avoid issues being posted about private addresses not working.

letsencrypt/models/letsencrypt.py
hbrunn: and ipv6 is not checked here at all, that also should happen

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class Letsencrypt(models.AbstractModel):
+    _name = 'letsencrypt'
+    _description = 'Abstract model providing functions for letsencrypt'
+
+    @api.model
+    def get_data_dir(self):
+        return os.path.join(_get_default_datadir(), 'letsencrypt')
```
sebalix: I'm testing this module currently, and I encounter a bug: `_get_default_datadir` returns the default data directory of Odoo (`~/.local/share/Odoo`), but does not honor the `data_dir` configuration option.
`tools.config.get('data_dir') or _get_default_datadir()` ?

hbrunn: sounds good, will do

hbrunn: (in case you change this anyways for your tests, a PR would be great)

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+                    '-reqexts', 'SAN',
+                ])
+            self.call_cmdline(cmdline)
+        return csr
+
+    @api.model
+    def cron(self):
+        domain = urlparse.urlparse(
+            self.env['ir.config_parameter'].get_param(
+                'web.base.url', 'localhost')).netloc
```
sebalix, line 139: If the `web.base.url` equals to `http://my.domain:8069`, the validation process produces this error:

```
  File "/home/odoo/odoo/server/openerp/http.py", line 537, in _handle_exception
  File "/home/odoo/odoo/server/openerp/http.py", line 574, in dispatch
  File "/home/odoo/odoo/server/openerp/http.py", line 310, in _call_function
  File "/home/odoo/odoo/server/openerp/service/model.py", line 118, in wrapper
  File "/home/odoo/odoo/server/openerp/http.py", line 307, in checked_call
    return self.endpoint(*a, **kw)
  File "/home/odoo/odoo/server/openerp/http.py", line 803, in __call__
  File "/home/odoo/odoo/server/openerp/http.py", line 403, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo/server/addons/web/controllers/main.py", line 948, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo/server/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/home/odoo/odoo/server/openerp/api.py", line 268, in wrapper
  File "/home/odoo/odoo/server/openerp/api.py", line 439, in old_api
    result = new_api(recs, *args, **kwargs)
  File "/home/odoo/odoo/server/openerp/api.py", line 443, in new_api
    result = [method(rec, *args, **kwargs) for rec in self]
  File "/home/odoo/odoo/addons_oca_server_tools/cron_run_manually/ir_cron.py", line 72, in run_manually
    return method(*args)
  File "/home/odoo/odoo/server/openerp/api.py", line 266, in wrapper
    return new_api(self, *args, **kwargs)
  File "/home/odoo/odoo/addons_oca_server_tools/letsencrypt/models/letsencrypt.py", line 141, in cron
    account_key, csr, acme_challenge, log=_logger, CA=DEFAULT_CA)
  File "/home/odoo/odoo/addons_oca_server_tools/letsencrypt/models/acme_tiny.py", line 105, in get_crt
    raise ValueError("Error requesting challenges: {0} {1}".format(code, result))
ValueError: Error requesting challenges: 400 {"type":"urn:acme:error:malformed","detail":"Error creating new authz :: Invalid character in DNS name","status":400}
```

Is it related to the presence of `:`? If so, maybe throws an explicit error message if the `ParseResult(...).netloc` != `ParseResult(...).hostname` (I don't know if it is the right fix) and invite the administrator to run on ports 80/443 with a correct `web.base.url`.

hbrunn, line 139: letsencrypt simply doesn't support port numbers in host names. You need a server listening on port 80. This should be added to https://github.com/OCA/server-tools/pull/347/files#diff-9b7e43b14857ae1b9294fe0125a6846fR73 I think

letsencrypt/controllers/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from . import letsencrypt
```
yajo: [This should be `main` instead.](https://www.odoo.com/documentation/8.0/reference/guidelines.html#file-naming)

letsencrypt/controllers/letsencrypt.py
```diff
@@ -0,0 +1,20 @@
...
+from openerp.http import request
+from openerp.tools.misc import file_open
+
+
+class Letsencrypt(http.Controller):
+    @http.route('/.well-known/acme-challenge/<filename>', auth='none')
+    def acme_challenge(self, filename):
+        try:
+            return file_open(
+                os.path.join('letsencrypt', 'static', 'acme-challenge',
```
yajo: Shouldn't this be inside `openerp.tools.config.filestore(request.env.dbname)`?

letsencrypt/controllers/letsencrypt.py
```diff
@@ -0,0 +1,20 @@
...
+from openerp import http
+from openerp.http import request
+from openerp.tools.misc import file_open
+
+
+class Letsencrypt(http.Controller):
+    @http.route('/.well-known/acme-challenge/<filename>', auth='none')
+    def acme_challenge(self, filename):
+        try:
+            return file_open(
```
yajo: Could this be inside a `with` block?

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
+# flake8: noqa
```
yajo: Remove this to let Travis do its job, please.

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
...
+#!/usr/bin/env python
```
yajo: python2

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
...
+import argparse, subprocess, json, os, sys, base64, binascii, time, hashlib, re, copy, textwrap, logging
+try:
+    from urllib.request import urlopen # Python 3
+except ImportError:
+    from urllib2 import urlopen # Python 2
+
+#DEFAULT_CA = "https://acme-staging.api.letsencrypt.org"
```
yajo: Remove this, it's useless.

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
...
+        "\n".join(textwrap.wrap(base64.b64encode(result).decode('utf8'), 64)))
+
+def main(argv):
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=textwrap.dedent("""\
+            This script automates the process of getting a signed TLS certificate from
+            Let's Encrypt using the ACME protocol. It will need to be run on your server
+            and have access to your private account key, so PLEASE READ THROUGH IT! It's
+            only ~200 lines, so it won't take long.
```
yajo: This last statement makes no sense to me. Since when sysadmins need to read the source of every script they run? They need to read docs, no sources...

Anyway, I'm afraid this file should not be inside `/models`. Maybe inside a `/bin` folder?

And what's most important... Are you sure you are not reinventing the wheel? [The official letsencrypt client is already a pip-packaged python app](https://pypi.python.org/pypi/letsencrypt), so it seems much easier and cleaner at first sight to just use it instead of writing this whole file... What do you think?

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
...
+    parser.add_argument("--account-key", required=True, help="path to your Let's Encrypt account private key")
+    parser.add_argument("--csr", required=True, help="path to your certificate signing request")
+    parser.add_argument("--acme-dir", required=True, help="path to the .well-known/acme-challenge/ directory")
+    parser.add_argument("--quiet", action="store_const", const=logging.ERROR, help="suppress output except for errors")
+    parser.add_argument("--ca", default=DEFAULT_CA, help="certificate authority, default is Let's Encrypt")
+
+    args = parser.parse_args(argv)
+    LOGGER.setLevel(args.quiet or LOGGER.level)
+    signed_crt = get_crt(args.account_key, args.csr, args.acme_dir, log=LOGGER, CA=args.ca)
+    sys.stdout.write(signed_crt)
```
yajo: In case this script is still needed, I'd thank a parameter for a write file, with `default=sys.STDOUT`

letsencrypt/models/acme_tiny.py
```diff
@@ -0,0 +1,199 @@
...
+
+if __name__ == "__main__": # pragma: no cover
+    main(sys.argv[1:])
```
yajo: This is the default, so if you just add `argv=None` to `main()`'s declaration, you save it for free.

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+        if not os.path.isdir(data_dir):
+            os.makedirs(data_dir)
+        account_key = os.path.join(data_dir, 'account.key')
+        if not os.path.isfile(account_key):
+            _logger.info('generating rsa account key')
+            self.call_cmdline([
+                'openssl', 'genrsa', '-out', account_key,
+                str(DEFAULT_KEY_LENGTH),
+            ])
+        assert os.path.isfile(account_key), 'failed to create rsa key'
```
yajo, line 62: Please do not use assert for production.

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+        account_key = self.generate_account_key()
+        csr = self.generate_csr(domain)
+        manifest, manifest_path = file_open(
+            'letsencrypt/__openerp__.py', pathinfo=True)
+        manifest.close()
+        acme_challenge = os.path.join(
+            os.path.dirname(manifest_path),
+            'static', 'acme-challenge')
+        if not os.path.isdir(acme_challenge):
+            os.makedirs(acme_challenge)
```
yajo, line 145: This chunk of code is quite weird.

Are you opening the `__openerp__.py` file just to get the module's path to create a folder inside it? :hushed:

First, you have the `__file__` magical variable to know your addon's dir, and second you should never write there. Addons should ideally be mounted in readonly folders. You have the filestore for writing, in a per-database basis.

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,157 @@
...
+                    'letsencrypt.chain_certificate_address',
+                    'https://letsencrypt.org/certs/'
+                    'lets-encrypt-x1-cross-signed.pem')
+            )
+            crt.write(chain_cert.read())
+            chain_cert.close()
+            _logger.info('wrote %s', crt.name)
+        self.call_cmdline([
+            'sh', '-c', self.env['ir.config_parameter'].get_param(
+                'letsencrypt.reload_command', ''),
```
yajo: Why the last `''`?

letsencrypt/models/acme_tiny.py
sebalix: `acme_tiny.py` is a lightweight client for Let's Encrypt (replacing the official one) which can be found here: https://github.com/diafygi/acme-tiny , it should not be changed.

hbrunn: exactly, that's why I think comments below are not relevant

letsencrypt/models/letsencrypt.py
hbrunn: to pass en empty string instead of `False` to sh in case the parameter doesn't exist

hbrunn:
@Yajo thanks for you review, fixed your points not related to the library

letsencrypt/models/letsencrypt.py
hbrunn, line 62: of course, what else?

yajo, line 62: if and raise. FTR https://docs.python.org/2.7/reference/simple_stmts.html#assert

hbrunn, line 62: to me, this is a clear case of can't happen in the list of places to consider putting assertions in https://wiki.python.org/moin/UsingAssertionsEffectively

letsencrypt/models/acme_tiny.py
yajo: Then the best would be that upstream merges https://github.com/diafygi/acme-tiny/pull/59 and we install this by pip

letsencrypt/models/letsencrypt.py
yajo, line 62: nice link, thanks :+1:

letsencrypt/models/letsencrypt.py
yajo: Opening an empty subprocess when you can simply skip it with an `if` or `try` makes little sense IMHO. Please change that.

letsencrypt/models/acme_tiny.py
lasley: IMO it's best to include this directly in the source because it is in beta, but also for security concerns of a direct pip install of something handling private keys.

letsencrypt/models/letsencrypt.py
lasley: I agree with @sebalix - We're going to get in trouble by excluding ~50% of the private class B networks. That said, I can't think of a better solution here other than converting the IP to INT, then determining between (eg 10/8 is 167772160 - 184549375)

letsencrypt/models/letsencrypt.py
```diff
@@ -0,0 +1,155 @@
...
+                domain.startswith('172.16.') or\
+                domain.startswith('192.168.'):
+            raise exceptions.Warning(
+                _("Let's encrypt doesn't work with private addresses!"))
+
+    @api.model
+    def generate_csr(self, domain):
+        domains = [domain]
+        i = 0
+        while self.env['ir.config_parameter'].get_param(
```
lasley, line 100: Would kill a lot of lines, and make slightly more readable if `self.env['ir.config_parameter'].get_param` was assigned as a var like `get_param`. Just a thought, not a blocker

lasley:
This is awesome, so glad you made @hbrunn - this is the next step in the era of SSL all the things! Comment or two inline but this is 👍  from me

letsencrypt/models/acme_tiny.py
yajo: [You can `pip install acme-tiny` now](https://github.com/diafygi/acme-tiny/pull/59#issuecomment-214826624), so please remove this file.

hbrunn:
@sebalix @Yajo thanks to @antespi, your comments have been addressed

letsencrypt/README.rst
```diff
@@ -22,6 +22,23 @@ the SSL version.
...
+For installing OpenSSL binary you can use you distro package manager. For Debian
+and Ubuntu would be:
+
+    sudo apt-get install openssl
+
+For installing ACME-Tiny python module you can use PIP package manager:
+
+    sudo pip install acme-tiny
+
+For installing ACME-Tiny python module you can use PIP package manager:
```
yajo: Copy-paste typo here.

letsencrypt/models/letsencrypt.py
```diff
@@ -8,26 +9,26 @@
...
     _name = 'letsencrypt'
     _description = 'Abstract model providing functions for letsencrypt'
 
     @api.model
-    def get_data_dir(self):
-        return os.path.join(_get_default_datadir(), 'letsencrypt')
-
-    @api.model
-    def get_challenge_dir(self):
-        return os.path.join(self.get_data_dir(), 'acme-challenge')
```
yajo: What's the point of getting this out of the model? Shouldn't there be a dir per database?

letsencrypt/models/letsencrypt.py
```diff
@@ -74,12 +75,20 @@ def generate_domain_key(self, domain):
 
     @api.model
     def validate_domain(self, domain):
-        if domain in ['localhost', '127.0.0.1'] or\
-                domain.startswith('10.') or\
-                domain.startswith('172.16.') or\
-                domain.startswith('192.168.'):
+        local_domains = ['localhost', 'localhost.localdomain']
```
yajo: A simple `cat /etc/hosts` reveals that you should add `localhost6.localdomain6` and `localhost6`, please.

letsencrypt/models/letsencrypt.py
```diff
@@ -149,7 +158,11 @@ def cron(self):
...
             _logger.info('wrote %s', crt.name)
-        self.call_cmdline([
-            'sh', '-c', self.env['ir.config_parameter'].get_param(
-                'letsencrypt.reload_command', ''),
-        ])
+        reload_cmd = self.env['ir.config_parameter'].get_param(
+            'letsencrypt.reload_command', False)
+        if reload_cmd:
+            _logger.info('reloading webserver...')
+            self.call_cmdline(['sh', '-c', reload_cmd])
```
yajo, line 168: Isn't it enough without the `sh -c` part? To allow this module to work outside Linux.

Besides, that `letsencrypt.reload_command` should be group-restricted to admin, to avoid any other user to read/write to it. It's a potential security flaw. It should be warned as such in the README too.

letsencrypt/models/letsencrypt.py
hbrunn: yes, that was my misunderstanding about my own code too. Here's the full story: https://github.com/hbrunn/server-tools/pull/2#discussion-diff-63349648

letsencrypt/models/letsencrypt.py
```diff
@@ -75,7 +75,10 @@ def generate_domain_key(self, domain):
 
     @api.model
     def validate_domain(self, domain):
-        local_domains = ['localhost', 'localhost.localdomain']
+        local_domains = [
```
yajo, line 78: Although it is OK and these are just 4 elements, I think it could be good to separate this list in another method. Also if you use a `set()` instead of a list, you get a performance boost for `in` operator. Not a blocker, just a tip you might find useful :wink:

letsencrypt/README.rst
```diff
@@ -0,0 +1,164 @@
...
+======================
+
+This module uses ``openssl`` to generate CSRs suitable to be submitted to
+letsencrypt.org. In order to do this, it copies ``/etc/ssl/openssl.cnf`` to a
+temporary and adapts it according to its needs (currently, that's just adding a
+``[SAN]`` section if necessary). If you want the module to use another configuration
+template, set config parameter ``letsencrypt.openssl.cnf``.
+
+After refreshing the certificate, the module attempts to run the content of
+``letsencrypt.reload_command``, which is by default ``sudo service nginx reload``.
```
yajo, line 84: I'm not sure I want a default value for that, and less one that starts by `sudo`. There are too many scenarios for not being agnostic in this IMHO. If you have to provide a default, I'd use the non-existing `letsencrypt-server-reload` command and tell the sysadmin to write an executable script in `/usr/local/bin/letsencrypt-server-reload` that implements the behavior they want.

I think that's the easiest way to automate any possible scenario. Every sysadmin knows how to write a 1-line script.

letsencrypt/models/letsencrypt.py
hbrunn, line 168: no, we need some shell to allow multiple commands there. Sh shells are very wide-spread, so I don't see a problem there.

good idea about the group, done

letsencrypt/README.rst
hbrunn, line 84: don't agree

yajo, line 84: any reasons?

letsencrypt/models/letsencrypt.py
yajo, line 168: if you need multiple commands, you can write `sh -c 'you-command'` in the config parameter already. And if you follow https://github.com/OCA/server-tools/pull/347/files#r63487176, you can write those multiple commands in the extra script, no matter the OS.

letsencrypt/README.rst
hbrunn, line 84: it's quite arbitrary what the default is, it will always work for some people and not for some other. But given nginx is the semi-standard for odoo installations, I don't want to invest any extra work and possible support requests for your proposal

yajo, line 84: Having the proxy in the same machine of Odoo is being deprecated in favor of 1-process-per-container solutions such as Docker, which should be the semi-standard from now on, and would work much better with my proposal.

Please don't take it badly, but "I don't want to work more" does not seem very convincing as you may guess. Mostly when the extra work would be nothing on your part because if you already have that setting in your production environments, it's under a noupdate tag, so it would not change; maybe your new instances would require an additional 1-line script, but you will thank me if any day you decide to migrate to Docker (or another proxy).

Please bear in mind we are a community here and everyone has his _how this should be done_ book. On such sysadmin-related modules, being as most system-agnostic as possible is very desirable, mostly when the effort is second to none.

.travis.yml
```diff
@@ -31,6 +31,11 @@ install:
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
+  - pip install python-ldap
```
moylop260: What about use `requirements.txt` file in root project path https://github.com/OCA/maintainer-quality-tools/pull/311?

letsencrypt/README.rst
hbrunn, line 84: I'll be happy to merge whatever you propose against this branch if it makes sense, but for stuff where there's no right and wrong as in this case, yes, even a split-second of my own time is too much.

.travis.yml
hbrunn: thanks, forgot about that, fixed

letsencrypt/README.rst
sebalix, line 84: @Yajo As long as we can change the `letsencrypt.reload_command` parameter to whatever we want, there is no problem IMO.

yajo, line 84: It just difficults automatic deployments in environments different to yours. Handcrafted systems give me rash, but nevermind, we can open PR later if needed :wink:

yajo:
Could you fix runbot?

hbrunn:
no, because runbot doesn't look at `requirements.txt` or `.travis.yml`

moylop260:
Thanks 👍

moylop260:
> no, because runbot doesn't look at `requirements.txt` or `.travis.yml`

Maybe we need use a [runbot travis2docker](https://github.com/Vauxoo/runbot-addons/tree/9.0/runbot_travis2docker) 😄

yajo:
@gurneyalex Can you fix runbot please? Otherwise all PR after this one will fail.

yajo:
:+1: BTW

antespi:
:+1: Thanks all for developing and reviewing. This is an awesome tool!

PR #349, opened 2016-02-03
Fix comparator equals when value is returned as a list.

clopez:
- At least with an OpenLDAP server, the values returned are always lists even when there is only one value. For example:

```
    {'cn': ['Carlos Lopez'],
     'employeeType': ['employee'],
     'givenName': ['Carlos'],
     'loginShell': ['/bin/bash'],
     }
```
- Therefore, using the operator 'equals' for 'employeeType' == 'employee' will fail because the value is inside a list.
- This patch checks if the value returned is a list and then picks the first field to do the comparision. If the value returned is not a list then it behaves as before.

### DIFF
--- a/users_ldap_groups/users_ldap_groups.py
+++ b/users_ldap_groups/users_ldap_groups.py
@@ -40,8 +40,15 @@ class LDAPOperator(models.AbstractModel):
             (value in ldap_entry[1][attribute])
 
     def equals(self, ldap_entry, attribute, value, ldap_config, company):
-        return attribute in ldap_entry[1] and \
-            unicode(value) == unicode(ldap_entry[1][attribute])
+        if attribute in ldap_entry[1]:
+            if isinstance(ldap_entry[1][attribute], list):
+                if len(ldap_entry[1][attribute]) > 1:
+                    return False
+                ldap_value = ldap_entry[1][attribute][0]
+            else:
+                ldap_value = ldap_entry[1][attribute]
+            return unicode(value) == unicode(ldap_value)
+        return False
 
     def query(self, ldap_entry, attribute, value, ldap_config, company):
         query_string = Template(value).safe_substitute(dict(


users_ldap_groups/users_ldap_groups.py
```diff
@@ -40,8 +40,13 @@ def contains(self, ldap_entry, attribute, value, ldap_config, company):
             (value in ldap_entry[1][attribute])
 
     def equals(self, ldap_entry, attribute, value, ldap_config, company):
-        return attribute in ldap_entry[1] and \
-            unicode(value) == unicode(ldap_entry[1][attribute])
+        if attribute in ldap_entry[1]:
+            if type(ldap_entry[1][attribute]) == type([]):
```
hbrunn: use `isinstance(ldap_entry[1][attribute], list)`

users_ldap_groups/users_ldap_groups.py
```diff
@@ -40,8 +40,13 @@ def contains(self, ldap_entry, attribute, value, ldap_config, company):
...
+                ldap_value = ldap_entry[1][attribute][0]
```
hbrunn, line 47: to follow the semantics of equality, I think you should return `False` here if `len(ldap_entry[1][attribute]) > 1`

clopez:
I have instead opted for converting the list to a string by joining the fields to do the comparision.

I'm also patching the compartor contains now because is broken also (behaves like equals) because it was checking for `'employee' in ['employee']`

hbrunn:
@clopez you first change is fine, but please revert the second. The contains operator is meant for exactly what it does: If a list value contains some value that is equal to the value in question

clopez:
So the contains value is not meant to allow comparing `"substring" in "string"` ???

clopez:
Ok. I have reverted the change on comparator equals and returning false if the list has more than one field. It makes sense if equals is meant to compare if a list value contains some value rather than comparing if string is inside substring like I was thinking before

hbrunn:
@clopez thanks, now provided the CLA :+1:

clopez:
and fixed a typo on the comparasion.

What is the CLA?

hbrunn:
@clopez the contributor license agreement which is mandatory to be signed in order to submit code to OCA repositories: http://odoo-community.org/page/website.cla

clopez:
Really? what a burden for such a small and trivial patch :(

hbrunn:
Well, that's how it is. If you don't want to sign it, give your code as a present to somebody who has signed it, then this person can submit the code (and reap all the fame and kudos deriving from it) with her own copyright

hbrunn:
@clopez but of course, this is a ome time thing, and assuming you'll propose more very useful patches, it should be worth the trouble

clopez:
Hi,

I don't plan to contribute more to odoo than this patch. So signing a CLA just for it is too much overhead.

Please feel free to take my patch, and commit it under your name reaping all the fame and kudos ;).

I hereby make the copyright on the patches attached to this PR public domain.

Regards

hbrunn:
@clopez thank, even though it's a pity. Anyways, I'll close this issue

PR #294, opened 2015-11-16
[8.0] add auth_password_settings from PR#134

hhgabelgaard:
This is an attempt to make a more "clean" commit history of PR#134
Unfortunately I don't know enough git to extract the relevant history from #134 (git git rebase -s recursive -X theirs origin/8.0 didn't do it for me) - I've "just" attributed the first commit to @dhp-denero - I hope that's acceptable

PR #162, opened 2015-05-04
add auth_password_settings module to enforce odoo user to set strong password

dhp-denero:
Hello Contributors,

As I was facing strange issue of rebase original branch, I removed my old cloned branch and create new branch. So now all well here. 
Please advice me if still some change required to merge it to original branch.

Thanks,

Dhaval Patel

PR #134, opened 2015-02-26
add auth_password_settings module

dhp-denero:

PR #350, opened 2016-02-05
[ADD] Add the module security_safe_password

tungocbui:
This module will help to make sure user password is strong enough by adding some constraints (at least 1 upper case, 1 lower case, 1 number, long enough). It also supports the check with cracklib package.


pedrobaeza:
This module looks very similar to https://github.com/OCA/server-tools/pull/162, which was proposed a long time ago. Maybe both can join forces to get an unified OCA module.

tungocbui:
@pedrobaeza: thks for pointing that pull request. I just took a quick look on that pull request and indeed the idea / purpose of both pull requests is almost the same. Do you have any idea on when the pull request [162](https://github.com/OCA/server-tools/pull/162) will be merged into OCA? On our side, we can try to contribute the check with "cracklib" package after that module is merged.

Please share with me if you have any other ideas on how we should help to have a unified OCA module.

Have a nice day!

pedrobaeza:
As far as I see, the PR got stopped because some comments were not attended by the author. I have commented in that PR if the author wants to give you the possibility to enhance that module or if he's going to do it himself.

hhgabelgaard:
#294 is my attempt to implement #134/#162 - But currently targeting v8.
@tungocbui Does your module support password reset from the login page? In v8 I need to override web_auth_reset_password in AuthSignupHome from auth_signup to get the validation messages displayed

tungocbui:
@hhgabelgaard: yes, it does. You will find that we overwrite the function "change_password" in the file "security_safe_password/controllers/main.py" to get the validation messages displayed.

tungocbui:
what should i do on this PR now? there is no feedback from #162 until now...

hhgabelgaard:
Good question - What is the next step?

It seems like a lot of people need/want this - We have the original PR #134, #162 and my #294 - all for v8 and now this new implementation for v9

How do we get #294 and #350 merged? What is missing? Should we make them more alike to ease upgrading?

dreispt:
@tungocbui @hhgabelgaard I see no point in having two different implementations on 9.0 (this #350) and 9.0 (#294).
We should create a version that takes the best from each implementation.
It would be nice to agree on which PR the work should be based. 
I suggest for authorship to be attributed to both.

hhgabelgaard:
#294 is for 8.0

But I agree, no point in having to functionally identicall modules

tungocbui:
@dreispt i totally agree with your point. So could you (or anyone in OCA) please take a look and suggest which module should be used as a base?

dreispt:
I just went through the code. The #294 seems like a richer implementation. #350 adds a cracklib check.
IMO we should work on #294, and then add the cracklib feature, as an option (only if the lib is installed) and then port it to 9.0 (should be straightforward AFAICS).

tungocbui:
Thks for your input. It's ok for me. So i will close this pull request? 

Let me know if i can help on any parts in #294.

dreispt:
Yes I think we should close this, if you don't mind.
It would be great if you could help with #294. First by doing a second code review, and when it's merged, by adding your cracklib feature and doing the port to 9.0.

tungocbui:
Ok for me. Please close it.

I will spend some time to take a look at #294 (but only next week)

PR #351, opened 2016-02-05
[8.0] Add sql export module (migration from v7)

florian-dacosta:


sql_export/sql_export.py
```diff
@@ -0,0 +1,142 @@
...
+    def export_sql_query(self):
+        for obj in self:
+            today = datetime.datetime.now()
+            today_tz = fields.Datetime.context_timestamp(
+                obj, today)
+            date = today_tz.strftime(DEFAULT_SERVER_DATETIME_FORMAT)
+            output = StringIO.StringIO()
+            query = "COPY (" + obj.query + ")  TO STDOUT WITH " + \
+                    obj.copy_options
+            self.env.cr.copy_expert(query, output)
```
StefanRijnhart: Call me paranoid, but I can imagine that there could be some methods of exploitation that none of us can think about, even with the blacklist of disallowed statement terms. Did you consider executing the query in a breakpoint that you rollback to afterwards?

florian-dacosta: I think there is no real risk because the query is executed with copy_expert and it does not accept 'update', 'drop', etc either.

But I agree with you, it can't hurt to be paranoid, I have added the rolllback statement.

Thank you for the review!

sql_export/tests/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+from . import test_sql_query
+checks = [
```
StefanRijnhart: in 8.0, you can drop the 'checks' declaration.

StefanRijnhart:
Looks good otherwise! I'd still be wary to grant any user the privilege to enter queries because they still might retrieve fetchmail passwords or other sensitive data, but I guess that is up to the admin user who to give access to that.

:+1:

florian-dacosta:
Not any user are able to enter queries. Only user of this group : 'group_sql_request_editor' are able to create/modify/delete queries.

sql_export/sql_export.py
```diff
@@ -0,0 +1,146 @@
...
+    @api.model
+    def check_query_syntax(self, vals):
+        if vals.get('query', False):
+            vals['query'] = vals['query'].strip()
+            if vals['query'][-1] == ';':
+                vals['query'] = vals['query'][:-1]
+            try:
+                self.env.cr.execute(vals['query'])
+            except:
+                self.env.cr.rollback()
```
lmignon: remove the call to rollback here.
the rollback is done in the 'finally' block.

sql_export/sql_export.py
```diff
@@ -0,0 +1,146 @@
...
+    def export_sql_query(self):
+        for obj in self:
+            today = datetime.datetime.now()
+            today_tz = fields.Datetime.context_timestamp(
+                obj, today)
+            date = today_tz.strftime(DEFAULT_SERVER_DATETIME_FORMAT)
+            output = StringIO.StringIO()
+            query = "COPY (" + obj.query + ")  TO STDOUT WITH " + \
+                    obj.copy_options
+            self.env.cr.execute("SAVEPOINT export_query")
```
lmignon: IMO it's safer to rollback to your savepoint in a `finally` block and use an random name for your savepoint;

``` python
name = 'export_query_%s' % uuid.uuid1().hex()
self.env.cr.execute("SAVEPOINT %s" % name)
try:
    self.env.cr.copy_expert(query, output)
    output.getvalue()
    new_output = base64.b64encode(output.getvalue())
    output.close()
finally:
    self.env.cr.execute("ROLLBACK TO SAVEPOINT %s" % name)
```

florian-dacosta: Ok, thank you for the advice, I adjust the code.

sebalix:
Tested, it works if there is no prohibited keywords in the SQL query. E.g. the following query is prohibited:

`SELECT * FROM my_table WHERE action='update';`

Maybe add a known issue in the README? (https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst#known-issues--roadmap)

Otherwise :+1:

florian-dacosta:
You are right, I just improved the README file!

StefanRijnhart:
You could consider dropping the blacklist because you now rollback anyway.

florian-dacosta:
Yes, but in case of prohibited query like "UPDATE res_partner SET ...". the copy_expert will fail and show an ugly error.

Because of the rollback, it is no big deal if the user create a prohibited query and execute it, but we still don't want him to do be able to do it...

sql_export/security/sql_export_security.xml
```diff
@@ -0,0 +1,21 @@
...
+    </record>
+
+    <record model="ir.rule" id="sql_export_restric_access_user_or_group">
+        <field name="name" >SQL Export users and groups rules</field>
+        <field name="model_id" ref="model_sql_export"/>
+        <field eval="1" name="perm_read"/>
+        <field eval="0" name="perm_create"/>
+        <field eval="0" name="perm_write"/>
+        <field eval="0" name="perm_unlink"/>
+        <field name="domain_force">['|', ('user_ids','=',user.id), ('group_ids','in', [x.id for x in user.groups_id])]</field>
```
moylop260, line 12: Is right the `*_ids` with `=`?
`('user_ids','=',user.id)`

moylop260:
Please, use [Standard-version-number](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers) in manifest file

sql_export/wizard/wizard_file.py
```diff
@@ -0,0 +1,115 @@
...
+                variable_dict[field.name] = self[field.name]
+        if "%(company_id)s" in sql_export.query:
+            variable_dict['company_id'] = self.env.user.company_id.id
+        if "%(user_id)s" in sql_export.query:
+            variable_dict['user_id'] = self._uid
+        format_query = self.env.cr.mogrify(sql_export.query, variable_dict)
+        query = "COPY (" + format_query + ")  TO STDOUT WITH " + \
+                sql_export.copy_options
+        name = 'export_query_%s' % uuid.uuid1().hex
+        self.env.cr.execute("SAVEPOINT %s" % name)
```
moylop260: What about use `with savepoint()`?
Check the `savepoint` section from [never-commit-the-transaction](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction)

sql_export/tests/test_sql_query.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_prohibited_queries_creation(self):
+        prohibited_queries = [
+            "upDaTe res_partner SET name = 'test' WHERE id = 1",
+            "DELETE FROM sql_export WHERE name = 'test';",
+            "  DELETE FROM sql_export WHERE name = 'test'   ;",
+            """DELETE
+            FROM
+            sql_export
+            WHERE name = 'test'
```
moylop260, line 51: Could you add the test with sql-injection?
`SELECT id FROM sql_export;DELETE FROM sql_export`

sql_export/security/sql_export_security.xml
StefanRijnhart, line 12: Yes, this should work fine.

sql_export/tests/test_sql_query.py
florian-dacosta: Done, with some delay...

StefanRijnhart:
Hurray! Now let's get this merged. :+1: again.

legalsylvain:
Please, wait, I'd like to refactor it for more modularity. (I'm currently creating an other module, that needs some of this feature). I propose a PR against this PR during this week. Thanks.

legalsylvain:
Refactoring done here : review welcome.
https://github.com/akretion/server-tools/pull/13

legalsylvain:
Hi all, 
I finished the refactoring. (available for review [here](https://github.com/OCA/server-tools/pull/351/commits/120fe2063f3324090241d99742a16d7ca1ea83ca))
In a nutshell, I moved a lot of code in a new module sql_request_abstract to share functionnalities with a new module in reporting-engine *bi_sql_editor*, that allow users to create graph view, based on a normal / materialized view, written in SQL (PR available [here](https://github.com/OCA/reporting-engine/pull/102), description available [here](https://github.com/grap/reporting-engine/tree/8.0_ADD_bi_sql_editor/bi_sql_editor))

@StefanRijnhart, @sebalix , @moylop260 : agree to keep your :+1: ?

thanks.

legalsylvain:
Ping @ all + @mourad-ehm, could you review and cherry pick in your 9.0 work ? 
thanks !

mourad-ehm:
Thanks @legalsylvain for the good work. 
I have just a proposition to improve the use of parameters in sql query. 
Now, we should create a fields for each parameter on the sql.file.wizard object.
I think that use a generic JSON field, is more suitable than a regular fields. 
The idea is to create a text field named sql_params on sql.file.wizard object. The field allow to store params values to use it on sql query {'param_1" : val 1, 'param_2' : val2 , ...). The param names can be extracted form sql query. To display the values on wizard, we can overwrite methods, read, write and field_view_get to transform a json to a "odoo view". 
Otherwise :+1:

legalsylvain:
Hi,
I talked with @mourad-ehm. the remark is more on a road map. So I propose to merge this PR, to allow him to get the refactoring in this PR. (https://github.com/OCA/server-tools/pull/535)

thanks.

PR #352, opened 2016-02-05
[FIX] keep _register_hook's signature and call super

eLBati:


hbrunn:
:+1:

base_optional_quick_create/model.py
```diff
@@ -50,14 +48,18 @@ def _register_hook(self, cr, ids=None):
...
 
+    def _register_hook(self, cr):
+        self._patch_quick_create(cr, self.search(cr, SUPERUSER_ID, []))
+        return super(IrModel, self)._register_hook(cr)
+
     def create(self, cr, uid, vals, context=None):
-        res_id = super(ir_model, self).create(cr, uid, vals, context=context)
-        self._register_hook(cr, [res_id])
+        res_id = super(IrModel, self).create(cr, uid, vals, context=context)
+        self._patch_quick_create(cr, [res_id])
```
yajo, line 57: Does it make any sense to patch the creation method after the record has been created?

base_optional_quick_create/model.py
```diff
@@ -50,14 +48,18 @@ def _register_hook(self, cr, ids=None):
...
         return res_id
 
     def write(self, cr, uid, ids, vals, context=None):
         if isinstance(ids, (int, long)):
             ids = [ids]
-        res = super(ir_model, self).write(cr, uid, ids, vals, context=context)
-        self._register_hook(cr, ids)
+        res = super(IrModel, self).write(cr, uid, ids, vals, context=context)
+        self._patch_quick_create(cr, ids)
```
yajo, line 64: What's the relation between patching `name_create` and `write()`?

base_optional_quick_create/model.py
```diff
@@ -50,14 +48,18 @@ def _register_hook(self, cr, ids=None):
                     model_obj.check_quick_create = True
         return True
 
+    def _register_hook(self, cr):
```
yajo, line 51: Shouldn't you use `_patch_method()` instead?

eLBati, line 51: Do you have a suggestion about how to do it in this case? (Where `_register_hook` has 2 parameters and `_patch_quick_create` 3)

base_optional_quick_create/model.py
StefanRijnhart, line 57: That is on a different level. After the creation of the model, we patch the model's name_create method which applies when records of that model are being created through the quick create option.

base_optional_quick_create/model.py
StefanRijnhart, line 64: name_create should be patched when the model has avoid_quick_create set. Strictly speaking, patching should only take place when this field is being written.

base_optional_quick_create/model.py
eLBati, line 57: I am not sure to understand.
Consider that this is the `create` of a new model. That is: when a new model is created, patch its `name_create` method

Edit: I did not see @StefanRijnhart answer :smirk:

base_optional_quick_create/model.py
eLBati, line 64: See https://github.com/OCA/server-tools/pull/352/files#r52468995

base_optional_quick_create/model.py
StefanRijnhart, line 51: I think @Yajo means to replace the contents of _patch_quick_create with a call to _patch_method. It basically does the same thing. You'd still trigger _patch_quick_create from _register_hook.

StefanRijnhart, line 51: You can check base_action_rule for an example of how to use _patch_method.

hbrunn:
I'm working on a mixin that simplifies monkey patching and encourages good practices, you might just want to use that: https://github.com/hbrunn/social/blob/8.0-mail_follower_custom_notification/mail_follower_custom_notification/models/base_patch_models_mixin.py - after sufficient testing and review, I'm planning to release the mixin as a module in server-tools. Note: This is development code, use at you own risk

base_optional_quick_create/model.py
yajo, line 51: Yes I meant that. It seems the standard way to do it.

base_optional_quick_create/model.py
yajo, line 57: Still it does not seem a good trigger point to do it. If you quick-create a record before slow-creating one, it would still enable it.

The logical flow IMHO should be like:

```
_inherit = 'ir.model'
@api.cr
def _register_hook(self):
  self.search([("avoid_quick_create", "=", True)])._patch_quick_create()
  return super(...)._register_hook()

@api.multi
def _patch_quick_create(self):
  @api.model
  def new_name_create_method(self):
    ...
  for s in self:
    self.env[s.model]._patch_method("name_create", new_name_create)
```

And add a constraint to `avoid_quick_create` column that patches or restores the method depending on what changed.

eLBati, line 57: > Still it does not seem a good trigger point to do it. If you quick-create a record before slow-creating one, it would still enable it

Why? I don't think this happens.

Consider these are the `create` and `write` methods of `ir.model` model, not of the model itself.

For example: when the `account.invoice` (record of `ir.model`) is modified, setting `avoid_quick_create` = True, patch `name_create` of `account.invoice` model

base_optional_quick_create/model.py
eLBati, line 51: Ok, thanks. I modified it

base_optional_quick_create/model.py
yajo, line 57: I hope you get my point with https://github.com/eLBati/server-tools/pull/3. I did not test it, just showing what I mean.

StefanRijnhart:
:+1: tested and code review. Maybe best to leave refactoring using @hbrunn's method once it's ready.

PR #353, opened 2016-02-06
add treeview2csv

tonygalmiche:
My first module
Thanks for your help


pedrobaeza:
Thanks for the contribution, but I think this feature is already present in https://github.com/OCA/web/tree/8.0/web_export_view (although to XLS). And if not, you can propose an improvement over that one.

tonygalmiche:
indeed

I did not know this module

thank you for the answer

it will be for another time

InfoSaône http://www.infosaone.com/
Tél. 03 80 47 93 81 / Port. 06 19 43 39 31
Mise en place d'outils de travail collaboratif / Intégrateur ERP odoo 
(OpenERP) et GED Dynacase

PR #354, opened 2016-02-10
[8.0][html_image_url_extractor] Image extractor from HTML fields.

yajo:
This technical utility allows the developer to get a list of image URLs from any piece of HTML.

You can use it for example, to get the cover image from a blog post (https://github.com/OCA/website/pull/163), or to create a slider with all images from it.


html_imgs/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [
+            "lxml.html",
+        ],
+    },
+    "depends": [
+        "base",
```
elicoidal: base or website?

yajo: It makes sense to use it mostly in website, but there's no real dependency, so I'm not adding it.

yajo:
Could somebody please tag this as "needs review"?

pedrobaeza:
Done

sebalix:
Would there be an interest to have just one module (#354 + #355) for all these kind of utilities? They both add functions to `ir.fields.converter`.

yajo:
The `ir.fields.converter` is chosen just because the name somehow fits the purpose of these modules, but they really don't have much to do with the rest of that model's logic, and don't seem very related among them. But let's hear what others think.

sebalix:
Yes, I was just wondering about the future if we create one module for one function. But it is probably easier to maintain/migrate like that.

pedrobaeza:
Call the module _html_image_url_extractor_.

yajo:
I already have some modules that depend on this one in production, and some PR open linked to it. If you don't mind, please let me keep the name to avoid much extra work on that. Besides, it matches the style of the just merged #355, and will be combined with it in most cases.

pedrobaeza:
Sorry, but the name include plurals and it doesn't reflect the purpose of the module. Another name can be suggested, but not the original one with plural. Although now I see that the one you mention in the PR is not the most suitable, at least it doesn't contain plural.

It's very easy to change the name in your installation with openupgradelib. There's a direct method for doing it. I do it via the shell.

yajo:
I decided to include plurals this time because it provides a generator, which must always be iterated.

Anyways I'm OK with what you say. Just keep in mind this will take _much longer_ than a simple `mv && git push`.

yajo:
Module name fixed. Rebased, squashed.

html_image_url_extractor/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================
+Images from HTML field
+======================
+
+This module extracts image URLs from any chunk of HTML, in appearing order.
```
pedrobaeza: This module includes a method that extracts...

html_image_url_extractor/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+======================
+Images from HTML field
```
pedrobaeza: Images URLs from...

pedrobaeza:
Little comments, but :+1:

yajo:
All fixed.

carlosdauden:
:+1:

rafaelbn:
:+1: we are already using this module

rafaelbn:
cc @antespi

Issue #339, opened 2016-01-26
Odoo runs much slower with auditlog module installed


wim-a:
I installed the auditlog module in our Odoo 8 application.

Since this module is installed, the Odoo runs much slower. One of my own developed processes runs less than 2 minutes when the auditlog is uninstalled, but with the auditlog installed it takes about 90 minutes.

I have rules installed for 30 models (f.e. partner, product, product variant, accounts, invoices, journal entries and items, purchase orders, manufacturing orders), and for each of these modules, I log writes, deletes, and creates. I never log to reads.

sebalix:
Hi Wim,

If you do not need to log everything starting from a specific method, you can control the log creation with the context key `auditlog_disabled`. Auditlog itself uses it to not log its own operations (https://github.com/OCA/server-tools/blob/8.0/auditlog/models/rule.py#L213).

Another feature I would like to add is to choose the log level for each rule:
- Full log: make a snapshot of the data before/after the CRUD operation and logs the difference, as it is now. It is slow, but even computed fields are logged if they change.
- Fast log: just log the `vals` dictionary for `write` and `create` operations. There is no snapshot of the data before/after, so it should be very fast.

wim-a:
Thanks, Sébastien, I'll change my own module for auditlog_disabled. 

I like fast log feature (certainly for some of the models), would it be a big change to the existing module?

About the full log and computed fields, do you mean that also not-stored computed fields are logged?

hbrunn:
@wim-a technically, he does a `read()` on all records twice, so yes, if you have expensive computed fields, they are probably the culprit. If you have control over the code and are not interested in the computed fields anyways during logging, check for the context key @sebalix mentions in your function and do nothing then.

Make sure you use a version with https://github.com/OCA/server-tools/pull/301 (merged to main branch three weeks ago) merged, this eases the pain already a lot in some cases.

The fast log sounds like a very good idea, I don't think that costs more than a couple of hours to implement.

wim-a:
Thanks for the info, Holger. How can we proceed for having the fast log feature available?

hbrunn:
Ask your favorite Odoo programmer to quote this new feature for you, and let them do it if you like their offer

wim-a:
Sébastien, Holger,

I've modified the code of the auditlog module on my server for fast logging.

In this version, it works as follows:
- The auditlog_rule table has an additional field action_rule in which the system administrator can define if this model is full or fast logged.
[...]

PR #357, opened 2016-02-11
[IMP] auditlog - Full log / Fast log

sebalix:
Ability to choose the log type on the rule: Full log (existing one, complete but slow) and Fast log (data input only, faster).

A first implementation was done by @wim-a, available here: https://github.com/osiell/server-tools/commit/1f2152cf5c7b4c6533c96cb1e2f3972a9eccaf01 (see #339)

I rewritten the feature as this implementation was still making `read` operations. Now the _Fast log_ type doesn't trigger `read` or `name_get()` calls anymore.
As a side effect, the logged data is different, e.g. changing a 2many field will log `[(6, 0, [1])]` in the _Fast log_ type and `[(1, u'Name')]` in _Full log_.

Some benchmarks: updating the `category_id` field on 200 partners takes ~17s in _Full log_, and ~3s in _Fast log_ (5x faster).

### DIFF
--- a/auditlog/models/log.py
+++ b/auditlog/models/log.py
@@ -40,7 +40,7 @@ class AuditlogLog(models.Model):
         'auditlog.http.session', string=u"Session")
     http_request_id = fields.Many2one(
         'auditlog.http.request', string=u"HTTP Request")
-    type = fields.Selection(
+    log_type = fields.Selection(
         [('full', u"Full log"),
          ('fast', u"Fast log"),
          ],
--- a/auditlog/models/log.py
+++ b/auditlog/models/log.py
@@ -40,6 +40,11 @@ class AuditlogLog(models.Model):
         'auditlog.http.session', string=u"Session")
     http_request_id = fields.Many2one(
         'auditlog.http.request', string=u"HTTP Request")
+    type = fields.Selection(
+        [('full', u"Full log"),
+         ('fast', u"Fast log"),
+         ],
+        string=u"Type")
 
 
 class AuditlogLogLine(models.Model):
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -88,6 +88,16 @@ class AuditlogRule(models.Model):
         u"Log Creates", default=True,
         help=(u"Select this if you want to keep track of creation on any "
               u"record of the model of this rule"))
+    log_type = fields.Selection(
+        [('full', u"Full log"),
+         ('fast', u"Fast log"),
+         ],
+        string=u"Type", required=True, default='full',
+        help=(u"Full log: make a diff between the data before and after "
+              u"the operation (log more info like computed fields which were "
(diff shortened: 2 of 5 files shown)


StefanRijnhart:
Nice! :+1:

auditlog/models/log.py
```diff
@@ -40,6 +40,11 @@ class AuditlogLog(models.Model):
         'auditlog.http.session', string=u"Session")
     http_request_id = fields.Many2one(
         'auditlog.http.request', string=u"HTTP Request")
+    type = fields.Selection(
```
sebalix: I'm wondering: maybe I should change this name as it is a reserved keyword... s/type/log_type/

pedrobaeza: Yes, you should change it to `log_type` for example.

sebalix:
I'm done with this PR.

auditlog/models/log.py
```diff
@@ -40,6 +40,11 @@ class AuditlogLog(models.Model):
         'auditlog.http.session', string=u"Session")
     http_request_id = fields.Many2one(
         'auditlog.http.request', string=u"HTTP Request")
+    log_type = fields.Selection(
+        [('full', u"Full log"),
+         ('fast', u"Fast log"),
+         ],
+        string=u"Type")
```
dreispt: It would be nice to add a `help` text with the explanation you gave for the PR.

auditlog/models/rule.py
```diff
@@ -261,42 +292,79 @@ def read(self, *args, **kwargs):
...
+        @api.multi
+        def write_fast(self, vals, **kwargs):
+            self = self.with_context(auditlog_disabled=True)
+            rule_model = self.env['auditlog.rule']
+            # Log the user input only, no matter if the `vals` is updated
+            # afterwards as it could not represent the real state
+            # of the data in the database
+            vals2 = dict(vals)
+            old_vals2 = dict.fromkeys(vals2.keys(), False)
+            old_values = dict((id_, old_vals2) for id_ in self.ids)
```
dreispt, line 330: Do we really need these two lines? AFAICS they produce a ids dict with a fields dict, with empty values.

dreispt:
:+1:

auditlog/models/rule.py
sebalix, line 330: Yes we need this dictionary with empty values to generate a result when we call the `diff.changed()` in the `create_logs()` method. Another way would be to pass `None` instead of the `old_values` empty dict and to write `diff.changed() or diff.added()` to be compatible with both _Full_ and _Fast_ logging, but it was not clear enough for me to do that way, and I would like to avoid possible _side effects_ with such expression.

sebalix:
I forgot to increment the version number.

sebalix:
Now it is ready :)

hbrunn:
:+1:

PR #361, opened 2016-02-15
[SEC] fetchmail_attach_from_folder: fix unsafe eval

gurneyalex:


fetchmail_attach_from_folder/model/fetchmail_server.py
```diff
@@ -253,7 +253,7 @@ def fields_view_get(self, cr, user, view_id=None, view_type='form',
                 if field.tag == 'field' and field.get('name') in modifiers:
                     field.set('modifiers', simplejson.dumps(
                         dict(
-                            eval(field.attrib['modifiers'],
+                            safe_eval(field.attrib['modifiers'],
```
sebalix, line 257: I think it needs to be imported?

antespi:
:+1:

fetchmail_attach_from_folder/model/fetchmail_server.py
gurneyalex, line 257: @sebalix good catch. It's fixed now.

sebalix, line 257: @gurneyalex it is strange that the previous build did not notice it, it was all green https://travis-ci.org/OCA/server-tools/builds/109303252

sebalix:
:+1:

fetchmail_attach_from_folder/model/fetchmail_server.py
gurneyalex, line 257: @sebalix the module is not ported to v8, so travis does not care :smile_cat:

hbrunn:
:+1:

PR #368, opened 2016-02-16
[WIP] Add base_stage_state

dreispt:


lasley:
@dreispt We created a few similar things for this in vertical-medical, then started abstracting into something similar to what you have here.

Seems like a good idea to base off of this. Am I correct that these test failures are outside the scope of this PR?

dreispt:
Please use this. I'm using it production, but this PR might need some love.
I'll try to look into it.

dreispt:
@lasley Correction: I'm using this concept in production, but this this particular module, that takes the abstraction a little further.
I did some fixes, will try to get a green runbot, but the configuration menu items for the is missing. I need to add them

But you're welcome to play with it in the current form a see if you find anything that should be fixed.

dreispt:
It's green now, although it has no UI and no tests.

base_stage_state/views/stage_state.xml
```diff
@@ -0,0 +1,386 @@
+<?xml version="1.0"?>
```
lasley, line 1: Is this just a placeholder for the planned UI? Looks like a direct copy of project KanBan?

base_stage_state/models/base_stage_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+        'Kanban State',
+        track_visibility='onchange',
+        default='normal',
+        copy=False,
+        help="The kanban state indicates the workflow readiness:\n"
+             " * Normal is the default situation\n"
+             " * Blocked indicates something is preventing progress\n"
+             " * Ready for next stage indicates it is ready"
+             " to be pulled to the next stage")
+    color = fields.Integer('Color Index')
```
lasley, line 41: Did we maybe want to add in the KanBan legend fields as well? `legend_blocked`, etc.

base_stage_state/models/base_stage.py
```diff
@@ -0,0 +1,52 @@
...
+from openerp import models, fields, api
+
+
+class BaseStageSet(models.Model):
+    """
+    Organize Stages in Sets
+    """
+    _name = 'base.stage.set'
+    _description = 'Base Stage Set'
+    name = fields.Char(translate=True)
```
lasley, line 14: Maybe a unique validation to avoid naming conflicts between the stage groups? Totally cosmetic, but seems appropriate IMO

base_stage_state/models/base_stage.py
```diff
@@ -0,0 +1,52 @@
...
+        return [
+            ('draft', 'New'),
+            ('open', 'In Progress'),
+            ('pending', 'Pending'),
+            ('done', 'Done'),
+            ('cancelled', 'Cancelled')]
+
+    @api.model
+    def get_default_stage(self, model):
+        domain = [('model_id', '=', model), ('fold', '=', False)]
```
lasley, line 34: Does this work? I don't see a `model_id` field declaration in this model.

lasley:
Sweet, thanks for the update @dreispt - this is solid, and can replace a lot of code I wrote in Vertical Medical.

Curious - what UI is being planned here? An abstract KanBan view?

base_stage_state/models/base_stage_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @api.model
+    def _get_default_stage(self):
+        """ Gives default stage_id """
+        return self.env['base.stages'].get_default_stage(self._name)
+
+    stage_id = fields.Many2one(
+        'base.stage', 'Stage',
+        track_visibility='always', index=True,
+        domain="[('model_name', '=', self._name)]",
```
lasley, line 21: I do not see a field declaration for `model_name` in `base.stage`

base_stage_state/views/stage_state.xml
dreispt, line 1: Yes, the UI part is a WIP. I'm working on it.

base_stage_state/models/base_stage_mixin.py
dreispt, line 41: That was introduced only on 9.0, didn't check if it is easy to replicate it in 8.0.
It's a good idea, but I would leave it out of a first release. I can add it to the roadmap list though.

base_stage_state/models/base_stage.py
dreispt, line 14: Nice to have.

base_stage_state/models/base_stage.py
dreispt, line 34: You're probably right, I'll look at that.

base_stage_state/models/base_stage_mixin.py
dreispt, line 21: Will check that.

dreispt:
Well, Odoo does not have "abstract views", you will have to add the view parts yourself.
But the module could provide some sample XML for a quick start.
The UI provided will be in the Settings menu, for the Stage Sets and Stages configuration tables.

hbrunn:
@dreispt actually, we can do something like an abstract view: Just define some view on your abstract model, and tell users to override it with view mode `primary` for their own model

base_stage_state/models/base_stage_mixin.py
lasley, line 41: Ohhhh cool I didn't know it was a 9.0 thing, thanks for that clarification. I was just comparing with a vertical medical module I made for the same thing. Roadmap sounds like a good plan; I doubt it's an easy back port.

Not sure if it was clear, but I will take on the upgrade to 9.0 if you would like. This is most definitely something I need :)

lasley:
@dreispt - We just merged a nearly identical module in to our vertical-medical incubator, and have it up for [PR in OCA](https://github.com/OCA/vertical-medical/pull/121). I think it would be better to continue under the base module here instead though. 

Would you be opposed to us combining the two modules and superseding both PRs? It would be for version 9 though.

dreispt:
Not at all. I won't be able to work on this soon.
Please propose your module here,. and add any idea you find interesting here.
Please ping me on the PR and I'll review it.

lasley:
Awesome thanks @dreispt - @obulkin will be submitting the PR in the next few days 😄

lasley:
@dreispt - OCA/web#490 created as proposal for superseding this PR. Would you mind double checking that we got everything in there that you did here, then closing this one? I could technically close, but I think it would be better to leave both open during confirmation - correct me if I'm under the wrong thought process

PR #369, opened 2016-02-18
[8.0][WIP][migratk] Migration tools collection.

yajo:
This module provides an object-oriented toolkit to perform migrations easily.

Currently WIP, as I'm using it in some migrations to check for errors, but I decided to release it now to get some feedback.

All logic comes from an old toolkit, so all untested methods are decorated with `@wip`. Please don't be too inquisitive while reviewing those, they are just a proof of concept. OTOH, the methods that don't have it should be ready.

I know OCA already has openupgradelib, but this follows a slightly different design, mostly focused in:
- Rollback all if something fails.
- OOP.
- Module-version upgrades (not necessarily Odoo-version upgrades, for which probably openupgrade will be better).

It seems useful to have this as a module. I'll send an example migration so you can see how this works.

@antespi


pedrobaeza:
I don't think we should have 2 tools in OCA for the same task. openupgradelib also serves for minor versions migrations (that's why now it's independent and put in pip).

About rollback, it's already done in Odoo core/OpenUpgrade, so I also don't see the need.

For me :-1:

yajo:
You have an example in https://github.com/OCA/social/pull/45

yajo:
> About rollback, it's already done in Odoo core/OpenUpgrade, so I also don't see the need.

AFAIK this is not true. Please correct me if I'm wrong, but recently I wrote some migrations and found that if I run any migration outside a `with cr.savepoint()` and something fails, then the database is stuck in a broken status. The next time the update is run, it's unpredictable. You need to fix your script, delete that database, restore it and run it again. A total loss of time, as explained in https://github.com/OCA/openupgradelib/pull/30 that now is automatically avoided. Just fix, reboot Odoo and see if it works now.

Another advantage above openupgradelib is that it is object-oriented, and that here you have partial access to `self.env`. Those things allow for more familiar, readable and error-prone code in migrations.

Please don't get me wrong, I'm not saying openupgradelib is wrong or that this _must_ be merged, it's just that I already had this job done and it works quite well, so I'm just defending it a little bit :sweat_smile:

pedrobaeza:
Maybe the mechanism has some lacks, but the point is to improve it instead of inventing another one.

yajo:
Do you think we could invite somebody from there to do a quick review to this and https://github.com/OCA/social/issues/45, just to see if this could be interesting for openupgradelib?

pedrobaeza:
Maybe @StefanRijnhart or @hbrunn can express their opinion, as they also participate actively on OpenUpgrade.

StefanRijnhart:
Yes, this should go into openupgradelib. Also see my revised review of https://github.com/OCA/openupgradelib/pull/30

yajo:
Thanks, let me post it there and continue discussion in the right place.

PR #374, opened 2016-02-24
[8.0][base_multi_image] Base module for models with multiple images (a gallery)

yajo:
Following what was discussed in https://github.com/OCA/product-attribute/pull/112, this is the base part of the module stack, without website dependencies.

This module allows other submodules to implement multiple images in very few steps.

@rafaelbn


rafaelbn:
Thanks
cc @pedrobaeza @antespi @sergio-incaser @carlos-incaser

carlosdauden:
It's easy to refresh the image to make it look when selected?
For the rest ok :+1:

rafaelbn:
Please @pedrobaeza @antespi @dreispt  could you take a look here?

base_multi_image/README.rst
```diff
@@ -0,0 +1,91 @@
...
+
+To develop a module based on this one:
+
+* See module ``product_multi_image`` as an example.
+* You have to inherit model ``base_multi_image.owner`` to the model that needs
+  the gallery::
+
+    class MyOwner(models.Model):
+        _name = "mymodule.name"
+        _inherit = "base_multi_image.owner"
```
pedrobaeza: The most common example will inherit from an existing model and this abstract one, so I think it's better to put it here:

```
class ...:
    _name = "name"
    _inherit = ["base_multi_image.owner", "base_model"]
```

base_multi_image/README.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Original implementation
+-----------------------
+This module is inspired in previous module *product_images* from OpenLabs
+and Akretion.
+
+Contributors
+------------
+
+* Sharoon Thomas
```
pedrobaeza: Why putting this name? The original code has nothing to do with this, only the idea of "having multiple images".

base_multi_image/models/owner.py
```diff
@@ -0,0 +1,101 @@
...
+    image_main_small = fields.Binary(
+        string="Small image",
+        compute="_get_multi_image",
+        inverse="_set_multi_image_main_small",
+        store=False)
+
+    @api.multi
+    @api.depends('image_ids')
+    def _get_multi_image(self):
+        """Get a the main image for this object.
```
pedrobaeza: remove the middle "a"

base_multi_image/models/owner.py
```diff
@@ -0,0 +1,101 @@
...
+
+        This is provided as a compatibility layer for submodels that already
+        had one image per record.
+        """
+        for s in self:
+            s.image_main = False
```
pedrobaeza: You don't need to set to False. ORM now makes this automatically (not at the beginning of the new API, though)

base_multi_image/models/owner.py
```diff
@@ -0,0 +1,101 @@
...
+            s.image_main_medium = False
+            s.image_main_small = False
+            if s.image_ids:
+                s.image_main = s.image_ids[0].image_main
```
pedrobaeza: If you put `[:1]`, you don't have to put the `if` sentence

andhit-r:
LGTM :+1:

base_multi_image/README.rst
dreispt: +1

base_multi_image/README.rst
```diff
@@ -0,0 +1,91 @@
...
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Rafael Blasco <rafabn@antiun.com>
+* Jairo Llopis <yajo.sk8@gmail.com>
```
dreispt, line 94: No one from Akretion?

base_multi_image/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © 2009 Sharoon Thomas Open Labs Business Solutions
+# © 2015 Antiun Ingeniería S.L. - Jairo Llopis
```
dreispt, line 2: I notice that the authors here differ from the `__openerp__.py`, both in the headers and in the authors key...

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+import urllib
+import os
+import logging
+from openerp import models, fields, api, exceptions, _
+from openerp import tools
+
+_logger = logging.getLogger(__name__)
+
+
+class ImageABC(models.Model):
```
dreispt: (nitpick alert) This is not really an Abstract base Class, is it?

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+        required=True)
+    owner_model = fields.Char(
+        required=True)
+    storage = fields.Selection(
+        [('url', 'URL'), ('file', 'OS file'), ('db', 'Database')],
+        required=True,
+        default='db')
+    name = fields.Char(
+        'Image title',
+        required=True,
```
dreispt: What if I don't care about image titles?
Can this be optional?

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+
+    owner_id = fields.Integer(
+        "Owner",
+        required=True)
+    owner_model = fields.Char(
+        required=True)
+    storage = fields.Selection(
+        [('url', 'URL'), ('file', 'OS file'), ('db', 'Database')],
+        required=True,
+        default='db')
```
dreispt: I'm curious about this default. 
It comes unexpected since the Odoo default for "attached files" is the filesystem.

base_multi_image/README.rst
pedrobaeza, line 94: Akretion has nothing to do with this. Why do you say that?

base_multi_image/models/__init__.py
pedrobaeza, line 2: I'm not included, for example, and as said, Sharoon Thomas has nothing to do here because he was the original author of a module for multiple images, not this one, that has been totally redesigned, and I added the possibility of loading from URL, db, file...

base_multi_image/models/image.py
pedrobaeza: But binary fields are stored in the DB. Do you mean to mimick filestore here? The module doesn't have that option. Maybe it can be added an storage option called 'attachment'.

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+        if self.path and os.path.exists(self.path):
+            try:
+                with open(self.path, 'rb') as f:
+                    return base64.b64encode(f.read())
+            except Exception as e:
+                _logger.error("Can not open the image %s, error : %s",
+                              self.path, e, exc_info=True)
+        else:
+            _logger.error("The image %s doesn't exist ", self.path)
+
```
dreispt, line 102: I was expecting this to use the file storage logic Odoo uses for attached files...

pedrobaeza:
I'm thinking that is interesting to have a controller that allows to retrieve these images via a URL, so that you can link them in any other site.

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+        return False
+
+    @api.multi
+    def _get_image_from_url(self):
```
dreispt, line 106: Should we consider using api cache here?

base_multi_image/models/image.py
pedrobaeza, line 102: This is intended for files that already exists on your server, for example a static folder that you share with Magento or similar. I have already talked about the possibility of having a fourth option called 'attachment' that uses the attachment logic. It's indeed a good improvement.

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+                vals = tools.image_get_resized_images(
+                    s.with_context(bin_size=False).image_main)
+            except:
+                vals = {"image_medium": False,
+                        "image_small": False}
+            s.update(vals)
+
+    @api.multi
+    def _check_filestore(self):
+        """check if the filestore is created, and do it otherwise."""
```
dreispt: The file path, not the filestore.
When you say filestore I think of Odoo's attachment filestore, but AFAICS that's not what it is meant here.

base_multi_image/models/image.py
```diff
@@ -0,0 +1,181 @@
...
+            dir_path = os.path.dirname(s.path)
+            try:
+                if not os.path.exists(dir_path):
+                    os.makedirs(dir_path)
+            except OSError as e:
+                raise exceptions.Warning(
+                    _('The image filestore cannot be created, %s') % e)
+
+    @api.model
+    def _make_pretty(self, name):
```
dreispt: I suggest the name `_make_name_pretty`

base_multi_image/models/image.py
pedrobaeza, line 106: That was desirable, but for now, there's no cache mechanism. Do you know any quick mechanism?

base_multi_image/models/image.py
pedrobaeza: Hehe, yes, this method doesn't make anybody in a photo image to look prettier :stuck_out_tongue_closed_eyes:

base_multi_image/models/image.py
hbrunn, line 106: @pedrobaeza https://github.com/OCA/OCB/blob/8.0/openerp/tools/cache.py#L49

base_multi_image/models/owner.py
```diff
@@ -0,0 +1,101 @@
...
+                s.image_main_medium = s.image_ids[0].image_medium
+                s.image_main_small = s.image_ids[0].image_small
```
dreispt: I dislike picking the first image you get, it is inflexible.
Also notice that the images model does not have an order by attribute.

I suggest adding a "sequence" field on images.
Then the user would be free to choose the presentation order and it would be OK for the first one to be the main image.

base_multi_image/README.rst
```diff
@@ -0,0 +1,91 @@
...
+* You have to inherit model ``base_multi_image.owner`` to the model that needs
+  the gallery::
+
+    class MyOwner(models.Model):
+        _name = "mymodule.name"
+        _inherit = "base_multi_image.owner"
+
+        # If you need this, you will need ``post_init_hook_for_submodules``
+        old_image_field = fields.Binary(related="image_main", store=False)
+
```
dreispt, line 46: Can you also offer here the XML we would typically add on the Forms wherev this feature is added?

dreispt:
Nice feature, smart implementation, well done!
I made a few comments on the design an implementation.

base_multi_image/README.rst
yajo: Refactoring does not mean the old versions have never been contributed.

pedrobaeza: That's not a refactoring. It's a totally new module.

pedrobaeza: The only thing that remains is the idea of having "multiple images"

base_multi_image/models/__init__.py
yajo, line 2: I respected the original headers in the previous versions. Removing copyrights is a bit disrespectful at my sight.

However new files I usually just put mine, since they are totally new. Anyway, the CLA shares the copyright with OCA and grants them rights to do whatever, so this is mostly anecdotal.

Anybody wishing to know the real contributors can do so by `git blame`, line by line, and as long as they are listed in the README, seems legit to me.

pedrobaeza, line 2: Well, as this has been totally fragmented, I doubt git blame give the proper information. But the facts are these:
- I started from product_multi_image from Sharoon Thomas.
- In one moment in the development, there were 0 lines of code from the original (and I'm not talking about API rewrite).
- Then, you took this and split it in a base + additional modules.

So for me it's clear: the copyright should be shared between you and me.

pedrobaeza, line 2: And besides, this PR only contains 2 commits from you, so no possibility for `git blame`.

base_multi_image/models/image.py
guewen, line 102: Beware with that, because if you have a lot of products with several images, you will end up by exhausting the number of inodes in the directory and / or degrade the performance when the number of files increases. We had this exact problem for a customer and it was a real pain to correct afterwards. This is why generally files are split in subdirectories (odoo puts the files in a directory named after the first 2 chars of the name of the files, which is a hash btw, magento creates 2 levels of subdirectories with the first 2 chars of the name).

Example for Odoo:

```
├── 08
│   └── 084c97c3dafa6981786f472e63f2196dd3d9ffdd
├── 27
│   └── 2719fbd9be57673729beee772aaac15039fe872d
```

Example for Magento:

```
├── p
│   ├── r
│   |   └── product.jpg
```

I think that the product_images module was doing that as well.

But reusing the Odoo attachments would be safer as it already handles more things (renaming, ...).

yajo, line 102: Nice catch. However, when you use a file image here, it's expected that the image is already in the system. You just provide the path and Odoo reads it. No file saving on disk.

base_multi_image/models/image.py
yajo, line 106: Fixed

yajo:
All should be fixed.

base_multi_image/models/image.py
pedrobaeza, line 102: @guewen, in this case Odoo doesn't handle that. This mode is thought for sharing the same files for the e-commerce/mediastore and Odoo, so you don't have to duplicate the storage. Imagine for example that you have a Magento store with thousands of products with multiple images. You don't want to double the space needed for that photos. You can mount the Magento image folder in your system, and program the connector to put the proper path. Another alternative is the URL method, that patches the Magento photo URL. Providing these alternatives in the base module, the programmers can have alternatives to implement their strategies.

dreispt:
@Yajo following discussion above, it would be nice to add to the README Roadmap, the idea of adding a "attachment" storage mode. Maybe someone contributes that in the future.

dreispt:
> I'm thinking that is interesting to have a controller that allows to retrieve these images via a URL, so that you can link them in any other site.

@pedrobaeza I think that should go on the website companion module. I think @yajo also isolated that in a module.

base_multi_image/models/image.py
dreispt, line 102: I can see the current implementation useful for the use case where people copy images to a network file share, usually organized with in subdirectories, and then link them to the Document.

For an approach where users create a directory "Product X" and then place the photos inside, it would be nice for the user to then be asked for the path and then attach to the document (the product) all images inside it. This could be done through a wizard. 
I think it's at least an idea to add to the README Roadmap section.

pedrobaeza:
@dreispt, I thought about putting on website module, but it's something similar to report, that has a controller for getting reports even without website installed.

antespi:
LGTM :+1: 
Thanks @Yajo

dreispt:
@pedrobaeza OK. My concern in only to avoid pushing the website dependency. That could be added now ot put in the roadmap.

pedrobaeza:
@dreispt, you don't need website for making a controller. See the report example.

base_multi_image/models/image.py
guewen, line 102: > Nice catch. However, when you use a file image here, it's expected that the image is already in the system. You just provide the path and Odoo reads it. No file saving on disk.

...

> in this case Odoo doesn't handle that. This mode is thought for sharing the same files for the e-commerce/mediastore and Odoo, so you don't have to duplicate the storage.

I get it! So that's fine :)

rafaelbn:
Thanks! :+1: the first piece

pedrobaeza:
@Yajo, can you add the missing features commented to the Known issues section?

rafaelbn:
Which known issues @pedrobaeza ?

pedrobaeza:
It has been discussed all over the PR: store as attachment and so on

yajo:
Yes, hold on.

yajo:
There you have the known issues.

About the controller, [Odoo already includes `website.image_url` that makes the dirty job for you](https://github.com/OCA/website/pull/171/files#diff-740f7661ca4e6b6b7669aa76f0ff4c4bR29), so I see no need for now.

This should be ready to merge then.

yajo:
:tada:

PR #384, opened 2016-03-08
[8.0] Issue with xmlid and field doesn't exists

kryskool:
When a reference exists in ir.model.data for a field, but the field in ir.model.fields doesn't exists anymore, it raised a MissingError exception

### DIFF
--- a/database_cleanup/model/purge_modules.py
+++ b/database_cleanup/model/purge_modules.py
@@ -21,6 +21,7 @@
 
 from openerp import pooler
 from openerp.osv import orm, fields
+from openerp.exceptions import MissingError
 from openerp.modules.module import get_module_path
 from openerp.tools.translate import _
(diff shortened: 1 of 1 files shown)


database_cleanup/model/purge_modules.py
```diff
@@ -54,7 +55,11 @@ def _module_data_uninstall(self, cr, uid, modules_to_remove, context=None):
                 ctx[MODULE_UNINSTALL_FLAG] = True
                 field = self.pool[this.model].browse(
                     cr, uid, this.res_id, context=ctx)
-                if not self.pool.get(field.model):
```
tremlin, line 57: Thanks for the bug report. In the case that you describe that the fields does not exist in `ir.model.fields` anymore, shouldn't then `field` already evaluate to `False` and the condition `if not field or not self.pool.get(field.model):` suffice to avoid the error? I think the `try`/`catch` is a bit too much.

PR #385, opened 2016-03-11
Fix bug that made previous images to disappear.

yajo:
Fixes https://github.com/OCA/product-attribute/pull/135#issuecomment-191358505. (when merged).

Although this is a breaking change, it current submodules don't use hooks (the only breaking part). The current implementation was useless anyway.

@rafaelbn @carlos-incaser

### DIFF
--- a/base_multi_image/README.rst
+++ b/base_multi_image/README.rst
@@ -55,6 +55,20 @@ To develop a module based on this one:
         }"
         mode="kanban"/>
 
+* If the model you are extending already had an image field, and you want to
+  trick Odoo to make those images to multi-image mode, you will need to make
+  use of the provided :meth:`~.hooks.pre_init_hook_for_submodules`, like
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why it has to be done in pre instead of post?

yajo:
Because on post images are already gone.

pedrobaeza:
OK, I understand. What about the signature change?

yajo:
Signature?

pedrobaeza:
The signature is the number and the name of the arguments of a method

yajo:
ah sorry, pre_init_hooks lack the registry argument.

pedrobaeza:
OK, thanks for the explanation. I didn't remember that.

:+1: 

I'm going to fast-track to avoid people that starts using this module uses this wrong hook.

yajo:
Thank you :wink:

PR #386, opened 2016-03-14
Modules for data cleanup (reset) modules

mtelahun:
These modules allow an administrator to remove database content in a clean way. In addition to removing model records it also removes ancillary records such as properties, attachments, etc. While it is based on the database_cleanup module it does not alter the database structure.

While the modules do not cover the whole list of content that can be removed I think they are good basic starting point. The modules included in the PR are:

o content_cleanup - base module; does not do anything by itself
o content_cleanup_account - remove accounting data (but not chart of accounts or taxes)
o content_cleanup_account_voucher - remove accounting voucher data
o content_cleanup_purchase - remove purchase orders and related data
o content_cleanup_sale - remove sale orders and related data
o content_cleanup_stock - remove stock move and related data (including inventories)


libreec:
Hi, I want to create a similar module, but I'm using internal Odoo functions to do this, like account_cancel and unlink(), it's a little bit slower, but the idea is avoid to leave data undeleted or delete needed data.

As everything is managed by Odoo, It gives me the security of a useful database.

I didn't tried to use this method, but, if it's working, I'd like to use it instead.

Have you used this module with validated and paid invoices, with payments and with production orders?, is this working without leaving corrupted data?.

Thanks.

Issue #271, opened 2015-10-14
Migration to version 9.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0
# Modules to migrate
- [x] admin_technical_features by @njeudy in #318
- [x] auditlog - ~~By @holdenrehg - #454~~ - By @sebalix - #603 
- [ ] auth_admin_passkey - By @rvalyi - #522
- [x] auth_brute_force - By @Yajo - #1255  
- [ ] auth_dynamic_groups
- [ ] auth_from_http_basic
- [ ] auth_from_http_basic_logout
- [x] auth_from_http_remote_user @lmignon #457
- [ ] auth_saml - #274
- [x] auth_signup_verify_email - by @Yajo #436 
- [x] auth_supplier - By @Yajo - #434 
- [x] auto_backup - by @lasley - #526 
- [X] base_concurrency - #342
- [x] base_custom_info - By @Yajo - #435
- [x] ~~base_debug4all~~ - Not needed anymore (in core) - Removed in #573 
- [x] base_external_dbsource by @naousse in #513 
- [ ] base_field_serialized
- [x] base_import_match - By @cubells - #649
- [ ] base_ir_filters_active
- [x] base_multi_image - Assigned to @atchuthan - #402
- [x] base_name_search_improved - By @nicomacr - #621 
- [x] base_optional_quick_create #387, see also https://github.com/OCA/server-tools/pull/300#issuecomment-159908187
- [ ] base_search_fuzzy - By @darshan-serpent - #675 
- [x] base_suspend_security - #333
- [x] base_report_auto_create_qweb - By @damdam-s - #438
- [x] base_user_gravatar - #338
- [x] base_user_role - By @sebalix - #641 
- [x] configuration_helper @yvaucher #411 
- [x] cron_run_manually #380 (Deleted because is now on core)
- [x] database_cleanup - By @hbrunn - #469
- [x] datetime_formatter - By @cubells - #605 
- [x] dbfilter_from_header - #382
- [x] disable_openerp_online > disable_odoo_online - #399
- [ ] email_template_template
- [ ] fetchmail_attach_from_folder
- [ ] fetchmail_notify_error_to_sender
- [x] html_image_url_extractor - @cubells - #624 
- [x] html_text - @cubells - #625
- [ ] import_odbc
- [x] inactive_session_timeout > auth_session_timeout - #379
- [ ] ir_config_parameter_viewer
[...]

PR #387, opened 2016-03-15
[WIP] from 8.0 to 9.0 base_optional_quick_create

JonathanNEMRY:
- bump version to 9.0.1.0.0 and make it installable
- apply a views/models directory arch into the module
- fix xpath anchor on ir_model views (osv_memory to transient)
- add README.rst
- apply some new api


pedrobaeza:
This modules shouldn't be migrated to 9.0. As discussed on https://github.com/OCA/server-tools/pull/300#issuecomment-159908187. We mustn't duplicate features between several modules. The module to be migrated is: https://github.com/OCA/web/tree/8.0/web_m2x_options

sbidoul:
@pedrobaeza `base_optional_quick_create` is a feature accessible to the administrator and applies globally.

OTOH, if I understand well, `web_m2x_options` can achieve the same result but requires changing the option on all widgets in all the views where they are used.

So IMO it's a different use case and `base_optional_quick_create` must be kept.

pedrobaeza:
No, you can apply it globally with a system parameter. Please read the discussion between @StefanRijnhart and me on the link provided to get background about it.

sbidoul:
@pedrobaeza which link?

pedrobaeza:
Sorry, I though I copied that link, but it was wrong: https://github.com/OCA/server-tools/pull/300#issuecomment-159908187 (copied also on the original quote).

sbidoul:
I see. We have no budget to work on this at the moment so we'll leave it as is for the moment. I updated #271.

StefanRijnhart:
I still don't agree with @pedrobaeza. Apparently, base_optional_quick_create has quite a number of sponsors here, and it's technically in a pretty good condition now.

sbidoul:
rebased and added setup.py cc/ @JonathanNEMRY @Olivier-LAURENT

JonathanNEMRY:
Thanks @sbidoul ! :+1: 
Regarding last @StefanRijnhart 's comment  do you now agree to take this one as a valid candidate for 9.0 @pedrobaeza ?

pedrobaeza:
OK, go ahead

sbidoul:
Tested on runbot. :+1:

StefanRijnhart:
Thanks! :+1:

atchuthan:
Tested on runbot :+1:

JonathanNEMRY:
Thank you all!

PR #389, opened 2016-03-16
[8.0][FIX] base_report_auto_create_qweb

tafaRU:
Handle case in which the qweb view name contains a dot, see https://github.com/odoo/odoo/blob/8.0/addons/account/views/report_overdue.xml#L4 as reference.


pedrobaeza:
:+1:

sebalix:
:+1:

eLBati:
:+1:

PR #391, opened 2016-03-16
[ADD] allow to push values constructed by expressions

hbrunn:


tremlin:
By the current module development guidelines, the version should be bumped to `8.0.1.1.0` because a module upgrade is necessary. Besides this technicality code looks good.

Issue #235, opened 2015-08-25
Supacalendar ist not able to handle recurrency meetings


mh-manatec:
When i track calendar.event and configure recurrent meeting the server says:

Traceback (most recent call last):
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 537, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 574, in dispatch
    result = self._call_function(*_self.params)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/service/model.py", line 113, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 307, in checked_call
    return self.endpoint(_a, *_kw)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 803, in __call__
    return self.method(_args, *_kw)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 403, in response_wrap
    response = f(_args, *_kw)
  File "/opt/odoo/odoo8/opt/addons/web/controllers/main.py", line 948, in call_button
    action = self._call_kw(model, method, args, {})
  File "/opt/odoo/odoo8/opt/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/api.py", line 241, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/opt/odoo/odoo8/local/lib/python2.7/site-packages/odoo-8.0-py2.7.egg/openerp/api.py", line 363, in old_api
    result = method(recs, _args, *_kwargs)
[...]

PR #392, opened 2016-03-16
Fix problem with recurrent events

danimaribeiro:
they change the type of id which causes an error retrieving the id
#235


danimaribeiro:
@mh-manatec Can you test it?

hbrunn:
:+1:

pedrobaeza:
:+1:

Issue #332, opened 2016-01-13
restricting field access per field, per record, depending on other values


hbrunn:
Hi all,

we've come far already with field access restrictions, but I'm not aware of a general and simple way to restrict access to fields based on values of the record (I'm not talking about the UI, but hard restrictions you can't circumvent).

I'm tinkering with a mixin that is supposed to make this task easy: https://gist.github.com/hbrunn/776143edc5e5f6e08367

With this, the following code is sufficient for my client to have only a couple of fields of the partner model readable/editable if the current user is not either the partner's salesperson or in the sales team assigned to the partner (the latter are custom fields):

``` python
class ResPartner(models.Model):
    _inherit = ['restrict.field.access.mixin', 'res.partner']
    _name = 'res.partner'

    @api.multi
    def _restrict_field_access_get_field_whitelist(self, action='read'):
        whitelist = [
            'name', 'parent_id', 'is_company', 'firstname', 'lastname',
            'infix', 'initials',
        ] + super(ResPartner, self)\
            ._restrict_field_access_get_field_whitelist(action=action)
        if action == 'read':
            whitelist.extend(['section_id', 'user_id'])
        return whitelist

    @api.multi
    def _restrict_field_access_is_field_accessible(self, field_name,
                                                   action='read'):
        result = super(ResPartner, self)\
            ._restrict_field_access_is_field_accessible(
                field_name, action=action)
        if result or not self:
            return result
        return all(this.section_id in self.env.user.section_ids or
                   this.user_id == self.env.user
                   for this in self)

    @api.multi
    @api.onchange('section_id', 'user_id')
    @api.depends('section_id', 'user_id')
    def _compute_restrict_field_access(self):
        return super(ResPartner, self)._compute_restrict_field_access()

    @api.model
    def _restrict_field_access_inject_restrict_field_access_domain(
            self, domain):
        domain[:] = expression.AND([
            domain,
            [
                '|',
                ('section_id', 'in', self.env.user.section_ids.ids),
                ('user_id', '=', self.env.user.id),
            ],
        ])
```

simple enough, right? Note that we can have a different set of fields writable than readable.

Now my questions:
- am I doing something stupid and similar functionality exists already?
- I'm inclined to polish the mixin a bit and PR it as a module here, do we want that?
- any features missing?

hbrunn:
ping @gurneyalex @pedrobaeza @StefanRijnhart - I'd be curious about your opinion on this one

hbrunn:
[...]

PR #396, opened 2016-03-21
[ADD] base_mixin_restrict_field_access

hbrunn:
This is the result of https://github.com/OCA/server-tools/issues/332:
# 
# Restrict field access

This module was written to help developers restricting access to fields in a
secure and flexible manner.

If you're not a developer, this module is not for you as you need to write code
in order to actually use it.
# Usage

``` python
        # inherit from the mixin
        _inherit = ['restrict.field.access.mixin', 'res.partner']
        _name = 'res.partner'

        @api.multi
        def _restrict_field_access_get_field_whitelist(self, action='read'):
            # return a whitelist (or a blacklist) of fields, depending on the
            # action passed
            whitelist = [
                'name', 'parent_id', 'is_company', 'firstname', 'lastname',
                'infix', 'initials',
            ] + super(ResPartner, self)\
                ._restrict_field_access_get_field_whitelist(action=action)
            if action == 'read':
                whitelist.extend(['section_id', 'user_id'])
            return whitelist

        @api.multi
        def _restrict_field_access_is_field_accessible(self, field_name,
                                                       action='read'):
            # in case the whitelist is not enough, you can also decide for
            # specific records if an action can be carried out on it or not
            result = super(ResPartner, self)\
                ._restrict_field_access_is_field_accessible(
                    field_name, action=action)
            if result or not self:
                return result
            return all(this.section_id in self.env.user.section_ids or
                       this.user_id == self.env.user
                       for this in self)

        @api.multi
        @api.onchange('section_id', 'user_id')
        @api.depends('section_id', 'user_id')
        def _compute_restrict_field_access(self):
            # if your decision depends on other fields, you probably need to
            # override this function in order to attach the correct onchange/
            # depends decorators
            return super(ResPartner, self)._compute_restrict_field_access()

        @api.model
        def _restrict_field_access_inject_restrict_field_access_domain(
                self, domain):
            # you also might want to decide with a domain expression which
            # records are visible in the first place
            domain[:] = expression.AND([
                domain,
                [
                    '|',
                    ('section_id', 'in', self.env.user.section_ids.ids),
                    ('user_id', '=', self.env.user.id),
                ],
            ])
```

The example code here will allow only reading a few fields for partners of
which the current user is neither the sales person nor in this partner's sales
team.


hbrunn:
travis' complaint is a false positive discussed here: https://github.com/OCA/maintainer-quality-tools/pull/307

pedrobaeza:
You can add a comment for disabling the check on that line:

```
# pylint: disable=<warning-code>
```

hbrunn:
@pedrobaeza thanks

pedrobaeza:
Almost done:

```
base_mixin_restrict_field_access/tests/test_base_mixin_restrict_field_access.py:5:1: F401 'RegistryManager' imported but unused
```

hbrunn:
saw it, thanks. Weird nothing complained about this one before. Anyways

base_mixin_restrict_field_access/README.rst
```diff
@@ -0,0 +1,118 @@
...
+team.
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* the code contains some TODOs which should be done
```
pedrobaeza, line 92: Better to specify here the big issues: read_group, not so secure call and the rest of possible operations not covered.

pedrobaeza:
Little comment, but :+1:

guewen:
:+1:

dannyadair:
This doesn't seem to apply to CSV exports?

To reproduce please see this code:

`
from openerp.osv import osv
from openerp import api

class hr_employee(osv.osv):
    _inherit = ['restrict.field.access.mixin', 'hr.employee']
    _name = 'hr.employee'

```
@api.multi
@api.onchange('user_id')
@api.depends('user_id')
def _compute_restrict_field_access(self):
    """Re-determine field access when the employee's user changes."""
    return super(hr_employee, self)._compute_restrict_field_access()

@api.multi
def _restrict_field_access_is_field_accessible(self, field_name,
                                               action='read'):
    """Determine field access dynamically."""
    if field_name == 'birthday':
        return False
    result = super(hr_employee, self)._restrict_field_access_is_field_accessible(
        field_name, action=action
    )
    if result or not self:
        return result
    return True
```

`

(in the full code I work with two lists FIELDS_ONLY_SELF_OR_MANAGER and FIELDS_ONLY_MANAGER corresponding to the web interface's "Public Information"/"Personal Information"/"HR Settings" tabs)

So birthdate should always be inaccessible. And indeed, "Date of birth" on the form shows no value although there is one (even for the admin user). However, CSV export doesn't seem to care, and will always happily contain it.

When I click "Export To File" then _restrict_field_access_is_field_accessible() gets called with field_name 'birthday' and action 'read' but I assume that's just a side effect of reading the full record somewhere along the line.

I briefly looked at odoo.openerp.fields.convert_to_export()
Maybe inherit and use the same lists of restricted fields there to ensure they come out blank?
The field names themselves can't seem to be hidden but of course it's the value that's most important.

see also related https://github.com/odoo/odoo/issues/12077

hbrunn:
thanks for the pointer, I updated https://github.com/OCA/server-tools/pull/396/files#diff-9797ae3b8487be8a1f1cd8e6faea9ad8R14 accordingly.

The solution will be to check field access in https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L859 and drop fields from the list as appropriate.

As the project for which I started this got suspended for a while, I won't be working on this any time soon, but I'll be happy to merge your PRs on this branch. As it seems you need this to be air tight, don't forget to also fix `read_group` as the comment above indicates. Setting this to work in progress meanwhile.

hbrunn:
PS: The list of fields to export is compiled in https://github.com/OCA/OCB/blob/8.0/addons/web/controllers/main.py#L1291 if you want to change it

dannyadair:
Thanks Holger, I monkeypatched web.ExportFormat.base to drop the fields. get_fields() looks like a cleaner location, and seems to have everything that's needed (model, request's user).

Hadn't noticed the read_group() issue until now, thanks for mentioning this. I'm going to Germany for a month after next week so this won't go in production here either anytime soon. :-)

dannyadair:
Took me also a while to get back to this.
Please check out https://github.com/hbrunn/server-tools/pull/3

hbrunn:
@guewen @pedrobaeza you've thumbed this up, can we merge if there's a green runbot?

pedrobaeza:
Do you plan to squash a bit the commit history?

hbrunn:
I did not, but you're right, that should happen. Done

PR #397, opened 2016-03-21
[ADD] base_ir_filters_active

hbrunn:


pedrobaeza:
Please put the name in singular. Isn't convenient to use "active" word in the technical name (_base_ir_filter_active_)?

hbrunn:
@pedrobaeza good point about `active`, makes sense for a technical module. For plural/singular, I sticked to the model name: https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_filters.py#L27 - agreeable? (I don't want to have to change the name several times, to let's get it right on the first rename)

pedrobaeza:
OK, as the model name is in plural, then let's stick to that plural.

hbrunn:
@pedrobaeza rename done

pedrobaeza:
:+1: 

Thanks

guewen:
:+1:

PR #403, opened 2016-04-01
FIX super_calendar images URL in README

eLBati:


tafaRU:
:+1:

sebalix:
:+1:

pedrobaeza:
Please use the same technique as in here to allow to see the image in all the possible places: https://github.com/OCA/web/pull/336

hbrunn:
@eLBati I had to think twice what @pedrobaeza means, but the proposal is to have absolute urls (so the change should be to add a slash at the beginning)

pedrobaeza:
Yeah, sorry for not being clear, but I mean exactly that: put the `/` at the beginning with the absolute path. It seems to work in both GitHub and apps. I'm not so sure about in Odoo itself.

eLBati:
@pedrobaeza @hbrunn thanks, done

hbrunn:
:+1:

PR #404, opened 2016-04-02
[ADD] email_template_qweb

hbrunn:
# 
# QWeb for email templates

This module was written to allow you to write email templates in QWeb instead
of jinja2. The advantage here is that with QWeb, you can make use of
inheritance and the `call` statement, which allows you to reuse designs and
snippets in multiple templates, making your development process simpler.
# Usage

- Select `QWeb` in the field `Body templating engine`
- Select a QWeb view to be used to render the body field
- Apart from QWeb's standard variables, you also have access to `object` and
  `email_template`, which are browse records of the current object and the
  email template in use, respectively.

Demo data contains an example on how to separate corporate identity from a
template's content.


hbrunn:
for me, this module is a full replacement for `email_template_template`: https://github.com/OCA/server-tools/blob/8.0/email_template_template/__openerp__.py#L27 - should I include deleting this one in this PR or do we want to keep it around in case someone made heavy use of this in 7.0 or 6.1 and it's not feasible to move the templates to qweb during migration?

elicoidal:
@hbrunn what about the repo reporting-engine?

hbrunn:
@elicoidal this has nothing to do with reporting, so I wouldn't see why. What's your reasoning here? If we don't want it in server-tools, I think the crm repo is a good place, at least it contains a couple of addons dealing with mails

pedrobaeza:
It's difficult to classify this module indeed. What about OCA/social?

hbrunn:
social is a good choice too I think. So do we agree on moving this to social? And what do you think about email_template_template?

pedrobaeza:
OK for moving. What do you mean with email_template_template? Do you refer to the name of the module? I see it OK.

hbrunn:
no, what I wrote in the first comment: https://github.com/OCA/server-tools/pull/404#issuecomment-204694356 - I wrote `email_template_template` to support inheritance which you get for free here. Given the code is a bit clumsy and qweb templates are a lot more elegant in my opinion, I won't continue developing this module and don't see a reason (except for having a lot of legacy templates) why anyone else should

pedrobaeza:
Yeah, for sure. That module is useless with this new module, but people may use it for this stable version, so we should keep it, and mark in v9 that the module won't be migrated in favor of the new one.

hbrunn:
it was never ported to 8.0: https://github.com/OCA/server-tools/blob/8.0/email_template_template/__openerp__.py#L103 - so I'd suggest making PRs to remove the module in 8.0 and 9.0, and putting a note in 7.0's readme that the module is superseded by this one.

pedrobaeza:
OK, perfect then. Do you proceed?

hbrunn:
yes, I will. Just not right now, I'm on something else currently

elicoidal:
:+1:  for the move to OCA/social

atchuthan:
@hbrunn I was looking for this module... thanks
tested the functionality at Runbot :+1:

lasley:
Excellent module, also tested on Runbot. Code looks to be inline with what I know of OCA standards as well :+1:

hbrunn:
moved to https://github.com/OCA/social/pull/55

jholze:
@hbrunn just a question, i couldn't check it in a odoo instance. Will this work with the email editor ? So editing a mail template will be possible as it is ?

hbrunn:
no

PR #407, opened 2016-04-05
[FIX] Do not die miserably when you get `False` as the HTML input. Test it.

yajo:
Title says it all. This fixes an ugly bug in the module. Simple fix, simple test.

@rafaelbn


hbrunn:
:+1:

pedrobaeza:
:+1:

sebalix:
:+1:

PR #408, opened 2016-04-06
[ADD] users_ldap_populate: migrate functionality added to 6.1 after 7.0 port

hbrunn:
This module was ported in mid 2013 from https://github.com/Therp/Therp-Addons/tree/6.1/users_ldap_populate, but we kept developing it for our 6.1 customers. This brings both versions to the same level.

I considered taking the opportunity to move the code to new standards, but this would create a lot of noise in the diff, so we can better do this afterwards (now our last customer using this has migrated away from 6.1, so from our side, probably no further 6.1 developments will happen)

### DIFF
--- a/users_ldap_populate/README.rst
+++ b/users_ldap_populate/README.rst
@@ -1,0 +1,72 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+=============
+LDAP Populate
+=============
+
+This module allows to prepopulate the user database with all entries in the
+LDAP database.
+
+Installation
+============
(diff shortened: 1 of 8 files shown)


gurneyalex (approved):

gurneyalex:
@hbrunn I like this new feature. Much wanted. Do you mind if I squash when merging?

hbrunn:
@gurneyalex not at all, go ahead! (I'd do it myself if there were signed commits involved, but this ancient one is more convenient via the github UI)

gurneyalex:
@hbrunn I will take care of having these forward ported to 9.0 and 10.0

PR #409, opened 2016-04-06
[PORT] [9.0] record_archiver

yvaucher:


pedrobaeza:
Interesting module! In a talk with @odony, he told me that they tried something similar with a customer, but at the end, Postgres doesn't scale well with large tables, and they ended up having a twin table where they dump old data.

Do you have any stat on this?

yvaucher:
@pedrobaeza We already had it on 7.0 and used it for quite a while (aprox. 7 month): 
https://github.com/OCA/server-tools/tree/7.0/record_archiver

We used it on a database of 50 Gb to speed up a bit things. I don't have any stats to show though.

yvaucher:
Just added a little PEP8 fix in tests

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
...
+        recs.invalidate_cache()
+        _logger.info(
+            'Archived %s %s older than %s',
+            len(recs.ids), model_name, expiration_date)
+
+    @api.multi
+    def archive_records(self):
+        """ Call the archiver for several record lifespans """
+        for lifespan in self:
+            self._archive_lifespan_records()
```
atchuthan: use lifespan instead of self
**lifespan**._archive_lifespan_records()

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
...
+    @api.multi
+    def _archive_domain(self, expiration_date):
+        """ Returns the domain used to find the records to archive.
+
+        Can be inherited to change the archived records for a model.
+        """
+        model = self.env[self.model_id.model]
+        domain = [('write_date', '<', expiration_date)]
+        if 'state' in model._columns:
+            domain += [('state', 'in', ('done', 'cancel'))]
```
atchuthan, line 67: If we set the archive feature for a custom model where the 'state' does not contain 'done' or 'cancel' options, then this domain will fail(Invalid leaf error).  
It would be best to get the domain from the FORM view itself.

yvaucher, line 67: No, the state needs one of the 2 states to work. If one of the two is missing, it doesn't matter. There is no check on the domain compared values.
This is the default behaviour as `'done'` and `'cancel'` are standard states.

If you made a model than doesn't respect this, you can still override `_archive_domain` in a new module to extend this function.

record_archiver/models/record_lifespan.py
yvaucher: Good catch

yvaucher:
Fixed loop error

record_archiver/models/record_lifespan.py
atchuthan, line 67: okay

atchuthan:
code reviewed :+1:

odony:
@pedrobaeza for the record, the problem with archiving is not exactly that PostgreSQL does not scale for large tables, but rather that mixing 2 "independent" types of records in the same tablespace is inefficient. In order to get a speed improvement you'll need to add indexes on the `active` column (there is none by default). Then imagine you have 30% of active records, and 60% of inactive records: because those are randomly organized inside the table data files, using an index to access the active records will actually cause more work (due to random access in many files) than sequentially reading all data files and skipping the inactive one. The PostgreSQL query planner will detect this and will always perform full table scans, making the index useless.

That said, deactivating records might still provide a performance boost if the data organization makes it possible (e.g. if you have 95% deactivated records, which may take time), or if you use other techniques, like table partitioning, which may require modified queries.

Of course you could decide to archive records for other reasons than performance...

pedrobaeza:
Thanks for the insights, @odony! I didn't remember well the details. I only remember the conclusion that it's not the best option for lot of cases. But now all is clear.

record_archiver/README.rst
```diff
@@ -0,0 +1,71 @@
...
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_.  In case of trouble, please
+check there if your issue has already been reported.  If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+server-tools/issues/new?body=module:%20
+record_archiver%0Aversion:%20
+9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
yvaucher, line 46: This has just changed 
https://github.com/OCA/maintainer-tools/pull/180

guewen:
@odony thanks a lot for your comment! Indeed, an index is useless in most cases on a boolean field, and we didn't add indexes on the `active` field. Still, IIRC we did observe general performance UX improvements (in m2o, searches, ...) in 7.0 after using this addon on a large database (but we have no numbers though..), I guess because after a `SELECT ... WHERE active = True` there is less rows to process. 

All in all, this addon probably doesn't improve posgres' performance, and should not be seen as a postgres archiving technique, rather as an odoo archiving one :)

It's worth to check if if we notice an improvement or not on 9.0 using this technique.

grindtildeath (approved):
As this was migrated to 11.0 and is already merged for a newer version, I guess the original PR should be merged as well.

dreispt:
@guewen I'm late into the party, but I recently learned about partial indexes: you add a WHERE clause to the Index, and it will be created only for the records meeting that criteria.
There indexes are preferred to the full table ones.
So, if you add `WHERE active=true` to the Index, you will get much faster index scans for most of the queries.
This could be a nice recommendation to add to the module's docs.

PR #418, opened 2016-05-03
[FIX] logger db_backup for pysftp

bealdav:
If I don't install the module, I don't want a warning, please.
ping @Yajo

### DIFF
--- a/auto_backup/models/db_backup.py
+++ b/auto_backup/models/db_backup.py
@@ -18,7 +18,7 @@ _logger = logging.getLogger(__name__)
 try:
     import pysftp
 except ImportError:
-    _logger.warning('Cannot import pysftp')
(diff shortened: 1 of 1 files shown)


auto_backup/models/db_backup.py
```diff
@@ -18,7 +18,7 @@
 try:
     import pysftp
 except ImportError:
-    _logger.warning('Cannot import pysftp')
+    _logger.info('Cannot import pysftp')
```
tafaRU, line 21: Shouldn't be better to use debug level as written at https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror ?

yajo, line 21: Indeed.

bealdav:
FIxed PR here #419

PR #420, opened 2016-05-11
[8.0][ADD][base_import_match] Import deduplicator.

yajo:
By default, when importing data (like CSV import) with the `base_import`
module, Odoo follows this rule:
1. If you import the XMLID of a record, make an **update**.
2. If you do not, **create** a new record.

This module allows you to set additional rules to match if a given import is an
update or a new record.

This is useful when you need to sync heterogeneous databases, and the field you
use to match records in those databases with Odoo's is not the XMLID but the
name, VAT, email, etc.

After installing this module, the import logic will be changed to:
1. If you import the XMLID of a record, make an **update**.
2. If you do not:
   1. If there are import match rules for the model you are importing:
      1. Discard the rules that require fields you are not importing.
      2. Traverse the remaining rules one by one in order to find a match in
         the database.
         1. If one match is found:
            1. Stop traversing the rest of valid rules.
            2. **Update** that record.
         2. If zero or multiple matches are found:
            1. Continue with the next rule.
         3. If all rules are exhausted and no single match is found:
            1. **Create** a new record.
   2. If there are no match rules for your model:
      1. **Create** a new record.


base_import_match/models/base_import.py
```diff
@@ -0,0 +1,220 @@
...
+# © 2016 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+from openerp.exceptions import except_orm as ValueError  # TODO remove in v9
+from openerp import SUPERUSER_ID  # TODO remove in v10
+
+
+class BaseImportMatch(models.Model):
+    _name = "base_import.match"
```
dreispt, line 10: Not that it makes a real difference, but shouldn't this be `base.import.match`?

base_import_match/models/base_import.py
```diff
@@ -0,0 +1,220 @@
...
+        """Wrap the model after writing."""
+        result = super(BaseImportMatch, self).write(vals)
+
+        if "model_id" in vals or "model_name" in vals:
+            for s in self:
+                self._load_autopatch(s.model_name)
+
+        return result
+
+    # TODO convert to @api.model_cr in v10
```
dreispt, line 69: Why not do it now?

yajo, line 69: `@api.model_cr` does not exist until v10.

base_import_match/models/base_import.py
yajo, line 10: Core `base_import` module uses the `base_import` namespace for its models.

base_import_match/models/base_import.py
```diff
@@ -0,0 +1,220 @@
...
+                    match.ensure_one()
+
+                # You hit this because...
+                # a. No match. Odoo must create a new record.
+                # b. Multiple matches. No way to know which is the right
+                #    one, so we let Odoo create a new record or raise
+                #    the corresponding exception.
+                # In any case, we must do nothing.
+                except ValueError:
+                    continue
```
dreispt: Maybe this could be an option in the match rule: a checkbox for "Error if multiple matches found?"

yajo: Good idea, but I'll add it to roadmap, it's not my current target.

base_import_match/models/base_import.py
```diff
@@ -0,0 +1,220 @@
...
+
+                # Give a valid XMLID to this row
+                row["id"] = match._BaseModel__export_xml_id()
```
dreispt: Missing a dot: `match._BaseModel.__export_xml_id()`

dreispt: Shouldn't it be just `match.__export_xml_id()` ?

base_import_match/models/base_import.py
dreispt: Roadmap idea: add a log to write these exceptions to, so that end users can monitor it and decide what to do to each individual record: ignore or force load.

dreispt:
I know this pain, and currently solve it by providing my own XML Ids at import time, generated based on the key attributes, such as the Customer VAT or a Contract code.

This is an Interesting alternative solution for that problem.
I felt the code somewhat hard to read at some points, and wish it could be clearer, but that doesn't make it wrong.

It's +1 for me once comments fixed and Travis is green.

yajo:
I know it's hard to read, but the import feature in Odoo is quite cumbersome. I had to dig a lot in the code to make this work.

I refactored it in c0f7ceb a bit to make it compatible with company import by VAT, and added that profile by default. Please test the new version, it's quite better.

base_import_match/models/base_import.py
yajo: That's too much IMO. I did not want to change the import UI with this, just the inner behavior.

base_import_match/models/base_import.py
yajo: This must be what it is :wink:

moylop260:
Could you check rst-syntax?

``` bash
base_import_match/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:30 Unexpected indentation.
base_import_match/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:63 Unexpected indentation.
base_import_match/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:59 Inline strong start-string without end-string.
```

More info here: https://travis-ci.org/OCA/server-tools/jobs/129966566#L430

base_import_match/models/base_import.py
dreispt, line 10: ok

base_import_match/models/base_import.py
dreispt, line 69: ok

dreispt:
:+1:

moylop260:
@Yajo 
FYI I used http://rst.ninjs.org/ to fix [current errors](https://travis-ci.org/OCA/server-tools/jobs/132766029#L423-L427) of `base_import_match/README.rst` file.

And applying the following patch:

``` patch
diff --git base_import_match/README.rst base_import_match/README.rst
index f1c6577..52609b4 100644
--- base_import_match/README.rst
+++ base_import_match/README.rst
@@ -23,20 +23,26 @@ After installing this module, the import logic will be changed to:

 - If you import the XMLID of a record, make an **update**.
 - If you do not:
+
   - If there are import match rules for the model you are importing:
-      - Discard the rules that require fields you are not importing.
-      - Traverse the remaining rules one by one in order to find a match in
-        the database.
-          - Skip the rule if it requires a special condition that is not
-            satisfied.
-          - If one match is found:
-            - Stop traversing the rest of valid rules.
-            - **Update** that record.
-          - If zero or multiple matches are found:
-            - Continue with the next rule.
-          - If all rules are exhausted and no single match is found:
-            - **Create** a new record.
+
+    - Discard the rules that require fields you are not importing.
+    - Traverse the remaining rules one by one in order to find a match in the database.
+
+        - Skip the rule if it requires a special condition that is not
+          satisfied.
+        - If one match is found:
+
+          - Stop traversing the rest of valid rules.
+          - **Update** that record.
+        - If zero or multiple matches are found:
+
+          - Continue with the next rule.
+        - If all rules are exhausted and no single match is found:
+
+          - **Create** a new record.
   - If there are no match rules for your model:
+
     - **Create** a new record.

 By default 2 rules are installed for production instances:
@@ -59,6 +65,7 @@ To configure this module, you need to:
 #. If the rule must be used only for certain imported values, check
    *Conditional* and enter the **exact string** that is going to be imported
    in *Imported value*.
+
    #. Keep in mind that the match here is evaluated as a case sensitive
       **text string** always. If you enter e.g. ``True``, it will match that
       string, but will not match ``1`` or ``true``.

```

Now don't show rst errors:
screen shot 2016-05-25 at 10 40 00 am

moylop260:
👍

dreispt:
:+1:

yajo:
@moylop260 please pr me with that

moylop260:
You can use this patch with `git apply moy.patch` or `patch -p0 <moy.patch`
Just save it in a file.
Could you try it or you prefer a pr?

moylop260:
@Yajo [yajo#1](https://github.com/Yajo/server-tools/pull/1/files)

yajo:
:green_heart:

base_import_match/models/base_import.py
```diff
@@ -0,0 +1,297 @@
...
+    @api.multi
+    @api.onchange("model_id")
+    def _onchange_model_id(self):
+        self.field_ids.unlink()
+
+    @api.model
+    def create(self, vals):
+        """Wrap the model after creation."""
+        result = super(BaseImportMatch, self).create(vals)
+        self._load_autopatch(result.model_name)
```
dreispt, line 46: @Yajo I'm curious because I don't understand how this works:
AFAICS you are here specifically patching the BaseImportMatch model, not the BaseModel.
So this patch won't affect the other models as expected.
But probably I'm missing something.

yajo, line 46: This only patches the modules found in configuration. That's why that method is called when creating, deleting, writing and restarting Odoo: it patches configured models and unpatches those that not.

dreispt, line 46: OK, I can see.

PR #423, opened 2016-05-16
[FIX] auto_backup: bad reference to field sftp_private_key

jabibi:


auto_backup/models/db_backup.py
```diff
@@ -4,25 +4,27 @@
...
 from openerp.service import db
-import logging
+
 _logger = logging.getLogger(__name__)
 try:
     import pysftp
 except ImportError:
-    _logger.warning('Cannot import pysftp')
-
+    _logger.warning("Cannot import pysftp")
```
pedrobaeza: Why change these lines with different quotes? They act the same, so please reduce the diff of the PR not changing them.

pedrobaeza:
Change the version of the module according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

auto_backup/models/db_backup.py
jabibi: I am rewriting the entire module to add ftp/ftps support. I just make this PR because I found an error on the current code

pedrobaeza: Your rewrite is not efficient because you are changing things that don't need to be changed. If you find an error, please make a PR changing only the lines of the error.

jabibi:
clean diff and version increase
let me know if there is something more to change
Thanks

pedrobaeza:
Thanks :+1:

hbrunn:
:+1:

PR #425, opened 2016-05-17
9.0 - Let's Encrypt

lasley:
Hi All,

I've been lurking & waiting for this to be merged on 8, so we can get in 9 as well. No changes were required outside of versioning.

Thanks to all that actually did do work on this module. It's been a long time dream for self-deploying SSL, I am so happy to see it a reality!


letsencrypt/data/ir_cron.xml
```diff
@@ -0,0 +1,14 @@
...
+<openerp>
+    <data noupdate="1">
+        <record id="cronjob" model="ir.cron">
+            <field name="name">Update letsencrypt certificates</field>
+            <field name="interval_type">weeks</field>
+            <field name="interval_number">11</field>
+            <field name="numbercall">-1</field>
+            <field name="model">letsencrypt</field>
+            <field name="function">cron</field>
+            <field name="nextcall">2016-01-01</field>
```
moylop260, line 11: What about use a dynamic value here?

moylop260:
👍 
Thanks for your contribution

letsencrypt/data/ir_cron.xml
lasley, line 11: Would there be any functional difference? Isn't a cron immediately fired if it is past the nextcall?

moylop260, line 11: I don't remember...
I just saw the following example [odoo/mail](https://github.com/odoo/odoo/blob/96e085355d2e29d81ace06e08dc76ea5668846ef/addons/mail/data/mail_data.xml#L23)
If you consider what don't affect then is good for me too

lasley, line 11: Ahh in that sample, they are generating a cron for +7 days. We want an instant trigger here 😉

moylop260, line 11: Thank you!

hbrunn:
:+1:, runbot error is fixed by now

sebalix:
:+1:

letsencrypt/README.rst
```diff
@@ -0,0 +1,165 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza: Missing target link

lasley:
Oops seems I missed that last comment. Target link added!

PR #434, opened 2016-05-26
[MIG][9.0][auth_supplier] Migrate.

yajo:
Update to v9, with mostly cosmetic changes.

@Tecnativa https://github.com/OCA/server-tools/issues/271


moylop260:
👍 thanks

auth_supplier/views/auth_supplier_view.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
```
pedrobaeza: Change this to <odoo> to use the new "cool" tag (and to avoid the need to change it when the other tag will be deprecated)

pedrobaeza:
:+1:

auth_supplier/README.rst
```diff
@@ -0,0 +1,75 @@
...
+Auth Supplier
+=============
+
+This module was written to extends the functionality of auth signup
+and allows the user to create an account as a supplier.
+
+Configuration
+=============
+
+* To enable users to create accounts:
```
pedrobaeza: You can include here that "Developer mode" should be activated first to have access to "Technical features".

auth_supplier/README.rst
```diff
@@ -0,0 +1,76 @@
...
+* Enable *Allow external users to sign up*.
+* Enable *Activate the customer portal*.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Log out.
+* In home page, press *Sign in*.
```
sebalix: s/Sign in/Sign up/ (tested on runbot)

auth_supplier/views/auth_supplier_view.xml
```diff
@@ -0,0 +1,18 @@
...
+<odoo>
+
+    <template id="auth_supplier.fields"
+              inherit_id="auth_signup.fields"
+              name="Auth Signup/ResetPassword form fields">
+        <xpath expr="//div[@class='form-group field-login']" position="before">
+            <div class="form-group field-type">
+                <label for="account_type" class="control-label">Account Type</label>
+                <select class="form-control" name="account_type">
+                    <option value="customer" selected="selected">Customer</option>
```
sebalix, line 11: I tested the 'Customer' option as it seems to be the default behaviour (auth_signup), but the resulting partner is not a customer (not checked). Should we rename this option 'Portal user', 'Default', 'User' or something else?

yajo, line 11: Not sure if that would confuse users. Maybe we should set it as a customer. @sergio-teruel @carlosdauden?

sebalix, line 11: If it is set as a customer, the module name should be `auth_customer_supplier` or `auth_partners`, otherwise it is confusing.

sebalix:
A little remark, maybe mention it as a Known issue, but :+1:

auth_supplier/views/auth_supplier_view.xml
yajo, line 11: Renaming modules is a pain for migrations. I agree, but I don't see it's worth the effort.

sebalix, line 11: You are right. It was just to point out the problem.
In fact, a module `auth_signup_account_type` could be the proper solution, where `ir.config.parameter` entries like `auth_signup_account_type.{TYPE}` to enable them on the UI, and each `TYPE` could be created/overloaded to load related data => `customer` binded to `{'customer': True}`, etc, and everyone should be able to implement the type they want in addition to the `default` one which is just there for the default `auth_signup` behaviour.

A new module for later!

yajo, line 11: Indeed a better design, but I hope you understand this time this is just a forward port.

pedrobaeza, line 11: @sebalix, indeed it's a good idea. Maybe you can develop that options. We will gladly review them.

sebalix, line 11: @pedrobaeza I put that on my TODO list, maybe we will need that for a customer.
@Yajo of course, no problem to merge this module as-is ;)

PR #437, opened 2016-05-27
[ADD] base_search_improved: new module

dreispt:


base_search_improved/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+
+Usage
+=====
+
+On any search box, in list or kanban views, type your expressions
+and the magic will be automatically done.
+
+For example, in a demo database, on the Contacts / Customers try searching
+for "john brown". With this module installed will will see the result
```
pedrobaeza: remove one of the duplicated `will`

base_search_improved/README.rst
```diff
@@ -0,0 +1,93 @@
...
+
+===============
+Improved Search
+===============
+
+This module extends the search functionality to add simple shortcuts for
+more complex search conditions.
+
+Features:
+* Fuzzy search: use wildcards ("%") in place of spaces
```
pedrobaeza, line 13: You must use `*` instead of spaces for the wildcard, or even better, you are allowed to use also `%` in the searches, so better to teach users to use it, that to introduce a system that don't allow to make specific searches.

dreispt, line 13: `%` is already allowed in searched - I use that every time. It's true that many are more used to `*` and we could replace it with `%`.
IMO using the space as a wildcard can make the search much more friendly.

pedrobaeza, line 13: But it makes it more unpredictable. You lose control over your search (and it also has a performance penalty). It should be at least optional.

dreispt, line 13: It could be made optional.
Why do you say there is a performance penalty? 
AFAICS there is no significant additional burden to the SQL.

pedrobaeza, line 13: AFAIK search performance decreases with wildcards.

base_search_improved/models/base_model.py
```diff
@@ -0,0 +1,30 @@
...
+            order=None, count=False: value if count else value.ids
+
+        @api.returns('self', upgrade=upgrade, downgrade=downgrade)
+        def new_search(self, cr, user, args, offset=0, limit=None,
+                       order=None, context=None, count=False):
+            for i, cond in enumerate(args):
+                # fuzzy search: replace spaces with wildcards
+                if len(cond) == 3 and cond[1] == 'ilike':
+                    args[i] = cond[0], cond[1], cond[2].replace(' ', '%')
+            return new_search.origin(self, cr, user, args, offset, limit,
```
moylop260, line 27: What about the following special case: [Search by invoice.number](https://github.com/odoo/odoo/blob/5f9ca1b89becb6867d0755cf7cb3d7f9c533e7af/addons/account/models/account_invoice.py#L874-L877)?
- Data:
  - invoice1: 123 987
  - invoice2: 123 789 987
- Original behaviour:
  - Search: "123 987" returns invoice1
- New behaviour:
  - Search: "123 987" returns invoice1 and invoice2

dreispt, line 27: The wildcard replacement affects only 'ilike' operations. The '=' operation you point at wouldn't be affected.
But I see the concern on changing the global behaviour.
Maybe we could activate it based on a flag in the `context`.
It would then be easy to edit Window Actions to active it.

moylop260, line 27: ok, thanks for clarify

base_search_improved/README.rst
lasley, line 13: In terms of performance- really the only degradation is due to SQL needing to perform a table scan, instead of allowing index seek.

base_search_improved/models/base_model.py
lasley, line 27: I like this context idea. Use cases that immediately come to mind for this module are pretty record specific IMO, so defining it in the window action makes sense

base_search_improved/README.rst
```diff
@@ -0,0 +1,93 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
+
+Known issues / Roadmap
+======================
+
+Additional serach tokens are welcome.
```
elicoidal, line 52: search

dreispt:
This was completely rewritten and will be replaced by two other modules.
The first one is #439, the second one, for extended search, will come later.
Thanks for your reviews here!

PR #439, opened 2016-06-02
[ADD] base_name_search_improved: friendly and powerful  name search

dreispt:
Extends the name search feature to use fuzzy matching methods, and  allowing to search in additional related record attributes.


base_name_search_improved/README.rst
```diff
@@ -0,0 +1,96 @@
...
+   :alt: License: AGPL-3
+
+====================
+Improved Name Search
+====================
+
+Extends the name search feature to use fuzzy matching methods, and 
+allowing to search in additional related record attributes.
+
+The name search is the lookup feature to select a related rcord.
```
angelmoya: "related rcord." > "related record."

base_name_search_improved/README.rst
```diff
@@ -0,0 +1,96 @@
...
+Installation
+============
+
+No specific requirements.
+
+
+Configuration
+=============
+
+The fuzzy search is snabled on all Models
```
angelmoya: "snabled" > "enabled"

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,60 @@
...
+        def make_name_search():
+
+            @api.model
+            def name_search(self, name='', args=None,
+                            operator='ilike', limit=100):
+                # Regular name search
+                res = name_search.origin(
+                    self, name=name, args=args, operator=operator, limit=limit)
+
+                if not res and self._rec_name:
```
OSguard: i think it is better that you should only support 'positive' operator: like, ilike, = 
And i think 'negative' operator 'not ilike' and '!=' should not have this improvment

OSguard:
Did you check the Odoo Core test which fail?

base_name_search_improved/models/ir_model.py
dreispt: That should be a rare even, since name_search is seldom called directly with a specific operator.
But it is wise to add that. I can restrict only to `like`, `ilike` and `=`.

dreispt:
I'll check the tests.

base_name_search_improved/tests/test_name_search.py
```diff
@@ -0,0 +1,35 @@
...
+        model_partner.name_search_ids = phone_field
+        self.Partner = self.env['res.partner']
+        self.partner1 = self.Partner.create(
+            {'name': 'Johann Gambolputty of Ulm',
+             'phone': '555-555-555'})
+
+    def test_NameSearchPhone(self):
+        """Name Search on Phone field"""
+        res = self.Partner.name_search('555-555-555')
+        self.assertTrue(res)
```
moylop260: Please, Could you validate that the result returned is the expected?

base_name_search_improved/tests/test_name_search.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_NameSearchUnordered(self):
+        """Name Search on unordered words"""
+        res = self.Partner.name_search('ulm gambol')
+        self.assertTrue(res)
```
moylop260: same case

base_name_search_improved/tests/test_name_search.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_NameSearchFalsePositive(self):
+        """Name Search no False Positive"""
+        res = self.Partner.name_search('ulm 555-555')
+        self.assertFalse(res)
```
moylop260, line 44: <s>Same case</s> N/A

dreispt:
Tests fixed!

The algorithm was improved: to the two existing passes (ordered words, unordered words), a first one was added, for full text including spaces.
A test was added for this case.

moylop260:
Thanks!
👍

dreispt:
CI is now green.

moylop260:
Thanks @dreispt

josemoralesp:
I verified this PR with some models and the result was the expected. 

My first worry was the answer speed. I used some models with more than one hundred thousand records and the answer was quick, combining the different way to search added by this PR.
:+1:

yajo:
:+1:

yajo:
BTW fuzzy finding is you search _drt_ and get _dreistp_, it woud be nice too

dreispt:
@josemoralesp Thanks: I worried about the performance, thus the design where the extra searches are tried in layers, and only after the regular name search fails to find results. I'm glad that you confirmed that the performance is not significantly affected.

@yajo Indeed, that is a candidate for an extra search layer. That could go in a later improvement PR.

Can we merge?

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,61 @@
...
+                    for rec_name in [self._rec_name] + other_names:
+                        domain = [(rec_name, operator, name.replace(' ', '%'))]
+                        recs = self.search(domain, limit=limit)
+                        if recs:
+                            return recs.name_get()
+                    # Try unordered word search on each of the search fields
+                    for rec_name in [self._rec_name] + other_names:
+                        domain = [(rec_name, operator, x)
+                                  for x in name.split() if x]
+                        recs = self.search(domain, limit=limit)
```
yajo: With these tries you end up performing 3 separate searches, and results will be more unpredictable. Is it not better to just add all those tries to the domain and perform only 1 predictable query including all methods?

pedrobaeza:
You can check https://www.postgresql.org/docs/9.5/static/fuzzystrmatch.html for even "fuzzier" searches.

moylop260:
@josemoralesp 
Could you share your last implementation of fuzzy string match that commented @pedrobaeza?

base_name_search_improved/models/ir_model.py
dreispt: That would have the same effect of just using the last of the three searched (the most relaxed one).
The several layers try to improve the relevance of the results.
If I try "555 777" and there is such a number, I don't need the search to return me a "777 555" matching result.
But if not match to "555 777" was found, the algorithm goes for a more relaxed filer, and tried matching any phone containing "555" and "777", in that order, such as "555 xxx 777".
If it still doesn't find anything, then it tries that again, but ignoring word order, so that "777 xxx 555" would match.

Maybe you could try it with real data and confirm if the results are user friendly.

dreispt:
@pedrobaeza That's a good pointer. That could be added in another PR as an enhancement.

base_name_search_improved/models/ir_model.py
yajo: Yes, I refer exactly to what you say. I'd drop any of the stricter methods.

Imagine you have duplicated partners: John Travolta and Travolta John. If I search for "jo tra" I'd expect both to show up always.

What most search engines or fuzzy finders I tried do is _sort_ by best match, not actually _hide_ worse matches. I'd expect John Travolta to be above Travolta John, unless I search for "tra jo".

I think code will shrink and UX will improve if you just drop all the other methods. Well, it's just my opinion, of course. Not blocker.

josemoralesp:
I am agree with @pedrobaeza to use the string distance theory Levenshtein of postgres extension fuzzystrmatch.

I used this extension for create an script to join similar products imported from another system to Odoo, to facility the process to create the variants a their attributes. The result using that extension was the expected. 

Investigating about string distance I changed the theory used I decided use jaro string distance theory its the best of all theory of string distance(as I read and I experience what I used). 

My steps used to add the extension were:

apt-get install postgresql-server-dev-all
cd /usr/local/src
git clone https://github.com/eulerto/pg_similarity.git
cd pg_similarity
export USE_PGXS=1
make
make install

Then in postgres you need to execute the following line

CREATE EXTENSION pg_similarity;

I used the following way:

SELECT id FROM product_template WHERE jaro(field_in_table, string_to_compare) > 0.856;

The result ranges returned by jaro method  are from 0 to 1(statistics). Where 1 is when the strings compared are the same.

To read more about Jaro distance [here](https://www.wikiwand.com/en/Jaro%E2%80%93Winkler_distance).

I hope this information can help.

Best Regards.

base_name_search_improved/models/ir_model.py
dreispt: It won't shorten the code, it will increase it, since I would have to collate the several results.
But I like your "sort results by best match" suggestion.
I'll implement that.

pedrobaeza:
I'm using it here: https://github.com/OCA/crm/pull/99

dreispt:
Just added a new commit, implementing @Yajo suggestion and adding @pedrobaeza suggestion to the roadmap.

PS: stepping back and thinking about workflows, the feature improvement commit should have been a second PR, don't you think?

jholze:
Hi @dreispt do you think this will work performant with a DB of 100.000 partners ? We think about using it in a big project.

dreispt:
@jholze The design was made with performance concerns in mind: the additional search strategies only kick in when the "standard" one yields no results. But it needs battle testing to confirm it's performance. If using additional fields to search, you should make sure that they are db indexed.

base_name_search_improved/models/ir_model.py
```diff
@@ -20,35 +20,38 @@ def make_name_search():
...
+                        domain += [('id', 'not in', [x[0] for x in results])]
+                        recs = self.search(domain, limit=limit - len(results))
+                        results.extend(recs.name_get())
+                    return results
                 # Regular name search
                 res = name_search.origin(
                     self, name=name, args=args, operator=operator, limit=limit)
-
+                # Extended name search is only used on some operators
                 allowed_ops = ['ilike', 'like', '=']
```
yajo: If you use a set here, you'd get a ridiculous performance boost in the next sentence.

base_name_search_improved/models/ir_model.py
```diff
@@ -20,35 +20,38 @@ def make_name_search():
             @api.model
             def name_search(self, name='', args=None,
                             operator='ilike', limit=100):
+                # Reusable parts
+                def extend_name_results(domain, results, limit):
+                    if len(results) < limit:
+                        domain += [('id', 'not in', [x[0] for x in results])]
+                        recs = self.search(domain, limit=limit - len(results))
```
yajo: If you store `len(results)` in a variable, you get another ridiculous performance boost avoiding to count them twice.

yajo:
:+1: code review

base_name_search_improved/models/ir_model.py
dreispt: Yes, and I could define the set outside this function, so that it's one less op to perform each time.

base_name_search_improved/models/ir_model.py
dreispt: I don't think I can: the `result` variable is potentially different on each call here.

yajo: I mean before line 25

dreispt: I see now. I can do that.

base_name_search_improved/README.rst
```diff
@@ -28,16 +29,18 @@ For example, Customers could be additionally searched by City or Phone number.
...
-But if no results are found, then sdditional search methods are tried until
-some results are found. The sepcific methods used are:
+But if not enough results are found, then additional search methods are tried.
+The specific methods used are:
 
 - Try regular search on each of the additional fields
 - Try ordered word search on each of the search fields
 - Try unordered word search on each of the search fields
 
+All results found are presented in that order, hopefully prenting them ordered
```
yajo: prenting?

base_name_search_improved/models/ir_model.py
```diff
@@ -20,35 +33,29 @@ def make_name_search():
...
                     for rec_name in other_names:
                         domain = [(rec_name, operator, name)]
-                        recs = self.search(domain, limit=limit)
-                        if recs:
-                            return recs.name_get()
+                        res = extend_name_results(domain, res, limit)
                     # Try ordered word search on each of the search fields
-                    for rec_name in [self._rec_name] + other_names:
+                    for rec_name in all_names:
                         domain = [(rec_name, operator, name.replace(' ', '%'))]
```
yajo, line 60: `%` would fail if operator is `=`

jgrandguillaume:
:+1: functional test ok. Great module !

lfreeke:
:+1: Functional test.

dreispt:
Introduced a few fixes and improvements, including the ones suggested by @Yajo.
Also made the searched field list cached.
AFAICS you now need a restart for changes on searched fields to take effects.
I tested in a real db with 20k Partners and the performance looks good.

jholze:
@dreispt Sounds good. We have similar tickets in the current sprint to add fuzzy like search in some frontend and backend fields. We already did a solution, but we used a plugin to have fuzzy search within PostgreSQL instead of odoo. We have a database with around 60k of partners. Looks like a good testing stack :)

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,76 @@
...
+from openerp import models, fields, api
+from openerp import SUPERUSER_ID
+from openerp import tools
+
+
+# Extended name search is only used on some operators
+ALLOWED_OPS = set(['ilike', 'like'])
+
+
+@tools.ormcache(skiparg=0)
```
yajo, line 14: AFAIK skiparg is not used in new api methods.

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,76 @@
...
+    return rec_name + other_names
+
+
+def _extend_name_results(self, domain, results, limit):
+    result_count = len(results)
+    if result_count < limit:
+        domain += [('id', 'not in', [x[0] for x in results])]
+        recs = self.search(domain, limit=limit - result_count)
+        results.extend(recs.name_get())
+    return results
```
yajo, line 29: Why not put these 2 methods as `@api.model` to let them be extended or reused easily?

base_name_search_improved/models/ir_model.py
dreispt, line 14: I read the underlying implementation, and AFAICS the skipargs is needed.

base_name_search_improved/models/ir_model.py
dreispt, line 29: These is just helper code, wrapped in function for convenience.
They are not `ir.model` methods, since the `self` parameter passed to them is the searched Model, and not `ir,.model`.

base_name_search_improved/models/ir_model.py
yajo, line 14: Ah sorry, you're right. I started using it in v9 and got used to the new system.

yajo:
I can repeat my :+1: code review

christophlsa:
As @jholze wrote we currently use a fuzzy search in our frontend. There we use the pg_trgm extension and search with the operator "%" with sorting the results with "similarity()". This solution was written in plain sql. Now we want to use this functionality also in the backend and also in a general solution like proposed in this PR. I thought about to extend your solution to maybe implement a new search parameter (maybe also "%"). But to use this operator the corresponding field should have a "gin" index. That's why I think the solution should add this feature only for specific char/text fields.

Do you think it's a good solution to add such a new operator and how would you implement the index generation? Currently you can set the index=True parameter to the definition of the field. I'm not sure if I can also add such new options in field definitions. What do you think about this?

Sorry if this is not the right place but I think we want to do some similar thing. :)

dreispt:
@christophlsa I believe you are referring to the `search` function, rather than `name_search` extended here (use only when filling in related values). I have plans for a similar module extending `search` with additional "operators", but that is a different work.

However, AFAIK the "%" operator is already supported for "ilike" searches.
So today, searching for like "Johann%Ulm" would match "Johann Gambolputty of Ulm".
If I'm right, no point for a new operator.

dreispt:
Can we merge this?

yajo:
Please :wink:

lasley:
Code review 👍

christophlsa:
@dreispt Yes, I would change the search method. But "%" is not the same as LIKE. It comes from the trgm module: https://www.postgresql.org/docs/current/static/pgtrgm.html

With this operator you can for example search for `WHERE name % 'Jon Moller'` and you can find entries like "John Miller". But you have to first create a GIN index for this column. That's why I'm asking if you or someone else has an idea how you would implement the creation of an additional index by adding for example new parameters to the fields.

dreispt:
I see. That could be added here, an as additional name search layer. It is in the roadmap, but Pedro [suggested](https://github.com/OCA/server-tools/pull/439#issuecomment-224843944) using [fuzzystrmatch](https://www.postgresql.org/docs/9.5/static/fuzzystrmatch.html).

My plan to extend the `search` method is to interpret the first character ins a `ilike` search and give it special meaning. So, we could use '<' or '>'. Adding a '%' for fuzzy search would also work.

PR #440, opened 2016-06-02
[9.0][FIX][auth_supplier] Right permissions.

yajo:
Right now a normal user cannot manage supplier portal access from the wizard because group was not marked as `is_portal`.

With this fix, you can enable supplier portal for a given partner.

BTW some unneeded code has been removed.

@Tecnativa

### DIFF
--- a/auth_supplier/__openerp__.py
+++ b/auth_supplier/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': "Auth Supplier",
-    'category': 'Tools',
+    'category': 'Portal',
     'version': '9.0.1.0.0',
     'depends': [
         'auth_signup',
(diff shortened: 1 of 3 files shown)


auth_supplier/__openerp__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     'name': "Auth Supplier",
-    'category': 'Tools',
+    'category': 'Portal',
     'version': '9.0.1.0.0',
```
pedrobaeza: Change version to 9.0.1.1.0

pedrobaeza:
Minor thing, but :+1:

moylop260:
👍

sebalix:
:+1:

PR #442, opened 2016-06-03
[9.0][IMP][auth_supplier] Remove useless group. Mark as customer.

yajo:
The whole purpose of this module was to be able to separate a customer and a supplier portals.

Given the status of the upstream implementations of portals, it is not possible to have a supplier portal without portal permissions. Thus, it would be required to make the group this module was creating inherit from `base.group_portal`, which would give suppliers access to all customers portal. As such, that group becomes useless.

From now on, this distinction should be made with simply checking if `user.customer` or `user.supplier` is `True`. To allow that, I'm implementing what was discussed at https://github.com/OCA/server-tools/pull/434#discussion-diff-64883758.

Tests had to be removed because https://github.com/odoo/odoo/issues/12237 makes the created record to disappear, so there's no way to actually check the created user was created with customer or supplier flag on.

@Tecnativa


pedrobaeza:
:+1:

moylop260:
👍

PR #443, opened 2016-06-03
Add base_import to dependencies of base_technical_features

ghost:
In Odoo 9.0 base_import/tests/test_cases.py line 90, there is a test named test_shallow.

This test takes for granted that the user admin does not have the group base.group_no_one.

So if you start your server and run the tests like this:

./start_odoo -d my_db_9.0 -i base_technical_features,base_import --log-level test --test-enable

The test fails because base_technical_features adds base.group_no_one to the user admin, which is 100% logical by the way.

I wish to add base_import to the dependencies of base_technical_features so that the unit tests of base_import do not fail and so that my ci server is happy.

### DIFF
--- a/base_technical_features/__openerp__.py
+++ b/base_technical_features/__openerp__.py
@@ -13,6 +13,9 @@
         'views/res_users.xml',
         'data/res_users.xml',
     ],
+    "depends": [
+        'base_import',
+    ],
     "license": "AGPL-3",
     "installable": True,
 }


joaoalf:
lgtm :+1:

moylop260:
👍 
I saw that the module `base_import` is `auto_install: True` based on `web`
And `web` is `auto_installed` based on `base`

lmignon:
:-1: Sorry but I'm not in favor of adding a dependency to a module to solve a problem that occurs when you run tests due to the way you run your tests. In any way you can't predict the order in which Odoo will install the modules required by you server. Therefore, in a lot of cases, when some modules introduce some constrains not expected by others, it's common to have tests failing.To avoid/minimize this kind of error you should run your tests in 2 steps as it's done by MQT in travis.
1. install all the dependencies of the modules to test  `./start_odoo -d my_db_9.0 -i base` **without** `--test-enable`
2. intstall the modules to test `./start_odoo -d my_db_9.0 -i base_technical_features --test-enable`

ghost:
@lmignon True, this is a better idea.

:-1: for myself

PR #444, opened 2016-06-04
[ADD] odoolint: Runtime lint module

moylop260:
This module is result of the following [previous discuss](https://github.com/OCA/maintainer-quality-tools/issues/331)

After merge:
- [ ] PR for MQT to auto-install this module and process the warning messages.
- [ ] Migrate it to version 9.0


dreispt:
Isn't this best as a PR for MQT?

moylop260:
@dreispt 
Thanks for feedback.

It's a module then if we create a PR for MQT, after to use it in our local we'll need download all MQT structure to install it

dreispt:
MQT has a small footprint, I don't see the issue.
I argue the opposite: keeping it in server-tools will force it as a dependency for all OCA repos, when we start using it as an additional layer for CI checks.

moylop260:
Ok @dreispt thank you
I can move it without problem...
Maybe the middle point is a new project.

But we could wait the feedback from other people to take a decision

dreispt:
Yes, let's hear from others.
As for having it in it's own repo, I don't see the point:
MQT is there for that, and all build already clone and use it...

pedrobaeza:
Maybe this deserves a specific repo...

odoolint/README.rst
```diff
@@ -0,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========
+odoolint
+========
+
+This module add runtime lint checks that don't are detected by odoo.
```
ghost: that are not detected

odoolint/models/ir_model_data.py
```diff
@@ -0,0 +1,128 @@
...
+                self.name, self.module_real, self.file_name,
+                imd_new['module_real'], imd_new['file_name'],
+            )
+            return False
+        return True
+
+    @api.multi
+    def _check_xml_id_unachievable(self, xmlid):
+        """Check a unachievable xml_id referenced
+        """
```
ghost, line 78: @moylop260 can you explain more this method. I'm not sure what it does? Please add a more detailled docstring

moylop260, line 78: Thanks for feedback!
Fixed!

odoolint/README.rst
moylop260: Thanks for feedback!
Fixed!

moylop260:
FYI I added tests and support for `auto_install` cases and we are using it in [vx/mqt](https://github.com/vauxoo/maintainer-quality-tools) production.

I'm waiting for decision of best location for modules used in MQT.

@lepistone @guewen @gurneyalex 
Your feedback is welcome

dreispt:
For me this is clearly MQT. CI tested repost all clone MQT and then use it's scripts and tools for the tests. From a functional PoV, is is closely related to MQT. From a technical/practical PoV having it there is the simplest solution.

moylop260:
@dreispt What is the better folder for this module if we decide to move it for MQT?

moylop260:
@dreispt FYI https://github.com/OCA/maintainer-quality-tools/pull/336

moylop260:
Following the recommendation of @dreispt I migrated the module to MQT
However, I have many question yet...

1) What is the folder to use?
2) How we will use the different version 8.0, 9.0... of the module in MQT branch master?
3) For all Is a good idea download a full MQT project to use this module in our local?

dreispt:
Regarding 3), the vision is that MQT is a command line tool to run tests. If so, people would download it for all it's features, and this additional checking would be just one more.

moylop260:
Im agree with you
Indeed, that is the reason for what I dont see a odoo module in MQT as a good idea.
The runtime check is a odoo module that needs full environment to detect comon error easiest.
Like as demo referenced from data or unreachable xml_id with my current tree dependencies.

A good place for this module could be server-tools but the main problem tha you saw is a full download of other modules from mqt. Im agree.
The other solution is add it to mqt directly but the main problem is the version management of the module and the out of the scope of have a command line tool.

The mix solution is a repository for modules used from MQT.

By example the
odoo-profiler to get python stats to know tha methods are slow or many times called.
Odoo-pgbadger to get postgresql stats to know that sql are slow or many times executed.
doolint to get runtime checks where a full environment is required and core methods used.

What do you think?

moylop260:
@dreispt FYI We are waiting for you

dreispt:
@moylop260 I'm sorry, I understand that discussion here is closed and https://github.com/OCA/maintainer-quality-tools/pull/336 is a WIP. Do you need my comments there?

PR #446, opened 2016-06-06
[FIX] fix broken menuitem reference for Audit on odoo 9

holdenrehg:
The `menu_audit` menu item was broken on odoo 9 even though this module is listed under the 9.0 branch on the repo.

I updated it to fall under `base.menu_custom` instead of `base.menu_reporting`.

### DIFF
--- a/auditlog/views/auditlog_view.xml
+++ b/auditlog/views/auditlog_view.xml
@@ -3,7 +3,7 @@
     <data>
 
         <menuitem id="menu_audit" name="Audit"
-            parent="base.menu_reporting" sequence="50"
+            parent="base.menu_custom" sequence="50"
             groups="base.group_system"/>
 
 


moylop260:
👍

lasley:
Thanks for the submission @holdenrehg 

It looks like this module has not yet been upgraded to v8, as can be seen in the manifest file's version attribute: https://github.com/OCA/server-tools/blob/9.0/auditlog/__openerp__.py#L24 - so I have to 👎 

Please take a look at the OCA migration guide for info on how to complete the migration of this module - https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0

holdenrehg:
Thanks @lasley , I'll walk through that and migrate this module over if no one else is in the process already.

PR #450, opened 2016-06-09
[PORT] module secure_uninstall to v8

bealdav:
Odoo erase all data managed by uninstalled modules.

To prevent catastrophe this module ask password before proceeding to module uninstallation.


secure_uninstall/README.rst
```diff
@@ -0,0 +1,68 @@
...
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================
+Secure Uninstall
+================
+
+Ask Master Password ('admin_passwd' key from config file)
+before to proceed to module uninstallation
```
pedrobaeza: s/before to proceed to/before proceeding with

secure_uninstall/README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+
+Configuration
+=============
+
+Check 'admin_passwd' is set in Odoo configuration file
```
pedrobaeza: s/Check/Check that

secure_uninstall/README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+Usage
+=====
+
+Try to uninstall a module
```
pedrobaeza:
```
#. Go to Settings > Local modules.
#. Click on one module.
#. Click on Uninstall button.
#. You will need to enter admin password before proceeding.
```

secure_uninstall/i18n/secure_uninstall.pot
```diff
@@ -0,0 +1,87 @@
+# Translation of Odoo Server.
```
pedrobaeza: Don't put POT file. It's automatically generated.

pedrobaeza:
Good icon!

secure_uninstall/models/module.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def button_uninstall(self, cr, uid, ids, context=None):
+        # can't be convert in new api because use button_uninstall fct
+        # in _button_immediate_function() is buggy in new api.
+        # Next v10 will solve it.
+        if not context:
+            context = {}
+        if 'uninstall_authorized' in context:
+            ctx = context.copy()
+            del ctx['uninstall_authorized']
```
pedrobaeza: Why do you remove this context key? I think it's not needed.

secure_uninstall/models/module.py
```diff
@@ -0,0 +1,72 @@
...
+            super(Module, self).button_uninstall(
+                cr, uid, ids, context=ctx)
+            return self._button_immediate_function(
```
pedrobaeza: Why do you need to manually execute this method?

secure_uninstall/README.rst
```diff
@@ -0,0 +1,68 @@
...
+help us smashing it by providing a detailed and welcomed feedback.
+
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
sebalix: To update as the icon has changed.

bealdav: HI @sebalix, I don't know what it has changed. This is the exact copy of maintainer tools README template ?

secure_uninstall/models/module.py
bealdav: Because if I let this key, I obtain:
`maximum recursion depth exceeded while calling a Python object` traceback

secure_uninstall/models/module.py
bealdav: If i don't do that, module fall in 'To be removed' state. I want uninstall it immediately and I don't want to 'Apply Scheduled Upgrades' menu ;-)

secure_uninstall/README.rst
sebalix: I was thinking about the icon used in this module which is not the OCA default one, so the link used here is meaningless?

bealdav:
Failed travis test comes from `fetchmail_notify_error_to_sender`module

```
Field(s) `arch` failed against a constraint: Invalid view definition
Error details:
Field `error_notice_template_id` does not exist
Error context:
View `fetchmail.server.form`

[view_id: 378, xml_id: fetchmail_notify_error_to_sender.view_email_server_form, model: fetchmail.server, parent_id: 242]" while parsing /home/travis/build/OCA/server-tools/fetchmail_notify_error_to_sender/fetchmail_view.xml:5, near

<record model="ir.ui.view" id="view_email_server_form">
```
No problem with `secure_uninstall`

secure_uninstall/models/module.py
```diff
@@ -0,0 +1,44 @@
...
+
+from openerp import models, fields, api
+from openerp.exceptions import Warning as UserError
+from openerp.tools.config import config
+
+
+def _get_authorized_password():
+    """ You can define your own authorized keys
+    """
+    return [config.get("secure_uninstall")]
```
rvalyi: Please could we have instead: [config.get("secure_uninstall"), config.get("admin_passwd")] ?

Rationale: using admin_passwd only could be a security issue because admin_passwd has a lot of power and not all installations provide it (Odoo SA SaaS don't for instance), this is why a new password such as secure_uninstall is required indeed. But on the other hand, in an installation where we do provide an admin_passwd, then the user who knows this password can already totally screw the installation if he wishes, so I think we should also permit him to uninstall any module using this same admin_passwd without having to set up and communicate yet another password in this case.

sylvainc:
very useful, thanks! :+1:

rvalyi:
:+1: thanks!

EBII:
efficace

moylop260:
Could you check the beta lint messages from build # 1 of travis?
https://travis-ci.org/OCA/server-tools/jobs/154802840#L442-L443

``` bash
secure_uninstall/models/module.py:27: [C8107(translation-required), BaseModuleUpgrade.upgrade_module] String parameter of raise "UserError" requires translation. Use _("Missing configuration key\n--------------------\n'secure_uninstall' configuration key is not set in \nyour Odoo server configuration file: please set it a value")
secure_uninstall/models/module.py:34: [C8107(translation-required), BaseModuleUpgrade.upgrade_module] String parameter of raise "UserError" requires translation. Use _("Password Error\n--------------------\nProvided password '%s' doesn't match with 'Master Password'\n('secure_uninstall' key) found in the Odoo server configuration file .\n\nResolution\n-------------\nPlease check your password and retry or cancel")
```

moylop260:
LGTM (small comment)

sebastienbeau:
:+1: LGTM code review

bealdav:
@moylop260 fixed by https://github.com/OCA/server-tools/pull/547
thanks to notice me

PR #454, opened 2016-06-12
[MIG] Migrate auditlog module from 8.0 to 9.0

holdenrehg:
The majority of the module was already following new api conventions and not using any deprecated methods, so there was only minor clean up to move to over to 9.0
- Update documentation to point to the new auditlog menu locations. These were changed because the 8.0 version was referencing menus that do not exist in 9.0
- Change version from 8.0.1.0.0 to 9.0.1.0.0
- Make the module installable again
- Remove an unused parameter from pre-migration.py for versioning
- Update classes to follow OCA new api conventions, camel case instead of snake case
- Fix typos and remove commented out blocks of code that were irrelevant


holdenrehg:
Here is a reference to my CLA signature https://github.com/odoo/odoo/pull/12305

moylop260:
@holdenrehg 
Thanks for you contribution.

> Here is a reference to my CLA signature odoo/odoo#12305

This one is CLA of odoo. Could you sign your CLA of oca https://odoo-community.org/page/cla too?

moylop260:
Could you check the [license](https://travis-ci.org/OCA/server-tools/jobs/137134463#L438)?
- `auditlog/__openerp__.py:22: [C8102(manifest-required-key), ] Missing required key "license" in manifest file`

More info [here](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules)

holdenrehg:
@moylop260 Checked and updated the license field. [See here](https://github.com/OCA/server-tools/pull/454/commits/233bb87ba5ef636cfd6ef95ab5b955d4f1293a4a) for the commit.

moylop260:
Thank you
After CLA detail... LGTM

hbrunn:
you should follow the procedure outlined in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#howto to get 8.0's commits. This allows us to simply cherry pick changes afterwards and keeps history intact.

holdenrehg:
Sounds good @hbrunn I'll pull in that 8.0 history and update the PR.

holdenrehg:
@hbrunn It appears that all the 8.0 commit history is already there. There was an updated translations file that I pulled from the 12th after I migrated the module.

hbrunn:
you seem to run the commands mentioned above on your current branch, this won't give you the desired results. You need to start out from an upstream 9.0 branch, and then do do the stuff so that the commit history includes all of those: https://github.com/OCA/server-tools/commits/8.0/auditlog

I also see that there are some commits that make the module better, but are not reflected in this PR, like https://github.com/OCA/server-tools/commit/3f9152ccbefabe3b4802e2944ab5040eb0fad106

holdenrehg:
@hbrunn So is the proper way to perform the actions on an upstream branch and then merge into this PR? It will add 70+ commits to this PR. Want to make sure I'm handling this the correct way according to OCA conventions.

hbrunn:
yes, that's exactly what we want

Just do all of this in a fresh branch and push overwrite this one. Or just close this PR and create a new one with your new branch.

holdenrehg:
Ah thanks @hbrunn

auditlog/README.rst
```diff
@@ -1,29 +1,42 @@
...
-Then, check logs in the `Reporting / Audit / Logs` menu.
 
-During installation, it will migrate any existing data from the `audittrail`
-module (rules and logs).
+Then, check logs in the `Settings / Technical / Audit / Logs` menu. You can group them by
+user sessions, date, data model or HTTP requests:
+
+Get the details:
+
+.. image:: /auditlog/static/description/log.png
```
sebalix, line 23: I set all these image URLs in my last PR, but it seems they do not work properly on apps.odoo.com (https://apps.odoo.com/apps/modules/8.0/auditlog/). We should change them to relative ones like `static/description/IMAGE.png` in order to have them displayed.

auditlog/__openerp__.py
```diff
@@ -1,38 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': "Audit Log",
-    'version': "8.0.1.0.0",
+    'version': "9.0.1.0.0",
     'author': "ABF OSIELL,Odoo Community Association (OCA)",
+    'license': "AGPL-3",
     'website': "http://www.osiell.com",
+    'license': 'AGPL-3',
```
hbrunn: two is a crowd already: https://travis-ci.org/OCA/server-tools/jobs/137559872#L433

hbrunn:
runbot error unrelated

auditlog/README.rst
holdenrehg, line 23: Relative paths do not work when testing. I'll leave these as absolute for now.

holdenrehg:
@hbrunn stripped out that duplicated license key, also thanks for mentioning unrelated runbot errors. Saw the same, related to other modules.

sebalix:
Tested on Runbot, :+1:

auditlog/migrations/8.0.1.0/pre-migration.py
```diff
@@ -1,26 +1,9 @@
...
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2015 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from openerp.addons.auditlog import migrate_from_audittrail
 
 
-def migrate(cr, version):
+def migrate(cr):
```
sebalix: I did not see this. We should not change the signature of the method here, even if the version parameter is not used. It seems that the migrate method is always called with this parameter: https://github.com/odoo/odoo/blob/9.0/openerp/modules/migration.py#L152

pedrobaeza: In fact, this file should be totally removed, as the migration doesn't apply to the version 9.

sebalix: It's right.

auditlog/models/http_request.py
```diff
@@ -0,0 +1,60 @@
...
+from openerp.http import request
+
+
+class AuditlogHTTPRequest(models.Model):
+    _name = 'auditlog.http.request'
+    _description = u"Auditlog - HTTP request log"
+    _order = "create_date DESC"
+    _rec_name = 'display_name'
+
+    display_name = fields.Char(u"Name", compute="_display_name")
```
moylop260, line 15: `s/_display_name/_compute_display_name/g`

hbrunn, line 15: any reason not to overwrite `name_get` instead? This will be the display name by default, with the added benefit that `name_get` also returns something sensible. Then you also don't need to set `_rec_name`. cf https://github.com/OCA/OCB/blob/9.0/openerp/models.py#L1664

holdenrehg, line 15: What's better in this scenario? `name_get` seems to be less elegant if you still keep the computed `display_name` field. If you get rid of `display_name` and compute directly in `name_get` that has worse performance since it will be computed on every view load that has this model defined.

``` python
class AuditlogHTTPRequest(models.Model):

    ...

    display_name = fields.Char(u"Name", compute="_compute_display_name")

    @api.multi
    @api.depends('create_date', 'name')
    def _compute_display_name(self):
        for httprequest in self:
            create_date = fields.Datetime.from_string(httprequest.create_date)
            tz_create_date = fields.Datetime.context_timestamp(
                httprequest, create_date)
            httprequest.display_name = u"%s (%s)" % (
                httprequest.name or '?',
                fields.Datetime.to_string(tz_create_date))

    @api.multi
    def name_get(self):
        res = []
        for request in self:
            res += [(request.id, request.display_name)]
        return res
```

sebalix: Is this a blocking point to migrate on 9.0? At best rename `_display_name` to `_compute_display_name`.

sebalix (approved):

pedrobaeza:
Please rebase the branch to make it mergeable

pedrobaeza:
Superseeded by #603

Issue #348, opened 2016-02-03
Map User Groups broken on module users_ldap_groups with Odoo 8


clopez:
Hi,

We have an Odoo server 8 configured with `auth_ldap` and `users_ldap_groups`.

In order to map users to Odoo groups from an LDAP attribute we use the feature `Map User Groups` of the module `users_ldap_groups`.

This feature is working correctly on the first login of the user (when the user is created). For example, using the following rules

```
LDAP attribute 'employeeType' contains value 'localworker' then assing to OpenERP group LocalWorkers
LDAP attribute 'employeeType' contains value 'remoteworker' then assing to OpenERP group RemoteWorkers
```

Everything works as expected on the first login of the user. It is mapped to the right group.

However, if we change this value on the LDAP server, the group is not changed after the user logs outs and logs in again. We tried restarting the odoo server. Didn't helped. Only deleting the user from Odoo makes the thing work.

I have been debuggin the issue, and it seems that `get_or_create_user()` is only executed when the user is created (first login), but not when the user already has an entry on the Odoo server

hbrunn:
did you check the box 'Only ldap groups'? If not, this is the expected behavior

hbrunn:
Note: this flag causes all group memberships being wiped on login and reconstructed from ldap group mappings, so you need a group mapping for all groups the user should have

clopez:
Yes, I checked it. I also put a log in the first line of `get_or_create_user()` that is where such function should be called to check if at least that function was being invoked or not. And is only executed the fist time the user logs (new user)

hbrunn:
Odoo's algorithm here is that it first uses the given credentials to do a local login, and if this works, never arrives at the ldap part, cf. https://github.com/OCA/OCB/blob/8.0/addons/auth_ldap/users_ldap.py#L252 - this means it's crucial that the local login fails, any possibility the user in question gets her local password changed to the ldap password by some mechanism? Check the user's `password` and `password_crypt` columns in the database.

hbrunn:
in case this happens because the users "change" their passwords via Odoo's UI, https://github.com/OCA/server-tools/pull/287 might come to rescue

clopez:
So, if I undertand this right. When an user is created via `auth_ldap` their credentials are stored inside Odoo's database and the next time it logs Odoo don't checks the LDAP but the Odoo database for the password??? 

I have tested to change the password from inside the Odoo web interface for an user that was automatically created by the `auth_ldap` module.

[...]

PR #455, opened 2016-06-15
only ldap groups fix

luc-demeyer:
Without this fix, the "map_groups" is never called for existing users, hence a change of access rights in LDAP is not reflected in the odoo groups.


hbrunn:
the function is never called for existing users _with a password in the odoo database_. The local login _must_ fail, only then the user is considered a ldap user. Clear out your passwords in the database, then the code works unaltered.

luc-demeyer:
@hbrunn,
please reopen since your statement is not correct, at least not for the current version of Odoo 8:

cf. v8 official addons, auth_ldap/users_ldap.py, line 251:

```
def _login(self, db, login, password):
    user_id = super(users, self)._login(db, login, password)
    if user_id:
        return user_id
```

Hence the get_or_create_user call is never executed.

hbrunn:
yes, and what do you think is the reason a user id comes back? Because the login worked locally, so it's not a ldap account from the point of view of odoo. As said before: Clear out your local passwords in res_users, then you get an actual ldap login and then you're fine

luc-demeyer:
@hbrunn 
Please close a PR only after the issue is clearly identified and fixed.

More in depth: 
I checked the res_users password field before making the fix.
Clearing out the password from the db also doesn't apply to the specific project I made this patch for since we work with 'Create User' set to True and no 'Template User'. 
In such a case the first connection works fine with the current OCA module, the user gets created with the correct security groups. At the second login the 'get_or_create_user' call is never entered into.

Now reading the code:

the auth_ldap does a super of the _login method.
In the base _login method the '_check_credentials' is called, this one will indeed never get into the ldap credential checking if the standard login/password is succesful.
When this standard login fails (e.g. because of cleared out password or newly created LDAP user), the code performs the ldap_authenticate call (cf. auth_ldap, users_ldap.py, check_credentials method) and if that call is succesful it returns (no raise -> final result is user_id being returned after def _login and no call to 'get_or_create_user').

hbrunn:
wow, I'm so sorry, din't look good enough at the changes in auth_ldap. My apologies! :+1:

lanouettea:
I am experiencing the same issue and would be voting for this pull request. Users gets created with the correct groups.  But the call to map the groups never gets called after initial user creation.

I confirm that the password are NOT saved in database for any user. I stepped through the login process with pdb and see the comportment described by @luc-demeyer. I also believe that issue #348 is caused by the same issue.

luc-demeyer:
we are currently running with this fix on 3 production databases, hence it's safe to merge.

sebalix (approved):
LGTM, :+1:

yvaucher (approved):
LGTM

PR #456, opened 2016-06-15
super_calendar: Improve images in README, to show in apps

eLBati:
see https://apps.odoo.com/apps/modules/8.0/super_calendar/

### DIFF
--- a/super_calendar/README.rst
+++ b/super_calendar/README.rst
@@ -23,13 +23,15 @@ phone calls, you can create the following lines
 
 Meetings:
 
-.. image:: /super_calendar/static/description/meetings.png
-   :width: 400 px
+.. figure:: static/description/meetings.png
+   :alt: meetings
+   :width: 600 px
 
 Phonecalls:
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why figure?

eLBati:
@pedrobaeza this is the only way I found to make it work on both apps.odoo.com and github. I saw mis_builder follows this way: https://raw.githubusercontent.com/OCA/account-financial-reporting/8.0/mis_builder/README.rst

pedrobaeza:
OK, I didn't know. 

Maybe we should include an example of this in the template in https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

elicoidal:
and I would add the px marker as well.
IMO 400 is maybe too small and 800 too big so I suggest 600

pedrobaeza:
@eLBati are you going to take into account @elicoidal's suggestions?

eLBati:
@pedrobaeza I should be able to do them next week

eLBati:
@pedrobaeza @elicoidal done, thanks

PR #457, opened 2016-06-15
[9.0][MIG] auth_from_http_remote_user

lmignon:
Migration from 8.0


sbidoul:
The DEBUG logging mentioned in the README is not there in this version. It should be re-added or removed from the README.

Otherwise :+1: (manual test and code review)

StefanRijnhart:
:+1:

moylop260:
👍 
Do you want fix the following lint?

``` bash
auth_from_http_remote_user/tests/test_res_users.py:27: [C8104(class-camelcase), test_res_users] Use `CamelCase` "TestResUsers" in class name "test_res_users". You can use oca-autopep8 of https://github.com/OCA/maintainer-tools to auto fix it.
```

lmignon:
good catch @moylop260 . Change done.

sbidoul:
The runbot error was fixed in #464 so I think we can merge.

lmignon:
@pedrobaeza This one is ready to merge....

Issue #168, opened 2015-05-19
Migration to version 8


max3903:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-8
# Modules to migrate
- [x] auth_admin_passkey
- [ ] auth_from_http_basic
- [ ] auth_from_http_basic_logout
- [x] base_external_dbsource by @Lauris-v in #41 
- [ ] base_optional_quick_create
- [ ] configuration_helper
- [x] cron_run_manually
- [x] database_cleanup
- [x] dbfilter_from_header
- [ ] defer_parent_store_computation_import
- [x] disable_openerp_online
- [ ] document_export_from_db
- [ ] email_template_dateutil
- [ ] email_template_template
- [x] fetchmail_attach_from_folder
- [x] import_odbc by @Lauris-v in #41
- [x] mail_environment
- [x] mass_editing
- [ ] module_parent_dependencies
- [x] scheduler_error_mailer
- [ ] sentry_logger
- [x] server_environment
- [ ] server_environment_files
- [ ] sql_export #351 
- [x] super_calendar by @alejandrosantana in #139 
- [ ] tree_view_record_id
- [ ] users_ldap_groups
- [x] users_ldap_mail
- [x] users_ldap_populate
- [ ] web_context_tunnel
# Obsolete and deprecated modules

pedrobaeza:
Closing as this is not in use anymore.

PR #41, opened 2014-08-21
base_external_dbsource and import_odbc modules adopted for v8.0

Lauris-v:
Moved both modules out of `__unported__` directory and reworked form views for both modules as they were not displayed correctly.
No other changes were made because functions were working as supposed.

PR #461, opened 2016-06-21
Make import_odbc installable

mileo:
As mentioned in migration  https://github.com/OCA/server-tools/issues/168 and pr https://github.com/OCA/server-tools/pull/41, This module should be installable.

### DIFF
--- a/import_odbc/__openerp__.py
+++ b/import_odbc/__openerp__.py
@@ -95,8 +95,8 @@ Contributors
         'import_odbc_demo.xml',
     ],
     'test': [],
-    'installable': False,
-    'active': False,
+    'installable': True,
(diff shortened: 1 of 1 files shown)


dreispt:
Part of the migration work is bring it up to latest code standards. So I guess the PR will need additional clean up work.

mileo:
@dreispt Ok , we will be responsible for fix it.

Oh if we had optimistic merges... This feature was already available.

dreispt:
'll have to clarify that:
A green build is a hygienic requirement. Stable branches must be green at all time.
Optimistic merging is mostly about functional criteria: merge if a PR is correct but needing improvements.

PR #465, opened 2016-06-23
Improvements for the auth_brute_force module (ban user, custom ban message, automatically unban, de l10n)

christophlsa:
I added some improvements to the auth_brute_force module which we need:
- add option to use the user instead of the ip for checking of failed attempts and banning (use ir.config_parameter 'auth_brute_force.attempt_ban_type' and set to 'user', default is 'ip')
- add option to print custom error message (use ir.config_parameter 'auth_brute_force.attempt_ban_message', empty means no custom message)
- add option to automatically unban remotes after time (use ir.config_parameter 'auth_brute_force.attempt_ban_time' to set the minutes, default is 30 minutes, 0 means no automatically unbanning)
- add new attempt type 'unbanned' to log when the remote was unbanned to avoid automatically banning after the remote was unbanned and typed the wrong password again
- added german translations

### DIFF
--- a/auth_brute_force/README.rst
+++ b/auth_brute_force/README.rst
@@ -7,13 +7,13 @@ Tracks Authentication Attempts and Prevents Brute-force Attacks
 
 This module registers each request done by users trying to authenticate into
 Odoo. If the authentication fails, a counter is increased for the given remote
-IP. After a defined number of attempts, Odoo will ban the remote IP and
-ignore new requests.
+IP or user. After a defined number of attempts, Odoo will ban the remote IP or
+user and ignore new requests.
 This module applies security through obscurity
 (https://en.wikipedia.org/wiki/Security_through_obscurity),
 When a user is banned, the request is now considered as an attack. So, the UI
-will **not** indicate to the user that his IP is banned and the regular message
-'Wrong login/password' is displayed.
+will **not** (unless an error message is set) indicate to the user that his IP
+is banned and the regular message 'Wrong login/password' is displayed.
 
 This module realizes a call to a web API (http://ip-api.com) to try to have
 extra information about remote IP.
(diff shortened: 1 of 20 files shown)


dreispt:
The review by Mr.Travis found a few issues:

```
Start lint check just in modules changed
************* Module auth_brute_force
auth_brute_force/__init__.py:1: [E7901(rst-syntax-error), ] README.rst:90 Duplicate implicit target name: "usage".
************* Module auth_brute_force.__openerp__
auth_brute_force/__openerp__.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
************* Module auth_brute_force.controllers.controllers
auth_brute_force/controllers/controllers.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
************* Module auth_brute_force.models.res_authentication_attempt
auth_brute_force/models/res_authentication_attempt.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
************* Module auth_brute_force.models.res_banned_remote
auth_brute_force/models/res_banned_remote.py:1: [C8201(no-utf8-coding-comment), ] No UTF-8 coding comment found: Use `# coding: utf-8` or `# -*- coding: utf-8 -*-`
```

@moylop260 I wonder why the build is green?

moylop260:
We have those check as `beta` yet from a long time ago and just for modules changed.
Maybe is time of choose which ones are no-beta now and just for all modules.

dreispt:
@moylop260 I thought that `beta` means that only the changed files are checked, not the full repo.

christophlsa:
So should I change something here? I think I did not changed the parts Prof Travis mentioned. ^^

christophlsa:
One note: In our project we had a bug: If you type the wrong login then in the login template some parts were shown which should only appear for users with the right permissions. I found out that the call to `request.session.authenticate` set the `request.uid` to `False` if the login failed. `False`seems to mean: "use the administrator as user. So in the web module they save the `request.uid` before they do the call and reset this value if the login failes. I added this fix in my last commit.

gurneyalex:
looks nice. Could you please rebase?

christophlsa:
Rebased

PR #466, opened 2016-06-27
[ADD] field_char_transformed

hbrunn:
# 
# Transform data on character fields

This module was written to give programmers a convenient way to transform text input before it is written into the database or before it is read from it.

As a practical example, FieldCharStripped is provided, which strips whitespace from the value before it is committed.
# Configuration

This module is a dependency to be used by programmers, not end users. There's no configuration to be done via the UI.
# Usage

Depend on this module and import the field classes you need::

```
from openerp.addons.field_char_transformed import FieldCharTransformed, FieldCharStripped
```


moylop260:
LGTM

field_char_transformed/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+    from openerp.addons.field_char_transformed import FieldCharTransformed, FieldCharStripped
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/server-tools/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/server-tools/issues/new?body=module:%20field_char_transformed%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
elicoidal, line 31: Hi @hbrunn, thanks for the contribution!
Can you please use the latest README template and remove this last line?

field_char_transformed/README.rst
```diff
@@ -0,0 +1,61 @@
...
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Holger Brunn <hbrunn@therp.nl>
+
+Do not contact contributors directly about help with questions or problems concerning this addon, but use the `community mailing list <mailto:community@mail.odoo.com>`_ or the `appropriate specialized mailinglist <https://odoo-community.org/groups>`_ for help, and the bug tracker linked in `Bug Tracker`_ above for technical issues.
```
elicoidal, line 46: This is not part of the template. I would suggest that you make a PR against https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst if you wish to add a section/text.

field_char_transformed/field_char_transformed.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields
+from openerp.osv import fields as low_level_fields
```
dreispt, line 5: Why not `from openerp import fields as low_level_fields` ?

field_char_transformed/field_char_transformed.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.transform:
+            return self.transform(result)
+        return result
+
+    def convert_to_write(self, value, target=None, fnames=None):
+        result = super(FieldCharTransformed, self).convert_to_write(
+            value, target=target, fnames=fnames)
+        if self.transform:
+            return self.transform(result)
+        return result
```
dreispt, line 39: I'm wondering if we shouldn't allow to have different functions for each of these operations.
This would allow for some "encoding" on write, and then do the reverse "decoding" on read.
I can see some exciting possibilities for that feature.

field_char_transformed/field_char_transformed.py
hbrunn, line 5: this gets us v8 fields, and we import that already in the live above

field_char_transformed/field_char_transformed.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def to_column(self):
+        result = super(FieldCharTransformed, self).to_column()
+        if result and isinstance(result, low_level_fields.char):
+            result = ColumnCharTransformed(
+                transform=self.transform, **result._args)
+        return result
```
dreispt, line 46: Sorry for the ignorance: what is this method for? How is it different from the read operation? Does it have some overlapping with name_get?

field_char_transformed/field_char_transformed.py
hbrunn, line 39: the problem is that the names are a bit misleading. `convert_to_read` is also called in some cases when writing values and the other way around, so you can't really trust the names which 'direction' you're busy with currently. So in all generality, that's not really possible. For scalar fields, it would work the way you expect as long as no function fields are involved.

field_char_transformed/field_char_transformed.py
hbrunn, line 46: not at all. This function takes care of the magic to translate between old and new style fields: https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L709

lasley:
👍  functional and code

field_char_transformed/README.rst
hbrunn, line 46: done quite a while ago in https://github.com/OCA/maintainer-tools/pull/190

JordiBForgeFlow:
:+1: functional and code. If we applied this to the field 'name' in res.partner as 

from openerp.addons.field_char_transformed import FieldCharTransformed, FieldCharStripped
from openerp import api, models

    _inherit = 'res.partner'

```
name = FieldCharStripped()
```

Create partner:
image

Once it's saved:

image

moylop260:
👍

PR #468, opened 2016-06-29
[9.0][FIX] base_multi_image - New storage backend 'Filestore'

sebalix:
New storage backend _Filestore_ to link an existing attachment record + Fix the `pre_init_hook_for_submodules()` function to extract the images from the `ir_attachment` table for binary fields initialized with the `attachment=True` parameter.

Required to fix https://github.com/OCA/product-attribute/pull/161

I'm not quite satisfied with the pre_init_hook function, it works but it is not really elegant and error prone. Any advices?

cc @Yajo @atchuthan @pedrobaeza

### DIFF
--- a/base_multi_image/models/image.py
+++ b/base_multi_image/models/image.py
@@ -40,7 +40,8 @@ class Image(models.Model):
         readonly=True)
     attachment_id = fields.Many2one(
         'ir.attachment',
-        string='Attachment')
+        string='Attachment',
+        domain="[('index_content', '=', 'image')]")
     file_db_store = fields.Binary(
         'Image stored in database',
         filters='*.png,*.jpg,*.gif')
--- a/base_multi_image/models/image.py
+++ b/base_multi_image/models/image.py
@@ -190,7 +190,7 @@ class Image(models.Model):
                 _('You must provide an attached file for the image.'))
 
     @api.constrains('storage', 'attachment_id')
-    def _check_store(self):
+    def _check_attachment_id(self):
         if self.storage == 'filestore' and not self.attachment_id:
             raise exceptions.ValidationError(
                 _('You must provide an attachment for the image.'))
--- a/base_multi_image/models/image.py
+++ b/base_multi_image/models/image.py
@@ -28,7 +28,8 @@ class Image(models.Model):
     owner_model = fields.Char(
         required=True)
     storage = fields.Selection(
-        [('url', 'URL'), ('file', 'OS file'), ('db', 'Database')],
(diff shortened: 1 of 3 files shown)


sebalix:
Ready for reviews.

base_multi_image/models/image.py
```diff
@@ -174,3 +188,9 @@ def _check_store(self):
         if self.storage == 'db' and not self.file_db_store:
             raise exceptions.ValidationError(
                 'You must provide an attached file for the image.')
+
+    @api.constrains('storage', 'attachment_id')
+    def _check_store(self):
+        if self.storage == 'filestore' and not self.attachment_id:
+            raise exceptions.ValidationError(
```
moylop260, line 195: `base_multi_image/models/image.py:195: [C8107(translation-required), Image._check_store] String parameter of raise "ValidationError" requires translation. Use _('You must provide an attachment for the image.')`

moylop260:
👍

sebalix:
I would like to discuss about what we should do about these `Binary(attachment=True)` fields. There is several solutions.

The pros of the implementation proposed in this PR:
- allow to select the same image (attachment) several times for different galleries without any additional space
- small PostgreSQL dumps as the images are stored in the filestore
- no data loss on migrated fields when uninstalling the module

Cons:
- the UX with this `attachment_id` field isn't very convenient

Another method is to add a new binary field `file_db_store_attachment` with `attachment=True`, linked to a new backend `db_attachment`. This way we can manage these two kind of storage: binary in database (`db`) and binary as attachment (`db_attachment`). Then the pre_init_hook function could migrate `Binary(attachment=True)` fields (e.g. in `product.template` https://github.com/odoo/odoo/blob/9.0/addons/product/product.py#L538) by updating the relevant data in the `ir_attachment` table (`res_model`, `res_id` and `res_field`).

Pros:
- small PostgreSQL dumps as the images are stored in the filestore
- good UX, a normal binary field on the form as it is now

Cons:
- bad UX, choose between Database and Database/Attachment is unclear (maybe add the possibility to enable/disable the different storage backends for integrators, another problem for another PR)
- unable to select the same image several times, each use needs to be stored
- uninstalling the module will not recover the migrated `ir.attachment` records

lasley:
IMO the implementation in this PR is great. (code review 👍  btw)

I would also say that the migration of image records into an unrecoverable table could be a big red flag in terms of using this module. As a sys admin, I naturally shy away from something that migrates my data without a way to walk it back.

base_multi_image/models/image.py
sebalix, line 195: Fixed.

atchuthan:
code review :+1:

PR #469, opened 2016-06-29
[ADD] migrate database_cleanup

hbrunn:


database_cleanup/model/purge_wizard.py
```diff
@@ -1,78 +0,0 @@
-# -*- coding: utf-8 -*-
```
moylop260, line 1: In order to save the log history of commits by files... Could you use `git mv` instead of `git rm && git add`?

database_cleanup/models/purge_columns.py
```diff
@@ -0,0 +1,123 @@
...
+
+            self.logger.info(
+                'Dropping column %s from table %s',
+                line.name, model_pool._table)
+            self.env.cr.execute(
+                """
+                ALTER TABLE "%s" DROP COLUMN "%s"
+                """ % (model_pool._table, line.name))
+            line.write({'purged': True})
+            self.env.cr.commit()
```
moylop260: Is posible use other commit way based on [our guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction)?

database_cleanup/models/purge_tables.py
```diff
@@ -64,75 +44,62 @@ def purge(self, cr, uid, ids, context=None):
                 AND confrelid::regclass = '%s'::regclass;
                 """ % line.name)
```
moylop260: Is posible avoid sql-injection here?
`s/% line.name)/, (line.name,))/g`

database_cleanup/models/purge_tables.py
```diff
@@ -64,75 +44,62 @@ def purge(self, cr, uid, ids, context=None):
...
 
         # Cannot pass table names as a psycopg argument
         known_tables_repr = ",".join(
             [("'%s'" % table) for table in known_tables])
-        cr.execute(
+        self.env.cr.execute(
             """
             SELECT table_name FROM information_schema.tables
             WHERE table_schema = 'public' AND table_type = 'BASE TABLE'
             AND table_name NOT IN (%s)""" % known_tables_repr)
```
moylop260: Is posible avoid sql-injection here?
`s/% known_tables_repr/, (known_tables_repr,)/g`

database_cleanup/models/purge_columns.py
```diff
@@ -0,0 +1,123 @@
+# -*- coding: utf-8 -*-
+# © 2014-2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import _, api, fields, models
+from openerp.exceptions import UserError
+from openerp.osv import fields as legacy_fields
```
moylop260: What about use `import openerp.fields.fields as legacy_fields`?

database_cleanup/README.rst
```diff
@@ -1,15 +1,78 @@
...
+    :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* ...
```
moylop260: `Unexpected possible title overline or transition.|Treating it as ordinary text because it's so short.`
Could we delete this unused section?

database_cleanup/README.rst
```diff
@@ -1,15 +1,78 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+================
+Database cleanup
+================
+
 Clean your OpenERP database from remnants of modules, models, columns and
 tables left by uninstalled modules (prior to 7.0) or a homebrew database
```
moylop260, line 10: Is valid the `(prior to 7.0)` yet?

database_cleanup/models/purge_data.py
```diff
@@ -0,0 +1,71 @@
...
+    _name = 'cleanup.purge.line.data'
+
+    data_id = fields.Many2one('ir.model.data', 'Data entry')
+    wizard_id = fields.Many2one(
+        'cleanup.purge.wizard.data', 'Purge Wizard', readonly=True)
+
+    @api.multi
+    def purge(self):
+        """
+        Unlink data entries upon manual confirmation.
```
moylop260: Use docstring without empty-newline in the first line, reference: https://www.python.org/dev/peps/pep-0257/#specification

database_cleanup/README.rst
hbrunn, line 10: yes

database_cleanup/model/purge_wizard.py
hbrunn, line 1: that's what I did: https://github.com/OCA/server-tools/commits/d37bfe7be687d222a212d43f77a064d6ee8e6a42/database_cleanup/model/purge_wizard.py - github has trouble following file renames sometimes

database_cleanup/models/purge_columns.py
hbrunn: possible. What's wrong with this?

database_cleanup/models/purge_columns.py
hbrunn: no. It really needs to be the current transaction, otherwise stuff goes very wrong

database_cleanup/models/purge_tables.py
hbrunn: no, we need the cast

database_cleanup/models/purge_columns.py
moylop260: Use `from openerp.osv import fields` show the lint error:
`database_cleanup/models/purge_columns.py:6: [W0402(deprecated-module), ] Uses of a deprecated module 'openerp.osv'`

Other good proposal is use `# pylint: disable=deprecated-module` comment in the same line.

database_cleanup/models/purge_tables.py
moylop260: Ok I got it
Then we need add a custom method to raise a sql-injection in `line.name`
Because the field is filled from UI this have too much risk.

hbrunn:
@moylop260 thanks, points taken or answered

database_cleanup/model/purge_wizard.py
moylop260, line 1: Thank for clarify

database_cleanup/models/purge_columns.py
moylop260: Thanks for clarify!
following the rule:
`All cr.commit() calls outside of the server framework from now on must have an explicit comment explaining why they are absolutely necessary, why they are indeed correct, and why they do not break the transactions...`

You can add a comment to justify the commit and a pylint-disable-comment `# pylint: disable=invalid-commit` too to avoid the check:
`database_cleanup/models/purge_columns.py:47: [E8102(invalid-commit), CleanupPurgeLineColumn.purge] Use of cr.commit() directly - More info https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction`

hbrunn: turn out we only need two of the commits, thanks for nagging about that

database_cleanup/models/purge_tables.py
hbrunn: done in https://github.com/OCA/server-tools/pull/469/commits/50e09adb9313e04ae9e2f0a0c7a125f468a1e89c

database_cleanup/models/purge_columns.py
StefanRijnhart: Those commits were put in for usability. If you sweep a large number of tables, and some of them might not be removable for whatever reason, a transaction roll back is really unnecessary here and annoying.

database_cleanup/README.rst
```diff
@@ -1,15 +1,73 @@
...
-Caution! This module is potentially harmful and can *easily* destroy the
-integrity of your data. Do not use if you are not entirely comfortable
-with the technical details of the OpenERP data model of *all* the modules
-that have ever been installed on your database, and do not purge any module,
-model, column or table if you do not know exactly what you are doing.
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+    :alt: Try me on Runbot
+    :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+For further information, please visit:
```
elicoidal: Remove this  part

database_cleanup/README.rst
```diff
@@ -1,15 +1,73 @@
...
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/database_cleanup/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/database_cleanup/issues/new?body=module:%20database_cleanup%0Aversion:%209.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
elicoidal: update to latest README template (you can remove this part)

hbrunn: @elicoidal done

moylop260:
👍 
Thanks for contribution
Excellent work in this refactory!

luistriana28:
👍 
Nice work!

StefanRijnhart:
:+1:, thanks!

PR #470, opened 2016-06-29
8.0 database cleanup backport 9 improvements

hbrunn:


database_cleanup/identifier_adapter.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from psycopg2.extensions import ISQLQuote
+
+
+class IdentifierAdapter(ISQLQuote):
```
StefanRijnhart, line 7: How about adding this to openupgradelib and add the dependency in this module?

StefanRijnhart, line 7: BTW. just read about AsIs. How does this compare to

```
from psycopg2.extensions import AsIs
self.env.cr.execute('select * from %s', (AsIs('res_partner'),))
```

hbrunn, line 7: that's the same as string formatting for the purposes of this module

StefanRijnhart, line 7: OK. You gotta explain that to me over a beer next time. For now :+1: ~~when Travis becomes green~~

hbrunn:
travis error unrelated

sebalix:
:+1:

NL66278:
:+1:

PR #472, opened 2016-07-03
[ADD] PoC for base_encrypted_field

hbrunn:
This is a quite early proof of concept for client side encryption for Odoo fields.

Things to do before we can even think of merging:
- [ ] implement reusing groups and adding users to new groups
- [ ] add a UI for administering encryption groups
- [ ] better UI for encrypting/decrypting fields (for example, this relies on labels being available)
- [ ] make it clean when a field is encrypted/decrypted/not available for a specific user
- [ ] refactor the cascading deferred calls away for better modularity
- safeguard against data loss, for example by
  - [ ] removing all users/keys from an encryption group
  - [ ] resetting PGP keys (we need a re-encrypt existing keys)


hbrunn:
So I created a key pair for admin, you'll find this in the preferences. The only encrypted field at the moment is demo's signature. To decrypt this, click the little lock next to the field label, fill in the passphrase `hello world`, and you should see the signature. On some browsers, you'll need to force SSL to use crypto functions.
If you want to encrypt some other field, click the key icon next to its label. It will ask to reuse or to create an encryption group. Only the latter is implemented, so fill in some name for a new group and click create. Don't forget to save, now an encrypted version of your text is stored in the database.

Whoever is funny enough to generate a new key pair has destroyed all data encrypted with the one before.

lasley:
Playing with this on my local right now and am running into an error, so I replicated in Runbot both in a new user and in a sys params. The sys param one is throwing an error, which is also below. 

Process I am following, just to make sure I'm not stupid:
- Edit
- Input Text
- Lock
- Pass & Group wizards
- Save (text is now ciphered)
- Click Unlock

In the users no error is displayed, but the signature text isn't decrypted. Users in RunBot are:
- New User
- Second

In sys params though, we may have a clue:

Settings => System Parameters => [Key] Test Encrypt => Unlock:

```
Uncaught Error: Syntax error, unrecognized expression: HJTX3gAAAAOZHEh4q+vrkWoLqkiI7utsnDrvxsiyyRPrvxGnew6uV0hgnupmbNDCGUOzKXDRC1f7fcR0WaZldBpDaWo7OTuLZrYT0xDkk8e904hsUoJj5KnUjWeG5aTCdzW6a4ylPZe0yo+4UX3JhI9q1Bo83Y8Z7pb1Sp+9iGn2J+wh0hj5bUMh/lzjWT7W/+cB3gIZSM75NztnZIgevQamOqJwGhbxaGa/5bwrAUS5QvTuc0KDYQFh5duBoDz8vE6puGVP2xG7xMtx6BP171Pk+g==
```

hbrunn:
this looks like two different problems. Will have a look, but not right now.

lasley:
Cool, I'm diving into the code as well. Spent yesterday learning and playing with OpenPGP so I can actually be useful here. Once I have my bearings, I can take some of the tasks from you if you would like.

base_encrypted_field/models/res_users.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    pgp_private_key = fields.Text('Private pgp key', readonly=True)
```
elicoidal, line 10: @hbrunn thanks for the PoC! looks promising
I was under the impression that the private key was not leaving the client. Did I misunderstand the concept?

lasley, line 10: Both keys are stored on the server, but there should be a password on the private key which only the client knows. All of the crypto methods are taking place client side, so the password never touches the server.

elicoidal, line 10: Clear to me thanks! @lasley

lasley:
@hbrunn - When you get some time take a look at #673. I think I may have gotten us on the right track maybe.

hbrunn:
this sounds very familiar! I think as soon as we have something like the Odoo crypto API, we can abstract from the actual implementation, as long as we assume some symmetric and some asymmetric "encryption" (quotes because the symmetric part doesn't actually need to be encryption). From the Odoo point of view, the hard part is key management and widgets that integrate with it. As soon as we have that, it shouldn't matter if some keys come from a smart card, a brain or where ever else

lasley:
@hbrunn Yeah that's kind of my thoughts on this - I figured that a working PoC with one technology could then just be stripped down into interface and adapter modules once we know the idea is solid. Given the module design, the abstract could simply just be external system file management - allowing for things like S3 storage and all that too.

Let me know if you come up with any cool ideas, or thoughts on the PoC implementation. This will be my side project for the next month or so I'm thinking.

hbrunn:
I will, good to know we're on the same page here. Can't say if/when I'll work on this in the coming months, but surely will share whatever it is

PR #473, opened 2016-07-04
[FIX] singleton error if we saved the current session two times

hbrunn:
For multi process servers, we can create the auditlog session multiple times, which results in a singleton error when returning the id. The proper solution to this would be a lock on the table spanning from the search operation to the create, but this would impact speed quite negatively, so I'd rather opt for the duplicate records.

The indexes have nothing to do with the problem, but seem to be a good idea to me.


dreispt:
:+1:

pedrobaeza:
:+1:

PR #480, opened 2016-07-12
[ADD] field_rrule

hbrunn:
# 
# Repetition rules

This module was written to offer a field type that holds rrules according to RFC 2445.
# Usage

- depend on it
- say `from openerp.addons.field_rrule.field_rrule import FieldRRule`
- use `FieldRRule` like any other field
- on forms, use `widget="rrule"`
- have a look at `demo/res_partner.*`

Technically, this is a wrapper around serialized fields. The value always will be a subclass of dateutil's `rruleset`. For technical reasons, this class overrides `__iter__`, so if you need a proper `rruleset`, call the value: `my_browse_record.my_field_of_type_rrule()` - this gives you a vanilla `rruleset`.
# Known issues / Roadmap
- support the unimplemented features of rrules
- support rdates, exdates, exrules
- consider multiple widgets with different feature sets


hbrunn:
@StefanRijnhart this is the widget to be used in https://github.com/OCA/bank-payment/pull/281

field_rrule/README.rst
```diff
@@ -0,0 +1,71 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+================
+Repetition rules
+================
+
+This module was written to offer a field type that holds rrules according to RFC 2445.
```
dreispt, line 9: Hmm, I read that [RFC2445](https://tools.ietf.org/html/rfc2445) is obsoleted by [RFC5545](https://tools.ietf.org/html/rfc5545)...

field_rrule/README.rst
```diff
@@ -0,0 +1,71 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+* depend on it
+* say ``from openerp.addons.field_rrule import FieldRRule``
+* use ``FieldRRule`` like any other field
+* on forms, use ``widget="rrule"``
+* have a look at `demo/res_partner.*`
```
dreispt: use double quotes

field_rrule/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Technically, this is a wrapper around serialized fields. The value always will be a subclass of dateutil's ``rruleset``. For technical reasons, this class overrides ``__iter__``, so if you need a proper ``rruleset``, call the value: ``my_browse_record.my_field_of_type_rrule()`` - this gives you a vanilla ``rruleset``.
+
+If you want to pass a default, use the internal representation you'll find in the database - a list of dictionaries with the keyword arguments to be passed to rrule's constructor and a `type` field that for now can only be `rrule`: A context of ``{"default_rrule": [{"count": 1, "freq": 1, "type": "rrule", "interval": 1, "bymonthday": [1]}]}`` would give you a default for the field ``rrule`` which occurs one time at the first of the month.
```
dreispt, line 24: It would be nice if `repr(<rrule object>)` would do this for us (or alternatively some helper function). But aboslutely not an indispensable feature.

field_rrule/demo/res_partner.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
dreispt, line 1: Since there isn't a cleaner way to do it, I suggest having the demo as a separate module `field_rrule_demo`.
This would allow us to see the module in action on Runbot.

dreispt:
I wonder what should be the naming convention for new field types: `field_xxx` or `base_field_xxx`?

field_rrule/tests/test_field_rrule.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestFieldRrule(TransactionCase):
+    def test_field_rrule(self):
+        pass
```
dreispt: I bet the code coverage is not that great 😉

field_rrule/README.rst
lasley, line 9: Quick review looks like this complies with 5545, just need to update the ReadMe.

More info https://tools.ietf.org/html/rfc5545#appendix-A

hbrunn, line 9: I linked what the author of dateutil claims to have implemented: https://labix.org/python-dateutil#head-470fa22b2db72000d7abe698a5783a46b0731b57

field_rrule/README.rst
hbrunn, line 24: it does: https://github.com/OCA/server-tools/pull/480/files#diff-04e747b1caf1bcf4b384e6a3c4b8258aR64, but that's not going to help in a form definition or the xml describing an action

hbrunn:
about the name: [precedent](https://github.com/OCA/server-tools/tree/8.0/field_char_transformed) says no `base_`, and this would also conflict with point two from the [conventions](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules)

StefanRijnhart:
:+1: tested.

field_rrule/demo/res_partner.py
hbrunn, line 1: I'm fervently against that, but maybe I've a better offer: https://github.com/OCA/server-tools/pull/480/files#diff-9cc3eb31c10fb83195573350fb4de22bR8

lasley:
Code + functional, including the fanciness in the post load hook for Runbot 👍

field_rrule/README.rst
dreispt, line 9: OK

field_rrule/demo/res_partner.py
dreispt, line 1: Wow, you're breaking new ground there!
If I read it correctly, you are capable of conditionally import the code in `demo/`, right?
I think we should improve the conditions: the demo code should be loaded as long as the `demo` server option is enabled.

field_rrule/hooks.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import socket
+import inspect
+
+
+def post_load_hook():
+    """do some trickery to have demo data/model on runbot, but nowhere else"""
```
StefanRijnhart, line 9: Why only on odoo-community.org?

field_rrule/demo/res_partner.py
hbrunn, line 1: this unfortunately doesn't work, because modules are loaded server wide, but demo or not is a property of a database. no chance to determine which database is going to be used during server startup

field_rrule/hooks.py
hbrunn, line 9: what would be the alternative?

field_rrule/demo/res_partner.py
dreispt: Could we examine the `config` command line options to check if `--demo` was used?

field_rrule/hooks.py
dreispt, line 9: I say we leave it like that for now, but a small voice in my head says that there is a better way 👼

dreispt:
:+1:

field_rrule/hooks.py
StefanRijnhart, line 9: Sorry, I misread. I thought you could check if demo mode was activated. This is both very clever and a horrible hack. What was wrong with the inclusion of a dedicated demo module again?

hbrunn, line 9: I think it would clutter our repos a lot if modules that don't demo themselves come with demo modules, especially in server-tools where we have many foundational modules that don't do anything by themselves. This is only meant as convenience for the reviewers to be able to test the thing on runbot without having to clone and install the branch locally.

field_rrule/README.rst
lasley, line 9: Just to note - 5545 differences were mainly in implementation & error validation. While this could be something the library would handle, it's all easily avoided module-side too. IMO it would be advantageous to claim 5545, or at least note that this module attempts to conform - because it does currently.

gurneyalex:
:+1:

PR #478, opened 2016-07-06
Fix name search to honor additional domain if provided

dreispt:
Fixes #476

PR #481, opened 2016-07-13
[FIX] Respect domain set on x2many relational fields

codingforfun:
The improved base name search in `base_name_search_improved` module doesn't respect the domain set on relational field. Without this fix one gets all matching results from the target model of the relation not only the ones that match the search string **AND** the domain set on the field.

### DIFF
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -49,20 +49,25 @@ class ModelExtended(models.Model):
                 enabled = self.env.context.get('name_search_extended', True)
                 # Perform extended name search
                 if enabled and operator in ALLOWED_OPS:
+                    field_domain = args or []
                     # Support a list of fields to search on
                     all_names = _get_rec_names(self)
                     # Try regular search on each additional search field
                     for rec_name in all_names[1:]:
-                        domain = [(rec_name, operator, name)]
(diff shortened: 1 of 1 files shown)


codingforfun:
@hbrunn @dreispt 
I think you maybe like to review.

base_name_search_improved/models/ir_model.py
```diff
@@ -49,19 +49,20 @@ def name_search(self, name='', args=None,
...
                     all_names = _get_rec_names(self)
                     # Try regular search on each additional search field
                     for rec_name in all_names[1:]:
-                        domain = [(rec_name, operator, name)]
+                        domain = field_domain + [(rec_name, operator, name)]
                         res = _extend_name_results(self, domain, res, limit)
                     # Try ordered word search on each of the search fields
                     for rec_name in all_names:
-                        domain = [(rec_name, operator, name.replace(' ', '%'))]
+                        domain = field_domain + [(rec_name, operator, name.replace(' ', '%'))]
```
OSguard: flake8 breakse CI here

OSguard:
:+1:

OSguard:
Das ist leider ein duplicate zu #478

codingforfun:
Closed because this was a duplicate to #478

PR #489, opened 2016-08-02
[FIX] purge uninstalled uninstallable modules instead of deleting

hbrunn:
This is necessary for module which were installed some time, then uncleanly uninstalled and only afterwards became uninstallable.

### DIFF
--- a/database_cleanup/model/purge_modules.py
+++ b/database_cleanup/model/purge_modules.py
@@ -106,13 +106,18 @@ class CleanupPurgeWizardModule(orm.TransientModel):
 
     def find(self, cr, uid, context=None):
         module_pool = self.pool['ir.module.module']
+        purge_line_pool = self.pool['cleanup.purge.line.module']
         module_ids = module_pool.search(cr, uid, [], context=context)
         res = []
         for module in module_pool.browse(cr, uid, module_ids, context=context):
             if get_module_path(module.name):
                 continue
             if module.state == 'uninstalled':
-                module_pool.unlink(cr, uid, module.id, context=context)
+                purge_line_pool.purge(
+                    cr, uid, [
+                        purge_line_pool.create(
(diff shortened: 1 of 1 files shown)


database_cleanup/model/purge_modules.py
```diff
@@ -106,13 +106,18 @@ def default_get(self, cr, uid, fields, context=None):
...
         res = []
         for module in module_pool.browse(cr, uid, module_ids, context=context):
             if get_module_path(module.name):
                 continue
             if module.state == 'uninstalled':
-                module_pool.unlink(cr, uid, module.id, context=context)
+                purge_line_pool.purge(
+                    cr, uid, [
+                        purge_line_pool.create(
+                            cr, uid, {'name': module.name}, context=context),
```
dreispt, line 119: It's a style nitpick, but IMO the could would be more readable if the `create` was done in it's own instruction.

dreispt:
:+1:

moylop260:
👍 
Could you check the build of oca/ocb?

gurneyalex:
maybe #507 will help in getting a green travis.

dreispt:
#507 is merged. I guess a rebase is needed.

hbrunn:
@dreispt how so?

dreispt:
@hbrunn AFAICS you need the changes #507 changes here to get a green build...

hbrunn:
@dreispt that's not how git works. Merging this will just add the extra commit on the python file, the travis definition will be quite unaffected

dreispt:
@hbrunn True for the target branch, but not for the PR: to get a green build on the PR checks the rebase is needed.
Anyway, it seems that #507 was not the issue, since I restarted the build and it is green now...

StefanRijnhart:
:+1:

PR #494, opened 2016-08-05
[ADD] stock_enterprise_remove

woodbrettm:
This module removes any settings marked as enterprise from the stock res_config view.

Cheers,
Brett

@lasley


lasley:
LGTM 👍

elicoidal:
:+1:

rvalyi:
:+1:

pedrobaeza:
You have merged all of this stuff including the SVG icon that it's not needed. Now, the repository is going to contain a diff history adding them and lately removing it, making very big (well, bigger than before). The only solution is to amend the commits, but this means that some people making `git pull` will receive an error. What do you think?

PR #505, opened 2016-08-05
Fix travis build on 9.0 branch

gurneyalex:
the 9.0 branch of the project is red, with really strange errors about computed fields. There have been such errors since the merge of #469 but for some reason they were neither fatal nor breaking the build. 

Splitting `database_cleanup` tests in .travis.yml seems to let the other modules go green, but the tests for database_cleanup are red.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
-  - UNIT_TEST="1"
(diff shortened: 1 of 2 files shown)


pedrobaeza:
It's still red...

woodbrettm:
The exact same error is showing up in my builds as well.

gurneyalex:
@pedrobaeza @bmw95 I updated the description of the PR. 

@hbrunn @StefanRijnhart your help on this would be very welcome.

StefanRijnhart:
There is a security check to see if the user actually has access to the purge wizard menus to prevent unauthorized access (because no access rules apply to transient models). Because the server is not in debug mode, access to technical menus is not available. This raises the security error.

gurneyalex:
@StefanRijnhart I'm still a bit worried about the failures in other module's tests when database_cleanup is present.

PR #507, opened 2016-08-08
split database_cleanup tests in travis config

gurneyalex:
try to find out why travis is red on 8.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
:+1: I had to include it in #508 as well.

gurneyalex:
@pedrobaeza if you could spare a few minutes, I'd welcome a fast track :smile_cat:

pedrobaeza:
Hehe, OK. Although I prefer to use another technique instead of splitting builds, but I understand that this is enough consumed time to check why the errors or inhibit in some tests the database_cleanup behavior.

:+1:

PR #508, opened 2016-08-09
[RFR] Explicit access rights so that tests can run

StefanRijnhart:
Fixes #505

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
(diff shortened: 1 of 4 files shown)


gurneyalex:
thanks for looking into this.

StefanRijnhart:
Travis is green!

gurneyalex:
:+1: 

@StefanRijnhart any idea why splitting the tests of dabase_cleanup is required?

StefanRijnhart:
@gurneyalex I was on holiday but it would be good to find out what goes wrong indeed. Need to coordinate with @hbrunn

PR #515, opened 2016-08-10
9 attachment base synchronize

florian-dacosta:
Migrate and rename attachment_metadata
This module is a base module which is meant to be used with module like external_file_location or file_email, which I will put in others PR

The goal is to extend ir.attachment to add some feature
- a hash system to control file integrity with the imported file.
- A run method, used in other modules to transform the file data (for instance we can have a method to import bank statement, a method to import account move (with account_move_base_import) or even a method used to do a classic import)

For instance, we have a module external_file_location which is able to get files from ftp or sftp, then others complementary module will implement the action to run on the imported file.


attachment_base_synchronize/models/attachment.py
```diff
@@ -0,0 +1,104 @@
...
+    external_hash = fields.Char(
+        help="File hash comes from the external owner of the file.\n"
+             "If provided allow to check than downloaded file "
+             "is the exact copy of the original file.")
+    attachment_id = fields.Many2one(
+        'ir.attachment', required=True, ondelete='cascade',
+        help="Link to ir.attachment model ")
+    file_type = fields.Selection(
+        selection=[],
+        string="File type",
```
angelmoya, line 33: better "Field Type" ?

angelmoya (commented):
I create a PR to add tests, then will be OK

only a question, Why is it necesary to create new enviroment to set done?

angelmoya (approved):
it's OK

PR #517, opened 2016-08-12
[Add] New module: Disable field translation globally per model

codingforfun:
For _data only_ models like products it often doesn't makes much sense to translate stuff. Translating field contents of these models (e.g. product description) often causes unnecessary trouble and isn't  necessary. 
This module provides means to disable translation of field contents for all fields on per model base. This includes fields of model customizations inherited from other modules. 

The case I made it for is that some of our customers with small to medium businesses do not want to translate content models like products and stuff because it's more work to maintain for them. But still they want system models like menus and stuff to get translated.

Of course we could override fields with `translate=False` options, but this is cumbersome if we also want to use other customizations that e.g. introduce new fields. 

The problem with translatable fields and your language is not the default language and you edit data, you change only your translation and not the original data. So removing the translation makes it for the user easy that you have this data only in one language.

Also some customers have multi-cultural employees but only sell on the german market, so their employees want to use the user interface in their language, but the product data for example should be consistent independent of the interface language the employee who edited last has set.

### DIFF
--- a/base_no_translation/README.rst
+++ b/base_no_translation/README.rst
@@ -1,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Disable model translation
+=========================
+
+Provides means to globally disable translation of field contents for
+all fields on a per Model base.
+Overrides `translate=True` of all fields of a model (including inherited
+custom fields of other models).
+
+For example to disable translation of all product data (description etc.).
+Inherit from `product.product` (and `product.template`) and set class
+attribute `self._disable_field_translations` of the model to False.
+
+
+Installation
+============
+
+No specific requirements.
+
+
+Configuration
+=============
+
+To disable translation of all fields for a given model you need to inherit
+that model and set class attribute `self._disable_field_translations`
+of the model to True.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/serevr-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Peter Hahn <peter.hahn@initos.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Take into account that in a multi-language company or for example in an multi-language e-commerce site, you need the translations of these fields.

codingforfun:
@pedrobaeza 
I'm not sure if I understand what you are trying to point out. 

Of course this module is optional as always and should be used only if it applies to your use case. 

The case I made it for is that some of our customers with small to medium businesses do not want to translate content models like products and stuff because it's more work to maintain for them. But still they want system models like menus and stuff to get translated. 

Of course we could override fields with `translate=False` options, but this is cumbersome if we also want to use other customizations that e.g. introduce new fields.

OSguard:
build fail unrelated as v8 is broken

The problem with translatable fields and your language is not the default language and you edit data, you change only your translation and not the original data. So removing the translation makes it for the user easy that you have this data only in one language.

pedrobaeza:
OK, I get the use case. I have also feel sometimes that the translation system is a bit problematic when using a non-English language. For example, you duplicate a product, and in some moment, you find a "Product name (copy)" in a report, view or field, and it's due to the English translation remains with the provisional name given to the copied record. I have thought about having some kind of mechanism to copy back to English the new "translation" in these cases, but I haven't seen possible side effects.

Going back to topic, please add a little test to assure that this will work from now on: you can set any of the base models that have a translatable field with this possibility, and check passing in context different `lang` variables to check if the value is the same.

codingforfun:
Also some customers have multi-cultural employees but only sell on the german market, so their employees want to use the user interface in their language, but the product data for example should be consistent independent of the interface language the employee who edited last has set.

pedrobaeza:
You can add all of this feedback to the description of the module to let clear the intention of it.

base_no_translation/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+No specific requirements.
+
+
+Configuration
+=============
+
+To disable translation of all fields for a given model you need to inherit
+that model and set class attribute 'self._disable_field_translations'
+of the model to False.
```
gurneyalex: This sounds very counter intuitive: there is a double negative here, I'd expect having to set `disable_field_translations` to `True` in order to disable the translations.

base_no_translation/model.py
```diff
@@ -0,0 +1,20 @@
...
+
+from openerp import models
+
+
+class ModelExtended(models.Model):
+    _inherit = 'ir.model'
+
+    def _register_hook(self, cr):
+        for m in self.pool.models.itervalues():
+            if getattr(m, '_disable_field_translations', False):
```
gurneyalex, line 13: This is different from what the README says (and actually what I expect... So the README needs fixing :smile_cat:)

codingforfun, line 13: You'r right. The README needs fixing.

base_no_translation/README.rst
```diff
@@ -0,0 +1,67 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=========================
+Disable model translation
+=========================
+
+Provides means to globally disable translation of field contents for
+all fields on a per Model base.
+Overrides 'translate=True' of all fields of a model (including inherited
```
gurneyalex: I'm nitpicking here: could you please change these quotes ('translate=True') to backquotes (`translate=True`). These are markers in ReST for `interpreted text` which is generally used for code excerpts. 

Same would need to be done in the rest of the file.

gurneyalex:
small change suggested. 

Globally :+1: on this.

bealdav:
:+1: thanks

codingforfun:
The test @pedrobaeza  asked for is still missing. Sorry, no time at the moment.

PR #518, opened 2016-08-15
auth_admin_passkey to new API

rvalyi:
in order to minimize the diff with v9 (another migration PR to v9 is also coming)

note: I don't think it's possible to convert the file res_users in 8.0 to the new API however because all methods are called from an overload of the authenticate method written in the old style API.

cc @legalsylvain

### DIFF
--- a/auth_admin_passkey/model/res_config.py
+++ b/auth_admin_passkey/model/res_config.py
@@ -20,57 +20,51 @@
 #
 ##############################################################################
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
res_user.py is not migrated to new API.

rvalyi:
Travis was stuck, so superseded by https://github.com/OCA/server-tools/pull/519

PR #519, opened 2016-08-15
auth_admin_passkey to new API

rvalyi:
cc @legalsylvain 
@pedrobaeza sorry I didn't master the API wizardry enough, but I finally got it an migrated res_users.py too. I amended the push and pushed again.
Travis didn't start in the old PR so I closed it an created a new one.


pedrobaeza:
Thanks for the changes. Please check Travis errors (some are related, although I think the main build isn't).

rvalyi:
@pedrobaeza I amended the commit and fixed the style. As you can see the Travis error on fetchmail is unrelated, so tests are passing.

rvalyi:
@pedrobaeza hey I won the heisenbug Travis lottery after a few attempts: all green. So if you could review again... The idea is I migrated the API just before an other v9 PR port and also the feature extension we discussed a year ago...

auth_admin_passkey/model/res_users.py
```diff
@@ -1,55 +1,40 @@
...
 
 import datetime
 
 from openerp import SUPERUSER_ID
-from openerp import pooler
+from openerp import registry
 from openerp import exceptions
-from openerp.osv.orm import Model
+from openerp import models, api
 from openerp.tools.translate import _
```
pedrobaeza: This can be also imported in `openerp` namespace directly

auth_admin_passkey/model/res_users.py
```diff
@@ -1,55 +1,40 @@
...
+# Copyright (C) 2013-2014 GRAP (http://www.grap.coop)
+# @author Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 import datetime
 
 from openerp import SUPERUSER_ID
-from openerp import pooler
+from openerp import registry
 from openerp import exceptions
```
pedrobaeza: You can join all of these imports (and order them in alphabetic)

auth_admin_passkey/model/res_users.py
```diff
@@ -116,22 +101,24 @@ def authenticate(self, db, login, password, user_agent_env):
                     self._send_email_passkey(cr, user_id, user_agent_env)
                 else:
                     self._send_email_same_password(cr, login)
-                cr.commit()
+                # we need to commit as we need to create the mails
+                # and we are not in a method with standard cursor management
+                cr.commit()  # pylint: disable=invalid-commit
```
pedrobaeza: I think there is an argument in send_mail called `autocommit` or similar that makes the same.

rvalyi:
@pedrobaeza done. How does it look now?

pedrobaeza:
Thanks for the changes.

:+1:

lasley:
LGTM as well 👍

auth_admin_passkey/model/res_config.py
```diff
@@ -1,76 +1,53 @@
...
-    def get_default_auth_admin_passkey_send_to_user(
-            self, cr, uid, ids, context=None):
-        icp = self.pool['ir.config_parameter']
+    @api.model
+    def get_default_auth_admin_passkey_send_to_user(self, fields):
         return {
-            'auth_admin_passkey_send_to_user': safe_eval(icp.get_param(
-                cr, uid, 'auth_admin_passkey.send_to_user', 'True')),
+            'auth_admin_passkey_send_to_user':
+            self.env["ir.config_parameter"].get_param(
```
legalsylvain: Hi,
Why did you removed safe_eval here, and not here : 
https://github.com/OCA/server-tools/pull/519/commits/8025a5db364bb551160a32e5f763081e70111b68#diff-70199c6feb20b249a2794f26b51b5105R35

legalsylvain:
Hi raphael,
Thanks a lot for porting this module. Just one question in the first commit about safe_eval.
Otherwise LGTM. Thanks !

sbidoul:
@rvalyi since you are busy refactoring this module, do you have an opinion on #529  ?

dvdhinesh:
I am facing an Internal server error when testing this PR on runbot.

Issue:
When a user login with the same password as admin, it tries to send an email for intimation.

Traceback:

```
result = dispatch(method, params)
return fn(*params)
return res_users.authenticate(db, login, password, user_agent_env)
self._send_email_same_password(cr, login)
TypeError: _send_email_same_password() takes exactly 2 arguments (3 given)
```

PR available @: https://github.com/akretion/server-tools/pull/6

PR #520, opened 2016-08-15
Fixes server error in initial user add.

tstonis:
When logging in for a new user, I would get a "Record does not exist or has been deleted." error and internal server error even though the user is created. I couldn't find the exact reason, but has something to do with setting values['email'] when the new user is created/copied (setting values['name'] is okay...). I changed the code so that after the new user is copied/created, it is then modified with the other LDAP attributes. I tried for a long time to find out the root cause (something related to object caching it looks like), but I had to cut my losses.

(Sorry I'm a relative GitHub and OCA newbie, so I apologize in advance for any poor decorum for contributing to this repository).

### DIFF
--- a/users_ldap_mail/users_ldap_model.py
+++ b/users_ldap_mail/users_ldap_model.py
@@ -19,19 +19,22 @@
 #
 ##############################################################################
 
-#from openerp.osv import fields, orm
-from openerp import models, fields, api, _
+from openerp import models, fields
 from openerp import SUPERUSER_ID
 
 import logging
 _logger = logging.getLogger(__name__)
-import pdb
+
 
 class CompanyLDAP(models.Model):
     _inherit = 'res.company.ldap'
 
-    name_attribute = fields.Char('Name Attribute', help="By default 'cn' is used. or ActiveDirectory you might use 'displayName' instead.", default='cn')
-    mail_attribute = fields.Char('Email Attribute', help="LDAP attribute to use to retrieve email address.",default='mail')
+    name_attribute = fields.Char('Name Attribute',
+        help=("By default 'cn' is used. or ActiveDirectory you might use "
+              "'displayName' instead."), default='cn')
+    mail_attribute = fields.Char('Email Attribute', 
+        help="LDAP attribute to use to retrieve email address.",
+        default='mail')
 
     def get_ldap_dicts(self, cr, ids=None):
         """
@@ -51,10 +54,10 @@ class CompanyLDAP(models.Model):
         """, args)
         return cr.dictfetchall()
 
-
     def get_or_create_user(self, cr, uid, conf, login, ldap_entry,
                            context=None):
-        user_id = super(CompanyLDAP, self).get_or_create_user(cr, uid, conf, login, ldap_entry, context)
+        user_id = super(CompanyLDAP, self).get_or_create_user(cr, uid, conf,
+                        login, ldap_entry, context)
 
         mapping = [
             ('name', 'name_attribute'),
@@ -66,10 +69,10 @@ class CompanyLDAP(models.Model):
                 if conf[conf_name]:
                     values[value_key] = ldap_entry[1][conf[conf_name]][0]
             except KeyError:
-                _logger.warning('No LDAP attribute "%s" found for login  "%s"' % (
-                    conf.get(conf_name), values.get('login')))
+                _logger.warning('No LDAP attribute "%s" found for login  "%s"'
+                    % (conf.get(conf_name), values.get('login')))
 
         new_user = self.pool['res.users'].browse(cr, SUPERUSER_ID, user_id)
-        new_user.write( values ) 
+        new_user.write(values) 
 
         return user_id
--- a/users_ldap_mail/users_ldap_model.py
+++ b/users_ldap_mail/users_ldap_model.py
@@ -29,10 +29,13 @@ _logger = logging.getLogger(__name__)
 class CompanyLDAP(models.Model):
     _inherit = 'res.company.ldap'
 
-    name_attribute = fields.Char('Name Attribute',
+    name_attribute = fields.Char(
+        'Name Attribute',
         help=("By default 'cn' is used. or ActiveDirectory you might use "
-              "'displayName' instead."), default='cn')
-    mail_attribute = fields.Char('Email Attribute', 
+              "'displayName' instead."),
+        default='cn')
+    mail_attribute = fields.Char(
+        'Email Attribute',
         help="LDAP attribute to use to retrieve email address.",
         default='mail')
 
@@ -56,8 +59,8 @@ class CompanyLDAP(models.Model):
 
     def get_or_create_user(self, cr, uid, conf, login, ldap_entry,
                            context=None):
-        user_id = super(CompanyLDAP, self).get_or_create_user(cr, uid, conf,
-                        login, ldap_entry, context)
+        user_id = super(CompanyLDAP, self).get_or_create_user(
+            cr, uid, conf, login, ldap_entry, context)
 
         mapping = [
             ('name', 'name_attribute'),
@@ -69,10 +72,11 @@ class CompanyLDAP(models.Model):
                 if conf[conf_name]:
                     values[value_key] = ldap_entry[1][conf[conf_name]][0]
             except KeyError:
-                _logger.warning('No LDAP attribute "%s" found for login  "%s"'
+                _logger.warning(
+                    'No LDAP attribute "%s" found for login  "%s"'
                     % (conf.get(conf_name), values.get('login')))
 
         new_user = self.pool['res.users'].browse(cr, SUPERUSER_ID, user_id)
-        new_user.write(values) 
+        new_user.write(values)
 
         return user_id
--- a/users_ldap_mail/users_ldap_model.py
+++ b/users_ldap_mail/users_ldap_model.py
@@ -19,32 +19,23 @@
 #
 ##############################################################################
 
-from openerp.osv import fields, orm
+#from openerp.osv import fields, orm
+from openerp import models, fields, api, _
+from openerp import SUPERUSER_ID
 
 import logging
-_log = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
+import pdb
 
-
-class CompanyLDAP(orm.Model):
+class CompanyLDAP(models.Model):
     _inherit = 'res.company.ldap'
-    _columns = {
-        'name_attribute': fields.char(
-            'Name Attribute', size=64,
-            help="By default 'cn' is used. "
-                 "For ActiveDirectory you might use 'displayName' instead."),
-        'mail_attribute': fields.char(
-            'E-mail attribute', size=64,
-            help="LDAP attribute to use to retrieve em-mail address."),
-    }
 
-    _defaults = {
-        'name_attribute': 'cn',
-        'mail_attribute': 'mail',
-    }
+    name_attribute = fields.Char('Name Attribute', help="By default 'cn' is used. or ActiveDirectory you might use 'displayName' instead.", default='cn')
+    mail_attribute = fields.Char('Email Attribute', help="LDAP attribute to use to retrieve email address.",default='mail')
 
     def get_ldap_dicts(self, cr, ids=None):
         """
-        Copy of auth_ldap's funtion, changing only the SQL, so that it returns
+        Copy of auth_ldap's function, changing only the SQL, so that it returns
         all fields in the table.
         """
         if ids:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You can take the occasion to migrate to new API. Check Pylint errors and please sign CLA: https://odoo-community.org/page/cla

tstonis:
I think I have done what is required to migrate to the new API. Please correct me if there is a way to decorate or declare the methods that are overridden so that they bridge properly.

users_ldap_mail/users_ldap_model.py
```diff
@@ -19,32 +19,29 @@
...
+        'Name Attribute',
+        help=("By default 'cn' is used. or ActiveDirectory you might use "
+              "'displayName' instead."),
+        default='cn')
+    mail_attribute = fields.Char(
+        'Email Attribute',
+        help="LDAP attribute to use to retrieve email address.",
+        default='mail')
 
     def get_ldap_dicts(self, cr, ids=None):
```
lasley, line 42: `api.multi`

users_ldap_mail/users_ldap_model.py
```diff
@@ -60,18 +57,26 @@ def get_ldap_dicts(self, cr, ids=None):
         """, args)
         return cr.dictfetchall()
 
-    def map_ldap_attributes(self, cr, uid, conf, login, ldap_entry):
-        values = super(CompanyLDAP, self).map_ldap_attributes(
-            cr, uid, conf, login, ldap_entry)
+    def get_or_create_user(self, cr, uid, conf, login, ldap_entry,
```
lasley, line 60: `api.model`

users_ldap_mail/users_ldap_model.py
```diff
@@ -19,32 +19,29 @@
 #
 ##############################################################################
```
lasley, line 20: Please convert all headers to the new style - https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

users_ldap_mail/users_ldap_model.py
```diff
@@ -60,18 +57,26 @@ def get_ldap_dicts(self, cr, ids=None):
...
         for value_key, conf_name in mapping:
             try:
                 if conf[conf_name]:
                     values[value_key] = ldap_entry[1][conf[conf_name]][0]
             except KeyError:
-                _log.warning('No LDAP attribute "%s" found for login  "%s"' % (
-                    conf.get(conf_name), values.get('login')))
-        return values
+                _logger.warning(
+                    'No LDAP attribute "%s" found for login  "%s"'
```
lasley, line 76: Use translation here `_('No LDAP attribute "%s" found for login  "%s"') % ..`

users_ldap_mail/users_ldap_model.py
```diff
@@ -60,18 +57,26 @@ def get_ldap_dicts(self, cr, ids=None):
...
+                    % (conf.get(conf_name), values.get('login')))
+
+        new_user = self.pool['res.users'].browse(cr, SUPERUSER_ID, user_id)
```
lasley, line 79: Once you switch this out for new api, this will be `self.env['res.users'].sudo().browse()`

Although I don't see the sudo/superuser in the previous code. Can you please elaborate as to what problem this is circumventing?

lasley:
@tstonis Thanks for the contribution. I have put some comments inline in order to hopefully assist w/ the API upgrade. Please let me know if further clarification is required.

users_ldap_mail/users_ldap_model.py
tstonis, line 42: I tried api.multi before and it gave me an error (expecting more arguments, etc) because it assumes the signature is (cr, uid, ids, *arguments, context), but that is not the signature in res.company.ldap for the function being overridden.

users_ldap_mail/users_ldap_model.py
tstonis, line 79: The previous code was overriding map_ldap_attributes, but the error I am fixing is because attributes like 'email' cause some kind of a cache error when added to the login dictionary and calling get_or_create_user. My fix is to override get_or_create_user instead and then modify the user. I used superuser, because that is the way the browse of res.users to get the user object is done in res.company.ldap. I hope this explains.

users_ldap_mail/users_ldap_model.py
lasley, line 42: Yuck, you're right 😑

users_ldap_mail/users_ldap_model.py
lasley, line 79: Got it, this makes sense because not all users can edit themselves. Thanks for the explanation

lasley:
@tstonis - I just tried to replicate this issue on my machine, and I am not able to. I am wondering if we instead have something else going on here.

Let me make sure I am understanding the issue correctly - when a new user with a mapped email logs into your server, instead of the user being correctly created and allowing login it is instead popping up a `Record does not exist or has been deleted.` error?

I assume previously existing users work fine?

Can you paste the whole traceback from your server log please?

tstonis:
@lasley Yes, that's correct. I am binding to an Active Directory server to get the mail and user name attribute. I have tried this on two machines, with two different databases, and both had the same problem. Below is the full stack trace:

2016-08-12 13:41:42,373 66945 INFO Cardolite_Trading werkzeug: 10.10.20.247 - - [12/Aug/2016 13:41:42] "POST /web/login HTTP/1.0" 500 -
2016-08-12 13:41:42,395 66945 ERROR Cardolite_Trading werkzeug: Error on request:
  File "/Library/Python/2.7/site-packages/werkzeug/serving.py", line 177, in run_wsgi
    execute(self.server.app)
  File "/Library/Python/2.7/site-packages/werkzeug/serving.py", line 165, in execute
    application_iter = app(environ, start_response)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/service/server.py", line 246, in app
    return self.app(e, s)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/service/wsgi_server.py", line 184, in application
    return application_unproxied(environ, start_response)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/service/wsgi_server.py", line 170, in application_unproxied
    result = handler(environ, start_response)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 1494, in **call**
    return self.dispatch(environ, start_response)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 1468, in **call**
    return self.app(environ, start_wrapped)
  File "/Library/Python/2.7/site-packages/werkzeug/wsgi.py", line 588, in **call**
    return self.app(environ, start_response)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 1658, in dispatch
    result = ir_http._dispatch()
  File "/Users/admin/Documents/Odoo/eval/odoo/addons/web_editor/models/ir_http.py", line 16, in _dispatch
    return super(ir_http, self)._dispatch()
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/addons/base/ir/ir_http.py", line 186, in _dispatch
    return self._handle_exception(e)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/addons/base/ir/ir_http.py", line 157, in _handle_exception
    return request._handle_exception(exception)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 785, in _handle_exception
    return super(HttpRequest, self)._handle_exception(exception)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/addons/base/ir/ir_http.py", line 182, in _dispatch
    result = request.dispatch()
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 844, in dispatch
    r = self._call_function(*_self.params)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 320, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/service/model.py", line 118, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 313, in checked_call
    result = self.endpoint(_a, *_kw)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 963, in __call__
    return self.method(_args, *_kw)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 513, in response_wrap
    response = f(_args, *_kw)
  File "/Users/admin/Documents/Odoo/eval/odoo/addons/auth_signup/controllers/main.py", line 20, in web_login
    response = super(AuthSignupHome, self).web_login(_args, *_kw)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 513, in response_wrap
    response = f(_args, *_kw)
  File "/Users/admin/Documents/Odoo/eval/odoo/addons/web/controllers/main.py", line 484, in web_login
    uid = request.session.authenticate(request.session.db, request.params['login'], request.params['password'])
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 1126, in authenticate
    if uid: self.get_context()
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/http.py", line 1161, in get_context
    self.context = request.registry.get('res.users').context_get(request.cr, request.uid) or {}
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "<string>", line 2, in context_get

  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/tools/cache.py", line 85, in lookup
    value = d[key] = self.method(_args, *_kwargs)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/addons/base/res/res_users.py", line 428, in context_get
    res = getattr(user, k) or False
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/fields.py", line 835, in **get**
    return record._cache[self]
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/models.py", line 6132, in __getitem__
    return value.get() if isinstance(value, SpecialValue) else value
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/fields.py", line 39, in get
    raise self.exception
MissingError: (u'Record does not exist or has been deleted.', None)

lasley:
Interesting - I don't see this module in your traceback, so while this fix is helping, it is something else. Looks like it's dead during the res_users copy from the template user for some reason. Do you by any chance get this same error if you duplicate the Template User through UI?

tstonis:
@lasley I don't think it's the duplication, because I tried it with no template user and had the same problem. I don't know if it's related, but the email to the user does fail before this (because I don't have an SMTP server configured). Below is the log for that. I don't know how that email failing would affect things. Note that I made the change that I did because I narrowed it down to the map_ldap_attributes call using pdb and when I took the 'email' key out of the mapped attributes, it worked. If I left in just the 'name' attribute to map it worked okay... 

2016-08-12 13:39:13,086 66945 INFO Cardolite_Trading werkzeug: 10.10.20.247 - - [12/Aug/2016 13:39:13] "GET /web/login HTTP/1.0" 200 -
2016-08-12 13:41:42,217 66945 INFO Cardolite_Trading openerp.addons.base.ir.ir_mail_server: Mail delivery failed via SMTP server 'localhost'.
error: 61
Connection refused
2016-08-12 13:41:42,218 66945 ERROR Cardolite_Trading openerp.addons.mail.models.mail_mail: failed sending mail (id: 6) due to Mail Delivery Failed
Mail delivery failed via SMTP server 'localhost'.
error: 61
Connection refused
  File "/Users/admin/Documents/Odoo/eval/odoo/addons/mail/models/mail_mail.py", line 262, in send
    res = IrMailServer.send_email(msg, mail_server_id=mail.mail_server_id.id)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/api.py", line 248, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/api.py", line 490, in new_api
    result = method(self._model, cr, uid, _args, *_old_kwargs)
  File "/Users/admin/Documents/Odoo/eval/odoo/openerp/addons/base/ir/ir_mail_server.py", line 483, in send_email
    raise MailDeliveryException(_("Mail Delivery Failed"), msg)
MailDeliveryException: (u'Mail Delivery Failed', u"Mail delivery failed via SMTP server 'localhost'.\nerror: 61\nConnection refused"

lasley:
You're right that the failed email should have no correlation to the actual user creation - can confirm because I have an Odoo server using this code that isn't rigged w/ SMTP.

I am also under the belief that the email attribute is the key here, but I'm utterly confused as to what the problem could be. I realized that I have this code in use on nearly all of my customers' Odoos with no issues too, which is just even more peculiar.

I'm wondering if maybe this has to do with the AD schema itself, because you have already eliminated the DB issue w/ multiple Odoos. When you tried on the other Odoo, was this (and dependencies) the only module that was installed?

tstonis:
I don't think it's related to AD, because the LDAP call works just fine. I was really at my wits end trying to figure out the problem, which is why I ended up patching this when I narrowed it down to the email attribute being passed to the get_or_create_user. I didn't have time to test it from a fresh or minimum Odoo install. If you're not having a problem, and no one else is, we can just close this. Sorry if I wasted your time, but I appreciate your attention!

lasley:
Hah yup I know how it is after trying to figure something like this out 😛 

I'm thinking there was something else at play here, but got me as to what. I'm going to close this & if you are able to replicate at some point with a fresh install, or identify the plugin mixture causing the conflict, please let me know.

PR #521, opened 2016-08-17
[8.0][FIX][base_export_manager] Fix "Expected singleton" bug.

yajo:
If you had a field that got translated in more than 1 addon, you'd possibly get
to this error:

```
  File "/opt/odoo/0079_ahk_openerp/oca/base_export_manager/models/ir_exports_line.py", line 105, in _compute_label
    field.name)),
  File "/opt/odoo/common/openerp/v8/openerp/fields.py", line 825, in __get__
    record.ensure_one()
  File "/opt/odoo/common/openerp/v8/openerp/models.py", line 5355, in ensure_one
    raise except_orm("ValueError", "Expected singleton: %s" % self)
except_orm: ('ValueError', 'Expected singleton: ir.translation(4899, 703976)')
```

With this patch, now we let Odoo return the translated string by using its
standard method to do so, so we have to care for less.

@Tecnativa

### DIFF
--- a/base_export_manager/hooks.py
+++ b/base_export_manager/hooks.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openerp import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    """Loaded after installing the module.
+
+    ``ir.exports.line.name`` was before a char field, and now it is a computed
+    char field with stored values. We have to inverse it to avoid database
+    inconsistencies.
+    """
+    env = api.Environment(cr, SUPERUSER_ID, dict())
+    env["ir.exports.line"].search([
+        ("field1_id", "=", False),
+        ("export_id", "!=", False),
+        ("name", "!=", False),
+    ])._inverse_name()
--- a/base_export_manager/hooks.py
+++ b/base_export_manager/hooks.py
@@ -11,9 +11,10 @@ def post_init_hook(cr, registry):
     char field with stored values. We have to inverse it to avoid database
     inconsistencies.
     """
-    env = api.Environment(cr, SUPERUSER_ID, dict())
-    env["ir.exports.line"].search([
-        ("field1_id", "=", False),
-        ("export_id", "!=", False),
-        ("name", "!=", False),
-    ])._inverse_name()
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        env["ir.exports.line"].search([
+            ("field1_id", "=", False),
+            ("export_id", "!=", False),
+            ("name", "!=", False),
+        ])._inverse_name()
--- a/base_export_manager/models/ir_exports_line.py
+++ b/base_export_manager/models/ir_exports_line.py
@@ -14,7 +14,6 @@ class IrExportsLine(models.Model):
 
     name = fields.Char(
         required=False,
-        readonly=True,
         store=True,
         compute="_compute_name",
         inverse="_inverse_name",
--- a/base_export_manager/models/ir_exports_line.py
+++ b/base_export_manager/models/ir_exports_line.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # © 2015 Antiun Ingeniería S.L. - Antonio Espinosa
 # Copyright 2015-2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+# Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openerp import models, fields, api, exceptions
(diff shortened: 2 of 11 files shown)


yajo:
Please could somebody forward-patch this to v9 when merged? Or do I open a separate PR?

yajo:
Before https://github.com/OCA/server-tools/pull/521/commits/9ef4280b6c529cf36404365eb963cbe442deea29, if you had a buggy record in database, you got an endless error loop. Now you get this beautiful message:

captura de pantalla de 2016-08-17 13-40-42

I considered adding a backup column of `ir.exports.line.name`, but it's not necessary because now installation gets rolled back if anything fails, and it does not lose data.

base_export_manager/hooks.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def post_init_hook(cr, registry):
+    """Loaded after installing the module.
+
+    ``ir.exports.line.name`` was before a char field, and now it is a computed
+    char field with stored values. We have to inverse it to avoid database
+    inconsistencies.
+    """
+    env = api.Environment(cr, SUPERUSER_ID, dict())
```
pedrobaeza:
```
with api.Environment.manage():
```

base_export_manager/hooks.py
```diff
@@ -0,0 +1,19 @@
...
+    env["ir.exports.line"].search([
+        ("field1_id", "=", False),
+        ("export_id", "!=", False),
+        ("name", "!=", False),
+    ])._inverse_name()
```
pedrobaeza: I still think we should protect this and don't stop. OK, now we are not in a loop, but forcing to fix one by one each of the possible old fields is an annoyance.

yajo: But the option of leaving records in broken status is not very desirable too... :confused:

The only option then I can imagine is unlinking wrong records.

pedrobaeza: They are broken, yes, but you can correct all of them in a whole in the corresponding menu, and the break is only when trying to use the corresponding export profile, so it's not a big problem nor breaks the DB.

pedrobaeza: Also when migrating from previous versions, you get the loop in the update also (I'm just facing this right now).

pedrobaeza:
One thing I'm facing: can it be that the textual representation of the fields (field `name`) have changed? It seems that previously you have `account_id / code` for representing a sub field, and now it's `account_id/code` (without spaces).

base_export_manager/models/ir_exports_line.py
```diff
@@ -162,9 +143,17 @@ def _get_field_id(self, model, name):
         :param str name:
             Technical name of the field, like ``child_ids``.
         """
-        return self.env["ir.model.fields"].search(
+        result = self.env["ir.model.fields"].search(
             [("name", "=", name),
              ("model_id", "=", model.id)])
+        try:
```
pedrobaeza: Use `if not result.exists():` instead

yajo:
Just to clarify: https://github.com/OCA/server-tools/pull/521#issuecomment-240485437 was not a bug.

Is this now ready to review?

pedrobaeza:
Indeed it is

pedrobaeza:
:+1: the part I haven't done

pedrobaeza:
@max3903 can you check this, please?

base_export_manager/models/ir_exports_line.py
```diff
@@ -53,104 +62,106 @@ def _default_model1_id(self):
...
-                    ]).value or
-                    field.display_name)
-            s.label = ("%s (%s)" % ("/".join(parts), s.name)
-                       if parts and s.name else False)
+                    self.env[self.model_n(num).model]._fields[field.name]
+                    .get_description(self.env)["string"])
+            self.label = ("%s (%s)" % ("/".join(parts), self.name)
+                          if parts and self.name else False)
+        except KeyError:
+            pass
```
yajo: I'd thank a little comment about why this is necessary.

base_export_manager/models/ir_exports_line.py
```diff
@@ -53,104 +62,106 @@ def _default_model1_id(self):
...
+                _("It's not allowed to have more than 4 levels depth: "
+                  "%s") % self.name)
+        for num in range(1, 5):
+            if num > len(parts):
+                # Empty subfield in this case
+                self[self.field_n(num, True)] = False
+                continue
+            field_name = parts[num - 1]
+            model = self.model_n(num)
+            self.with_context(skip_check=True)[self.field_n(num, True)] = (
```
yajo: This skip trick also deserves a little comment.

base_export_manager/views/ir_exports_view.xml
```diff
@@ -47,33 +47,25 @@
                         <field name="model1_id" invisible="True"/>
                         <field name="model2_id" invisible="True"/>
                         <field name="model3_id" invisible="True"/>
+                        <field name="model4_id" invisible="True"/>
                         <field name="label"/>
-                        <field name="name"/>
+                        <field name="name" readonly="0"/>
```
yajo: I'd prefer it if you remove the readonly from the model itself.

base_export_manager/tests/test_ir_exports_line.py
```diff
@@ -34,3 +42,44 @@ def test_get_label_string(self):
...
+        self.assertEqual(line.field3_id, self.field_name)
+
+    def test_compute_name(self):
+        line = self.env['ir.exports.line'].create({
+            'export_id': self.export.id,
+            'field1_id': self.field_parent_id.id,
+            'field2_id': self.field_parent_id.id,
+            'field3_id': self.field_name.id,
+        })
+        self.assertEqual(line.name, 'parent_id/parent_id/name')
```
yajo: Could you add the 4th field here? It's a simple patch and covers more cases.

base_export_manager/tests/test_ir_exports_line.py
```diff
@@ -34,3 +42,44 @@ def test_get_label_string(self):
...
+            default_model1_id=self.export.model_id.id).create({
+                "name": "name",
+                "export_id": self.export.id,
+            })
+        self.assertEqual(line.model1_id, self.export.model_id)
+
+    def test_inverse_name(self):
+        line = self.env['ir.exports.line'].create({
+            'export_id': self.export.id,
+            'name': 'parent_id/parent_id/name',
```
yajo: Could you add the 4th field here? It's a simple patch and covers more cases.

yajo:
I'm thinking that maybe a good improvement would be to give this addon its own menu item. Non-admins may want to manage this too, given what this addon ultimately does is make that task easier. Maybe it could be handled by a group setting. Not a requirement though.

base_export_manager/models/ir_exports_line.py
lasley: IMO it would also be a little nicer if we guarded a little closer to the key error. We could then just sub the `pass` for a `return` to accomplish the same effect

base_export_manager/models/ir_exports_line.py
```diff
@@ -53,104 +62,106 @@ def _default_model1_id(self):
...
+        self.model3_id = (
+            self.field2_id.ttype and
+            "2" in self.field2_id.ttype and
+            ir_model.search([("model", "=", self.field2_id.relation)]))
 
-    @api.multi
+    @api.one
+    @api.depends("field3_id")
+    def _compute_model4_id(self):
+        """Get the related model for the third field."""
```
lasley, line 99: s/third/fourth

base_export_manager/models/ir_exports_line.py
```diff
@@ -53,104 +62,106 @@ def _default_model1_id(self):
...
-                # Prevent possible inexisting fields
-                pass
+        if not self.env.context.get('skip_check'):
+            lines = self.search([('export_id', '=', self.export_id.id),
+                                 ('name', '=', self.name)])
+            if len(lines) > 1:
+                raise exceptions.ValidationError(
+                    _("Field '%s' already exists") % self.name)
+
+    @api.onchange('name')
```
lasley, line 167: Isn't an `api.multi` required for onchange?

base_export_manager/models/ir_exports_line.py
yajo: I like the idea.

base_export_manager/models/ir_exports_line.py
yajo, line 167: `@api.multi` is implicit if the method signature does not start by `self, cr, uid, ...`. Its advantage is that makes methods callable from old api too, but if that's not a requirement, you don't need to put it. I'm more in the side of always decorating, but certainly it's not required.

max3903:
@pedrobaeza Sorry, I missed your message. What do you want me to check?

pedrobaeza:
Well, the idea is to see the improvements and give your bless.

max3903:
:+1: LGTM

rafaelbn:
Functionaly tested :+1:

base_export_manager/models/ir_exports_line.py
```diff
@@ -53,104 +62,106 @@ def _default_model1_id(self):
         """Default model depending on context."""
         return self.env.context.get("default_model1_id", False)
 
-    @api.multi
-    @api.depends("field1_id", "field2_id", "field3_id")
+    @api.one
```
dreispt: Isn't best to leave api.maulti here, since we have confirmation that Odoo intends to deprecated api.one ?

base_export_manager/models/ir_exports_line.py
dreispt: I concur with @yajo: it's not obvious why the `pass` is needed, an explaining comment would be welcome.

base_export_manager/models/ir_exports_line.py
pedrobaeza: I prefer `@api.one` while it exists. And I repeat that it's still not clear that it will be deprecated.

dreispt: I don't see the point for the extra work when doing the port to v9.
And when asking to reconsider deprecation, all comments I had were against keeping it. So I don't see must space for any hopes on that.

pedrobaeza: No, there's no work needed. `@api.one` is still in v9, v10, and counting.

dreispt: Won't Lint report errors?

pedrobaeza: No, they are only warnings. The PR in MQT was for removing the warning.

pedrobaeza:
@Yajo, can you attend last comments (including yours)?

pedrobaeza (approved):

pedrobaeza:
Merging as the problem in Travis fluctuates between builds.

PR #526, opened 2016-08-23
[9.0][MIG] auto_backup

lasley:
This migrates auto_backup to v9

Note this was mostly unchanged except for two self.ensure_ones in db_backup.py


auto_backup/data/backup_data.yml
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
elicoidal: Arenot yml test files deprecated?

auto_backup/tests/test_auto_backup.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TestsAutoBackup(common.TransactionCase):
+
+    def setUp(self):
+        super(TestsAutoBackup, self).setUp()
+
+    def new_record(self, method='sftp'):
+        vals = {
+            'name': u'Têst backup',
```
elicoidal: is the accent ('ê') intentional?

lasley: Not sure, but I left it assuming that it was testing for handling of special chars in the file name

auto_backup/data/backup_data.yml
lasley: Are they? I would admittedly be very happy about that. Where'd you see this?

lasley:
FYI I'll be adding more tests as well

auto_backup/data/backup_data.yml
elicoidal: Not sure where: I will try to track down...

sebalix: Maybe this? https://github.com/odoo/odoo/issues/6287#issuecomment-240459221 YAML is not listed

elicoidal: there is no mention of it in v9 doc: https://www.odoo.com/documentation/9.0/reference/testing.html

elicoidal: @sebalix nice catch but indeed that confirms my first comment

pedrobaeza: I confirm that YAML tests are deprecated on v9 and removed on v10

lasley: Awesome well that just made me incredibly happy - I hate the YML tests. I'll move this to Python

rvalyi: @pedrobaeza do you have any pointer about the fact they are removed in v10? or deprecated in v9? May be we should discuss it at the OCA... I think it's a bit late to require no YAML test anymore in v9 given the OCA is already late on that v9 transition...

pedrobaeza: You can see in the comment above (https://github.com/odoo/odoo/issues/6287#issuecomment-240459221) the removal part. About deprecation, I don't remember where it was said, but it has been stated several times.

I see no point in discussing this: if the author of the PR agrees to change the test, perfect. If not, better YML test than nothing. In v10 we will need to change them. It's not a blocking point.

rvalyi: @pedrobaeza well if we still accept YML tests on 9.0 that's fine for me. The contrary would have annoyed me cause it would have slowed migration a bit too much... On the contrary, 9.0 to 10.0 migration should be pretty straightforward so it will be easier to bite the bullet.

lasley: In this case we're talking maybe a ten minute conversion too, so really not too much skin off my back even if I didn't already want to see YML die in a fire 😄 

Just to note though, this isn't a test - it's the cron and message data.

lasley:
Alright comments handled and test coverage raised. Methods still missing coverage, but I figure they're probably fine given it was all pre-existing code:
- backup_log
- cleanup
- cleanup_log

max3903 (approved):

ivantodorovich (approved):
LGTM 👍

lmignon:
@max3903 @lasley @pedrobaeza This merge must be reverted. The commit history is not preserved!
https://github.com/OCA/server-tools/issues/619

max3903:
@lmignon Done

PR #522, opened 2016-08-18
9.0 auth admin passkey

rvalyi:
same as 8.0 https://github.com/OCA/server-tools/pull/519
but also migrated the view and enable the module on 9.0
cc @legalsylvain @pedrobaeza

PR #527, opened 2016-08-23
Authpasskey9

JayVora-SerpentCS:
V9 migration for the auth_admin_passkey module.


pedrobaeza:
There's already an ongoing migration on https://github.com/OCA/server-tools/pull/522

PR #531, opened 2016-08-26
[9.0][ADD] Password Security Settings

lasley:
# 
# Password  Security

This module allows admin to set company-level password security requirements
and enforces them on the user.

It contains features such as
- Password expiration days
- Password length requirement
- Password minimum number of lowercase letters
- Password minimum number of uppercase letters
- Password minimum number of numbers
- Password minimum number of special characters
# Configuration
- Navigate to company you would like to set requirements on
- Click the `Password Policy` page
- Set the policies to your liking.

Password complexity requirements will be enforced upon next password change for
any user in that company.
## Settings & Defaults

These are defined at the company level:

| Name | Default | Description |
| --- | --- | --- |
| password_expiration | 60 | Days until passwords expire |
| password_length | 12 | Minimum number of characters in password |
| password_lower | True | Require lowercase letter in password |
| password_upper | True | Require uppercase letters in password |
| password_numeric | True | Require number in password |
| password_special | True | Require special character in password |


lasley:
~~Hi @oca-clabot - @jamesfoster85 was covered under LasLabs ECLA when his commit was made~~ I squashed away from his commit anyways now 😄

lasley:
Going to add tests to cover the controller, but would love to know anyone's thought in the interim 😄

res_users_password_security/README.rst
```diff
@@ -0,0 +1,94 @@
...
+and enforces them on the user.
+
+It contains features such as
+
+* Password expiration days
+* Password length requirement
+* Password minimum number of lowercase letters
+* Password minimum number of uppercase letters
+* Password minimum number of numbers
+* Password minimum number of special characters
```
elicoidal: One common requirement for password security is the impossibility to repeat passwords. Would it be a good idea to add (maybe in roadmap?)

res_users_password_security/README.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Configuration
+=============
+
+* Navigate to company you would like to set requirements on
```
elicoidal: use enumerated list for procedures ("#. ")

res_users_password_security/README.rst
```diff
@@ -0,0 +1,94 @@
...
+Password complexity requirements will be enforced upon next password change for
+any user in that company.
+
+
+Settings & Defaults
+-------------------
+
+These are defined at the company level:
+
++---------------------+---------+------------------------------------------+
```
elicoidal: just a note: why dont use simple rst table format? much easier to handle and maintain. In your case you dont need the complex format

res_users_password_security/models/res_users.py
```diff
@@ -0,0 +1,87 @@
...
+        return super(ResUsers, self).write(vals)
+
+    @api.multi
+    def password_match_message(self):
+        company_id = self.company_id
+        message = [
+            'Password must be %d characters or more.' %
+            company_id.password_length
+        ]
+        message.append('Must contain the following:')
```
elicoidal: What if all checkboxes are set to false? you would still get the message while not necessary, right?

res_users_password_security/tests/company.yml
```diff
@@ -0,0 +1,10 @@
+-
```
elicoidal: yml tests are deprecated. Please use py tests.

res_users_password_security/views/res_company_view.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+    Copyright 2015 LasLabs Inc.
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<openerp>
```
elicoidal: odoo

res_users_password_security/views/res_company_view.xml
```diff
@@ -0,0 +1,42 @@
...
+    <data>
```
elicoidal: you can remove

res_users_password_security/README.rst
```diff
@@ -0,0 +1,94 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
elicoidal: be careful: be consistent in your licensing. Everywhere in your headers you state LGPL

elicoidal:
@lasley Thanks for the module !!!!

res_users_password_security/README.rst
lasley: We're talking about the reuse of old passwords right? 

Will probably need to add something to limit the amount of resets in a timespan as well then - I know that's how I get around these types of requirements 😛

elicoidal: yes indeed.

lasley: I gave it a valiant effort, but my strategy seems to be a bust. These `pbkdf2_*` hashes return a different value each time, similar to encryption. This defeats the normal strategy of saving and referencing the pass hashes directly for use as a history. This just blew my limited knowledge of cryptography out of the water - crazy smart people with their crazy math.

I'm open to ideas if you have them. The last commit removes my attempt if you want to see what I had.

```
>>> pbkdf2_sha512.encrypt("password")
'$pbkdf2-sha512$25000$4ByDsPa.lxJCaE2JkVLK.Q$dVDsbr49NHfKEQWYdjunuqbz2a2MGQmj0rfF3Bh519StAUouvCyYag50EL7kjBckyArqDRBdrwDHkeWeHva5DA'
>>> pbkdf2_sha512.encrypt("password")
'$pbkdf2-sha512$25000$RYixlnKOMab0vjemtNa6lw$d/8zZUv4Ttjfj31/LrXulIk8EuyPPptGpwWllm8jEtU0Ehoe7UfjPmzdn8KwpM8o6eTShCFjn7TvfFzSpf.J3Q'
>>> pbkdf2_sha512.encrypt("password")
'$pbkdf2-sha512$25000$AQBgLEVoLSVEKIUwZuzd.w$BJkX.U1YJTO5Q.g5I/1kh4guNF0qG/1/cYJ8amhFB0JSRT65oPtuJJtwGvNNUGOcdElKuNsXyyitAhcWZdK6Ng'
```

hbrunn: modern hashing algorithms don't only depend on the input string, but also a salt (another input string) and the amount of rounds: https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#interface
Which is a good thing, and you shouldn't make an attempt to fix the salt or whatever. Just take the strings as they are, and try new passwords against them. The result contains those values, so you don't need to worry about it: https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#format-algorithm

lasley: I guess I've only seen crap implementations like Magento then 😛 . 

This make sense though, so basically instead of trying to reference hash vs hash, I need to validate the non-hashed password against the previous X hashed. Piece of cake, thanks @hbrunn

lasley: 👍  worked like a champ, thanks again! Implementing the minimum reset time then I think we'll be good.

res_users_password_security/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.0.2',
```
lasley: FYI I have to leave this version number, this is already in the wild under our repos. We are migrating here and will be having our's removed from the app store for a hopefully seamless transition.

res_users_password_security/README.rst
elicoidal: :+1: excellent. 
If I understand well you will have to save a history of hashes so that we make sure the user doesnot repeat passwords.

lasley: Yup that's implemented in [`res.users.pass.history`](https://github.com/laslabs/server-tools/blob/9d20afcbaebc520818ebb8f9462d7932ebe5c48d/res_users_password_security/models/res_users_pass_history.py#L8) 😄

res_users_password_security/__openerp__.py
elicoidal: let's say this is not yet critical until merge. After the merge we should stick strictly to the rule

lasley: Once merged, people that are using the `9.0.1.0.1` version from my repo will need a way to indicate the change won't they? I honestly have no idea how the Odoo App store works, but I figured these version numbers has some relevance to the auto update?

elicoidal: My point is follow your internal logic as this is not important at which number you start in the OCA repo. What is important though is that after the merge the numbers are dealt logically

res_users_password_security/README.rst
elicoidal: Great :)

res_users_password_security/__openerp__.py
lasley: Ahh cool thanks for elaborating 👍

lasley:
Alright assuming this build goes green, I'm done here 😄 

Will squash my commits once we get the thumbs

res_users_password_security/models/res_company.py
```diff
@@ -0,0 +1,51 @@
...
+    password_special = fields.Boolean(
+        'Special',
+        default=True,
+        help='Require special characters',
+    )
+    password_history = fields.Integer(
+        'History',
+        default=30,
+        help='Disallow reuse of this many previous passwords - use negative '
+             'number for infinite',
```
elicoidal: "and 0 to disable"

res_users_password_security/models/res_users.py
```diff
@@ -0,0 +1,158 @@
...
+        if company_id.password_lower:
+            message.append('* ' + _('Lowercase letter'))
+        if company_id.password_upper:
+            message.append('* ' + _('Uppercase letter'))
+        if company_id.password_numeric:
+            message.append('* ' + _('Numeric digit'))
+        if company_id.password_special:
+            message.append('* ' + _('Special character'))
+        if len(message):
+            message = [_('Must contain the following:')]
```
elicoidal: don't you overwrite the message variable here?

lasley:
Thanks for the catches @elicoidal - handled

dreispt:
AFAICS we already have one (or more) PR for this feature.
Did you check #294 ?

lasley:
@dreispt - I was unaware of that PR, but there are too many issues with it in my opinion & it doesn't seem like it is being worked. 

Main non-starter in my eyes is the breaking of the inheritance chain, massive code duplication & 0% test coverage. 

You're welcome to block this PR in favor of that if you would like, but unfortunately I will not be fixing that one as there are more features here, it is tested, and we are not breaking inheritance chains.

dreispt:
I don't want to block anything, we just need to avoid duplicate PRs.
The same story happened with the previous one - there were two different PRs for similar features.
I involved both authors in a discussion of the pros and cons of each, and finally the most promising PR was selected and the other one was closed.

So, we need to invite the other PR's author(s) here, compare the features here with the referenced PR. If  the other PR won't go forward, at least there may be good ideas there that could be picked up here, and we could ensure that this PR also fits the use cases for the other PR.

lasley:
Cool I understand the motivation, thanks for elaborating @dreispt

@hhgabelgaard - I seem to have duplicated your functionality from #294. I propose that we merge this PR instead, then backport to v9 in order to provide the same features of your PR plus
- Handling for lowercase/uppercase rules
- Minimum password length (aka user cannot reset for 24 hours since last reset)
- Disallow use of X last passwords
- Password expiration

res_users_password_security/views/res_company_view.xml
```diff
@@ -0,0 +1,42 @@
...
+                        <group>
+                            <field name="password_numeric" />
+                            <field name="password_special" />
+                        </group>
+                    </group>
+                </page>
+            </xpath>
+        </field>
+    </record>
+
```
bealdav: May be a module you don't know could have been used here.
https://github.com/OCA/server-tools/tree/9.0/configuration_helper
It allow to define settings easily on settings page.
Non blocking for me

res_users_password_security/README.rst
```diff
@@ -0,0 +1,90 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==================
+Password  Security
```
bealdav: I think module technical name could be 'password_security'. 
Display name and technical name should be the same (except plural maybe), I think at least.
2 words is quickly read, and when we read password, everybody think about users.

bealdav:
I have tested this code with V8. It works, except <odoo> instead of <openerp>  in xml files

Thanks for this really good job fully tested.
:+1: 

@hhgabelgaard are you ok to close your PR  #294 and vote +1 here

bealdav (approved):

res_users_password_security/controllers/main.py
```diff
@@ -0,0 +1,90 @@
...
+        redirect = user_id.partner_id.signup_url
+        return http.redirect_with_hash(redirect)
+
+    @http.route()
+    def web_auth_signup(self, *args, **kw):
+        try:
+            return super(PasswordSecurityHome, self).web_auth_signup(
+                *args, **kw
+            )
+        except PassError, e:
```
moylop260: Could you use `except PassError as e:` instead, please?

moylop260: More info: https://www.python.org/dev/peps/pep-3110

hhgabelgaard:
I agree, this seems better than #294 
+1

res_users_password_security/controllers/main.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        qcontext = self.get_auth_signup_qcontext()
+        if (
+            request.httprequest.method == 'POST' and
+            qcontext.get('login') and
+            'error' not in qcontext and
+            'token' not in qcontext
+        ):
+            login = qcontext.get('login')
+            user_ids = request.env.sudo().search([('login', '=', login)])
```
moylop260: Could you use `limit=1`, please?

res_users_password_security/controllers/main.py
```diff
@@ -0,0 +1,90 @@
...
+            if not user_ids:
+                user_ids = request.env.sudo().search(
```
moylop260: Could you use `limit=1`, please?

res_users_password_security/controllers/main.py
```diff
@@ -0,0 +1,90 @@
...
+                    [('email', '=', login)],
+                )
+            if len(user_ids) == 1:
+                user_ids[0]._validate_pass_reset()
```
moylop260: With `limit=1` and new api you don't need a `if` validation here.

``` diff
user_ids = request.env.sudo().search(... limit=1)
...
- if len(user_ids) == 1:
-    user_ids[0]._validate_pass_reset()
+ user_ids._validate_pass_reset()
```

res_users_password_security/models/res_users_pass_history.py
```diff
@@ -0,0 +1,25 @@
...
+class ResUsersPassHistory(models.Model):
+    _name = 'res.users.pass.history'
+    _description = 'Res Users Password History'
+
+    _order = 'user_id, date desc'
+
+    user_id = fields.Many2one(
+        string='User',
+        comodel_name='res.users',
+        index=True,
```
moylop260: If we delete a user: Should we store the pass history?
If the answer is `no`: you could use `ondelete=cascade`

moylop260 (changes requested):

res_users_password_security/models/res_users.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @api.multi
+    def _password_has_expired(self):
+        self.ensure_one()
+        if not self.password_write_date:
+            return True
+        write_date = fields.Datetime.from_string(self.password_write_date)
+        today = fields.Datetime.from_string(fields.Datetime.now())
+        days = (today - write_date).days
+        return (days > self.company_id.password_expiration)
```
moylop260: We don't need the parens here
`return days > self.company_id.password_expiration`

res_users_password_security/tests/test_password_security_home.py
```diff
@@ -0,0 +1,270 @@
...
+
+IMPORT = 'openerp.addons.res_users_password_security.controllers.main'
+
+
+class EndTestException(Exception):
+    """ It allows for isolation of resources by raise """
+
+
+class MockResponse(object):
+    def __new__(self):
```
moylop260: for `__new__` method use `cls` instead of `self`

res_users_password_security/tests/test_password_security_home.py
```diff
@@ -0,0 +1,270 @@
...
+    def __init__(self):
+        super(MockPassError, self).__init__('Message')
+
+
+class TestPasswordSecurityHome(TransactionCase):
+
+    def setUp(self):
+        super(TestPasswordSecurityHome, self).setUp()
+        self.Controller = main.PasswordSecurityHome
+        self.controller = self.Controller()
```
moylop260: What about use a name of variable different with `snake_case` for `Controller`?

res_users_password_security/tests/test_password_security_session.py
```diff
@@ -0,0 +1,56 @@
...
+class EndTestException(Exception):
+    """ It allows for isolation of resources by raise """
+
+
+class TestPasswordSecuritySession(TransactionCase):
+
+    def setUp(self):
+        super(TestPasswordSecuritySession, self).setUp()
+        self.Controller = main.PasswordSecuritySession
+        self.controller = self.Controller()
```
moylop260: What about use a name of variable different with `snake_case` for `Controller`?

lasley:
Thanks for the reviews @bealdav & @moylop260 - all changes made outside of the `configuration_helper`. You are correct that I was unaware of the existence of this module, but will need to play around with it a bit first I think.

moylop260 (approved):
Thanks

moylop260:
Could you review the red build of travis related with this module, please?
https://travis-ci.org/OCA/server-tools/jobs/162977224

lasley:
Alright we should be good on Travis and your last comments @moylop260

moylop260:
ups lint show

``` bash
password_security/tests/test_password_security_home.py:235:17: F841 local variable 'search' is assigned to but never used
password_security/tests/test_password_security_home.py:264:17: F841 local variable 'search' is assigned to but never used
password_security/tests/test_password_security_home.py:266:17: F841 local variable 'user' is assigned to but never used
```

lasley:
Woot build 🍏

bealdav:
:+1: thanks

dreispt:
🎉

Issue #529, opened 2016-08-24
[auth_admin_passkey] User account is locked each time we use auth_admin_passkey


nowheresly:
We are using auth_admin_passkey in our odoo 8.0 instance with the auth_ldap module.
Our LDAP is an Active Directory (AD).
Unfortunately, each time we use the admin password to log in, the corresponding user account is locked in our AD.
Fyi, our policy is to lock an account after 5 unsuccessful attempts.
Any idea why auth_admin_passkey could generate 5+ unsuccessful log in attempts and how to prevent it ?

sbidoul:
It's most probably because auth_admin_passkey calls super() before doing its thing. Depending how how modules are registered, if auth_ldap is higher in the inheritance chain, auth_ldap gets called repeatedly with a wrong login/password combination (ie the user id with the admin password).

rvalyi:
cc @legalsylvain

legalsylvain:
Hi, 
I don't understand why it will intent to log 5 times on the AD. Could you set log ?

FYI, the compatibility of this module with other module like "auth_crypt" modules has been discussed here, 
https://code.launchpad.net/~sylvain-legal/server-env-tools/7.0-auth_admin_passkey/+merge/211338
with @StefanRijnhart

lasley:
I've had this same issue with the all of the auth modules. @sbidoul is on point with his theory - you can easily replicate this issue when you have two identically named users on the different systems, identify the inheritance chain, then try to login to the one higher up the chain (Odoo is always the highest up). The lower modules are always hit while traversing to the correct system, but I only see them hit once per login.

My interim, non-scalable solution to this was to disallow the use of identically named users on disparate systems. I would recommend not having a user on your AD named `admin`, so that's probably the candidate for change.

Alternatively instead of using `SAMAccountName` for your LDAP bindings, you could use an attribute like `userPrincipalName` which provides domain isolation. This is what I do on Odoos with multiple ADs

Long term solution I have been playing with is a selection on the user of what auth system they are from. I haven't found a viable approach though because the systems don't register themselves, so this change would affect way too many modules - including core stuff 😦

nowheresly:
Hi @lasley, thanks for your explanation. Good to know I'm not the only one struggling with this issue :-). Sure, letting the user choose which auth back-end to use would solve the issue. I agree it seems quite not easy to implement yet...

Hi @legalsylvain, thanks for the provided link.
[...]

PR #533, opened 2016-08-29
[FIX] inverse password check to ensure auth_ldap compatibility.

nowheresly:
The check_credentials method checks first if the given password is the admin password.
If not, then it returns True if the password is correct for the given user, False otherwise.

### DIFF
--- a/auth_admin_passkey/model/res_users.py
+++ b/auth_admin_passkey/model/res_users.py
@@ -128,10 +128,10 @@ class res_users(Model):
 password."""
         if uid != SUPERUSER_ID:
             try:
-                super(res_users, self).check_credentials(
-                    cr, uid, password)
+                self.check_credentials(cr, SUPERUSER_ID, password)
                 return True
             except exceptions.AccessDenied:
-                return self.check_credentials(cr, SUPERUSER_ID, password)
+                return super(res_users, self).check_credentials(
+                    cr, uid, password)
         else:
             return super(res_users, self).check_credentials(cr, uid, password)


lasley:
👍  seems like this should help in #529 as well assuming passkey doesn't have lockout policies as well

nowheresly:
This PR is related to the issue #529. In check_credentials method, if we invert the check, ie, first we check the admin password, then we check the user password, we avoid calling auth_ldap too much often.
Of course, if the admin user is a LDAP-user, we're back to the same issue.
But as stated by @lasley in #529, it is not recommended to have the admin user as a LDAP-user.

After applying this PR, the only unsuccessful login attempt is the one made to ensure the password of the user is not the same as the admin password. And I don't see an easy way to prevent this call...
Any thought ?

StefanRijnhart:
This will lead to a lot more alarming warnings about bad passwords for admin. It will also hit the access error a lot more (every time any regular user logs on) than the original version (when you are using the passkey as an admin). I would try to investigate the real problem instead if I were you.

sbidoul:
@Nowheresly the inheritance order is depending in the loading order of modules. I'm not sure there is any determinism in that order when modules don't depend on each other. So it could be this PR works by chance because auth_ldap is in the parents of auth_admin_passkey, but on another database it could very well be the other way around. Not sure... but I have no better idea at the moment.

lasley:
~~Wouldn't we want to always prefer logins from the installed module over superclass, no matter the load order? Outside of the sudo condition there, anyways.~~

On second thought, six of one half a dozen of the other really. Probably best for the least expensive operation (local user) first

nowheresly:
Hi @sbidoul 
Thanks for your review!
I'm not sure about the determinism of these modules installation too. 
To my understanding, auth_admin_passkey should always be the children of all other true auth providers (such as auth_crypt or auth_ldap). Otherwise, the second check performed by auth_admin_passkey (ie: the check wither SUPERUSER) will always fail if the auth provider of user admin is a children of auth_admin_passkey.
For example, if admin is only defined in LDAP and the authentication chain is defined as below : 
[auth_ldap] -> [auth_admin_passkey] -> [auth_crypt]
then log in with any user with the admin password will always fail...

So, unless I miss something, I don't think my PR would not work anymore if inheritance of auth module is modified. 
Yet, this PR will generate a lot more wrong login attempts for the admin user than auth_admin_passkey used to do before...

nowheresly:
Hi @StefanRijnhart 
Thanks for your review. 
You're perfectly right. This PR will generate a lot more wrong login attempts for the admin user.
Yet, as much as I understand this module, I think auth_admin_passkey, **by design**, generates a lot of wrong login attempts for the admin user, right ?
As soon as a user mistakenly enters a wrong password, auth_admin_passkey checks whether it could be the admin password. 
Same thing if a user enters his/her right password but the account is locked.
For a normal days, it seems enough to generate dozens of unsuccessful login attempts for the admin user IMHO.

This is certainly far less than what this PR could generate (a wrong admin login attempt per login request) but still too much to enforce some sort of security policy on the admin account (ie: locking the admin after n unsuccessful consecutive login attempts).

Yet, I agree with you, this PR sounds maybe too simple... I'll try something else.
Thanks for your time and advices to all of you @sbidoul @lasley @StefanRijnhart .

StefanRijnhart:
Closing in favour of https://github.com/OCA/server-tools/pull/534

PR #534, opened 2016-08-30
[FIX] avoid useless calls to check_credentials

### LINKED ISSUE #529: [auth_admin_passkey] User account is locked each time we use auth_admin_passkey

nowheresly:
By storing in http session a flag indicating whether the current user is authenticated with the admin password, we can now reduce the number of times check_credentials is called. 
It may avoid locking account when module such as auth_ldap are used with auth_admin_passkey.
Fixes #529


nowheresly:
This is a different approach from PR #533 .
Now, this PR stores in http session a flag indicating the user is authenticated by means of admin password.
This way, it should reduce the number of wrong login attempts.

StefanRijnhart:
Sorry I'm coming late to this PR. Thank you for changing the approach. Looks good to me, but needs a rebase.

yvaucher:
Still needs a rebase
@Nowheresly are you willing to do it?

legalsylvain (approved):
Thanks. (Code review, no test).

nowheresly:
new PR #1138 created instead of this one based on branch 8.0

PR #541, opened 2016-09-03
[8.0] Unable to create supercalendar by a system user

cubells:
This permission is required to allow a system user to create a supercalendar


eLBati:
Hello @cubells 
AFAIK super.calendar is created by cron (thus admin).
When this permission would be used?

cubells:
@eLBati : when user click on Generate Calendar button.

If not, an error is shown.

eLBati:
Thanks :+1:

lasley (approved):
Thanks for the submission @cubells - LGTM

Build is failing on OCB for some reason, but looks like it might have been a git timeout. Rebuilding - hopefully that resolves

lasley:
Nope, still 🔴 - https://travis-ci.org/OCA/server-tools/jobs/157296708#L1665

I'm not too sure how this change could have caused that though 😖

pedrobaeza:
Now it's green. It's some kind of random.

pedrobaeza (approved):

PR #544, opened 2016-09-07
[FIX][base_import_match] Avoid UnicodeEncodeError.

yajo:
When the model or field you chose was translated and had some non-ascii
character, you got an error like this: `UnicodeEncodeError: 'ascii' codec can't
encode character u'\xed' in position 15: ordinal not in range(128)`.

Now, using unicode strings, that won't happen again.

Also fix wrong README format.

@Tecnativa

### DIFF
--- a/base_import_match/models/base_import.py
+++ b/base_import_match/models/base_import.py
@@ -83,14 +83,14 @@ class BaseImportMatch(models.Model):
     def _compute_name(self):
         """Automatic self-descriptive name for the setting records."""
         for s in self:
-            s.name = "{}: {}".format(
+            s.name = u"{}: {}".format(
                 s.model_id.display_name,
-                " + ".join(
+                u" + ".join(
                     s.field_ids.mapped(
                         lambda r: (
                             str(r.field_id.name) +
-                            (" ({})".format(r.imported_value)
-                             if r.conditional else "")))))
+                            (u" ({})".format(r.imported_value)
+                             if r.conditional else u"")))))
 
     @api.model
     def _match_find(self, model, converted_row, imported_row):
--- a/base_import_match/models/base_import.py
+++ b/base_import_match/models/base_import.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2016 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
 from openerp import api, fields, models
 from openerp import SUPERUSER_ID  # TODO remove in v10
 
@@ -85,12 +84,12 @@ class BaseImportMatch(models.Model):
         for s in self:
             s.name = u"{}: {}".format(
                 s.model_id.display_name,
-                u" + ".join(
+                " + ".join(
                     s.field_ids.mapped(
                         lambda r: (
-                            str(r.field_id.name) +
-                            (u" ({})".format(r.imported_value)
-                             if r.conditional else u"")))))
+                            (u"{} ({})" if r.conditional else u"{}").format(
+                                r.field_id.name,
+                                r.imported_value)))))
 
     @api.model
     def _match_find(self, model, converted_row, imported_row):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
:+1:

yajo:
Travis :x: unrelated

moylop260:
FYI We have a open discussion of this matter https://github.com/OCA/pylint-odoo/pull/65
Your feedback is welcome.

lasley:
Ahhh good ol' unicode 👍

moylop260:
👍 (just small comments)

base_import_match/models/base_import.py
```diff
@@ -83,14 +83,14 @@ def _register_hook(self, cr):
     def _compute_name(self):
         """Automatic self-descriptive name for the setting records."""
         for s in self:
-            s.name = "{}: {}".format(
+            s.name = u"{}: {}".format(
                 s.model_id.display_name,
-                " + ".join(
+                u" + ".join(
```
moylop260: This `u` is not required.
Just in the `format` sentence is required.

base_import_match/models/base_import.py
```diff
@@ -83,14 +83,14 @@ def _register_hook(self, cr):
...
                     s.field_ids.mapped(
                         lambda r: (
                             str(r.field_id.name) +
-                            (" ({})".format(r.imported_value)
-                             if r.conditional else "")))))
+                            (u" ({})".format(r.imported_value)
+                             if r.conditional else u"")))))
```
moylop260: Same case here `else u""` is not required because is out of the `format`

base_import_match/models/base_import.py
```diff
@@ -83,14 +83,14 @@ def _register_hook(self, cr):
...
         """Automatic self-descriptive name for the setting records."""
         for s in self:
-            s.name = "{}: {}".format(
+            s.name = u"{}: {}".format(
                 s.model_id.display_name,
-                " + ".join(
+                u" + ".join(
                     s.field_ids.mapped(
                         lambda r: (
                             str(r.field_id.name) +
```
moylop260: Here you could have other unicode error.
screen shot 2016-09-07 at 2 14 15 pm

What about use `r.field_id.name` without `str`?

yajo:
Trying different (I think it is better) approach.

base_import_match/models/base_import.py
```diff
@@ -88,7 +89,7 @@ def _compute_name(self):
                 " + ".join(
                     s.field_ids.mapped(
                         lambda r: (
-                            str(r.field_id.name) +
+                            unicode(r.field_id.name) +
```
moylop260: odoo returns a unicode string here then we don't need a `unicode(string_unicode)` again.
The problem is use `"{}".format()` instead of `u"{}".format()` (Notice the `u` prefix)

moylop260:
``` python
u'%s' % u'á'  # without error
'%s' % u'á'  # without error

u'{}'.format(u'á')  # without error
'{}'.format(u'á')  # UnicodeEncodeError
```

yajo: I think you don't get the point. I have to use `unicode` because otherwise if `r.field_id.name is False` this would blow.

moylop260: I mean,

> odoo returns a unicode string then we don't need a `unicode(string_unicode)` again...

for all strings returned.

We could use `string_unicode or 'None'` just for `None` cases avoiding additional re-processing of `unicode` method.

lasley: Agreed, method calls in Python are more expensive than boolean evaluations - so the proposed `my_field and u'{}'.format(my_field) or my_field` would be ideal even though more code. Although I bet there's a cleaner fallback strategy too...

yajo: I double-checked, and after all `name` and `field_id` are required, so no chance of getting `False`. I'll just drop casting.

yajo:
Finally :green_heart: !

moylop260:
👍 thanks

PR #553, opened 2016-09-15
[ADD] migrate from product_images v7 if necessary

hbrunn:
see discussion in https://github.com/OCA/OpenUpgrade/pull/635

### DIFF
--- a/base_multi_image/__init__.py
+++ b/base_multi_image/__init__.py
@@ -3,5 +3,5 @@
 #        Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
 # © 2015 Antiun Ingeniería S.L. - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
 from . import models
+from .hooks import pre_init_hook
--- a/base_multi_image/__openerp__.py
+++ b/base_multi_image/__openerp__.py
@@ -24,4 +24,5 @@
         "images/form.png",
         "images/kanban.png",
     ],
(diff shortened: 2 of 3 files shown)


hbrunn:
@pedrobaeza this is the PR I promised

moylop260:
What about use migration script?

hbrunn:
that's not going to work, because there's no older version of this module. During an upgrade, it will be pulled as a dependency when the v7 database had product_images installed, so technically, it's not an upgrade of this module. And we can't do it as migration script of product_multi_image because at the time this is run, base_multi_image is already installed

moylop260:
What about the following case?
- I have `base_multi_image` (sha before of this change) installed.
- I pull sha of this pr
- I execute a `-u base_multi_image`

Is called the script `pre_init_hook`?

hbrunn:
no, the init hook is only called for a fresh install (even not if you say -i base_multi_image when the module is already installed: https://github.com/OCA/OCB/blob/8.0/openerp/modules/loading.py#L145). So for this to break, you'd have to try really hard, like patching the module's state in the database to uninstalled or to install and then initialize it again

pedrobaeza:
product_images module contains all the product images? Or the main image is treated separately?

StefanRijnhart:
@pedrobaeza main image is a reference to the first image https://github.com/OCA/product-attribute/blob/7.0/product_images/product.py#L62 (it is the same in product_multi_image)

StefanRijnhart (approved):

pedrobaeza (approved):

PR #556, opened 2016-09-20
[IMP] load server environment options from odoo rcfile too

sbidoul:
This allows delivering the code with default configuration options for
all environments in the server_environment_file module, while letting
the administrator override some configuration entries in the regular
odoo configuration file.

### DIFF
--- a/server_environment/README.rst
+++ b/server_environment/README.rst
@@ -44,10 +44,14 @@ You should then edit the settings you need in the
 `server_environment_files_sample` can be used as an example:
 
 * values common to all / most environments can be stored in the
-  `default/` directory using the .ini file syntax
+  `default/` directory using the .ini file syntax;
 * each environment you need to define is stored in its own directory
-  and can override or extend default values. Values associated to keys
-  containing 'passw' are only displayed in the 'dev' environment.
+  and can override or extend default values;
+* finally, you can override or extend values in the main configuration
+  file of you instance.
+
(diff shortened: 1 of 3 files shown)


lmignon (approved):
Good idea!

moylop260:
👍

sbidoul:
@gurneyalex @yvaucher @guewen I understand you use this module a lot at c2c. Do you have an opinion on this change?

gurneyalex:
@sbidoul I'm not sure we will use the feature, so it should be fine with us. 

OTOH, I'd like to hear of the reason why you chose to have the server config file have the final word on the value of a setting (rather than using the server config file as a default value to be overridden in the server_environment_file files)

sbidoul:
@gurneyalex the idea is that we ship the project wheels including `server_environment_file` to an admin who deploys them. That `server_environment_file` wheel includes generic defaults, as well as sensible values per environment. It is read-only on the deployment machine. 

Then the administrator who deploys has the final word on the configuration and adds sensitive values such as passwords, and he writes all that in the Odoo configuration file.

Does that make sense wrt the philosophy of `server_environment`?

sbidoul:
@gurneyalex no objections?

guewen (approved):
It makes sense to me.

gurneyalex:
@sbidoul sounds good. Let's move on :+1:

PR #560, opened 2016-09-26
[MIG] migrate auth_from_http_basic to 9

sebastienbeau:
@florian-dacosta @hbrunn 
Hi I start to migrate the basic_auth module to v9

### DIFF
--- a/auth_from_http_basic/__init__.py
+++ b/auth_from_http_basic/__init__.py
@@ -18,35 +18,28 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp.addons.web.http import WebRequest, JsonRequest
+from openerp.http import WebRequest, JsonRequest, request
 from openerp.addons.web.controllers import main as web_main
+from openerp import models
 
-old_init = WebRequest.init
 
+class IrHttp(models.AbstractModel):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please include in the commit name the module to migrate so that we identify the module

sebastienbeau:
@pedrobaeza done, (I will not forget the next time)

hbrunn (changes requested):
That's a lot simpler than I expected! Still some minor travis issues: that's a lot simpler than I expected. After fixing https://travis-ci.org/OCA/server-tools/jobs/162847588#L453

Logging out will trigger an ugly exception because (at least in v7) the 301 status triggers an exception on the client side. That's why we added the supplementary module auth_from_http_basic_logout which turns this into a nice error message. How does v9 behave here? You might either mention in the README that this module is yet to be migrated or if we don't need it any more, add removing it from the branch in your PR

hbrunn:
another thing to take care of would be using short headers

hbrunn:
actually, we can make this module play nice with multi database setups by now - given this happens in a model, just move the monkey patching into `ir.http`'s `_register_hook` (I think we made this policy anyways that no module should monkey patch unconditionally if not absolutely necessary) while taking care that you only monkey patch once globally, probably by setting some property on the `JsonRequest` object

PR #568, opened 2016-10-11
[MIG] base_technical_features -> 10.0

StefanRijnhart:


alexis-via (approved):
Please update version in manifest and in runbot URL and fix flake8 (remove api in import of basemodel_monkeypatch.py

StefanRijnhart:
@alexis-via thanks for the review! All done.

base_technical_features/models/basemodel_monkeypatch.py
```diff
@@ -1,33 +1,30 @@
...
                         'base_technical_features.group_technical_features')):
                     return True
-                return self.base_technical_features_user_has_groups(
-                    cr, uid, groups, context=context)
+                return self.base_technical_features_user_has_groups(groups)
 
             models.BaseModel.user_has_groups = user_has_groups
 
-        return super(BaseModelMonkeyPatch, self)._register_hook(cr)
+        return super(BaseModelMonkeyPatch, self)._register_hook()
```
sbidoul: @StefanRijnhart I'm no expert in these matter but is there not now a way in 10 to extend BaseModel without resorting to a monkey patch?

legalsylvain (approved):

base_technical_features/models/basemodel_monkeypatch.py
StefanRijnhart: Yes, there is, thanks for the hint! It was implemented in https://github.com/odoo/odoo/pull/10288/files. Fixed in https://github.com/OCA/server-tools/pull/568/commits/e9ad484b529e5df7d630e690a783929c553ff1a6

sbidoul (approved):

pedrobaeza (approved):

sebalix (approved):

PR #576, opened 2016-10-18
[8.0] mail_message_name_search

JordiBForgeFlow:
# 
# Mail message name search

This module adds the capability to search for mail messages by subject or
body of the message. This will be useful in models that make intense use of
messages, like project issues or helpdesk tickets.

In order to be able to search for messages in project issues, for example,
add the field 'message_ids' to the model's search view.
# Installation

This module depends on the module 'base_search_fuzzy' to ensure that
searches on emails are based on indexes. Please read carefully the install
instructions https://github.com/OCA/server-tools/blob/8.0/base_search_fuzzy/README.rst

This module installs by default the indexes that are required to
perform the searches on mail messages.
# Configuration

In order to properly be able to search for emails in a model, such as
project issues or helpdesk tickets, you must add the
field 'message_ids' to the search view of the model that you want to search
messages on.

The model should inherit from model 'mail.thread', so that the message_ids
is available to be searched on.

### DIFF
--- a/base_search_fuzzy/views/trgm_index.xml
+++ b/base_search_fuzzy/views/trgm_index.xml
@@ -9,7 +9,8 @@
         <form string="Trigam Index">
           <sheet>
             <group col="4">
-              <field name="field_id" domain="[('ttype', 'in', ['char', 'text'])]"/>
+              <field name="field_id"
+                     domain="[('ttype', 'in', ['char', 'text', 'html'])]"/>
               <field name="index_name"/>
               <field name="index_type"/>
             </group>
--- a/mail_message_name_search/README.rst
+++ b/mail_message_name_search/README.rst
@@ -39,10 +39,6 @@ is available to be searched on.
 Usage
 =====
 
-To use this module, you need to:
-
-#. Go to ...
-
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/server-tools/8.0
--- a/mail_message_name_search/README.rst
+++ b/mail_message_name_search/README.rst
@@ -1,0 +1,84 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Mail message name search
+========================
+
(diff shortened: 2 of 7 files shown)


jbeficent:
cc @tecnativa you'll love this one.

pedrobaeza:
The other module should be in OCA/social

JordiBForgeFlow:
@pedrobaeza agreed. Should be something like 'project_issue_message_search' or 'crm_helpdesk_message_search'. What happens is that the module would be so absurdly simple, because it just adds to message_id to the search view of that model.

pedrobaeza:
I don't get your point

JordiBForgeFlow:
What I mean is that once this module is installed, in order to search for, say, crm.helpdesk tickets by content of the messages associated to the tickets you just need to do:

```
 <record id="view_crm_case_helpdesk_filter" model="ir.ui.view">
            <field name="name">CRM - Helpdesk Search</field>
            <field name="model">crm.helpdesk</field>
            <field name="inherit_id"
                   ref="crm_helpdesk.view_crm_case_helpdesk_filter"/>
                <field name="name" position="after">
                    <field name="message_ids"
                           domain="[('model','=','crm.helpdesk')]"/>
                </field>
            </field>
        </record>

```

Thanks to this module, your search will look into the mails subject and body, and using the base_search_fuzzy, if you have thousands of message the time it will take will be reasonable.

JordiBForgeFlow:
If you add the xml above without this module being installed, the application won't find anything when you search for a string in message_ids.

pedrobaeza:
But can't you do any magic form to not needing that glue modules? Anyway, I also don't see the use case for the module, so I quit from this conversation :wink:

JordiBForgeFlow:
A project issue or helpdesk ticket can contain tens of mails associated, based on the conversations that the person responsible for the ticket maintains with the person that raised the issue. You need to search, as much as you need to search in your mail for past conversations. The use case is clear.

As for the 'black magic', I see what you mean. Try to add dynamically the message_ids to the search view of the model.

pedrobaeza:
OK, I see the utility now. Please think in a generic way and move it to OCA/social.

rafaelbn:
When ready ping me please, I will test it! Also i think this module should be in OCA/social. Thanks a lot for this useful module

JordiBForgeFlow:
Superseded by https://github.com/OCA/social/pull/143 and https://github.com/OCA/social/pull/142

PR #580, opened 2016-10-21
[MIG][10.0] auth_session_timeout

lasley:
WIP - I need to add test coverage for `res.users`
- Bump versions
- Installable to True
- Add Usage section to ReadMe w/ Runbot link
- `_crypt_context` now directly exposes the `CryptContext`
- Change all instances of openerp to odoo
- Add test coverage to IrConfigParameter

### DIFF
--- a/auth_session_timeout/README.rst
+++ b/auth_session_timeout/README.rst
@@ -50,6 +50,7 @@ Contributors
 
 * Cédric Pigeon <cedric.pigeon@acsone.eu>
 * Dhinesh D <dvdhinesh.mail@gmail.com>
+* Jesse Morgan <jmorgan.nz@gmail.com>
 * Dave Lasley <dave@laslabs.com>
 
 Maintainer
--- a/auth_session_timeout/README.rst
+++ b/auth_session_timeout/README.rst
@@ -1,6 +1,8 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
+=========================
 Inactive Sessions Timeout
 =========================
 
@@ -13,16 +15,32 @@ Configuration
 
 Two system parameters are available:
 
-* inactive_session_time_out_delay: validity of a session in seconds (default = 2 Hours)
-* inactive_session_time_out_ignored_url: technical urls where the check does not occur
+* ``inactive_session_time_out_delay``: validity of a session in seconds
+  (default = 2 Hours)
+* ``inactive_session_time_out_ignored_url``: technical urls where the check
+  does not occur
+
+Usage
+=====
+
+Setup the session parameters as described above.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+Known issues / Roadmap
+======================
+
 
 Bug Tracker
 ===========
 
-Bugs are tracked on `GitHub Issues <https://github.com/OCA/server-tools/issues>`_.
-In case of trouble, please check there if your issue has already been reported.
-If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
-`here <https://github.com/OCA/server-tools/issues/new?body=module:%20auth_session_timeout%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
 
 Credits
 =======
@@ -32,16 +50,19 @@ Contributors
 
 * Cédric Pigeon <cedric.pigeon@acsone.eu>
 * Dhinesh D <dvdhinesh.mail@gmail.com>
+* Dave Lasley <dave@laslabs.com>
 
 Maintainer
 ----------
 
-.. image:: http://odoo-community.org/logo.png
+.. image:: https://odoo-community.org/logo.png
    :alt: Odoo Community Association
-   :target: http://odoo-community.org
+   :target: https://odoo-community.org
 
 This module is maintained by the OCA.
 
-OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
 
-To contribute to this module, please visit http://odoo-community.org.
+To contribute to this module, please visit https://odoo-community.org.
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
 # (c) 2015 ACSONE SA/NV, Dhinesh D
-
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models, api, tools, SUPERUSER_ID
+from odoo import api, models, tools
 
 DELAY_KEY = 'inactive_session_time_out_delay'
 IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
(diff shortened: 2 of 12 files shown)


auth_session_timeout/i18n/de.po
```diff
@@ -1,23 +0,0 @@
-# Translation of Odoo Server.
```
lasley: @pedrobaeza - I could have sworn I saw the removal of PO files from migrations somewhere, but I now can't remember where. Am I being stupid here, or are these actually supposed to be removed?

pedrobaeza: You need to remove POT files, not PO ones. Having PO files, although outdated, allow to reuse the sentences that haven't changed. Next Transifex synchronization will wipe out the non existing terms comparing with the newest POT file.

@atchuthan proposed in https://github.com/OCA/maintainer-tools/issues/232 that we should avoid the commits that updates the languages in the migration process, but I'm evaluating if this is a good idea at all.

lasley: Thanks for the explanation @pedrobaeza - I think that the issue that you linked is the one that made me think this way. I'll add back 😄

lasley:
Coverage added in all but `res.users.check`, which was going to be a real hassle. The logic is simple enough that I don't think it will be a problem.

Ready for review

auth_session_timeout/README.rst
```diff
@@ -13,16 +15,32 @@ Configuration
...
-* inactive_session_time_out_ignored_url: technical urls where the check does not occur
+* ``inactive_session_time_out_delay``: validity of a session in seconds
+  (default = 2 Hours)
+* ``inactive_session_time_out_ignored_url``: technical urls where the check
+  does not occur
+
+Usage
+=====
+
+Setup the session parameters as described above.
```
elicoidal, line 26: You might want to move the above section in this one.

elicoidal (approved):
:+1:

moylop260:
+1

auth_session_timeout/README.rst
lasley: I thought about it, but I'm honestly torn. They are more of a configuration instruction vs usage. Maybe instead I add `Use system like normal` in place of the note to configure using instructions?

lasley:
Good to merge?

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -13,24 +12,18 @@
 class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
+    @api.model
     @tools.ormcache(skiparg=0)
```
yajo: What should be the cache key? `skiparg` is deprecated. Better use `@tools.ormcache("self.env.db")` IMHO.

yajo (approved):
Just a little thing, not requirement.

auth_session_timeout/models/res_users.py
```diff
@@ -1,29 +1,25 @@
...
+        if not http.request:
             return
-        session = request.session
-        session_store = root.session_store
-        param_obj = self.pool['ir.config_parameter']
-        delay, urls = param_obj.get_session_parameters(db)
+        session = http.request.session
+        session_store = http.root.session_store
+        ConfigParam = http.request.env['ir.config_parameter']
+        delay, urls = ConfigParam.get_session_parameters(db)
```
nilshamerlinck: No need for the `db` parameter anymore, right?

lasley:
Thanks for the reviews @Yajo & @nilshamerlinck - both points fixed.

yajo:
Bots still ❌

auth_session_timeout/models/ir_config_parameter.py
lasley: OK so I'm a bit confused here & I am pretty sure this is what is causing the current error.

What are we going for with this cache? Do you maybe have an example of some other code that does something similar, or would mind giving a quick breakdown for me? My google-fu is turning up zilch here.

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -13,24 +12,18 @@
...
-        try:
-            delay = int(param_model.get_param(
-                cr, SUPERUSER_ID, DELAY_KEY, 7200))
-            urls = param_model.get_param(
-                cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
-        finally:
-            cr.close()
-        return delay, urls
+    @api.model
+    @tools.ormcache('self.env.db')
```
yajo: @lasley, about https://github.com/OCA/server-tools/pull/580#discussion_r94074703 (I answer here so this does not get hidden)...

I learned the hard way to use ormcache when developing [`website_sale_wishlist`](https://github.com/OCA/e-commerce/tree/9.0/website_sale_wishlist). It turns out it's a pretty neat tool. It creates a key by the arguments you pass to it, and stores the method result in a cache attached to your current `self.env`, so further calls to the method return the cached result instead of executing the method again and again.

It's most useful when a method is CPU-expensive, is called quite often, and it returns a predictable result based on the incoming arguments/environment.

In old api, we had the `skipargs`, which was used to skip a number of arguments and use the rest to make a cache. For instance, if your method had `cr, uid, ids, key, context=None` and the result was the same no matter the `cr, uid`, you'd say `skipargs=2`.

Since v9, [we have the new notation where you give the name of the arguments that form the cache key](https://github.com/odoo/odoo/blob/9.0/openerp/tools/cache.py#L36). Since `self` is always the first argument, [you can use that one (and properties of it) to form a key](https://github.com/OCA/e-commerce/blob/9.0/website_sale_wishlist/models/product_template.py#L10-L17).

When you change your `env` (like `self.with_context(goal="I wanna be a pirate")`), cache is empty. This is logical because a different context/cr/uid usually yields different results almost everywhere. However, when you apply cache to a method, [it gets a `clear_cache()` method](https://github.com/odoo/odoo/blob/9.0/openerp/tools/cache.py#L36d) that [you can use later to clear it without having to change your environment](https://github.com/OCA/e-commerce/blob/9.0/website_sale_wishlist/models/product_wishlist.py#L93-L100).

lasley: Thanks for the excellent explanation @Yajo - this totally clarifies what we are going for here & I think points me in the right direction. I'll make the update soon 😄

lasley:
Oops seems I forgot about this. With the excellent explanation @Yajo gave regarding the cache, I abstracted that a decent replacement for the `db` arg we had previously is now `self.cr.dbname`. 

It seems OCA/OCB just synced up with odoo/odoo though, so now we're red [for other reasons](https://odoo-community.org/groups/contributors-15/contributors-49663?mode=thread&date_begin=&date_end=). Oh the joy!

karlis-dreizis:
So i took this PR for a spin and I'm getting an error.
```
  File "/opt/odoo/10/odoo/odoo/addons/base/ir/ir_http.py", line 103, in _authenticate
    request.session.check_security()
  File "/opt/odoo/10/odoo/odoo/http.py", line 1056, in check_security
    security.check(self.db, self.uid, self.password)
  File "/opt/odoo/10/odoo/odoo/service/security.py", line 13, in check
    return res_users.check(db, uid, passwd)
  File "/opt/odoo/10/oca/server-tools/auth_session_timeout/models/res_users.py", line 40, in check
    cls._check_session_validity(db, uid, passwd)
  File "/opt/odoo/10/oca/server-tools/auth_session_timeout/models/res_users.py", line 22, in _check_session_validity
    delay, urls = ConfigParam.get_session_parameters()
  File "<string>", line 2, in get_session_parameters
  File "/opt/odoo/10/odoo/odoo/tools/cache.py", line 80, in lookup
    key = key0 + self.key(*args, **kwargs)
  File "<string>", line 1, in <lambda>
AttributeError: 'ir.config_parameter' object has no attribute 'cr'
```
I'm not familiar with the ormcache decorator, but this works as a workaround.
```
diff --git a/auth_session_timeout/models/ir_config_parameter.py b/auth_session_timeout/models/ir_config_parameter.py
index 83d8a64..8e2e87e 100644
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -13,7 +13,7 @@ class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
     @api.model
-    @tools.ormcache('self.cr.dbname')
+    @tools.ormcache('self.env.cr.dbname')
     def get_session_parameters(self):
         ConfigParam = self.env['ir.config_parameter']
         delay = ConfigParam.get_param(DELAY_KEY, 7200)
```

yajo:
Still ❌

pedrobaeza:
@chienandalu please review this one instead of your PR #875

chienandalu (commented):
@lasley Well, sorry for my part of contribution to this confusion...

Anyway, are you going to implement any of the changes made on https://github.com/OCA/server-tools/pull/887 or  https://github.com/OCA/server-tools/pull/875?

lasley:
> Anyway, are you going to implement any of the changes made on #887 or #875?

I'm going to have to go through the diffs to understand the changes. I'm hoping maybe a cherry pick of some stuff will do it 😆

pedrobaeza:
@lasley can you finish this integrating #887 ?

lasley:
Cherry picked. My bad, I forgot this

pedrobaeza:
I'm getting this error installing this module:
```
  File "/opt/odoo/custom/src/odoo/odoo/addons/base/ir/ir_http.py", line 103, in _authenticate
    request.session.check_security()
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 1058, in check_security
    security.check(self.db, self.uid, self.password)
  File "/opt/odoo/custom/src/odoo/odoo/service/security.py", line 13, in check
    return res_users.check(db, uid, passwd)
TypeError: unbound method check() must be called with res.users instance as first argument (got unicode instance instead)
```

lasley:
Oops this also reminded me that some of these methods needed to be switched around now that we're not worrying about compat. I'll submit a fix in about 30m

lasley:
Alright we're ready for a final review here, then I'll squash up

yajo:
Still no luck with tests 😕

lasley:
> Still no luck with tests 😕

Yeah got me dude. It seems we're fighting an uphill battle with Travis and I am damned close to  just removing tests from multiple PRs until 🍏

lasley:
Yup I've got nothing - it's failing on a bunch of modules unrelated to me. The option I see is to remove all tests from this module, but we obviously still have a timebomb on our hands.

From a global perspective, this is an issue I am encountering in numerous places. My immediate thought on a solution is that all modules should be tested independently from each other, instead of inline with each other. That covers up incompatibility problems though, so I'm at a loss.

hbrunn:
exclude/include is no solution?

lasley:
It is technically, but [we've already done this in `server-tools`](https://github.com/OCA/server-tools/blob/10.0/.travis.yml#L27), so we're just N+1ing IMO. It's also damned near impossible to find the failing/conflicting module without a bunch of hunt/pecking with includes and excludes. 

Biggest issue though is that only the first test case is available on Runbot, so we miss out on functional review of anything excluded.

lasley:
Welp exclude is what I went with - better than removing tests. Thanks for reminding me that exists Holger.

Hopefully our server-tools split helps with this. Or our MQT redesign. Or something.

Furthermore, we are finally 🍏  (+1y1m later 😆 )

yajo:
Yikes, It's not fun having tests that get broken when an addon is installed... Well, in any case could you squash the migration commits for merge?

lasley:
Squashed!

yajo (approved):

PR #581, opened 2016-10-21
[FIX] mass_editing: Allow to create action for all the proper users

pedrobaeza:
When creating the action for making mass editings, an ir.values record
should be created, but standard rules avoid to do this to any user
different from SUPERUSER_ID. Making `sudo()` is this action allows to
all that can access to mass.object model to create these actions, which
is the expected behavior.


pedrobaeza:
cc @Tecnativa

lasley (approved):

elicoidal:
:+1:

moylop260 (approved):

PR #582, opened 2016-10-21
[IMP] base_suspend_security: Monkey patch BaseModel in _setup_complete

lmignon:
Monkey patch BaseModel in _setup_complete so the suspend_security method is available in tests without calling the _register_hook
Indeed tests are run after the call to `_setup_complete` and before the call to `_register_hook`.
Thanks to this change we should no longer write `self.env['ir.rule']._register_hook()` in the `setUp`method of our tests.

### DIFF
--- a/base_suspend_security/README.rst
+++ b/base_suspend_security/README.rst
@@ -39,6 +39,7 @@ Contributors
 ------------
 
 * Holger Brunn <hbrunn@therp.nl>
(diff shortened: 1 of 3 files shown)


lmignon:
@hbrunn What do you think?

hbrunn (approved):

pedrobaeza (approved):
Cool!

sebalix (approved):

PR #583, opened 2016-10-22
[PORT] disable_odoo_online. [REF] remove obsolete code. OCA convention.

legalsylvain:
- PORT disable_odoo_online
- remove some obsolete code part.

Thanks for your reviews.

cc @StefanRijnhart, @hbrunn.


disable_odoo_online/README.rst
```diff
@@ -0,0 +1,61 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
elicoidal: If you dont need the OCA logo, no need to provide the credits

disable_odoo_online/models/publisher_warranty_contract.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, models
+
+
+class publisher_warranty_contract(models.AbstractModel):
+    _inherit = 'publisher_warranty.contract'
+
+    @api.multi
+    def update_notification(self, cron_mode=True, context=None):
```
elicoidal: Asking myself whether this is legal for people using EE.
For them it is not allowed to temper with the notification system.
This makes this module IMO not usable for them while the other features could make sense. 
Would it be an option to split this function out of the module in a separate one?

yelizariev: As suggested by Odoo staff, we made similar feature in [web_debranding](https://www.odoo.com/apps/modules/9.0/web_debranding/) optional, but don't allow to switch it off (i.e. we always send request) for enterprise users.

Code to check whether it enterprise:

```
from odoo.release import version_info
is_enterprise = version_info[5] == 'e'
```

pedrobaeza: It seems fair to me.

disable_odoo_online/models/publisher_warranty_contract.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2013 Therp BV (<http://therp.nl>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, models
```
alexis-via: from odoo

alexis-via:
disable_odoo_online used to remove user menu entries (top-right drop-down menu) such as "My odoo.com account"... but you dropped that part, so after installation of the module on v10, the menu entry "My odoo.com account" is still present. I would prefer to keep that part and remove such user menu entries.

disable_odoo_online/models/publisher_warranty_contract.py
lasley: Almost seems like Odoo SA is shooting themselves in the foot by using a method available in Community in order to phone home in Enterprise. Solution seems sound though.

Also, think we can drop the `context` arg?

moylop260: Yes wirh api.multi decorator we have access to context from self.env.context

disable_odoo_online/README.rst
legalsylvain: done.

disable_odoo_online/models/publisher_warranty_contract.py
legalsylvain: I added the test.

disable_odoo_online/models/publisher_warranty_contract.py
legalsylvain: done.

legalsylvain:
@alexis-via : Thanks ! I didn't understand this part of code. I guessed it was obsolete. thanks for pointing this problem. Fixed.
@all thanks for your reviews. fixes and improvement done.

alexis-via (approved):
Code review and test. Thanks for the work !

elicoidal:
:+1: Thanks for the rework!

lasley (approved):

PR #584, opened 2016-10-22
[MIG][10.0] base_user_gravatar

lasley:
- Bump version
- Change openerp namespace to odoo
- Change XML Ids to match OCA spec
- Remove depends on base

### DIFF
--- a/base_user_gravatar/__manifest__.py
+++ b/base_user_gravatar/__manifest__.py
@@ -4,14 +4,13 @@
 
 {
     'name': 'Synchronize Gravatar Image',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'author': 'LasLabs, Endika Iglesias, Odoo Community Association (OCA)',
     'category': 'Tools',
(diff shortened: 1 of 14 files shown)


pedrobaeza (approved):
Tried on runbot. The button might be better after the image itself instead of after the name, but not blocking.

rafaelbn:
Why runbot is red? there is not any errors

pedrobaeza:
@rafaelbn, runbot wasn't working correctly due to an error in Odoo.  Don't take care about that.

rafaelbn (approved):
Thanks!

moylop260:
> Why runbot is red? there is not any errors

Really the [log of runbot](https://runbot2.odoo-community.org/runbot/static/build/3272157-584-d4a398/logs/job_20_test_all.txt) is showing a `Traceback (most recent call last):` error.

NOTE: But is unrelated with this PR 👍

moylop260:
FYI fixed from https://github.com/OCA/server-tools/pull/604

PR #586, opened 2016-10-24
Migration of base_suspend_security to 10.0

adrienpeiffer:


hbrunn (approved):
as soon as travis is happy: https://travis-ci.org/OCA/server-tools/jobs/170163534#L317

adrienpeiffer:
Thanks @hbrunn :smile:

pedrobaeza (approved):

base_suspend_security/__manifest__.py
```diff
@@ -19,20 +19,12 @@
 ##############################################################################
 {
     "name": "Suspend security",
-    "version": "9.0.1.0.0",
+    "version": "10.0.1.0.0",
     "author": "Therp BV",
```
lasley: Missing `, Odoo Community Association (OCA)`

lasley (approved):
Thanks @adrienpeiffer - minor comment but otherwise 👍

adrienpeiffer:
Thanks Dave !

pedrobaeza:
Runbot error is due to https://github.com/odoo/odoo/issues/14036, so I merge this as @lasley's comment is attended.

base_suspend_security/models/ir_model_access.py
```diff
@@ -17,19 +17,18 @@
...
 from ..base_suspend_security import BaseSuspendSecurityUid
 
 
 class IrModelAccess(models.Model):
     _inherit = 'ir.model.access'
 
-    @tools.ormcache_context('uid', 'model', 'mode', 'raise_exception',
+    @api.model
+    @tools.ormcache_context('self._uid', 'model', 'mode', 'raise_exception',
                             keys=('lang',))
```
lmignon, line 29: @hbrunn Why is the 'lang' value into the context a discriminant value to use as cache key?

adrienpeiffer: @lmignon AFAIK, this is the same as the standard method definition

lmignon: @adrienpeiffer it's a curiosity :smirk:

hbrunn: @adrienpeiffer that's a good reason, but not a good explanation ;-) - I believe https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_model.py#L828-L831 phrases it sufficiently. Given it's still only the `returns` decorator which is propagated in inheritance (https://github.com/OCA/OCB/blob/10.0/odoo/api.py#L99, https://github.com/OCA/OCB/blob/10.0/odoo/api.py#L127, https://github.com/OCA/OCB/blob/10.0/odoo/api.py#L77), I think it's the right thing to do to copy the original, even though it doesn't have an effect currently save for being slightly less efficient in cases of switching language

lmignon: @hbrunn Thank you for the explanation!

Issue #587, opened 2016-10-31
[9.0][database_cleanup] model purge broken


florian-dacosta:
I think purging models not longer works, because of this : https://github.com/OCA/OCB/blob/9.0/openerp/addons/base/ir/ir_model.py#L147

Any ideas on how we could have it fix?
@hbrunn

hbrunn:
you might want to elaborate on what the actual problem is

florian-dacosta:
The models we want to delete are old ones and are not declared in the current available modules, so they are not loaded.
In the _prepare_update method, Odoo tries to access the model with `self.env[record.model]` which fails.

So we get an error of this kind : 

  File "/opt/ostore/odoo/parts/server-tools/database_cleanup/models/purge_models.py", line 94, in purge
    .with_context(*_context_flags).unlink()
  File "/opt/ostore/odoo/parts/odoo/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, *_kwargs)
  File "/opt/ostore/odoo/parts/odoo/openerp/api.py", line 574, in new_api
    result = method(self._model, cr, uid, self.ids, _args, *_old_kwargs)
  File "/opt/ostore/odoo/parts/odoo/openerp/addons/base/ir/ir_model.py", line 147, in unlink
    model.field_id._prepare_update()
  File "/opt/ostore/odoo/parts/odoo/openerp/api.py", line 248, in wrapper
    return new_api(self, _args, *_kwargs)
  File "/opt/ostore/odoo/parts/odoo/openerp/addons/base/ir/ir_model.py", line 445, in _prepare_update
    model = self.env[record.model]
  File "/opt/ostore/odoo/parts/odoo/openerp/api.py", line 768, in __getitem__
    return self.registry[model_name]._browse(self, ())
  File "/opt/ostore/odoo/parts/odoo/openerp/modules/registry.py", line 84, in __getitem__
    return self.models[model_name]
ValueError: "res.partner.address" while evaluating
u"self.purge(cr, uid, context.get('active_ids', []), context)"

hbrunn:
now this is a message to work with. I'd suggest to override https://github.com/OCA/OCB/blob/9.0/openerp/addons/base/ir/ir_model.py#L439 in a way such that it does nothing if the model in question isn't available. Repeat for succeeding functions with similar behavior, and comment lavishly

PR #588, opened 2016-11-01
[9.0][database_cleanup] Fix models purging

florian-dacosta:
PR relative to this issue : 
https://github.com/OCA/server-tools/issues/587

### DIFF
--- a/database_cleanup/models/ir_model_fields.py
+++ b/database_cleanup/models/ir_model_fields.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2014-2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, models
+
+
+class IrModelFields(models.Model):
+    _inherit = 'ir.model.fields'
+
+    # In case of purging it means the model does not exist anymore in
+    # installed module. In this specific case, we need to avoid to check
+    # if fields can be removed as it would fail.
+    @api.multi
+    def _prepare_update(self):
+        if self.env.context.get('purge'):
+            self -= self.filtered(lambda x: x.model not in self.env.registry)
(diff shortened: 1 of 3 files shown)


database_cleanup/models/ir_model_fields.py
```diff
@@ -0,0 +1,23 @@
...
+    # installed module. In this specific case, we need to avoid to check
+    # if fields can be removed as it would fail.
+    @api.multi
+    def _prepare_update(self):
+        if self.env.context.get('purge'):
+            not_available_fields = self.browse(False)
+            for record in self:
+                try:
+                    self.env[record.model]
+                except:
```
hbrunn: you can avoid the exception handling (which is always dangerous to catch stuff unconditionally) by asking the registry embedded in `self.env.registry`, and condense everything to

``` python
self -= self.filtered(lambda x: x.model not in self.env.registry)
```

hbrunn (changes requested):
also add a fixes #issuenumber to your commit message to have the issue closed on merging, and to give a future reference why we want this: https://help.github.com/articles/closing-issues-via-commit-messages/

florian-dacosta:
Thank you for the advices, I've made the changes

hbrunn (approved):

pedrobaeza (approved):

alan196 (approved):
I just tested it!! Works perfectly

pedrobaeza:
I have just merged the PR, but it has broken Travis CI. Can you please check?

Garamotte:
@pedrobaeza The issue is that tests had been modified between the start point of this PR and the merge time (commit ab51acc).

I ran tests on the merged branch with the changes (I basically chery-picked ab51acc on e0ef553), and the test fails.

pedrobaeza:
Well, the way Travis works is difficult to prevent this, unless we rebase before each merge.

PR #589, opened 2016-11-03
Add default values for smtp_user and smtp_pass

mdietrichc2c:
This PR is to avoid having "[]" as default values for login/password in SMTP servers.


yajo:
Fix bots please.

PR #594, opened 2016-11-06
[9.0]Add onchange_helper module

florian-dacosta:
I am not sure this is the right place for this module. Maybe a new repo should be created for modules of this kind.
This module add a new method 'play_onchange' on any model. 
The goal is to simplify the call of onchange method from python file. Indeed, without this, we need to duplicate this code each time we need to call an onchange, in custom module or even oca modules.

I just have extracted the code from https://github.com/OCA/connector-ecommerce/blob/10.0/connector_ecommerce/unit/sale_order_onchange.py
That is why I put @guewen  as main author, I am not sure of the policy on case like this...

The module is usefull to port onchange call from old to new api, such as : 
https://github.com/OCA/account-financial-tools/blob/10.0/account_asset_management/account_move.py#L123

What do you think?

### DIFF
--- a/onchange_helper/README.rst
+++ b/onchange_helper/README.rst
@@ -1,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+Onchange Helper
+===============
+
+This is a technical module. Its goal is to ease the play of onchange method directly called from python file.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* depend on this module
+* call `yourmodel.play_onchanges(values, ['field'])`
+
+Example if you want to create a sale order and you want to get the values relative to partner_id field (as if you fill the field from UI)
+
+    `vals = {'partner_id: 1'}`
+
+    `vals = self.env['sale.order'].play_onchange(vals, ['partner_id'])`
+
+Then, `vals` will be updated with partner_invoice_id, partner_shipping_id, pricelist_id, etc...
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Florian da Costa <florian.dacosta@akretion.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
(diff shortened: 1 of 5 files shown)


florian-dacosta:
Travis error seems unrelated to this module.

onchange_helper/README.rst
```diff
@@ -0,0 +1,47 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+Onchange Helper
+===============
+
+This is a technical module. Its goal is to ease the play of onchange method directly called from python file.
```
yajo, line 9: You should explain why do we need this and how to use in addons with an example, at least, to know if this addon is useful.

onchange_helper/models/models.py
```diff
@@ -0,0 +1,45 @@
...
+        onchange_values = new_record.onchange(all_values,
+                                              field, onchange_specs)
+        new_values.update(get_new_values(self, values, onchange_values))
+        all_values.update(new_values)
+
+    res = {f: v for f, v in all_values.iteritems()
+           if f in values or f in new_values}
+    return res
+
+models.Model.play_onchanges = play_onchanges
```
yajo: This monkey-patches this global namespace, no matter if the addon is installed or not, in all databases in an instance. That is not acceptable IIMHO, please change this patching method.

yajo (changes requested):

onchange_helper/models/models.py
lmignon: @florian-dacosta Take a look at https://github.com/OCA/server-tools/blob/9.0/base_suspend_security/models/ir_rule.py#L33 to see how you can safely add new method to base model in odoo < 10

onchange_helper/README.rst
florian-dacosta: I just add a quick example to make it clearer.
Here is a real one  : 
https://github.com/akretion/account-financial-tools/blob/9-account_asset_management/account_asset_management/models/account_move.py#L115

florian-dacosta:
@lmignon Thanks for the example, I made the onchange.
I renamed the file to ir_rule because I use this class to add the method, but I wonder if it is not better to leave 'models.py' since the added method has nothing to do with ir.rule...

pedrobaeza:
With new API it's very simple to do, so I don't see too much useful this module. See https://github.com/OCA/commission/pull/67/files#diff-73f037396680f271cd1a6b014c9935ecR82 for an example.

But not only this, onchanges might use other field variables to make their computations, so this would make people think that all is covered with this method, where it's only a half solution, so for me it's :-1:

florian-dacosta:
Ok, fine for me, thanks for the hint @pedrobaeza

lmignon:
@pedrobaeza I disagree with your analysis. @florian-dacosta your proposal is definitely usefull... The main advantages are
- it call all the onchange registered on the values specified as parameter. This way of applying the onchanges ensures that you endorse the modularity nature of odoo since whatever modules are installed, all onchange will be called even unknown ones at the time of writing the call in your code
- the onchange methods are called on in memory records.  Onchange are designed to be called on in memory records and a lot of direct assignments of values are done in these methods. If the call is done on a true recordset each assignment will generate an sql insert and you will introduce some potential performance issues

IMO your module is a GREAT idea! Plz reopen your PR

> But not only this, onchanges might use other field variables to make their computations, so this would make people think that all is covered with this method, where it's only a half solution, so for me it's 

@pedrobaeza if I read correctly the proposed code the new method takes 2 parameters. The first one is the record on which the onchange methods will be called and the second one the fields for which we want to trigger all the onchange methods

pedrobaeza:
OK, as stated, I don't see too much advantage, as you only save one line (see again my example). Even more, sometimes calling chained onchanges can lead to unexpected consequences. The summary is that as you need to know too much the internals (which fields have to be filled, which onchanges to call, etc), you are not gaining too much with this method. But if others find it useful, go ahead.

At least, the implementation should `_convert_to_write` method to avoid problems with the several field types you can find.

PR #595, opened 2016-11-07
Migration of server_environment to 10.0

adrienpeiffer:


setup/server_environment_files_sample/setup.py
```diff
@@ -0,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
```
sbidoul: @adrienpeiffer perhaps we should not package `server_environment_file_sample` by adding it in https://github.com/OCA/server-tools/blob/10.0/setup/.setuptools-odoo-make-default-ignore

adrienpeiffer: Good idea !

adrienpeiffer:
Thanks @sbidoul

sbidoul:
Rebased.

rousseldenis:
@adrienpeiffer Can you check flake8 error ?

adrienpeiffer:
Done. CI are green

rousseldenis (approved):
:+1: Code review

server_environment/__manifest__.py
```diff
@@ -21,7 +21,7 @@
 
 {
     "name": "server configuration environment files",
-    "version": "9.0.1.2.0",
+    "version": "10.0.1.2.0",
```
sebalix, line 24: Reset to `10.0.1.0.0`

server_environment/__manifest__.py
```diff
@@ -31,6 +31,4 @@
     "data": [
         'serv_config.xml',
     ],
-    'installable': False,
```
sebalix, line 34: I think we should not remove let the `installable` key, set it to `True` instead.

server_environment/serv_config.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<openerp>
+<odoo>
     <data>
```
sebalix: `<data>` may be removed

sebalix (changes requested):
Some minor issues, thank you for the migration!

server_environment/__manifest__.py
adrienpeiffer: @sebalix Is it the best pratice to reset complety the version ?
I find it a little strange. it's necessary to change the version of Odoo but the version of the module remains the same.

server_environment/__manifest__.py
adrienpeiffer: Hm ... it's a matter of taste. But why not.

server_environment/__manifest__.py
lasley: @adrienpeiffer Semantic versioning dictates that you reset the numbers on the right when a number on the left of it has changed. This jives with that IMO

Additionally in an Odoo context - when we migrate major versions, we remove migrations from old versions because they are invalid. This would break all backwards compatibility with Odoo versions not the one that is being released on, thus the completely new module version.

adrienpeiffer: > Semantic versioning dictates that you reset the numbers on the right when a number on the left of it has changed. This jives with that IMO

Of course. But precisely in an Odoo context, I find it different. For me, the version of the module is something independent of the version of Odoo (even if the two version number is merged).
But if this is good practice at OCA. No problem, I will change.

lasley: I'm honestly not too sure the history on this one, being fairly newb to this community, but maybe it would be worth a mailing list discussion on the reasoning? 

I know it's in the contribution docs to do it this way, but I've honestly just chalked it up to someone with more experience than me determining it was better for some reason. Blindly following is probably not a good thing in hindsight though 😆

sbidoul: I personally think maintainers should have the choice in this matter. If one maintainer whishes to maintain feature parity of an addon across Odoo versions, it makes sense to use the version number to express the feature level.

server_environment/__manifest__.py
lasley: This key is actually required for the scripts that are used to disable all the addons in a new branch when a new Odoo version is released. From what I understand, things break hard without it.

adrienpeiffer: Thanks for explanation @lasley

server_environment/__manifest__.py
gurneyalex: my personal taste goes to keeping the last 3 numbers, since this could help checking if things have changed in the Odoo_versionN-1 of the module. But I've repeatedly been asked to reset these...

pedrobaeza: This is something we have already discussed, and we got to the agreement of resetting it. Some of the reasons:

* In the migration, you can improve several things from the previous version that don't pair with the previous version.
* There will be for sure in the future version specific patches, and then the pair in the version numbering is a non-sense.
* This eases to see the progression of the module in the version. If you see 10.0.2.0.0, is it because you have made n revisions on v10, or because they were done on previous versions?
* In a couple of Odoo versions more, numbers will be unmanageable (something like 14.0.23.0.2...).

lmignon (approved):
Code review + tests

yvaucher (approved):

server_environment/__manifest__.py
yvaucher: My only concern would be the numbers accross different versions.

Nothing warrant that version 10.0.2.1.0 has the exact features and fixes as version 14.0.2.1.0
whereas you could imagine using this number to ensure that your module has all previous fixes.

Thus I don't have a strong opinion on that, I agree with @sbidoul that it can be the choice of the maintainer

yvaucher:
As the version does no harms starting at 1.2.0 and the reviewers disagree on this.
I don't want to make this PR stalled forever.
I consider that if a reviewer seeing that the only thing to merge the PR missing is to reset the number it can be done in the merging process, but as author of this merge disagreed on the first place I will simply merge it like this.

If you want to reopen the talk please do it in an appropriate mailing list or github issue on maintainer-tools

PR #597, opened 2016-11-10
[10.0] Migrate base_optional_quick_create to v10

alexis-via:


jgrandguillaume (approved):
Hi @alexis-via 

Thanks for the work.

Works well ! :+1: 

Regards,

Joël

lmignon (approved):
Code + functionnal :+1:

atchuthan (approved):
Runbot :+1:

pedrobaeza:
You should use the new base model inheritance instead of monkey-patching.

pedrobaeza:
Why runbot fails? I have launched a rebuild to see

espo-tony:
:+1: 

I guess the runbot fails because of:

> 2017-03-17 03:14:54,259 207 WARNING openerp_test odoo.fields: Field date.range.type.company_id: parameter 'select' is no longer supported; use 'index' instead.
> 2017-03-17 03:14:54,276 207 WARNING openerp_test odoo.fields: Field date.range.company_id: parameter 'select' is no longer supported; use 'index' instead.
> 2017-03-17 03:14:54,276 207 WARNING openerp_test odoo.fields: Field date.range.type_id: parameter 'select' is no longer supported; use 'index' instead.

Models _date.range.type_ and _date.range_ are defined in the module _date_range_, and in the latest version they use the correct parameter '_index_' instead of '_select_'. The author of the PR could need to update his local branch (the one from which this PR was generated) as it keeps the old version of the module and cause the warnings into the runbot.

Anyway surely is not related to this migration, which looks to be working.

espo-tony (approved):

PR #599, opened 2016-11-10
FIX secure_uninstall

bealdav:
## Steps to reproduce: 
- Install 'secure_uninstall' module
- Try uninstall another module providing a bad key

## Current behavior:
- Module state switch 'to remove' state, then it will finally be uninstalled with future updates of other modules.

## Expected behavior:
- Module state rollback to 'installed' state if uninstall is not authorized.

### DIFF
--- a/secure_uninstall/README.rst
+++ b/secure_uninstall/README.rst
@@ -40,6 +40,11 @@ Usage
    :target: https://runbot.odoo-community.org/runbot/149/server-tools
 
 
+Known issues / Roadmap
+======================
+* This module can be uninstall properly: needs to be fixed soon
+* This module allowed us to discover this bug in core Odoo https://github.com/odoo/odoo/issues/14452: hoping it will fixed
+
 Bug Tracker
 ===========
 
--- a/secure_uninstall/models/module.py
+++ b/secure_uninstall/models/module.py
@@ -24,6 +24,7 @@ class BaseModuleUpgrade(models.TransientModel):
     def upgrade_module(self):
         for elm in self:
             if not config.get("secure_uninstall"):
+                self.rollback_state_modules()
                 raise UserError(_(
                     "Missing configuration key\n--------------------\n"
                     "'secure_uninstall' configuration key "
@@ -31,6 +32,7 @@ class BaseModuleUpgrade(models.TransientModel):
                     "your Odoo server configuration file: "
                     "please set it a value"))
             if elm.uninstall_password not in _get_authorized_password():
+                self.rollback_state_modules()
                 raise UserError(_(
                     "Password Error\n--------------------\n"
                     "Provided password '%s' doesn't match with "
@@ -42,3 +44,15 @@ class BaseModuleUpgrade(models.TransientModel):
             # keep this password in db is insecure, then we remove it
             elm.uninstall_password = False
         return super(BaseModuleUpgrade, self).upgrade_module()
+
+    def rollback_state_modules(self):
+        modules = self.env['ir.module.module'].browse(
+            self._context.get('active_ids'))
+        if modules:
+            module_ids = modules.downstream_dependencies(
+                exclude_states=['uninstalled', 'uninstallable'])
+            module_ids = module_ids + list(modules._ids)
+            self.env['ir.module.module'].browse(
+                module_ids).button_uninstall_cancel()
+            self._cr.commit()
+        return True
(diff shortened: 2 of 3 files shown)


legalsylvain:
Hi @bealdav, I can not figure out how modules switch to 'to remove' state if password is bad. The current code of upgrade_module seems to prevent super call, if password is bad. Can you point how it occurs ?
Thanks !

bealdav:
errors come from 2016-11-10 16:52:01,265 6583 ERROR openerp_test openerp.sql_db: Programming error: column "database_cleanup_test" does not exist

LINE 1: select database_cleanup_test from res_users

bealdav:
@legalsylvain Without my module, before to create a record in 'base.module.upgrade' odoo write module state to 'to remove'.
If you check cancel button on the wizard, state is yet written but to 'installed' this time.

Raise prevent to rewrite state field, except if you use cr.commit()

bealdav:
@legalsylvain you can check without this module and test with write method on `ir.module.module`. 
module state field is set 'to remove' value before `base.module.upgrade` wizard is opened. State field is reset to 'installed' if installation is cancelled

https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/module/wizard/base_module_upgrade.py#L76

Then in my case before raise, I need to rollback to old state.

hparfr:
👍

legalsylvain:
Could you wait a time. I'd like to do extra test. Thanks.

mourad-ehm:
:+1:

bealdav:
@legalsylvain it's somewhat an emergency fix.
As it this module can damage erp.
I think it's better to merge (2 ok), and if there is a better way to do things it can be done later.
Except if you can evaluate it today, I propose to merge.
my 2 cents

legalsylvain:
Hi. 
Well I'm still not in favor of this patch : 
secure_uninstall installed, if you click on "uninstall" module, (on the module form)

you have 3 ways : 
1. You set a valid password and click on confirm. (--> OK)
2. You click on cancel (--> OK (*))
3. You set an invalid password and then click on cancel (--> OK (*))
4. You do wathever you want and close the tab with the top right button --> KO

(*) 2 + 3 : It' not possible because the field password is required. You should set it required to False. I tested and it works. https://github.com/OCA/server-tools/blob/8.0/secure_uninstall/views/module_view.xml#L16

4 : This is a core bug, that should be fixed by a core fix, not in this module. BUG : 

- create a new db, without any custom module. (just web)
- install by ui 'mail'
- click on 'uninstall' and then close the tab
- the module is marked as 'to be uninstalled'

This is really not the expected behaviour.

Extra bug on this module : I can not uninstall 'secure_uninstall'. It freezes. Did you tried ?

regards.

bealdav:
Ok nice sylvain for your wide and valuable review.

as I see, 3 bugs:

- A: bug you describe as 4/
- B: bug "I can not uninstall 'secure_uninstall'. It freezes": needs to be fixed
- C: bug without this patch: customer can loose data with most standard way to use ERP (not close tab)

C is an emergency bug for customers which use this module now without this patch (I don't want forget them)

With or without this patch there is no impact on A or B.

For A I need define an issue on Odoo
For B I need to investigate

While I investigate, and Odoo SA review, this PR stay there and somebody can loose some data because of A.

On this base what do you think to approve PR now ? 
If OK, I'll make the issue right now and investigate for B with a PR, for sure.

Thanks for your time

legalsylvain:
Hi @bealdav , thanks for your answer. Your analysis seems correct. 

For C point, I see two minor improvments : 
- could you accept this PR, that allows user to click on "cancel" button ? (https://github.com/akretion/server-tools/pull/8). (As this field is required, it's not possible for the time being)

- Just add a text in the readme file / description part, to mention :
-> Know Issues / Roadmap : B point description you just wrote
-> Note : This module fixes A point. (with link to your issue)

So, all will be clear for the admin user that will install this module.
After that, no problem to merge quickly this PR. I understand, that it's a problematic bug.

kind regards.

bealdav:
https://github.com/odoo/odoo/issues/14452

bealdav:
@legalsylvain all changes done, thanks

legalsylvain:
Thanks !

bealdav:
@legalsylvain not approve ?

bealdav:
thanks

Issue #585, opened 2016-10-24
auth_signup_verify_email: user lang not stored on signup


simahawk:
The base module `auth_signup` uses `do_signup` https://github.com/odoo/odoo/blob/9.0/addons/auth_signup/controllers/main.py#L99

but `auth_signup_verify_email` skips it

https://github.com/OCA/server-tools/blob/9.0/auth_signup_verify_email/controllers/main.py#L43

which is correct for its behavior but is misses the user language retrieved by current request lang.

pedrobaeza:
Can you please do a PR amending this?

simahawk:
@pedrobaeza there you go ;)

pedrobaeza:
Cool! I'll review it. I close this issue as there's an opened PR.

PR #602, opened 2016-11-12
[fix] auth_signup_verify_email user lang

simahawk:
Addresses #585

### DIFF
--- a/auth_signup_verify_email/controllers/main.py
+++ b/auth_signup_verify_email/controllers/main.py
@@ -15,6 +15,10 @@ except ImportError:
 
 
 class SignupVerifyEmail(AuthSignupHome):
+
+    # TODO: make this configurable
+    DEFAULT_USER_LANG = 'en_US'
+
     @http.route()
     def web_auth_signup(self, *args, **kw):
         if (http.request.params.get("login") and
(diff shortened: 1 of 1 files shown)


auth_signup_verify_email/controllers/main.py
```diff
@@ -23,6 +27,17 @@ def web_auth_signup(self, *args, **kw):
         else:
             return super(SignupVerifyEmail, self).web_auth_signup(*args, **kw)
 
+    def _get_user_lang(self):
+        """Retrieve user language."""
+        langs = http.request.env['res.lang'].search_read([], ['code'])
+        supported_langs = [
+            lang['code'] for lang in langs
+        ]
+        lang = self.DEFAULT_USER_LANG
```
simahawk: @pedrobaeza maybe for this we can take the website language, what do you think?

pedrobaeza: Maybe website default language here?

simahawk: @pedrobaeza I commented the same but didn't submit the review comment :face_with_head_bandage:

pedrobaeza: hehe, then go ahead with that language default.

simahawk:
@pedrobaeza done

pedrobaeza (approved):
When CIs finish correctly

PR #603, opened 2016-11-13
[MIG] 9.0 - auditlog

sebalix:
Supersedes #454 (no response)

Migration done following the instructions [from the wiki](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#technical-method-to-migrate-a-module-from-80-to-90-branch) + Cherry-picked the commits done by @holdenrehg in #454.


pedrobaeza:
Thanks for the followup. Please squash all the "OCA Transbot" commits in one, and check Travis status.

sebalix:
Module tested on runbot (full log, fast log, user session, HTTP requests...), works like expected.

Travis fails on the `database_cleanup` module.

sebalix:
As soon as this module is merged I'll port it on 10.

sebalix:
@pedrobaeza sorry I didn't see the lint errors, it's fixed now.

I also migrated the last methods to the new API, PR ready for reviews now.

auditlog/models/autovacuum.py
```diff
@@ -0,0 +1,41 @@
...
+        data_models = (
+            'auditlog.log',
+            'auditlog.http.request',
+            'auditlog.http.session',
+        )
+        for data_model in data_models:
+            records = self.env[data_model].search(
+                [('create_date', '<=', fields.Datetime.to_string(deadline))])
+            nb_records = len(records)
+            records.unlink()
```
pedrobaeza, line 37: Maybe we need to launch a real "VACUUM" here?

pedrobaeza (approved):
Little comment, but overall :+1:

auditlog/models/autovacuum.py
sebalix: Why not, but I would prefer in a new PR as it's a new feature.
About the VACUUM, I was thinking that Odoo already performs this operation through `ir.autovacuum` https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_autovacuum.py but it's not a real VACUUM neither.

lasley (approved):
Thanks @sebalix - I agree with your eval that adding the actual vacuum should be a separate PR, which would also allow for easy backports

lasley:
Travis is 🔴  but not on your module and I think it's been resolved already. Restarting build; this can be merged if 🍏

sebalix:
PR rebased but it's failing on `database_cleanup`.

pedrobaeza:
This PR is supposed to fix it: https://github.com/OCA/server-tools/pull/612

sebalix:
Rebased to include #612 , Travis is now green

pedrobaeza:
OK, merging. Thanks for the PR.

Issue #146, opened 2015-03-30
Module to manage user roles


sebalix:
Hi,

As a standard group is a notion a bit too technical sometimes, we developed a module to manage functional roles. A role is just an aggregation of groups. It helps us to manage a lot of user accounts over time without dealing with a collection of technical groups or "user template" account to duplicate when we need to create a new one (not really maintainable).
The code and description is hosted here: https://github.com/osiell/base_user_role/

I have not encountered similar modules, so I'm asking:
- is this module irrelevant? (we're missing a point or something within the standard functionalities)
- if not, is it interesting to maintain it in OCA? (in this repo I suppose)

Regards,

matthieuchoplin:
Hi,

Maybe this module could do what you want: https://github.com/Smile-SA/odoo_addons/tree/8.0/smile_access_control

sebalix:
Hi,
Thank you for sharing your work, this is exactly this kind of functionality we miss with Odoo.
I just tested your module, it works great! Sharing user fields is really useful (groups & home action especially).

I only miss one feature: apply several profiles on a user. I mean, a person could have many jobs in a company, each job giving access to menus/actions and so on. This combination of profiles can not really interpreted as a new profile to be linked to the relevant users I think (and it's not the purpose, as we want to combine them sometimes).

Do you plan to support this?
Your private/public fields feature is really interesting, maybe I could implement it in our own module instead.

matthieuchoplin:
I am glad you find this module interesting. I think it is better to speak to @corentinpb about it as he is the main maintainer.

yelizariev:
I can suggest my module 
https://apps.odoo.com/apps/modules/8.0/access_base/

It uses more simple approach. It just creates special category "Custom User Groups" and puts access groups of that category to the top of "Access Rights" tab of user form. 
So you create a group that is collection of a technical groups. And you can use that group as a role.
You can apply such role to your test user to check everything.
It allows to apply several roles for users.

sebalix:
Your module is very similar to ours in fact. From what I see, the differences are:
- you always have the possibility to update groups of a user manually (a matter of taste, starting from several tens of users to manage we prefer to disallow this possibility, but this kind of behaviour could be optional),
[...]

PR #608, opened 2016-11-15
[ADD] 8.0 - New module 'base_user_role' to manage user roles efficiently

sebalix:
Note : we are using this module on version 6.0, 7.0 and 8.0 I just added the support for dates recently on the 8.0 version. I'll migrate it to 9.0 and 10.0 soon.

Related issue: #146 

User roles
========

This module was written to extend the standard functionality regarding users and groups management. It helps creating well-defined user roles and associating them to users.

It can become very hard to maintain a large number of user profiles over time, juggling with many technical groups. For this purpose, this module will help you to:

  * define functional roles by aggregating low-level groups,
  * set user accounts with the predefined roles (roles are cumulative),
  * update groups of all relevant user accounts (all at once),
  * ensure that user accounts will have the groups defined in their roles (nothing more, nothing less). In other words, you can not set groups manually on a user as long as there is roles configured on it,
  * activate/deactivate roles depending on the date (useful to plan holidays, etc)
  * get a quick overview of roles and the related user accounts.

That way you make clear the different responsabilities within a company, and are able to add and update user accounts in a scalable and reliable way.

Configuration
=============

To configure this module, you need to go to *Configuration / Users / Roles*, and create a new role. From there, you can add groups to compose your role, and then associate users to it.


sebalix:
About Travis: flake8 errors and tests failure not related to this module.

jcdrubay:
Very good contribution (high quality of code, useful, well documented).
I hope it can be reviewed and merged soon :+1:

base_user_role/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    'name': 'User roles',
+    'version': '8.0.1.1.0',
+    'category': 'Tools',
+    'author': 'ABF OSIELL, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'maintainer': 'ABF OSIELL',
+    'website': 'http://www.osiell.com',
+    'depends': [
```
lasley, line 13: You can remove this key. `base` dependency is implicit.

base_user_role/models/role.py
```diff
@@ -0,0 +1,84 @@
...
+    group_id = fields.Many2one(
+        'res.groups', required=True, ondelete='cascade',
+        readonly=True, string=u"Associated group")
+    line_ids = fields.One2many(
+        'res.users.role.line', 'role_id', string=u"Users")
+    user_ids = fields.One2many(
+        'res.users', string=u"Users", compute='_compute_user_ids')
+
+    @api.multi
+    @api.depends('line_ids')
```
lasley: `line_ids.user_id`

base_user_role/models/role.py
```diff
@@ -0,0 +1,84 @@
...
+    def write(self, vals):
+        res = super(ResUsersRole, self).write(vals)
+        self.update_users()
+        return res
+
+    @api.multi
+    def update_users(self):
+        """Update all the users concerned by the roles identified by `ids`."""
+        users = self.mapped('user_ids')
+        if users:
```
lasley: There shouldn't be need for this guard, mapped will just return an empty Recordset that should work with the other method just fine due to record loop

base_user_role/models/role.py
```diff
@@ -0,0 +1,84 @@
...
+    _name = 'res.users.role.line'
+    _description = 'Users associated to a role'
+
+    role_id = fields.Many2one(
+        'res.users.role', string=u"Role")
+    user_id = fields.Many2one(
+        'res.users', string=u"User")
+    date_from = fields.Date(u"From")
+    date_to = fields.Date(u"To")
+    is_active = fields.Boolean(u"Active", compute='_compute_is_active')
```
lasley: `active` is the Odoo norm for this col - although I'm with you on prefixing bool cols with linking verbs like `is`, `has`, etc. It also comes with some defaults filters and such, so makes it worth using despite the hurt to naming OCD

That said, it also needs to be `store=True` to allow all of that

base_user_role/models/user.py
```diff
@@ -0,0 +1,53 @@
...
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    role_line_ids = fields.One2many(
+        'res.users.role.line', 'user_id', string=u"Role lines")
+    role_ids = fields.One2many(
+        'res.users.role', string=u"Roles", compute='_compute_role_ids')
+
+    @api.multi
+    @api.depends('role_line_ids')
```
lasley: role_line_ids.role_id

lasley (changes requested):
This is awesome, thanks @sebalix 

Code, no test. Comments inline

lasley:
Oh please also check Travis and Runbot 😉

base_user_role/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_res_users_role,access_res_users_role,model_res_users_role,"base.group_erp_manager",1,1,1,1
```
jcdrubay, line 2: From travis https://travis-ci.org/OCA/server-tools/jobs/176051212#L1703

> The model res.users.role.line has no access rules, consider adding one. E.g. 

`access_res_users_role_line,access_res_users_role_line,model_res_users_role_line,"base.group_erp_manager",1,0,0,0`

sebalix:
Thank you for the review, I'll fix these points soon

base_user_role/__openerp__.py
pedrobaeza: No, you should depend on your module at least in base, but base is implicit when depending on other module.

base_user_role/models/role.py
sebalix: Indeed, but I was unable to set a working `domain` on the XML views to display all the lines (as we don't want to hide them from the administrator in any case, we just want to filter them when setting groups). If I set the domain on the Python field, it works fine, but then it does not dispense to filter the lines in the `ResUsers.set_groups_from_roles` method. And as I've seen no benefit to use the standard `active` field I ended up adding my own.

base_user_role/security/ir.model.access.csv
sebalix: Fixed

base_user_role/models/role.py
sebalix: Or I could rename it to 'is_enabled'?

base_user_role/models/role.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ResUsersRoleLine(models.Model):
+    _name = 'res.users.role.line'
+    _description = 'Users associated to a role'
+
+    role_id = fields.Many2one(
+        'res.users.role', string=u"Role", ondelete='cascade')
+    user_id = fields.Many2one(
+        'res.users', string=u"User")
+    date_from = fields.Date(u"From")
```
susport, line 66: Date from may be default the date now of the create moment.
Good contribution.
👍

sebalix: Dates are optional, no dates = role activated immediately

base_user_role/models/role.py
lasley: I like the `is_enabled` idea - it is far enough away from `active` that people won't be trying to change it all the time. I know I would have probably done this during a migration had I not already gotten the reasoning behind it.

sebalix: It's done!

sebalix:
I think all the points are fixed.

lasley:
Travis is still mad 🔴

sebalix:
You mean I have to fix lint errors of others modules in this one? Tests fail on `fetchmail_notify_error_to_sender`.

lasley (approved):
Hmm yeah ok the build looks broken on 8.0 branch in the same way, so you are correct that we should not be fixing it here.

sebalix:
I fixed the other modules in #616 to get green builds.

sebalix:
Rebased

sbidoul:
@adrienpeiffer you should have a look at this.

andhit-r:
Tested functionality on runbot :+1:

andhit-r:
@mikevhe18 @azmimr67 you guys should take a look at this PR and help review

lasley:
We're approved for merge. @sebalix - do you prefer to squash your own commits or would you rather me do it? I don't think I've gotten this preference from you yet, but sorry if I'm double asking.

sebalix:
@lasley You can squash them, no trouble

lasley:
Thanks for the awesome new module @sebalix! Do you have any plans to migrate it forward?

sebalix:
@lasley Thank you! Yes, I plan to migrate it to v9 (see #641) and v10 right after.

PR #611, opened 2016-11-16
[9.0] Add oauth_provider module

Garamotte:
A new module that allows to use Odoo as OAuth provider.
The OAuth protocol is used to delegate authentication, and authorization to access some resources.

This module really need security reviews/tests, as I'm not fully comfortable with the OAuh protocol.

The RFC defines four profiles: Implicit grant, Authorization code grant, Password credentials grand and User credentials grant.
Only Implicit grant and Authorization code grant are implemented yet, and I don't plan to implement the others for now. The work should not be very hard, as most parts are common, but I don't actually need these profiles, and don't have time to do it for the moment.

Note: I created the PR to allow you to begin testing the module.
@pedrobaeza said that this repo will be split into several smaller repos, this PR will need to be recreated on one of the new repos when they are created.

### DIFF
--- a/oauth_provider/controllers/main.py
+++ b/oauth_provider/controllers/main.py
@@ -1,0 +1,290 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 SYLEAM
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import json
+import logging
+import werkzeug.utils
+import werkzeug.wrappers
+from datetime import datetime
+from openerp import http, fields
+from openerp.addons.web.controllers.main import ensure_db
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import oauthlib
+    from oauthlib import oauth2
+except ImportError:
+    _logger.debug('Cannot `import oauthlib`.')
+
+
+class OAuth2ProviderController(http.Controller):
+    def __init__(self):
+        super(OAuth2ProviderController, self).__init__()
+
+    def _get_request_information(self):
+        """ Retrieve needed arguments for oauthlib methods """
+        uri = http.request.httprequest.base_url
+        http_method = http.request.httprequest.method
+        body = oauthlib.common.urlencode(
+            http.request.httprequest.values.items())
+        headers = http.request.httprequest.headers
+
+        return uri, http_method, body, headers
+
+    def _check_access_token(self, access_token):
+        """ Check if the provided access token is valid """
+        token = http.request.env['oauth.provider.token'].search([
+            ('token', '=', access_token),
+        ])
+        if not token:
+            return False
+
+        oauth2_server = token.client_id.get_oauth2_server()
+        # Retrieve needed arguments for oauthlib methods
+        uri, http_method, body, headers = self._get_request_information()
+
+        # Validate request information
+        valid, oauthlib_request = oauth2_server.verify_request(
+            uri, http_method=http_method, body=body, headers=headers)
+
+        if valid:
+            return token
+
+        return False
+
+    def _json_response(self, data=None, status=200, headers=None):
+        """ Returns a json response to the client """
+        if headers is None:
+            headers = {'Content-Type': 'application/json'}
+
+        return werkzeug.wrappers.BaseResponse(
+            json.dumps(data), status=status, headers=headers)
+
+    @http.route('/oauth2/authorize', type='http', auth='user', methods=['GET'])
+    def authorize(self, client_id=None, response_type=None, redirect_uri=None,
+                  scope=None, state=None, *args, **kwargs):
+        """ Check client's request, and display an authorization page to the user,
+
+        The authorization page lists allowed scopes
+        If the client is configured to skip the authorization page, directly
+        redirects to the requested URI
+        """
+        client = http.request.env['oauth.provider.client'].search([
+            ('identifier', '=', client_id),
+        ])
+        if not client:
+            return http.request.render(
+                'oauth_provider.authorization_error', {
+                    'title': 'Unknown Client Identifier!',
+                    'message': 'This client identifier is invalid.',
+                })
+        oauth2_server = client.get_oauth2_server()
+
+        # Retrieve needed arguments for oauthlib methods
+        uri, http_method, body, headers = self._get_request_information()
+        try:
+            scopes, credentials = oauth2_server.validate_authorization_request(
+                uri, http_method=http_method, body=body, headers=headers)
+            # Store only some values, because the pickling of the full request
+            # object is not possible
+            http.request.httpsession['oauth_scopes'] = scopes
+            http.request.httpsession['oauth_credentials'] = {
+                'client_id': credentials['client_id'],
+                'redirect_uri': credentials['redirect_uri'],
+                'response_type': credentials['response_type'],
+                'state': credentials['state'],
+            }
+            if client.skip_authorization:
+                # Skip the authorization page
+                # Useful when the application is trusted
+                return self.authorize_post()
+        except oauth2.FatalClientError as e:
+            return http.request.render(
+                'oauth_provider.authorization_error', {
+                    'title': 'Error: {error}'.format(error=e.error),
+                    'message': e.description,
+                })
+        except oauth2.OAuth2Error as e:
+            return http.request.render(
+                'oauth_provider.authorization_error', {
+                    'title': 'Error: {error}'.format(error=e.error),
+                    'message': 'An unknown error occured! Please contact your '
+                    'administrator',
+                })
+
+        oauth_scopes = client.scope_ids.filtered(
+            lambda record: record.code in scopes)
+        return http.request.render(
+            'oauth_provider.authorization', {
+                'oauth_client': client.name,
+                'oauth_scopes': oauth_scopes,
+            })
+
+    @http.route(
+        '/oauth2/authorize', type='http', auth='user', methods=['POST'])
+    def authorize_post(self, *args, **kwargs):
+        """ Redirect to the requested URI during the authorization """
+        client = http.request.env['oauth.provider.client'].search([
+            ('identifier', '=', http.request.httpsession.get(
+                'oauth_credentials', {}).get('client_id'))])
+        if not client:
+            return http.request.render(
+                'oauth_provider.authorization_error', {
+                    'title': 'Unknown Client Identifier!',
+                    'message': 'This client identifier is invalid.',
+                })
+        oauth2_server = client.get_oauth2_server()
+
+        # Retrieve needed arguments for oauthlib methods
+        uri, http_method, body, headers = self._get_request_information()
+        scopes = http.request.httpsession['oauth_scopes']
+        credentials = http.request.httpsession['oauth_credentials']
+        headers, body, status = oauth2_server.create_authorization_response(
+            uri, http_method=http_method, body=body, headers=headers,
+            scopes=scopes, credentials=credentials)
+
+        return werkzeug.utils.redirect(headers['Location'], code=status)
+
+    @http.route('/oauth2/token', type='http', auth='none', methods=['POST'],
+                csrf=False)
+    def token(self, client_id=None, client_secret=None, redirect_uri=None,
+              scope=None, code=None, grant_type=None, username=None,
+              password=None, refresh_token=None, *args, **kwargs):
+        """ Return a token corresponding to the supplied information
+
+        Not all parameters are required, depending on the application type
+        """
+        ensure_db()
+
+        # If no client_id is specified, get it from session
+        if client_id is None:
+            client_id = http.request.httpsession.get(
+                'oauth_credentials', {}).get('client_id')
+
+        client = http.request.env['oauth.provider.client'].sudo().search([
+            ('identifier', '=', client_id),
+        ])
+
+        if not client:
+            return self._json_response(
+                data={'error': 'invalid_client_id'}, status=401)
+        oauth2_server = client.get_oauth2_server()
+
+        # Retrieve needed arguments for oauthlib methods
+        uri, http_method, body, headers = self._get_request_information()
+        credentials = {'scope': scope}
+
+        # Retrieve the authorization code, if any, to get Odoo's user id
+        existing_code = http.request.env[
+            'oauth.provider.authorization.code'].search([
+                ('client_id.identifier', '=', client_id),
+                ('code', '=', code),
+            ])
+        if existing_code:
+            credentials['odoo_user_id'] = existing_code.user_id.id
+        # Retrieve the existing token, if any, to get Odoo's user id
+        existing_token = http.request.env['oauth.provider.token'].search([
+            ('client_id.identifier', '=', client_id),
+            ('refresh_token', '=', refresh_token),
+        ])
+        if existing_token:
+            credentials['odoo_user_id'] = existing_token.user_id.id
+
+        headers, body, status = oauth2_server.create_token_response(
+            uri, http_method=http_method, body=body, headers=headers,
+            credentials=credentials)
+
+        return werkzeug.wrappers.BaseResponse(
+            body, status=status, headers=headers)
+
+    @http.route('/oauth2/tokeninfo', type='http', auth='none', methods=['GET'])
+    def tokeninfo(self, access_token=None, *args, **kwargs):
+        """ Return some information about the supplied token
+
+        Similar to Google's "tokeninfo" request
+        """
+        ensure_db()
+        token = self._check_access_token(access_token)
+        if not token:
+            return self._json_response(
+                data={'error': 'invalid_or_expired_token'}, status=401)
+
+        token_lifetime = (fields.Datetime.from_string(token.expires_at) -
+                          datetime.now()).seconds
+        # Base data to return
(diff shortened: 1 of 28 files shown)


Garamotte:
Travis errors unrelated (from database_cleanup and auth_signup_verify_email modules).

Garamotte:
Last little thing done (and rebased on 9.0).
You can add replace the WIP label by needs review.

oauth_provider/controllers/main.py
```diff
@@ -0,0 +1,316 @@
...
+            'oauth_provider.authorization', {
+                'oauth_client': client.name,
+                'oauth_scopes': oauth_scopes,
+            })
+
+    @http.route(
+        '/oauth2/authorize', type='http', auth='user', methods=['POST'])
+    def authorize_post(self, *args, **kwargs):
+        """
+        Redirect to the requested URI during the authorization
```
lasley: Not a blocker or anything, just FYI - Docstrings should start on the same line that the quote is started on. The first line (in a multi) should be a complete summary, followed by a blank line, followed by more description if necessary. Check out [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) for more

oauth_provider/controllers/main.py
```diff
@@ -0,0 +1,316 @@
...
+        oauth2_server = token.client_id.get_oauth2_server()
+
+        # Retrieve needed arguments for oauthlib methods
+        uri = http.request.httprequest.base_url
+        http_method = http.request.httprequest.method
+        body = oauthlib.common.urlencode(
+            http.request.httprequest.values.items())
+        headers = http.request.httprequest.headers
+
+        valid, oauthlib_request = oauth2_server.verify_request(
```
lasley: This token & argument retrieval followed verification block seems to be repeated a lot. It would be nice if it was just a helper method instead.

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+class OdooValidator(RequestValidator):
+    """
+    OAuth2 validator to be used in Odoo
+    """
+    def authenticate_client(self, request, *args, **kwargs):
+        """
+        Authenticate the client
+        """
+        raise NotImplementedError
+        # return False
```
lasley: Please remove this commented code

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+
+class OdooValidator(RequestValidator):
+    """
+    OAuth2 validator to be used in Odoo
+    """
+    def authenticate_client(self, request, *args, **kwargs):
+        """
+        Authenticate the client
+        """
+        raise NotImplementedError
```
lasley: No client auth implementation?

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+        # return False
+
+    def authenticate_client_id(self, client_id, request, *args, **kwargs):
+        """
+        Ensure client_id belong to a non-confidential client
+        """
+        request.client = http.request.env['oauth.provider.client'].search([
+            ('identifier', '=', client_id),
+        ])
+        request.odoo_user = http.request.env['oauth.provider.client'].env.user
```
lasley: Why the `env` and `user` from `oauth.provider.client`? It's the same `env` and `user` isn't it, or is there some fancy context switching that I am missing?

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+                ('refresh_token', '=', token),
+            ])
+            db_token.refresh_token = False
+
+    def rotate_refresh_token(self, request):
+        """
+        Determine if the refresh token has to be renewed
+        Called after refreshing an access token
+        """
+        return True
```
lasley: This doesn't seem very useful, or is it an interface for child classes?

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        Store the authorization code into the database
+        """
+        redirect_uri = http.request.env['oauth.provider.redirect.uri'].search([
+            ('name', '=', request.redirect_uri),
+        ])
+        http.request.env['oauth.provider.authorization.code'].sudo().create({
+            'code': code['code'],
+            'client_id': request.client.id,
+            'user_id': request.odoo_user.id,
```
lasley, line 168: Is it ever possible for this to be hit without the `authenticate_client_id` being called? (real question)

oauth_provider/oauth2/validator.py
```diff
@@ -0,0 +1,253 @@
...
+        return request.client.identifier == client_id and set(scopes).issubset(
+            set(request.client.mapped('scope_ids.code')))
+
+    def validate_user(
+            self, username, password, client, request, *args, **kwargs):
+        """
+        Ensure the usernamd and password are valid
+        """
+        raise NotImplementedError
+        # return False
```
lasley: Remove commented code

oauth_provider/security/oauth_provider_security.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        ##############################################################################
```
lasley: Wait... how is this not breaking XML? Please use the shorter header though, and put just under the xml version. This is how I do it https://github.com/OCA/vertical-medical/blob/9.0/medical/views/medical_menu.xml#L3 - although we don't have specific guidelines in OCA on the XML part AFAIK

oauth_provider/security/oauth_provider_security.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
```
lasley: `<odoo>` and can remove data tag

oauth_provider/templates/authorization.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
lasley: Odoo tag and remove data

oauth_provider/templates/authorization.xml
```diff
@@ -0,0 +1,60 @@
...
+    <data>
+        ##############################################################################
```
lasley: Same note regarding use of smaller header

oauth_provider/tests/test_oauth_provider_client.py
```diff
@@ -0,0 +1,110 @@
...
+    def test_grant_response_type_fields(self):
+        """
+        Check the value of the grant_type and response_type fields
+        """
+        # Default : Web Application
+        client = self.new_client({'identifier': 'default'})
+        self.assertEqual(client.grant_type, 'authorization_code')
+        self.assertEqual(client.response_type, 'code')
+
+        # Web Application
```
lasley: IMO the scope of all these tests is too large. It's best to make smaller test methods in order to provide a better view of things when stuff is broken vs. execution stopping on the first broken test.

lasley (changes requested):
Awesome thanks @sylvain-garancher - comments inline. Also my bad on leaving the WIP tag for so long, missed the comment.

oauth_provider/controllers/main.py
Garamotte: Didn't know that, thanks.
I'll have to change my comments on many other PRs :D

oauth_provider/oauth2/validator.py
Garamotte: This file is an implementation of the interface provided by the `oauthlib` module.
As the user is already logged on Odoo, we have the user in the `request.env`, so I didn't need to implement this method (unless I miss something).

oauth_provider/oauth2/validator.py
Garamotte: I leaved this line here to remind the return value's type, but I may add this in the docstring.

oauth_provider/oauth2/validator.py
Garamotte: I think this is a bad copy/paste from the previous line, I'll fix that :)

oauth_provider/oauth2/validator.py
Garamotte: This is the implementation of an interface.
The default value in the interface is True, si this is not needed, but I preferred to implement all methods. We can remove this without any issue, if it is better.

oauth_provider/controllers/main.py
Garamotte: You're right, I'll provide a helper method (which will be useful for modules depending on this one too).

oauth_provider/oauth2/validator.py
Garamotte: Yes, because the user is already logged in Odoo (the first route has `auth='user'`).
The `authenticate_client_id` should be required for the two unimplemented profiles (which need to privide credentials on the first request).

oauth_provider/security/oauth_provider_security.xml
Garamotte: Oops, I kept the old templates for my XML files :)

oauth_provider/security/oauth_provider_security.xml
Garamotte: This is part of our own templates since at least v5, I think this is time to update our templates.
I'll change that.

oauth_provider/oauth2/validator.py
lasley: Can we instead just add a note about the method being primarily for inheritance by child classes in the docstring? This is typically what Odoo SA does with interfaces I think, and would make it more clear for a module developer doesn't understand all the base code.

oauth_provider/oauth2/validator.py
lasley: Docstring would be the preferred please 😄

oauth_provider/oauth2/validator.py
lasley: Sweet I had a feeling I was missing a piece from the auth workflow. Thanks for the explanation

oauth_provider/controllers/main.py
lasley: Hah it's one of those things that you can't unsee once you know it, right?!

oauth_provider/oauth2/validator.py
lasley: Nope nope, was definitely me missing something. Can we add a docstring note that this is an interface then?

oauth_provider/tests/test_oauth_provider_client.py
Garamotte: You advise to split in one test method per possible value ?

I left all in a single method, because every test in this method are really close, but I don't know all best practices about tests.

lasley: Yeah the smaller the test the better. I go to great lengths to follow a 1:1 method:assertion ratio, but sometimes it's not possible given context and setup required. 

Consider an example using these tests. If the response_type for the web application is wrong, we won't know if it's right for Mobile, Legacy, and Backend because the test will stop execution at the first failed assertion. 

It's incredibly valuable to have the whole picture when debugging failing tests because it can help point us closer to the actual error, or specific edge case that is being debugged (maybe only X of the Y are failing).

Garamotte: Ok, I understand. I'll split the tests when it is possible ;)

Garamotte:
@lasley I updated the PR with all of your comments.

I also created a common class for both controllers tests (initialization and many tests are the same for all controllers).
Is it done right ? I prefixed all methods by an `_` because Odoo ran tests from the common class in addition to running them from the specific classes. I'd like to avoid the need to redefine each common test in the custom test classes, if possible, but didn't found how to do that.

Garamotte:
Rebased on 9.0, but runbot seems broken.
Is it caused by the Docker build ? Do I need to update some configuration files in the repo ?

lasley (approved):
@sylvain-garancher LGTM OCA/runbot-addons#111 for Runbot issue, which is definitely not related to anything you did

moylop260 (approved):
LGTM 
I'll review the runbot issue

Garamotte:
I'll finally have to implement the password credentials grant (token request by providing username and password), and add it in this PR.

Garamotte:
A short reminder for myself : Check if everything works well with multiple databases (unauthenticated routes may not work if no database has been selected).

Garamotte:
I also have a PoC for using [JWT](https://jwt.io) instead of a randomly generated token.

Can I add it in this module, or provide another module later to add this feature ?
It requires another python dependency (pyjwt), which is not required for simple OAuth2 authentication.

lasley:
IMO JWT should be a separate module. Not everyone is going to want it (although I do and can't wait :P).

Garamotte:
@lasley If you want to take a look, I added the JWT module here: https://github.com/syleam/server-tools/tree/9.0-add-oauth_provider-jwt-module

I'll create for the new module a PR after this one is merged, but I still have some bugs to fix here.

Garamotte:
I re-added WIP in the PR name.
I have some issues with the web application profile, and I'm currently developing the legacy application profile.

Garamotte:
Waiting for tests by the customer, but the Legacy application profile (password credentials grant type) is now implemented.

I'm not very proud of the `test_full_process` methods (in each controllers test class), but everything else should be ok.

Garamotte:
Tests improvements:
- Better split controller's test stuff from real test methods
- Added `login` and `logout` to avoid the need to mock the env in each method
- I finally found how to include test methods in separate classes without the need to call them explicitely in the real test class

I also fixed the web and legacy application behaviours (`/oauth2/token` using POST, etc.)...
Still needs tests to confirm that this now works as expected in OAuth's RFC.

Garamotte:
New changes:
- Specified HTTP status code for errors (400/401)
- Updated tests to check the HTTP status code after each request

Garamotte:
Tests split is finished, I think this is good now, so I removd the WIP tag in title.
Still waiting for a confirmation from my customer for a potential cors issue.

Garamotte:
After a rebase on 9.0, Travis is failing on many other modules, but not on this module's tests.

Garamotte:
After many tests, my customer has no new issue now, using the "legacy application" profile.
The cors issue doesn't seem to be linked to this module.

lasley (approved):
Thanks @sylvain-garancher - code looks good. Also confirmed that the CI fail isn't you and is instead #689

Garamotte:
I have an issue with the default values for UUID fields (`oauth_identifier` on `res.users`).
The default looks like this : https://github.com/OCA/server-tools/pull/611/files#diff-4a2885ce924a60c01c46e75c1663da33R14

When installing the module, all existing users get the same UUID as identifier... which is not what we want. How could we generate a new random UUID for each user ?
This is because the default_get method is only called once for all existing records : https://github.com/odoo/odoo/blob/5f63878ff6af844b6eab8d233a3f52ee7622eccd/openerp/models.py#L2396-L2419

I initially though about a `pre_init_hook` which could create the column in database, then populate it (to allow the unique constraint be added on the first run), but there will still be an issue if a script creates users in batches.
Maybe overriding the create method in addtion to a `pre_init_hook` ? Any other idea ?

lasley:
We circumvented this in medical by moving the default generation into an overrided create method instead. Looks like that would still work here? Basiccally:


```
class TestClass:
  def create(self, vals):
    vals['unique_key'] = uuid4()
    return super(TestClass, self).create(vals)
```

Downfall of this is from a UI perspective, because the default won't exist until save. We considered it not a problem in our context, and I don't think it would be here either - the val is read-only for humans anyways.

Garamotte:
I added the `pre_init_hook` for now, as our main issue was the initial install of the module, kept in a separate commit for the review, will be squashed later.

I also added the missing `copy=False` attribute on the two uuid fields (`oauth_identifier` on `res.users` and `identifier` on `oauth.provider.client`). Creating user accounts with the `auth_signup` module is done by duplicating the template user, and all new users had the same identifier.

Do you still see a need for overriding the `create` method ?
When creating several users at once, either the identifier will be in data (and if the same value appears multiple times, the unique constraint will raise), or the identifier will not be in data (in this case, the default function should be called for each new user).

lasley:
@> Do you still see a need for overriding the create method ?

Nope, I'd say you covered the bases. We ended up having to implement the create I think due to layers and layers of delegate inheritance at play & some connectors that bulk add data in ways a user normally wouldn't.

Looks like lint is 🔴  though. The other two failures are known, one of which is being resolved in #719 & the other should clear itself per Pedro's email earlier.

Garamotte:
Lint fixed, rebased on 9.0 to benefit from the tests fixes of the database_cleanup module.
Everything looks good.

Garamotte:
Rebased on 9.0 to resolve the requirements.txt conflict.
This module is now used in production by our customer, since a couple of weeks, without any issue :)

lasley (approved):
We've been in prod for a few weeks now too. Excellent work, @sylvain-garancher

Garamotte:
Rebased on 9.0 to fix the `requirements.txt` conflict.

lasley:
Thanks for another badass module @sylvain-garancher

PR #612, opened 2016-11-17
[FIX] database_cleanup: Fix test

moylop260:
- Follow recommendation from https://github.com/odoo/odoo/pull/13458#issuecomment-248901816
- Fix blocking tables issues
- Delete trash information

For the merger: I @moylop260 authorize a `squash and merge` here in order to leave the commits comments like as history.


database_cleanup/tests/test_database_cleanup.py
```diff
@@ -78,3 +89,9 @@ def test_database_cleanup(self):
         with self.assertRaises(ProgrammingError):
             with self.registry.cursor() as cr:
                 self.env.cr.execute('select * from database_cleanup_test')
+        with self.registry.cursor() as cr2:
+            # Release the delete of ir_module_module pending
+            self.env.cr.rollback()
+            cr2.execute(
+                "DELETE FROM ir_module_module WHERE id=%s", (module.id,))
```
hbrunn: if this is necessary, the test has failed. Or better: Then we have the problem that the commits the module needs to do shouldn't be committed. Maybe it's better to defuse `cr.commit` at the beginning of the test and reinstate it here?

database_cleanup/models/purge_models.py
```diff
@@ -87,11 +87,15 @@ def purge(self):
                     pass
                 except AttributeError:
                     pass
+            if line.name not in self.env:
```
hbrunn: do we need this now that the test runs at the end of everything? The problem is that for the actual use case of the module, this will happen very often, and reloading the registry for every unknown model will take very long. The alternative would be to check for the testing flag and do it only then if it's still necessary for the test

database_cleanup/models/purge_models.py
```diff
@@ -87,11 +87,15 @@ def purge(self):
...
+                # Reload env
+                self.pool.setup_models(self._cr, partial=(not self.pool.ready))
             self.env['ir.model.relation'].search([
                 ('model', '=', line.name)
             ]).with_context(**context_flags).unlink()
-            self.env['ir.model'].browse([row[0]])\
-                .with_context(**context_flags).unlink()
+            model = self.env['ir.model'].browse([row[0]])
+            if line.name in self.env:
```
hbrunn: this defeats the purpose of purging undefined modules

hbrunn (changes requested):

database_cleanup/tests/test_database_cleanup.py
moylop260: If you run 2 times this test, you will a contraint error of ir_module_module name unique.
It's because the process: `self.env['ir.module.module'].create(...)` and `RegistryManager.registries[self.env.cr.dbname] = original_registry` have a custom cursor with commit (different to current `self.env.cr`)

Then the test purge using `self.env.cr` and work fine because that cursor have a delete.
But all tests use `rollback` at final.
Then you will have a dirty database with a module called: `database_cleanup_test`

I mean, without this patch
- `cursor_x`

``` bash
BEGIN;
INSERT INTO ir_module_module
COMMIT;
END;
```
- `self.env.cr`

``` bash
BEGIN;
.... # many other transactions
DELETE FROM ir_module_module
ROLLBACK;
END;
```

database_cleanup/models/purge_models.py
moylop260: I know but I didn't find the way to fix the issue (I revert this change to see the error from travis):

``` bash
   ======================================================================
    ERROR: test_database_cleanup (openerp.addons.database_cleanup.tests.test_database_cleanup.TestDatabaseCleanup)
    `   File "/root/build/OCA/server-tools/database_cleanup/tests/test_database_cleanup.py", line 58, in test_database_cleanup
    `     purge_models.purge_all()
    `   File "/.repo_requirements/odoo/openerp/api.py", line 248, in wrapper
    `     return new_api(self, *args, **kwargs)
     `   File "/root/build/OCA/server-tools/database_cleanup/models/purge_wizard.py", line 51, in purge_all
     `     self.mapped('purge_line_ids').purge()
     `   File "/.repo_requirements/odoo/openerp/api.py", line 248, in wrapper
     `     return new_api(self, *args, **kwargs)
     `   File "/root/build/OCA/server-tools/database_cleanup/models/purge_models.py", line 94, in purge
     `     model.with_context(**context_flags).unlink()
     `   File "/.repo_requirements/odoo/openerp/api.py", line 248, in wrapper
     `     return new_api(self, *args, **kwargs)
     `   File "/.repo_requirements/odoo/openerp/api.py", line 574, in new_api
     `     result = method(self._model, cr, uid, self.ids, *args, **old_kwargs)
     `   File "/.repo_requirements/odoo/openerp/addons/base/ir/ir_model.py", line 147, in unlink
     `     model.field_id._prepare_update()
     `   File "/.repo_requirements/odoo/openerp/api.py", line 248, in wrapper
     `     return new_api(self, *args, **kwargs)
     `   File "/.repo_requirements/odoo/openerp/addons/base/ir/ir_model.py", line 445, in _prepare_update
     `     model = self.env[record.model]
     `   File "/.repo_requirements/odoo/openerp/api.py", line 768, in __getitem__
     `     return self.registry[model_name]._browse(self, ())
     `   File "/.repo_requirements/odoo/openerp/modules/registry.py", line 84, in __getitem__
     `     return self.models[model_name]
     ` KeyError: u'x_database.cleanup.test.model'
```

database_cleanup/tests/test_database_cleanup.py
hbrunn: I see. Well, it's not beautiful, but I don't think it's worth the time to get this cursor's commit patched away to by starting a patch fest of the involved methods. Ignore this comment then

pedrobaeza:
@moylop260, please rebase that I have fixed 2 small PEP8 problems.

database_cleanup/tests/test_database_cleanup.py
```diff
@@ -34,16 +48,18 @@ def test_database_cleanup(self):
...
             'model': 'x_database.cleanup.test.model',
         })
+
         self.env.cr.execute(
             'insert into ir_attachment (name, res_model, res_id, type) values '
             "('test attachment', 'database.cleanup.test.model', 42, 'binary')")
         self.registry.models.pop('x_database.cleanup.test.model')
         self.registry._pure_function_fields.pop(
             'x_database.cleanup.test.model')
+        self.registry.setup_models(self.env.cr, partial=False)
```
hbrunn: right, doing this actually makes this a good model, so the next line fails because there's nothing to purge. Just pop the model from the registry immediately, then this goes well

moylop260:
@hbrunn 
Thanks for feedback

IMHO this is a real issue from [openerp/addons/base/ir/ir_model.py#L147](https://github.com/odoo/odoo/blob/eed11c57bb531cf308594dce475e68070bee9a31/openerp/addons/base/ir/ir_model.py#L147)
Inside `model.field_id._prepare_update()` odoo is using [`self.env[record.model]`](https://github.com/odoo/odoo/blob/eed11c57bb531cf308594dce475e68070bee9a31/openerp/addons/base/ir/ir_model.py#L445) added from https://github.com/odoo/odoo/pull/13978

But here we don't have a model in environment.
Then we have the key error by this reason.

moylop260:
That line is not exists [from 8.0 version](https://github.com/odoo/odoo/blob/205b5542ca274c2c7dc501b4f30e82e53b45a7bf/openerp/addons/base/ir/ir_model.py#L184) then if I comment this one the test pass fine!

Then we should create a fix to odoo/odoo before.

It mean that `database_cleanup` it is not working for 9.0 for now to delete models.

moylop260:
We have false red builds by this issue.
Now is false green just for `database_cleanup` since the problem inject from https://github.com/odoo/odoo/pull/13978

IMHO we should avoid the false red for other builds with this PR
What do you think?

sbidoul:
@hbrunn what's the status here? Is this good for merging to get the 9.0 branch green?

hbrunn (approved):

PR #614, opened 2016-11-18
[MIG][10.0] auto_backup

lasley:
* Change openerp to odoo
* Bump version
* Rename manifest
* Add pysftp to requirements


auto_backup/README.rst
```diff
@@ -0,0 +1,108 @@
...
+Configuration
+=============
+
+Go to *Settings -> Configuration -> Configure Backup* to
+create your configurations for each database that you needed
+to backups.
+
+Usage
+=====
+
```
lasley, line 27: We ran into an issue where backup execution was dying when running inside of Odoo due to the timeouts configured on the instance. We had to make a small script to run this system outside of the main Odoo instance to circumvent this, which is described [in this blog post](https://blog.laslabs.com/2016/10/running-python-scripts-within-odoos-environment/).

If this is something other people have run into as well, maybe transferring some of that info to this ReadMe would be of use?

max3903:
@lasley Why don't you add a link to your blog post in the module description (know issues section) ?

lasley:
Good call, that's exactly where it should go!

yajo:
Why did we lose commits?

lasley:
Squash and merge? The cherry-picked is definitely the commit for this on 9.0

yajo:
I think the common strategy is to preserve others' commits (except those from transbot, that get squashed) when migrating versions. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

It has the advantage that all previous contributors get subscribed to the PR automatically. They're probably interested.

lmignon:
@lasley Where is the commit histrory. We must preserve the history! The history is now empty also in 9.0.

lmignon (changes requested):
The commit history from 8.0 must be preserved

lasley:
Alright I did the required git-fu for fixing the commit history after merge of #620. Better?

lmignon (approved):
code review only

max3903 (approved):

PR #617, opened 2016-11-20
[9.0][ADD] server_environment_ir_config_parameter

sbidoul:


server_environment_ir_config_parameter/README.rst
```diff
@@ -0,0 +1,82 @@
...
+To configure this module, you need to add a section ``[ir.config_parameter]`` to
+you server_environment_files configurations, where the keys are the same
+as would normally be set in the Systems Parameter Odoo menu.
+
+When first using a value, the system will read it from the configuration file
+and override any value that would be present in the database, so the configuration
+file has precedence.
+
+The user cannot write, create, or delete System Parameters that are defined in the
+configuration files.
```
moylop260: What about use this configuration file similar to data xml file at start server?
I mean, odoo.conf to odoo.xml format (using StringIO) with `xml_id` and `noupdate="0"` to support changes and loading from [convert_xml_import](https://github.com/odoo/odoo/blob/6b55032e1ac86030fbdf2e057c71bbec0aea504a/openerp/tools/convert.py#L914)

moylop260: Here a example using [tools.convert_file](https://github.com/odoo/odoo/blob/6b55032e1ac86030fbdf2e057c71bbec0aea504a/addons/account_asset/tests/test_account_asset.py#L11-L13)

Here a example using directly [tools.convert_xml_import](https://github.com/OCA/l10n-spain/blob/741a47ac77b2f25386c3fa1a472ca2e51fb6b76b/l10n_es_toponyms/wizard/l10n_es_toponyms_wizard.py#L54)

We can use the last one but but with `StringIO` like as:

``` python
def serv_conf2xml(conf):
    # .... code here
    pass

fp_xml = StringIO(serv_conf2xml(serv_config.get('ir.config_parameter')))
tools.convert_xml_import(fp_xml, ...)
```

sbidoul: @moylop260 I use the mechanism that exists in the server_environment module and is very powerful.

moylop260: Both options use the mechanism that exists in the server_environment module and both are very powerful.

The disadvantage of a inherit `@ormcache('uid', 'key')` is that will be use 2 caches:
1. ormcache for you inherited method
2. ormcache for original one

But with the option of load parameters directly to database from configuration file you won't need add python logic because will use the same original powerful way and will work with new version of odoo (for 10.0 this section change).

By the way, your idea is very good (Thanks!)

lmignon: @moylop260 IMO one advantage of the conf file is that it's easily understood by system administrators and less errors prone than xml for this kind of audience. I like the idea of being able to display the environment ribbon just by specifying a value in my configuration file.

sbidoul:
@moylop260 actually I think I don't really understand the alternative you propose. Do you want to add XML support to server_environment?

server_environment_ir_config_parameter/README.rst
moylop260: @lmignon I like the configuration file too
I'm not talking about use a xml like as file input I'll explain it better (I hope) in the main description. (see https://github.com/OCA/server-tools/pull/617#issuecomment-261802331)

moylop260:
@sbidoul 
Ok, sorry for my bad explanation.
I'll try it again with a example.

# Input 

Configuration file with `[ir.config_parameter]` section like as:

``` conf
[wkhtml2pdf]
lib_path = /myHome/lib/wkhtmltopdf-linux-i386-0-9-9

[ir.config_parameter]
ircp_from_config=config_value
```

# Process

(see `new methods` section)

# Output

Database updated, inserting records to `ir_config_parameter` table with new parameters at start odoo server.

# New methods

``` python
def serv_conf_parameter2xml(conf_parameter)
    """:param conf_parameter: Server configuration of `ir.config_parameter`
    :returns: xml odoo data string like as the following example:
<?xml version='1.0' encoding='utf-8'?>
<openerp>
    <data noupdate="0">
        <record model="ir.config_parameter" id="record_id">
            <field name="key">{{ key_parameter }}</field>
            <field name="value">{{ value_parameter }}</field>
        </record>
        ...
        <!-- for this example will be: -->
        <record model="ir.config_parameter" id="ircp_from_config">
            <field name="key">ircp_from_config</field>
            <field name="value">config_value</field>
        </record>
        ...
        <record model="ir.config_parameter" id="record_id_N">
            <field name="key">{{ key_parameter_N }}</field>
            <field name="value">{{ value_parameter_N }}</field>
        </record>
    </data>
</openerp>
    """
    # ... Code using a template (maybe jinja) to auto-create this xml file based on configuration file.

def load_server_conf_parameter():
     # ...
     parameters_from_configuration_file = serv_conf['ir.config_parameter']
     xml_str = serv_conf_parameter2xml(parameters_from_configuration_file)
     fp_xml = StringIO.StringIO(xml_str)
     # This way update the database using xml_id engine to create.
     # This way will support updates using xml_id engine.
     # You won't need change or inherit the current python logic of `ir.config_parameter` model.
     # You won't need a duplicated validation because the database have a unique() constraint
     # You won't need problems managing `ormcached` for the inherited method and the original one
     # This way is compatible with all odoo versions
     tools.load_xml_import(fp_xml, ...)
```

sbidoul:
@moylop260 thanks for the clarification, I understand better now.

I can see some advantages to loading the values in the database at server startup.

The problem is that we need to load them with the right record ids if the parameters are also provided by data files in modules (like web_environment_ribbon does for instance), otherwise there will be duplicate key errors when loading the modules. So we'd need additional information in the config file to provide the record external ids. 

I'm also worried about the xml file not correctly enforcing the value from the config file, depending on the module loading order.

That's why I came up with this compromise of reading the config file in get_param.

In the meantime I pushed a workaround for the double ormcache issue you mentioned. I also changed the write/create behaviour to override with the config values instead of raising an error.

lmignon:
@moylop260 Thank you for the clarification. For sure  I misunderstood  your comment.

lmignon (approved):
Good idea! (Code review only)

moylop260:
I'm not sure... but I won't block this one...

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -0,0 +1,54 @@
...
+SECTION = 'ir.config_parameter'
+
+
+class IrConfigParameter(models.Model):
+
+    _inherit = 'ir.config_parameter'
+
+    def get_param(self, cr, uid, key, default=False, context=None):
+        value = super(IrConfigParameter, self).get_param(
+            cr, uid, key, default=None, context=context)
```
moylop260, line 19: `default=default`

moylop260 (changes requested):
I'm worried about a silently change of values without a warning.
It could be good add a flag to identify the parameters where is extracted from configuration file and set readonly the name and value field

server_environment_ir_config_parameter/models/ir_config_parameter.py
sbidoul: `default=None` is on purpose. The test for default value is at the end of the method.

sbidoul:
@moylop260 

> It could be good add a flag to identify the parameters where is extracted from configuration file and set readonly the name and value field

That's a good idea. But I'm a bit short on time at the moment. Would it be ok for you if I add that in the roadmap?

moylop260 (approved):

sbidoul:
Roadmap updated, waiting for #612 to check travis.

sbidoul:
Green after rebase. Good to go.

sbidoul:
@rousseldenis

guewen (approved):
Excellent! I had this in head since a while.

PR #624, opened 2016-11-28
[9.0] [MIG] html_image_url_extractor module

cubells:
* Updated README.rst

@Tecnativa


pedrobaeza (approved):

html_image_url_extractor/models/ir_fields_converter.py
```diff
@@ -0,0 +1,72 @@
...
+        :param str html_content:
+            HTML contents from where to extract the images.
+
+        :param int limit:
+            Only get up to this number of images.
+
+        :param bool fail:
+            If ``True``, exceptions will be raised.
+        """
+        # Parse HTML
```
lasley, line 30: Parsing HTML with Regex is never really the way to go, please see [this famous answer](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454) on SO for an amazing elaboration.

Why not just `xml.dom.minidom.parseString(html).getElementsByTagName('img')` + other logic?

pedrobaeza: This is only a migration of something done on v8. As this extract utility is very specific, I wouldn't touch this too much...

lasley (changes requested):
Thanks @cubells but I'm 👎  on this, please see my inline comment regarding the use of Regexes for this purpose.

html_image_url_extractor/models/ir_fields_converter.py
lasley: But when we bring something into OCA aren't we guaranteeing that it works by OCA standards? This definitely does not work in all instances - I've been burned by this before.

pedrobaeza: Why is not working? It's in v8 and used by many modules without problems

lasley: Alright wait maybe I'm misinterpreting something here. The regex is being applied against an element node, so we're good. My bad I misread there.

lasley (approved):
In light of inline convo, this was a stupid me situation. LGTM! 👍

PR #625, opened 2016-11-28
[9.0] [MIG] html_text module

cubells:
* Updated README.rst

@Tecnativa


pedrobaeza (approved):

html_text/models/ir_fields_converter.py
```diff
@@ -0,0 +1,73 @@
...
+        except (TypeError, etree.XMLSyntaxError, etree.ParserError):
+            if fail:
+                raise
+            else:
+                _logger.exception("Failure parsing this HTML:\n%s",
+                                  html_content)
+                return ""
+
+        # Get words
+        words = u"".join(doc.xpath("//text()")).split()
```
tarteo, line 54: Why do you join and split after? Isn't it possible to do this: doc.xpath("//text()").split() ?

tarteo (approved):
👍 Just a question

html_text/models/ir_fields_converter.py
pedrobaeza: This is because you can have several text elements. They are joined, and then split by words.

PR #628, opened 2016-11-30
[FIX][9.0] base_multi_image: Adhere to image delete bypass

lasley:
At long last I have figured out the bug plaguing OCA/product-attribute#199, and it is actually in the base module. The images were being unlinked during the variant addition due to the call to super in the `unlink` method. 

It seems there was an intent for a context override, but it never made it into this module. It has now!

* Add catch in owner unlink to allow for image delete bypass via context


pedrobaeza (approved):
I don't know why this was removed in 9.0 migration, as in 8.0 it's present in https://github.com/OCA/server-tools/blob/8.0/base_multi_image/models/owner.py#L96...

Anyway, let's fast-track this to have it working again.

sebalix:
Thank you @lasley !

PR #629, opened 2016-11-30
[FIX][10.0] base_multi_image: Adhere to image delete bypass

lasley:
This is a forward port of #628 & should probably not be reviewed until it is merged in order to not duplicate effort.

* Add catch in owner unlink to allow for image delete bypass via context

- [x] #628


pedrobaeza (approved):
Fast-tracked also

PR #634, opened 2016-12-05
[10.0][MIG] Module Prototyper

tarteo:


module_prototyper/__manifest__.py
```diff
@@ -22,8 +22,8 @@
 
 {
     'name': 'Module Prototyper',
-    'version': '9.0.0.1.0',
-    'author': 'Savoir-faire Linux, Odoo Community Association (OCA), Sudokeys',
+    'version': '10.0.0.1.0',
```
max3903: Should be `10.0.1.0.0`

max3903 (changes requested):
LGTM. No test.

max3903 (approved):

module_prototyper/data/README.rst
```diff
@@ -2,33 +2,37 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-===========
-Module name
-===========
+==============
+{module_title}
```
moylop260, line 6: Real name

module_prototyper/models/module_prototyper.py
```diff
@@ -291,11 +293,12 @@ def save_icon(self):
...
     @api.model
-    def generate_module_openerp_file_details(self):
+    def generate_module_openerp_file_details(self, api_version):
         """Wrapper to generate the __openerp__.py file of the module."""
+        fn_inc_ext = '{}.py'.format(api_version.manifest_file_name)
         return self.generate_file_details(
-            '__openerp__.py',
-            '__openerp__.py.template',
+            fn_inc_ext ,
+            '{}.template'.format(fn_inc_ext),
```
moylop260: `u'{}.template''.format...`
To avoid a ascii error similar to https://github.com/OCA/manufacture/issues/146

Or apply `%()s`:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms
`Prefer % over .format(), prefer %(varname) instead of positional. This is better for translation and clarity.`

module_prototyper/models/module_prototyper_api_version.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, fields
+
+
+class ModulePrototyperApiVersion(models.Model):
+    _name = 'module_prototyper.api_version'
+
+    name = fields.Char(
+        string='Name'
```
moylop260: `If the technical name of the field (the variable name) is the same to the string of the label, don't put string parameter for new API fields, because it's automatically taken. If your variable name contains "_" in the name, they are converted to spaces when creating the automatic string and each word is capitalized. (example: old api 'name': fields.char('Name', ...) new api 'name': fields.Char(...))`
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields

moylop260 (approved):
Just small changes

tarteo:
@moylop260 Thanks for your feedback, really helpful! But the first comment "Real name" I don't quite understand. Can you explain this to me, please?

module_prototyper/data/README.rst
sebalix: @moylop260 : this README is a template, no change is required here

moylop260: My bad
Sorry forget it

Issue #619, opened 2016-11-24
[9.0] auto_backup The commit history is lost in 9.0


lmignon:
We must preserve the commit history
https://github.com/OCA/server-tools/commits/8.0/auto_backup
https://github.com/OCA/server-tools/commits/9.0/auto_backup

pedrobaeza:
Closing as this has been solved.

antespi:
There is a problem when merging this addon in 9.0 branch, there is no auto_backup addons at all.
Can you review this?

Thanks all @lasley, @pedrobaeza, @lmignon

lasley:
What the heck - that PR (#620) has no files changed in it anymore. I can guarantee that wasn't the case before though because #614 is based on it. 

I'm not sure what happened here guys. I guess I can resubmit the PR, but what about all the commits that made it on but I guess are blank?

antespi:
ok, @lasley, you can resumit the PR in order to see if that fix it

lasley:
What about the 54 commits already on 9.0 branch that are seemingly empty?

lasley:
Closing, this was resolved in #636

PR #620, opened 2016-11-24
[MIG][9.0] Redo auto_backup 

lasley:
Reverts and reapplies #526. Fixes #619

PR #636, opened 2016-12-07
[FIX][9.0] auto_backup: Missing files

lasley:
This reverts commit 0111be2e599efb616170b4586bde80a73b12ba2c, which was committed directly to 9.0 & pulled the files from #620 that contained an identical commit 8a078fc4ec9df1993fcb4d498d0a4508eb7a5ecf.

This is the cleanest way that I see to resolve https://github.com/OCA/server-tools/issues/619#issuecomment-265404988. 

The alternative is another migration, which will add another 50 some odd commits.

Anyone have thoughts?

- [x] #614


antespi (approved):
LGTM :+1:

lasley:
Dammit I just realized with this then we'll circle back around to the original #620. I guess re-migrating the module will be the best solution in hindsight. I'll just backport from my 10.0 #614 once merged.

antespi:
Ok, thanks

lasley:
I'm going to look at this in the next few days. I'm in conflict ~~heaven~~ hell in any sort of migration direction due to the translation files being on 9 branch. 

I guess I'm just going to start again, delete the migrations, then re-upgrade this module to 9. I really only needed it in 10 though, so bottom of my work stack TBH seeing as this is will be my third time migrating it to a version I don't need.

lasley:
Ok I think I got this to the state it was on merge on 9. There's this dirty hack I had to do in 8c45b49 to delete the translations that are already there so that the commits could be walked back onto the branch. 

It seems there are some missing translation files in this PR though, but everything that I do ends up with them not there. I'm hoping maybe that will correct itself when Transifex tries to update the branch?

antespi:
Thanks @lasley :+1:

antespi:
Password for runbot: Admin:1234567890

susport:
@antespi  With that password I can not authenticate in runbot

lasley:
Username is `admin` Password is `Admin:1234567890`. Hah this repo is fun now that `password_security` is a thing

antespi:
@lasley thanks for your clarification and sorry @susport for the confusion.
Now, OCA/server-tools runbots have this issue

lasley:
I'm probably going to provide a fix for that soon. Something to disable the security on the admin user unless an option is explicitly enabled I'm thinking

lasley:
I'm going to go ahead and merge this. It's pretty broken on 9.0 branch as it is & this was all approved a bunch of times, so I don't see the harm. Worse case scenario it's still broken, right? 😛

PR #325, opened 2016-01-06
mass_editing: Avoid errors if 'model_id' is not a list of ids.

hpestano:
When I try to make this query

```
@api.model
def get_default_fields(self):
    field_pool = self.env['ir.model.fields']
    fields = ['name', 'street', 'zip', 'city', 'email']
    ids = [field_pool.search([('model_id', '=', "'res.partner'"), ('name', '=', f)]).id for f in fields]
    return ids
```

I get this error

```
ValueError: invalid literal for int() with base 10: 'es.partne'
```

That is why i suggest this change

PR #638, opened 2016-12-08
mass_editing: Fix issue with ir.model.fields implementation of search

mvaled:
Without 'mass_editing' installed you can search the 'ir.model.fields' like this:

    ir_model_fields.search([('model_id', '=', 'res.partner')])

and it works.

However mass_editing changes the 'search' method and broke this type of query.  This commit fixes the 'search' method so that it allows the legacy behavior.

Closes #325 for a new PR.


mvaled:
I kept digging about the usage of 'model_id' to test a string.  In all our codebase I found two usages:

- One in an old (now unused) module from Julius Networks.
- A single addon of one of our customer.

Can I safely switch the 'model_id' for 'model'?  I found only two cases where model_id.model is not the same as model:

```
SELECT DISTINCT imf.model, m.model, imf.model = m.model
       FROM ir_model_fields imf JOIN ir_model m ON imf.model_id = m.id WHERE imf.model <> m.model;
         model          |         model         | ?column? 
------------------------+-----------------------+----------
 crm.case.resource.type | crm.tracking.campaign | f
 crm.case.channel       | crm.tracking.medium   | f
```

But I wonder if that's an error.

StefanRijnhart:
Thank you for taking up this one! This hack in the model's search method allowed for a domain containing a stringified list of ids. It has been obsolete since https://github.com/OCA/server-tools/commit/4442424, and can safely be removed (i.e. remove the whole file).

mvaled:
Hi @StefanRijnhart, I'll move your suggestion to the PR #638 which is the one standing for this issue.

mvaled:
Opps, I didn't notice this is the PR...

mvaled:
Done in 09732a89

StefanRijnhart (approved):
Thanks! Tested on runbot.

pedrobaeza (approved):

PR #640, opened 2016-12-09
[10.0] [MIG] Migration of server_environment_ir_config_parameter

rousseldenis:


sbidoul:
@rousseldenis can you please forward port this too https://github.com/OCA/server-tools/pull/676 ?

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -51,4 +51,4 @@ def write(self, vals):
                 newvals = dict(vals, value=serv_config.get(SECTION, key))
             else:
                 newvals = vals
-            super(IrConfigParameter, rec).write(newvals)
+        return super(IrConfigParameter, rec).write(newvals)
```
sbidoul: @rousseldenis this does not look right. The write must be inside the loop as we overwrite whatever is provided in vals with the values found in the config file, so each property may end up with a different value. So this commit changes the expected behaviour.

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -51,4 +51,5 @@ def write(self, vals):
                 newvals = dict(vals, value=serv_config.get(SECTION, key))
             else:
                 newvals = vals
-        return super(IrConfigParameter, rec).write(newvals)
+            super(IrConfigParameter, rec).write(newvals)
+        return True
```
sbidoul: Ah yes, fixed in next commit, okay.

lmignon:
This one depends of #595

lmignon:
@sbidoul #676 forward ported

lmignon (approved):
Code review + test

sbidoul:
rebased

sbidoul:
This one is now finished and migrated to new API.

simahawk (approved):

leemannd (approved):
Code review and tested. Thank you! :+1:

atchuthan (approved):
code review :+1:

guewen (approved):
Nice!

PR #641, opened 2016-12-09
[MIG] 9.0 - Migrate 'base_user_role'

sebalix:
@lasley here it is


base_user_role/data/ir_cron.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 ABF OSIELL <http://osiell.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
```
lasley: odoo noupdate="1"

base_user_role/migrations/8.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
lasley: Past version migration can be removed

lasley (approved):
Thanks @sebalix

base_user_role/migrations/8.0.1.1.0/post-migration.py
sebalix: Fixed

base_user_role/data/ir_cron.xml
sebalix: Fixed

base_user_role/views/user.xml
```diff
@@ -0,0 +1,37 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+<record id="view_res_users_form_inherit" model="ir.ui.view">
+    <field name="name">res.users.form.inherit</field>
+    <field name="model">res.users</field>
+    <field name="inherit_id" ref="base.view_users_form"/>
+    <field name="arch" type="xml">
+        <notebook position="inside">
+            <page string="Roles">
```
sbidoul, line 12: What would you think about putting the Roles as first page in the notebook, ie before the standards groups page. As admins using this module will almost never look at the detailed groups, it's probably better to show roles first.

base_user_role/models/role.py
```diff
@@ -0,0 +1,83 @@
...
+    _inherits = {'res.groups': 'group_id'}
+    _description = "User role"
+
+    group_id = fields.Many2one(
+        'res.groups', required=True, ondelete='cascade',
+        readonly=True, string=u"Associated group")
+    line_ids = fields.One2many(
+        'res.users.role.line', 'role_id', string=u"Users")
+    user_ids = fields.One2many(
+        'res.users', string=u"Users", compute='_compute_user_ids')
```
sbidoul, line 25: Since roles are groups too, it would be nice to put them in their own category so they are more easily distinguished from regular groups in the groups configuration page.

sbidoul (approved):
@sebalix excellent module! I made a couple of ergonomy suggestions but as this is a migration PR, :+1:

base_user_role/models/role.py
sebalix: This is exactly how we do when declaring our roles as XML records, it could be done by default by associating them to a 'User roles' category indeed.

sebalix: @sbidoul Done

base_user_role/models/role.py
```diff
@@ -23,6 +23,10 @@ class ResUsersRole(models.Model):
         'res.users.role.line', 'role_id', string=u"Users")
     user_ids = fields.One2many(
         'res.users', string=u"Users", compute='_compute_user_ids')
+    group_category_id = fields.Many2one(
+        related='group_id.category_id',
+        default=lambda cls: cls.env.ref(
+            'base_user_role.ir_module_category_role').id)
```
sbidoul, line 29: Cool, thanks for adding this. Would it work too with a simple override ```category_id = fields.Many2one(default=...)```?

sebalix: It does not work without the `related` part. About the field's name, I don't know what is best.

sbidoul: so weird to me... anyway it works fine, thanks!

sbidoul (approved):

sebalix:
2 thumbs up and 5 days, ready to merge I think ;) I'll start the migration to v10 soon.

pedrobaeza:
Can you please squash your fixes in the review process commits?

sebalix:
@pedrobaeza Done, I kept the commits from @sbidoul and the last one related to the default module category as they are new features which could be backported on the 8.0 version.

pedrobaeza:
Thanks for the comprehensive squashing! I'll merge when CIs finish.

Issue #622, opened 2016-11-25
Manage external accounts in Odoo:


mourad-ehm:
After reflection with the akretion team, we propose the following
## **Context :**
Odoo is dealing to communicate with several external services (ie 3thd party APIs).
Credentials needs to be stored securely and accessible in a clearly defined manner. 
Currently each module implement this feature differently : it's a bad practice, error prone and waste of time to reimplement it.
We propose a specification of a generic module to manage accounts.

**## Solution proposed :**

### Features:

- store account credentials
- multiple account per provider
- access restriction per society and/or per security group
- store aditionnal data (config)
- ensure data security by encrypting the identification data.

### Design:
1 - store credential information in table with few columns (module_name, company_id, password, data). data will be stored in json and contain additionnal info ie {'login':'', 'key1': 'value1', 'key2': 'value2'},
2- encrypting key will be stored in config file. Potentially reduce damage if a database is leaked.
3- provide a method that allows to provides identification data (decrypting) (create a abstract object that can be inherited by business object ex stock.picking).
5 - restrict read/write by default to admin and a group "external account managers”.
6 - each consumer modules provides a dict for registration : a dict of expected info

** Example of table :**
| name            |       type      | identifier                       | company_id |  login     | password | json data                                                                       | 
----------------------------------------------------------------
| Laposte        |        lapost   | laposte_warehouse1 | 1                    | 122732| ******         | {'customerId': 51221, 'agencyCode: 'paris0701'} |
| Laposte        |        lapost   | laposte_warehouse2 | 2                    | 122732| ******         | {'customerId': 53232, 'agencyCode: 'nice1323'} |
| Service aws |     aws         | s3_storage                   | 1                    | 45877  | ******         | {'preferedRegion':'EU'} |

**Management of multiple accounts by service provider :**

In certain use case : for their shipping, the company can use the same authentification (login/password) with different customer number. For example customer number for parent company and an other customer number for subsidiary. Or customer number by white brand ext. 
the sub-account use used for either statical or  accounting reasons (ex: Invoice parent company and subsidiary separately) 

** There 2 approaches to manage several accounts by service provider:**
1 – duplicate the account and modify only  the json column (customernumber, my_other_param', ..)
2 - Split the table in to object account and account line : 
* Account contain fields Name, account type, Company,  login, password
*Account line contain : account_id, key, value (json)

### Use Case:
- carriers: ups, laposte, geodis
- marketplace: ebay, amazon, magento, ...
- ftp/sftp

[...]

Issue #471, opened 2016-07-01
RFC: Encrypted Fields Module


lasley:
In some secure contexts, at-rest encryption is required at the database layer. My specific context being Protected Health Information covered under HIPAA. Some PCI data (Cardholder name, etc) is also required to be encrypted at rest.

I made a crude module a while back in order to meet this need - https://github.com/laslabs/odoo-base/blob/9.0/fields_encrypted/fields.py

It has worked great in some long term tests so we are about to begin the build out of the real module. Assuming this is something that OCA wants, I will bring it here instead.

I'm probably forgetting some details, but from a high level this is what I am thinking. Please let me know if you have any suggestions or if there is some overlap with pre-existing resources.

Currently the module:
- Provides `EncryptedText` field w/ encrypt and decrypt in what seem to be the appropriate convert methods
- Has a hard coded key - obviously bad
- Will basically thrash your data if you ever change the hard coded key - definitely worse

The plan:
- Move the key into a secure storage
  - I'm thinking this will just be a file that gets loaded into memory on Odoo boot
- Switch to PgCrypto instead of encrypting in Python
- It would be nice if I could add an `encrypt` arg in all fields, instead of a specific field for encryption. This would allow for usage on existing columns.
  - Assuming this is possible, some sort of migration up and down would be required in order to circumvent data destruction
- Possibility of specifying encryption key to use on column. By name? Or maybe the signature..

Pitfalls:
- Searching an encrypted col is a challenge. Doesn't hurt for anything I can think of this being used for, but worth mentioning.

hbrunn:
I'm toying with something like that too for a while: https://github.com/hbrunn/server-tools/tree/8.0-base_encrypted_field/base_encrypted_field
My approach is to have a group of records share a symmetric encryption key that is encrypted with a public key generated for every user. Also the architecture tries hard to have all crypto being done on the client side so that the key never leaves the client's computer.

lasley:
I like what you have here, particularly the client-side aspect & PGP vs AES. My implementation's security depends on separation of server infrastructure, and still has a weak point of exploitation from the Odoo filesystem with an attack surface as wide as the application's keys.

Am I correct in reading that this is storing the public and private keys in the Odoo database? I am trying to ward against an attack at the database layer, so anything stored in that becomes a weak point in the event of a DB dump.

[...]

PR #644, opened 2016-12-12
Keychain: account manager for external systems

hparfr:
See readme for more info


lasley:
This is very dangerous IMO & I am explicitly 👎 

This is enforcing bad practice and the entire keychain can be easily compromised in the UI with code similar to the following:

```
model = env['keychain.account']

log(
    str(env['keychain.account'].search([]).read()),
)
```

Given the current security, it wouldn't even require an admin to run this exploit.

There is also no way to expire the existing key in the event of compromise, which is a 100% necessity with anything encryption.

I'm sorry that I am not providing anything helpful to this PR, but this is because [we have been toiling over this for a while](https://github.com/OCA/server-tools/issues/471) & have not been able to come up with a production viable solution at this point. 

I would not trust my data with either this module or the PoCs we have already built for this purpose, thus the 👎 . Security is no joke, and a simple disclaimer is not enough IMO.

hparfr:
Again, the aim of this module is to store credentials like smtp server, carriers, market places, magento, ftp servers and so on and don't leak them with db dumps or search_read. 

It's credentials you need in odoo and it's credentials which are currently stored in plain text by the modules and are easily recoverable with your snippet.

With keychain module, at runtime you need to do 2 actions to decrypt a password : 
- get the record from the ORM (which is filtered with odoo's ACL)
- get the key from the config manager (which is only accessible by python)

From a security point of view : the current state with odoo out of the box is very terrible. With this keychain module, it's certainly not perfect but a little more secure.


> This is enforcing bad practice 

Which one ?

> and the entire keychain can be easily compromised in the UI with code similar to the following:

No with keychain supported modules you will get : 
- if you are not in authorized group  : access_error 
- else : an encrypted password

With other modules, you will get : 
- if you are not in authorized group  : access_error 
- else : a clear text password

>There is also no way to expire the existing key in the event of compromise, which is a 100% necessity with anything encryption.

In the event of a compromise you have to change all your passwords.
If you want key expiration or x509 certificates your PR is welcome :) I hope the current design of this module allows this kind of enhancement.

>but this is because we have been toiling over this for a while & have not been able to come up with a production viable solution at this point.

It seams your requirements (storing credit card or personal / medical data) are very different.
You need conform to norms (like PCI-DSS) and your responsibility may be engaged if you leak this kind of data : this is not the aim of this module.

hparfr:
Responding to comments on https://github.com/OCA/server-tools/issues/622#issuecomment-266999694

>This module will be available on the App Store. I can guarantee that it will be installed on production servers by administrators that did not read the disclaimer in the ReadMe.

This module is unusable / useless if no modules consumes his services. I hope, the developer who add this module as a dependecy of his own module will read the readme. If he don't  (which is plausible) we can't do anything about it.

>I do also like the idea of a separate service. I created a Golang app recently to do some key management & other useful things, and have been thinking of integrating it into Clouder

It's a good idea to externalize the storage of sensitive data, but it may be overkill to store smtp credentials or delivery carrier accounts.

rvalyi:
@hparfr @lasley I won't comment about the module because I'm not a security expert. That being said, usually a good way to inject secrets in an app is via environment variables.

These can be managed with tools such https://www.vaultproject.io if something more high level than editing these secrets over ssh is required.

lasley (changes requested):
@rvalyi - Ohhhhh I like that. I'm going to see about making a connector for VaultProject after I play around with it a bit more.

@hparfr - I agree on the need, but this is adding a tad bit of security and also the possibility of mass data corruption if the key needs to be revoked. The tradeoff is more dangerous than it is worth IMO.

Obviously the community can override my feelings, but they are firm on this. My project was doing nearly the exact same things you are here & I abandoned it because it is flawed from the ground up.

I recommend reading into OWASP guidelines on these subjects. They explain it all much better then me, and you will notice that nearly nothing of the guidelines is covered. 

Again, I agree it isn't covered in Odoo either, but you are specifically advertising this module as a `keychain`. That name implicitly states you are managing the keys in a safe fashion.

* https://www.owasp.org/index.php/Key_Management_Cheat_Sheet
* https://www.owasp.org/index.php/Cryptographic_Storage_Cheat_Sheet
* https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet

hparfr:
Thanks @lasley for having taking time to expose your arguments.

lasley:
@hparfr - I honestly feel real bad about blocking this & am trying to come up with a viable solution for the key management + a proposal to make this more secure. I'll update you.

bealdav:
Hi @lasley,

Ok there are more secure ways with external and complex system, but our all customers can't pay to have an erp (aka all in one features system) and 36 connectors with external system to maintain over the time.
It depends of criticity of these keys

@hparfr proposal use a secure way, I think

>     Fernet is built on top of a number of standard cryptographic primitives. Specifically it uses:
>
>        AES in CBC mode with a 128-bit key for encryption; using PKCS7 padding.
>        HMAC using SHA256 for authentication.
>        Initialization vectors are generated using os.urandom().

It ensures all password are not clear text.

It's a v9 module. If in v10 or v11 you provide a better way to do it, then we'll migrate to the new solution, but please don't break continuous improvement of the community, it's a step in the right direction

https://en.wikipedia.org/wiki/Continual_improvement_process

:+1:

lasley:
@bealdav - This isn't simply about security. This is also about potential data destruction with no avenue of resolve. Encryption isn't just flipping a few bits, then flipping them back with the appropriate key when needed - you also need to build a way for re-encryption of data against new keys & the revocation of old. This is standard in any application that encrypts anything, and any product without it is dangerous. Continuous improvement does not help the user of the application in the middle of a breach scenario that is faced with the decision of having to destroy all their encrypted data in order to revoke a key. Revocation is not a feature of an encryption system, it is a requirement.

bealdav:
@lasley If security was the main point, you'd haven't choose odoo at all, don't you think ?
Here it's better secure than odoo standard practice. Nowdays any module store its password with no rule, some time in clear text.
Rome was not built in one day and here it's v9, an old version, now.
Improvement will comes after in v10
You can propose a feature to provide revocation, no problem.
@hparfr could you consider to add revocation in roadmap ?

@lasley Thanks for your contribution on security subjects. It's a real important for us to have expert like you on this matter, just let community go step by step to real good practice as you suggest.

legalsylvain:
Hi,
I'm not expert in the security domain, but I find the design of this module very good. To steal uncrypted credentials you have to have access to the database AND the config openerp file. So I see it as something secured.
Maybe there are other possibilities, but for the time being, this is a good solution, and **better than nothing**, specially because there are actually some OCA / not OCA modules with clear credentials (not secure), or others with clear token (not very secure) so :+1: for the concept.

 @bealdav said :
> It's a v9 module. If in v10 or v11 you provide a better way to do it, then we'll migrate to the new solution, but please don't break continuous improvement of the community, it's a step in the right direction

:+1: on that point, too.

@OCA/board , @StefanRijnhart, @others, could you give your PoV ?

kind regards.

bealdav:
ping @guewen @yvaucher

mourad-ehm:
:+1:  because this module make one step in the right direction. It brings a solution to managing credentials for external applications which not exist now in odoo.

sebastienbeau:
Hi, after reading twice the comment of @lasley, I also think that we will need a process for updating the key (we should be able to unencrypt the data and encrypt it again with the new key). But I do not think we should stop this PR for that. We should add this feature in the README (in the roadmap, @hparfr  can you add it?), so everyone know that this feature is missing, and we can add it later. In any case if we want to change the key an administrator can still do it by changing it manually (extracting password, change the key and setting back the password).
So :+1:

bguillot:
I agree with @legalsylvain @bealdav @mourad-ehm 

This can be improved but we need this improvement to have something to manage credentials because this is better than what we have right now.

:+1:

elicoidal:
I agree with @lasley and his feedback about revocation. This is a must.
Nevertheless I would not block this PR which provides an additional step toward better password management.
I wish we would have a core and complete method to encrypt password but until then :+1:

lasley:
Everyone tossing thumbs on this - please use the Github review feature. Otherwise, it is likely your vote will not be tallied.

bealdav:
Error on some other modules: html_image_url_extractor, server_environment, 
not keychain.

bealdav:
ping @mourad-ehm @bguillot @sebastienbeau @elicoidal @legalsylvain @rvalyi  see lasley comment for thumbs on top of the PR

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+This module allows you to store credentials of external systems.
+
+* All the crendentials are stored in one place: easier to manage and to audit.
+* Multi-account made possible without effort.
+* Store additionnal data for each account. 
+* Validation rules for additional data.
+* Have different account for different environments (prod / test / env / etc).
+
+
+By default passwords are encrypted with a key stored in odoo config.
```
elicoidal: add a comma after default
s/odoo config/Odoo config file

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+It's far from an ideal password storage setup, but it's way better 
+than password in clear text in the database.
+It can be easily replaced with another system. See "Security" chapter below.
```
elicoidal: s/with/by (?)

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+Accounts may be: market places (Amazon, Cdisount, ...), carriers (Laposte, UPS, ...)
```
elicoidal: Cdiscount

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+* Have different account for different environments (prod / test / env / etc).
+
+
+By default passwords are encrypted with a key stored in odoo config.
+It's far from an ideal password storage setup, but it's way better 
+than password in clear text in the database.
+It can be easily replaced with another system. See "Security" chapter below.
+
+Accounts may be: market places (Amazon, Cdisount, ...), carriers (Laposte, UPS, ...) 
+or any third party system called from odoo.
```
elicoidal: Odoo

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+This module is usefull for developers.
```
elicoidal: aimed

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+The logic to choose between accounts is done in dependant modules.
```
elicoidal: "will be achieved in dependent modules"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+
+==========
+Uses cases
+==========
+
+For delivery, each warehouse send parcels with his own account to the same carrier.
```
elicoidal: I am not sure to understand this sentence in the usage

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+
+Configuration
+=============
+
+After the installation of this module, you need to add some entries in odoo's config file :
```
elicoidal: remove column at the end of line (duplicated)

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+Configuration
+=============
+
+After the installation of this module, you need to add some entries in odoo's config file : 
+(etc/openerp.cfg) :
+
+> keychain_key = fyeMIx9XVPBBky5XZeLDxVc9dFKy7Uzas3AoyMarHPA=
+
+You can generate keys with `python keychain/bin/generate_key.py`
```
elicoidal: add a period

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+* Implement validation of user entered data:
+
+.. code:: python
+
+    class LaposteAccount(models.Model):
+        # ...
+        def _roulier_laposte_validate_data(self, data):
+            return len(data.get("agencyCode") > 3)
+
+* In your code, fetch the account :
```
elicoidal: remove space before ":"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+            'password': account.get_password()
+        }
+
+
+In this example, an account is randomly picked. Usually this is set according 
+to rules specific for each client.
+
+You have to restrict user access of your methods with suspend_security().
+
+Warning: _init_data and _validate_data should be prefixed with your namespace !
```
elicoidal: remove space before "!"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+Choose python naming function compatible name.
+
+Switching from prod to dev
+==========================
+
+You may adopt one of the following stragegies:
```
elicoidal: "strategies"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+* store your dev accounts in production db using the dev key
+* import your dev accounts with odoo builtin methods like a data.xml (in a dedicated module).
+* import your dev accounts with your own migration/cleanup script
+* etc.
+
+Note: only the password field is unreadable without the right key, login and data fields
```
elicoidal: s/right/proper (?)

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+between accounts.
+
+Usage (for user)
+================
+
+Go to *settings / keychain*, create a record with the following 
+
+* Namespace: type of account (ie: Laposte)
+* Name : human readable label "Warehouse 1"
+* Technical Name: name used by a consumer module (like "wharehouse_1")
```
elicoidal: "warehouse_1"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+* Login: login of the account
+* Password_clear : For entering the password in clear text (not stored unecrypted)
```
elicoidal: "uncrypted" or "unencrypted"?

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+* Password : password encrypted, unreadable without the key (in config)
+* data: a JSON string for additionnal values (additionnal config for the account, like : `{"agencyCode": "Lyon", "insuranceLevel": "R1"})`
```
elicoidal: s/data/Data

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+* Environment : usually prod or dev or blank (for all)
+
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/{repo_id}/{branch}
```
elicoidal: update the target link and remove the comments that follows

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
+
+Known issues / Roadmap
+======================
+- Account inheritence is not supported out of the box (like define common settings for all environments)
```
elicoidal: In general try to use lines up to 80/100 characters to improve readability (non-blocking / nice to have)

elicoidal: out-of-the-box
s/define/defining?

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+- Adapted to work with `server_environnement` modules
+- Key expiration
+- Import passwords from data.xml
+
+Security
+========
+Common sense : Odoo is not a safe place for storing sensitive data. But sometimes you don't have much other possibilities. This module is designed to store credentials of things like carrier account, smtp, api keys... but definitively not for credits cards number, medical records et alia.
```
elicoidal: s/much/any

elicoidal: s/things/data.
Not sure what is "alia"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+
+By default, passwords are stored encrypted in the db using symetric encryption [Fernet : https://cryptography.io/en/latest/fernet/]. The encryption key is stored in openerp.tools.config.
+
+Threats even with this module installed :
+
+- unauthorized odoo user want to access data: access is rejected by odoo security rules
```
elicoidal: "Odoo"

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+- authorized odoo user try to access data with rpc api : he get the passwords encrypted, he can't recover because the key and the decrypted password are not exposed through rpc
```
elicoidal: s/get/gets

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+If your dev server is compromised, hacker can't decrypt your prod passwords since you have different keys between dev and prod.
+
+If you want something more secure : don't store any sensitive data in odoo, use an external system as a proxy, you can still use this module for storing all other data related to your accounts.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
elicoidal: please update target link

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+
+Credits
+=======
+
+Akretion
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
elicoidal: if you do not use the logo, no need for this line

keychain/README.rst
```diff
@@ -0,0 +1,232 @@
...
+------------
+
+* Raphaël Reverdy <raphael.reverdy@akretion.com>
+
+Funders
+-------
+
+The development of this module has been financially supported by:
+
+* Akretion
```
elicoidal: you might want to add a URL

keychain/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+    "category": "Uncategorized",
+    "website": "https://akretion.com/",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [
+            'cryptography'],
+        "bin": [],
```
elicoidal: remove empty keys

keychain/models/keychain.py
```diff
@@ -0,0 +1,201 @@
...
+        for rec in self:
+            rec.password = rec._encode_password(
+                rec.clear_password, rec.environment)
+
+    @api.model
+    def retrieve(self, domain):
+        """Search accounts for a given domain.
+
+        Environment is added by this function.
+        Use this instead of search() to benift from environment filtering.
```
elicoidal: benefit

keychain/models/keychain.py
```diff
@@ -0,0 +1,201 @@
...
+    @staticmethod
+    def _serialize_data(data):
+        return json.dumps(data)
+
+    @staticmethod
+    def _parse_data(data):
+        try:
+            return json.loads(data)
+        except ValueError:
+            raise ValidationError(_("Data not valid JSON"))
```
elicoidal: "Data contains invalid JSON format"

keychain/models/keychain.py
```diff
@@ -0,0 +1,201 @@
...
+                "key. Unless you can recover the previous key, "
+                "this password is unreadable."
+            ))
+
+    @classmethod
+    def _get_cipher(cls, force_env=None):
+        """Return a cipher using the keys of environments.
+
+        force_env = name of the env key.
+        Usefull for encoding against one precise env
```
elicoidal: "useful"

keychain/models/keychain.py
```diff
@@ -0,0 +1,201 @@
...
+                if key and len(key) > 0  # remove False values
+            ]
+
+        if force_env:
+            envs = [force_env]
+        else:
+            envs = cls._retrieve_env()  # ex: ('dev', False)
+        keys = _get_keys(envs)
+        if len(keys) == 0:
+            _logger.info('on declanche un warning')
```
elicoidal: "Triggering a warning"

keychain/models/keychain.py
```diff
@@ -0,0 +1,201 @@
...
+            raise Warning(_(
+                "No 'keychain_key_%s' entries found in config file. "
+                "Use a key like : %s" % (envs[0], Fernet.generate_key())
```
elicoidal: s/like :/similar to:

elicoidal (changes requested):
Mostly README review
No code review

hparfr:
Thanks @elicoidal for your precise review.

hparfr:
Requested changes taken in account.
This PR is now ready to be merged.

Thanks everyone for your comments and reviews.

hparfr:
What may I do for this PR to be accepted ?
* fixed requested changes
* tests ok

bealdav:
ping @elicoidal @lasley

lasley:
Sorry but I cannot sign off with this potential for data loss. A PSC will need to dismiss my review for merge. There's definitely enough thumbs though and it sounds like @elicoidal was onboard with the risk, so you should be good here soon.

elicoidal (approved):

elicoidal:
Can you check travis?
Can you add in the readme a disclaimer/warning with a link to this discussion?

hparfr:
Travis is failing on other modules when database_cleanup is included
https://travis-ci.org/OCA/server-tools/builds/199916286

lasley:
@hparfr - The build was fixed in #719, so you should be able to clear the issue here with a rebase on the current 9 🚀

PR #642, opened 2016-12-10
[IMP][9.0] base_external_dbsource: Refactor & Split by source

lasley:
This is a refactoring of the `base_external_dbsource` module, hopefully allowing for easier addition and more isolation of the independent data sources.

Note that all adapter modules are `auto_install` to `True` in order to not break existing installations while still allowing for the modules to be split logically. In v11 we can then add the proper Python depends, which are already in the manifests but commented out.

Due to the above note, I also had to submit all of the modules in one PR. Separating them will break things horribly. Sorry for the gigantic PR

Depends:
- [x] #637

PR #623, opened 2016-11-25
[10.0] Add Firebird database support to base_external_dbsource

mstuttgart:
* Add Firebird database support to `base_external_dbsource` module using the `fdb` python package https://pypi.python.org/pypi/fdb/
* Code adjust and update README.rst from module.

PR #578, opened 2016-10-19
[MIG][10.0] base_external_dbsource

jlzhou:
Adapted to 10.0
#565

PR #659, opened 2016-12-20
[10.0][IMP] base_external_dbsource: Refactor & Split by source

lasley:
This is the v10 forward port of #642, which heavily refactors `base_external_dbsource` for modularity & splits the sources into independent modules.

Note that all adapter modules are auto_install to True in order to not break existing installations while still allowing for the modules to be split logically. In v11 we can then add the proper Python depends, which are already in the manifests but commented out.

Due to the above note, I also had to submit all of the modules in one PR. Separating them will break things horribly. Sorry for the gigantic PR

Depends:
- [x] #658


pedrobaeza:
But what about https://github.com/OCA/server-tools/pull/578?

lasley:
@pedrobaeza That migration is considered in this. Sorry I should have said this is the Forward port of #642, it's a feature port not a v9=>10 migration. Will update my message.

dreispt:
Please rebase.

lasley:
Rebased!

aldo774:
LGTM :+1:

base_external_dbsource_firebird/models/base_external_dbsource.py
```diff
@@ -0,0 +1,56 @@
...
+                continue
+            kwargs[key.lower()] = value
+        return fdb.connect(**kwargs)
+
+    @api.multi
+    def execute_fdb(self, sqlquery, sqlparams, metadata):
+        with self.connection_open_fdb() as conn:
+            cur = conn.cursor()
+            cur.execute(sqlquery % sqlparams)
+            rows = cur.fetchall()
```
dreispt, line 55: Improvement idea: It would be nice if we could return an iterable, wither here or on a specific method for that.

dreispt (approved):
Excellent work! Several notches above the original implementation.

base_external_dbsource_firebird/models/base_external_dbsource.py
lasley: I'm good with that - it would be fairly easy too, because `execute` conforms to the iterator protocol.

Downside, which isn't present in this FDB implementation, is that the meta couldn't be returned. TBH I don't see a use-case for the meta, so again I'm good with the change, but I'm sure others have a use case maybe?

It would be nice if there were a way to get the meta from the query without actually grabbing the first row off the stack. I'm not sure how to do that though.

dreispt: It's just an idea for future improvement; IMO it's best to go in small steps, one at a time.

lasley: I like it, so I'll add to Roadmap. I think it may be kind of difficult to make it work in a guaranteed backwards compatible manner, so we'll probably see it in v11.

base_external_dbsource_firebird/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Credits
+=======
+
+Contributors
+------------
+
+* Daniel Reis <dreis.pt@hotmail.com>
+* Maxime Chambreuil <maxime.chambreuil@savoirfairelinux.com>
+* Gervais Naoussi <gervaisnaoussi@gmail.com>
+* Dave Lasley <dave@laslabs.com>
```
mstuttgart, line 62: @lasley Please, add me as Contributors. The firebird support was provide by me.
https://github.com/OCA/server-tools/pull/623

base_external_dbsource/README.rst
```diff
@@ -67,6 +71,7 @@ Contributors
 * Maxime Chambreuil <maxime.chambreuil@savoirfairelinux.com>
 * Gervais Naoussi <gervaisnaoussi@gmail.com>
 * Michell Stuttgart <michellstut@gmail.com>
```
mstuttgart, line 73: @lasley please, remove me of contributors. I only provide firebird support. I din't work in this module core code.

mstuttgart:
@lasley I am contributors only Firebird feature. Please add me as contributor in *base_external_dbsource_firebird* and remove me from *base_external_dbsource*

lasley:
Noted contributors issues fixed in #796

mstuttgart:
Yes, i open it. Thanks!

Issue #633, opened 2016-12-02
[RFC] base_elasticsearch


lasley:
This module will provide a connection to an Elasticsearch cluster and underlying nodes via the `elasticsearch-py` module. A requirement of this module is LGPL licensing, which rules out the use of the `connector` library.

It will allow for CRUD on Elasticsearch indices & documents. The goal is to not persist anything beyond the cluster connection data and node hostnames in the Odoo database as part of this module. Depending on architectural limitations, I may also have to store the indices and document types - but maybe in a TransientModel so they'll get vacuumed.

The only views provided will be for setup and maintenance of the cluster/node meta.

It would be nice if I can seamlessly emulate an Odoo Recordset, but with Elasticsearch data and without persisting any of the data. This may be possible with some creative method overloading.

If the Recordset emulation is possible, it would also be nice if some abstract Document view was provided. While it's not really something that is required, it would be nice to directly query and view the Elastic results Odoo sees when in a pinch.

I'm wondering if anyone has pursued anything like this, particularly with the point about seamlessly emulating a Recordset only using ephemeral storage?

rvalyi:
Hello @lasley at Akretion we probably started what you need:
https://github.com/akretion/connector-nosql
also see the solr branch https://github.com/akretion/connector-nosql/tree/9.0-solr
cc @sebastienbeau 
please let's work on this together. We have a great experience with integrating Odoo with Solr and Algolia NoSQL data stores already.

lasley:
Nice, thanks @rvalyi - it does indeed look like you have started what I need. 

Unfortunately one of the requirements of my project is that it has to be LGPL - I should have mentioned that in my RFC in hindsight (edited to include it). The LGPL requirement rules out the use of `connector`, and also rules out inclusion of this logic into `base_external_dbsource`. This requirement is due to my planned use of this module as part of my core SaaS billing infrastructure.

I see a lot of export logic here, but nothing for reading the data. Am I correct in that evaluation?

How did you circumvent the need to store data in Odoo?

dreispt:
@lasley I'm the author for the logic in the `base_external_dbsource`. I looked at the history and the only other original contribution is adding the Firebird DB connections. If it's important, we could consider relicensing that module to LGPL.
In fact, since they are just technical tools, I wouldn't see a problem for the server-tools repo to be as much LGPL as possible.

lasley:
[...]

PR #660, opened 2016-12-20
[10.0][MIG] base_external_dbsource_elasticsearch

lasley:
* Add base_external_dbsource_elasticsearch for CRUD operations on Elasticsearch

This is the forward port of #645 & related to RFC #633 

Depends:
- [x] #659 

External Database Source - Elasticsearch
========================================

This module extends ``base_external_dbsource``, allowing you to connect to
foreign Elasticsearch databases.

Installation
============

* Install ``elasticsearch`` python library

Configuration
=============

Database sources can be configured in Settings > Configuration -> Data sources.

Usage
=====

To use this module:

* Go to Settings > Database Structure > Database Sources
* Click on Create to enter the following information:

* Datasource name 
* Pasword
* Connector: Choose the database to which you want to connect
* Connection string : Specify how to connect to database


base_external_dbsource/README.rst
```diff
@@ -19,43 +19,45 @@ Configuration
...
 Known issues / Roadmap
 ======================
 
+* Find a way to remove or default the CA certs dir
+* Add concept of multiple connection strings for one source (multiple nodes)
+* Add a ConnectionEnvironment that allows for the reuse of connections
+* Message box should be displayed instead of error in ``connection_test``
+* Remove old api compatibility layers (v11)
+* Implement better CRUD handling
+
```
lasley, line 53: * Add map reduce

lasley:
Ok so now that I've finished planning this out properly from the PoC that it is currently, I am going to need a more opinionated connector I think.

We have two ways to go about this - I can add the opinionation here & maybe set a precedence for the other modules. The alternative is to simply create an `elasticsearch` module based on this core in order to provide the more proprietary handling. Both will be about the same amount of work at this point, which is why I bring it up now.

Following is a preliminary data design of what I am thinking. Notice that most of the data models are actually Transient, which is because I don't really want to store any data permanently. I figure if someone does, it's as simply as inheriting the Transient cores with regular models to provide context.

image

Anyone have any thoughts on something like this? cc @dreispt @YannickB @laslabs

dreispt:
Dependency merged. Please rebase.

PR #661, opened 2016-12-20
[IMP] Dynamic placeholder(depends connector value)

aldo774:
Changing value of *connector* field will automatically set value in *conn_string* field

Depends on:
- [x] #659


guewen (approved):

sebalix:
This PR is not compatible with #659 I think.
The connstring should be generated by the respective module once #659 is merged.

aldo774:
I made a rebase on this branch with modifications of https://github.com/OCA/server-tools/pull/659, this PR is dependent of https://github.com/OCA/server-tools/pull/659

dreispt:
@Aldo774 You're welcome to review #659

base_external_dbsource/models/base_external_dbsource.py
```diff
@@ -172,6 +171,48 @@ def execute(
         else:
             return rows
 
+    @api.onchange('connector')
+    def _change_placeholder_str_connect(self):
+        """
+        Changing value of connector field will automatically
+        set value in conn_string field.
+        """
+        if self.connector == 'fdb':
```
lasley, line 180: IMO these should all be in the independent modules & use an interface strategy in the core similar to what I developed for the others. The core module should only have a concept of postgres, and that should still be in an independent adapter method. This avoids the giant unmanageable nested if chains, in favor of a more flat approach - which jives with [PEP-20](https://www.python.org/dev/peps/pep-0020/) better IMO.

I even made a helper method, example:
* https://github.com/multidadosti-erp/server-tools/blob/b231d610b5b0de3e8c5e9d0c63156e6d50899247/base_external_dbsource/models/base_external_dbsource.py#L110

Then some of the corresponding adapter methods that it actually calls:
* https://github.com/multidadosti-erp/server-tools/blob/b231d610b5b0de3e8c5e9d0c63156e6d50899247/base_external_dbsource/models/base_external_dbsource.py#L336
* https://github.com/multidadosti-erp/server-tools/blob/b231d610b5b0de3e8c5e9d0c63156e6d50899247/base_external_dbsource_firebird/models/base_external_dbsource.py#L36

High level info on this interface/adapter implementation is in the [Class level doc block](https://github.com/multidadosti-erp/server-tools/blob/b231d610b5b0de3e8c5e9d0c63156e6d50899247/base_external_dbsource/models/base_external_dbsource.py#L21). This block should be updated with any new methods as well.

lasley (changes requested):
Thanks for the feature @Aldo774! As was mentioned, this is incompatible with the new version though. Please take a look at my inline comment & let me know if you have questions.

We would also want to bump the version of all affected modules up a patch version, as there is a new feature being implemented.

Please also add some tests on the new feature.

PR #666, opened 2016-12-23
[FIX][mail_environment][server_environment] Prevent ImportError.

### LINKED ISSUE #488: Depends server_environment missed
The module server_environment depends of server_environment_files, it not exist. ¿Maybe server_environment_files_samples?

yajo:
Fixes #488.

It should fix Travis in OCA/website, like in https://github.com/OCA/website/pull/303#issuecomment-268797438.

@Tecnativa

### DIFF
--- a/mail_environment/__init__.py
+++ b/mail_environment/__init__.py
@@ -1,2 +1,10 @@
 # -*- coding: utf-8 -*-
-from . import env_mail
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from . import env_mail
+except ImportError:
+    _logger.info("ImportError raised while loading module.")
+    _logger.debug("ImportError details:", exc_info=True)
--- a/server_environment/__init__.py
+++ b/server_environment/__init__.py
@@ -18,4 +18,12 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from .serv_config import serv_config, setboolean
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from .serv_config import serv_config, setboolean
+except ImportError:
+    _logger.info("ImportError raised while loading module.")
+    _logger.debug("ImportError details:", exc_info=True)


pedrobaeza (approved):

mail_environment/__init__.py
```diff
@@ -1,2 +1,10 @@
 # -*- coding: utf-8 -*-
-from . import env_mail
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from . import env_mail
+except ImportError:
+    _logger.info("ImportError raised while loading module.")
```
petrus-v, line 9: Could be nice to tell which module is not able to be imported!

suggested:

```
   _logger.info("ImportError raised while loading env_mail module.")

File: server_environment/__init__.py [hunk 2]
```diff
@@ -18,4 +18,12 @@
...
 ##############################################################################
-from .serv_config import serv_config, setboolean
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from .serv_config import serv_config, setboolean
+except ImportError:
+    _logger.info("ImportError raised while loading module.")
```
petrus-v, line 28: same comment here!

petrus-v (approved):
small typo otherwise you get my :+1:

mail_environment/__init__.py
yajo: No need, the logs already say where do error come from, that's implicit by `getLogger(__name__)`

petrus-v:
In fact the real matter is the missing dependency(sever_environement_files) for server_environment, isn't it?

yajo:
But AFAICS that addon [should be created manually](https://github.com/OCA/server-tools/tree/9.0/server_environment#installation) in each environment.

OTOH, the real problem is [not controlling the ImportError](https://github.com/OCA/pylint-odoo/pull/68). That's a bug.

petrus-v:
ok thanks for those informations! i've never used this module :) I was reading that on the history on other PR...

yajo:
Me neither, but this bug is messing up other repos' bots... 😆

dreispt (approved):

dreispt:
@Yajo sorry for nagging, but you shouldn't merge your own PRs.

DarioLodeiros:
the hard import: https://github.com/OCA/server-tools/blob/8.0/server_environment/serv_config.py#L32
is the problem, when you run Odoo, this import is a real problem becouse it causes boot failure even if it is not installed.
I think that isn't enought this fix... ¿?

moylop260:
What about use the new standard of ImportError to know the exactly package with the error?

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror
```bash
try:
    import external_dependency_python_N
    import external_dependency_python_M
    EXTERNAL_DEPENDENCY_BINARY_N_PATH = tools.find_in_path('external_dependency_binary_N')
    EXTERNAL_DEPENDENCY_BINARY_M_PATH = tools.find_in_path('external_dependency_binary_M')
except (ImportError, IOError) as err:
    _logger.debug(err)
```

yajo:
Hi @dreispt, this had already 3x👍; isn't that enough to merge? (Sorry if I missed something, I've been PSC for not so long).

@DarioLodeiros, I tried that. The problem is that if you catch the ImportError but still continue executing the whole module, another exception would be probably raised because the module was not loaded properly. I decided to catch the exception at a higher level to abort loading the rest of the file(s) if the import fails. After all, this is a quick patch for an addon that I don't use but breaks my environment, its actual developers should do a better patch if they need it 😉.

@moylop260 If I understand fine, you mean that you want to get the ImportError traceback. Actually I'm using a feature of the logging module, [the `exc_info=True` flag](https://github.com/OCA/server-tools/pull/666/files#diff-81332b35f063c24510597cd358dfcd03R10), which outputs current exception traceback automatically appended to the message. I use an INFO to display the import error, but leave the traceback at DEBUG level. Also, [all imports with possible failures are grouped](https://github.com/OCA/server-tools/pull/666/files#diff-e50635d23d671f97c9c8592d7f053a1dR26), so what exactly are you missing here? 😕

sbidoul:
@Yajo why was this module suddenly creating problems? It had not been causing any problem for years.

yajo:
No idea, but [it happened](https://github.com/OCA/website/pull/303#issuecomment-269050419).

DarioLodeiros:
The source of the problem is the hard import of a module that does not exist... server_environment_files
If you clone the repository, actually it break the enviroment... i think that thisn't admissible... ¿?¿

yajo:
Indeed. I mean I don't know how come it was not failing before 😁

sbidoul:
`mail_environment` was never auto imported (unless installed in a database) because it has no `static` directory. So there is no reason it would start failing suddenty. @Yajo I could not find the travis failure you mention. 

#617 created some noise because that module had a `static` directory causing it to be auto imported. But that is fixed in #676. So please guys do not rush at the risk of doing more harm than good.

yajo:
Nice to know that, although could you please detail what did I harm here?

PR #669, opened 2016-12-23
[10.0][FIX] password_security: Validate admin pass

### LINKED ISSUE #652: Password security is causing testing to be difficult
`password_security` is causing things to be more difficult than it should be on a shared Runbot by causing us to change the root password on functional tests.

The fix is simple, stop the password security checks from acting on the root user when in Demo mode. This will allow us to test the module using the demo user, but not keep locking each other out on Runbot.

It turns out I don't actually know how to identify demo mode though. If someone can point me in a direction on that, I'd be happy to fix. Otherwise I'll do some discovery at some point soon, I just wasn't able to find anything quickly & didn't want to forget this.

lasley:
* Add current time as password_write_date for admin user in demo, disabling the reset prompt - fixes #652 

Forward port of #668

### DIFF
--- a/password_security/__manifest__.py
+++ b/password_security/__manifest__.py
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'author': "LasLabs, Odoo Community Association (OCA)",
     'category': 'Base',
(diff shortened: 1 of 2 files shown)


yajo (approved):

moylop260 (approved):

ovnicraft:
2 👍  so is ready to merge ?

sebalix (approved):

Issue #667, opened 2016-12-23
server_environment breaks existing installations


lasley:
`server_environment` breaks existing installations by raising an Exception in the global namespace if there is a missing config variable - https://github.com/OCA/server-tools/blob/10.0/server_environment/serv_config.py#L39

I feel that this should instead be handled similar to our import guards, otherwise pulling this repo breaks things badly. It just happened to us on a production server.

If the `Exception` is truly necessary, it should be moved further back into the module itself. It should also be changed into an `EnvironmentException` at the very least.

@t3ddftw will be submitting a fix soon.

pedrobaeza:
Closing as #670 is merged.

PR #670, opened 2016-12-23
[FIX] server_environment: Protect import

pedrobaeza:
Trying to fix the mess that this module merge has provoked.

Fast-tracking if all goes right.

### DIFF
--- a/server_environment/serv_config.py
+++ b/server_environment/serv_config.py
@@ -19,6 +19,7 @@
 #
 ##############################################################################
 
+import logging
 import os
 import ConfigParser
 from lxml import etree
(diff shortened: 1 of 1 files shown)


lasley:
Re #667

lasley (approved):
👍  pending CI 🍏

tedsalmon:
Hey @pedrobaeza,

The `Exception` will still raise because you're only catching the `ImportError`.  I actually fixed this earlier in this branch https://github.com/LasLabs/server-tools/blob/9.0-guard-server-environment-exceptions/server_environment/serv_config.py and came here to submit the PR :P 

Thanks!

server_environment/serv_config.py
```diff
@@ -29,29 +30,37 @@
...
+    _dir = os.path.dirname(server_environment_files.__file__)
 
-if not os.path.exists(ck_path):
-    raise Exception(
-        "Provided server environment does not exist, "
-        "please add a folder %s" % ck_path
-    )
+    if not system_base_config.get('running_env', False):
+        raise Exception(
+            "The parameter 'running_env' has not be set neither in base "
```
tedsalmon, line 46: Though I know you didn't write this exception, It could use a little sprucing up. Might I suggest:
```
            "The parameter 'running_env' is not set in the base config file, "
            "option -c or in openerprc.\n"
            "We strongly discourage the use of an rc file and instead "
            "recommend you explicitly add the following options into your "
            "config:\n[options]\nrunning_env = [dev|prod]"
```

server_environment/serv_config.py
```diff
@@ -29,29 +30,37 @@
...
-ck_path = os.path.join(_dir, system_base_config['running_env'])
+    _dir = os.path.dirname(server_environment_files.__file__)
 
-if not os.path.exists(ck_path):
-    raise Exception(
-        "Provided server environment does not exist, "
-        "please add a folder %s" % ck_path
-    )
+    if not system_base_config.get('running_env', False):
+        raise Exception(
```
tedsalmon, line 45: EnvironmentError makes more sense than a plain `Exception`

server_environment/serv_config.py
```diff
@@ -29,29 +30,37 @@
...
+            "config file option -c or in openerprc.\n"
+            "We strongly recommend against using the rc file but instead use "
+            "an explicit config file with this content:\n"
+            "[options]\nrunning_env = dev"
+        )
+
+    ck_path = os.path.join(_dir, system_base_config['running_env'])
+
+    if not os.path.exists(ck_path):
+        raise Exception(
```
tedsalmon, line 56: `EnvironmentError` makes more sense than a plain `Exception`

server_environment/serv_config.py
```diff
@@ -29,29 +30,37 @@
...
 
-from openerp.addons import server_environment_files
-_dir = os.path.dirname(server_environment_files.__file__)
+_logger = logging.getLogger(__name__)
 
 # Same dict as RawConfigParser._boolean_states
 _boolean_states = {'1': True, 'yes': True, 'true': True, 'on': True,
                    '0': False, 'no': False, 'false': False, 'off': False}
 
-if not system_base_config.get('running_env', False):
```
tedsalmon, line 39: Why not move this logic (L39:L54) into `_load_config()` and then wrap the `serv_config = _load_config()` in a `try/except` for EnvironmentError?

tedsalmon (changes requested):
Made some suggestions based on my previous comment :)

pedrobaeza:
@t3ddftw, thanks for the suggestion, but as this is only a PR for fixing the error on branches with this module not installed, is out of bounds to refactor it, so I prefer to keep it as is, because I'm not even using it.

P.S.: cryptic login you have, hehe

pedrobaeza:
OK, fast-tracking as stated. Please make another PR for the suggestion you made.

tedsalmon:
@pedrobaeza,

Yeah, we aren't using this either but similarly to you, we ran into issues deploying this repo. I'll submit another PR as you suggested as I don't like the fact that you can still run into an `Exception` if you have `server_environment_files` available but no `running_env` config option set :).

Haha yes, this has been my GitHub login for too long to give it up now.

Thanks!

PR #672, opened 2016-12-23
[9.0][IMP] server_environment: Update Error Guards

tedsalmon:
* Convert Exceptions to EnvironmentErrors
* Move config checking into function `_load_config()`
* Guard execution of `_load_config()` against EnvironmentErrors and log info
* Update exception verbiage

### DIFF
--- a/server_environment/serv_config.py
+++ b/server_environment/serv_config.py
@@ -32,35 +32,18 @@ from .system_info import get_server_environment
 
 _logger = logging.getLogger(__name__)
 
+try:
+    from openerp.addons import server_environment_files
+    _dir = os.path.dirname(server_environment_files.__file__)
+except ImportError:
+    _logger.info('ImportError raised while trying to load module: '
+                 '`server_environment_files`')
+    _logger.debug('ImportError Details: ', exc_info=True)
+    server_environment_files = False
+
 # Same dict as RawConfigParser._boolean_states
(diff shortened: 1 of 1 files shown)


lasley (approved):
Yes, exceptions out of global namespace. Thanks @t3ddftw 👍 

The exceptions should be translated though IMO.

sbidoul (changes requested):
It is important that a missing `running_env` remains a hard error preventing the server to start, rather than a quiet log entry. At first glance this PR changes this behaviour.

lasley:
@sbidoul - Without this change, simply cloning the `server-tools` repo into Odoo breaks installations badly. It seems to me that no matter the reasoning, breaking existing installations with modules that are not installed should not happen. Am I incorrect in this evaluation?

sbidoul:
@lasley I thought this particular issue you mention was resolved by #670. This PR seems to go further. IMO, preventing the server to start when `running_env` is absent is a design feature of `server_environment` and must be preserved.

lasley:
@sbidoul - That only partially resolves the issue because the Exception is still raised inside of the global namespace if the import passes. Odoo still attempts the module load, which ends us with an emergency line call in the middle of the night/early morning.

 I see your #676 & will review that. Thanks!

tedsalmon:
@lasley,

The changes in #676 indeed resolve the issue. I'll go ahead and close this PR :).

Thanks!

PR #673, opened 2016-12-25
[ADD] red_october: External Crypto Handling

lasley:
This module adds support for secure, external encryption and decryption of data using [CloudFlare Red October](https://blog.cloudflare.com/red-october-cloudflares-open-source-implementation-of-the-two-man-rule/).

It also allows for the creation of vaults, management of users, and delegation of decryption rights to the vaults.

Technical
---------

Crypto is implemented by providing a ``red_october`` type to ``ir.attachment``, which utilizes a remote Red October instance to seamlessly encrypt and decrypt data on the fly.

It also provides some fields & form widgets to allow for similar operation to internal fields with seamless encryption:

* `.fields.RedOctoberChar` - Acts like `odoo.fields.Char`

Developer Implementation
========================

While this module does technically provide the ability to create standalone encrypted files, that is not feasible from a usability perspective.

Instead, the preferred approach to allowing encryption is to use the provided fields in a model and subsequently add into the view.

`Model Example:`
``` python
    from odoo.addons.red_october.fields import RedOctoberChar

    class MyModel(odoo.models):
        _name = 'my.model'
        encrypted_field = RedOctoberChar()
```

`View Example:`
``` xml
    <record id="my_model_view_form" model="ir.ui.view">
        <field name="name">My Model Form View</field>
        <field name="model">my.model</field>
            <form string="My Model">
                <field name="encrypted_field" />
            </tree>
        </field>
    </record>
```
Installation
============

* A working Red October installation is required to use this module, which can easily be launched using Docker: ``docker run laslabs/alpine-red-october``
* Install the Python Red October library ``pip install git+https://github.com/LasLabs/python-red-october``

Usage
=====

Create the Vault
----------------

* Go to `Settings => Crypto => Vaults` to create a new vault.
* Once created, a vault must be initialized with an administrative user using the `Action => Init Vault` button from the vault form view.
  * If the vault was already activated externally, make sure to check the `Already Active` box.

Crypto Profiles
---------------

Crypto Profiles are currently 1:1 with Odoo Users. If a user does not have a Crypto Profile, it will be created on demand using the Odoo password as the initial password.

Crypto Menu
-----------

The Crypto Menu can be accessed in the top navigation bar by using the dropdown menu with a lock icon.

The following actions are available in the Crypto Menu:
* Change Profile Password
* Delegate Decryption Rights

Encryption/Decryption
---------------------

* Navigate to a form view implementing an encrypted field
  * When in demo mode, a sample encrypted field is available under the `Preferences` tab in the `res.users` form view

If there is a value saved in the encrypted field, the user will be immediately prompted for their decryption password.

If there is no value in the field, the user is not prompted for a password until they edit the record, add a value, then save the record.

Once a password is entered, it will be used to decrypt all encrypted fields until the user refreshes their browser.

The user that initially adds the encrypted value will be the owner of the file. They will need to subsequently grant ownership others if they also need access to the data.

Note that there must be a valid rights delegation on the Vault in order to decrypt a file, even if you are the only owner.

Delegate Decryption Rights
--------------------------

Proper rights delegation is required to allow the Red October Vault to decrypt data using your credentials.

Once decryption rights have expired, a Vault will no longer be able to decrypt your data, even with the correct password. Adding decryption rights to the Vault again will restore its ability to decrypt your data.

Note that a Vault can only maintain one delegation at a time per user, so any new delegations will replace existing delegations for the same user.

* Click the `Crypto Menu`
* Click the name of the Vault you want to delegate rights to
* Enter the delegation information:
  * Number of Uses - How many decryptions are allowed for this delegation
  * Date of Expiration - When will this delegation expire, if the Number of Uses has not been exceeded
  * Password - Password to the currently active Crypto Profile

Change Password
---------------

New Crypto Profiles are created using the current user's Odoo password, which is stored as plain text in the Odoo request session. This is insecure, and it is recommended that all users immediately change their password before working with sensitive data.

* Click the `Crypto Menu`
* Click `Change Password`
* Enter the requested information

Known issues / Roadmap
======================

* Add caching for most methods
* Allow transferring files between vaults.
* Add delegation uses & delta to ``red.october.file.owner``.
* Proper handling for multiple profiles per user
* All model methods that touch the password or decrypted data need to be moved from
  Odoo models into standard objects. This will provide more security, because the Odoo
  model inheritance magic can be utilized to inject malicious code with a module. Using
  the standard Python inheritance will require that malicious code is placed directly in
  the base module in order to be used.
* Audit Javascript XSS vulnerabilities
* Company rules for automatic ownership of files
* Company rules to enforce password change on Crypto Profiles (initial, and recurrent)
* Company rules to enforce password complexity requirements (glue with
  `OCA/server-tools/password_security`?)


lasley:
Alright I have finally finished my PoC, which actually works quite nicely. I'll update the original post above once I finalize things & update the ReadMe, but below is a summary of the additions.

A new encryption profiles menu has been added to the primary navigation bar:
image


By default, profiles are created on the fly using the session password. Users can change their password using the profiles menu.

image


Encryption is provided with the `.fields.RedOctober` field, which comes with a text widget. The user will be prompted for their password as soon as the field is present in a view, with non-blank data - or if they try to add a value that was previously blank.

image

Once the user enters their password, it will persist in the profile object until refresh. All fields that the user has access to decrypt will show the decrypted value. Inaccessible fields will be shown as so:

image

The best part is that the password never touches the database in any way. The Odoo server does have access to the password and the decrypted data for the duration of the request, but this is the same weakness that would be present in a transparent database encryption as well. 

I still have a lot of edge cases to weed out and the code is incredibly crappy, but it works. I also didn't add any of the delegation handling yet, so that has to be done manually via the Red October API at the moment.

Anyone have thoughts on this implementation? Note that you can ignore the `base_fields_ephemeral` module, which will be submitted in another PR but was required for my testing. 

cc @laslabs @hbrunn

lasley:
Alright my technical implementation is now done for the most part, I just need to add unit tests. I welcome anyone to try this - the ReadMe has now also been updated & the PR description reflects its contents.

red_october/README.rst
```diff
@@ -0,0 +1,201 @@
...
+* Add caching for most methods
+* Allow transferring files between vaults.
+* Add delegation uses & delta to ``red.october.file.owner``.
+* Proper handling for multiple profiles per user
+* All model methods that touch the password or decrypted data need to be moved from
+  Odoo models into standard objects. This will provide more security, because the Odoo
+  model inheritance magic can be utilized to inject malicious code with a module. Using
+  the standard Python inheritance will require that malicious code is placed directly in
+  the base module in order to be used.
+* Audit Javascript XSS vulnerabilities
```
hbrunn, line 167: if you're interested, I was in contact with people from http://codeblau.de for a review of my encrypted fields approach. They still have a hefty price tag, but agreed to negotiate given they also appreciate open source projects
The founder of this company is mostly known for being the author of https://www.fefe.de/dietlibc

red_october/controllers/main.py
```diff
@@ -0,0 +1,108 @@
...
+        type='http',
+        auth='user',
+        methods=['POST'],
+    )
+    def crypt(self, command, user_id, password, data, vault_id=None,
+              owner_ids=None, delegation_min=1):
+        user = http.request.env['red.october.user'].browse(
+            int(user_id),
+        )
+        command = getattr(http.request.env['red.october.file'], command)
```
hbrunn: should we whitelist stuff here? Or at least forbid `command` to start with an underscore? I don't have an exploit at hand, but I feel this could be problematic, because you allow arbitrary functions here

red_october/fields/red_october_char.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def convert_to_cache(self, value, record, validate=True):
+        if isinstance(value, buffer):
+            return str(value)
+        return value
+
+    def read(self, records):
+        # values are stored in attachments, retrieve them
+        attachments = self._get_attachments(records).with_context(
+            bin_size=False,
```
hbrunn: for performance, you might want to check if the current context has this not set. in this case, you don't need to create a new context, which will lose the caches

red_october/fields/red_october_char.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        data = {att.res_id: att.datas for att in attachments}
+        _logger.debug(data)
+        for record in records:
+            record._cache[self.name] = data.get(record.id, False)
```
hbrunn: is it important to have `False` instead of `None`?

red_october/models/ir_attachment.py
```diff
@@ -0,0 +1,62 @@
...
+            ))
+
+    @api.multi
+    @api.depends('type')
+    def _compute_red_october_file_id(self):
+        """ It sets the existing Crypto file or creates a new one. """
+        for record in self:
+            if record.type != self.RED_OCTOBER:
+                continue
+            if self.env.context.get('__crypt'):
```
hbrunn: is it security relevant that anyone can manipulate the context?

red_october/models/red_october_delegation.py
```diff
@@ -0,0 +1,95 @@
...
+    @api.depends('date_expire', 'num_expire')
+    def _compute_active(self):
+        for record in self:
+            record.active = all((
+                record.num_expire, record.delta_expire.total_seconds() > 0
+            ))
+
+    @api.multi
+    def delegate(self, password):
+        """ It runs a remote delegation & deactivates existing active locals.
```
hbrunn: i noticed this before, but now I speak up: In python, documentation tends to be in the active tense: `Run a remote delegation & deactivate existing active locals`

red_october/models/red_october_delegation.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Any existing locals are deactivated because a vault can only maintain
+        one active delegation per user.
+        
+        Args:
+            password (str): Password for the associated Red October user.
+        """
+        if self._context.get(self.CONTEXT_SKIP):
```
hbrunn: also here: I cannot judge the security implications, but anyone can set a context

red_october/models/red_october_file.py
```diff
@@ -0,0 +1,186 @@
...
+        """ It will not allow vault transfers.
+
+        Raises:
+            NotImplementedError: If a vault transfer is being attempted.
+        """
+        if vals.get('vault_id'):
+            changed = self.filtered(
+                lambda r: r.vault_id.id != vals['vault_id']
+            )
+            if len(changed):
```
hbrunn: your don't need to test the length for a recordset: https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5062

red_october/models/red_october_user.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @api.multi
+    @api.constrains('user_id', 'vault_ids')
+    def _check_user_id_vault_ids(self):
+        for record in self:
+            results = self.search([
+                ('user_id', '=', record.user_id.id),
+                ('vault_ids', 'in', record.vault_ids.ids),
+            ])
+            if len(results) > 1:
```
hbrunn, line 68: you should rely on the implementation of truthyness of recordsets

red_october/models/red_october_user.py
```diff
@@ -0,0 +1,206 @@
...
+        return res
+
+    @api.multi
+    def update(self, vals):
+        """ It updates the user on the remote vaults. """
+        for record in self:
+            role = None
+            active = vals.get('is_active')
+            is_admin = vals.get('is_admin')
+            if is_admin is not None and is_admin != record.is_admin:
```
hbrunn, line 96: can I use this to inject evil values in the cache?

red_october/models/red_october_user.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @api.model
+    def change_current_user(self, ro_user_id):
+        """ It changes the current session user to the provided.
+
+        Args:
+            ro_user_id (int): ID of the RedOctoberUser to add as the current
+                session default.
+        """
+        setattr(request.session, self.SESSION_USER_ATTR, ro_user_id)
```
hbrunn, line 118: can I hijack another user's session with this?

red_october/models/red_october_user.py
```diff
@@ -0,0 +1,206 @@
...
+    def read_current_user(self):
+        user = self.get_current_user()
+        return user and user.read() or {}
+
+    @api.model
+    def get_user_profiles(self):
+        """ It returns the current user's profiles. """
+        users = self.search([
+            ('user_id', '=', self.env.user.id),
+        ])
```
hbrunn: `self.mapped` might be simpler here

red_october/models/res_company.py
```diff
@@ -0,0 +1,34 @@
...
+
+class ResCompany(models.Model):
+
+    _inherit = 'res.company'
+
+    default_red_october_id = fields.Many2one(
+        string='Default Crypto Vault',
+        comodel_name='red.october.vault',
+        store=True,
+        compute="_compute_default_red_october_id",
```
hbrunn: we have a different convention for that: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

hbrunn (changes requested):
I want to make clear that I'm an interested security layman at best, so this is only an Odoo related review.
Red October is totally unknown to me, so I can't judge if any communication between the module and this service is problematic. (In my own mind, giving your secrets to 3rd parties is highly problematic, but you'll have thought about that before I guess)

red_october/fields/red_october_char.py
lasley: Mainly just that `False` seems to correspond with what Odoo typically puts as a blank field value. Bugs the hell out of me personally, so let me know if this isn't the case.

red_october/models/ir_attachment.py
lasley: It was! I refactored around this and forgot to remove the hook - good catch

red_october/models/red_october_user.py
lasley: This is a constraint, so we're testing for more than one. I think we're good in this instance?

red_october/models/red_october_user.py
lasley: Hmmmmm probably. I can refactor around this now that I wrote all the JS 👍

red_october/models/red_october_user.py
lasley: Yes, but only in the sense that you would appear logged in as that user. You would still need to provide a decryption password into the UI - so no big deal.

The decryption password is stored in the browser memory (within the Crypto Menu) due to the Odoo frame not actually changing as you navigate the backend. It is never stored in the session, although the user does default with the Odoo password from the session if they are not created through the Vault initialization. This weakness is then circumvented by changing the password via the Crypto Menu

red_october/models/red_october_user.py
lasley: Interesting - how could I use `self.mapped` here? `self.filtered` perhaps?

General question on this - I assumed these methods first gathered all records, then filtered/mapped. Are they operating directly within the DB, before iterating the cursor?

red_october/README.rst
```diff
@@ -0,0 +1,203 @@
...
+The following actions are available in the Crypto Menu:
+* Change Profile Password
+* Delegate Decryption Rights
+
+Encryption/Decryption
+---------------------
+
+* Navigate to a form view implementing an encrypted field
+  * When in demo mode, a sample encrypted field is available under the `Preferences`
+    tab in the `res.users` form view
```
tedsalmon, line 99: I think the indent here is causing some rendering issues in the README. This appears quoted in the Github view.

red_october/README.rst
```diff
@@ -0,0 +1,203 @@
...
+----------------
+
+* Activate Developer Mode
+* Go to `Settings => Crypto => Vaults` to create a new vault.
+  * Make sure to uncheck the `Verify Cert` option if your Red October server is using a self
+    signed cert.
+* Once created, a vault must be initialized with an administrative user using the
+  `Action => Init Vault` button from the vault form view.
+  * If the vault was already activated externally, make sure to check the
+    `Already Active` box.
```
tedsalmon, line 76: I think the indent here is causing some rendering issues in the README. This appears quoted in the Github view.

red_october/README.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Note that a Vault can only maintain one delegation at a time per user, so any new delegations
+will replace existing delegations for the same user.
+
+* Click the `Crypto Menu`
+* Click the name of the Vault you want to delegate rights to
+* Enter the delegation information:
+  * Number of Uses - How many decryptions are allowed for this delegation
+  * Date of Expiration - When will this delegation expire, if the Number of Uses has not been
+    exceeded
```
tedsalmon, line 134: Indent.

red_october/wizards/red_october_vault_activate.py
```diff
@@ -0,0 +1,87 @@
...
+            ))
+        if self.is_active:
+            self.vault_ids.write({
+                'is_active': True,
+            })
+            self.unlink()
+            return
+        for vault in self.vault_ids:
+            vault.activate(self.admin_user_id, self.admin_password)
+        self.unlink()
```
tedsalmon: I believe this is causing a bug where when initiating a new, inactive vault you get an Odoo Warning about missing records.

tedsalmon (commented):
I've left some comments for things I found in my functional tests :)

red_october/wizards/red_october_vault_activate.py
lasley: Damn this was a hack to remove my ephemeral fields. I'll JS this too then

lasley:
@hbrunn - Thanks for the review! Most of the items fixed, will nail the outstanding soon.

Regarding Red October - I thought a lot about this, and my stance is that the external system actually provides a layer of protection that we would not otherwise have. We will assume that Red October itself is secure, because CloudFlare uses it and [says it's secure](https://blog.cloudflare.com/red-october-cloudflares-open-source-implementation-of-the-two-man-rule/). Taking the crypto out of it is huge IMO, and leave us with simple API interactions.

> Red October is a cryptographically-secure implementation of the two-person rule to protect sensitive data. From a technical perspective, Red October is a software-based encryption and decryption server. The server can be used to encrypt a payload in such a way that no one individual can decrypt it. The encryption of the payload is cryptographically tied to the credentials of the authorized users.

With this, we also gain a layer of isolation between our encrypted data and our keys/delegation - which is something we would not otherwise have without the introduction of a new system.

Now what's left to us is specifically watching where the password and decrypted data touch. Everything else is irrelevant, assuming that the connection between Odoo and Red October is also secure. This module does not provide a way to use a Red October server that is HTTP, so the least secure there would be unchecking the `Verify Cert` box in the Vault.

Here's my data architecture diagram of this. Most is obviously irrelevant, but the left hand side is valid in the context of this conversation. Data labelled as `2 - Confidential` (red) in the diagram is the password and decrypted data.

system architecture flow

One of the big things that's a killer for us in terms of security right now is Odoo's model inheritance structure. This magic is actually going to work against us, because to be truly secure (IMO) we cannot allow a simple module installation to expose data that should be otherwise treated as confidential (such as by overriding and catching the `decrypt` method args/return). 

I'm still playing with possible ways for this, but I think the ideal is the creation of a new decorator `@secure(2)`. The number sent as an arg to the decorator will be the amount of bases that can exist on the class. If the amount of bases exceeds allowed, we know we have inheritance & we should fail. This is manual and shitty though, so I'm totally open to some magical idea if you have one 🚀

red_october/fields/red_october_char.py
hbrunn: just ask the field what its null value is: https://github.com/OCA/OCB/blob/10.0/odoo/fields.py#L767

red_october/models/red_october_user.py
hbrunn: right, you check for greater than one

red_october/README.rst
lasley: I'll keep this in mind - at the moment I'm not too worried, but an external audit will need to happen as we start scaling. Thanks!

hbrunn:
I pass the question for the magical idea on, because unfortunately, I don't have it. But there's a nice Metal song about California, if I expect your place of living right: https://www.youtube.com/watch?v=ILZ_lyZVyd8
For forbidding unwanted inheritance, I'd suggest passing down some token in the inheritance chain, and only functions validating it are actually run

red_october/models/red_october_user.py
hbrunn: i think we could make odoo a lot faster by parsing `.mapped` and `.filtered` and emit sql for it. but as things are currently, it's getting stuff and just then filter it

hbrunn: if not clear: all model functions operate on local

lasley:
I'm right next door to California, so close enough! And that metal sent me into a nice Spotify rabbit hole, so I thank you for that 🎉 

Interesting idea with the token - basically a CSRF for classes. I'll run with that, thanks!

chris001:
@lasley 
This looks really nice.  Would be great to have on 11.0.  I notice both Travis build and Runbot build failed, but this was 7 months ago, and the build log isn't available now.  Any updates?

lasley:
@chris001 - I haven't touched this for a few months. At the moment, the module is fully functional though outside of some usability things - primarily with decryption rights delegation. I amend/pushed so that you can functionally test if you'd like.

This is a personal project of mine and doesn't have any specific customer funding, so the progress is slow going - particularly given the feature complexity. I hoped to have it done by end of this year, but we've been particularly busy as of late.

PR #676, opened 2016-12-26
[FIX] do not auto load server_environment_ir_config_parameter

sbidoul:
When a module has a static directory, it is preloaded by the
Odoo http server. This breaks installations which do not have
nor need server_environment_file addon.

This PR removes the static directory which only contains an icon anyway.

See also #670 and #672


sbidoul:
@lasley

lasley (approved):
@sbidoul Hmmm well that's an interesting little fact to know. I wonder if this is also what causes our import error/guard conundrum. 

Can we maybe add a note or something in the ReadMe regarding the fact that this directory is intentionally left out of this module. I see this getting added back during a migration TBH.

Change looks safe enough. @t3ddftw can you please confirm that this can supersede #672.

sbidoul:
>  I wonder if this is also what causes our import error/guard conundrum.

It surely is. I once proposed https://github.com/odoo/odoo/pull/8591 without much echoes.

README updated.

lasley:
> It surely is. I once proposed odoo/odoo#8591 without much echoes.

Thanks for the additional info @sbidoul. I think maybe I'm going to try to propose just a simple import error guard there & hope it gets traction instead. My PRs typically end the same way as that one though 😦

server_environment/README.rst
```diff
@@ -77,6 +77,10 @@ Known issues / Roadmap
 * it is not possible to set the environment from the command line. A
   configuration file must be used.
 * the module does not allow to set low level attributes such as database server, etc.
+* do not add a ``static`` directory to this addon nor any ``OCA/serer-tool`` module that
```
lasley: s/serer-tool/server-tool

yajo (approved):

PR #677, opened 2016-12-26
[FIX] base_multi_image: Use env in uninstall_hook

lasley:
* Uninstall hook should use new API instead of old, which doesn’t exist anymore

This should fix the Travis fails in OCA/product-attribute#201

### DIFF
--- a/base_multi_image/hooks.py
+++ b/base_multi_image/hooks.py
@@ -78,7 +78,7 @@ def uninstall_hook_for_submodules(cr, registry, model):
     """
     env = api.Environment(cr, SUPERUSER_ID, dict())
     Image = env["base_multi_image.image"]
-    images = Image.search(env, [("owner_model", "=", model)])
+    images = Image.search([("owner_model", "=", model)])
     images.unlink()
 
 
--- a/base_multi_image/hooks.py
+++ b/base_multi_image/hooks.py
@@ -78,8 +78,8 @@ def uninstall_hook_for_submodules(cr, registry, model):
     """
     env = api.Environment(cr, SUPERUSER_ID, dict())
     Image = env["base_multi_image.image"]
-    ids = Image.search(env, [("owner_model", "=", model)])
-    ids.unlink()
+    images = Image.search(env, [("owner_model", "=", model)])
+    images.unlink()
 
 
 def table_has_column(cr, table, field):
--- a/base_multi_image/hooks.py
+++ b/base_multi_image/hooks.py
@@ -76,8 +76,7 @@ def uninstall_hook_for_submodules(cr, registry, model):
         Model technical name, like "res.partner". All multi-images for that
         model will be deleted
     """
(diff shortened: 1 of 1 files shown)


base_multi_image/hooks.py
```diff
@@ -76,9 +76,10 @@ def uninstall_hook_for_submodules(cr, registry, model):
         Model technical name, like "res.partner". All multi-images for that
         model will be deleted
     """
-    Image = registry["base_multi_image.image"]
-    ids = Image.search(cr, SUPERUSER_ID, [("owner_model", "=", model)])
-    Image.unlink(cr, SUPERUSER_ID, ids)
+    env = api.Environment(cr, SUPERUSER_ID, dict())
+    Image = env["base_multi_image.image"]
+    ids = Image.search(env, [("owner_model", "=", model)])
```
pedrobaeza: Better to call the variable `records` or `recs`

pedrobaeza:
`registry` still exists on v10: https://github.com/OCA/OCB/blob/10.0/odoo/modules/registry.py#L23.

What is the value of the argument `registry` passed to the function?

lasley:
@pedrobaeza Registry is valid, I just figured that while the env was there I should use it. Is that not the case?

lasley:
The real error I am circumventing is the `cr, SUPERUSER_ID` being sent into the `.search` method: https://travis-ci.org/OCA/product-attribute/jobs/180247258#L559 - the switch from registry just seemed like bonus while I was there honestly.

pedrobaeza:
Have you tried to use directly `registry`, but without `cr` and `SUPERUSER_ID`?

lasley:
Yeah but an environment is still required as the first argument, so I don't see the different from just using the environment itself. Both of these work:

```
    env = api.Environment(cr, SUPERUSER_ID, dict())
    Image = env["base_multi_image.image"]
    images = Image.search([("owner_model", "=", model)])
    images.unlink()
```

```
    env = api.Environment(cr, SUPERUSER_ID, dict())
    Image = registry["base_multi_image.image"]
    images = Image.search(env, [("owner_model", "=", model)])
    images.unlink()
```

But the former seems cleaner and more OOP IMO

I am about to double check all of this in a new install though just to make sure

pedrobaeza:
OK, thanks for the tests.

lasley:
Hmmm yeah ok so that worked too, without the env. In hindsight I'm actually not sure why it worked the first time around when I passed an env to the `.search`, but it definitely does. 

Simplified version, which I'm pretty sure is what you were thinking of, is pushed. Confirmed it works too 😄

pedrobaeza (approved):
Yeah, indeed that was my idea.

To the merger: please squash on merge

lasley:
@tedsalmon please review. Solves this - https://travis-ci.org/OCA/product-attribute/builds/190700970 from OCA/product-attribute#201

StefanRijnhart (approved):

PR #680, opened 2016-12-28
[WIP] Add module multi_localization

aricaldeira:
This module extends the functionality of the base module
to allow installing and using different localizations in
the same database, without one interfering with the other,
both visually and functionally.


aricaldeira:
cc: @rvalyi  @renatonlima @leorochael @mileo

leorochael:
@aricaldeira, please give us some examples of how this module would work

multi_localization/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2016 Taŭga Tecnologia - Aristides Caldeira <aristides.caldeira@tauga.com.br>
+# License AGPL-3 or later (http://www.gnu.org/licenses/agpl)
+#
+
+{
+    'name': u'Multi localization',
+    'version': '10.0.0.0.0',
```
mileo: Please use 10.0.1.0.0

multi_localization/models/res_company.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2016 Taŭga Tecnologia - Aristides Caldeira <aristides.caldeira@tauga.com.br>
+# License AGPL-3 or later (http://www.gnu.org/licenses/agpl)
+#
+
+
+from __future__ import division, print_function, unicode_literals
```
mileo: Please remove this unused imports

mileo (changes requested):

multi_localization/__manifest__.py
mileo: You can see all the OCA Guidelines here:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

aricaldeira:
Code channged as requested

pedrobaeza:
But what is the purpose of this module?

mileo:
@pedrobaeza 

The main purpose is have many localisations installed in the same database, working gently together.

In a install that you have 2 companies:
1. Brazilian 
2. Spanish

If this two companies share the same customers:

For l10r-brazil user some fields like vat and trademark they are useless most of the time. As well as cnpj and partner_fiscal_type ( 2 required fields in l10n_brazil) are useless to l10n-spanish.

This module make views dynamic as user context like when you use user as company 1 you will see properties from company A. 

The next step is handle methods that changed by two locations like action_number of invoice:

https://github.com/OCA/l10n-spain/blob/10.0/l10n_es_account_invoice_sequence/models/account_invoice.py#L35

https://github.com/OCA/l10n-brazil/blob/10.0/l10n_br_account_product/models/account_invoice.py#L512

aricaldeira:
Ok, responding to both @leorochael  and @pedrobaeza :

There's been an ongoing discussion among the l10n-br team, as how to isolate, both visually and functionally (where needed), the new fields, views and methods needed for Odoo to run properly in Brazil, from other l10ns that may be installed on the same database;

For the visual isolation of the l10n, we needed some method to overlook some inherited views, that may replace key fields in the parent/root view, usually to add a new domain to the field, domain that must be obeyed only when dealing with Brazilian companies or environments.

Some of such fields are the CNPJ and IE fields on res_partner model; CNPJ is pretty much the VAT number of Brazil, but IE is an entirely different beast, that has no counterpart on the core implementation of the res_partner model.

Those fields are required in a Brazilian environment, so, in order to a res_partner to be considered really a customer or supplier, those fields must be properly filled;

We cannot simply inherit view_partner_form, put those fields there as required, as it used to be done, because it breaks the form for other companies in the same database, that are not Brazilian companies;

The same applies to the partner_id fields in the sale_order_form view, purchase_order_form, etc. They all need a domain=[('cnpj_cpf', '!=', False)] on the form, which also breaks things for others;

To deal with it, the only way we could think is to somehow not merge our l10n-br inherited views when the active user and company doesn't need them, because they do not speak Portuguese and are not Brazilian.

That way the l10n-br team will be able to change any view, in any needed way, inserting required fields, new domains etc., affecting only those companies and/or users that are based in Brazil and must respect Brazilian l10n rules.

As for the inherited methods, we are still discussing the best approach: at first we thought of a new decorator, api.localization (say @api.localization('L10N-BR')), to be applied to the inherited methods; the decorator would test if the company/user must follow the localized method, and just call super if not;

This decorator would be applied not only to inherited methods, but to some new methods to: for the 2 fields described above, the model must implement also 2 methods _check_cnpj_cpf and _check_ie, both of them with de @depends decorator applied; since the fields are required depending on the circumstance of the company being Brazilian, a decorator would save us from lots and lots of ifs;

But, the decorator approach is not settled yet;

pedrobaeza:
Do you know the module https://github.com/OCA/server-tools/tree/9.0/base_view_inheritance_extension? I think is what you need.

aricaldeira:
Pedro,

Thanks, I didn't know that module.

But, I analysed it, and I think it doesn't do quite what we need.

It allows to do more complex changes when inheriting a view, but I didn't see how it could make a view NOT to be inherited/merged when the use case (l10n selected by company/user) requires it not to, as it seems to still allways apply/merge the views.

The method used here completly ignores some views, as if they were not installed at all.

PR #682, opened 2016-12-30
[IMP] Update LetsEncrypt and improve readme

kenneth-vkd:
I have updated the readme.rst file to be a little more helpful for users that are not technically experienced in nginx configuration as well as fixed issues that prevented Odoo 10 form generating a LetsEncrypt certificate.

### DIFF
--- a/letsencrypt/README.rst
+++ b/letsencrypt/README.rst
@@ -2,7 +2,6 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-=============================================
 Request SSL certificates from letsencrypt.org
 =============================================
 
@@ -89,7 +88,7 @@ You'll also need a matching sudo configuration, like::
 
     your_odoo_user ALL = NOPASSWD: /usr/sbin/service nginx reload
 
-Further, if you force users to https, you'll need something like for nginx::
+Further, if you would like users to be automatically redirected to https, you'll need something like for nginx::
 
     if ($scheme = "http") {
         set $redirect_https 1;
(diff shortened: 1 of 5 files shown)


letsencrypt/README.rst
```diff
@@ -101,20 +100,71 @@ Further, if you force users to https, you'll need something like for nginx::
...
+    server {
+            listen 443;
+            listen 80;
+            if ($scheme = "http") {
+                    set $redirect_https 1;
+            }
+            if ($request_uri ~ ^/.well-known/acme-challenge/) {
+                    set $redirect_https 0;
+            }
+            if ($request_uri ~ ^/pos/web/) {
```
hbrunn, line 120: why this?

letsencrypt/README.rst
```diff
@@ -101,20 +100,71 @@ Further, if you force users to https, you'll need something like for nginx::
         rewrite ^   https://$server_name$request_uri? permanent;
     }
 
+A complete configuration file could look like this::
```
hbrunn, line 103: a user who needs the below probably also needs instructions what to change where. You probably should add comments about that if you want this to be foolproof

letsencrypt/__manifest__.py
```diff
@@ -3,9 +3,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Let's encrypt",
-    "version": "9.0.1.0.0",
+    "version": "1.0.0",
```
hbrunn, line 6: the version should be `10.0.1.0.0`

letsencrypt/__manifest__.py
```diff
@@ -3,9 +3,10 @@
...
     "author": "Therp BV,"
               "Tecnativa,"
+              "VK Data ApS,"
```
hbrunn, line 9: contributor I'd understand, but author seems a bit over the top to me

hbrunn (changes requested):

letsencrypt/README.rst
```diff
@@ -101,20 +100,71 @@ Further, if you force users to https, you'll need something like for nginx::
...
+            proxy_read_timeout 720s;
+            proxy_connect_timeout 720s;
+            proxy_send_timeout 720s;
+
+            proxy_set_header X-Forwarded-Host $host;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_set_header X-Real-IP $remote_addr;
+
+            ssl on;
```
lasley, line 136: These should match current Mozilla recommended settings - https://wiki.mozilla.org/Security/Server_Side_TLS

There's a generator here - https://mozilla.github.io/server-side-tls/ssl-config-generator/

Might also want to add a comment with that link.

NL66278 (changes requested):
Glad this will be moved to 10.0. Please take into account remarks by hbrunn and lasley, and then it will look good to me.

yvaucher:
I close this as no activity since 2017-3-2

PR #683, opened 2016-12-30
[9.0][IMP] base_name_search_improve

jjscarafia:
### Improvements
Several improvements on the great module base_name_search_improve:

1. add boolean "add_smart_search" to ir.model, if you set it True for any model (for eg. "res.partner"), on the top of the search view this field will be add to allow you to search by the same "base_name_search_improve" criteria
2. ~~add "name_search_exact_ids" to ir.model: as for "name_search_ids" you can select some fields but when doing the search it will try to search on thos fields with a "=ilike" criteria, if any result found, return it and don't keep going (useful to use, for eg, with identifications or internal reference for partners, products, etc).~~ DEPRECEATED
3. add "name_search_domain" on ir.model: you can add any domain you want there, for eg you can use in projects with "[('state', '=', 'open)]" so the smart search and the name_search will only return open projects
4. ~~by default the module use an space (" ") as separator to try the different searches, we add the possibility to change it by setting a parameter, by default we keep " " as separator~~ DEPRECEATED
5. by default, in this module, if for eg. you add the fields "name" and "brand" to product model and you try type "Speaker Boss" and you have a product with name="Speaker" and brand="Boss" it won't be returned as a result, we change search method so this cases are covered.

What do you think about this? Any suggestion?

### help needed:
Some more ideas buy I need some help:

- if you use the "Smart Search" computed field on a search view and you search something like "Speaker Boss 15" it would be great that search is rendered as 3 separate searches over "Smart Search" so, if you user want to remove, for eg, "15" and choose "17" he can do it. Of course user could enter "Speaker", later "Boss", and later "15", but that way it takes more time because on each search odoo will return the results
- I believe that the changes of point "5" can be improved, any suggestion?
- Modify the _add_magic_fields method on the _register_hook, I couldn't make it work

seleccion_024
seleccion_023

### DIFF
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -203,17 +203,6 @@ class ModelExtended(models.Model):
                 return res
             return fields_view_get
 
-        # def patch_add_magic_fields():
-        #     @api.model
-        #     def _add_magic_fields(self):
-        #         res = _add_magic_fields.origin(self)
-        #         if 'smart_search' not in self._fields:
-        #             self._add_field('smart_search', fields.Char(
-        #                 automatic=True, compute='_compute_smart_search',
-        #                 search='_search_smart_search'))
-        #         return res
-        #     return _add_magic_fields
-
         @api.multi
         def _compute_smart_search(self):
             return False
@@ -259,9 +248,5 @@ class ModelExtended(models.Model):
                     'name_search', make_name_search())
                 Model._patch_method(
                     'fields_view_get', patch_fields_view_get())
-                # TODO improove this, not sure why but it dones works when
-                # patching this way
-                # Model._patch_method(
-                #     '_add_magic_fields', patch_add_magic_fields())
 
         return super(ModelExtended, self)._register_hook(cr)
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -2,10 +2,12 @@
 # © 2016 Daniel Reis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import models, fields, api
+from openerp import models, fields, api, _
 from openerp import SUPERUSER_ID
 from openerp import tools
-
+from lxml import etree
+from ast import literal_eval
+from openerp.exceptions import ValidationError
 
 # Extended name search is only used on some operators
 ALLOWED_OPS = set(['ilike', 'like'])
@@ -21,6 +23,38 @@ def _get_rec_names(self):
     return rec_name + other_names
 
 
+@tools.ormcache(skiparg=0)
+def _get_rec_exact_names(self):
+    "List of fields to exact search into"
+    model = self.env['ir.model'].search(
+        [('model', '=', str(self._model))])
+    other_names = model.name_search_exact_ids.mapped('name')
+    return other_names
+
+
+@tools.ormcache(skiparg=0)
+def _get_add_smart_search(self):
+    "Add Smart Search on search views"
+    return self.env['ir.model'].search(
+        [('model', '=', str(self._model))]).add_smart_search
+
+
+@tools.ormcache(skiparg=0)
+def _get_name_search_domain(self):
+    "Add Smart Search on search views"
+    name_search_domain = self.env['ir.model'].search(
+        [('model', '=', str(self._model))]).name_search_domain
+    if name_search_domain:
+        return literal_eval(name_search_domain)
+    return []
+
+
+@tools.ormcache(skiparg=0)
+def _get_separator(self):
+    return self.env['ir.config_parameter'].get_param(
+        'base_name_search_improved.separator', default=" ")
+
+
 def _extend_name_results(self, domain, results, limit):
     result_count = len(results)
     if result_count < limit:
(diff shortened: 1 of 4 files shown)


jjscarafia:
@dreispt your review is more than welcome as your are the creator of this module! Thanks for the great work!

base_name_search_improved/models/ir_model.py
```diff
@@ -21,6 +23,38 @@ def _get_rec_names(self):
...
+    name_search_domain = self.env['ir.model'].search(
+        [('model', '=', str(self._model))]).name_search_domain
+    if name_search_domain:
+        return literal_eval(name_search_domain)
+    return []
+
+
+@tools.ormcache(skiparg=0)
+def _get_separator(self):
+    return self.env['ir.config_parameter'].get_param(
```
moylop260: The method `get_param` of `ir.config_parameter` model for v9.0 has lru cache from original modules (is not the case for 8.0)
[ir_config_parameter.py#L70](https://github.com/odoo/odoo/blob/a71cdd02ec6f0a738b48f0bf792de81e17ce240c/openerp/addons/base/ir/ir_config_parameter.py#L70)
Then maybe we don't need a ormcache here.

base_name_search_improved/models/ir_model.py
```diff
@@ -29,25 +63,82 @@ def _extend_name_results(self, domain, results, limit):
...
+
+_add_magic_fields_original = models.BaseModel._add_magic_fields
+
+
+@api.model
+def _add_magic_fields(self):
+    res = _add_magic_fields_original(self)
+
+    if 'base_name_search_improved' in self.env.registry._init_modules:
+        if 'smart_search' not in self._fields:
```
moylop260: What about use `and` instead of two nested `if` in order to reduce cyclomatic complexity

base_name_search_improved/models/ir_model.py
```diff
@@ -29,25 +63,82 @@ def _extend_name_results(self, domain, results, limit):
...
+        string='Name Search Exact Fields',
+        help="If we found exact matches for this fields then we return only "
+        "this results and we don't keep going")
+    name_search_domain = fields.Char()
 
-    def _register_hook(self, cr, ids=None):
+    @api.multi
+    @api.constrains('name_search_domain')
+    def check_name_search_domain(self):
+        for rec in self:
```
moylop260: What about `for rec in self.filtered('name_search_domain'):`?

base_name_search_improved/models/ir_model.py
```diff
@@ -29,25 +63,82 @@ def _extend_name_results(self, domain, results, limit):
...
+            if rec.name_search_domain:
+                name_search_domain = False
+                try:
+                    name_search_domain = literal_eval(rec.name_search_domain)
+                except:
+                    pass
```
moylop260: Catch the exception and add a logger with the exception.message

base_name_search_improved/models/ir_model.py
```diff
@@ -29,25 +63,82 @@ def _extend_name_results(self, domain, results, limit):
...
+                if enabled:
+                    # we add domain
+                    args = args or [] + _get_name_search_domain(self)
+
+                # first we search for an exact match, if we found any, we
+                # return it
+                if name and enabled and operator in ALLOWED_OPS:
+                    exact_fields_names = _get_rec_exact_names(self)
+                    for rec_name in exact_fields_names:
+                        recs = self.search(args + [(rec_name, '=ilike', name)])
```
moylop260: And the `limit`?

base_name_search_improved/models/ir_model.py
```diff
@@ -65,19 +156,111 @@ def name_search(self, name='', args=None,
...
-                                  for x in name.split() if x]
+                    # we only perform this search if we have at least one
+                    # separator character
+                    separator = _get_separator(self)
+                    if separator in name:
+                        domain = []
+                        for word in name.split(separator):
+                            word_domain = []
+                            for rec_name in all_names:
+                                word_domain = (
```
moylop260: Could you help us with a comment to explain this piece of code (I didn't understand it, sorry)?

moylop260 (changes requested):
Could you increase the coverage adding a few tests, please?

base_name_search_improved/models/ir_model.py
jjscarafia: Thanks for the review and the suggestions @moylop260 !
Let me explain and please suggest me a better way to this if you have one. Let me show with an example (the same of the tests, it makes more sense with other case like the one explained on point 5 of this PR.
1. We have a partner named "Johann Gambolputty of Ulm" with phone "+351 777 333 555"
2. We want to search that partner and we remember his name start with "Johann" and his phone has "555"
3. Then we search "Johann 555"
4. I solve it by:
   1. splitting the search with the separator " "
   2. for each work we perform a search on each field with and or, for "Johann" it would be "['|', ('name', 'ilike', 'Johann'), ('ref', 'ilike', 'Johann')]
   3. we append each of those searches with and and, it would be: " ['&', '|', ('name', 'ilike', 'Johann'), ('ref', 'ilike', 'Johann'), '|', ('name', 'ilike', '555'), ('ref', 'ilike', '555')]

Any suggestion?

jjscarafia:
@moylop260 to increase test coverage I should add a test that makes a search over a search view, any suggestion on how to achieve that?

dreispt:
@jjscarafia Thank you for this, the "Smart Search" is a great idea, and is something I was missing in this module.
About 5., I wish that the implementation could be made more expressive/clear. Is the current state it seems a little confusing, but I admit it's not easy.
About 4., why the configurable separator? Do you have particular use in mind? 
About 2., I feels that these exact ids are not necessary, given the "layered" sequential search tries.

jjscarafia:
Thanks @dreispt for the review.
About 5: I agree with you, I'm going to work on it, any suggestion is welcome
About 4: it was related to the need help, first point "if you use the "Smart Search..."  somehow, the space of separator, would perform several searches joined with and "&", so to make it clearer on the search view, some customers prefers to use other character on the search like "Speaker;Boss;15". But I think the really better option would be the one I propose to split the search over several searches (but I still don't know how). So, I agree with you, let's remove this and I will try later to make this graphical split over search view
About 2: I've the same filling as you but you confirm it to me, removed.

jjscarafia:
NOTE: travis errors are related to this merge/pr https://github.com/OCA/server-tools/pull/588

PR #685, opened 2017-01-03
ADD module active_previous_actived_cron

bealdav:
+Use case:
+
+Your ERP is in production, and you have an issue.
+You should stop the automatical tasks to understand what is the real problem.
+before to repair it.
+
+Usage
+=====
+
+* Click on button in General Settings : all tasks (crons) are now inactive
+* Fix the issue
+* Click one more time on button: all tasks previously active, and only them,
+  are now active.
+
+
+When the production is repaired, you may forget to active one or more crons.
+Some of them may run important features.
+
+With this module, you can't forget anymore.

====

- [ ] add ui to inactive / active crons
- [ ] rename module


bealdav:
I don't know how to improve codecov.
How to test unit install and install of modules ?

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,32 @@
...
+    def init(self):
+        crons = self.env['ir.cron'].search([])
+        if crons:
+            vals = {'active': False, 'active_old': True}
+            crons.write(vals)
+
+    def button_uninstall(self):
+        cron_m = self.env['ir.cron']
+        crons = cron_m.with_context(active_test=False).search(
+            [('active_old', '=', True)])
```
lasley, line 34: @bealdav - couldn't you just create a cron matching this query, then test the write? Running the method will trigger the super, which will 🍏  that too

bealdav: thanks @lasley

bealdav:
I'm open to rename my module if any good idea

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,38 @@
...
+    _inherit = 'ir.cron'
+
+    active_old = fields.Boolean(
+        string='Intial active state',
+        help="State of the cron before 'Active Previous Active Cron' "
+             "module installation.")
+
+    def _inactive_crons(self):
+        crons = self.search([])
+        if crons:
```
hbrunn: I don't think this check is necessary, a `write` on an empty recordset is basically a noop

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,38 @@
...
+            vals = {'active': False, 'active_old': True}
+            crons.write(vals)
+
+    def _active_previous_crons(self):
+        crons = self.with_context(active_test=False).search(
+            [('active_old', '=', True)])
+        if crons:
```
hbrunn: same here

active_previous_actived_cron/tests/test_reactive_cron.py
```diff
@@ -0,0 +1,24 @@
...
+        cron_m = self.env['ir.cron']
+        # cron creation
+        vals = {
+            'name': 'Test cron',
+        }
+        cron = cron_m.create(vals)
+        # inactive
+        cron_m._inactive_crons()
+        crons = cron_m.search([])
+        self.assertEqual(len(crons), 0, "All crons should be inactive")
```
hbrunn: `assertFalse` is a bit more concise and (at least to me) simpler to read

active_previous_actived_cron/tests/test_reactive_cron.py
```diff
@@ -0,0 +1,24 @@
...
+        cron_m._active_previous_crons()
+        crons = cron_m.search([])
+        self.assertTrue(cron in [x for x in crons],
```
hbrunn: this is not v6 any more, you can (and should) use [recordset](https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5058) [arithmetics](https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5095)

active_previous_actived_cron/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    'summary': 'Unactive crons on install, reactive previously '
+               'active one on module uninstall.',
+    'version': '10.0.1.0.0',
+    'author': 'Akretion, Odoo Community Association (OCA)',
+    'category': 'Tools',
+    'depends': [
+        'base',
+    ],
+    'website': 'http://www.akretion.com/',
+    'data': [
```
hbrunn, line 16: delete empty key

hbrunn (changes requested):
I must admit I don't see the use case for this module, but you wouldn't have written it without one, so I don't mind.

But I think there's a conceptual problem: Installing/uninstalling a module causes a registry reload, which can impact debugging to the degree of masking the problem you currently want to debug. Why don't you provide a means to flip crons on and off via some button on some `ir.config.settings` or the like? Then you save the registry reload.

And as reminder to turn on the crons again, how about implementing `ir.needaction_mixin`?

hbrunn:
@pedrobaeza probably has two ideas about the rename!

pedrobaeza:
hehe, now that I'm the renamer guy, I suggest to use `cron_activation_persistent` or `cron_activation_changeset`, depending on the final approach you give to the module. I also suggest to make something like a changeset: put a name to an specific cron activation status, having the possibilities to save as much as you want. Include two by default: all activated and all deactivated. Having a button for enabling a certain cron changeset will do the rest, including your use case.

legalsylvain (changes requested):
Hi @bealdav, 

Thanks for this contribution.

Point 1. @hbrunn  said
> Why don't you provide a means to flip crons on and off via some button on some ir.config.settings or the like?

+1. An ir.config.parameter, managed by an ir.config.settings should be better.

Point 2. One of the base cron is autovacuum_job. I guess it's not very good to disable it : 

- it is the only one to be enabled, by default on each installation
- it is the only one that is not set as noupdate="1", that supposes that it should be allways active. (or will be reactivated at each update)

What I see, if you're agree :

* In your class module add a list : 

        _KEEP_ALIVE_MODULE = [
            'base.autovacuum_job',
        ]

* filter on your search, with crons which ids are not in the list defined.

what do you think ?

Point 3. Possible improvment (not a blocking point). When I restore a production database, I'd like to be sure that all cron are disabled by default, specially because crons send mails, or communicate with third apps. So in a second time, we could image that if a key is present on the config file, when we run the odoo instance, crons are disabled too. (my personal use case, I don't know if other OCA members should be interested too).

Point 4. no PoV about the name of the module.

Kind regards.

bealdav:
Hi all, 

Point 1:
OK, no needs to uninstall to switch back to the previous cron state.
but I prefer to have button on settings page than ir.config.settings, what do you think ?

Point 2:
Why not, Ok

Point 3:
I'm ok with you, it could be interesting for dev purpose. Which key name ?

legalsylvain:
Point 1 : The installation of this module should not disable the cron too, right ?

point 3 : "disable_cron_tasks" ? I let

bealdav:
> The installation of this module should not disable the cron too

Yes

bealdav:
Thanks for your suggestions, I will rework my PR

pedrobaeza:
@bealdav, have you seen my suggestions?

bealdav:
yes, but no sure to have completely understood them except if they are in the other proposals

pedrobaeza:
OK, then nevermind

active_previous_actived_cron/__manifest__.py
bealdav: I'll add a file for UI

bealdav:
ping @legalsylvain

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
...
+            [('active_old', '=', True)])
+        crons.write({'active': True})
+        self.env['ir.config_parameter'].set_param(PARAM, False)
+
+    @api.model
+    def _callback(self, model_name, method_name, args, job_id):
+        if config.get(PARAM):
+            disable_cron = self.env['ir.config_parameter'].get_param(PARAM)
+            if not disable_cron:
+                self._inactive_crons()
```
bealdav, line 43: needs to decide the behavior here according cron is in KEEP_ALIVE_CRONS or not

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
...
+
+    active_old = fields.Boolean(
+        string='Intial active state',
+        help="State of the cron before 'Active Previous Active Cron' "
+             "module installation.")
+
+    def _inactive_crons(self):
+        alive_crons = [self.env.ref(x).id for x in KEEP_ALIVE_CRONS]
+        # some crons shouldn't be inactived
+        crons = self.search([('id', 'not in', alive_crons)])
```
legalsylvain, line 27: & ('active', '=', True)

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
...
+        vals = {'active': False, 'active_old': True}
+        crons.write(vals)
+        self.env['ir.config_parameter'].set_param(PARAM, True)
+
```
legalsylvain, line 31: replace line 28-> 30 by : 
    
    if crons:
        vals = {'active': False, 'active_old': True}
        crons.write(vals)
        self.env['ir.config_parameter'].set_param(PARAM, True)

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
+# coding: utf-8
+# © 2017 David BEAL @ Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+from odoo.tools import config
+
+
+PARAM = 'disable_cron_tasks'
```
legalsylvain, line 9: rename into CRON_TASK_DISABLED = 'cron_task_disabled'

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
...
+        crons = self.search([('id', 'not in', alive_crons)])
+        vals = {'active': False, 'active_old': True}
+        crons.write(vals)
+        self.env['ir.config_parameter'].set_param(PARAM, True)
+
+    def _active_previous_crons(self):
+        crons = self.with_context(active_test=False).search(
+            [('active_old', '=', True)])
+        crons.write({'active': True})
+        self.env['ir.config_parameter'].set_param(PARAM, False)
```
legalsylvain, line 36: replace line 35 + 36 by : 

    if crons:
        crons.write({'active': True})
        crons.write(vals)
        self.env['ir.config_parameter'].set_param(CRON_TASK_DISABLED, False)

active_previous_actived_cron/models/ir_cron.py
legalsylvain: Add a new line DISABLE_CRON_TASK = 'disable_cron_task'

active_previous_actived_cron/models/ir_cron.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @api.model
+    def _callback(self, model_name, method_name, args, job_id):
+        if config.get(PARAM):
+            disable_cron = self.env['ir.config_parameter'].get_param(PARAM)
+            if not disable_cron:
+                self._inactive_crons()
+        return super(IrCron, self)._callback(
+            model_name, method_name, args, job_id)
```
legalsylvain, line 45: replace full function by 

    to_disable =  config.get(DISABLE_CRON_TASK) or\
        self.env['ir.config_parameter'].get_param(DISABLE_CRON_TASK)
    disabled = self.env['ir.config_parameter'].get_param(PARAM)

    if to_disable and not disabled:
        self._inactive_crons()
    if not to_disable and disabled:
        self._active_previous_crons()
    return super(IrCron, self)._callback(
        model_name, method_name, args, job_id)

active_previous_actived_cron/README.rst
```diff
@@ -0,0 +1,70 @@
...
+Use case:
+
+Your ERP is in production, and you have an issue.
+You should stop the automatical tasks to understand what is the real problem.
+before to repair it.
+
+Usage
+=====
+
+* Click on button in General Settings : all tasks (crons) are now inactive
```
legalsylvain, line 18: Add : 
1. by UI, on a running instance ....

active_previous_actived_cron/README.rst
```diff
@@ -0,0 +1,70 @@
...
+* Fix the issue
+* Click one more time on button: all tasks previously active, and only them,
+  are now active.
+
+
```
legalsylvain, line 23: 2. By config parameter, adding a key 'disable_cron_task' = True. ....

legalsylvain (changes requested):
I reviewed and propose some changes. Adding a new ir.config;parameter that has to be switch by UI. 'disable_cron_task'.

call me if it's not clear.

yvaucher:
@bealdav are you willing to work on this ?

PR #687, opened 2017-01-06
[9.0][ADD] auth_totp: MFA Support

obulkin:
* Add relevant fields and methods to the `res.users` model
* Overload `check_credentials` in `res.users` to allow for logins using an MFA login token rather than a password
* Add the `res.users.authenticator` and `res.users.device` models, along with appropriate ACLs and record rules
* Add the `res.users.authenticator.create` wizard model and an associated view to facilitate creation of `res.users.authenticator` records
* Extend `base.view_users_form_simple_modif` with fields needed to manage the new functionality
* Add an `AuthTotp` controller that inherits from `Home` in the `web` module and an associated view to introduce MFA logic to the login process
* Add several new exception classes that inherit from `AccessDenied`


auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===========
+MFA Support
+===========
+
+This module adds support for MFA using TOTP (time-based, one-time passwords).
```
lasley, line 9: Break this up. Try to max out around 2-3 sentences per paragraph when tech writing.

Get with me in chat at some point and we can go over some points. Preferably next week after SMD's newb @kellylougheed starts, so we can kill multiple explanations with the same stone.

auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
...
+via the "Change My Preferences" view and an associated wizard. After logging 
+in normally, users with MFA enabled are taken to a second screen where they 
+have to enter a password generated by one of their authentication apps and 
+are presented with the option to remember the current device. This creates a 
+secure, HTTP-only cookie that allows subsequent logins to bypass the MFA step.
+
+Installation
+============
+
+To install this module, simply follow the standard install process, making
```
lasley: Too many words. Bullet point the steps, which is just a pip install.

auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
...
+
+To install this module, simply follow the standard install process, making 
+sure that the PyOTP library is available in your environment. The module will 
+not function properly without PyOTP, which can easily be added via pip: 
+``pip install pyotp``.
+
+Configuration
+=============
+
+By default, the trusted device cookies introduced by this module have a
```
lasley, line 28: Two paragraphs

auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+Known Issues / Roadmap
+======================
+
+Known Issues
+------------
+
+* Due to an Odoo bug, the module does not uninstall cleanly, leaving the
```
lasley: Too many commas - remove the first

auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
lasley: Let's do LGPL-3

Future reference, if we aren't inheriting anything and we are designing it inside of Odoo - I want LGPL always. If we are inheriting, we need to use AGPL if there is an AGPL in the inheritance chain. Otherwise, LGPL

Outside of Odoo, MIT if possible. Licensing sucks dude, (dealwithit)

auth_totp/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
```
lasley: 2016-2017 - we wrote this in both years, copyright is from first write point to last write point.

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from datetime import datetime, timedelta
```
lasley, line 5: Ok so import alpha is weird. Take a look at https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports and let me know in chat with questions

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+from openerp.http import Response, request
+from openerp.addons.web.controllers.main import Home
+from ..exceptions import MfaTokenInvalidError, MfaTokenExpiredError
+
+
+class JsonSecureCookie(SecureCookie):
+    serialization_method = json
+
+
+class AuthTotp(Home):
```
lasley, line 19: carriage return between class and method definition

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+    @http.route()
+    def web_login(self, *args, **kwargs):
```
lasley, line 22: Add docblock please. What does this method do, what args does it expect, what does it return? The latter two are answered but I remain questioning the WHAT without actually reading the method.

Reading a method normally isn't a problem, but this one is >100 lines. While I don't think that we should break it all up or anything, because the logic seems useless elsewhere, we still need to be careful with such dense code.

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+        response = super(AuthTotp, self).web_login(*args, **kwargs)
+
+        if not request.params.get('login_success'):
+            return response
+
+        user = request.env['res.users'].sudo().browse(request.uid)
+        if not user.mfa_enabled:
+            return response
+
+        cookie_key = 'trusted_devices_%s' % user.id
```
lasley: We're expecting an int, so we should use the decimal string replace `%d`. This gives us some implicit type checking that will fail in conditions, such as `user.id == False`

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+            '/auth_totp/login',
+            query={
+                'mfa_login_token': user.mfa_login_token,
+                'redirect': request.params.get('redirect'),
+            },
+            keep_hash=True,
+        )
+
+    @http.route('/auth_totp/login', type='http', auth='none', methods=['GET'])
+    def mfa_login_form(self, *args, **kwargs):
```
lasley: mfa_login_get

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+        return request.render('auth_totp.mfa_login', qcontext=request.params)
+
+    @http.route('/auth_totp/login', type='http', auth='none', methods=['POST'])
+    def mfa_login(self, *args, **kwargs):
```
lasley: mfa_login_post

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+        token = request.params.get('mfa_login_token')
+        try:
+            user = request.env['res.users'].user_from_mfa_login_token(token)
+        except (MfaTokenInvalidError, MfaTokenExpiredError) as exception:
+            if isinstance(exception, MfaTokenInvalidError):
```
lasley: Python exceptions are built to not need to do this. Check it:

```
try:
    armageddon()
except BruceWillisException:
    print "Saved it with a nuke"
except OregonTrailException:
    print "You have died of dysentery. No need to wait for space death."
```

We've got ourselve's an if without the processing of said if. EAFP ftw!

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+
+        user = request.env['res.users'].sudo().browse(request.uid)
+        if not user.mfa_enabled:
+            return response
+
+        cookie_key = 'trusted_devices_%s' % user.id
+        device_cookie = request.httprequest.cookies.get(cookie_key)
+        if device_cookie:
+            config_model = request.env['ir.config_parameter']
+            secret = config_model.sudo().get_param('database.secret')
```
lasley: Nitpicky, but let's centralize our sudos at the top and add a comment as to why we are doing this. 

```
SudoConfig = request.env['ir.config_parameter'].sudo()
SudoUser = request.env['res.users'].sudo()
```

Sudo is a dangerous beast. We should always make this stand out as much as possible in the code, as well as provide context as to the reasoning. Otherwise, a questioning of the sudo will become a yearly occurrence in migrations.

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+            if isinstance(exception, MfaTokenInvalidError):
+                msg = 'Your MFA login token is not valid. Please try again.'
+            if isinstance(exception, MfaTokenExpiredError):
+                msg = 'Your MFA login token has expired. Please try again.'
+
+            return http.local_redirect(
+                '/web/login',
+                query={
+                    'redirect': request.params.get('redirect'),
+                    'error': msg,
```
lasley: What about the non-English folk? `_(msg)`

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+                token = temp_user.mfa_login_token
+        request.session.authenticate(request.db, user.login, token, user.id)
+
+        redirect = request.params.get('redirect')
+        if not redirect:
+            redirect = '/web'
+        response = Response(http.redirect_with_hash(redirect))
+
+        if request.params.get('remember_device'):
+            device_model = request.env['res.users.device']
```
lasley: Move this to a method in the model to create the cookie

auth_totp/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2017 LasLabs Inc.
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
```
lasley: odoo noupdate="1" & remove data tag

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+        comodel_name='res.users.authenticator',
+        inverse_name='user_id',
+        string='Authentication Apps/Devices',
+        help='To delete an authentication app, remove it from this list. To'
+        ' add a new authentication app, please use the button to the right.',
+    )
+    mfa_login_token = fields.Char()
+    mfa_login_token_exp = fields.Datetime()
+    trusted_device_ids = fields.One2many(
+        comodel_name='res.users.device',
```
lasley, line 28: string attr

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+        inverse_name='user_id',
+    )
+
+    @api.multi
+    @api.constrains('mfa_enabled', 'authenticator_ids')
+    def _check_enabled_with_authenticator(self):
+        for record in self:
+            if record.mfa_enabled and not record.authenticator_ids:
+                raise ValidationError(
```
lasley: Translate errors `ValidationError(_(`

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from datetime import datetime, timedelta
```
lasley, line 5: See import order/spacing comment

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+            raise MfaTokenInvalidError()
+
+        user = self.search([('mfa_login_token', '=', token)])
+        user._user_from_mfa_login_token_validate()
+
+        return user
+
+    @api.multi
+    def _user_from_mfa_login_token_validate(self):
+        if not len(self) == 1:
```
lasley: `if len(self) != 1`

Although a better way to do this would be `self.ensure_one()`, which raises an `AssertionError` which you could catch and re-raise as the custom one.

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+            raise MfaTokenInvalidError()
+        if self.mfa_login_token_exp < fields.Datetime.now():
```
lasley: This is not the droid you're looking for in terms of math. This is a string comparison right now.

You're looking for a lil something like this instead I thinks:

```
from datetime import datetime
...
token_exp = fields.Datetime.from_string(self.mfa_login_token_exp)
if token_exp < datetime.now():
```

datetime.datetime != fields.Datetime

auth_totp/models/res_users_authenticator.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from openerp import api, fields, models
+
+_logger = logging.getLogger(__name__)
+try:
```
lasley, line 9: Carriage return between logger and try

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+
+    name = fields.Char(
+        string='Authentication App/Device Name',
+        help='A name that will help you remember this authentication'
+        ' app/device',
+        required=True,
+        index=True,
+    )
+    secret_key = fields.Char(
+        default=lambda s: pyotp.random_base32(),
```
lasley, line 32: required?

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    qr_code_tag = fields.Html(
+        compute='_compute_qr_code_tag',
+        string='QR Code',
+        help='Scan this image with your authentication app to add your account'
```
lasley: add trailing comma

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+        string='Associated User',
+        help='This is the user whose account the new authentication app/device'
+        ' will be tied to',
+        readonly=True,
+        index=True,
+    )
+    confirmation_code = fields.Char(
+        string='Confirmation Code',
+        help='Enter the latest six digit code generated by your authentication'
+        ' app',
```
lasley: tab in, visually align the `'` with each other

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+        return self.env['res.users'].browse(user_id)
+
+    @api.multi
+    @api.depends(
+        'secret_key',
+        'user_id.display_name',
+        'user_id.company_id.display_name',
+    )
+    def _compute_qr_code_tag(self):
+        for record in self:
```
lasley, line 70: `for record in self.filtered(lambda r: r.user_id)` & remove ln69+70

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+                continue
+
+            totp = pyotp.TOTP(record.secret_key)
+            provisioning_uri = totp.provisioning_uri(
+                record.user_id.display_name,
+                issuer_name=record.user_id.company_id.display_name,
+            )
+            provisioning_uri = urllib.quote(provisioning_uri)
+
+            qr_width, qr_height = 300, 300
```
lasley: Ints are immutable, trick like this would be cleaner:
```
qr_width = qr_height = 300
```

```
>>> test_a = test_b = 300
>>> test_a
300
>>> test_b
300
>>> test_b = 400
>>> test_a
300
>>> test_b
400
```

But don't do it with mutables:

```
>>> test_a = test_b = []
>>> test_a.append(1)
>>> test_a
[1]
>>> test_b
[1]
```

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+            tag_base = '<img src="/report/barcode/?type=QR&amp;'
```
lasley, line 82: This is nasty. Two things - construct the src first, then use lxml to make sure you didn't eff up the escaping.

```
from lxml import etree
src = 'qr_insanity'
img =etree.Element('img', src=src)
etree.tostring(img)
```

I made the above up, but should be close enough

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+            record.qr_code_tag = tag_base + tag_params
+
+    @api.multi
+    def action_create(self):
+        self.ensure_one()
+        self._perform_validations()
+        self._create_authenticator()
+
+        action_data = self.env.ref('base.action_res_users_my').read()[0]
```
lasley, line 96: Not totally sure what you're going for here, but I think `record.convert_to_write()` maybe? 

https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=convert_to_write

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+        action_data.update({'res_id': self.user_id.id})
+        return action_data
+
+    @api.multi
+    def _perform_validations(self):
+        totp = pyotp.TOTP(self.secret_key)
+        if not totp.verify(self.confirmation_code):
+            raise ValidationError(
```
lasley: l10n

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,115 @@
...
+                'Your confirmation code is not correct. Please try again,'
+                ' making sure that your MFA device is set to the correct time'
+                ' and that you have entered the most recent code generated by'
+                ' your authentication app.'
+            )
+
+    @api.multi
+    def _create_authenticator(self):
+        self.env['res.users.authenticator'].create({
```
lasley, line 113: This expects singleton

lasley (changes requested):
Excellent job @obulkin - comment inline

lasley:
Please also check 🔴  build

auth_totp/README.rst
```diff
@@ -0,0 +1,103 @@
...
+  custom uninstall logic via an uninstall hook.
+
+Roadmap
+-------
+
+* Make the various durations associated with the module configurable. They are 
+  currently hard-coded as follows: 15 minutes to enter an MFA confirmation 
+  code after logging in using a password, 30 days before the MFA session 
+  expires and the user has to log in again, 30 days before the trusted device 
+  cookie expires.
```
tedsalmon: Each item should be a sub-bullet point, IMO

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+            device = device_model.sudo().create({'user_id': user.id})
+            config_model = request.env['ir.config_parameter'].sudo()
+            secret = config_model.get_param('database.secret')
+            device_cookie = JsonSecureCookie({'device_id': device.id}, secret)
+            cookie_lifetime = timedelta(days=30)
+            cookie_exp = datetime.now() + cookie_lifetime
+            device_cookie = device_cookie.serialize(cookie_exp)
+            cookie_key = 'trusted_devices_%s' % user.id
+            cookie_security = config_model.get_param('auth_totp.secure_cookie')
+            security_flag = False if cookie_security == '0' else True
```
tedsalmon: This can be simplified as `security_flag = cookie_security != '0'` 

or if you want to cast the variable: `security_flag = bool(int(cookie_security))`

auth_totp/controllers/main.py
```diff
@@ -0,0 +1,124 @@
...
+        response = Response(http.redirect_with_hash(redirect))
+
+        if request.params.get('remember_device'):
+            device_model = request.env['res.users.device']
+            device = device_model.sudo().create({'user_id': user.id})
+            config_model = request.env['ir.config_parameter'].sudo()
+            secret = config_model.get_param('database.secret')
+            device_cookie = JsonSecureCookie({'device_id': device.id}, secret)
+            cookie_lifetime = timedelta(days=30)
+            cookie_exp = datetime.now() + cookie_lifetime
```
tedsalmon: I would use `utcnow()` just because you never know what timezone the server is running

auth_totp/wizards/res_users_authenticator_create.xml
```diff
@@ -0,0 +1,41 @@
...
+
+<odoo>
+    <record id="res_users_authenticator_create_view_form" model="ir.ui.view">
+        <field name="name">MFA App/Device Creation Wizard</field>
+        <field name="model">res.users.authenticator.create</field>
+        <field name="arch" type="xml">
+            <form string="Authenticator Info">
+                <header/>
+                <sheet>
+                    <span>Please provide a name for your app/device. Then scan the QR code below to add this account to your authenticator app and enter in the six digit code produced by the app.</span>
```
tedsalmon: Use a div and remove the `<br />`

tedsalmon (changes requested):
Requested changes inline :)

auth_totp/wizards/res_users_authenticator_create.xml
lasley: Yup any time you're using a `<br />`, you're probably doing it wrong

This might be good as a bootstrap panel or well.

auth_totp/README.rst
obulkin: There are only two sentences here. Does each one really need to be a separate paragraph?

lasley: Nope on second read it's good

auth_totp/README.rst
obulkin: I believe that comma is needed there (http://www.wheaton.edu/Academics/Services/Writing-Center/Writing-Resources/The-Comma) given how the sentence is currently constructed, but I can get rid of it by writing the following instead: `The module does not uninstall cleanly due to an Odoo bug`

lasley: Sweet I like that better 👍

auth_totp/controllers/main.py
obulkin: Are you sure? I just checked again, and I think I'm following the guidelines correctly.

auth_totp/controllers/main.py
obulkin: Sure. When you say that it's >100 lines, I assume that you're including the logic in the parent version, so does the docblock have to describe that as well?

lasley: Sorry I meant method block, two methods here that are dense - total about 100 lines. You only need worry about what you're doing in the docblock, no need to document super

auth_totp/controllers/main.py
obulkin: The logic that precedes this step should make it impossible for `user.id` to not be an integer. At this point, it's equivalent to `request.uid`, which is the result of a call to `authenticate()` with `False` excluded (https://github.com/odoo/odoo/blob/9.0/addons/web/controllers/main.py#L475-L476) and hence must be a valid ID of a saved user record.

lasley: Yes but there is a `%d` for interpolating decimal integers. `%s` is string. See more - https://docs.python.org/2/library/string.html#format-specification-mini-language

auth_totp/controllers/main.py
obulkin: Sure. The downside is that `sudo()` would now get called twice each time the method runs, as opposed to when necessary, but I guess that's not a major concern.

auth_totp/controllers/main.py
obulkin: This approach would require writing out the `return` statement twice, so I went with the conditionals to make things more DRY.

lasley: Why? 
```
try:
 ...
except:
  msg = 'whatever'
return http.local_redirect(..
```

auth_totp/controllers/main.py
obulkin: `user.id` would also never be a decimal, but I'm fine with switching over. I just don't see how it helps in this particular case.

auth_totp/controllers/main.py
obulkin: The `return` should only happen if there's an exception, so it needs to be part of the `except` block(s), and with this pattern there would be two such blocks.

auth_totp/controllers/main.py
lasley: Basically it's just the formatting language. In the example of English it's the difference between you saying "The Lake" and "Lake Tahoe". Sure both work, but one is more specific and the correct syntax grammatically.

auth_totp/controllers/main.py
obulkin: Could you elaborate on the reasoning here? It doesn't make sense for the model to actually add the cookie to the `Response` object since building the response seems to be very much in the domain of what the controller should do. This means the model can at most create an unserialized Werkzeug cookie, and that in turn means that both the model and the controller would now need access to the `JsonSecureCookie` class. Ultimately, that seems significantly more complicated, with no clear benefits.

auth_totp/controllers/main.py
obulkin: Fair enough. Should I do that [here](https://github.com/OCA/server-tools/pull/687/files/1fa0f2064df52752e02ba6c6b1d56f118f1bd2e7#diff-86a6308a179237cf51cf7238129c39adR62) as well?

auth_totp/controllers/main.py
obulkin: I'm not a fan of the casting approach because the conversion to an int generates exceptions that would require extra error handling, but I could go with the first option. Personally, I like the ternary, but it does seem to be unpopular in the Python community for some reason.

auth_totp/models/res_users.py
obulkin: As far as I can tell, the order is correct, so is it just the lack of whitespace between groups that's a problem? Since there are only 6 lines, I felt that having them in one block would be fine.

auth_totp/models/res_users.py
obulkin: Since this field is not exposed in any views, I thought it would be fine to leave out the `string` argument. Does this mean that `mfa_login_token` and `mfa_login_token_exp` need it as well?

auth_totp/models/res_users.py
obulkin: My understanding is that the string format used by `field.Datetime` should ensure that string comparisons like this will produce the right result conceptually. Isn't that how searches against `Datetime` fields work in the ORM?

auth_totp/models/res_users_authenticator.py
obulkin: Why? There's no carriage return in the example given here: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports.

auth_totp/wizards/res_users_authenticator_create.py
obulkin: Sure. I think I left that off because this field isn't exposed in the view and has a default, but it doesn't hurt to add it on.

auth_totp/models/res_users.py
lasley: The searches work in ORM because in SQL you query with the dates as a string (because they're stored as a string YYYY-MM-DD HH:MM:SS). That's also why they're stored as a string.

In Python, you must work with them as an object if you want to math/compare.

auth_totp/wizards/res_users_authenticator_create.py
obulkin: Isn't it more efficient to just perform the check as part of the explicit loop? `filtered()` is probably looping through the recordset, creating a new recordset object, and so on.

auth_totp/wizards/res_users_authenticator_create.py
obulkin: The goal was to generate a valid return value for the `action_create` method by using a pre-existing action that has almost all of the required info. It doesn't look like `_convert_to_write()` would really help with this, and using `read` was your suggestion at the time, IIRC.

auth_totp/wizards/res_users_authenticator_create.py
obulkin: I'm ensuring that the recordset is a singleton in `action_create()`, so I figured there was no reason to do it again here or in `_perform_validations()`.

auth_totp/controllers/main.py
tedsalmon: This is true; if `cookie_security` isn't an str of a number you'd run into issues.

I hate ternary because of PHP. See:

Horse

obulkin: Lol. Looks like some parens could really help to clear that up.

moylop260: Ternary is good to me but you have a boolean to get other boolean.
I mean, the comparison returns a boolean: `security_flag = cookie_security != '0'` directly

auth_totp/controllers/main.py
moylop260: Do works translation for variables?
I think so, this comment is for `msg = _('Your MFA ...')` section

auth_totp/controllers/main.py
moylop260: You could use the `exception.message` to unified case.
```python
      try:
           user = request.env['res.users'].user_from_mfa_login_token(token)
       except (MfaTokenInvalidError, MfaTokenExpiredError) as exception:
           msg = exception.message
```

auth_totp/controllers/main.py
moylop260: Where is declared the variable `msg` if there is not a exception?

obulkin: The `return` only happens if there is an exception. Removing the whitespace on L65 may be a good way to make this more clearly visually.

moylop260: Thanks for clarify.
Your code is fine to me... I just didn't see because there are comments

auth_totp/controllers/main.py
obulkin: @moylop260 - Unfortunately, that option is off the table because these inherit from `AccessDenied`, whose constructor doesn't accept a message: https://github.com/odoo/odoo/blob/9.0/openerp/exceptions.py#L52

Edit: Actually, nevermind. I could go with this approach. Thoughts, @lasley?

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+    @api.model
+    def check_credentials(self, password):
+        try:
+            super(ResUsers, self).check_credentials(password)
+        except AccessDenied:
+            user = self.sudo().search([
+                ('id', '=', self.env.uid),
+                ('mfa_login_token', '=', password),
+            ])
+            user._user_from_mfa_login_token_validate()
```
moylop260, line 53: This inherited method requires a `return` to be consistent with original one

auth_totp/models/res_users.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @api.multi
+    def generate_mfa_login_token(self, lifetime_mins=15):
+        char_set = string.ascii_letters + string.digits
+
+        for record in self:
+            record.mfa_login_token = ''.join(
+                random.SystemRandom().choice(char_set) for __ in xrange(20)
```
moylop260: 20 is a minor number then you could use `range` to long term support

auth_totp/models/res_users_authenticator.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    user_id = fields.Many2one(
+        comodel_name='res.users',
+        ondelete='cascade',
+    )
+
+    @api.multi
+    @api.constrains('user_id')
+    def _check_has_user(self):
+        for record in self:
```
moylop260: What about `self.filtered(lambda r: not r.user_id).unlink()`?

auth_totp/models/res_users_authenticator.py
```diff
@@ -0,0 +1,55 @@
...
+            if not record.user_id:
+                record.unlink()
+
+    @api.multi
+    def validate_conf_code(self, confirmation_code):
+        for record in self:
```
moylop260, line 50: Should be `ensure_one()` or is required a loop in many records?

auth_totp/controllers/main.py
lasley: Yeah all exceptions should have a message. I think this is a 👍

auth_totp/controllers/main.py
lasley: Yeah I'm also good with ternaries, and it was fine in this instance if there wasn't the more simple approach of simply using the bool

auth_totp/models/res_users.py
obulkin: There's no explicit return in the original method: https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/res/res_users.py#L442

auth_totp/models/res_users_authenticator.py
obulkin: A user can have multiple authenticators so a loop is required here.

auth_totp/models/res_users.py
moylop260: I know.
A no explicit return in the original method is a implicit `None` return.
If this return value is changed from other inherited or original one you method won't change it

You could see the following example: https://github.com/odoo/odoo/blob/9.0/addons/auth_crypt/auth_crypt.py#L59

moylop260: ...or this one: https://github.com/odoo/odoo/blob/9.0/addons/auth_oauth/res_users.py#L119

auth_totp/controllers/main.py
lasley: Nope that's fine

auth_totp/controllers/main.py
lasley: The clarity is worth it IMO, sudo doesn't do much in terms of power

auth_totp/models/res_users.py
obulkin: K. I see where you're going with this. I'll add a `return` to L45.

auth_totp/controllers/main.py
lasley: A cookie is data. All data construction should take place in models. Only data adaptation can take place in controllers. You are correct that not all of this logic needs to be moved over, just the stuff that defines the data that we expect elsewhere.

This also allows for someone to inherit your data model and inject values into the cookie, or completely override the operation of it altogether. As it stands here, in order to change the cookie, you have to re-implement the entire method.

More general advice - If you find yourself creating and updating records in a controller, you should probably rethink the architecture (unless you are simply passing values to a create method or something - like a CRUD method).

auth_totp/models/res_users.py
lasley: No prob

auth_totp/models/res_users.py
lasley: Nah this hardline rule only extends to relations. Mainly because if it is ever presented on a view, _id or _ids does not correctly define the fact that a user is looking at recordset

auth_totp/wizards/res_users_authenticator_create.py
lasley: If our data models require it, we should always require it. Considering what is or is not exposed in a view is out of scope for a model.

auth_totp/wizards/res_users_authenticator_create.py
lasley: Sure, then you have to test the line. I'm good with either.

obulkin: I have a test case for it [right here](https://github.com/OCA/server-tools/pull/687/files/1fa0f2064df52752e02ba6c6b1d56f118f1bd2e7#diff-6a7cc46d985f940dd490e7713d097887R55). 😄

lasley: There was one missing a test case, but I can't identify it until the tests are passing. As long as the negative case is tested, the more explicit is fine & more efficient as you noted.

auth_totp/models/res_users.py
obulkin: FYI, it looks like `ensure_one()` raises a `ValueError`.

obulkin:
Thanks for the feedback, everyone. I've pushed up a new commit that should reflect the suggestions and any subsequent discussions. All the tests look good, but the Travis build is failing due to the issue discussed [here](https://github.com/OCA/server-tools/issues/689).

lasley (approved):

lasley:
@obulkin - please rebase on current OCA/9.0 to 🍏  the build

obulkin:
@lasley - All the tests look good without that. I guess the branch that this was based on didn't have anything in it that needed to be changed. All the issues must have been over on the OCA:9.0 side.

lasley:
Oh sweet I guess my rebuild did it then. It was on the OCA/9.0 side, which was why I recommended rebase. I guess I still don't fully understand Travis cross-merge

lasley:
@tedsalmon - please update your review if your requested changes were attended to.

tedsalmon (approved):
My apologies for not updating previously. Changes LGTM now 👍

PR #694, opened 2017-01-10
[10.0][ADD] Module base_import_security_group

espo-tony:
This module makes importing data from CSV files optional for each user, allowing it only for those users belonging to a specific group.

This module is a reviewed migration to v10 from the same module proposed on v8 with the PR #154 not merged yet.


espo-tony:
The runbot fail doesn't look like depending from this module.

pedrobaeza:
For the name, I will prefer _base_import_security_group_ following other OCA names like https://github.com/OCA/social/tree/9.0/mass_mailing_security_group

espo-tony:
@pedrobaeza For the name, I kept the consistence with the original v8 module proposed with the PR #154 . I'm in favor of changing the name, but shouldn't we also align the other name?

pedrobaeza:
Well, we can change both PRs to this name. The other I suggest is also good: base_import_restrict, but not consistent with the mentioned module.

espo-tony:
I changed the name of the module in _base_import_security_group_ and suggested this change also on the other PR.

pedrobaeza:
Thanks!

StefanRijnhart:
I checked in runbot and the group renders as a drop down with only one option. Maybe we should remove the group category so that it shows up under 'Other' permissions?

espo-tony:
@StefanRijnhart 
The group renders as a drop-down with only one option just because on the runbot is currently available only that group in the category technical settings, also due to the very few modules installed on it. As soon as other groups will be added to that category (installing other modules, like 'sale' for example), the drop-down will disappear and the category will be rendered as a paragraph with several groups shown as check-box.
In my opinion this group is better placed into the "Technical Settings" category rather than "Other". What do you think?

StefanRijnhart:
@espo-tony yes alright, thanks for the clarification.

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,41 @@
...
+        messages = []
+        info = {}
+        messages.append(dict(info, type='error', message=msg, moreinfo=None))
+        res = {'ids': None, 'messages': messages}
+    return res
+
+
+# Monkey patch function
+# A Because BaseModel _name = None
+BaseModel.load = load_import_optional
```
StefanRijnhart: I think in 10.0 you can override the 'base' model instead of monkey patching core methods. See https://github.com/OCA/server-tools/blob/10.0/base_technical_features/models/base.py for an example.

StefanRijnhart (changes requested):

espo-tony:
@StefanRijnhart thank you for the tip. I removed the monkey patch and extended the Base.load method instead;

base_import_security_group/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
astirpe: Could you add the target tag as it is in the [README template](https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst)?

base_import_security_group/README.rst
```diff
@@ -0,0 +1,66 @@
...
+- https://www.odoo.com/forum/help-1
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/server-tools/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/server-tools/issues/new?body=module:%base_import_security_group%0Aversion:%2010.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
astirpe: Could you update this part, according to the README template?

base_import_security_group/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Group-based permissions for importing CSV files
```
astirpe: Could you include the module title this way:
```
==============
{module_title}
==============
```
as it is shown in the [README template](https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst)?

base_import_security_group/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+
+{
+    'name': 'Optional CSV import',
+    'version': '10.0.1.0.0',
+    'category': 'Server tools',
+    'summary': 'Group-based permissions for importing CSV files',
+    'license': 'AGPL-3',
+    'author': 'Odoo Community Association (OCA), '
+              'Alejandro Santana <alejandrosantana@anubia.es>, '
+              'ONESTEiN BV',
```
astirpe: Could you please rename the author to "Onestein"?

base_import_security_group/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# © 2015 Anubía, soluciones en la nube,SL (http://www.anubia.es)
+# © 2017 ONESTEiN BV (http://www.onestein.eu)
```
astirpe: Could you replace the symbol with "Copyright" (as per the module template) and rename the author to "Onestein"?

base_import_security_group/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+        'web',
+        'base_import',
+    ],
+    'data': [
+        'security/base_import_security_group_security.xml',
+        'views/base_import.xml',
+    ],
+    'installable': True,
+    'application': False,
+    'auto_install': False,
```
astirpe: Could you remove defaults?

base_import_security_group/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © 2015 Anubía, soluciones en la nube,SL (http://www.anubia.es)
+# © 2017 ONESTEiN BV (http://www.onestein.eu)
```
astirpe: Could you replace the symbol with "Copyright" (as per the module template) and rename the author to "Onestein"?

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# © 2015 Anubía, soluciones en la nube,SL (http://www.anubia.es)
+# © 2017 ONESTEiN BV (http://www.onestein.eu)
```
astirpe: Could you replace the symbol with "Copyright" (as per the module template) and rename the author to "Onestein"?

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,38 @@
...
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    @api.model
+    def load(self, fields, data):
+        '''Overriding this method we only allow its execution
+        if current user belongs to the group allowed for CSV data import.
+        An exception is raised otherwise, and also log the import attempt.
+        '''
+        current_user = self.env['res.users'].browse(self.env.uid)
```
astirpe: Cannot use `self.env.user` instead of the browse() here?

base_import_security_group/static/src/js/import.js
```diff
@@ -0,0 +1,33 @@
+odoo.define('web.ListImport', function (require) {
+    "use strict";
+    var core = require('web.core');
+    var ListView = require('web.ListView');
+    var _t = core._t;
+    var _lt = core._lt;
+    var QWeb = core.qweb;
```
astirpe: Are all these `var` necessary?

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,38 @@
...
+        allowed_group = 'base_import_security_group.group_import_csv'
+        allowed_group_id = self.env.ref(
+            'base_import_security_group.group_import_csv',
+            raise_if_not_found=False)
+        if not allowed_group_id or \
+                (current_user and current_user.has_group(allowed_group)):
```
astirpe: Is there a case that `current_user` is False?

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.model
+    def load(self, fields, data):
+        '''Overriding this method we only allow its execution
+        if current user belongs to the group allowed for CSV data import.
+        An exception is raised otherwise, and also log the import attempt.
+        '''
+        current_user = self.env['res.users'].browse(self.env.uid)
+        allowed_group = 'base_import_security_group.group_import_csv'
+        allowed_group_id = self.env.ref(
```
astirpe: Cannot use the just defined variable above to replace the string?

astirpe (changes requested):

astirpe (approved):

espo-tony:
Branch Rebased.

espo-tony:
@StefanRijnhart please, can you check the changes you requested and eventually approve them?

StefanRijnhart:
Thanks for the updates! You merged rather than rebased, but we can squash this change rather than merge. :+1: for this change.

About @astirpe's request to remove all references to Anubia, is this module derived from @alejandrosantana's work or not? If it is, we have to keep the credits of course. He has signed the OCA CLA so inclusion of his work is unproblematic.

astirpe:
@StefanRijnhart I didn't request to remove all references to Anubia

espo-tony:
@StefanRijnhart This work is explicitly derived from the work of @alejandrosantana , as declared in the PR description itself. Of course all the references to his work are kept in this PR, both in authors and contributors as well as in the header of all files.

StefanRijnhart:
@astirpe Sorry, I misunderstood!

StefanRijnhart (approved):

base_import_security_group/models/models.py
```diff
@@ -0,0 +1,39 @@
...
+            res = super(Base, self).load(fields=fields, data=data)
+        else:
+            msg = ('User (ID: %s) is not allowed to import data '
+                   'in model %s.') % (self.env.uid, self._name)
+            _logger.info(msg)
+            messages = []
+            info = {}
+            messages.append(
+                dict(info, type='error', message=msg, moreinfo=None))
+            res = {'ids': None, 'messages': messages}
```
StefanRijnhart, line 38: You might want to replace the occurrences of 'None' with 'False', as None cannot be serialized through xmlrpc.

StefanRijnhart:
I refactored the 8.0 version here: https://github.com/OCA/server-tools/pull/743. Maybe you can adopt the phantom_js test.

espo-tony:
@StefanRijnhart 
Thank you for the approval.
I'm also trying to adopt the phantom_js tests as you hinted, but despite my efforts and several different attempts I can't get them running correctly. I continue to incur in a "_RuntimeError: cannot join current thread_" error, raised after that the phantom_js test itself ran successfully.

If you are willing to open a PR over this, adding the phantom_js tests, I'll be happy to merge it; otherwise I think it would be better to get this migration merged and then eventually to open a new PR in order to add phantom_js tests to this module.

StefanRijnhart:
@espo-tony OK, I'll try to have a look later today.

StefanRijnhart:
@espo-tony here you go: https://github.com/espo-tony/server-tools/pull/1

espo-tony:
@StefanRijnhart 
Thank you very much for your help. Unfortunately the build fails on those tests. I'll see if I can find time today to watch at the errors and try to solve them.

espo-tony:
@StefanRijnhart 
From the build log I think it could be a problem with some library used by this repository. I locally ran the tests and they work correctly.

StefanRijnhart:
I would try to upgrade postgresql first, we needed this for the 8.0 version of this module: https://github.com/OCA/server-tools/pull/743/commits/34ffedeca4383821176edaedc3fe669a4ea086a2

If that does not work, try upgrading phantomjs. In .travis.yml:

```
-before_install:
 - "export PATH=$PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH"
 - "if [ $(phantomjs --version) != '2.1.1' ]; then rm -rf $PWD/travis_phantomjs; mkdir -p $PWD/travis_phantomjs; fi"
 - "if [ $(phantomjs --version) != '2.1.1' ]; then wget https://assets.membergetmember.co/software/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2; fi"
 - "if [ $(phantomjs --version) != '2.1.1' ]; then tar -xvf $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis_phantomjs; fi"
 - "phantomjs --version"
```

StefanRijnhart:
Interesting! With the phantomjs update, the test works in principal but there is a functional problem where the waiting condition $('button.o_list_button_add').length never becomes true for user demo. To see if this is caused by an interaction with other modules, maybe you could test this module on Travis in isolation using

TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="base_import_module_security_group"

StefanRijnhart:
Thank you for your continuous work here! I see you are adding the demo user to the group of erp managers. What I noticed when I was working with this test myself was that the state of the database in phantomjs does not reflect such changes. It is probably running in another transaction.

Indeed, if the add button is not showing up on the list view, that could be because the demo user does not have creation rights on the partner category model. Which would be a mystery in itself.

Would you mind if I do some tests myself in another dummy PR?

espo-tony:
@StefanRijnhart 
Thank you for your precious help. With my latest commit it seems I managed to get a working build, and indeed the problem was the demo user not having the rights for partner creation. I solved evaluating the waiting condition against a different element which is always present.
The weird thing, as you also remarked is that on my local environment the tests were properly working, meaning that the demo user has a different access right configuration between the build here and the one I was testing locally (though, it's also weird that even after an explicit assignation of the right group, the demo user on the build wasn't able to see the create button).

Feel free to add more tests (I think in the next days I'll also refactor the existing test) if you wish.
Again thank you for your help.

StefanRijnhart:
Oh, I see you changed the action to the res users action https://github.com/espo-tony/server-tools/pull/1/files#diff-42657ea13cced82fd0efda56ecf8fef0R29. The demo user indeed does not have the required permissions for that, which is why I used the partner category action: https://github.com/espo-tony/server-tools/pull/1/files#diff-42657ea13cced82fd0efda56ecf8fef0R29. Do you want to try that one again?

StefanRijnhart:
And yes, explicit assignment of groups in the test class itself is not reflected in the phantomjs environment in 10.0 (it works in 8.0), which makes testing with phantomjs quite a bit less interesting IMHO.

espo-tony:
Yes, it was one of my first attempts to make this test working (when the problem still was the Phantom JS version) and it only introduced a new issue. Indeed using the partner category action the waiting condition works also on the create button. Still I think, also for the future, I quite prefer the other condition, as the element _div.o_view_manager_content_ is always loaded regardless of the user and its groups.

About the explicit assignment of groups in the test class I wonder if making explicit use of the write method would solve the problem, or if the creation of a new user (passing the group in the create parameters) would do it.

StefanRijnhart:
But if you test for div.o_view_manager_content and then check for the missing button, you are actually testing if the user has create permission instead if import permission. So I like it better as it is.

Congratulations with your green branch!

espo-tony:
The test still stays on the import button ( _button.o_button_import_ ), but as the test should run only after the target page is loaded, we have to use a ready criteria, which is the 3rd parameter of the function _HttpCase.phantom_js_.
As we don't care, for the purposes of this test (and talking in general, for the purposes of most tests), about if the user has the create permission on the object, I think a more appropriate ready criteria would be using a surely loaded element of the page rather than the create button. Why you say that doing so we would test if the user has create permission instead of import permission?
I'll leave the ready criteria on this test as it is, by the way.

espo-tony:
@pedrobaeza 
Can we merge this?

pedrobaeza:
Please squash the commits per logical set/author.

espo-tony:
@pedrobaeza 
Squash done

pedrobaeza:
On 8.0 PR, we talked about a module renaming to _base_import_restrict_ or similar, which also follow. I still see that name better, although I'm not going to block the PR for this reason. What do you think?

On the other side, README talks about CSV, but now Odoo also imports Excel files, and there's an outdated reference to help forum which was removed from template. Can you check this?

espo-tony:
@pedrobaeza  Thanks for signaling. I updated the module in order to comply with the latest OCA conventions.

About the name, while I find both names well fitting, I think that it would be better to keep the _base_import_security_group_ because it was also already merged in v8 with that name.
Moreover, browsing through the 8.0 PRs #154 #743 I see that a lot of discussion was done about this module name and that after multiple changes it was agreed to call it _base_import_security_group_. What do you think?

StefanRijnhart:
@pedrobaeza I think, especially in a late stage during the review, it would be wise to hold back a little bit with requests to change module names in PRs. When a PR is open for a longer time, people might already be using these modules in production.

(BTW it was suggested by *you* in in this *very thread* to change the name to base_import_security_group https://github.com/OCA/server-tools/pull/694#issuecomment-271568143. We did update the name then in this PR and in the PR for 8.0)

pedrobaeza:
Yes, you're totally right. Sorry for the noise. I only read partially the 8.0 PR, without seeing the other PR and these comments. I always try to be homogeneous with the current conditions, but that conditions may change through time or be incorrectly read like I did.

StefanRijnhart:
@pedrobaeza alright, thanks for the merge!

espo-tony:
Thank you all!

PR #698, opened 2017-01-13
[9.0] Add Spanish translation and sort imports

arkhan:


arkhan:
for password_security

pedrobaeza:
Use Transifex for making the translations

PR #700, opened 2017-01-16
9.0 mig qweb_usertime

GillesTephaneMeyomesse:


pedrobaeza:
Welcome to OCA, Gilles! Does Acsone an ECLA (Entity CLA) signed or are they individual?

pedrobaeza:
Looking at the module, maybe its functionality is duplicated with datetime_formatter. Have you tried it? Anyway, this module should go better to OCA/reporting-engine I think.

lmignon:
@pedrobaeza IMO this functionality is not provided by `datetime_formatter` This module adds a new qweb expression to get a 'datetime.now' value formatted according to  an expected format. MOREOVER the datetime value is rendered by taking care of the user timezone.

pedrobaeza:
OK, no problem. How do you see to move the module to OCA/reporting-engine with the name _report_qweb_usertime_? This repo is quite crowded and the goal is very closed to the other repo.

lmignon:
@pedrobaeza The module can be used everywhere where QWEB is used. (website, report, ...).  Therefore `OCA/reporting-engine` is not the right place for this module and the name should be preserved.

pedrobaeza:
OK, but AFAIK, Odoo web client already handles user tz in datetime fields. I don't get the user case, but I'm not going to block this anymore because you are going to really hate me :tongue:

lmignon:
@pedrobaeza If you put `datetime.now()` into a qweb `t-esc` element , the tz will not be taken into account because the result of `datetime.now` on the server is UTC. But when you add a `t-field` in a report or a webpage, the qweb template engine will automatically take into account the locale... https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_qweb/fields.py#L192

lmignon:
@pedrobaeza I'll recheck if we can find a 'standard way' to correctly display datetime.now in a report...

pedrobaeza:
Thanks for the explanation. I get the user case: the typical "printing time" in the footer. I'm going also to investigate if there's a standard way of doing that as well.

GillesTephaneMeyomesse:
@pedrobaeza After some more investigations and a retry of an unsuccessful attempt, we finally found the standard way to do it without an additional module :
``` xml
 <t t-esc="context_timestamp(datetime.datetime.now()).strftime('%H:%M:%S')"/>
```
Therefore we close this PR

pedrobaeza:
Cool that you found it! Then I remove the module from the repo and put the explanation in the migration issue.

PR #701, opened 2017-01-16
[FIX] Request id no longer exists after concurrency rollback

StefanRijnhart:
The request id stored on the http request can be rolled back when a concurrency error occurs. Upon retry, the id does not exist anymore.

Related logs look like this:
```
  File "/opt/odoo/parts/odoo/openerp/sql_db.py", line 234, in execute
TransactionRollbackError: could not serialize access due to concurrent update
2017-01-16 05:31:59,837 25491 INFO odoo [-][-][-] openerp.service.model: SERIALIZATION_FAILURE, retry 1/5 in 0.3513 sec...
[...]
IntegrityError: insert or update on table "auditlog_log" violates foreign key constraint "auditlog_log_http_request_id_fkey"
DETAIL:  Key (http_request_id)=(2263) is not present in table "auditlog_http_request".
```

### DIFF
--- a/auditlog/models/http_request.py
+++ b/auditlog/models/http_request.py
@@ -2,6 +2,7 @@
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from psycopg2.extensions import AsIs
 from openerp import models, fields, api
 from openerp.http import request
 
@@ -47,7 +48,14 @@ class AuditlogHTTPRequest(models.Model):
         httprequest = request.httprequest
         if httprequest:
             if hasattr(httprequest, 'auditlog_http_request_id'):
-                return httprequest.auditlog_http_request_id
+                # Verify existence. Could have been rolled back after a
+                # concurrency error
+                self.env.cr.execute(
+                    "SELECT id FROM %s WHERE id = %s", (
+                        AsIs(self._table),
+                        httprequest.auditlog_http_request_id))
+                if self.env.cr.fetchone():
+                    return httprequest.auditlog_http_request_id
             vals = {
                 'name': httprequest.path,
                 'root_url': httprequest.url_root,


auditlog/models/http_request.py
```diff
@@ -2,6 +2,7 @@
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from psycopg2.extensions import AsIs
```
sebalix, line 5: One line space between third modules and Odoo imports?

sebalix (approved):
Minor change, otherwise :+1:

auditlog/models/http_request.py
StefanRijnhart: Mmm, this is not linted and it's not even in the guidelines. https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports only explicitely mentions the order of the imports, even if it uses newlines to clarify the grouping in the example. Seems like a cornercase to me, no?

sebalix: Yes it is, that's why I am not blocking the PR ;) I'm used to do this to clarify the Python dependencies (like the example you linked), and it must be difficult for a quality checker to guess that I think.

moylop260 (approved):
LGTM
I didn't know AsIs method to avoid sql injection from table section
Thanks!

auditlog/models/http_request.py
moylop260: In the guidelines recommend us run `isort`
This lint detect this case and auto-fix.
But it is optional.
If you don't want is goos to us

sebalix:
@StefanRijnhart Do you think this error could occurred on 10.0? To know if there is an interest to include it in https://github.com/OCA/server-tools/pull/707

StefanRijnhart:
I would be surprised if it did not occur in 10.0, so please have it included there.

sebalix:
Two approvals, is someone available to merge this PR please ? ;)

PR #598, opened 2016-11-10
Migrating MassEditing to v10

JayVora-SerpentCS:

PR #702, opened 2017-01-16
[MIG][mass_editing] from 9.0 to 10.0

JonathanNEMRY:


pedrobaeza:
There's already another PR for the migration of this module: https://github.com/OCA/server-tools/pull/598. Have you checked it?

JonathanNEMRY:
Thanks! I didn't notice it! I will propose the difference between both into it

-jne

PR #703, opened 2017-01-18
[10.0][ADD] auth_totp: MFA Support

lasley:
Forward port of #687

MFA Support
===========

This module adds support for MFA using TOTP (time-based, one-time passwords). 
It allows users to enable/disable MFA and manage authentication apps/devices 
via the "Change My Preferences" view and an associated wizard. 

After logging in normally, users with MFA enabled are taken to a second screen 
where they have to enter a password generated by one of their authentication 
apps and are presented with the option to remember the current device. This 
creates a secure, HTTP-only cookie that allows subsequent logins to bypass the 
MFA step.

Installation
============

1. Install the PyOTP library using pip: ``pip install pyotp``
2. Follow the standard module install process

Configuration
=============

By default, the trusted device cookies introduced by this module have a 
``Secure`` flag and can only be sent via HTTPS. You can disable this by going 
to ``Settings > Parameters > System Parameters`` and changing the 
``auth_totp.secure_cookie`` key to ``0``, but this is not recommended in 
production as it increases the likelihood of cookie theft via eavesdropping.

Usage
=====

Install and enjoy.


lasley:
Added PR review fixes from #687 & rebased on current 10.0. This is ready for review

dreispt:
Great module! The logo should be changed to a company neutral one.

auth_totp/README.rst
```diff
@@ -0,0 +1,105 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl.html
+   :alt: License: LGPL-3
+
+===========
+MFA Support
```
pedrobaeza: Is it support for MFA (multi-factor authentication) or TOTP? Explain anyway the letters.

dreispt:
I tested this locally and found a problem: I can't activate the "MFA Enabled?" because it's not editable, even though I successfully added a device. What am I doing wrong?

image


Also, shouldn't the regular User form have these new fields available, so that the Administrator can, for example, revoke access to devices?

auth_totp/README.rst
lasley: TOTP is an implementation of MFA. This should probably be:
```
MFA Support Using TOTP
```

lasley:
@obulkin - please take a look at https://github.com/OCA/server-tools/pull/703#issuecomment-301459734

Also - I've noticed that the LasLabs icon has been in a lot of your submissions (looks like it made it to v9 this time too). Has this since been resolved now that you have switched template repos?

obulkin:
@dreispt - Good catch on the bug. It was a permissions issue stemming from the fact that non-admin users aren't allowed to write to `res.users` records. There is a [workaround](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L180) that makes this possible when users are editing their own record, which I've expanded to include some MFA fields.

As for the regular user form, the MFA fields are not present there because the module wasn't designed to give admins access to MFA settings. I don't recall the reasoning for this, and it seems like a good feature to have, so I'll go ahead and add it.

auth_totp/models/res_users.py
```diff
@@ -13,6 +13,12 @@
 class ResUsers(models.Model):
     _inherit = 'res.users'
 
+    @classmethod
+    def _build_model(cls, pool, cr):
+        ModelCls = super(ResUsers, cls)._build_model(pool, cr)
+        ModelCls.SELF_WRITEABLE_FIELDS += ['mfa_enabled', 'authenticator_ids']
+        return ModelCls
+
```
dreispt, line 21: This depends on an implementation detail and not on the API.
What is the point? Just make the fields writeable?

obulkin: Is there some other mechanism for controlling access to a subset of a model's fields? The idea is that a user should be able to edit her email info or MFA authenticators but not other fields on her record, like the groups that she belongs to.

dreispt: I feel that we can avoid the `_build_model` internal method, and achieve the same effect extending the `write` method instead. What do you think?

obulkin: That would work too, but the new logic would now be triggered every time there's a write operation as opposed to once per registry, which doesn't seem like a good tradeoff.

FWIW, Odoo SA code has a couple examples of this being done in ``__init__``, but all the custom ``__init__`` logic has been [deprecated in v10](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=SELF_WRITEABLE_FIELDS&type=) and ``_build_model`` is [now calling ``__init__`` just for legacy support](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L576), which is what led me to put it in ``_build_model``.

dreispt: OK, leave as it is then.

dreispt:
Let me know when you're done with the fixes.

obulkin:
@dreispt - The fixes are in place, and I've added view changes and record rules that should allow admins to toggle MFA and read/unlink authenticators. This addresses all of the PR comments so far.

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    user_id = fields.Many2one(
+        comodel_name='res.users',
+        default=lambda s: s._default_user_id(),
+        required=True,
+        string='Associated User',
+        help='This is the user whose account the new authentication app/device'
+             ' will be tied to',
+        readonly=True,
+        index=True,
```
dreispt, line 49: Add `ondelete='cascade'` to avoid this:
image

dreispt:
Tested locally: I activated MFA for a user, logged out, logged in, and on the "MFA Confirmation Code" page I get this error:
```
2017-05-26 13:26:08,599 21359 INFO oxide-env werkzeug: 10.8.16.43 - - [26/May/2017 13:26:08] "POST /auth_totp/login HTTP/1.1" 500 -
2017-05-26 13:26:08,601 21359 ERROR oxide-env werkzeug: Error on request:
  File "/home/dr/work/oxide-env/lib/python2.7/site-packages/werkzeug/serving.py", line 209, in run_wsgi
    execute(self.server.app)
  File "/home/dr/work/oxide-env/lib/python2.7/site-packages/werkzeug/serving.py", line 199, in execute
    for data in application_iter:
  File "/home/dr/work/oxide-env/lib/python2.7/site-packages/werkzeug/wsgi.py", line 704, in __next__
    return self._next()
  File "/home/dr/work/oxide-env/lib/python2.7/site-packages/werkzeug/wrappers.py", line 81, in _iter_encoded
    for item in iterable:
TypeError: 'Response' object is not iterable
```

auth_totp/wizards/res_users_authenticator_create.py
obulkin: Noted. Seems like the default for wizard models should just be ``cascade`` or ``set null``, not ``restrict``.

lasley: Maybe we should add a pylint for this?

cc @moylop260

obulkin:
@dreispt - I can't replicate the Werkzeug error locally or on Runbot. Any chance you can give me more info?

auth_totp/wizards/res_users_authenticator_create.py
moylop260: What is the new pylint, sorry?

lasley:
@dreispt - I'm also not able to replicate on Runbot. Which user were you using?

Oneirag:
Hi all
I tried to use in odoo 10 the branch for odoo 9 but it does not work in the website. I check the code of this pull request and I guess it would not work either. So I would recommend two tiny changes for the code to make it work:
First, in __manifest__.py add the website dependency
`'depends': [
         'report',
         'web',
            'website',
     ],`
and then in controllers/main.py add website=True to the decorator of mfa_login_get
`    @http.route('/auth_totp/login', type='http', auth='none', methods=['GET'], website=True)
`
otherwise it would awfuly crash when rendering the the verification code page in the website.
Hope this helps.

obulkin:
Thanks, @Oneirag. I've confirmed the bug and will be adding a fix. We don't want to introduce a dependency on ``website`` here so it'll likely come in the form of a glue module.

Oneirag:
Hi all. 
I removed the `website `dependency in `__manifest__.py` but kept the change in the decorator of `mfa_login_get`. For me it works for me both with and without the `website` module installed.
If you can also try it maybe problem solved!

obulkin:
@Oneirag - I did give that a try and it generally works because the decorator seems to ignore the ``website`` argument if that module is not installed. However, this means that if ``auth_totp`` is installed first and ``website`` is added afterwards, the MFA login process will be broken until the server is restarted. This isn't a great outcome, so the glue module solution seems like the only option here.

obulkin:
The ``website`` compatibility fix is up now, and the PR should be ready for review again.

@Oneirag - In case you're curious, a glue module was ultimately not required. The error I mentioned had to do with order of inheritance and the fact that ``website`` also overloads the same [login method](https://github.com/odoo/odoo/blob/10.0/addons/website/controllers/main.py#L82), where it looks for a ``login_success`` key that ``auth_totp`` was not properly modulating.

lasley (commented):
Excellent investigation @obulkin 

Does the website bug also affect v9? If so, let's cherry pick and submit a PR

Oneirag:
Hi all
I tested and indeed it works. Thank you very much for your efficent works guys and looking forward to helping in any other thing if needed!

lasley:
Thanks for the confirmation @Oneirag. Please accept the PR using the Github review feature in the Files Changed tab so that we can officially count the approval. Thanks!

obulkin:
@lasley - It does. I've created a PR that brings all of these bug fixes and the admin improvement over to v9: https://github.com/OCA/server-tools/pull/898

Oneirag (approved):
It works for me with and without website module installed

hbto:
Hello community,

A long time ago I just begun a similar module which was a pet project just for fun
In there I tried several approaches for TFA/MFA.

from using OTP / GridCard 
Twilio / Whatsapp.

Something that I just realized about MFA/TFA is that the seeds to generate OTPs 
are stored in the same table/database where the data that is intended to be protected
is stored, 

in order to have a stronger implementation of MFA you have to separate those two
which I know will render it more complex but that is the price of MFA.

No security measure come without a price and price is cut from usability for
end user and complexity for those in charge of deploying.

a few days ago a colleague of mine @oscarolar pointed me out to this [post](https://www.theverge.com/2017/7/10/15946642/two-factor-authentication-online-security-mess)

I just remembered this scene, you want to protect your house:
- just raise a fence.
- buy a new door / MultiLock One.
- realize your dog has to go out to make poo.
- get tired with opening and closing the doors, 
- and open a new tiny door for your dog which requires no security.

if you realize what is the morale on this scene then you are with me on my concerns on 
deploying properly MFA.

You have to come to realize that nobody or the minimum amount of people
could/should be able to handle seeds generated for MFA.
and that you don't store all your eggs in the same basket.

in the next comment I will bring up my pet project so that we compare 
what was done there, my mistakes/concepts and try provide more help on this project.

Regards.

dreispt:
I'm sorry, I just tested it on Runbot and get an error after submitting the "MFA Confirmation Code".

2017-07-13 10_02_31-500 internal server error


I disabled the "auth_totp.secure_cookie" parameter, craetes a "test" account, logged in, registered my phone, and then tried to login.

lasley:
@hbto - Thank you for your comment, I'd love to see what you have. Regarding the security, I agree that this implementation is not secure by default.

Final plans for this module will leverage #673, in which I provide an external encryption/decryption service powered with [Cloudflare Red October](https://blog.cloudflare.com/red-october-cloudflares-open-source-implementation-of-the-two-man-rule/). We will be using similar strategy on all our confidential data. Flow diagram is in https://github.com/OCA/server-tools/pull/673#issuecomment-294044930

obulkin:
@dreispt - I can't seem to reproduce this one either, both locally and on Runbot. If you can get it to trigger again, could you maybe SSH in and provide some more info from the logs? I wonder if you're seeing the same ghostly Werkzeug error that we couldn't replicate earlier.

lasley:
@obulkin - run logs are available https://runbot2.odoo-community.org/runbot/static/build/3298370-703-b5fc80/logs/job_30_run.txt

obulkin:
@lasley - Thanks. That's good to know. However, it doesn't really help in this case as I can't seem to recreate the failure.

lasley:
Hmmm I'm not able to replicate this either. I just created a new user, enabled MFA on my phone, logged out, logged in & was able to submit the token fine. I tried the remember this device flow as well, hoping it would maybe be a security issue.

obulkin:
@dreispt - I've pushed up a commit that should fix your Werkzeug error. Thanks to the Firefox mention in https://github.com/OCA/server-tools/issues/908, I was finally able to track it down to inconsistent behavior in [``redirect_with_hash``](https://github.com/odoo/odoo/blob/10.0/odoo/http.py#L159), which returns a ``Response`` object when the browser is Firefox and a plain string in every other case.

grindtildeath:
@obulkin Works well with your last commit on Firefox :+1:

grindtildeath (approved):

dreispt (approved):

dreispt:
Thanks for the trouble fixing this @obulkin

PR #713, opened 2017-01-24
[10.0][FIX] Move DB libs to travis file

lasley:
This is a stop gap to fix build while we determine how to accomplish OCA/maintainer-quality-tools#427

* Move libraries that require development headers to travis file to band-aid build


lasley:
Hrmmm well that build failure was unexpected. It almost seems like the apt packages are being skipped in lint, but I don't see how that could be possible.

lasley:
I added a switch to not install during lint checks. Not the prettiest solution, but I believe we're good for review here.

pedrobaeza (approved):
Good for me

moylop260 (approved):

pedrobaeza:
Let's merge this one for solving the issues that are raising all across 10.0

PR #681, opened 2016-12-30
[10.0][MIG] base_kanban_stage

lasley:
This is a forward port of #679 

* Rename manifest
* Change openerp references to odoo
* Bump version

PR #716, opened 2017-01-27
[9.0][FIX] base_kanban_stage: Add group_expand attr

lasley:
Simple fix for issue described in https://github.com/OCA/server-tools/pull/681#issuecomment-275572678:

* Add group_expand attr to abstract kanban model, allowing for empty columns
* Fix `_read_group_stage_ids` for new API

cc @dreispt @obulkin

### DIFF
--- a/base_kanban_stage/models/base_kanban_abstract.py
+++ b/base_kanban_stage/models/base_kanban_abstract.py
@@ -38,6 +38,7 @@ class BaseKanbanAbstract(models.AbstractModel):
         help='The Kanban stage that this record is currently in',
         default=lambda s: s._default_stage_id(),
         domain=lambda s: [('res_model.model', '=', s._name)],
+        group_expand='_read_group_stage_ids',
     )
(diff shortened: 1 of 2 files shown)


dreispt (approved):

lasley:
Oops it seems I didn't update the tests - I noticed this in the v10 PR & have cherry picked here. I forget how important CI is until it goes away! I fixed the v9 build with #719 in light of this almost-whoopsie as well.

lasley:
In a conversation with @obulkin, we determined that I implemented this wrong for 9. I'm going to wait on our CIs then come back to this. Sorry to bring you in with it broken @dreispt!

[The way it should be done in v9](https://github.com/odoo/odoo/blob/9.0/openerp/fields.py#L304)

lasley:
Ok yup so this is irrelevant in 9, my bad. Implementation is confirmed good & I can actually remove a few more lines in the pending v10 PR. I love removing code 😄

PR #718, opened 2017-01-27
Migration of auth_signup_verify_email to 10.0

sharkende:
Module set to installable
Odoo don't let signup without password

### DIFF
--- a/auth_signup_verify_email/controllers/main.py
+++ b/auth_signup_verify_email/controllers/main.py
@@ -34,7 +34,7 @@ class SignupVerifyEmail(AuthSignupHome):
             values["email"] = values.get("login")
 
         # Remove password
-        values["password"] = "password"
+        values["password"] = ""
         sudo_users = (http.request.env["res.users"]
                       .with_context(create_user=True).sudo())
 
--- a/auth_signup_verify_email/controllers/main.py
+++ b/auth_signup_verify_email/controllers/main.py
@@ -34,7 +34,7 @@ class SignupVerifyEmail(AuthSignupHome):
             values["email"] = values.get("login")
 
         # Remove password
-        values["password"] = False
+        values["password"] = "password"
         sudo_users = (http.request.env["res.users"]
                       .with_context(create_user=True).sudo())
 
--- a/auth_signup_verify_email/controllers/main.py
+++ b/auth_signup_verify_email/controllers/main.py
@@ -3,8 +3,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
-from openerp import _, http
-from openerp.addons.auth_signup.controllers.main import AuthSignupHome
+from odoo import _, http
+from odoo.addons.auth_signup.controllers.main import AuthSignupHome
 
 _logger = logging.getLogger(__name__)
 
--- a/auth_signup_verify_email/__manifest__.py
+++ b/auth_signup_verify_email/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Verify email at signup",
     "summary": "Force uninvited users to use a good email for signup",
-    "version": "9.0.1.0.0",
+    "version": "10.0.1.0.0",
     "category": "Authentication",
     "website": "http://www.tecnativa.com",
     "author": "Antiun Ingeniería S.L., "
@@ -12,7 +12,7 @@
               "Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
-    'installable': False,
+    'installable': True,
     "external_dependencies": {
         "python": [
             "lxml",


sharkende:
Hi pedrobaeza,
Because I'm new I would like to know if there is anything more I need to do for the pull request ?
Thanks in advance.

pedrobaeza:
If you have signed the CLA, you just have to wait that at least 2 reviewers validate your work. See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#review for more information.

sharkende:
Ok thanks

Fenkiou:
Any news on this one ? Changes in this PR aren't that big, could someone take a look to validate it? @Yajo and @moylop260 were the ones that migrated it from 8.0 to 9.0.

Thanks!

pedrobaeza:
If you want this to be reviewed, review other PRs from the same repo or others and ask in exchange to review this one.

nhomar (approved):

elicoidal:
You probably have to adapt a bit the README file: for example the runbot URL and refresh with newest template.

auth_signup_verify_email/controllers/main.py
```diff
@@ -34,7 +34,7 @@ def passwordless_signup(self, values):
             values["email"] = values.get("login")
 
         # Remove password
-        values["password"] = False
+        values["password"] = ""
```
yajo, line 37: I cannot understand the reasoning behind this change. Could you explain please?

yajo (changes requested):

auth_signup_verify_email/controllers/main.py
sharkende: Well Odoo 10.0 doesn't allow to signup without password (False) but it work with an empty password/string ("").

yajo: But I have searched through the code and cannot see where it breaks, could you point me to it please? The field is not required after all...

sharkende:
@elicoidal Ok for the runbot url but the newest template ? I don't see anything different. Can you explain please ?

auth_signup_verify_email/controllers/main.py
sharkende: Well I have search it too and to be honest I don't remember where it breaks. I just know that it works :/

yajo: Hmm I don't say it doesn't work, but I suspect it works also with previous implementation, and it's not a good practice to store empty strings in fields. I don't feel comfortable skipping good practices precisely in the password field 🤔, so please elaborate this need before merging.

Fenkiou: If we use the module as is with the password set to `False`, we can't signup:

```
  File "/Users/alex/Documents/workspace/odoo/addons/10.0/server-tools/auth_signup_verify_email/controllers/main.py", line 43, in passwordless_signup
    sudo_users.signup(values, qcontext.get("token"))
  File "/Users/alex/Documents/workspace/odoo/odoo10/addons/auth_signup/models/res_users.py", line 71, in signup
    self._signup_create_user(values)
  File "/Users/alex/Documents/workspace/odoo/odoo10/addons/auth_signup/models/res_users.py", line 98, in _signup_create_user
    raise SignupError(ustr(e))
SignupError: secret must be unicode or bytes, not bool
```

Can't find why this is raised, but, as stated in the code for the password field: "Keep empty if you don't want the user to be able to connect on the system." and the default value is `''` not `False`

I think even for odoo 9, the password should be set to `''`

yajo: OH I see, thanks! 😊 It's because of the secret computation. It needs strings to work.

yajo (approved):

moylop260:
FYI continuation of this PR https://github.com/Vauxoo/server-tools/pull/90

legalsylvain:
The Merge of this PR seems to have broken Travis. (Commit n° 385e9b7b2c1ce1ffad5d047f944ee40508e65816)

https://github.com/OCA/server-tools/commits/10.0

@moylop260, @sharkende  Could you take a look ? 

thanks.

Fenkiou:
@legalsylvain I don't know how travis is able to run this module's tests as they are still using openerp package.

For me, to run the tests, I had to:
- change openerp to odoo in imports
- use @post_install(True) on the class (otherwise tests aren't run at all..)
- replace database commits with solution shared by @Yajo here : https://github.com/odoo/odoo/issues/12237#issuecomment-254781169

Also, for having the test pass in case of a running smtp server, I had to set the timeout to 20sec not sure that's relevant

florian-dacosta:
I think https://github.com/OCA/server-tools/pull/762 may fix these tests.

PR #719, opened 2017-01-27
[9.0][FIX] database_cleanup: Isolate build

### LINKED ISSUE #689: Travis build for 9.0 is red!
I [see](https://travis-ci.org/OCA/server-tools/jobs/188673110) problems in `html_image_url_extractor`, `html_text`, `auth_signup_verify_email` and `database_cleanup`.

lasley:
* Isolate `database_cleanup` into its own build in Travis file to fix #689

Admittedly I didn't RTFM of MQT, but I'm guessing this is probably the way to do it. If build fails, I'll look into it!

Note that Runbot will only have the first Tests build as the usable, so Runbot is now not actually testing the `database_cleanup` module due to this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="database_cleanup"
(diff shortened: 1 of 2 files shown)


lasley:
d72ba5b was also required, because the [associated PR](https://github.com/odoo/odoo/pull/13978/files#r88654967) was merged. We should be good now, outside of the known 🔴  on odoo/odoo

pedrobaeza (approved):
We totally forgot this!

pedrobaeza:
@dreispt please review

dreispt (approved):

lasley:
Wooo 🍏  v9!!! 🎉

pedrobaeza:
Yeah!

PR #729, opened 2017-02-08
Fixed image resize bug

wpichler:
When used with product_template for example - it will overwrite the original image with the smaller images which get automatically generated by product_template. It does not make sense here to store the smaller images - because we do resize the images on the fly !


pedrobaeza:
You are putting product as example, but does it fit in all models? This is one is the generic one, and the small and medium image is for each sub-image I think. Please check this and also the Travis status.

pedrobaeza:
This branch is incorrect with lots of commits. Please create a separate one and then commit there your change attending my comments. I close meanwhile.

PR #732, opened 2017-02-10
[10.0][ADD] base_directory_file_download

astirpe:
This module allows to view and download the files contained in a directory on the server.

### DIFF
--- a/base_directory_file_download/README.rst
+++ b/base_directory_file_download/README.rst
@@ -1,0 +1,78 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+========================
+Directory Files Download
+========================
+
+View and download the files contained in a directory on the server.
+
+This functionality can have impacts on the security of your system,
+since it allows to download the content of a directory.
+Be careful when choosing the directory!
+
+Notice that, for security reasons, files like symbolic links
+and up-level references are ignored.
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Set the group "Download files of directory" for the users who need this functionality.
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Settings -> Downloads -> Directory Content
+#. Create a record specifying Name and Directory of the server
+#. Save; a list of files contained in the selected directory is displayed
+#. Download the file you need
+#. In case the content of the directory is modified, refresh the list by clicking the button on the top-right of the form
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Andrea Stirpe <a.stirpe@onestein.nl>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/base_directory_file_download/models/ir_filesystem_file.py
+++ b/base_directory_file_download/models/ir_filesystem_file.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017-2018 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+import os
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+from odoo.tools import human_size
+
+_logger = logging.getLogger(__name__)
+
+
+class IrFilesystemDirectoryLine(models.TransientModel):
+    _name = 'ir.filesystem.file'
+
+    name = fields.Char(required=True)
+    filename = fields.Char()
+    file_content = fields.Binary(compute='_compute_file')
+    stored_filename = fields.Char()
+    directory_id = fields.Many2one(
+        'ir.filesystem.directory',
+        string='Directory'
+    )
+
+    @api.multi
+    def _file_read(self, fname, bin_size=False):
+
+        def file_not_found(fname):
+            raise UserError(_(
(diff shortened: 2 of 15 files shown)


base_directory_files_download/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+
+Known issues / Roadmap
+======================
+
+* ...
```
hbrunn: remove this section

base_directory_files_download/models/ir_filesystem_directory.py
```diff
@@ -0,0 +1,89 @@
...
+        compute='_compute_file_ids',
+        string='Files'
+    )
+    file_count = fields.Integer(compute='_file_count', string="# Files")
+
+    @api.multi
+    def get_dir(self):
+        self.ensure_one()
+        dir = self.directory
+        if dir and dir[-1] != '/':
```
hbrunn: say `os.path.join(dir, '')`

base_directory_files_download/models/ir_filesystem_directory.py
```diff
@@ -0,0 +1,89 @@
...
+        'ir.filesystem.file',
+        compute='_compute_file_ids',
+        string='Files'
+    )
+    file_count = fields.Integer(compute='_file_count', string="# Files")
+
+    @api.multi
+    def get_dir(self):
+        self.ensure_one()
+        dir = self.directory
```
hbrunn: `dir` is a python keyword, better use another name

base_directory_files_download/models/ir_filesystem_directory.py
```diff
@@ -0,0 +1,89 @@
...
+            dir += '/'
+        return dir
+
+    @api.multi
+    def _compute_file_ids(self):
+        File = self.env['ir.filesystem.file']
+        for dir in self:
+            dir.file_ids = None
+            if dir.get_dir():
+                for file in dir._get_directory_files():
```
hbrunn: `file` is a keyword too

base_directory_files_download/models/ir_filesystem_directory.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @api.multi
+    def reload(self):
+        self.onchange_directory()
+
+    @api.multi
+    def copy(self, default=None):
+        self.ensure_one()
+        default = dict(default or {}, name=_("%s (copy)") % self.name)
+        return super(IrFilesystemDirectory, self).copy(default=default)
```
hbrunn: what's the rationale in making this whole think persistent in the database with all the sync trouble you get as opposed to have a purely virtual model for the files which behave as computed field here?

hbrunn (changes requested):

base_directory_files_download/models/ir_filesystem_directory.py
astirpe: Indeed a pure virtual model would be an elegant solution. I've tried to create an abstract model for handling the files but solution didn't work. The point is that the function of the computed field (file_ids) should return a recordset. Since I'm creating recordsets, the system assumes they are persistent. Do you have any alternative idea?

hbrunn: the trick is to define an `AbstractModel` for the files, and in your compute function, use `new` as opposed to `create`: https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5000. At least in 8 and 9, this works. The records then behave as if they were newly created, but not yet saved

base_directory_files_download/models/ir_filesystem_file.py
```diff
@@ -0,0 +1,57 @@
...
+            raise Warning(_(
+                '''Error while reading file %s.
+                Maybe it was removed or permission is changed.
+                Please refresh the list.''' % fname))
+
+        self.ensure_one()
+        r = ''
+        directory = self.directory_id.get_dir() or ''
+        full_path = directory + fname
+        if not (directory and os.path.isfile(full_path)):
```
hbrunn: what a shame I missed this in the first review: There's no code that prevents me from creating a record of this type with `stored_filename` equal to `../../../etc/passwd` or the like. Use `normpath` and maybe also `realpath` and check that the result is contained by the directory the file record is linked to

base_directory_files_download/models/ir_filesystem_directory.py
astirpe: @hbrunn You suggested a great elegant solution, thank you!

Unfortunately, after many attempts, I have to say that the trick doesn't work as I was expecting.

I tried the creation of files, converting the model into an `AbstractModel` and using the method `new` instead of `create`. The mechanism itself works!

What I want is to use such mechanism in the function of computed field `file_ids`: in this case it doesn't work very well. What happens is that the lines on the form remain empty. This behavior seems a bit strange to me; in facts, if I convert the computed field `file_ids` into a non-computed field and if I run the same function (eg.: calling the function from the onchange), then the mechanism works! Another strange thing is that the `file_count` field works well in every case! It correctly displays the length of field `file_ids`, even if the lines are not displayed.

I tried it in both V9 and V10: behavior is the same.

The reason that I want the field `file_ids` remains as computed field is because I want the files are automatically recalculated every time I open the form. What do you think? Any suggestion?

hbrunn: this one got me curious, so I fiddled a bit with it myself: https://github.com/hbrunn/server-tools/tree/10.0-base_directory_files_download-AbstractModel

But well, ultimately, this isn't feasible because for a proper download link (as opposed to a data uri), we really need an id. With a some patching of the read method of the abstract model this might work, but I don't think that's worth the effort then. Sorry for making you waste time on this!

hbrunn (approved):

pedrobaeza:
Please look for a module name without plurals in it, as the guidelines state.

base_directory_files_download/models/ir_filesystem_directory.py
astirpe: Learning something interesting is never a waste of time. Thank you!

astirpe:
@pedrobaeza with a lot of fantasy I would rename it as `base_directory_file_download`. Is that ok?

yung-wang (approved):
tested on runbot and it is possible to download files from /

Please merge

base_directory_file_download/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+This functionality can have impacts on the security of your system,
+since it allows to download the content of a directory.
+Be careful when choosing the directory!
+
+Notice that, for security reasons, files like symbolic links
+and up-level references are ignored.
+
+Installation
+============
```
tarteo: This section seems a little unnecessary.

base_directory_file_download/models/ir_filesystem_file.py
```diff
@@ -0,0 +1,57 @@
...
+
+        def file_not_found(fname):
+            raise UserError(_(
+                '''Error while reading file %s.
+                Maybe it was removed or permission is changed.
+                Please refresh the list.''' % fname))
+
+        self.ensure_one()
+        r = ''
+        directory = self.directory_id.get_dir() or ''
```
tarteo: I think the 'or' here is unnecessary because in get_dir it always returns a string.

tarteo (approved):
Some nitpicking but none blocking.

astirpe:
Commits squashed

astirpe:
@pedrobaeza can this old PR be merged?

astirpe:
Thanks!

PR #733, opened 2017-02-13
Improve readme and fix dep

hparfr:
1° Readme: (in example) prefix function with _
2° Dep: add a depency to base_setup in order to use xml id "base.menu_config" in https://github.com/OCA/server-tools/blob/9.0/keychain/views/keychain_view.xml#L47

### DIFF
--- a/keychain/README.rst
+++ b/keychain/README.rst
@@ -99,7 +99,7 @@ Usage (for module dev)
 
     import random
 
-    def get_auth(self):
+    def _get_auth(self):
         keychain = self.env['keychain.account']
         if self.env.user.has_group('stock.group_stock_user'):
             retrieve = keychain.suspend_security().retrieve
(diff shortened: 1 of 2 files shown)


florian-dacosta:
:+1:

bealdav:
Please remove base, if you add base_setup. Otherwise, :+1: 
"Qui peut le + peut le moins"

hparfr:
>Please remove base, if you add base_setup. Otherwise, 👍

Fixed.

Thanks for your reviews

bealdav:
Why several commits here, for 2 lines ? --force

hparfr:
--forcing a push on a public branch is not a good thing
It breaks the history of the comments (like inline review).

tafaRU (approved):
LGTM even though I'd prefer commits to be squashed to only one.

legalsylvain:
thanks.

PR #727, opened 2017-02-07
[REF] mass_editing: migrate module to version 10.0

zaoral:
I made the minimum changes in order to run this module in 10.0.
Update README and descriptions of the repository and the module it self
to refer this module as a 10.0 module y not a 9.0 unported module.

PR #734, opened 2017-02-13
migrate module mass_editing to V10

yk2kus:
The module works in V10


LoisRForgeFlow:
It seems like this is already done in https://github.com/OCA/server-tools/pull/727

pedrobaeza:
Yes, please look before duplicating efforts.

PR #736, opened 2017-02-15
[ADD] allow creating missing indexes and purging properties

hbrunn:
this is not exactly a clean up operation in the sense of removing stuff, but in a broader sense of tidying up I think we can justify having this here. I wouldn't know to which other module to add this, and making this a standalone module seems a bit over the top to me

### DIFF
--- a/database_cleanup/views/menu.xml
+++ b/database_cleanup/views/menu.xml
@@ -56,6 +56,7 @@
             <field name="name">Create missing indexes</field>
             <field name="sequence" eval="70" />
             <field name="action" ref="cleanup_create_indexes_wizard_action" />
+            <field name="parent_id" ref="menu_database_cleanup"/>
         </record>
 
         <record model="ir.ui.menu" id="menu_purge_property">
--- a/database_cleanup/views/menu.xml
+++ b/database_cleanup/views/menu.xml
@@ -52,5 +52,12 @@
             <field name="parent_id" ref="menu_database_cleanup"/>
         </record>
 
+        <record model="ir.ui.menu" id="menu_purge_property">
+            <field name="name">Purge obsolete properties</field>
+            <field name="sequence" eval="70" />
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Well, I don't see something out of scope to have a module base_index_manager. Module database_cleanup is very heavy to be installed (it even modifies the model unlink and so on) for only managing/creating new indexes.

hbrunn:
I hope people install `database_cleanup` to do some cleanup, and then throw it away again? Let's see what other contributors say, maybe @StefanRijnhart as original author has an opinion?

StefanRijnhart:
Yes please!

StefanRijnhart (approved):

hbrunn:
@pedrobaeza @StefanRijnhart I added commits to purge properties too. The reason to lump this together in the same PR is that we'd get a lot of conflicts with two PRs after the first one is merged

database_cleanup/views/menu.xml
```diff
@@ -52,5 +52,18 @@
             <field name="parent_id" ref="menu_database_cleanup"/>
         </record>
 
+        <record model="ir.ui.menu" id="menu_create_indexes">
+            <field name="name">Create missing indexes</field>
+            <field name="sequence" eval="70" />
+            <field name="action" ref="cleanup_create_indexes_wizard_action" />
+        </record>
```
NL66278: Is this not missing the parent_id element? This will create top level menu entry...

NL66278:
Apart from the remark on the menu entry for creating indexes looks good to me, and tested backported version on 8.0. That worked well. So basically :+1:

hbrunn:
@NL66278 thanks for the pointer. Please use github's review mechanism in the future, this makes the approvals machine readable and a few scripts and filters look at this rather than the text people write

lasley:
Looks like this was approved long ago. Merging

PR #738, opened 2017-02-16
[ADD] base_domain_operator

hbrunn:
Custom domain operators
=======================

This module was written to allow developers to define custom domain operators. This can be very helpful for complicated selections or ir.rules.

Currently implemented operators
===============================

``parent_of``
  The inverse of ``child_of``: Searches for the parents of given records in the hierarchy, including the records themselves.

Usage
=====

- depend on it
- use one of the operators explained above in any of your domains

To add your own operators:
- in ``base.domain.operator``, define a function ``_operator_${your_operator}`` with decorator ``@api.model`` and signature ``leaf, expression``
- ``leaf`` is a 3-tuple of a domain proposition, ``expression`` an instance of ``openerp.osv.expression.expression``
- return a list of ``openerp.osv.expression.ExtendedLeaf`` instances that maps your operator to some expressions the original domain parser can evaluate
- note that you can use the internal ``inselect`` operator here if you ``pass internal=True`` to ``ExtendedLeaf``'s constructor
- be careful with using the ORM in those handlers, you're busy modifying a crucial part of the ORM
- take good care that you don't introduce SQL injections and other security problems

Known issues / Roadmap
======================

* given the upstream code is not exactly extension friendly, we'll have to reimplement a bunch of helper functions
* ``parent_of`` currently only supports the most basic use cases, this needs to be amended as needed
* another nice operator would be ``indomain``, probably also with a way to refer to the table's columns in the right hand side of expressions


base_domain_operator/README.rst
```diff
@@ -0,0 +1,77 @@
...
+#. depend on it
+#. use one of the operators explained above in any of your domains
+
+To add your own operators:
+
+#. in ``base.domain.operator``, define a function ``_operator_${your_operator}`` with decorator ``@api.model`` and signature ``leaf, expression``
+#. ``leaf`` is a 3-tuple of a domain proposition, ``expression`` an instance of ``openerp.osv.expression.expression``
+#. return a list of ``openerp.osv.expression.ExtendedLeaf`` instances that maps your operator to some expressions the original domain parser can evaluate
+#. note that you can use the internal ``inselect`` operator here if you pass ``internal=True`` to ``ExtendedLeaf``'s constructor
+#. be careful with using the ORM in those handlers, you're busy modifying a crucial part of the ORM
```
pedrobaeza: You can clarify that the danger is that you enter in an infinite recursion. Advice to use SQL instead.

base_domain_operator/README.rst
```diff
@@ -0,0 +1,77 @@
...
+#. take good care that you don't introduce SQL injections and other security problems
+
+Known issues / Roadmap
+======================
+
+* given the upstream code is not exactly extension friendly, we'll have to reimplement a bunch of helper functions
+* ``parent_of`` currently only supports the most basic use cases, this needs to be amended as needed
```
pedrobaeza: What do you refer about basic use cases? Can it be specified? I suppose for example that dot notation is not included.

base_domain_operator/expression.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api
+# pylint: disable=W0402
+from openerp.osv.expression import ExtendedLeaf, expression, is_operator,\
```
pedrobaeza, line 6: @moylop260 I think we should fine grain this pylint-odoo check, as even in 10.0 the expression class is here: https://github.com/odoo/odoo/blob/10.0/odoo/osv/expression.py.

base_domain_operator/expression.py
```diff
@@ -0,0 +1,62 @@
...
+    is_leaf
+
+
+class BaseDomainOperatorExtendedLeaf(ExtendedLeaf):
+    def __init__(self, leaf, model, join_context=None, internal=False):
+        if isinstance(leaf, ExtendedLeaf):
+            # TODO: think very long and hard if this introduces a security
+            # problem
+            internal = True
```
pedrobaeza, line 15: Looking at for what internal is needed, it's only used for expanding the admitted operators with `inselect` and `not inselect` (ref: https://github.com/odoo/odoo/blob/a587fde991c7b3dfdffe1310d7800c210764f866/openerp/osv/expression.py#L414). I think you should leave as is without changing it.

base_domain_operator/expression.py
```diff
@@ -0,0 +1,62 @@
...
+            leaf = leaf.leaf
+        super(BaseDomainOperatorExtendedLeaf, self).__init__(
+            leaf, model, join_context=join_context, internal=internal
+        )
+
+
+class Expression(expression):
+    def parse(self, cr, uid, context):  # pylint: disable=R8110
+        domain = []
+        env = api.Environment(cr, uid, context)
```
pedrobaeza, line 27: You can create the environment after the check (it's not too expensive, but you can save a bit or avoid side effects).

base_domain_operator/models/base_domain_operator.py
```diff
@@ -0,0 +1,75 @@
...
+from ..expression import BaseDomainOperatorExtendedLeaf, Expression,\
+    is_leaf_base_domain_operator
+
+
+class BaseDomainOperator(models.AbstractModel):
+    _name = 'base.domain.operator'
+    _description = 'Implement custom domain operators'
+
+    def _register_hook(self, cr):
+        if not isinstance(expression_module.expression, Expression):
```
pedrobaeza, line 20: Isn't there code for restoring the monkey-patching when uninstalling module?

base_domain_operator/tests/test_base_domain_operator.py
```diff
@@ -0,0 +1,29 @@
...
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp.tests.common import TransactionCase
+
+
+class TestBaseDomainOperator(TransactionCase):
+    def test_base_domain_operator(self):
+        self.env['base.domain.operator']._register_hook()
+        # parent_of with an id
+        child = self.env.ref('base.res_partner_main1')
```
pedrobaeza, line 15: I prefer to not depend on demo data as this tends to change a lot over versions and forces you to check new XML-IDs, if the data are the same...

It's very simple to create a partner hierarchy here.

base_domain_operator/models/base_domain_operator.py
```diff
@@ -0,0 +1,75 @@
...
+
+        fields = leaf[0].split('.')
+        model = self.env[expression.root_model._name]
+        field = model._fields[fields[0]]
+
+        if len(fields) > 1 or (
+                field.type not in ('many2one', 'many2many', 'one2many') and
+                field.name != 'id'
+        ):
+            raise NotImplementedError(
```
pedrobaeza, line 70: And here it is. My guess was right about the unimplemented thing :wink:

base_domain_operator/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Custom domain operators
+=======================
+
+This module was written to allow developers to define custom domain operators. This can be very helpful for complicated selections or ir.rules.
+
+Currently implemented operators
+===============================
+
+``parent_of``
+  The inverse of ``child_of``: Searches for the parents of given records in the hierarchy, including the records themselves.
```
pedrobaeza: Include here a comment saying that the operator is similar to the v10 one and can be removed when migrating the module to that version.

base_domain_operator/expression.py
hbrunn: I think being able to use `inselect` for the definition of other operators will be very helpful. The comment is meant to notify reviewers to look closely at this because of the security implications. To the best of my knowledge, this is not a problem...

base_domain_operator/expression.py
hbrunn: this environment is use in the next line already, and we definitely need one for the first operator check, so this wouldn't really save anything

base_domain_operator/models/base_domain_operator.py
hbrunn: good idea to add that, will do that some time later

simahawk (commented):
I've no experience in this but overall LGTM. I'd just include a simple example of `parent_of` in action in the README.

base_domain_operator/expression.py
moylop260: Could you help me to understand the point here?

hbrunn: I think what @pedrobaeza means is that the linter should allow importing from `osv.expression` (but keep barking at stuff from osv)
We *could* also import expression from models I think (https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L6241), but that's a bit convoluted in my opinion

moylop260: Ok I'm not sure that understand the point.

I mean, if you see the [pylint result](https://travis-ci.org/OCA/server-tools/jobs/205329573#L483) we don't have a message emitted about use `osv.expression` because we are skip it from https://github.com/OCA/pylint-odoo/pull/40

moylop260: oh, sorry
Now I saw the `pylint: disable` comment
This is a pylint-odoo issue because is a valid import...

moylop260: FYI It was fixed from https://github.com/OCA/pylint-odoo/pull/118
We didn't have `from openerp.osv.expression...` import case supported (Sorry)

hbrunn: I don't think the guy who is a driving force in CI-usage by the OCA has to apologize for anything on these concerns

daramousk:
@hbrunn Is there any help needed in order to get this merged?

hbrunn:
nobody approved this up to now. So a positive review would help

base_domain_operator/models/base_domain_operator.py
```diff
@@ -0,0 +1,108 @@
...
+            model = self.env[field.comodel_name]
+            prefix = field.comodel_name
+
+        right = leaf[2]
+        if not isinstance(right, Iterable):
+            right = [right]
+
+        return [
+            ExtendedLeaf(p, expression.root_model)
+            for p in parent_of_domain(left, right, model, prefix=prefix)
```
daramousk, line 86: Here the `right` is always a string, and `parent_of_domain` expects ids, which makes this always crash

hbrunn: do we need something like https://github.com/OCA/OCB/blob/8.0/openerp/osv/expression.py#L707 here? unfortunately, it's a nested function so we cannot just call it

hbrunn: (right will be an int when the user chooses a specific record in the search ui)

daramousk: That's what I though so too, but that is not the case either. Looking at https://github.com/OCA/OCB/blob/3e813d02cef79f3e8b9558bb5810ae2a81c8d1c3/addons/web/static/src/js/search.js#L1658 I see that it returns a list of (id, text_repr) and on https://github.com/OCA/OCB/blob/3e813d02cef79f3e8b9558bb5810ae2a81c8d1c3/addons/web/static/src/js/search.js#L2536 only the facet is taken into consideration

hbrunn: cool. please PR a fix that's resilient to have ids, so that it will behave nicely if called from code too - that's why I wrote this module in the first place

daramousk: All looks good now.

daramousk (approved):

PR #739, opened 2017-02-17
FIX warning in mass editing module when read

bealdav:
dynamic fields (fake fields) trigger a warning

This fix remove the warning

### DIFF
--- a/mass_editing/wizard/mass_editing_wizard.py
+++ b/mass_editing/wizard/mass_editing_wizard.py
@@ -183,6 +183,20 @@ class MassEditingWizard(orm.TransientModel):
             result['fields'] = all_fields
         return result
 
+    def read(self, cr, uid, ids, fields, context=None):
+        """ Without this call, dynamic fields defined in fields_view_get()
+            generate a log warning, i.e.:
+
+            openerp.models: mass.editing.wizard.read()
+                with unknown field 'myfield'
+            openerp.models: mass.editing.wizard.read()
+                with unknown field 'selection__myfield'
+        """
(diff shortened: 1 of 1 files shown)


bealdav:
ping @hparfr, @sebastienbeau

mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -183,6 +183,20 @@ def fields_view_get(
             result['fields'] = all_fields
         return result
 
+    def read(self, cr, uid, ids, fields, context=None):
+        """ Without this call, dynamic fields defined in fields_view_get()
+            generate a log warning, i.e.:
```
legalsylvain, line 188: I don't understand how to add dynamic fields in fields_view_get, do you have some code sample ? 
thanks.

hparfr:
👍

mass_editing/wizard/mass_editing_wizard.py
bealdav: In the wizard:
https://github.com/OCA/server-tools/blob/8.0/mass_editing/wizard/mass_editing_wizard.py#L31-L184

It's the way the module do to update data

sebastienbeau:
:+1: ok for me too, indeed as we create fake field (in field_view_get we need to remove this fake field from the read to avoid useless warning

bealdav:
#741 test with unit test and similar PR but for v10. Tests results are OK. I'm confident for this one too

chafique-delli:
it's ok for me :+1:

Issue #753, opened 2017-02-28
Letsencrypt fails to fetch key for all domains, if one alternate domain deleted.


NL66278:
Lets say you have several system parameters to use additional domains with letsencrypt:
letsencrypt.altname.1 - info@acme.com
letsencrypt.altname.2 - support@acme.com
letsencrypt.altname.3 - sales@acme.com
letsencrypt.altname.4 - conference@acme.com
letsencrypt.altname.5 - finance@acme.com

Now if you determine you no longer need support@acme.com and you delete the parameter with name letsencrypt.altname.2, letsencrypt will no longer include sales.., conference and finance in the certificate either.

This is due to the following code in models/letsencrypt.py:
```
 96     @api.model                                                                 
 97     def generate_csr(self, domain):                                            
 98         domains = [domain]                                                     
 99         i = 0                                                                  
100         while self.env['ir.config_parameter'].get_param(                       
101                 'letsencrypt.altname.%d' % i):                                 
102             domains.append(                                                    
103                 self.env['ir.config_parameter']                                
104                 .get_param('letsencrypt.altname.%d' % i)                       
105             )                                                                  
106             i += 1                                                             
107         _logger.info('generating csr for %s', domain)
```
A workaround would be to renumber all the altname parameters to make the numbering without gaps again. No need to explain this is cumbersome and easily overlooked. 
I think the solution would be to use all parameters that start with letsencrypt.altname.

NL66278:
See PR: https://github.com/OCA/server-tools/pull/755

pedrobaeza:
Already solved in PRs, so I close.

hbrunn:
Note: I regret my choice having this in different parameters. For subsequent versions, we should have a key `letsencrypt.altnames` whcih simply contains a space separated list of domains. For a transition phase, we can have both of course (or implement only this and write a migration)

PR #756, opened 2017-02-28
[9.0][FIX] Make letsencrypt resilient for alternate name removal.

NL66278:
Solve issue #753


pedrobaeza (approved):
Merging as well with the same reasoning.

PR #757, opened 2017-02-28
[10.0][FIX] Make letsencrypt resilient for alternate name removal.

NL66278:
Sole #753 even before module is migrated.


NL66278:
Seems Travis is failing for all 10.0 branches. No idea why, because I do not see any logs or other pointers at what might be wrong.

NL66278:
Now I see errors. Test on mail fails, because travis runs in environment were mail is not configured. So maybe we should either find a way to configure access to a mail server for testing purposes, or disable these tests on travis.

lasley (approved):
Code LGTM. I restarted Travis to hopefully clear the build

sebalix (approved):

Issue #758, opened 2017-03-01
[10.0] Travis broken


legalsylvain:
The recent merge of https://github.com/OCA/server-tools/pull/718 broke travis.

commit number : 385e9b7b2c1ce1ffad5d047f944ee40508e65816

see last commits history.
https://github.com/OCA/server-tools/commits/10.0

florian-dacosta:
@legalsylvain 
https://github.com/OCA/server-tools/pull/762 seems to fix travis

lasley:
Confirmed fixed, closing!

PR #759, opened 2017-03-01
[10.0] [FIX] auditlog: Duplicate logging

obulkin:
* Update ``_patch_methods`` and ``_revert_methods`` in ``auditlog.rule`` model to properly track whether a method has already been patched for logging purposes. This prevents duplicate logs from being produced in various cases (e.g. when a logging rule is created as part of a module install).

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -131,28 +131,28 @@ class AuditlogRule(models.Model):
             if getattr(rule, 'log_create') \
                     and not hasattr(model_model, check_attr):
                 model_model._patch_method('create', rule._make_create())
-                setattr(model_model, check_attr, True)
+                setattr(type(model_model), check_attr, True)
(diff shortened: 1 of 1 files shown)


obulkin:
Take a look, @lasley. The Travis errors are coming from the ``auth_signup_verify_email`` module.

lasley:
Travis red is due to #758

lasley (approved):
Thanks @obulkin 

This applies on the previous versions too, yeah?

obulkin:
@lasley - I believe so. In essence, the problem is that the attributes used to track auditlog monkey patches were being set at the recordset level, while the patching itself was of course happening at the class level.

lasley:
I rebased onto current 10.0, which should clear this build.

sebalix (approved):
:+1:

PR #762, opened 2017-03-06
[10.0] Fix base_kanban_stage tests

florian-dacosta:
I think there is a problem with base_kanban_stage tests, as it never leaves the test mode : 
https://github.com/OCA/server-tools/blob/10.0/base_kanban_stage/tests/test_base_kanban_abstract.py#L47

Tests in other module may fail because of that.

### DIFF
--- a/base_kanban_stage/tests/test_base_kanban_abstract.py
+++ b/base_kanban_stage/tests/test_base_kanban_abstract.py
@@ -48,6 +48,11 @@ class TestBaseKanbanAbstract(SavepointCase):
         cls._init_test_model(BaseKanbanAbstractTester)
         cls.test_model = cls.env[BaseKanbanAbstractTester._name]
 
+    @classmethod
+    def tearDownClass(cls):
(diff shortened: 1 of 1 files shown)


moylop260 (approved):

lasley:
@obulkin - please take a look

obulkin (approved):
👍  Looks good to me. ``enter_test_mode()`` makes some changes to the registry class that probably persist between test suites if they're not reversed by ``leave_test_mode()``: https://github.com/odoo/odoo/blob/10.0/odoo/modules/registry.py#L435

lasley (approved):

lasley:
Thanks @florian-dacosta

lasley:
@florian-dacosta - Did you plan on backporting to v9?

florian-dacosta:
@lasley 
No, actually I don't think the bug is present in version 9.

lasley:
Awesome, thanks again @florian-dacosta!

PR #763, opened 2017-03-06
[MIG] Migrate module attachment_base_synchronize to v10

florian-dacosta:
The tests fail because of : https://github.com/OCA/server-tools/pull/762

If I play only the test of this module, it is all ok, but if I play base_kanban_stage tests together, it fails.
Indeed, because of base_kanban_stage's tests, the cursor is already in test_mode.

### DIFF
--- a/attachment_base_synchronize/README.rst
+++ b/attachment_base_synchronize/README.rst
@@ -1,4 +1,3 @@
-
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
@@ -22,7 +21,7 @@ An example of the use of this module, can be found in the external_file_location
 Usage
 =====
 
-Go the menu Settings > Attachments
+Go the menu Settings > Technical > Database Structure > Meta Data Attachments
 
 You can create / see standard attachments with additional fields
 
@@ -30,7 +29,7 @@ You can create / see standard attachments with additional fields
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/149/8.0
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
 
 
 Known issues / Roadmap
@@ -43,13 +42,9 @@ Bug Tracker
 ===========
 
 Bugs are tracked on `GitHub Issues
-<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
 check there if your issue has already been reported. If you spotted it first,
-help us smashing it by providing a detailed and welcomed `feedback
-<https://github.com/OCA/
-server-tools/issues/new?body=module:%20
-attachment_metadata%0Aversion:%20
-8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+help us smash it by providing detailed and welcomed feedback.
 
 Credits
 =======
@@ -64,9 +59,7 @@ Contributors
 ------------
 
 * Valentin CHEMIERE <valentin.chemiere@akretion.com>
-* Sebastien BEAU <sebastian.beau@akretion.com>
-* Joel Grand-Guillaume Camptocamp
-* initOS <http://initos.com>
+* Florian da Costa <florian.dacosta@akretion.com>
 * Angel Moya <http://angelmoya.es>
 
 Maintainer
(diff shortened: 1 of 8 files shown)


attachment_base_synchronize/data/cron.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<openerp>
+<odoo>
```
lasley: `odoo noupdate="1"` & can remove data tag

attachment_base_synchronize/demo/attachment_metadata_demo.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<openerp>
+<odoo>
```
lasley: `odoo noupdate="1"` & can remove data tag

lasley (approved):
Minor comments, otherwise LGTM

florian-dacosta:
Thanks for the review @lasley 
I've updated the PR

lasley (approved):
Thanks @florian-dacosta

attachment_base_synchronize/models/attachment.py
```diff
@@ -56,8 +57,7 @@ def _compute_hash(self):
     @api.model
     def run_attachment_metadata_scheduler(self, domain=None):
         if domain is None:
-            domain = []
-        domain.append(('state', '=', 'pending'))
+            domain = [('state', '=', 'pending')]
```
Garamotte, line 60: Did you really want to change the behaviour here ?
The `('state', '=', 'pending')` tuple is now only added in the domain when the domain was `None`, instead of being always added like before.

A unit test calling this method with a domain in arguments would be fine.

Garamotte (changes requested):

cmsalmeida:
When will this be accepted?
Any prediction?

Garamotte:
This will be accepted when there are enough approves.
You can review it, test it, and approve if you find it good, or comment to ask for explanationsor changes if you find something wrong.

attachment_base_synchronize/models/attachment.py
florian-dacosta: That is wanted indeed.
The normal way is to just run the attachment at pending state, so the cron does not need any argument. But one can make a cron only with attachment at 'failed' state for instance.
So the idea is that we can ignore the generic domain `('state', '=', 'pending')` for more flexibility.

attachment_base_synchronize/README.rst
```diff
@@ -22,15 +21,15 @@ An example of the use of this module, can be found in the external_file_location
 Usage
 =====
 
-Go the menu Settings > Attachments
+Go the menu Settings > Technical > Database Structure > Meta Data Attachments
```
cmsalmeida, line 24: :+1:

attachment_base_synchronize/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_attachment_metadata_user,ir.attachment.metadata.user,model_ir_attachment_metadata,base.group_user,1,0,0,0
+access_attachment_metadata_user,ir.attachment.metadata.user,model_ir_attachment_metadata,,1,0,0,0
+access_attachment_metadata_user,ir.attachment.metadata.user,model_ir_attachment_metadata,base.group_no_one,1,1,1,1
```
cmsalmeida, line 3: They have same key, I've fixed that in #856 because it needed to pass tests.

cmsalmeida (approved):
It's running well.
Tested together with external_file_localtion in #856 
:+1:

paulius-sladkevicius:
Took the same path like @cmsalmeida :+1:

attachment_base_synchronize/security/ir.model.access.csv
lasley: What is the change you are recommending that @florian-dacosta  perform here?

Edit: (sorry for the accidental ping @sylvain-garancher)

attachment_base_synchronize/models/attachment.py
lasley: I agree with @florian-dacosta on this one. An alternative to changing the behavior is to use a context override to stop the hard coded domain, but IMO explicitly adding is better than explicitly removing.

lasley:
@sylvain-garancher - have your comments been attended to?

Garamotte (approved):
Yes :+1:

attachment_base_synchronize/security/ir.model.access.csv
florian-dacosta: @lasley There is a mistake because 2 security rules have the same id. @cmsalmeida fixed it in the PR of external_file_location migration. I guess it is enought, as for now external_file_location is the only module depending on this one.

lasley: Thanks for the elaboration @florian-dacosta. Works for me too!

PR #767, opened 2017-03-09
[10.0] FIX auto_backup SFTP cleanup

eLBati:
2017-03-09 13:07:36,067 25649 ERROR jdeal_2017-03-09 odoo.addons.auto_backup.models.db_backup: Cleanup of old database backups failed: %s
  File "server-tools/auto_backup/models/db_backup.py", line 242, in cleanup_log
    yield
  File "server-tools/auto_backup/models/db_backup.py", line 232, in cleanup
    remote.unlink(name)
  [..]
IOError: [Errno 2] No such file

### DIFF
--- a/auto_backup/models/db_backup.py
+++ b/auto_backup/models/db_backup.py
@@ -232,7 +232,7 @@ class DbBackup(models.Model):
                         for name in remote.listdir(rec.folder):
                             if (name.endswith(".dump.zip") and
                                     os.path.basename(name) < oldest):
-                                remote.unlink(name)
+                                remote.unlink('%s/%s' % (rec.folder, name))
 
     @api.multi
     @contextmanager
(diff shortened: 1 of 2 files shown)


auto_backup/models/db_backup.py
```diff
@@ -232,7 +232,7 @@ def cleanup(self):
                         for name in remote.listdir(rec.folder):
                             if (name.endswith(".dump.zip") and
                                     os.path.basename(name) < oldest):
-                                remote.unlink(name)
+                                remote.unlink('%s/%s' % (rec.folder, name))
```
sebalix, line 235: ~~Maybe we could use `posixpath.join(rec.folder, name)` instead?-~~

sebalix (approved):
Nevermind, it seems that paramiko which is behind pysftp assumes that the separator `/` is always used.

StefanRijnhart (approved):

PR #773, opened 2017-03-10
[8.0] Base bypass record rules by context

angedras:
Bypass permissions on fields visibility by field context

### DIFF
--- a/base_bypass_record_rules_by_context/README.rst
+++ b/base_bypass_record_rules_by_context/README.rst
@@ -1,0 +1,78 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================
+Bypass Record Rules
+===================
+
+This module bypasses permissions on fields visibility by field context.
+For example in a multi-company environment it allows to see the records of
+company B for a user of company A, even if the user has no visibility on
+these records (i.e. B is not an allowed company).
+
+Installation
+============
+
+No specific requirements.
+
+Configuration
+=============
+
+Add a context ref in a model view for the desired field with the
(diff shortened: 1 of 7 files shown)


base_bypass_record_rules_by_context/README.rst
```diff
@@ -0,0 +1,78 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
eLBati, line 51: Please add the icon

eLBati (approved):
thanks @angedras , just a remark

base_bypass_record_rules_by_context/models/ir_rule.py
```diff
@@ -0,0 +1,18 @@
...
+
+from openerp import api, models
+
+
+class IrRule(models.Model):
+    _inherit = 'ir.rule'
+
+    @api.model
+    def domain_get(self, model_name, mode='read'):
+        if 'bypass_record_rules' in self._context:
```
hbrunn, line 14: this basically turns off record rules, because anybody can set contexts. Very bad idea. I think this module is conceptually wrong and shouldn't be merged.

hbrunn (changes requested):

base_bypass_record_rules_by_context/models/ir_rule.py
eLBati: @hbrunn true. let us see if we can make it more solid

dreispt:
I have the same security concern as Holger.
A solution might be to have a flag on Record Rules, to allow them to be ignored based on a context marker.

PR #775, opened 2017-03-13
10.0 add base exception

mourad-ehm:
Split module sale_exception into 2 modules base_exception and sale exception (see https://github.com/OCA/sale-workflow/issues/408).


base_exception/README.rst
```diff
@@ -0,0 +1,65 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+Sale Exception
```
florian-dacosta: s/Sale Exception/Base Exception

base_exception/README.rst
```diff
@@ -0,0 +1,65 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
```
florian-dacosta, line 60: Whole README must be updated (description, version number, etc...)

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+    _order = 'active desc, sequence asc'
+
+    name = fields.Char('Exception Name', required=True, translate=True)
+    description = fields.Text('Description', translate=True)
+    sequence = fields.Integer(
+        string='Sequence',
+        help="Gives the sequence order when applying the test")
+    rule_group = fields.Selection(
+        [],
+        help="Rule group is used the group the rules that must validated "
```
florian-dacosta: s/Rule group is used the group/Rule group is used to group

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+    def detect_exceptions(self):
+        """returns the list of exception_ids for all the considered base.exceptions
+        """
+        exception_obj = self.env['exception.rule']
+        model_exceptions = exception_obj.sudo().search(
+            [('model', '=', self._name)])
+        sub_exceptions = exception_obj.sudo().search(
+            [('rule_group', '=', self.rule_group),
+             ('id', 'not in', model_exceptions.ids),
+             ])
```
florian-dacosta: Why have 2 search here? Instead of just one on field rule_group
Is it better for performance?

florian-dacosta (commented):
Small changes to do, otherwise, it seems good!

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @api.multi
+    def detect_exceptions(self):
+        """returns the list of exception_ids for all the considered base.exceptions
+        """
+        exception_obj = self.env['exception.rule']
+        model_exceptions = exception_obj.sudo().search(
+            [('model', '=', self._name)])
+        sub_exceptions = exception_obj.sudo().search(
+            [('rule_group', '=', self.rule_group),
```
florian-dacosta: self may be a recordset (when called from cron test_all_draft_orders from sale_exception for example), it will fail with the 'expect singlton' error.

base_exception/models/base_exception.py
mourad-ehm: Hi @florian-dacosta , We have need of exceptions applied to model (i.e sale order) and sub exceptions applied to line (i.e sale order line). 
I change the search to make 1 search then filter data on the result.

florian-dacosta: @mourad-ehm You are right, we need to have it separated, so it is ok to have 2 search, I do not think you have to change then, my mistake!

base_exception/README.rst
```diff
@@ -0,0 +1,60 @@
...
+==============
+
+This module provide an abstract model to manage customizable exceptions to be applied on different models (sale order, invoice, ...). It is not usefull for itself. You can see an exemple of implementation in the 'sale_exception' module. (sale-workflow repository).
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/167/10.0
```
florian-dacosta: It is not the right repo id, for server-tools it is 149

florian-dacosta (commented):
One last very small comment and it is good for me!

base_exception/README.rst
mourad-ehm: OK thanks.

florian-dacosta (approved):
:+1:  code review + tests (with module sale_exception)

bealdav:
Don't you think than .po must be removed ?

mourad-ehm:
Hi all, 
@jgrandguillaume @pedrobaeza @TheCloneMaster @atchuthan @sbidoul @bguillot 
@yvaucher  @rvalyi, @davbea 
What do you think about translations ? They come from sale_exception and several translated text are the same. 
If we replace, in all po file, sale_exception by base_exception, will it be useful for transsiflex to update the translation files?
If not, there is and other mean to reuse this translations ?

base_exception/i18n/am.po
```diff
@@ -0,0 +1,227 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# * sale_exception
+# 
+# Translators:
+msgid ""
+msgstr ""
+"Project-Id-Version: sale-workflow (9.0)\n"
```
angelmoya: Maybe could be better to generate new po files, and update this lines related with module sale_exception on project sale-workflow instead of base_exception on server-tools.

angelmoya:
:+1:

base_exception/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2011 Raphaël Valyi, Renato Lima, Guewen Baconnier, Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{'name': 'Exception Rule',
+ 'version': '10.0.1.0.0',
+ 'category': 'Generic Modules/Sale',
```
bealdav: missing 'summary' key

base_exception/README.rst
```diff
@@ -0,0 +1,60 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+Base Exception
+==============
+
+This module provide an abstract model to manage customizable exceptions to be applied on different models (sale order, invoice, ...). It is not usefull for itself. You can see an exemple of implementation in the 'sale_exception' module. (sale-workflow repository).
```
bealdav: s/exemple/example

base_exception/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+ 'author': "Akretion, Sodexis, Odoo Community Association (OCA)",
```
bealdav: Could you add 'Camptocamp' ?
A lot of code comes from them

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+    _description = "Exception Rules"
+    _order = 'active desc, sequence asc'
+
+    name = fields.Char('Exception Name', required=True, translate=True)
+    description = fields.Text('Description', translate=True)
+    sequence = fields.Integer(
+        string='Sequence',
+        help="Gives the sequence order when applying the test")
+    rule_group = fields.Selection(
+        [],
```
bealdav: could you add explicit parameter, there and below ?

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+        return all_exception_ids
+
+    @api.model
+    def _exception_rule_eval_context(self, obj_name, rec):
+        user = self.env['res.users'].browse(self._uid)
+        return {obj_name: rec,
+                'self': self.pool.get(rec._name),
+                'object': rec,
+                'obj': rec,
+                'pool': self.pool,
```
bealdav: pool or env ?

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+            all_exception_ids += exception_ids
+        return all_exception_ids
+
+    @api.model
+    def _exception_rule_eval_context(self, obj_name, rec):
+        user = self.env['res.users'].browse(self._uid)
+        return {obj_name: rec,
+                'self': self.pool.get(rec._name),
+                'object': rec,
+                'obj': rec,
```
bealdav, line 165: Is these 2 args are required whereas they have the same value ?

base_exception/models/base_exception.py
```diff
@@ -0,0 +1,215 @@
...
+                model_exceptions, sub_exceptions)
+            obj.exception_ids = [(6, 0, exception_ids)]
+            all_exception_ids += exception_ids
+        return all_exception_ids
+
+    @api.model
+    def _exception_rule_eval_context(self, obj_name, rec):
+        user = self.env['res.users'].browse(self._uid)
+        return {obj_name: rec,
+                'self': self.pool.get(rec._name),
```
bealdav, line 163: pool vs env ?

base_exception/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+ 'website': 'http://www.akretion.com',
+ 'depends': ['sale'],
```
bealdav: s/sale/base

base_exception/models/base_exception.py
mourad-ehm: These args are used to evaluate python code of exception. in this code we can use obj or object .

base_exception/models/base_exception.py
mourad-ehm: env and pool do not give the same result :
(Pdb) self.env[rec._name]         
sale.order()
(Pdb) self.pool.get(rec._name)
<class 'odoo.api.sale.order'>
I'm afraid that use of env can generate a side effect.

bealdav:
:+1:

PR #779, opened 2017-03-15
[MIG 10] keychain

EBII:


EBII:
It lacks the innitial commit

PR #755, opened 2017-02-28
[8.0][FIX] Make letsencrypt resilient for alternate name removal.

NL66278:

PR #783, opened 2017-03-20
[8.0] resilient letsencrypt extra

NL66278:
This includes the 8.0-resilient-letsencrypt PR: https://github.com/OCA/server-tools/pull/755

This PR has some improvements in error-handling and will make sure the main domain remains available, even if renewing of extra hostnames (altnames) fails.

### DIFF
--- a/letsencrypt/models/letsencrypt.py
+++ b/letsencrypt/models/letsencrypt.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2016 Therp BV <http://therp.nl>
+# © 2016-2017 Therp BV <http://therp.nl>
 # © 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import os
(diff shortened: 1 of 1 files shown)


letsencrypt/models/letsencrypt.py
```diff
@@ -132,23 +135,45 @@ def generate_csr(self, domain):
...
+            except Exception as ex:
+                # If certificate renewal fails for altname, we still will
+                # keep main domain with valid certificate:
+                _logger.warn(
+                    'Exception occured during certificate retrieval.'
+                    ' Retrying without altnames. Exception was: %s' %
+                    str(ex)
+                )
+                save_error = ex
+                csr = self.generate_csr(domain, ignore_altnames=True)
```
hbrunn, line 177: as you do this unconditionally, can't this create an infinite loop if there's a problem with the domain from `web.base.url`?

hbrunn (changes requested):

letsencrypt/models/letsencrypt.py
NL66278: No, because an exception raised in the exception handler, will not be covered by the previous try: So if there is a problem with web.base.url, the function will be ended by an exception immediately.

hbrunn: ah, this is within the cron function, thanks

hbrunn (approved):

pedrobaeza:
Please rebase and fix Travis.

NL66278:
Travis error occurs because web.base.url does not contain a valid non-local domain name. For the moment I do not see a way to fix this. The tests work locally in a database with a valid url.

yajo:
You'll have to fix the tests. Currently I see 2 options:

1. Use `mock.patch` to trick the test in some point.
2. Change `web.base.url` in the `setUp` method for the test(s).

I'm afraid we can't merge with ❌ in Travis.

hbrunn:
closing for inactivity

PR #784, opened 2017-03-20
v10 base_onchange_rule

bealdav:
Allow to store fields settings (default values, allowed values, etc)
to drive a specific behavior on any model/record

### Fictive use cases

**In sale order**

deliv1
**With related field definition**
When partner_shipping_id is 'Camptocamp' => Camptocamp pricelist is used whatever partner_id

-------------

deliv2
**With fix value**
When partner_shiping_id is 'Camptocamp' => 'eeee' pricelist is used whatever partner_id

------

**Demo Data**

onchange

onch_demo


base_onchange_rule/models/onchange_rule.py
```diff
@@ -0,0 +1,118 @@
...
+                lines[line.implied_record.id][line.field_id.name] = {
+                    'value': line.m2o_value.id or line.selection_value,
+                    'readonly': line.readonly,
+                    'domain': line.domain,
+                }
+                if (lines[line.implied_record.id][line.field_id.name]
+                        ['readonly']):
+                    readonly.append(line.field_id.name)
+                configs[rule.field_id.name] = lines
+        # print '    >>>> CONFIGS', configs, readonly
```
mourad-ehm: test code to delete (print ...)

base_onchange_rule/models/onchange_rule.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @api.model
+    def _update_nodes_from_rule(self, doc, fields_def, field, tag, attrs):
+        for path in ("//field[@name='%s']", "//label[@for='%s']"):
+            node = doc.xpath(path % field)
+            if node:
+                for current_node in node:
+                    current_node.set(tag, str(attrs))
+                    orm.setup_modifiers(current_node, fields_def[field])
+                    print "         NODE  ", current_node.values()
```
mourad-ehm: same remark (print)

base_onchange_rule/models/onchange_rule.py
```diff
@@ -0,0 +1,120 @@
...
+            rules = self._get_rules_from_model(src_model._name)
+            if rules:
+                doc = etree.XML(res['arch'])
+                params, readonly = (self.env['onchange.rule']
+                                    ._get_config_from_model(
+                                        src_model._name, rules))
+                fields_def = src_model.fields_get(
+                    allfields=None, attributes=None)
+                _logger.debug("  >>>> params, reado %s %s ", params, readonly)
+                for field in params.keys():
```
sebastienbeau: After reading mote the odoo API, maybe a cleaner solution for setting the onchange to True will be to inherit the method "_has_onchange" from models.py https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L5315

base_onchange_rule/models/onchange_rule.py
```diff
@@ -0,0 +1,120 @@
...
+                    for param in params[field]:
+                        self._update_nodes_from_rule(
+                            doc, fields_def, field, 'on_change', 1)
+                for field in readonly:
```
sebastienbeau: Be careful we should not set readonly True but something like "attrs: {'readonly': [('my_trigger_onchange_field', 'in', [3, 4..])]"
Also instead of reparsing the xml maybe we can plug the code for adding/updating the "attrs" here : https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_ui_view.py#L700

sebastienbeau (changes requested):

base_onchange_rule/models/onchange_rule.py
```diff
@@ -0,0 +1,120 @@
...
+                    field_value = values[field_name]
+                    param_field = configs[field_name][field_value]
+                    for field in param_field:
+                        res['value'][field] = param_field[field]['value']
+        return res
+
+    @api.model
+    def _get_rules_from_model(self, model_name):
+        model = self.env['ir.model'].search([('model', '=', model_name)])
+        return self.env['onchange.rule'].search(
```
mourad-ehm, line 213: I think, you can make one search call like this : 
return self.env['onchange.rule'].search(
            [('model_id.model', '=', model_name)])

base_onchange_rule/views/onchange_rule_view.xml
```diff
@@ -0,0 +1,109 @@
...
+                                        context="{'dest_field_id': onchange_field_id}"
+                                        attrs="{
+                                        'invisible': [('dest_val_type', '!=', 'fixed')]}"/>
+                                    <field name="dest_selection_value" 
+                                        attrs="{
+                                        'invisible': [('dest_val_type', '!=', 'fixed')]}"/>
+                                </group>
+                                <group colspan="4">
+                                    <field name="is_restrictive_rule"/>
+                                    <field name="onchange_domain" placeholer="{'partner_id': [('name', '=', 'Akretion')]}"/>
```
bealdav: s/placeholer/placeholder

sebastienbeau (approved):

gurneyalex:
@bealdav is this still "WIP" ?

gurneyalex (approved):

gurneyalex:
@bealdav could you rebase the PR just to make sure it is still green?

legalsylvain:
No need. Ocabot will test, rebasing, and if red, will not merge. Don't you think ?

PR #793, opened 2017-03-25
10.0 migrate auth_supplier

StephanRozendaal:
Migrate the auth_supplier module to 10.0

### DIFF
--- a/auth_supplier/__manifest__.py
+++ b/auth_supplier/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': "Auth Supplier",
     'category': 'Portal',
-    'version': '9.0.2.0.0',
+    'version': '10.0.1.0.0',
     'depends': [
         'auth_signup',
     ],
@@ -20,5 +20,5 @@
               'Odoo Community Association (OCA)',
     'website': 'http://www.incaser.es',
     'license': 'AGPL-3',
-    'installable': False,
+    'installable': True,
 }
--- a/auth_supplier/README.rst
+++ b/auth_supplier/README.rst
@@ -33,7 +33,7 @@ To use this module, you need to:
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/149/9.0
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
 
 Known issues / Roadmap
 ======================
--- a/auth_supplier/controllers/main.py
+++ b/auth_supplier/controllers/main.py
@@ -4,8 +4,8 @@
 # © 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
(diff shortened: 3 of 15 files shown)


auth_supplier/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     'name': "Auth Supplier",
     'category': 'Portal',
-    'version': '9.0.2.0.0',
+    'version': '10.0.1.0.0',
```
Fenkiou: The module version was 2.0.0, it shouldn't be downgraded.

pedrobaeza, line 10: No, it's correct as it is. The subversion should be reset on each new Odoo version.

Fenkiou: This is non sense. Odoo releasing a version each year, modules won't have same version despite sharing exact same code.

Also, referring to [this](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers), there is no indication on resetting the module version.

The two first digits are just here to tell that the module can be installed only on a specific odoo version.

For this module, the version should even be `10.0.2.1.0` if you consider adding new translations is a feature.

**EDIT**: Forget my last sentence as it's transbot that produced the translation files and I guess the files are presents on version 9.0 too. Even if I think that new languages should bump version number too. :)

pedrobaeza: Please see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#tasks-to-do-in-the-migration with the reference for the change of the version, and there's a rationale after this decision that is too large to explain here.

Fenkiou: Oh, missed that one, sorry. I would like to read the discussion on how this decision came up if you have a link.

Fenkiou: Just found the [issue](https://github.com/OCA/pylint-odoo/issues/59) talking about that, deleting my comment ;)

**EDIT**: Though deleting my comment would delete all comments but no :/
The first comment was:
> Module version was 2.0.0, it should not be downgraded

**EDIT2**: In fact, I'm not sure this issue is related to the decision taken

Tardo:
Looks good for me... and I need it for migrate 'website_portal_purchase' to v10.0

Tardo (approved):
Looks good for me... and I need it for migrate 'website_portal_purchase' to v10.0

yajo (approved):
Code OK

felixvillafranca (commented):
Now a field "Account Type" appears in the sign up form; but when sign up an error occurs:

screenshot 2017-05-30 at 17 47 13

Tardo:
@felixvillafranca For some reason not show "Password" and "Confirm Password"... In local works fine.

signup

pedrobaeza:
Check if there's any conflicting module

felixvillafranca:
Hi Tardo; 
I've been checking your module; and you're rigth, the module is Ok..
Password and confirm password were hidden due to "auth_signup_verify_email" was installed.
And becuase of the outgoing mail server wasn't not configured it failed.

So, excuse me and thanks for your work.

felixvillafranca (approved):
It works

PR #776, opened 2017-03-13
[10.0][MIG] auth_admin_passkey

Fenkiou:
This is a port of auth_admin_passkey from 8.0.

I changed the way the verification is done, but it works same as before.

PR #803, opened 2017-04-01
Fix menu in base_exception

eugen-don:
The placement of this menu item is causing this error. https://github.com/OCA/server-tools/pull/776#issuecomment-290450549

In adittion it disables all access to the general settings menu...

I suggest placing it unter technical settings, right at the bottom of it.


pedrobaeza:
No need to put a name attribute (it's caught from action name), and we avoid changes in the translations.

pedrobaeza (approved):
Squashing on merge.

elicoidal (approved):

PR #805, opened 2017-04-04
[FIX][mass_editing] unreliable view domain definition

jbchouinard:
The domain in the mass editing view relies on internal
details of the frontend implementation to get the list of
IDs from the many2many model_ids field.

This fix uses a more robust approach to get the list of IDs.

### DIFF
--- a/mass_editing/models/mass_object.py
+++ b/mass_editing/models/mass_object.py
@@ -36,6 +36,7 @@ class MassObject(models.Model):
 
     @api.onchange('model_id')
     def _onchange_model_id(self):
+        res = {}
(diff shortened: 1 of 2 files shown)


ghost:
:+1: makes sense

jbchouinard:
I tested it functionally on v10 Community and Enterprise, it works on my end.

RoelAdriaans (approved):
Tested, functionality works!
Tested in enterprise 10.

pedrobaeza:
This doesn't work for editing existing records, as the onchange_model_id is not executed there. That's why the domain is put on the view.

PR #813, opened 2017-04-18
[ADD] show the traceback, don't expect all exceptions to fill value

hbrunn:
the error messages are a lot more helpful for a programmer if they contain the traceback of the exception. I also added some HTML escaping and print the exception itself if its value property is Falsy


scheduler_error_mailer/ir_cron_email_tpl.xml
```diff
@@ -21,9 +21,13 @@
 <p>OpenERP tried to run the scheduler <em>${object.name or ''}</em> in the database <em>${ctx.get('dbname')}</em> but it failed. Here is the error message :</p>
```
elicoidal: Odoo?

elicoidal (approved):

dreispt (approved):

pedrobaeza (approved):

PR #816, opened 2017-04-20
[ADD] module module_uninstall_check

legalsylvain:
Add new module 'module_uninstall_check' that provides extra tools to know what is the impact of the uninstallation of a module.

On each module, you have now a extra button 'uninstallation impact'.

module_form

When you click on it, a wizard displays the list of models (SQL tables) and fields (SQL columns) that will be dropped).

sale_uninstallation

review welcome.


rvalyi:
@legalsylvain I'll review more deeply, but seems nice and very welcome. Also @sebastienbeau once made an ORM patch or monkey patch so that module uninstall would not drop the object tables. May be that would be the place in such a module to bring such a monkey patch (or override) to do that say based on some ENV var for instance to simply avoid dropping the tables/columns when this option is set? What do you think? @sebastienbeau could you find this patch again?

legalsylvain:
Hi @rvalyi ! Thanks for your review !

In a first time, the idea of this module is just to answer to the question _"Hey, If I click on 'uninstall module', what's going to happen, that will not be possible to revert ?"_. (and because, default behaviour of Odoo, when you cancel an uninstallation, is to keep modules in a 'to remove' state, that is a mess. [see odoo core](https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/module/module.py#L522))

In a second time (not implemented yet), I thought to :
- provide extra button on each line of the wizard _"preserve during uninstallation"_. This button could delete the xml id of the model (or of the field) So, if you uninstall the module, the process will not drop the according table (or column).
- provide an extra button _"change the ownership"_ for a fields, changing xml id. I mean : You developp a feature in module_a, with field_1. After, you refactor your module, creating another module module_b, with field_1, too. Finally, the module_a has to be removed, because obsolete. If you uninstall the module_a, it will drop the field_1, even if module_b declares the field_1. (tested on v9.0, I don't know if all series have the same behaviour). If before, we change the xml id of the field, from module_a, to module_b, it is possible to preserve the data. this is basically what openupgradelib does in [that function](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L407)

The pro of this approach could be to have the possibility to select what you want to preserve or delete. But, i'm very interested to see @sebastienbeau's patch.

module_uninstall_check/wizards/view_wizard_module_uninstall.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (C) 2017 - Today: GRAP (http://www.grap.coop)
+@author Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<openerp><data>
+
+    <record id="view_wizard_module_uninstall_form" model="ir.ui.view">
```
jcdrubay: The name of the files and xml_id don't seem to match conventions:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#naming-xml_id

I think it should be `wizard_module_uninstall.py` and `<record id="wizard_module_uninstall_view_form"`

module_uninstall_check/wizards/wizard_module_uninstall.py
```diff
@@ -0,0 +1,117 @@
...
+    def onchange_module_id(self):
+        model_data_obj = self.env['ir.model.data']
+        model_obj = self.env['ir.model']
+        field_obj = self.env['ir.model.fields']
+
+        for wizard in self:
+            wizard.model_line_ids = False
+            wizard.field_line_ids = False
+
+        for wizard in self.filtered(lambda x: x.module_id):
```
jcdrubay: Why not only `for wizard in self:`? Because field module_id is a required field so I guess self == self.filtered(lambda x: x.module_id).

module_uninstall_check/wizards/wizard_module_uninstall.py
```diff
@@ -0,0 +1,117 @@
...
+            model_ids = []
+            module_names = wizard.module_ids.mapped('name')\
+                + [wizard.module_id.name]
+            # Get Models
+            models_data = []
+            model_xml_ids = model_data_obj.search([
```
jcdrubay: I think the variable name should be `model_ids` as it contains database ids, not xml_ids.

module_uninstall_check/wizards/wizard_module_uninstall.py
```diff
@@ -0,0 +1,117 @@
...
+                    models_data.append((0, 0, {
+                        'type': 'model',
+                        'model_id': model.id,
+                    }))
+                    model_ids.append(model.id)
+            wizard.model_line_ids = models_data
+
+            # Get Fields
+            fields_data = []
+            field_xml_ids = model_data_obj.search([
```
jcdrubay: Rename `field_xml_ids` to `field_ids` because it does store database ids, not xml_ids.

module_uninstall_check/wizards/wizard_module_uninstall_line.py
```diff
@@ -0,0 +1,69 @@
...
+
+    TYPE_SELECTION = [
+        ('model', 'Model'),
+        ('field', 'Field'),
+    ]
+
+    wizard_id = fields.Many2one(
+        comodel_name='wizard.module.uninstall', required=True)
+
+    type = fields.Selection(
```
jcdrubay: Should we rename type to line_type to avoid conflicts with python builtin type?

module_uninstall_check/wizards/wizard_module_uninstall_line.py
```diff
@@ -0,0 +1,69 @@
...
+    field_ttype = fields.Selection(
+        string='Field Type', related='field_id.ttype', readonly=True)
+
+    field_model_name = fields.Char(
+        string='Field Model Name', related='field_id.model_id.model',
+        readonly=True)
+
+    @api.multi
+    @api.depends('model_id')
+    def _compute_model_row_qty(self):
```
jcdrubay: I think we could re-write in a more simple way like this:
```
for line in self.filtered(lambda x: x.line_type == 'model'):
    model_env = self.env[line.model_id.model]
    line.model_row_qty = model_env.search_count()  # Note: this will ignore inactive records. Not sure if it is meaningful to take them into account anyway.

```
I understand we might loose in term of performance but this should not be critical considering how often we will use this feature.

jcdrubay (changes requested):
Very nice and useful feature. Thanks :)

I am wondering if instead of adding a new button + adding a new object + new views, it wouldn't be better to  extend the native `base.module.upgrade` so all this information would be show when we use the existing button "Uninstall" and get this kind of heads up whether we remember to click on check or not. Maybe, this could be added to the roadmap and done at next module upgrade as it would involve quite a lot of refactoring.

bealdav:
:+1: thanks

max3903 (approved):

module_uninstall_check/wizards/view_wizard_module_uninstall.xml
legalsylvain: yes.

module_uninstall_check/wizards/wizard_module_uninstall.py
legalsylvain: you're right.

module_uninstall_check/wizards/wizard_module_uninstall.py
legalsylvain: as you can see, model_ids is used (as field_ids). I so renamed into all_model_ids, and all_field_ids.

module_uninstall_check/wizards/wizard_module_uninstall.py
legalsylvain: see above.

module_uninstall_check/wizards/wizard_module_uninstall_line.py
legalsylvain: yes.

legalsylvain:
Hi @jcdrubay. Thanks for your review. I changed my PR, according your remarks, except 2. 

> I am wondering if instead of adding a new button + adding a new object + new views, it wouldn't be better to extend the native base.module.upgrade so all this information would be show when we use the existing button "Uninstall" and get this kind of heads up whether we remember to click on check or not. 

Indeed, I first did what you said, but there is a native bad feature in Odoo, if you click on uninstall, and then cancel, related modules will stay in a 'to remove' state, that is error prone. I so prefer an extra form. Second reason, I was thinking to improve in a second step, this module, to have the possibility to select many modules. for exemple, in a tree view, selecting X modules, and displaying this wizard.

> I think we could re-write in a more simple way like this

No I'm not in favor to do that, because the result is not correct. We want to know the impact of the uninstallation of the module, not only the unarchived / active datas. furthermore, using SQL request allows to have extra information like db size, used in database. (I added table size, in the last commit)

@all I added database size feature. (see updated images in the header of the PR)

thanks for your review.

jcdrubay (approved):
LGTM :+1: 

Well done with the addition of all the database information!

legalsylvain:
Using this module in a "just openupgraded" version, i just found an error that occurs if the module is installed, but the code is not here. (key error on self.env). I set this PR in a 'WIP' state, and  I'll realize a little fix to manage this case in the next days.

regards.

mourad-ehm:
Thanks @legalsylvain for this useful feature.
code review :+1:

hparfr:
Very usefull to prevent the deletion of columns when uninstalling old modules. 
👍

PR #819, opened 2017-04-23
[FIX] mail_environment must not depend on server_environment_files

sbidoul:
Depending on server_environment is enough because server_environment_files is itself a dependency of server_environment.

Otherwise we'd need [this like in 9.0](https://github.com/OCA/server-tools/blob/14396ca2afbc7280caa552be5cb3925d420a0109/setup/mail_environment/setup.py#L5-L11).


lmignon (approved):

sbidoul:
@guewen @gurneyalex no objections here?

gurneyalex (approved):

PR #821, opened 2017-04-25
[9.0] Add oauth_provider_jwt module

Garamotte:
A small module to allow `oauth_provider` to deliver a JSON Web Token to the client, instead of a randomly generated token. By doing this, some user's information is included in the token, to avoid some extra queries to retrieve them using the token later.

As stated in the readme, JWT authentication (https://tools.ietf.org/html/rfc7523) is not (yet ?) supported.

### DIFF
--- a/oauth_provider_jwt/controllers/main.py
+++ b/oauth_provider_jwt/controllers/main.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 SYLEAM
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import werkzeug
+from openerp import http
+from openerp.addons import oauth_provider
+from openerp.addons.web.controllers.main import ensure_db
+
+
+class OAuth2ProviderController(
+        oauth_provider.controllers.main.OAuth2ProviderController):
+    @http.route(
+        '/oauth2/public_key', type='http', auth='none', methods=['GET'])
+    def public_key(self, client_id=None, *args, **kwargs):
+        """ Returns the public key of the requested client """
+        ensure_db()
+
+        client = http.request.env['oauth.provider.client'].sudo().search([
+            ('identifier', '=', client_id),
+        ])
+        return werkzeug.wrappers.BaseResponse(
+            client.jwt_public_key or '', status=200)
--- a/oauth_provider_jwt/models/oauth_provider_client.py
+++ b/oauth_provider_jwt/models/oauth_provider_client.py
@@ -1,0 +1,191 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 SYLEAM
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+from datetime import datetime, timedelta
+from openerp import models, api, fields, exceptions, _
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from oauthlib.oauth2.rfc6749.tokens import random_token_generator
+except ImportError:
+    _logger.debug('Cannot `import oauthlib`.')
+
+try:
+    import jwt
+except ImportError:
+    _logger.debug('Cannot `import jwt`.')
+
+try:
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives.asymmetric.ec import \
+        EllipticCurvePrivateKey
+    from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
+    from cryptography.hazmat.primitives.serialization import \
+        Encoding, PublicFormat, PrivateFormat, NoEncryption, \
+        load_pem_private_key
+    from cryptography.hazmat.primitives.asymmetric import rsa, ec
+except ImportError:
+    _logger.debug('Cannot `import cryptography`.')
+
+
+class OAuthProviderClient(models.Model):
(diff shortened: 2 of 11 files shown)


Garamotte:
@lasley Ready for reviews :)

oauth_provider_jwt/controllers/main.py
```diff
@@ -0,0 +1,26 @@
...
+
+import werkzeug
+from openerp import http
+from openerp.addons import oauth_provider
+from openerp.addons.web.controllers.main import ensure_db
+
+
+class OAuth2ProviderController(
+        oauth_provider.controllers.main.OAuth2ProviderController):
+    def __init__(self):
```
lasley: This can be removed

oauth_provider_jwt/models/oauth_provider_client.py
```diff
@@ -0,0 +1,187 @@
...
+        EllipticCurvePrivateKey
+    from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
+    from cryptography.hazmat.primitives.serialization import \
+        Encoding, PublicFormat, PrivateFormat, NoEncryption, \
+        load_pem_private_key
+    from cryptography.hazmat.primitives.asymmetric import rsa, ec
+except ImportError:
+    _logger.debug('Cannot `import cryptography`.')
+
+ALGORITHMS = {
```
lasley: This would be better in the class IMO.

Also these aren't really algorithms, but are instead `CRYPTOSYSTEMS`.

oauth_provider_jwt/models/oauth_provider_client.py
```diff
@@ -0,0 +1,187 @@
...
+                    raise exceptions.ValidationError(
+                        _('The private key doesn\'t fit the selected '
+                          'algorithm!'))
+
+    @api.multi
+    def generate_private_key(self):
+        """ Generate a private key for ECDSA and RSA algorithm clients """
+        for client in self:
+            algorithm_prefix = client.jwt_algorithm[:2]
+            if algorithm_prefix not in ALGORITHMS:
```
lasley: This basically repeats the conditions below, just slightly reversed. What about an `else: continue` below instead of this?

oauth_provider_jwt/models/oauth_provider_client.py
```diff
@@ -0,0 +1,187 @@
...
+            )
+
+    @api.multi
+    def _compute_jwt_public_key(self):
+        """ Compute the public key associated to the client's private key
+
+        This is only done for asymetric algorithms
+        """
+        for client in self:
+            if client.jwt_algorithm.startswith(('ES', 'RS', 'PS')) and \
```
lasley: IMO all the hard coded `ES`, `RS`, `PS` somewhat defeats the purpose of having the constant. What about:

```
if client.jwt_private_key and client.jwt_algorithm[:2] in ALGORITHMS:
```

Also note the boolean eval before the `in` eval - it's cheaper, so might as well knock any of those out first.

oauth_provider_jwt/models/oauth_provider_client.py
```diff
@@ -0,0 +1,187 @@
...
+                client.jwt_public_key = private_key.public_key().public_bytes(
+                    Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
+            else:
+                client.jwt_public_key = False
+
+    @api.model
+    def _generate_jwt_payload(self, request):
+        """ Generate a payload containing data from the client """
+        data = {
+            'exp': datetime.utcnow() + timedelta(seconds=request.expires_in),
```
lasley: Instead of generating utcnow three times, which is more costly & could technically be yielding diff dates, you should instead assign to a temp var for use

oauth_provider_jwt/models/oauth_provider_client.py
```diff
@@ -0,0 +1,187 @@
...
+            'aud': request.client.identifier,
+            'iat': datetime.utcnow(),
+            'user_id': request.client.generate_user_id(request.odoo_user),
+        }
+        if request.client.jwt_scope_id:
+            # Sudo as the token's user to execute the scope's filter with that
+            # user's rights
+            scope_data = request.client.jwt_scope_id.sudo(
+                user=request.odoo_user).get_data_for_model(
+                    'res.users', res_id=request.odoo_user.id)
```
lasley: This should be broken into distinct operations:

```
scope = request.client.jwt_scope_id.sudo(
    user=request.odoo_user,
)
scope_date = scope.get_data_for_model(
    'res.users', res_id=request.odoo_user.id,
)
```

It's one more line, but way more clear IMO

lasley (changes requested):

oauth_provider_jwt/models/oauth_provider_client.py
Garamotte: Thanks for the wording/vocabulary.
I renamed it and moved in the class.

oauth_provider_jwt/models/oauth_provider_client.py
Garamotte: I replaced this by an `else: raise`, to warn the user if he clicks on the button with the wrong algorithm.

lasley (approved):
Thanks @sylvain-garancher

Garamotte:
Reminder to myself : Add a note about the jwt_scope_id field in the readme.

gurneyalex (approved):

Issue #715, opened 2017-01-25
[RFC] Case Insensitive Logins


lasley:
[This module is actually pre-existing](https://github.com/LasLabs/odoo-base/tree/9.0/res_users_case_insensitive) & I completely forgot it existed until a customer needed it in their v10. I propose moving it to OCA for v10.

Basically the module just creates logins as all lowercase & adds a case insensitive search for them. It is meant to be installed on a new database, because it is possible that two logins could be conflicting in a pre-existing database due to case sensitivity.

It's been a while since we looked at it, so we'd probably give it some sort of overhaul to not make it so incredibly simplistic. An immediate feature that comes to mind is a hook to determine if there are logins that would conflict on a pre-existing database & convert all logins to lowercase if they don't.

Anyone have thoughts or a need for this?

dreispt:
I'm using LDAP auth with M$ Active Directory (tm), and usernames are case insensitive there.
Allowing the login username lookup to be case insensitive would be nice for the users.
Another nice improvement would be to trim whitespace around the username: I found that when typing them in mobile devices we can sometimes get an extra space at the end.

lasley:
> I'm using LDAP auth 

This is my use-case as well. We have also noticed a measurable amount of end user contact to our customers due to missing accounts that really just have a capital letter in the username.

> Another nice improvement would be to trim whitespace around the username:

Good call. So maybe we should generalize the name a bit, something like `user_login_usability`?

dreispt:
Sounds good.

dreispt:
I think `case_insensitive` is more meaningful.
I suggest `auth_user_insensitive`

lasley:
Closing to track in #826

PR #826, opened 2017-05-05
[10.0] [ADD]: auth_user_insensitive

tedsalmon:
This PR fixes https://github.com/OCA/server-tools/issues/715

This module will make all Odoo logins case insensitive by hooking into methods of `res.users`

cc @lasley


auth_user_insensitive/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
lasley: Switch up to LGPL please

auth_user_insensitive/README.rst
```diff
@@ -0,0 +1,56 @@
...
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Dave Lasley <dave@laslabs.com>
+* Ted Salmon <tsalmon@laslabs.com>
+
+
```
lasley: Too much whitespace

auth_user_insensitive/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "category": "Authentication",
+    "website": "https://laslabs.com/",
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    'installable': True,
+    "depends": [
+        "mail",
+    ],
+    "data": [],
```
lasley: Remove blank key

auth_user_insensitive/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+from odoo import api, fields, models
+
+
+class ResUsers(models.Model):
+
+    _inherit = 'res.users'
+
+    login = fields.Char(
+        'Login',
+        size=64,
```
lasley: Remove, [no size limit in core](https://github.com/odoo/odoo/blob/9e159ef2048574d179a9afb2226397e962aa5725/odoo/addons/base/res/res_users.py#L193)

lasley (changes requested):
Thanks @tedsalmon - comments inline

Please also make a pre init hook to make sure that there are no logins that will conflict once the module is installed (e.g. `dave` & `Dave` would conflict), and raise an exception if so. Make sure to [follow these directions](https://github.com/OCA/maintainer-tools/blob/master/template/module/hooks.py#L11) for avoiding broken db.

Please also add a post_init_hook to convert all logins to lowercase, otherwise there could be future conflicts without warning.

auth_user_insensitive/models/res_users.py
dreispt: Do we really need to redefine the `login` field?

auth_user_insensitive/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+        required=True,
+        help='Used to log into the system. Case insensitive.',
+    )
+
+    @api.model
+    def search(self, domain, *args, **kwargs):
+        """ Overload search to lowercase name domains. This can't be done in
+        a regular search method as the field is not computed
```
dreispt: Aren't ILIKE SQL searches already case insensitive?

auth_user_insensitive/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        for idx, _domain in enumerate(domain):
+            if _domain[0] == 'login':
+                lower = _domain[2].lower() if _domain[2] else False
+                domain[idx] = (_domain[0], _domain[1], lower)
+        return super(ResUsers, self).search(domain, *args, **kwargs)
+
+    @api.model
+    def create(self, vals, ):
```
dreispt: Remove the trailing comma

dreispt (commented):
I miss an extension to the authentication method, to "enforce" the lowercase at auth validation time:
Given a login name as an input, it first tried an exact match (with upper and lower case). If not found, it will then try to find a lowercase match.

auth_user_insensitive/models/res_users.py
lasley: It's just for the help attribute to say case insensitive. I think that it is worth the addition, but I don't guess it's a necessary thing to keep.

auth_user_insensitive/models/res_users.py
lasley: Yup, but `ILIKE` wouldn't cover all the operators. Also, it's slower than `=`

auth_user_insensitive/models/res_users.py
dreispt: AFAIK, for that you can use `login = fields.Char(help='...')`

auth_user_insensitive/models/res_users.py
dreispt: But isn't ILIKE the default operator used both on the search view and on the authentication method? If so, that's all we need.

lasley: Yup it's the default in the search view, but [The `_login` method (called by `authenticate`) looks to use `=`](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L456).

nhomar:
I honestly do not like the concept of this feature and it is too intrusive in the code for the value it adds in the system.

IMHO this should not be in OCA.

hbrunn:
@nhomar I think we should never dismiss modules for our own estimation of the value. usually people submit stuff because it fixes some problem, that's their value. let's look at the technical stuff, and indeed, it's tricky here

lasley:
Agreed with @hbrunn - I wouldn't have spent precious engineering hours developing this without a good reason:
* Active Directory is case insensitive. Case sensitive logins create conflicts, and are a security issue
* Old people, and otherwise tech-illiterates, generate a significant amount of support tickets for account merges because they capitalized their email during sign up, couldn't get reset emails, then created another account with different capitalization

hbrunn:
this module, as all modules which mess with login methods, might or might not interfere with ldap logins, or some other alternative auth mechanism. we can't use this as argument, because 1st there's currently no technical possibility to fix this properly (but oca modules at least could follow some dependency rule), and second we already have mutually exclusive auth-related modules, and that's the great thing: We only install what we want

hbrunn:
(I don't want to be the @pedrobaeza in the party here, but don't you think `user_insensitive` sends the wrong message? I think we can have the extra 5 bytes for `case_`...)

lasley:
I'm indifferent TBH, but we switched it due to https://github.com/OCA/server-tools/issues/715#issuecomment-279555759. @dreispt thoughts?

tedsalmon:
I think adding `case_` would make more sense but I went with what #715 called for.

Anyways, comments attended to!

auth_user_insensitive/models/res_users.py
dreispt: IMO it's more explicit to extend `_login` to use [`=ilike`](https://www.odoo.com/documentation/10.0/reference/orm.html#domains) instead of `=`. No ``search()` magic needed that way.
And this extension would run after and if the default `=` search fails.

dreispt:
@lasley its rude to call your users "insensitive" :wink:

lasley (approved):

auth_user_case_insensitive/models/res_users.py
```diff
@@ -0,0 +1,38 @@
...
+    @api.model
+    def search(self, domain, *args, **kwargs):
+        """ Overload search to lowercase name domains. This can't be done in
+        a regular search method as the field is not computed
+        """
+        for idx, _domain in enumerate(domain):
+            if _domain[0] == 'login':
+                lower = _domain[2].lower() if _domain[2] else False
+                domain[idx] = (_domain[0], _domain[1], lower)
+        return super(ResUsers, self).search(domain, *args, **kwargs)
```
dreispt: I dislike this. I still think that instead of hacking the `search()` method, we should extend the `_login`/`authenticate` method.

lasley: Technically we can just remove this and not add anything else. All new logins are being stored as lowercase, and we now have a hook to check for conflicting + convert existing logins

dreispt: True, but when we have case insensivity, an user expects to be able to login typing wither "eric", "Eric" or "ERIC".
To avoid hacking search(), we can just extend `_login` to lowercase the login before calling super(),

tedsalmon: Change has been made :). It was annoying to test, since `_login` uses a fresh cursor 😡, but it all worked out!

auth_user_case_insensitive/README.rst
```diff
@@ -0,0 +1,54 @@
...
+This module makes user logins case insensitive. It also overwrites the
+search method to allow these case insensitive logins to work on a database
+that previously had case sensitive logins.
+
+Installation
+============
+
+Install this module as you would any other. No further configuration is
+required.
+
```
dreispt, line 18: It is relevant to add that existing logins will all be converted to lowercase upon installation.

dreispt (approved):

auth_user_case_insensitive/README.rst
lasley: @tedsalmon please attend to this so we can merge

PR #830, opened 2017-05-06
[10.0] [ADD] user_immutable: Add Module

### LINKED ISSUE #810: [RFC] user_immutable: Immutable User Permission
This module will provide a new security group `Immutable`. Users with this permission will be protected from write or deletion from users that are not within this security group. 

This group would only be able to be assigned to users by users that are already in the group, and the builtin super user will be the only default with the permission.

The trick is that we cannot allow the built-in super user to circumvent this permission if they do not have permission explicitly defined. The goal is to allow for a customer controlled built-in super user, while still locking out the ability for them to delete my service users.

Does anyone know any existing code that I could base on for this, or maybe a strategy that's currently in use for similar?

tedsalmon:
User Immutable
==============

This module adds a group named `Immutable` which cannot be altered by users 
outside of that group. By default, the `Administrator` user is the only user
given access to this group on install. This module also adds protections 
against non-members granting/revoking membership to this group.

Installation
============

* Install module as normal

Usage
=====

Simply add any user to the `Immutable` group, provided that your login user 
is a member of that group already.

This fixes #810 

cc @lasley


tedsalmon:
Hmm, the build is failing due to a conflict with another module. I'll have to sort that out.

tedsalmon:
And it's fixed! Just had to sort out the demo data :)

user_immutable/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+Installation
+============
+
+* Install module as normal
+
+Configuration
+=============
+
+None
```
lasley: Remove blank sections

user_immutable/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Immutable Users",
+    "summary": "Add Immutable User Support",
+    "version": "10.0.1.0.0",
+    "category": "Authentication",
+    "website": "https://www.laslabs.com",
+    "author": "LasLabs Inc, Odoo Community Association (OCA)",
```
lasley: No Inc in the author

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that access to the `Immutable` group is
+        not given or removed by users without access """
+        immutable_id = -1
+        try:
+            immutable_id = self.env.ref(IMMUTABLE).id
+        except ValueError:
+            pass
```
lasley: ["Errors should never pass silently.](https://www.python.org/dev/peps/pep-0020/)

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,30 @@
...
+        if self.id == immutable_id and vals.get('users'):
+            if not self.env.user.has_group(IMMUTABLE):
+                raise AccessError(
```
lasley: `AccessError(_(`

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,30 @@
...
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that access to the `Immutable` group is
+        not given or removed by users without access """
+        immutable_id = -1
+        try:
+            immutable_id = self.env.ref(IMMUTABLE).id
+        except ValueError:
+            pass
+        if self.id == immutable_id and vals.get('users'):
```
lasley: This expects singleton

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, models
+from odoo.exceptions import AccessError
+
+IMMUTABLE = 'user_immutable.group_immutable'
```
lasley: Why defining this constant in two files?

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,43 @@
...
+class ResUsers(models.Model):
+
+    _inherit = 'res.users'
+
+    def _check_immutable(self):
+        """ Check to see if the user being edited is Immutable and if so,
+        make sure that the user performing the action has access """
+        if self.has_group(IMMUTABLE):
+            if not self.env.user.has_group(IMMUTABLE):
+                raise AccessError(
```
lasley: `AccessError(_(`

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,43 @@
...
+
+IMMUTABLE = 'user_immutable.group_immutable'
+
+
+class ResUsers(models.Model):
+
+    _inherit = 'res.users'
+
+    def _check_immutable(self):
+        """ Check to see if the user being edited is Immutable and if so,
```
lasley, line 16: [all doc blocks are out of spec](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that there are no alterations to users
+        whom are members of the `Immutable` group. """
+        self._check_immutable()
+        immutable = self.env.ref(IMMUTABLE)
+        has_group = self.env.user.has_group(IMMUTABLE)
+        if vals.get('in_group_%s' % immutable.id) and not has_group:
+                raise AccessError(
```
lasley: Translate

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,43 @@
...
+                    'access to it'
+                )
+        return super(ResUsers, self).write(vals)
+
+    @api.multi
+    def unlink(self):
+        """ Override unlink to verify that there are no deletions of users
+        whom are members of the `Immutable` group. """
+        self._check_immutable()
+        return super(ResUsers, self).unlink()
```
lasley: Untested

lasley (changes requested):
Thanks @tedsalmon

user_immutable/models/res_users.py
tedsalmon: I always screw these up. I even looked at that PEP earlier that day. (retardedsadpanda)

tedsalmon:
Alright, @lasley. Comments have been attended to :)

user_immutable/README.md
```diff
@@ -0,0 +1,68 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
```
lasley: This should be an RST

user_immutable/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Immutable Users",
+    "summary": "Add Immutable User Support",
+    "version": "10.0.1.0.0",
+    "category": "Authentication",
+    "website": "https://www.laslabs.com",
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
lasley: LGPL-3

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,36 @@
...
+class ResGroups(models.Model):
+
+    _inherit = 'res.groups'
+
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that access to the `Immutable` group is
+        not given or removed by users without access
+        """
+        immutable_id = -1
```
lasley:
```
immutable = self.env.ref(IMMUTABLE, raise_if_not_found=False)
```

Also before this do
```
if not vals.get('users') or self.env.user.has_group(IMMUTABLE):
    return super(ResGroups, self).write(vals)
```

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,36 @@
...
+        try:
+            immutable_id = self.env.ref(IMMUTABLE).id
+        except ValueError:
+            logger.info('ValueError raised due to `Immutable` group not yet '
+                        'having been created')
+        for rec in self:
```
lasley:
```
self.filtered(lambda r: r == immutable)
```

lasley (changes requested):

user_immutable/models/res_groups.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that access to the `Immutable` group is
+        not given or removed by users without access
+        """
+        if not vals.get('users') or self.env.user.has_group(IMMUTABLE):
+            return super(ResGroups, self).write(vals)
+        immutable = self.env.ref(IMMUTABLE, raise_if_not_found=False)
+        if len(self.filtered(lambda r: r == immutable)):
```
rco-odoo: The following is equivalent as more readable IMHO:
```python
    if immutable and immutable in self:
```

rco-odoo (commented):
Just a small hint...

user_immutable/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+                  'access to it')
+            )
+        return super(ResUsers, self).write(vals)
+
+    @api.multi
+    def unlink(self):
+        """ Override unlink to verify that there are no deletions of users
+        whom are members of the `Immutable` group
+        """
+        self._check_immutable()
```
Fenkiou: I'm not sure, but if you unlink multiple `res.users` at one time, the call to `_check_immutable()` that calls [has_group](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L542) will produce an error because `self` will be a recordset.

If I'm right, you should use a `for rec in self`loop.

tedsalmon:
@rco-odoo / @Fenkiou,

Comments attended to -- Thanks for your reviews!

gurneyalex (approved):

Fenkiou (approved):

lasley (approved):

PR #832, opened 2017-05-08
[Add] odoo server maintenance

newtratip:
I have new module is odoo_server_maintenance for lock user if that user is in maintenance mode so they will can not login but if user not in maintenance mode user can normal login


newtratip:
Sorry, wrong PR.

PR #833, opened 2017-05-08
[ADD] odoo server maintenance

newtratip:
I have new module is odoo_server_maintenance for lock user if that user is in maintenance mode but if that user is not in maintenance mode user will login normal.


odoo_server_maintenance/__openerp__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+{
+    'name': 'Server Lock for Maintenance Mode',
+    'version': '8.0.1.0.0',
+    'author': 'Ecosoft,Odoo Community Association (OCA)',
+    'category': 'Hidden',
+    'description': """
```
dreispt: please remote the description key, and place it in a README file, using the template https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

odoo_server_maintenance/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+    'data': [
+        'wizards/change_maintenance_mode_view.xml',
+        'views/res_users_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    'installable': True,
+    'auto_install': False,
+}
+
+# vim:expandtab:smartinden't:tabstop=4:softtabstop=4:shiftwidth=4:
```
dreispt: Remove

odoo_server_maintenance/views/res_users_view.xml
```diff
@@ -0,0 +1,36 @@
...
+            </field>
+        </record>
+
+        <record id="view_users_form" model="ir.ui.view">
+            <field name="name">res.users.form</field>
+            <field name="model">res.users</field>
+            <field name="inherit_id" ref="base.view_users_form"/>
+            <field name="arch" type="xml">
+                <field name="partner_id" position="after">
+                    <field name="maintenance_mode" attrs="{'readonly': [('id','=',1)]}"/>
```
dreispt: The flag is read only only for admin?

odoo_server_maintenance/wizards/change_maintenance_mode.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @api.multi
+    def select_maintenance_mode(self):
+        ResUsers = self.env['res.users']
+        context = self._context.copy()
+        active_ids = context.get('active_ids', [])
+        if len(active_ids) > 0:
+            for user in ResUsers.browse(active_ids):
+                if user.id == SUPERUSER_ID:
+                    raise ValidationError(_("%s is superuser, so can not "
```
dreispt: IMHO we should just ignore admin, instead of raising an error.

odoo_server_maintenance/wizards/change_maintenance_mode.py
```diff
@@ -0,0 +1,33 @@
...
+                                            "select maintenance mode.")
+                                          % (user.name))
+                user.write({'maintenance_mode': True, })
```
dreispt: Instead of having a for-loop with "n" writes, why not use just a single write?

odoo_server_maintenance/wizards/change_maintenance_mode.py
```diff
@@ -0,0 +1,33 @@
...
+        return {'type': 'ir.actions.act_window_close'}
+
+    @api.multi
+    def unselect_maintenance_mode(self):
```
dreispt: Not DRY: select and uneslect methods about about the same logic.
You could create a common "_change_maintenance_mode" with a boolean "value" parameter, used in the select and unselect methods.

odoo_server_maintenance/controllers/main.py
```diff
@@ -0,0 +1,26 @@
...
+                                               request.params['login'],
+                                               request.params['password'])
+            if uid is not False:
+                maintenance_mode = \
+                    request.env.user.browse(uid).maintenance_mode
+                if maintenance_mode:
+                    values['error'] = _("This user is in maintenance mode")
+                    if request.env.ref('web.login', False):
+                        return request.render('web.login', values)
+        return super(Home, self).web_login(redirect, **kw)
```
dreispt: AFAIK "routes" are lazy evaluated, so the redirect will not done until all inherited methods are run.
This means you could have the "super" call first, and then check for the maintenance mode. 
With this, no need to perform the authentication twice.

dreispt (changes requested):

pedrobaeza:
Isn't odoo word in the module name repetitive?

kittiu:
@newtratip 
- I recommend change module name to **user_maintenance_mode**
- Please add README file, and remove description.
- LINT_CHECK, although Green, still have many warning to fix
- Please make the test Green

odoo_server_maintenance/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+
+from . import models
+from . import wizards
+from . import controllers
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
kittiu: Remove # vim: line in all .py fileds (last line)

dreispt:
Hello @newtratip , need help?

newtratip:
Thank you for every comment. I will try edit it.

odoo_server_maintenance/controllers/main.py
newtratip: I ask for advice about authentication once.

odoo_server_maintenance/views/res_users_view.xml
newtratip: yes only admin will can't set maintenance mode

odoo_server_maintenance/wizards/change_maintenance_mode.py
newtratip: Thank you for recommend

dreispt:
Please post here when it's ready for new reviews.

newtratip:
OK dreispt :)

newtratip:
dreispt please check again. :)

kenvac:
bump!!
Any chance of getting it merge in 8.0 please?

pedrobaeza:
Yes if there are reviews...

user_maintenance_mode/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+User Maintenance Mode
+=====================
+
+Some companies need to use odoo server lock for some user to maintenance mode.
```
dreispt, line 9: For some or for all outside the "Maintenance" group?

user_maintenance_mode/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+
+{
+    'name': 'User Maintenance Mode',
+    'version': '8.0.1.0.0',
+    'author': 'Tharathip C.,Ecosoft,Odoo Community Association (OCA)',
+    'category': 'Hidden',
+    'description': """
```
dreispt: remove the description key

dreispt (commented):
Is the "maintenance mode" enabled by user? I have 300 users - that means that I have to enable it in all of them? 

What do you think on having the "maintenance mode" at the Group level rather than User? (For safety, the admin would be excluded from the maintenance mode lock out).

PR #836, opened 2017-05-11
[ADD] user_threshold: Create module

### LINKED ISSUE #811: [RFC] user_threshold: Impose Superficial User Limit
This module will provide the ability to limit the amount of non-portal/public user limits at the following levels:

* Total users per database - System parameter
* Total users per company - Company option

It will provide a security group for the ability to administer the amount of these users, and will restrict view/edit of the parameter & option to that group.

It will look for the environment variable `USER_THRESHOLD_HIDE` to be `1`. If it is, the security option should be removed from all users on all databases (during Odoo start). The security option should also be removed from the view so that administrators cannot add it back.

Also if this trigger is set, we will need to explicitly disable the ability for the sudo user to control these options. We will need to add explicit checks in the relevant models to circumvent this.

This will allow for a lockout of these permissions, and therefore a (possibly remote) configurable user threshold that cannot be circumvented without the proper user - in my case a service account.

We will also create a glue module `user_threshold_immutable` that will auto install if [user_immutable](https://github.com/OCA/server-tools/issues/810) is installed. That module will simply not count the immutable users towards the threshold.

Anyone know of anything existing, or have thoughts on implementation?

tedsalmon:
User Threshold
==============

This module adds the ability to limit the amount of non-portal/public that 
exist in the database and per-company. It also adds a group named 
`User Threshold Managers` which are  the only users who can alter the 
thresholds. This module also limits the  ability of users to add membership 
to the manager group to  pre-existing members. By default, `Administrator` 
is the only member of this group. There  is also a flag that can be set on 
users so that they do not count towards the user threshold.

Using the `USER_THRESHOLD_HIDE` environment variable, you can also hide the 
threshold exemption flag from users and the company setting for user 
threshold. Setting this flag will also remove threshold exemptions for any 
users who are not defined in the `USER_THRESHOLD_USER` environment variable.

Installation
============

* Install module as normal

Usage
=====

A system parameter named `user.threshold.database` is added by default with 
the value of '0' (Unlimited). Set this value to the total number of users 
you wish to allow in the database.

A field has been added to users to allow you to exempt them from the 
thresholds.

A field has been added to all companies, which allows you to define the max 
number of users that the company can have.

The following environment variables are available for your configuration
pleasure:

| Name | Description |
|------|-------------|
| USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those definied by the USER_THRESHOLD_USERS variable
| USER_THRESHOLD_USER | White list of users who are exempt from the 
threshold.

Fixes #811
cc @lasley

### DIFF
--- a/user_threshold/README.rst
+++ b/user_threshold/README.rst
@@ -3,27 +3,27 @@
    :alt: License: LGPL-3
 
 ==============
-User Immutable
+User Threshold
 ==============
 
-This module adds the ability to limit the amount of non-portal/public that 
-exist in the database and per-company. It also adds a group named 
-`User Threshold Managers` which are  the only users who can alter the 
-thresholds. This module also limits the  ability of users to add membership 
+This module adds the ability to limit the amount of non-portal/public
+users that exist in the database and per-company.
+
+It adds a group named `User Threshold Managers` which are  the only users
+who can alter the thresholds.
+
+This module also limits the  ability of users to add membership
 to the manager group to  pre-existing members. By default, `Administrator` 
-is the only member of this group. There  is also a flag that can be set on 
-users so that they do not count towards the user threshold.
+is the only member of this group.
+
+Additionally, there is a flag that can be set on users so that they do not
+count towards the user threshold.
 
 Using the `USER_THRESHOLD_HIDE` environment variable, you can also hide the 
 threshold exemption flag from users and the company setting for user 
 threshold. Setting this flag will also remove threshold exemptions for any 
 users who are not defined in the `USER_THRESHOLD_USER` environment variable.
 
-
-Installation
-============
-
-* Install module as normal
 
 Usage
 =====
@@ -44,8 +44,7 @@ pleasure:
 | Name | Description |
 |------|-------------|
 | USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those definied by the USER_THRESHOLD_USERS variable
-| USER_THRESHOLD_USER | White list of users who are exempt from the 
-threshold.
+| USER_THRESHOLD_USER | White list of users who are exempt from the threshold.
 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
--- a/user_threshold/README.rst
+++ b/user_threshold/README.rst
@@ -1,0 +1,91 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==============
+User Immutable
+==============
+
+This module adds the ability to limit the amount of non-portal/public that 
+exist in the database and per-company. It also adds a group named 
+`User Threshold Managers` which are  the only users who can alter the 
+thresholds. This module also limits the  ability of users to add membership 
+to the manager group to  pre-existing members. By default, `Administrator` 
+is the only member of this group. There  is also a flag that can be set on 
+users so that they do not count towards the user threshold.
+
+Using the `USER_THRESHOLD_HIDE` environment variable, you can also hide the 
+threshold exemption flag from users and the company setting for user 
+threshold. Setting this flag will also remove threshold exemptions for any 
+users who are not defined in the `USER_THRESHOLD_USER` environment variable.
+
+
+Installation
+============
+
+* Install module as normal
+
+Usage
+=====
+
+A system parameter named `user.threshold.database` is added by default with 
+the value of '0' (Unlimited). Set this value to the total number of users 
+you wish to allow in the database.
+
+A field has been added to users to allow you to exempt them from the 
+thresholds.
+
+A field has been added to all companies, which allows you to define the max 
+number of users that the company can have.
+
+The following environment variables are available for your configuration
+pleasure:
+
+| Name | Description |
+|------|-------------|
+| USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those definied by the USER_THRESHOLD_USERS variable
+| USER_THRESHOLD_USER | White list of users who are exempt from the 
+threshold.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+`<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Ted Salmon <tsalmon@laslabs.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/user_threshold/README.rst
+++ b/user_threshold/README.rst
@@ -24,6 +24,12 @@ threshold exemption flag from users and the company setting for user
 threshold. Setting this flag will also remove threshold exemptions for any 
 users who are not defined in the `USER_THRESHOLD_USER` environment variable.
 
+There are two modules available that also implement functionality similar to
+what is provided in this module but in a more abstract way. They are:
+
+https://github.com/it-projects-llc/access-addons/tree/10.0/access_limit_records_number
+https://github.com/it-projects-llc/access-addons/tree/10.0/access_restricted
+
 
 Usage
 =====
@@ -38,12 +44,11 @@ thresholds.
 A field has been added to all companies, which allows you to define the max 
 number of users that the company can have.
 
-The following environment variables are available for your configuration
-pleasure:
+The following environment variables are available for your configuration ease:
 
 | Name | Description |
 |------|-------------|
-| USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those definied by the USER_THRESHOLD_USERS variable
+| USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those defined by the USER_THRESHOLD_USERS variable
 | USER_THRESHOLD_USER | White list of users who are exempt from the threshold.
 
 
--- a/user_threshold/data/user_threshold_data.xml
+++ b/user_threshold/data/user_threshold_data.xml
@@ -2,19 +2,12 @@
 <!-- Copyright 2017 LasLabs Inc.
      License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
 
-<odoo noupdate="1">
+<odoo>
 
     <record id="group_threshold_manager" model="res.groups">
         <field name="name">User Threshold Manager</field>
         <field name="category_id" ref="base.module_category_hidden"/>
         <field name="users" eval="[(4, ref('base.user_root'))]" />
-    </record>
-
-    <record id="config_user_threshold" model="ir.config_parameter">
-        <field name="key">user.threshold.database</field>
-        <field name="value">0</field>
-        <field name="group_ids"
-               eval="[(4, ref('user_threshold.group_threshold_manager'))]" />
     </record>
 
 </odoo>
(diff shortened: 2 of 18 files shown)


lasley:
@tedsalmon your readme is for user immutable. Please update and fix your PR description

_Edit - It's just the title it seems. I fixed the PR description_

user_threshold/README.rst
```diff
@@ -0,0 +1,91 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==============
+User Immutable
+==============
+
+This module adds the ability to limit the amount of non-portal/public that
```
lasley: users that

user_threshold/README.rst
```diff
@@ -0,0 +1,91 @@
...
+   :alt: License: LGPL-3
+
+==============
+User Immutable
+==============
+
+This module adds the ability to limit the amount of non-portal/public that 
+exist in the database and per-company. It also adds a group named 
+`User Threshold Managers` which are  the only users who can alter the 
+thresholds. This module also limits the  ability of users to add membership
```
lasley: Should break this up. Lots of "This also" included for one paragraph

user_threshold/README.rst
```diff
@@ -0,0 +1,91 @@
...
+is the only member of this group. There  is also a flag that can be set on 
+users so that they do not count towards the user threshold.
+
+Using the `USER_THRESHOLD_HIDE` environment variable, you can also hide the 
+threshold exemption flag from users and the company setting for user 
+threshold. Setting this flag will also remove threshold exemptions for any 
+users who are not defined in the `USER_THRESHOLD_USER` environment variable.
+
+
+Installation
```
lasley: Remove empty section

user_threshold/data/user_threshold_data.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
+
+<odoo noupdate="1">
```
lasley: Security rules should be noupdate=0

user_threshold/data/user_threshold_data.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <record id="group_threshold_manager" model="res.groups">
+        <field name="name">User Threshold Manager</field>
+        <field name="category_id" ref="base.module_category_hidden"/>
+        <field name="users" eval="[(4, ref('base.user_root'))]" />
+    </record>
+
+    <record id="config_user_threshold" model="ir.config_parameter">
```
lasley: Break data up by model

user_threshold/models/ir_config_parameter.py
```diff
@@ -0,0 +1,48 @@
...
+from odoo.exceptions import AccessError
+
+MAX_DB_USER_PARAM = 'user.threshold.database'
+HIDE_THRESHOLD = str(os.environ.get('USER_THRESHOLD_HIDE', '')) == '1'
+
+
+class IrConfigParameter(models.Model):
+    _inherit = 'ir.config_parameter'
+
+    def _can_manipulate_th(self):
```
lasley: `_th`?

user_threshold/models/res_groups.py
```diff
@@ -0,0 +1,26 @@
...
+    _inherit = 'res.groups'
+
+    @api.multi
+    def write(self, vals):
+        """ Override write to verify that membership of the Threshold Manager
+        group is not able to be set by users outside that group
+        """
+        manager = self.env.ref(THRESHOLD_MANAGER, raise_if_not_found=False)
+        is_manager = self.env.user.has_group(THRESHOLD_MANAGER)
+        if len(self.filtered(lambda r: r == manager)) and not is_manager:
```
lasley: `if not is_manager and manager in self`

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    def __init__(self, pool, cr):
+        """ Override to check if env var to hide threshold configuration and
+        reset the database state is set. If it is, run those actions
+        """
+        if HIDE_THRESHOLD:
+            exempt_users = os.environ.get(
+                'USER_THRESHOLD_USER', ''
+            ).split(',')
```
lasley: Even if you think it's simple, never parse CSV by hand. Always use the csv module.

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+            cr.execute(
```
lasley: Did ORM significantly slow this down or something?

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+    def _check_thresholds(self):
+        """ Check to see if any user thresholds are met
+        Returns:
+            False when the thresholds aren't met and True when they are
+        """
+        domain = [
+            ('threshold_exempt', '=', False),
+            ('share', '=', False),
+        ]
+        db_users = len(self.env['res.users'].search(domain))
```
lasley: [`search_count`](https://github.com/odoo/odoo/blob/faba6daf7c8f95bfc865cc3f59297070c59e3e3b/odoo/models.py#L1489)

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+        max_db_users = int(self.env['ir.config_parameter'].get_param(
+            MAX_DB_USER_PARAM
+        ))
+        if max_db_users > 0 and db_users >= max_db_users:
+            return True
+        company = self.env.user.company_id
+        domain.append(('company_id', '=', company.id))
+        company_users = len(self.env['res.users'].search(domain))
```
lasley: Why bother with another search? 

```
company_user = db_users.filtered(lambda r: r.company_id == company)
```

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+        if company.max_users > 0 and company_users >= company.max_users:
+            return True
+        return False
+
+    @api.multi
+    def copy(self, default=None):
+        """ Override method to make sure the Thresholds aren't met before
```
lasley: So I keep telling you that you're writing your doc blocks out of spec, and you have yet to write one correctly. Please read the [PEP](https://www.python.org/dev/peps/pep-0257/) again very very very closely, paying particular attention to what your first line should look like.

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+        creating a new user
+        """
+        if self._check_thresholds():
+            raise ValidationError(_(
+                'Cannot add user - Maximum number of allowed users reached!'
```
lasley: The exclamation point isn't very professional IMO. Let's switch it to a statement instead of yelling at the user.

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+    def write(self, vals):
+        """ Override write to verify that membership of the Threshold Manager
+        group is not able to be set by users outside that group
+        """
+        th_group = self.env.ref(THRESHOLD_MANAGER)
+        user_is_manager = self.env.user.has_group(THRESHOLD_MANAGER)
+        if vals.get('threshold_exempt') and not user_is_manager:
+            raise AccessError(_(
+                'You must be a member of the `User Threshold Manager`'
+                ' group to grant threshold exemptions'
```
lasley: exemptions.

user_threshold/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+            ))
+        if vals.get('in_group_%s' % th_group.id) and not user_is_manager:
+            raise AccessError(_(
+                'You must be a member of the `User Threshold Manager`'
+                ' group to grant access to it'
```
lasley: it.

lasley (changes requested):

user_threshold/models/res_users.py
tedsalmon: Sure! Is there's a specific reason why though? Seems like bringing in Jon Cena to deal with a schoolyard bully :P

lasley: While it would be incredibly weird, is it technically possible for a login to contain a `,`?

user_threshold/models/res_users.py
tedsalmon: No, I just had to create an environment, since `self.env` is not yet available in the `__init__` and the `pool` is weird. I went ahead and just generated a cursor and environment.

user_threshold/models/res_users.py
tedsalmon: I can't use `search_count` if I want to use `.filtered()` later on the same object as the aforementioned returns a long, not a RecordSet

lasley: But `len` returns an int. How is that any different?

```
db_users = len(self.env['res.users'].search(domain))
```

user_threshold/models/ir_config_parameter.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @api.multi
+    def unlink(self):
+        """
+        Override to disallow deletion of the user threshold parameter
+        when the user does not have the right access
+        """
+        for rec in self.filtered(lambda r: r.key == MAX_DB_USER_PARAM):
+            if not self._can_manipulate():
+                raise AccessError(_(
```
tedsalmon: This line and L47 are showing as untested because the user I'm using for testing is not a member of the required group and therefore cannot read `r.key` on L33 and L46 respectively, causing an AccessError before we ever hit this one. If anyone has any ideas on circumventing this, I'm all ears :)

tedsalmon:
@lasley,

Fixed!

user_threshold/models/ir_config_parameter.py
lasley: Is there ever an instance in which the user would still be able to read `.key` without being able to manipulate?

user_threshold/README.rst
```diff
@@ -0,0 +1,90 @@
...
+
+A field has been added to all companies, which allows you to define the max 
+number of users that the company can have.
+
+The following environment variables are available for your configuration
+pleasure:
+
+| Name | Description |
+|------|-------------|
+| USER_THRESHOLD_HIDE | Hide all threshold settings and default the exempt users to those definied by the USER_THRESHOLD_USERS variable
```
elicoidal: defined

user_threshold/README.rst
```diff
@@ -0,0 +1,90 @@
...
+you wish to allow in the database.
+
+A field has been added to users to allow you to exempt them from the 
+thresholds.
+
+A field has been added to all companies, which allows you to define the max 
+number of users that the company can have.
+
+The following environment variables are available for your configuration
+pleasure:
```
elicoidal: s/pleasure/ease

user_threshold/README.rst
```diff
@@ -0,0 +1,90 @@
...
+to the manager group to  pre-existing members. By default, `Administrator` 
+is the only member of this group.
+
+Additionally, there is a flag that can be set on users so that they do not
+count towards the user threshold.
+
+Using the `USER_THRESHOLD_HIDE` environment variable, you can also hide the 
+threshold exemption flag from users and the company setting for user 
+threshold. Setting this flag will also remove threshold exemptions for any 
+users who are not defined in the `USER_THRESHOLD_USER` environment variable.
```
elicoidal, line 25: WHITELIST_THRESHOLD_USER?

user_threshold/models/ir_config_parameter.py
```diff
@@ -0,0 +1,51 @@
...
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+import os
+
+from odoo import _, api, models
+from odoo.exceptions import AccessError
+
+MAX_DB_USER_PARAM = 'user.threshold.database'
+HIDE_THRESHOLD = str(os.environ.get('USER_THRESHOLD_HIDE', '')) == '1'
```
elicoidal: Can you be consistent in the naming?

user_threshold/models/res_groups.py
```diff
@@ -0,0 +1,27 @@
...
+    def write(self, vals):
+        """ Override write to verify that membership of the Threshold Manager
+        group is not able to be set by users outside that group
+        """
+        manager = self.env.ref(THRESHOLD_MANAGER, raise_if_not_found=False)
+        if manager:
+            is_manager = self.env.user.has_group(THRESHOLD_MANAGER)
+            if not is_manager and manager in self:
+                raise AccessError(_(
+                    'You must be a member of the `User Threshold Manager` '
```
elicoidal, line 24: You should use this kind of message everywhere: it gives a better information for end user

elicoidal (changes requested):
Interesting module
Mostly details on my side

yelizariev:
There are two more common modules that allow to do the same

https://github.com/it-projects-llc/access-addons/tree/10.0/access_limit_records_number

https://github.com/it-projects-llc/access-addons/tree/10.0/access_restricted

lasley (approved):

lasley:
@tedsalmon - can you add a note in the readme pointing to the modules @yelizariev linked in https://github.com/OCA/server-tools/pull/836#issuecomment-301292630 please? We've gone too far to switch the implementation now, but IMO his is better so we should move towards something more abstract for future versions.

tedsalmon:
Done, @lasley!

gurneyalex (approved):

PR #845, opened 2017-05-25
FIX Allow using non_ASCII imap folder without encoding error

diggy128:
This patch allows fetching from an IMAP folder that its name contains non-ASCII characters (e.g. Greek) without leading to an encoding error.
I've opted to include a custom class instead of a python package as dependency because IMAP v4r1 specifies a slight variant of UTF7 to encode folder names.


yajo:
Hi there @diggy128 and welcome!

This feature you are adding needs IMHO some tests, because it's a very specific behavior, and that's the best way you can assure yourself it works forever and makes sense to future developers (although provided docs are OK for that too, but tests are best).

Could you supply them along in this pull request please? Thanks! 😊 

BTW, make travis go ✔️ too please.

PR #849, opened 2017-05-31
[FIX][mass_editing] v10 Fixing error when removing value for fields being…

BT-rmartin:
… translatable, as their translations were not removed.

### DIFF
--- a/mass_editing/README.rst
+++ b/mass_editing/README.rst
@@ -92,6 +92,7 @@ Contributors
 * Oihane Crucelaegui <oihanecrucelaegi@gmail.com>
 * Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
+* Raul Martin <raul.martin@braintec-group.com>
 
 Maintainer
 ----------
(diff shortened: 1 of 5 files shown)


yajo (changes requested):
Big amount of code, fixing a weird bug... Please add tests 😊

BT-rmartin:
@Yajo Tests is ready here as well

yajo (approved):
Code OK, let's wait for bots.

pedrobaeza:
Travis is failing. Please check.

rafaelbn (approved):
Thanks @BT-rmartin 

Only tested functionally in runbot :+1:

rafaelbn:
@Yajo we should take care this is included in v11

yajo:
Rebase please, it could fix travis here too

rafaelbn:
All bots are green :+1:  great!

rafaelbn:
@BT-rmartin I founf more fixes in https://github.com/OCA/server-tools/pull/1209

BT-rmartin:
Thanks @pedrobaeza @rafaelbn @Yajo for your reviews and comments. 
I will do it for v11 as well, but I noticed that module mass_editing now it's not in server-tools but in server-ux right?
Why do you mention #1209? To include their changes here or to review it?

BT-rmartin:
Here you have PR to v11 https://github.com/OCA/server-ux/pull/12

PR #851, opened 2017-06-07
[10.0][ADD] Improving the mail tracking value

TDu:
Improve the tracking of changed values on database fields ("mail.tracking.value" functionality) by adding many2many and one2many support.
As well as a more user friendly view to consult changes recorded.

### DIFF
--- a/mail_improved_tracking_value/__init__.py
+++ b/mail_improved_tracking_value/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
--- a/mail_improved_tracking_value/__manifest__.py
+++ b/mail_improved_tracking_value/__manifest__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Improved tracking value change',
+    'version': '10.0.1.0.0',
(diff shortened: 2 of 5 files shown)


pedrobaeza:
Please put this on OCA/social. Note that there's already a mail_tracking module there.

guewen:
> Please put this on OCA/social. Note that there's already a mail_tracking module there.

@pedrobaeza I'm not sure OCA/social is a good fit for this addon. This addon is an improvement of the `mail.tracking.value` system by providing support for *2many fields and a usable UI (the default one is useless). It contains the name `mail` in it only because the model for tracking values contains it, but this is hardly related to any social features. I proposed server-tools to @TDu because it is a relatively low-level and very generic improvement over the default one (it's more close to auditlog than a social feature).

Note that `mail_tracking` is really related to mail messages.

pedrobaeza:
But isn't `mail.tracking.value` related to mail and thus to OCA/social (all mail related modules goes there - don't think only in social like social networks...)?

guewen:
@pedrobaeza well the model's implemented in the `mail` addon and indeed is partly related to notifications about changes on records, but it can also be used as a tracking mechanism for the changes (audit log). Social is advertised as "Addons concerning odoo's social ERP features and messaging in general", if I were looking for this new addon (which does not improve the messaging part of `mail.tracking.value` but the tracking part), this is not where I would look. But anyway maybe there is no perfect place so if you insist we'll move it to OCA/social. Just let us know your final word :)

pedrobaeza:
Thanks for the explanation, Guewen. Indeed, there's no perfect place and this tracking system can be used for several purposes (although at the end, they are for "communicating" something).

But in this case, as you are using `mail` prefix and this repo is very crowded, I prefer to move it to OCA/social. Don't you mind?

guewen:
Ok @pedrobaeza, no problem, we move it there. Thanks!

PR #858, opened 2017-06-14
[10.0] add new module to manage technical user on company

Cedric-Pigeon:
Sometimes, when you launch batch processes, you don't want to use the current user for several reasons
- the current user can switch company before the end of the batch process
- you can encounter concurrent access error if the user works in the meantime

This module add a technical user on the company. You can use for exemple in cron process or in queue job.

### DIFF
--- a/base_technical_user/README.rst
+++ b/base_technical_user/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================
+Base Technical User
+===================
+
+This module extends the functionality of company management.
+It allows you to bind a technical user on the company in order to use it in
+batch processes.
+
+The technical user must
+- be inactive to avoid login
+- be in the required groups depending of what you need to do
+
+Usage
+=====
+
+If you install the module, you will find a tab on the company form allowing
+to define the technical user.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Cédric Pigeon <cedric.pigeon@acsone.eu>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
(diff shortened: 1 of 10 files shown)


Cedric-Pigeon:
@guewen this module aims to replace https://github.com/OCA/connector-magento/pull/212/files#diff-131bcecd325b739cf356543a96bb2747R11 in the multi-company PR for the magento connector

Cedric-Pigeon:
@sebastienbeau This could be useful for ShopInvader in the export for Algolia for example

base_technical_user/README.rst
```diff
@@ -0,0 +1,49 @@
...
+===================
+Base Technical User
+===================
+
+This module extends the functionality of company management.
+It allows you to bind a technical user on the company in order to use it in
+batch processes.
+
+The technical user must
+- be inactive to avoid login
```
guewen, line 14: I highly advise not to do that, been there, done that.

* you set some groups on your technical user, let's say one of the group is 'connector manager', all works well
* someday, you open the groups menu, edit the 'connector manager' group and add a user in the many2many, save it
* your technical user is no longer part of the 'connector manager' group and you don't notice it until things start to fail weirdly 

What happened? the many2many for users on the group does not include inactive users, so you just removed all the inactive users from the group by adding a new user.

What we do now is that we don't set a password on the user instead, it can't be used to log in.

guewen (changes requested):

base_technical_user/README.rst
Cedric-Pigeon: @guewen Interesting remark. I wasn't aware of that. Thanks. The other goal of making it inactive is to avoid Odoo taking care of this kind of users in the Enterprise contract.

guewen: @Cedric-Pigeon the `login_date` of such users will always be empty, maybe Odoo could not include such users in the Enterprise contract?

base_technical_user/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    user_tech_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Technical User",
+        help="This user can be used by process for technical purpose",
+        domain="[('company_id', '=', id), ('active', '=', False)]")
```
guewen: If we don't want to take the risk to disable the user (regarding [this comment](https://github.com/OCA/server-tools/pull/858#discussion_r125591029)), could you remove the domain on active = false?

Cedric-Pigeon: @guewen I just did it ;-)

guewen (approved):
Thanks! I'm still a uncomfortable with the recommendation to disable the user, but that's not a requirement for the module to work. :+1:

gurneyalex (approved):

PR #859, opened 2017-06-15
[FIX] password_security: Fix password stored and token to reset password invalid

moylop260:
Steps to reproduce issue:
1. Create a user with a valid password
2. Use the `Forget password` button
3. Use the same previous password

Saw:
Error of history password. Good!
Try to fix it your password and now the token to reset password is invalid. Wrong!
Check the fields user.password and is filled. Wrong!

Expected:
Error of history password and allow fix it and don't store user.password field.

Video:
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=4PDz2djXuPU)

If you don't want configure a email smtp apply the following patch:
```diff
diff --git a/addons/mail/models/mail_template.py b/addons/mail/models/mail_template.py
index 76d8ae741..316d96a3f 100644
--- a/addons/mail/models/mail_template.py
+++ b/addons/mail/models/mail_template.py
@@ -560,6 +560,7 @@ class MailTemplate(models.Model):
                 updated with given values dict
            :returns: id of the mail.message that was created
         """
+        raise_exception=False
         self.ensure_one()
         Mail = self.env['mail.mail']
         Attachment = self.env['ir.attachment']  # TDE FIXME: should remove dfeault_type from context
```

The query executed from the video is:
`psql openerp_test -c "SELECT login FROM res_users WHERE password != ''"`
`psql openerp_test -c "UPDATE res_users SET password=Null"`

MERGER
=======
Please merge just one commit using merge & squash of github

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -81,7 +81,9 @@ class ResUsers(models.Model):
             password_regex.append(r'(?=.*?\W)')
         password_regex.append('.{%d,}$' % company_id.password_length)
         if not re.search(''.join(password_regex), password):
-            raise PassError(_(self.password_match_message()))
+            self.env.cr.rollback()
+            self.invalidate_cache()
+            raise PassError(self.password_match_message())
         return True
 
     @api.multi
@@ -117,6 +119,8 @@ class ResUsers(models.Model):
             )
             delta = timedelta(hours=pass_min)
             if write_date + delta > datetime.now():
+                self.env.cr.rollback()
+                self.invalidate_cache()
                 raise PassError(
                     _('Passwords can only be reset every %d hour(s). '
                       'Please contact an administrator for assistance.') %
@@ -138,9 +142,10 @@ class ResUsers(models.Model):
                 recent_passes = rec_id.password_history_ids[
                     0:recent_passes-1
                 ]
-            if len(recent_passes.filtered(
-                lambda r: crypt.verify(password, r.password_crypt)
-            )):
+            if recent_passes.filtered(
+                    lambda r: crypt.verify(password, r.password_crypt)):
+                self.env.cr.rollback()
+                self.invalidate_cache()
                 raise PassError(
                     _('Cannot use the most recent %d passwords') %
                     rec_id.company_id.password_history
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -38,7 +38,7 @@ class ResUsers(models.Model):
     @api.multi
     def write(self, vals):
         if vals.get('password'):
-            self.check_password(vals['password'])
+            self._check_password(vals['password'])
             vals['password_write_date'] = fields.Datetime.now()
         return super(ResUsers, self).write(vals)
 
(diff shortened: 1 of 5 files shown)


lasley (approved):
Excellent catch @moylop260

password_security/models/res_users.py
```diff
@@ -81,6 +81,8 @@ def check_password(self, password):
             password_regex.append(r'(?=.*?\W)')
         password_regex.append('.{%d,}$' % company_id.password_length)
         if not re.search(''.join(password_regex), password):
+            self.env.cr.rollback()
+            self.invalidate_cache()
             raise PassError(_(self.password_match_message()))
```
yajo: Whoah, I didn't see this! Remove the call to `_()` here please.

password_security/models/res_users.py
```diff
@@ -81,6 +81,8 @@ def check_password(self, password):
             password_regex.append(r'(?=.*?\W)')
         password_regex.append('.{%d,}$' % company_id.password_length)
         if not re.search(''.join(password_regex), password):
+            self.env.cr.rollback()
```
yajo: I feel all is much more clean with `with self.env.cr.savepoint():`, which rolls back automatically if any exception is found inside the block (the same for all cases below).

yajo (approved):

password_security/models/res_users.py
moylop260: Sorry, but I didn't understand.
Where you have planned add a `savepoint`?


- Note: The module `auth_signup` has the `write`:
   
  - [auth_signup/models/res_users.py#L42](https://github.com/odoo/odoo/blob/676ba1bdb/addons/auth_signup/models/res_users.py#L42)

yajo: Oh yes sorry, forget about that.

Anyways, isn't transaction rolled back automatically when an exception is raised?

moylop260: Yeah, you are right!
I deleted the rollback and it is working just using invalidate_cache

For a strange reason the website frontend is not cleaning the cache.
Because the same exception using backend is working without clean cache.

So, maybe this is a issue related with website but I didn't find a way to reproduce it without a custom module and will be hardly that odoo fixed.

moylop260: Wait me I didn't test it with the real fail case... sorry.
I will come back with the news

moylop260: Yeah, I find the real issue.
The method `def _set_password(` don't run a rollback if an exception is raised.
So I just changed the inherited method to write validation
I removed rollback and clear caches

moylop260:
@lasley 
I reproduced this issue using frontend `/web/reset_password` and raising the password history exception.

I mean,

| exception raised          | UI                          | Issue detected |
|------------------|-----------------------------|----------------|
| password string rules   | User backend wizard         | No             |
| password string rules   | Frontend web/reset_password | No             |
| password history | User backend wizard         | No             |
| password history | Frontend web/reset_password | Yes            |

So I removed the rollbacks and invalidate_cache and moved the history check to rules string check method.
Thanks @Yajo for comments

Could you review it again, please?

egrivero (approved):

lasley (approved):
LGTM thanks @moylop260

PR #860, opened 2017-06-16
[8.0] Add module base_user_auth_log

alexis-via:
See README.rst for more info.

### DIFF
--- a/base_user_auth_log/models/res_users_auth_log.py
+++ b/base_user_auth_log/models/res_users_auth_log.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields, api, _
+from openerp.exceptions import Warning as UserError
+from datetime import datetime
+
+
+class ResUsersAuthLog(models.Model):
+    _name = 'res.users.auth.log'
+    _description = 'Users Authentication Logs'
+    _order = 'date desc'
+    _rec_name = 'date'
+
+    user_id = fields.Many2one(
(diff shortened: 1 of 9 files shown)


cdubuit (approved):
Test, no code review.
Great idea.

pedrobaeza:
https://github.com/OCA/server-tools/tree/8.0/auth_brute_force also logs logins, so I think this one is overlapping functionality.

legalsylvain (changes requested):
Hi @alexis-via, thanks for your contribution.

1. @pedrobaeza +1. 

@alexis-via, what did you not found in auth_brute_force to create this module. It could be better to improve auth_brute_force, instead of creating two ways to log authentication.
In a nutshell, auth_brute_force : 
- log authentications (as your module)
- ban users (by IP) that tried too many times to authenticate, with bad logins. (existing or not)

2. Tested on the last runbot instance, your module create 3 logs per try. Do you have an idea why ? 

image

kind regards.

legalsylvain:
3. Are you sure ```default=datetime.utcnow()``` is working correctly ? all the logs have the same authentication date, that is wrong.

base_user_auth_log/models/res_users_auth_log.py
```diff
@@ -17,8 +16,7 @@ class ResUsersAuthLog(models.Model):
         'res.users', string='User', ondelete='cascade', readonly=True)
     login = fields.Char(string='Login', readonly=True)
     date = fields.Datetime(
-        string='Authentication Date', required=True, readonly=True,
-        default=datetime.utcnow())
```
sbidoul: `default=datetime.utcnow`, without `()` should work.

dreispt:
Should this be closed in favour of `auth_brute_force`?

legalsylvain:
Closing. (Duplicate with auth brut force)
One idea of improvment could be to make a PR with one module to make logs, and another one auth brut force, depending of the first one, and adding brut force features.
(Need refactoring)

PR #863, opened 2017-06-20
Add module copy views

renatonlima:
Hi all,

follwing the discussion in the Brazilian location https://github.com/OCA/l10n-brazil/pull/544, I created this PR for discussion to add copy_views module in server-tools

@mileo and @aricaldeira

Can you introduce the features create by this module and add some tests in this module?

cc @rvalyi @mbcosta @pedrobaeza


AaronHForgeFlow:
What is the purpose of this module? The readme it's not clear

aricaldeira:
@aheficent 
It allows you to create the following structure:

module_A/views:
    * base_view.xml

module_B/views:
    * copied_from_base_view.xml (is a "clone" of the base_view; and it's arch field is kept in sync with base_view's arch, including all it's inheritance)
    * inherited_copied_view.xml (a view that inherits and changes the behaviour of the copied view, but does not affect the original base_view)

module_C/views:
    * inherited_base_view.xml (a view that inherits and changes the behaviour of the base view, and also affects the copied_from_base_view)

You can see it being used in here:
Base view: https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped/views/sped_documento_base_view.xml
Copied view, with minimal changes: https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped/views/sped_documento_emissao_nfe_view.xml
Copied view, with some more changes: https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped/views/sped_documento_emissao_nfce_view.xml
Copied view, with changes to item_ids views also: https://github.com/odoo-brazil/l10n-brazil-wip/blob/10.0-develop/sped/views/sped_documento_emissao_nfse_view.xml

The 3 copied views represent 3 types of fiscal documents in Brazil (of more than 30!), they all share the same functional logic, with some minor differences in required fields etc. Keeping all of them in the same table/model simplifies greatly dealing with changes in legislation, consolidate reports without 30 UNIONs, which is always, always nice, but having 30 or more views that, when needed, had to be all changed individually by hand was quite a pain... Hence, this module ;)

Sorry for the long and cumbersome explanation, but it was the best way I thought to give you not only the answer you wanted, but also the context and use case that sparkled the idea behind the module's behaviour.

hbrunn:
can't you do what you need to do with [`mode='primary'`](https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_ui_view.py#L169-L180)? Check how this is used in the views for `product.product` and `product.template`

Garamotte:
I agree with @hbrunn this corresponds exactly to the `primary="True"` behaviour.

You can create your base view, then inherit from multiple views using `primaty="True"` on each.
If a view inherits the base without `primary`, the changes are applied everywhere, if a view inherits the base with `primary`, the changes are applied only when you call this view explicitely (or one of its children).

renatonlima:
Hi guys,

@hbrunn thank you to remember this core feature.

@aricaldeira did this change to use primary attribute, I will close this PR.

PR #866, opened 2017-06-21
[add] base_fontawesome

simahawk:
started from OCA/website-cms/pull/13


base_fontawesome/README.rst
```diff
@@ -0,0 +1,44 @@
+.. image:: https://img.shields.io/badge/licence-lgpl--3-blue.svg
+   :target: http://www.gnu.org/licenses/LGPL-3.0-standalone.html
+   :alt: License: LGPL-3
+
+Base fontawesome
```
pedrobaeza: Line above

base_fontawesome/README.rst
```diff
@@ -0,0 +1,44 @@
...
+================
+
+Provide up to date `Fontawesome <http://fontawesome.io/>`_ resources.
+
+Current version: 4.7.0.
+
+
```
pedrobaeza: Add usage section indicating how to use this (or you just need to work the same way as standard? - just indicate it then).

pedrobaeza (changes requested):

yajo (approved):
Nothing besides @pedrobaeza's comments.

I'd just encourage you to open a PR to odoo master updating FA, if you care for this and want to stop maintaining this addon in the near future 😉

base_fontawesome/README.rst
simahawk: sorry, what do you mean? :)

simahawk:
oops, I totally forgot about this PR :stuck_out_tongue: 
I updated the README and I changed the version of the module to match FA one so that they stay in line.
I'm gonna check odoo master version and submit a PR if needed.

simahawk:
@Yajo they already updated it on master \o/

BTW tests are failing because of this issue https://github.com/OCA/server-tools/pull/1007

base_fontawesome/README.rst
simahawk: @pedrobaeza better now?

yajo:
Nice, please add a note to roadmap telling not to migrate to v11 😊

simahawk:
@Yajo not sure: **only as of today** v11 ships latest version but in the future is very likely that is going to be outdated so this module can still be useful to ship always the latest version of FA.

simahawk:
updated roadmap (no CI SKIP as I wanted the start a new build to include the fix for #1007)

simahawk:
Sounds like tests that were passing before are not passing now when running w/ `EXCLUDE="database_cleanup"`. Now it's `html_text`... @yajo any bell? https://travis-ci.org/OCA/server-tools/jobs/282692962#L1339-L1379

yajo:
Those are tests that are expected to fail. Really they should get decorated with `@mute_logger`, but in any case I don't think they are your real problem.

[The real failures seem to be in `oauth_provider`](https://travis-ci.org/OCA/server-tools/jobs/282692962#L1764-L1780), which [has the branch 9 broken](https://travis-ci.org/OCA/server-tools/branches), BTW.

I'm not blocking this PR for that, since the base branch is broken already. You already fought enough with Travis today IMHO 😊

simahawk:
@pedrobaeza @moylop260 can we merge this?
I'd like to unlock https://github.com/OCA/website-cms/pull/13 and to migrate it to 10.0

simahawk:
FTR reported here https://github.com/OCA/server-tools/issues/1011

PR #867, opened 2017-06-21
[10.0][FIX] server_environment_ir_config_parameter Avoid wrong import

rousseldenis:
Odoo loads modules that contain static directory, so it wanted to load server_environment_files but it does not exist


sbidoul (approved):

lmignon (approved):

rgarnau (approved):
merged this PR and it fixes the errors I was getting previously :+1:

sbidoul:
fast tracking, it's the same approach as in mail_environment

PR #869, opened 2017-06-22
[ADD][10.0] module mail_original_from

astirpe:
This module sets the field "email_from" based on the 'X-Original-From' value of the email header in case it exists, otherwise it takes the standard 'From' value.

It is useful to get correct sender information on messages from e-mail service providers which have implemented DMARC policies.


CasVissers-360ERP (approved):
Tested on multiple machines, 1 customer production machine.

astirpe:
https://github.com/odoo/odoo/pull/20911

mail_original_from/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================
+Mail Original From
+==================
+
```
JayVora-SerpentCS: Should not we rather put this section under usage?
https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

astirpe: Yes, thanks!

PR #872, opened 2017-06-22
[MIG][10.0][datetime_formatter] Migration 

yajo:
Special highlights:

- Fixed some license headers.
- Adhered to normal addon structure.
- Removed `exceptions.py` and use `UserError` instead.
- Allow submodules to import some constants from main namespace.

The rest is basically a standard migration.

@Tecnativa


pedrobaeza (approved):
Maybe you can remove Grupo ESOC from this version...

chienandalu (approved):
:+1:

PR #875, opened 2017-06-26
[MIG] auth_session_timeout: Migration to 10.0

chienandalu:
Auth Session Timeout
=================

This module was written to be able to kill (logout) all inactive sessions since
a given delay. On each request the server checks if the session is yet valid
regarding the expiration delay. If not a clean logout is operated.

cc @Tecnativa


inactive_session_timeout/models/ir_config_parameter.py
```diff
@@ -0,0 +1,30 @@
...
+
+DELAY_KEY = 'inactive_session_time_out_delay'
+IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
+
+
+class IrConfigParameter(models.Model):
+    _inherit = 'ir.config_parameter'
+
+    @api.model
+    @tools.ormcache(skiparg=0)
```
yajo: Don't use skiparg, [it's kept for backwards compatibility, but the new and preferred way of caching is by argument names](https://github.com/odoo/odoo/blob/10.0/odoo/tools/cache.py#L34-L47).

In any case, IMHO this whole file is not needed. [You should just use Odoo's native `get_param` method, which already has caching enabled](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_config_parameter.py#L53-L68).

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+    @classmethod
+    def _check_session_validity(cls, db, uid, passwd):
+        if not request:
+            return
+        session = request.session
+        session_store = root.session_store
+        with registry(db).cursor() as cr:
+            env = Environment(cr, SUPERUSER_ID, {})
+            param_obj = env['ir.config_parameter']
+            delay, urls = param_obj.get_session_parameters(db)
```
yajo: Again, you'll have to change this for 2 calls to `ir.config_parameter.get_param` if you delete `ir_config_parameter.py` file as suggested above.

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+                # and access time.
+                utime(path, None)
+        except OSError:
+            pass
+        return
+
+    @classmethod
+    def check(cls, db, uid, passwd):
+        res = super(ResUsers, cls).check(db, uid, passwd)
+        cls._check_session_validity(db, uid, passwd)
```
yajo: Why not just move all the code from `_check_session_validity` into this method? After all this one does nothing more...

Or even better, why not [move these checks into `check_credentials`](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L431-L436)? That one is a normal method (not a `@classmethod`) [and has a working `self.env` to work with](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L501).

yajo (changes requested):
Code needs a small refactor to make it more maintaiable 😉

chienandalu:
@Yajo Look at it now. Some notes:

- Seems that `_check_session_validity` is not going to be called allways. So I took a halfway solution.
- When the session expires, an exception raises. Do you think is convenient to catch it and explain to the user to reload browser and login again?
- Besides no update is set in default data, every time I update the module the parameters are reset.

inactive_session_timeout/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright 2015 ACSONE - Cédric Pigeon
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Inactive Sessions Timeout",
+    'version': '10.0.1.0.0',
+    'author': "ACSONE SA/NV, "
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://www.tecnativa.com',
```
pedrobaeza: Let original acsone URL.

inactive_session_timeout/data/ir_config_parameter_data.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+<odoo>
+
+    <data noupdate="1">
```
pedrobaeza: Why 2 data noupdate? Put all in the first tag (`<odoo noupdate="1">`)

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,54 @@
...
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    @api.model
+    def _check_session_validity(self):
+        if not request:
+            return
+        session = request.session
+        session_store = root.session_store
+        ipm = self.env['ir.config_parameter']
```
pedrobaeza: You have removed the other model and it was introduced for having a cache mechanism to not query so many times the DB, so please restore it as it was.

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,54 @@
...
+        delay = (int(delay) if delay else False)
+        urls = ipm.get_param(IGNORED_PATH_KEY, '')
+        urls = (urls.split(',') if urls else [])
+        deadline = time() - delay
+        path = session_store.get_session_filename(session.sid)
+        try:
+            if getmtime(path) < deadline:
+                if session.db and session.uid:
+                    session.logout(keep_db=True)
+            elif http.request.httprequest.path not in urls:
```
pedrobaeza: This is totally wrong (not your fault, as it was the same on v8), but the intended behavior for this is not achieved here:

* This should be checked before doing anything. See that if the request has been inactive, you'll get logout anyway.
* What you have to check is if one of the ignored path keys (variable `urls`) is not included in current http request, so the code should be something similar to:
  ```python
  if any([url in http.request.httprequest.path for x in urls]):
      return
  ```

pedrobaeza (changes requested):

inactive_session_timeout/models/res_users.py
chienandalu: It was @Yajo 's request, given that `get_param` has already a cached mechanism at its core: https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_config_parameter.py#L53-L68

pedrobaeza: OK, I see. Then it's enough indeed. Maybe it would be good to throw a comment saying that this is cached, so no performance penalty...

chienandalu:
@pedrobaeza Changes done

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,52 @@
...
+        ipm = self.env['ir.config_parameter']
+        delay = ipm.get_param(DELAY_KEY, 7200)
+        delay = (int(delay) if delay else False)
+        # Note that get_param is cached
+        urls = ipm.get_param(IGNORED_PATH_KEY, '')
+        urls = (urls.split(',') if urls else [])
+        deadline = time() - delay
+        path = session_store.get_session_filename(session.sid)
+        try:
+            if any([url in http.request.httprequest.path for url in urls]):
```
yajo: Sorry, this is incorrect for 2 reasons:

1. Performance-wise, doing `any([... for ...])` makes the generator be converted in a list and then iterated again under the `any`. You can save several iterations by just dropping the `[]` characters. Not critical here since the list will be short, but keep it in mind.
2. If you have `/longpolling,/shop`, then a request to `/blog/longpolling-is-your-friend` will give a false positive.

Instead of this, do:

```py
if any(map(http.request.httprequest.path.startswith, urls)):
```

yajo (changes requested):

chienandalu:
@Yajo @pedrobaeza Changes done

pedrobaeza (approved):

yajo (approved):

inactive_session_timeout/tests/test_inactive_session.py
```diff
@@ -0,0 +1,24 @@
...
+        cls.ipm = cls.env['ir.config_parameter']
+        cls.rus = cls.env['res.users']
+        cls.db = common.get_db_name()
+
+    def test_01_timeout(self):
+        """ Test timeout after delay """
+        self.ipm.set_param(DELAY_KEY, 1)
+        res = self.rus.authenticate(self.db, 'admin', 'admin', {})
+        time.sleep(1.1)
+        res = self.rus.browse(res).check(self.db, res, 'admin')
```
pedrobaeza: You are actually not checking here anything, and you also don't have any request. Better to put that this is a TODO and include in "Known issues" to finish tests. Or @Yajo, do you have a quick suggestion for testing this?

yajo: To have full coverage, you can either mock the session or authenticate through an HttpCase.

Also, after this code (that checks you are OK), I'd try mocking also the method that checks if the session time is old to make it think it is, and then check that you are kicked off.

chienandalu: I've commited a test with HttpCase but I still don't get a request. Can you provide an example to use as a guideline?

yajo: Yes, you need to authenticate through an http request to get a working request (redundant, but that's it). Check this out: https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L270

chienandalu: @Yajo I still can't see how to do such request :confused:

yajo: Did you try calling that method?

chienandalu: Yes, it's what I tried in my last commit: https://github.com/Tecnativa/server-tools/blob/c32b5308088e736c2347965276235a6c1878287d/inactive_session_timeout/tests/test_inactive_session.py#L24

inactive_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,55 @@
...
+                if session.db and session.uid:
+                    session.logout(keep_db=True)
+            else:
+                utime(path, None)
+        except OSError:
+            pass
+        return
+
+    @classmethod
+    def check(cls, db, uid, passwd):
```
yajo: Do not override this method. Override `check_credentials`. Raise an `AccessDenied` if the session is expired.

inactive_session_timeout/tests/test_inactive_session.py
```diff
@@ -0,0 +1,26 @@
...
+            self.rus = env['res.users']
+        self.db = common.get_db_name()
+
+    def test_01_timeout(self):
+        """ Test timeout after delay """
+        self.ipm.set_param(DELAY_KEY, 1)
+        self.assertTrue(self.ipm.get_param(DELAY_KEY), 1)
+        self.authenticate('admin', 'admin')
+        time.sleep(1.1)
+        res = self.rus.browse(1).check(self.db, 1, 'admin')
```
yajo: Assert that this should raise a AccessDenied exception

yajo (changes requested):

inactive_session_timeout/models/res_users.py
pedrobaeza: No, the idea is to return to logout, disauthorizing future operations and returning to login

pedrobaeza:
I have just found that the module was renamed to `auth_session_timeout` in v9. We should rename it too.

pedrobaeza:
See also https://github.com/OCA/server-tools/pull/887

chienandalu:
@pedrobaeza Implemented some of the good ideas of https://github.com/OCA/server-tools/pull/887
https://github.com/OCA/server-tools/pull/875/commits/7f38887d602356138afd269755f2b35b95a437ff

lasley:
IMO we should complete #887 before a merge of this one just so we make sure we have a proper migration path

pedrobaeza:
Well, if this one includes the changes of that PR, we assure migration path, isn't it?

lasley:
This is true, however there were some PR review recommendations that I don't think were attended to here. Reviewing and modifying in two PRs seems like duplicating effort IMO (even though I'm totally guilty of submitting multiple active PRs against the same module, so kettle calling the pot black here).

Also, the commit wasn't cherry picked so we've lost attribution + I'm not entirely certain which aspects of #887 were retained because the commit is squashed into another logical one & the [comment](https://github.com/OCA/server-tools/pull/875#issuecomment-313200501) noted that only some of the good ideas were implemented.

pedrobaeza:
OK, please try to speed up the other PR and we will wait for this.

chienandalu:
Hi, @lasley. Maybe the confusion has been originated because I didn't realize in first place that there was an already migrated version on v9. So now they have diverged quite a bit. From the last PR, I took the idea of split some of functionalities into separate methods and adapted it to my existing code. Although I've got no objection in giving fair attribution. Anyway I'll wait until the PR is fully reviewed :smiley: :+1:

pedrobaeza:
Closing in favor of #580

lasley:
😆  did I seriously review a module that I also had an open PR for? whoops

PR #876, opened 2017-06-26
[ADD] base_manifest_extension

hbrunn:
Extra options in manifest
=========================

This is a technical module to allow developers to make use of a couple of new keys in module manifests.

The following new keys are available currently:

depends_if_installed
    Your module will depend on modules listed here, but only if those modules are already installed. This is useful if your module is supposed to override behavior of a third module if present, but it's not a hard dependency of your module. Think of auth_* and their interactions.

Usage
=====

* depend on this module and use the keys described above

Roadmap
=======

* ``breaks`` to mark some modules as being incompatible. This also could be versioned (``'breaks': ['my_module<<0.4.2']``)
* ``demo_if_installed``, ``data_if_installed``, ``qweb_if_installed`` being dicts with module names as keys and a list of files as values in order to pull some data files in case some other module is installed
* ``rdepends_if_installed`` to make another installed module depend on the current module
* ``_if_module`` on models to load certain models only if the appropriate module is installed


hbrunn:
@PCatinean this is the module we talked about a while ago

PCatinean:
@hbrunn wow, I've been thinking about that conversation lately and you did it again. Just like the extra xml "position" attributes. Nicely done! :1st_place_medal:

PCatinean (approved):

base_manifest_extension/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        result.pop('depends_if_installed')
+        depends = result.setdefault('depends', [])
+        depends.extend(module for module, in cr.fetchall())
+
+    return result
+
+
+module.load_information_from_description_file =\
+    load_information_from_description_file
```
yajo: Are you aware this monkey patch will happen no matter if this addon is installed or just available?

hbrunn: yes, that's how monkey patches work. It might make sense to do this in a post load hook, but it's still instance global then, so we need to take care anyways that this doesn't interfere if it's not wanted. Given it only reacts to keys that shouldn't be in any standard manifest, I don't see a big problem

yajo: I do see it, uninstalled addons should not alter a database behavior. 😕 Can't you use `mock` instead somehow?

pedrobaeza:
I think the module name should be `base_module_extension`, as there are things in the roadmap that will be outside manifest (`_if_module` for example).

base_manifest_extension/__init__.py
hbrunn: this must be a server wide module anyways to catch all corner cases for load order. then we can indeed do this in `_register_hook` and we're happy

hbrunn:
@pedrobaeza `base_module_extension` is very generic. Better idea? I think the name is fine as it is, the `if_module` key is subordinate to fiddling with the manifest imho

pedrobaeza:
OK, let's continue with this name then.

base_manifest_extension/__init__.py
hbrunn: @Yajo your concerns should be addressed by https://github.com/OCA/server-tools/pull/876/commits/2923199994f4242bf361afc820d8c00c714dfdcb

elicoidal (approved):
LGTM

base_manifest_extension/hooks.py
```diff
@@ -0,0 +1,94 @@
...
+            tuple(['installed', 'to install', 'to upgrade']),
+            tuple(manifest['depends_if_installed']),
+        ),
+    )
+    manifest.pop('depends_if_installed')
+    depends = manifest.setdefault('depends', [])
+    depends.extend(module for module, in cr.fetchall())
+
+
+def _get_cr():
```
lasley, line 38: 😮

base_manifest_extension/hooks.py
```diff
@@ -0,0 +1,94 @@
...
+        if 'cr' in frame.f_locals and isinstance(frame.f_locals['cr'], Cursor):
+            cr = frame.f_locals['cr']
+            break
+    return cr
+
+
+def _get_graph():
+    graph = None
+    for frame, filename, lineno, funcname, line, index in inspect.stack():
+        # walk up the stack until we've found a cursor
```
lasley: s/cursor/graph

lasley (approved):
Very nice! One minor change, but it's just in a comment so nothing functional.

base_manifest_extension/hooks.py
hbrunn: done and rebased

PR #880, opened 2017-06-29
[10.0][FIX] sentry Avoid raven import

ghost:
Odoo loads modules that contain static directory, so it wanted to load module sentry that contains import to raven package


bealdav:
What do you think ?PR https://github.com/OCA/server-tools/pull/881

pedrobaeza:
See https://github.com/OCA/server-tools/pull/883 instead

PR #881, opened 2017-06-29
FIX when raven unavailable

bealdav:
Not tested but should works

@pedrobaeza @naglis

### DIFF
--- a/sentry/const.py
+++ b/sentry/const.py
@@ -12,6 +12,7 @@ try:
     import raven
     from raven.conf import defaults
 except ImportError:
+    raven = False
     _logger.debug('Cannot import "raven". Please make sure it is installed.')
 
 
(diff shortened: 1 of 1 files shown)


sentry/const.py
```diff
@@ -34,10 +35,10 @@ def split_multiple(string, delimiter=',', strip_chars=None):
 
 DEFAULT_TRANSPORT = 'threaded'
 TRANSPORT_CLASS_MAP = {
-    'requests_synchronous': raven.transport.RequestsHTTPTransport,
-    'requests_threaded': raven.transport.ThreadedRequestsHTTPTransport,
-    'synchronous': raven.transport.HTTPTransport,
-    'threaded': raven.transport.ThreadedHTTPTransport,
+    'requests_synchronous': getattr(raven, 'transport.RequestsHTTPTransport', False),
```
pedrobaeza, line 38: getattr is not recursive AFAIK. Better to move this to try section

bealdav: or add raven_import = True in try and False in exception and then add a conditional dict

pedrobaeza: Yes, it's another possibility

naglis:
Hi guys,

sorry for breaking `10.0`, did not really anticipate it.

AFAICT, this will not be enough, as imports from raven are used elsewhere at module level.

Had a [quick go at it](https://github.com/OCA/server-tools/compare/10.0...versada:fix_missing_raven), covering (I hope) all cases where imports from `raven` are used at top-level, although it's not pretty. Will test it more thouroughly tomorrow.

Please let me know if you have better ideas how to solve it (I especially dislike the part of [`SentryHandler` and `SanitizePasswordsProcessor` setting to `object`](https://github.com/versada/server-tools/blob/522c0cf7d5ae4b54d1cbc53cf0c8ed753982a643/sentry/logutils.py#L17-L18) if `raven` can't be imported).

bealdav:
:+1: thanks

bealdav:
Guidelines improvement https://github.com/OCA/maintainer-tools/pull/279/files

pedrobaeza:
Are you going to change this PR according the advise from @naglis?

bealdav:
No @naglis have done this branch https://github.com/OCA/server-tools/compare/10.0...versada:fix_missing_raven and is nice

Issue #790, opened 2017-03-22
[RFC] module_auto_update: Community Module Updater


lasley:
This module will provide a working automatic update system for Odoo v10 Community.

A `checksum_dir` column will be added to `ir.module.module`, which will represent the sha1 checksum of the directory (use [`checksumdir`](https://pypi.python.org/pypi/checksumdir/1.0.5)). A `checksum_installed` column will also be added that will represent the checksum of the module version that is currently installed.

The current `Updates` menu in Apps will be replaced with a menu of identical naming that will display a tree view of `ir.module.module` records that have an `checksum_dir` that is different than the `checksum_installed`. 

It might make sense to add a button in the header of the above tree view to allow for `Update Apps List` functionality, which would make for a more fluid process. Or maybe this should just be implicitly called when viewing the Updates list.

Modules to update will be selected in the tree view, with an option provided in the `Action` menu to schedule them for upgrade.

The upgrade itself will be scheduled using the standard Odoo mechanism of setting the `ir.module.status` to `to upgrade`. From there, the pre-existing `Apply Scheduled Upgrades` button in the UI will work.

To take it a step further, we will also add a server parameter to configure a cron job for automatic upgrade of modules at specific timings.

A nice to have feature would be to redirect all incoming web requests to a defined static template during upgrade, which would allow for a more graceful maintenance period.

Anyone have thoughts on this, or pre-existing code that I could base this on?

Yenthe666:
@lasley and anautomatic git fetch and git reset? I'd love that.

pedrobaeza:
Not everyone uses Git, so this is not possible.

:+1: for the community update feature.

lasley:
A few of our instances still aren't Dockerized, so I might actually be ok with adding git handling as an option in order to make legacy a bit easier to work with and push back upgrades a bit.

The one issue we'll run into, at least in my deploys, is that the Odoo user does not have write access to the module directories except in some rare instances (and even then, only their `./static` dir). Is this not the case for you @Yenthe666?

Yenthe666:
Hi @lasley,

Nope, we don't have that problem. Our Odoo user has access inside the `/odoo` folders exactly for this.

Regards,
Yenthe

Yenthe666:
Also @pedrobaeza this could be easily configurable as a setting to allow it for those who want it or use git.

pedrobaeza:
OK then

lasley:
[...]

PR #882, opened 2017-06-29
[ADD] module_auto_update: Create module

hughesbm:
Automate module updates by comparing module directory checksums.

Completes RFC #790

### DIFF
--- a/module_auto_update/README.rst
+++ b/module_auto_update/README.rst
@@ -1,0 +1,76 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==================
+Module Auto Update
+==================
+
+This module will automatically check for and apply module upgrades on a schedule.
+
+Upgrade checking is accomplished by comparing the SHA1 checksums of currently-installed modules to the checksums of corresponding modules in the addons directories.
+
+Installation
+============
+
+Prior to installing this module, you need to:
+
+#. Install checksumdir with `pip install checksumdir`
+#. Ensure all installed modules are up-to-date. When installed, this module will assume the versions found in the addons directories are currently installed.
+
+Configuration
+=============
+
+The default time for checking and applying upgrades is 3:00 AM (UTC). To change this schedule, modify the "Perform Module Upgrades" scheduled action.
+
+This module will ignore .pyc and .pyo file extensions by default. To modify this, create a module_auto_update.checksum_excluded_extensions system parameter with the desired extensions listed as comma-separated values.
+
+Usage
+=====
+
+Modules scheduled for upgrade can be viewed by clicking the "Updates" menu item in the Apps sidebar.
+
+To perform upgrades manually, click the "Apply Scheduled Upgrades" menu item in the Apps sidebar.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Brent Hughes <brent.hughes@laslabs.com>
+* Juan José Scarafía <jjs@adhoc.com.ar>
+
+Do not contact contributors directly about support or help with technical issues.
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/module_auto_update/__init__.py
+++ b/module_auto_update/__init__.py
@@ -1,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import models
+from . import wizards
+from .hooks import post_init_hook
--- a/module_auto_update/__manifest__.py
+++ b/module_auto_update/__manifest__.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Module Auto Update',
+    'summary': 'Automatically update Odoo modules',
+    'version': '10.0.1.0.0',
+    'category': 'Extra Tools',
+    'website': 'https://odoo-community.org/',
+    'author': 'LasLabs, '
+              'Juan José Scarafía, '
+              'Odoo Community Association (OCA)',
+    'license': 'LGPL-3',
+    'application': False,
+    'installable': True,
+    'post_init_hook': 'post_init_hook',
+    'external_dependencies': {
+        'python': [
+            'checksumdir',
+        ],
+    },
+    'depends': [
+        'base',
+    ],
+    'data': [
+        'views/module_views.xml',
+        'data/cron_data.xml',
+    ],
+}
--- a/module_auto_update/data/cron_data.xml
+++ b/module_auto_update/data/cron_data.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record model="ir.cron" id="module_check_upgrades_cron">
+        <field name="name">Perform Module Upgrades</field>
+        <field name="active" eval="True"/>
+        <field name="user_id" ref="base.user_root"/>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field name="nextcall" eval="(DateTime.now() + timedelta(days= +1)).strftime('%Y-%m-%d 3:00:00')"/>
+        <field name="model">base.module.upgrade</field>
+        <field name="function">upgrade_module</field>
+        <field name="args" eval="'()'"/>
+    </record>
+</odoo>
--- a/module_auto_update/hooks.py
+++ b/module_auto_update/hooks.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    installed_modules = env['ir.module.module'].search([
+        ('state', '=', 'installed'),
+    ])
+    for r in installed_modules:
+        r.checksum_installed = r.checksum_dir
--- a/module_auto_update/models/module.py
+++ b/module_auto_update/models/module.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+
+from odoo import api, fields, models
+from odoo.modules.module import get_module_path
+
+_logger = logging.getLogger(__name__)
+try:
+    from checksumdir import dirhash
+except ImportError:
+    _logger.debug('Cannot `import checksumdir`.')
+
+
+class Module(models.Model):
+    _inherit = 'ir.module.module'
+
+    checksum_dir = fields.Char(
+        compute='_compute_checksum_dir',
+    )
+    checksum_installed = fields.Char()
+
+    @api.depends('name')
+    def _compute_checksum_dir(self):
+        exclude = self.env["ir.config_parameter"].get_param(
+            "module_auto_update.checksum_excluded_extensions",
+            "pyc,pyo",
+        ).split(",")
+
+        for r in self:
+            r.checksum_dir = dirhash(
+                get_module_path(r.name),
+                'sha1',
+                excluded_extensions=exclude,
+            )
+
+    def _store_checksum_installed(self, vals):
+        if self.env.context.get('retain_checksum_installed'):
+            return
+        if 'checksum_installed' not in vals:
+            if vals.get('state') == 'installed':
+                for r in self:
+                    r.checksum_installed = r.checksum_dir
+            elif vals.get('state') == 'uninstalled':
+                self.write({'checksum_installed': False})
+
+    @api.multi
+    def button_uninstall_cancel(self):
+        return super(
+            Module,
+            self.with_context(retain_checksum_installed=True),
+        ).button_uninstall_cancel()
+
+    @api.multi
+    def button_upgrade_cancel(self):
+        return super(
+            Module,
+            self.with_context(retain_checksum_installed=True),
+        ).button_upgrade_cancel()
+
+    @api.model
+    def create(self, vals):
+        res = super(Module, self).create(vals)
+        res._store_checksum_installed(vals)
+        return res
+
+    @api.model
+    def update_list(self):
+        res = super(Module, self).update_list()
+        installed_modules = self.search([('state', '=', 'installed')])
+        upgradeable_modules = installed_modules.filtered(
+            lambda r: r.checksum_dir != r.checksum_installed,
+        )
+        upgradeable_modules.write({'state': "to upgrade"})
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super(Module, self).write(vals)
+        self._store_checksum_installed(vals)
+        return res
--- a/module_auto_update/tests/__init__.py
+++ b/module_auto_update/tests/__init__.py
@@ -1,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import test_module
+from . import test_module_upgrade
--- a/module_auto_update/views/module_views.xml
+++ b/module_auto_update/views/module_views.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Module Search View -->
+    <record id="module_view_search" model="ir.ui.view">
+        <field name="name">updates.module.search</field>
+        <field name="model">ir.module.module</field>
+        <field name="inherit_id" ref="base.view_module_filter"/>
+        <field name="arch" type="xml">
+            <field name="category_id" position="after">
+                <filter name="scheduled_upgrades" string="Scheduled Upgrades" domain="[('state', '=', 'to upgrade')]"/>
+            </field>
+        </field>
+    </record>
+
+    <!--Open Updates Action (updates apps list first)-->
+    <record id="module_action_open_updates" model="ir.actions.server">
+        <field name="name">Open Updates and Update Apps List Server Action</field>
+        <field name="model_id" ref="model_ir_module_module"/>
+        <field name="code">
+            if model.update_list():
+                action = {
+                    'name': 'Updates',
+                    'type': 'ir.actions.act_window',
+                    'res_model': 'ir.module.module',
+                    'view_type': 'form',
+                    'view_mode': 'tree,form',
+                    'target': 'main',
+                    'context': '{"search_default_scheduled_upgrades": 1}',
+                }
+        </field>
+    </record>
+
+    <!--Apps / Updates menu item-->
+    <menuitem
+        name="Updates"
+        action="module_action_open_updates"
+        id="module_menu_updates"
+        groups="base.group_no_one"
+        parent="base.menu_management"
+        sequence="20"/>
+
+    <!-- Menu in Settings > Technical for standard Updates link -->
+    <menuitem parent="base.menu_custom" sequence="27" name="Modules" id="menu_default_modules"/>
+
+    <!-- Moved standard Updates link -->
+    <record model="ir.ui.menu" id="base.menu_module_updates">
+        <field name="parent_id" ref="menu_default_modules"/>
+    </record>
+</odoo>
--- a/module_auto_update/models/__init__.py
+++ b/module_auto_update/models/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import module
--- a/module_auto_update/tests/test_module.py
+++ b/module_auto_update/tests/test_module.py
@@ -1,0 +1,205 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+import tempfile
+
+import mock
+
+from odoo.modules import get_module_path
+from odoo.tests.common import TransactionCase
+
+from .. import post_init_hook
+
+_logger = logging.getLogger(__name__)
+try:
+    from checksumdir import dirhash
+except ImportError:
+    _logger.debug('Cannot `import checksumdir`.')
+
+model = 'odoo.addons.module_auto_update.models.module'
+
+
+class TestModule(TransactionCase):
+
+    def setUp(self):
+        super(TestModule, self).setUp()
+        module_name = 'module_auto_update'
+        self.own_module = self.env['ir.module.module'].search([
+            ('name', '=', module_name),
+        ])
+        self.own_dir_path = get_module_path(module_name)
+        self.own_checksum = dirhash(
+            self.own_dir_path,
+            'sha1',
+            excluded_extensions=['pyc', 'pyo'],
+        )
+
+    @mock.patch('%s.get_module_path' % model)
+    def create_test_module(self, vals, get_module_path_mock):
+        get_module_path_mock.return_value = self.own_dir_path
+        test_module = self.env['ir.module.module'].create(vals)
+        return test_module
+
+    def test_compute_checksum_dir(self):
+        """It should compute the directory's SHA-1 hash"""
+        self.assertEqual(
+            self.own_module.checksum_dir, self.own_checksum,
+            'Module directory checksum not computed properly',
+        )
+
+    def test_compute_checksum_dir_ignore_excluded(self):
+        """It should exclude .pyc/.pyo extensions from checksum
+        calculations"""
+        with tempfile.NamedTemporaryFile(
+                suffix='.pyc', dir=self.own_dir_path):
+            self.assertEqual(
+                self.own_module.checksum_dir, self.own_checksum,
+                'SHA1 checksum does not ignore excluded extensions',
+            )
+
+    def test_compute_checksum_dir_recomputes_when_file_added(self):
+        """It should return a different value when a non-.pyc/.pyo file is
+        added to the module directory"""
+        with tempfile.NamedTemporaryFile(
+                suffix='.py', dir=self.own_dir_path):
+            self.assertNotEqual(
+                self.own_module.checksum_dir, self.own_checksum,
+                'SHA1 checksum not recomputed',
+            )
+
+    def test_store_checksum_installed_state_installed(self):
+        """It should set the module's checksum_installed equal to
+        checksum_dir when vals contain state 'installed'"""
+        self.own_module.checksum_installed = 'test'
+        self.own_module._store_checksum_installed({'state': 'installed'})
+        self.assertEqual(
+            self.own_module.checksum_installed, self.own_module.checksum_dir,
+            'Setting state to installed does not store checksum_dir '
+            'as checksum_installed',
+        )
+
+    def test_store_checksum_installed_state_uninstalled(self):
+        """It should clear the module's checksum_installed when vals
+        contain state 'uninstalled'"""
+        self.own_module.checksum_installed = 'test'
+        self.own_module._store_checksum_installed({'state': 'uninstalled'})
+        self.assertEqual(
+            self.own_module.checksum_installed, False,
+            'Setting state to uninstalled does not clear checksum_installed',
+        )
+
+    def test_store_checksum_installed_vals_contain_checksum_installed(self):
+        """It should not set checksum_installed to False or checksum_dir when
+        a checksum_installed is included in vals"""
+        self.own_module.checksum_installed = 'test'
+        self.own_module._store_checksum_installed({
+            'state': 'installed',
+            'checksum_installed': 'test',
+        })
+        self.assertEqual(
+            self.own_module.checksum_installed, 'test',
+            'Providing checksum_installed in vals did not prevent overwrite',
+        )
+
+    def test_store_checksum_installed_with_retain_context(self):
+        """It should not set checksum_installed to False or checksum_dir when
+        self has context retain_checksum_installed=True"""
+        self.own_module.checksum_installed = 'test'
+        self.own_module.with_context(
+            retain_checksum_installed=True,
+        )._store_checksum_installed({'state': 'installed'})
+        self.assertEqual(
+            self.own_module.checksum_installed, 'test',
+            'Providing retain_checksum_installed context did not prevent '
+            'overwrite',
+        )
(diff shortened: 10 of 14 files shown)


module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Module Auto Update",
+    "summary": "Automatically update Odoo modules",
+    "version": "10.0.1.0.0",
+    "category": "Uncategorized",
+    "website": "https://odoo-community.org/",
+    'author': 'Brent Hughes, '
```
lasley: `LasLabs` if on the clock. You go in Contributors section of ReadMe using LasLabs email.

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+              'Juan José Scarafía, '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
```
lasley: relicense to LGPL-3, this was OK'd in the RFC https://github.com/OCA/server-tools/issues/790#issuecomment-299258642

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
lasley: LGPL, apply to all

module_auto_update/README.rst
```diff
@@ -0,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
lasley: LGPL

module_auto_update/README.rst
```diff
@@ -0,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
lasley: LGPL

module_auto_update/README.rst
```diff
@@ -0,0 +1,83 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
```
lasley: lgpl

module_auto_update/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+from . import models
+from . import wizard
+from . import tests
```
lasley: Newline at end of file

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    'external_dependencies': {
+        'python': ['checksumdir'],
+    },
+    'data': [
+        'views/module_views.xml',
+        'views/module_schedule_upgrade.xml',
+        'security/ir.model.access.csv',
+        'data/cron_data.xml',
+    ],
+    'demo': [],
```
lasley: Remove empty keys

module_auto_update/data/cron_data.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
lasley: `<odoo noupdate="1">` & can remove data tags

module_auto_update/data/cron_data.xml
```diff
@@ -0,0 +1,30 @@
...
+        <record model="ir.cron" id="module_check_upgrades_cron">
+            <field name="name">Check for Module Upgrades</field>
+            <field name="active" eval="True"/>
+            <field name="user_id" ref="base.user_root"/>
+            <field name="interval_number">1</field>
+            <field name="interval_type">days</field>
+            <field name="numbercall">-1</field>
+            <field name="nextcall" eval="(DateTime.now() + timedelta(days= +1)).strftime('%Y-%m-%d 3:00:00')"/>
+            <field name="model" eval="'ir.module.module'"/>
+            <field name="function" eval="'check_upgrade_availability'"/>
```
lasley: Simple strings should just be in the XML text node. Eval is for python code:

`<function>check_upgrade_availability</function>`

module_auto_update/data/cron_data.xml
```diff
@@ -0,0 +1,30 @@
...
+    <data noupdate="1">
+        <record model="ir.cron" id="module_check_upgrades_cron">
+            <field name="name">Check for Module Upgrades</field>
+            <field name="active" eval="True"/>
+            <field name="user_id" ref="base.user_root"/>
+            <field name="interval_number">1</field>
+            <field name="interval_type">days</field>
+            <field name="numbercall">-1</field>
+            <field name="nextcall" eval="(DateTime.now() + timedelta(days= +1)).strftime('%Y-%m-%d 3:00:00')"/>
+            <field name="model" eval="'ir.module.module'"/>
```
lasley: `<field name="model">ir.module.module</field>`

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @api.multi
+    def _compute_checksum_dir(self):
+        for r in self:
+            r.checksum_dir = dirhash(get_module_path(r.name), 'sha1')
+
+    @api.multi
+    def button_immediate_upgrade(self):
+        res = super(Module, self).button_immediate_upgrade()
+
```
lasley: Remove the spacing - it's a bit much IMO

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+    @api.model
+    def check_upgrade_availability(self):
+        self.update_list()
+        self.determine_upgradability()
+
+    @api.depends('checksum_dir')
+    def determine_upgradability(self):
+        upgradable_modules = self.search(
+            [('state', 'in', ['installed', 'to upgrade', 'to remove'])]
```
lasley: Trailing comma

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+        return res
+
+    @api.model
+    def check_upgrade_availability(self):
+        self.update_list()
+        self.determine_upgradability()
+
+    @api.depends('checksum_dir')
+    def determine_upgradability(self):
+        upgradable_modules = self.search(
```
lasley: s/upgradable/upgradeable

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+            [('state', 'in', ['installed', 'to upgrade', 'to remove'])]
+        )
+
+        for r in upgradable_modules:
```
lasley: Filters ftw:

``` python
upgradable_modules = upgradable_modules.filtered(
    lambda r: r.checksum_dir != r.checksum_installed
)
upgradable_modules.write({
    'upgrade_available': True,
})
```

It's more lines, but it is less McCabe score & performs the write in one database query. Odoo ActiveRecord pattern may seem nice, but it hits the db each time - vs a write can update a crapton of records with the same data at once.

ActiveRecord in SQL:
``` sql
UPDATE my_table SET col = val WHERE id = 1;
UPDATE my_table SET col = val WHERE id = 2;
UPDATE my_table SET col = val WHERE id = 3;
UPDATE my_table SET col = val WHERE id = 4;
```

Write:

``` sql
UPDATE my_table SET col = val WHERE id in [1, 2, 3, 4];
```

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+            if r.checksum_dir != r.checksum_installed:
+                r.upgrade_available = True
+
+    @api.depends('checksum_dir')
+    def post_upgrade_actions(self):
+        for r in self:
+            r.checksum_installed = r.checksum_dir
```
lasley:
```
r.write({
    'checksum_installed': r.checksum_dir,
    'upgrade_available': False,
})
```

module_auto_update/views/module_schedule_upgrade.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
lasley: can remove data tag

module_auto_update/views/module_schedule_upgrade.xml
```diff
@@ -0,0 +1,25 @@
...
+            <field name="name">action_module_schedule_upgrade</field>
+            <field name="action_id" ref="action_server_module_schedule_upgrade"/>
+            <field name="value" eval="'ir.actions.server,' + str(ref('action_server_module_schedule_upgrade'))"/>
+            <field name="key">action</field>
+            <field name="model_id" ref="model_ir_module_module"/>
+            <field name="model">ir.module.module</field>
+            <field name="key2">client_action_multi</field>
+        </record>
+    </data>
+</odoo>
```
lasley: newline at EOF

module_auto_update/views/module_views.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
lasley: no data

module_auto_update/wizard/module_upgrade.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ModuleUpgrade(models.TransientModel):
+    _inherit = 'base.module.upgrade'
+
+    @api.multi
+    def upgrade_module(self):
+        mods = self.env['ir.module.module'].search(
+            [('state', '=', 'to upgrade')]
+        )
+
```
lasley: kill the spaces

lasley (changes requested):
Excellent first PR @hughesbm!

Comments inline

module_auto_update/models/module.py
yajo: Can't this be into `button_immediate_upgrade` directly?

module_auto_update/models/module.py
```diff
@@ -0,0 +1,61 @@
...
+_logger = logging.getLogger(__name__)
+try:
+    from checksumdir import dirhash
+except ImportError:
+    _logger.debug('Cannot `import checksumdir`.')
+
+
+class Module(models.Model):
+    _inherit = 'ir.module.module'
+    _rec_name = 'name'
```
yajo: You don't need this line.

module_auto_update/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_ir_module_module_user,ir_module_module user,model_ir_module_module,,1,0,0,0
```
yajo: Do you really need this file?

yajo (approved):
Code OK (apart from @lasley's comments)

module_auto_update/models/module.py
hughesbm: @Yajo It's also called from `upgrade_module` in `base.module.upgrade`; `button_immediate_upgrade` isn't used for applying scheduled upgrades.

module_auto_update/security/ir.model.access.csv
hughesbm: I guess not! Thanks! 👍

hughesbm:
The requested changes have been addressed. I still need to add tests, however.

module_auto_update/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import models
+from . import wizard
+from . import tests
```
tafaRU: This line is not needed as `tests` is automatically inspected by Odoo.

lasley:
Ping @jjscarafia

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Module Auto Update",
+    "summary": "Automatically update Odoo modules",
+    "version": "10.0.1.0.0",
+    "category": "Uncategorized",
```
yajo: I'd use `Extra Tools`

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+    _inherit = 'ir.module.module'
+
+    checksum_installed = fields.Char(
+        string="Installed Module Checksum",
+    )
+    checksum_dir = fields.Char(
+        string="Current Module Checksum",
+        compute='_compute_checksum_dir',
+    )
+    upgrade_available = fields.Boolean(
```
yajo: Could we just set state=to upgrade?

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @api.model
+    def check_upgrade_availability(self):
+        self.update_list()
+        self.determine_upgradeability()
+
+    @api.depends('checksum_dir')
+    def determine_upgradeability(self):
+        upgradeable_modules = self.search(
+            [('state', 'in', ['installed', 'to upgrade', 'to remove', ])]
```
yajo: If we use state to track upgradeable addons, then you should remove `to upgrade` from here.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+        res = super(Module, self).button_immediate_upgrade()
+        self.post_upgrade_actions()
+        return res
+
+    @api.model
+    def check_upgrade_availability(self):
+        self.update_list()
+        self.determine_upgradeability()
+
+    @api.depends('checksum_dir')
```
yajo: I'm not sure... does `api.depends` work if the method is not marked as a `compute` method on some field?

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+            [('state', 'in', ['installed', 'to upgrade', 'to remove', ])]
+        )
+        upgradeable_modules = upgradeable_modules.filtered(
+            lambda r: r.checksum_dir != r.checksum_installed
+        )
+        upgradeable_modules.write({
+            'upgrade_available': True,
+        })
+
+    @api.depends('checksum_dir')
```
yajo: Same about depends here

module_auto_update/views/module_views.xml
```diff
@@ -0,0 +1,77 @@
...
+            if model.update_list():
+                action = {
+                    'name': 'Updates',
+                    'type': 'ir.actions.act_window',
+                    'res_model': 'ir.module.module',
+                    'view_type': 'form',
+                    'view_mode': 'tree,form',
+                    'target': 'main',
+                    'context': '{"search_default_upgrade_available": 1}',
+                }
```
yajo: If we use the state as said above, could we move all the checksum computing into `update_list()`, so we don't have to create additional actions, menu items and methods? This way, if the admin uses the standard system to upgrade the addons list, he will always update their checksums too. It seems like what he'd expect to happen.

module_auto_update/data/cron_data.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field name="nextcall" eval="(DateTime.now() + timedelta(days= +1)).strftime('%Y-%m-%d 3:00:00')"/>
+        <field name="model">ir.module.module</field>
+        <field name="function">check_upgrade_availability</field>
+        <field name="args" eval="'()'"/>
+    </record>
+
+    <record model="ir.cron" id="base_module_upgrade_cron">
+        <field name="name">Perform Scheduled Module Upgrades</field>
```
yajo: I see no sense on having 2 crons. Checking the upgrades without upgrading makes no sense, as upgrading without checking what's to upgrade neither. Besides, you could have a race condition where there are more than 1 cron workers. Please have a Python method that does all and run it in a single cron.

module_auto_update/wizard/module_upgrade.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import api, models
+
+
+class ModuleUpgrade(models.TransientModel):
+    _inherit = 'base.module.upgrade'
+
+    @api.multi
+    def upgrade_module(self):
+        mods = self.env['ir.module.module'].search(
+            [('state', '=', 'to upgrade')]
```
yajo: Add `to install` too?

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+        compute='_compute_checksum_dir',
+    )
+    upgrade_available = fields.Boolean(
+        default=False,
+    )
+
+    @api.multi
+    def _compute_checksum_dir(self):
+        for r in self:
+            r.checksum_dir = dirhash(get_module_path(r.name), 'sha1')
```
yajo: One question... If we would have addons directory full of `.pyc` or `.pyo` files, and under a containerized frozen environment where those files are autoregenerated on each build, would that produce a different checksum? I guess it should be the same because if the pycompiled file is based on the same input python file, it should produce the same output, but did you check this? Or is it possible to ignore such files?

yajo (changes requested):
I feel this needs a little more simplification.

module_auto_update/models/module.py
lasley: I think you're correct here Jairo. The `pyc` files contain a `moddate` header, which is the date of compilation as I understand it. Due to this, the hash can and will change when the file is recomputed, regardless of whether there were actually any changes.

Luckily `checksumdir` has an `excluded_extension` key, which interestingly enough has `pyc` in it as the example:

```
sha256hash = dirhash(directory, 'sha256', excluded_extensions=['pyc'])
```

IMO we should use an array declared at class level to determine the extensions that should be excluded. We could also use a method, but I think that's a bit overengineered

yajo: The simplest balance between simplicity and configurability is an unexisting `ir.config_parameter` with sane defaults:

```python
exclude = self.env["ir.config_parameter"].get_param(
    "module_auto_update.checksum_excluded_extensions", 
    "pyc,pyo").split(",")
```

This code with a little note in the configuration section of README should do the trick.

lasley: @hughesbm - here's an example of [ir.config_param](https://github.com/OCA/website/blob/10.0/website_form_recaptcha/data/ir_config_parameter_data.xml)

lasley: Ummm but don't copy whatever the hell I did with the tab/space mixture there. Spaces4Lyfe!

module_auto_update/data/cron_data.xml
lasley: I'm good with automating the whole thing. Let's hide the `Apply Scheduled Upgrades` button in the UI then as well.

module_auto_update/models/module.py
lasley: It does not; you can only edit computed fields with depends methods. They technically run, but the changes aren't stored to DB because you are in the cache layer.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    checksum_dir = fields.Char(
+        string="Current Module Checksum",
+        compute='_compute_checksum_dir',
+    )
+    upgrade_available = fields.Boolean(
+        default=False,
+    )
+
+    @api.multi
```
lasley: `@api.depends('name')`

module_auto_update/models/module.py
yajo: Yeah, although you don't really need to provide a `ir.config_param` record if you add a fallback value as in the code I put above.

module_auto_update/models/module.py
yajo: Then this decorator makes no sense

module_auto_update/data/cron_data.xml
yajo: wouldn't it be better to extend that method and make the cron call it instead?

module_auto_update/models/module.py
lasley: Agreed - I was mainly just explaining for my newb 😉

module_auto_update/data/cron_data.xml
lasley: Ohhhh yes, I like that better. That way we can trigger a manual update without the cron

hughesbm:
In addition to the requested changes, I made a fairly large change to how `checksum_installed` is stored. The strategy of overriding module upgrade methods to store the checksum completely ignored module installation; modules would be immediately scheduled for upgrade after installation.

To deal with this, this module now uses the `@api.depends` decorator to watch for module state changes, and recomputes `checksum_installed` when state changes to `'installed'`. Cancellation of upgrades and uninstalls also changes the state to 'installed', however, so those methods were overridden to prevent undesired behavior.

module_auto_update/README.rst
```diff
@@ -6,38 +6,31 @@
...
+=============
 
-To change Scheduled Action settings:
+The default time for checking and applying upgrades is 3:00 AM (UTC). To change this schedule, modify the "Perform Module Upgrades" scheduled action.
 
-#. Enable developer mode.
-#. Navigate to Settings / Technical / Automation / Scheduled Actions
-#. Click "Check for Module Upgrades" or "Perform Scheduled Module Upgrades."
-#. Click "Edit", enter desired changes, then click "Save."
+This module will ignore .pyc and .pyo file extensions by default. To modify this, create a module_auto_update.checksum_excluded_extensions system parameter with the desired extensions.
```
lasley: with the desired extensions listed as comma separated values.

module_auto_update/models/module.py
```diff
@@ -17,49 +17,65 @@
...
-        string="Installed Module Checksum",
-    )
     checksum_dir = fields.Char(
-        string="Current Module Checksum",
         compute='_compute_checksum_dir',
     )
-    upgrade_available = fields.Boolean(
-        default=False,
+    checksum_installed = fields.Char(
+        compute='_compute_checksum_installed',
```
yajo: This field must not be computed. It only has to be updated when addon gets updated too, but it is a normal char field.

module_auto_update/models/module.py
```diff
@@ -17,49 +17,65 @@
...
+
+        for r in self:
+            if r.state == 'installed':
+                r.checksum_installed = r.checksum_dir
+            else:
+                r.checksum_installed = stored_checksums[r.id]
+
+    @api.multi
+    def button_uninstall_cancel(self):
+        return self.retain_checksum_installed(
```
yajo: You don't need these kind of weird hacks if `checksum_installed` is not a computed field.

yajo (changes requested):
The way you are using the computation of `checksum_installed` seems to go against the purpose of such kind of fields.

If you want to check when an addon's state changes to `installed`, but skip doing things under some situations, you should override `write` and `create` methods, and put something like this inside them:

```python
def write(self, vals):
  if vals.get("state") == "installed" and not self.env.context.get("retain_checksum_installed"):
    for one in self:
      one.checksum_installed = one.checksum_dir
  return super(Class, self).write(vals)
```

Also, on the methods where you need to retain the installed checksum:

```python
def medthod_X(self):
  return super(Class, self.with_context(retain_checksum_installed=True)).method_X()
```

This would follow the normal Odoo framework logic and hooks to make all work as expected and be written as expected 😉

module_auto_update/models/module.py
hughesbm: Yeah, this is a lot better. 😆  Thanks!

module_auto_update/models/module.py
```diff
@@ -39,36 +36,30 @@ def _compute_checksum_dir(self):
...
-    def retain_checksum_installed(self, func):
-        installed_checksums = self.mapped('checksum_installed')
-        res = func()
-        for index, r in enumerate(self):
-            r.checksum_installed = installed_checksums[index]
-        return res
+    @api.model
+    def create(self, vals):
+        self._store_checksum_installed(vals)
+        return super(Module, self).create(vals)
```
yajo: First, create the record, and then call to `result._store_checksum_installed(vals)`, or it will have no effect.

module_auto_update/models/module.py
```diff
@@ -79,3 +70,8 @@ def update_list(self):
         )
         upgradeable_modules.write({'state': "to upgrade", })
         return res
+
+    @api.multi
+    def write(self, vals):
+        self._store_checksum_installed(vals)
+        return super(Module, self).write(vals)
```
yajo: First write the record and then call the other method, or it could have undesired results.

module_auto_update/models/module.py
```diff
@@ -39,36 +36,30 @@ def _compute_checksum_dir(self):
...
-    @api.depends('state')
-    def _compute_checksum_installed(self):
-        stored_checksums = {}
-        for r in self.read(['checksum_installed']):
-            stored_checksums[r['id']] = r['checksum_installed']
-
-        for r in self:
-            if r.state == 'installed':
+    def _store_checksum_installed(self, vals):
+        if (vals.get('state') == 'installed' and
```
yajo: I'd also check `"checksum_installed" not in vals`, to avoid double writes if the field is being manually written.

module_auto_update/wizard/module_upgrade.py
yajo: Just noting this seems OK now.

module_auto_update/models/module.py
yajo: Just noting this seems OK now.

yajo (changes requested):

yajo:
BTW, you have conflicts.

module_auto_update/models/module.py
```diff
@@ -37,11 +37,14 @@ def _compute_checksum_dir(self):
...
+        if ('checksum_installed' not in vals and
                 not self.env.context.get('retain_checksum_installed')):
-            for r in self:
-                r.checksum_installed = r.checksum_dir
+            if vals.get('state') == 'installed':
+                for r in self:
+                    r.checksum_installed = r.checksum_dir
+            elif vals.get('state') == 'uninstalled':
+                for r in self:
+                    r.checksum_installed = ''
```
yajo: Set it to `False`, which is the standard in Odoo when we have an empty field.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+    _logger.debug('Cannot `import checksumdir`.')
+
+
+class Module(models.Model):
+    _inherit = 'ir.module.module'
+
+    checksum_dir = fields.Char(
+        compute='_compute_checksum_dir',
+    )
+    checksum_installed = fields.Char()
```
yajo, line 23: I feel this one should be read only, mostly now that you are exposing it in the UI.
You should also add `string` and `help` attributes to both fields for the same reason, to let the user know what's this.

yajo (changes requested):

module_auto_update/models/module.py
hughesbm: Makes sense. Does exposing these fields in the UI have value, or should I take it out again? I keep wavering on this.

lasley: The value would be for the human. If a human cares, expose it - otherwise don't. In this case, I'd say don't.

module_auto_update/hooks.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
lasley, line 9: This isn't tested. Also I don't think it's implemented correctly, because coverage should have a hit regardless of test status for hooks.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @api.depends('name')
+    def _compute_checksum_dir(self):
+        exclude = self.env["ir.config_parameter"].get_param(
+            "module_auto_update.checksum_excluded_extensions",
+            "pyc,pyo"
```
lasley: trailing comma

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+        ).split(",")
+
+        for r in self:
+            r.checksum_dir = dirhash(
+                get_module_path(r.name),
+                'sha1',
+                excluded_extensions=exclude
```
lasley: trailing comma

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+
+    def _store_checksum_installed(self, vals):
+        if ('checksum_installed' not in vals and
```
lasley: Simplify this conjoined if by moving `not self.env.context` above, inverse the logic, and return.

module_auto_update/models/module.py
lasley: `self.write({'checksum_installed': False})` and kill the loop

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+                    r.checksum_installed = r.checksum_dir
+            elif vals.get('state') == 'uninstalled':
+                for r in self:
+                    r.checksum_installed = ''
+
+    @api.multi
+    def button_uninstall_cancel(self):
+        return super(
+            Module,
+            self.with_context(retain_checksum_installed=True)
```
lasley: trailing comma

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+        ).button_uninstall_cancel()
+
+    @api.multi
+    def button_upgrade_cancel(self):
+        return super(
+            Module,
+            self.with_context(retain_checksum_installed=True)
```
lasley: trailing comma

module_auto_update/models/module.py
```diff
@@ -0,0 +1,83 @@
...
+        return res
+
+    @api.model
+    def update_list(self):
+        res = super(Module, self).update_list()
+        installed_modules = self.search([('state', '=', 'installed')])
+        upgradeable_modules = installed_modules.filtered(
+            lambda r: r.checksum_dir != r.checksum_installed
+        )
+        upgradeable_modules.write({'state': "to upgrade", })
```
lasley: no trailing comma. The point of the trailing comma is to allow for an option to be added without editing multiple lines in the diff. In the instance of a single line, this is a null point

module_auto_update/tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import test_module
+from . import test_module_upgrade
```
lasley: Newline at EOF

module_auto_update/wizard/module_upgrade.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class ModuleUpgrade(models.TransientModel):
+    _inherit = 'base.module.upgrade'
+
+    @api.multi
+    def upgrade_module_cancel(self):
+        return super(
+            ModuleUpgrade,
+            self.with_context(retain_checksum_installed=True)
```
lasley: trailing comma

lasley (changes requested):

hughesbm:
Runbot is currently failing when creating the odoo instance, not due to changes here as far as I can tell.

lasley:
Runbot fail due to https://github.com/OCA/maintainer-quality-tools/pull/464

lasley (approved):

elicoidal (approved):

moylop260 (approved):

elicoidal:
@yajo please review again (I think your remark was attended)

yajo (approved):

yajo:
It looks ready to merge now, just wondering if we could get rid of 505af1e77f312760b51b277a276862325760f2f0 and put everything in a single commit, seems to add no value at my eyes (maybe @jjscarafia has an opinion that we must respect) 😉

lasley:
> just wondering if we could get rid of 505af1e and put everything in a single commit

Agreed - this is pretty much a rewrite at this point. Only a bit of boilerplate was carried over from what I can tell.

hughesbm:
Everything is now in a single commit. Thanks for the fantastic code reviews! 😄

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    'external_dependencies': {
+        'python': [
+            'checksumdir',
+        ],
+    },
+    'data': [
+        'views/module_views.xml',
+        'data/cron_data.xml',
+    ],
+    'auto_install': True,
```
yajo: I was just about to merge, but saw this. Please do not set it to autoinstall. Also, add dependency on `base` module.

yajo (changes requested):

module_auto_update/__manifest__.py
hughesbm: Thanks! I have made the requested changes and squashed.

hughesbm: Apparently those changes weren't as consequence-free as I expected and a test is now failing. Working on it now.

hughesbm:
@Yajo The changes you requested are done, and I fixed the test that broke as a result. I'll clean up the commits after approval. Sorry about the mess, I got too hasty trying to wrap this up quickly.

yajo (approved):
Great! That's what I love about good tests like yours, they fail whey they must! 😉

lasley:
Congrats on your first module Brent! Not very often your first one gets to be this fun or popular!

Yenthe666:
Hooray! I'll test this next week too. Kudos!

PR #887, opened 2017-07-05
Module auth_session_timeout: Pluggability

jmorgannz:
---------------------------

- Refactor to allow other modules to inherit and augment or override the following:
  - Session expiry time (deadline) calculation
  - Ignored URLs
  - Final session expiry (with possibility to late-abort)
- Re-ordered functionality to remove unnecessary work, as this code is called very often.
- Do not expire a session if delay gets set to zero (or unset / false)

### DIFF
--- a/auth_session_timeout/__init__.py
+++ b/auth_session_timeout/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
+# (c) 2015 ACSONE SA/NV, Dhinesh D
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
--- a/auth_session_timeout/__init__.py
+++ b/auth_session_timeout/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D
+# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
--- a/auth_session_timeout/__openerp__.py
+++ b/auth_session_timeout/__openerp__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
+# (c) 2015 ACSONE SA/NV, Dhinesh D
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
--- a/auth_session_timeout/__openerp__.py
+++ b/auth_session_timeout/__openerp__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D
+# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
--- a/auth_session_timeout/__openerp__.py
+++ b/auth_session_timeout/__openerp__.py
@@ -9,7 +9,7 @@
     'summary': """
         This module disable all inactive sessions since a given delay""",
 
-    'author': "ACSONE SA/NV, Dhinesh D, Odoo Community Association (OCA)",
+    'author': "ACSONE SA/NV, Dhinesh D, Jesse Morgan, Odoo Community Association (OCA)",
     'maintainer': 'Odoo Community Association (OCA)',
     'website': "http://acsone.eu",
 
--- a/auth_session_timeout/__openerp__.py
+++ b/auth_session_timeout/__openerp__.py
@@ -9,7 +9,8 @@
     'summary': """
         This module disable all inactive sessions since a given delay""",
 
-    'author': "ACSONE SA/NV, Dhinesh D, Jesse Morgan, Odoo Community Association (OCA)",
+    'author': "ACSONE SA/NV, Dhinesh D, Jesse Morgan, \
+Odoo Community Association (OCA)",
     'maintainer': 'Odoo Community Association (OCA)',
     'website': "http://acsone.eu",
 
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -1,10 +1,9 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
+# (c) 2015 ACSONE SA/NV, Dhinesh D
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from openerp import models, api, tools, SUPERUSER_ID
-
 
 DELAY_KEY = 'inactive_session_time_out_delay'
 IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
@@ -14,27 +13,25 @@ class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
     @tools.ormcache(skiparg=0)
+    def get_session_parameters(self, db):
+        """Method no longer required by module auth_session_timeout.
+        Left here in case any other modules referenced it.
+        """
+        ir_config_parameter_model = self.env['ir.config_parameter']
+        delay = int(ir_config_parameter_model.sudo().get_param(DELAY_KEY, 7200))
+        urls = ir_config_parameter_model.sudo().get_param(IGNORED_PATH_KEY, '').split(',')
+        return delay, urls
+
+    @tools.ormcache(skiparg=0)
     def _auth_timeout_get_parameter_delay(self, db):
-        param_model = self.pool['ir.config_parameter']
-        cr = self.pool.cursor()
-        delay = False
-        try:
-            delay = int(param_model.get_param(
-                cr, SUPERUSER_ID, DELAY_KEY, 7200))
-        finally:
-            cr.close()
+        ir_config_parameter_model = self.env['ir.config_parameter']
+        delay = int(ir_config_parameter_model.sudo().get_param(DELAY_KEY, 7200))
         return delay
 
     @tools.ormcache(skiparg=0)
-    def _auth_timeout_get_parameter_urls(self, db):
-        param_model = self.pool['ir.config_parameter']
-        cr = self.pool.cursor()
-        urls = []
-        try:
-            urls = param_model.get_param(
-                cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
-        finally:
-            cr.close()
+    def _auth_timeout_get_parameter_ignoredurls(self, db):
+        ir_config_parameter_model = self.env['ir.config_parameter']
+        urls = ir_config_parameter_model.sudo().get_param(IGNORED_PATH_KEY, '').split(',')
         return urls
 
     @api.multi
@@ -42,6 +39,8 @@ class IrConfigParameter(models.Model):
         res = super(IrConfigParameter, self).write(vals)
         if self.key == DELAY_KEY:
             self._auth_timeout_get_parameter_delay.clear_cache(self)
-        if self.key == IGNORED_PATH_KEY:
-            self._auth_timeout_get_parameter_urls.clear_cache(self)
+            self.get_session_parameters.clear_cache(self)
+        elif self.key == IGNORED_PATH_KEY:
+            self._auth_timeout_get_parameter_ignoredurls.clear_cache(self)
+            self.get_session_parameters.clear_cache(self)
         return res
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -39,9 +39,7 @@ class IrConfigParameter(models.Model):
     def write(self, vals, context=None):
         res = super(IrConfigParameter, self).write(vals)
         if self.key == DELAY_KEY:
-            #self._auth_timeout_get_parameter_delay.clear_cache(self)
             self.get_session_parameters.clear_cache(self)
         elif self.key == IGNORED_PATH_KEY:
-            #self._auth_timeout_get_parameter_ignoredurls.clear_cache(self)
             self.get_session_parameters.clear_cache(self)
         return res
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D
+# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
@@ -14,23 +14,34 @@ class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
     @tools.ormcache(skiparg=0)
-    def get_session_parameters(self, db):
+    def _auth_timeout_get_parameter_delay(self, db):
         param_model = self.pool['ir.config_parameter']
         cr = self.pool.cursor()
         delay = False
-        urls = []
         try:
             delay = int(param_model.get_param(
                 cr, SUPERUSER_ID, DELAY_KEY, 7200))
+        finally:
+            cr.close()
+        return delay
+
+    @tools.ormcache(skiparg=0)
+    def _auth_timeout_get_parameter_urls(self, db):
+        param_model = self.pool['ir.config_parameter']
+        cr = self.pool.cursor()
+        urls = []
+        try:
             urls = param_model.get_param(
                 cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
         finally:
             cr.close()
-        return delay, urls
+        return urls
 
     @api.multi
     def write(self, vals, context=None):
         res = super(IrConfigParameter, self).write(vals)
-        if self.key in [DELAY_KEY, IGNORED_PATH_KEY]:
-            self.get_session_parameters.clear_cache(self)
+        if self.key == DELAY_KEY:
+            self._auth_timeout_get_parameter_delay.clear_cache(self)
+        if self.key == IGNORED_PATH_KEY:
+            self._auth_timeout_get_parameter_urls.clear_cache(self)
         return res
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -3,7 +3,7 @@
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, api, tools
+from openerp import models, api, tools, SUPERUSER_ID
 
 DELAY_KEY = 'inactive_session_time_out_delay'
 IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
@@ -12,39 +12,36 @@ IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
 class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
+    @tools.ormcache('db')
     def get_session_parameters(self, db):
-        """Method no longer required by module auth_session_timeout.
-        Left here in case any other modules referenced it.
-        """
+        param_model = self.pool['ir.config_parameter']
+        cr = self.pool.cursor()
         delay = False
         urls = []
-        with openerp.registry(db).cursor() as cr:
-            uid, context = self.env.uid, self.env.context
-            with api.Envionment.manage():
-                env = api.Envionment(cr, uid, context)
-                delay = self.with_env(env)._auth_timeout_get_parameter_delay()
-                urls = self.with_env(env)._auth_timeout_get_parameter_urls()
+        try:
+            delay = int(param_model.get_param(
+                cr, SUPERUSER_ID, DELAY_KEY, 7200))
+            urls = param_model.get_param(
+                cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
+        finally:
+            cr.close()
         return delay, urls
 
-    @tools.ormcache()
     def _auth_timeout_get_parameter_delay(self):
-        param_model = self.env['ir.config_parameter']
-        delay = int(param_model.sudo().get_param(DELAY_KEY, 7200))
+        delay, urls = self.get_session_parameters(self.pool.db_name)
         return delay
 
-    @tools.ormcache()
     def _auth_timeout_get_parameter_ignoredurls(self):
-        param_model = self.env['ir.config_parameter']
-        urls = param_model.sudo().get_param(IGNORED_PATH_KEY, '').split(',')
+        delay, urls = self.get_session_parameters(self.pool.db_name)
         return urls
 
     @api.multi
     def write(self, vals, context=None):
         res = super(IrConfigParameter, self).write(vals)
         if self.key == DELAY_KEY:
-            self._auth_timeout_get_parameter_delay.clear_cache(self)
+            #self._auth_timeout_get_parameter_delay.clear_cache(self)
             self.get_session_parameters.clear_cache(self)
         elif self.key == IGNORED_PATH_KEY:
-            self._auth_timeout_get_parameter_ignoredurls.clear_cache(self)
+            #self._auth_timeout_get_parameter_ignoredurls.clear_cache(self)
             self.get_session_parameters.clear_cache(self)
         return res
--- a/auth_session_timeout/models/ir_config_parameter.py
+++ b/auth_session_timeout/models/ir_config_parameter.py
@@ -3,7 +3,7 @@
 
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import models, api, tools, SUPERUSER_ID
+from openerp import models, api, tools
 
 DELAY_KEY = 'inactive_session_time_out_delay'
 IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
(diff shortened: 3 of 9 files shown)


jmorgannz:
I'm new at this (contributing); guess I might need to read up on the CLA.
The reason I did this is that I needed to add the ability to exempt certain users from the expiry functionality.

I didn't want to re-implement the entire thing, but the OCA copy was not easily pluggable.
I did this, then made my own module that adds the exempt-by-user functionality.

I am using this in production.

auth_session_timeout/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015 ACSONE SA/NV, Dhinesh D
+# (c) 2015 ACSONE SA/NV, Dhinesh D, Jesse Morgan
```
lasley: Please do not append to others' copyrights. Also, you did not change this file at all, so your claim to copyright on any aspect of it would be invalid.

Regarding the copyright headers & what happened here - we were significantly more lax on what happened with these headers previously. If you look at the history of this file, you'll see that Dhinesh D's copyright is actually invalid here too & I would argue it is invalid for the entire module. That argument isn't something to make for history though, but it's most certainly something to enforce going forwards.

Typically if you're going to claim copyright on a file in the header, you need to have written the whole thing. Otherwise, Git line-by-line attribution is an implicit copyright on things you do actually change.

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -14,23 +14,34 @@ class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
     @tools.ormcache(skiparg=0)
-    def get_session_parameters(self, db):
+    def _auth_timeout_get_parameter_delay(self, db):
```
lasley: This is changing the public external facing interface of a module that is already in production, and therefore is against a stable release policy.

Is the rename of this method actually necessary? If so, we'll need to provide a `get_session_parameters` method to use as an adaptation layer for old code.

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -14,23 +14,34 @@ class IrConfigParameter(models.Model):
...
+        param_model = self.pool['ir.config_parameter']
+        cr = self.pool.cursor()
+        urls = []
+        try:
             urls = param_model.get_param(
                 cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
         finally:
             cr.close()
-        return delay, urls
+        return urls
```
lasley: This is changing the external facing API. The adapter method would need to return compatibly

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -14,23 +14,34 @@ class IrConfigParameter(models.Model):
...
 
     @api.multi
     def write(self, vals, context=None):
         res = super(IrConfigParameter, self).write(vals)
-        if self.key in [DELAY_KEY, IGNORED_PATH_KEY]:
-            self.get_session_parameters.clear_cache(self)
+        if self.key == DELAY_KEY:
+            self._auth_timeout_get_parameter_delay.clear_cache(self)
+        if self.key == IGNORED_PATH_KEY:
```
lasley: `elif`

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
 class ResUsers(models.Model):
     _inherit = 'res.users'
 
-    def _check_session_validity(self, db, uid, passwd):
+    def _auth_timeout_urls_get(self, session, db, uid):
```
lasley: It could be argued that you are also changing the interfaces here, but they are private so you're somewhat in the clear. I would still recommend an adapter method though- you don't want to be the reason that someone's code breaks in production.

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
...
+        # Pluggable method for calculating ignored urls
```
lasley: Check out [PEP-257](https://www.python.org/dev/peps/pep-0257/), which describes Pythonic docblocks. This is more akin to a C docblock

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
...
+        # Defaults to stored config param
+        param_obj = self.pool['ir.config_parameter']
```
lasley: `self.env` - Pool is old API. Please don't use

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
...
+        return param_obj._auth_timeout_get_parameter_urls(db)
+
+    def _auth_timeout_deadline_calculate(self, session, db, uid):
+        # Pluggable method for calculating timeout deadline
+        # Defaults to current time minus delay using delay stored as session param
+        param_obj = self.pool['ir.config_parameter']
```
lasley: `self.env`

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
...
+        deadline = self._auth_timeout_deadline_calculate(session, db, uid)
+
+        # Check if deadline expired
+        expired = False
+        if deadline != False:
+            path = root.session_store.get_session_filename(session.sid)
+            try:
+                expired = getmtime(path) < deadline
+            except OSError:
+                pass
```
lasley: [Never pass silently](https://www.python.org/dev/peps/pep-0020/). Please add a `_logger.debug` here at minimum

auth_session_timeout/models/res_users.py
```diff
@@ -17,28 +17,70 @@
...
+        urls = self._auth_timeout_urls_get(session, db, uid)
+
+        # Update session last modified and access times
+        if http.request.httprequest.path not in urls:
+            try:
                 utime(path, None)
-        except OSError:
-            pass
+            except OSError:
+                pass
```
lasley: I know it's not your code, but you're already here & changing things. Never pass silently - `_logger.debug`

lasley (changes requested):
Thanks for the contribution @jmorgannz!

Information on the CLA is [here](http://odoo-community.org/page/website.cla). Let us know if you have questions.

I added some comments inline after reviewing. Also, it would be nice to see some tests for your added logic - we're dropping in coverage here.

auth_session_timeout/__init__.py
jmorgannz: Thanks for the heads-up. 
I did in fact only change the copyright headers on the files I modified to begin with, but when I was 'polishing it up' to submit as a pull request I went back and added to the other files.

I guess the thought was that the enter module would be subject to copyright rather than each file within it.

No matter, I will change it :)

jmorgannz:
@lasley Thanks for the input. Will investigate changes required!

auth_session_timeout/models/ir_config_parameter.py
jmorgannz: I was trying to be granular here.
With the new structure in ``ResUser`` each param is required in different places, so I split `get_session_parameters` in two.

Alternatively I could have just called `get_session_parameters` in all the now required places, using the parameter returned that I need and discarding the other; however I was feeling the need to be frugal given that from what I can see these methods are running in a fairly hot code-path.

The other thing that occurs to me now is that this code now opens more cursors than before.
Technically this new code would open and close a cursor for each parameter read, whereas they were both read in a single cursor before.

Am I barking up the wrong tree here?
Should I be being less frugal and instead trust the ORM layer to mitigate any repeat work ?

auth_session_timeout/models/res_users.py
jmorgannz: d

auth_session_timeout/models/ir_config_parameter.py
lasley: I 100% agree with your design choice. The solution here is to basically just create a `get_session_parameters` method that returns the output of the two new methods in a compatible tuple

_EDIT: I'll comment inline about the cursors_

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -14,23 +14,34 @@ class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
     @tools.ormcache(skiparg=0)
-    def get_session_parameters(self, db):
+    def _auth_timeout_get_parameter_delay(self, db):
         param_model = self.pool['ir.config_parameter']
         cr = self.pool.cursor()
```
lasley: Sounds like you're going for a real fix, I ❤️  it and thank you! Let's do like this in both of the cursor areas:

```
with openerp.registry(self.env.cr.dbname).cursor() as cr:
    ...operations...
```

This will kill the try/finally.

Now, I do have a question regarding this & forgive if it's basic - I don't know this module. Why are we opening new cursors in these spots?

jmorgannz: I wasn't sure why this was done originally, so followed the pattern in case it had good reason.

This is all complete conjecture, but:

Seems to me that the `ResUser.check` method could quite likely be called on requests that are at a stage that have not had their proper environment/context set up yet as part of request buildup/teardown.

I.e. Odoo should check a users validation credentials before opening them an active db cursor, etc.

Since this module hangs off `ResUser.check`, we get caught in that scenario.

The only scenario we are actually interested in though is (re)checks - we don't need to check for session expiry on the request that is logging a user into the system (but we do)

It would not surprise me if the manual cursor control was added as a workaround by the original developer(s) as they quickly realized that without them users could no longer log in!

jmorgannz: I have taken the manual cursors out for now, and everything still *appears* to work correctly.

In order to use a manual cursor with the newer API, I will have to create a new env instance, from what I can see.

auth_session_timeout/models/res_users.py
jmorgannz: FYI, self.env is not defined in these methods.
Was able to use request.env, but I'd rather find out why self.env isn't working.

jmorgannz:
Man, minimum line width of 79 in conjunction with strict line-continuation indent rules presents an interesting challenge.

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -15,22 +14,33 @@ class IrConfigParameter(models.Model):
...
-            urls = param_model.get_param(
-                cr, SUPERUSER_ID, IGNORED_PATH_KEY, '').split(',')
-        finally:
-            cr.close()
+        """Method no longer required by module auth_session_timeout.
+        Left here in case any other modules referenced it.
+        """
+        param_model = self.env['ir.config_parameter']
+        delay = int(param_model.sudo().get_param(DELAY_KEY, 7200))
+        urls = param_model.sudo().get_param(IGNORED_PATH_KEY, '').split(',')
```
jmorgannz: Method no longer using the passed in db for database access

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -5,15 +5,14 @@
...
 
 
 class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
-    @tools.ormcache(skiparg=0)
+    @tools.ormcache('db')
     def get_session_parameters(self, db):
         param_model = self.pool['ir.config_parameter']
         cr = self.pool.cursor()
```
jmorgannz, line 18: `self.pool` usage: reasoning listed in separate comment

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -28,9 +27,19 @@ def get_session_parameters(self, db):
             cr.close()
         return delay, urls
 
+    def _auth_timeout_get_parameter_delay(self):
+        delay, urls = self.get_session_parameters(self.pool.db_name)
+        return delay
```
jmorgannz, line 32: `self.pool` usage: reasoning listed in separate comment

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -28,9 +27,19 @@ def get_session_parameters(self, db):
...
+
+    def _auth_timeout_get_parameter_ignoredurls(self):
+        delay, urls = self.get_session_parameters(self.pool.db_name)
+        return urls
```
jmorgannz, line 36: `self.pool` usage: reasoning listed in separate comment

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
 class ResUsers(models.Model):
     _inherit = 'res.users'
 
-    def _check_session_validity(self, db, uid, passwd):
-        if not request:
+    def _auth_timeout_ignoredurls_get(self):
+        """Pluggable method for calculating ignored urls
+        Defaults to stored config param
+        """
+        param_model = self.pool['ir.config_parameter']
```
jmorgannz, line 27: `self.pool` usage: reasoning listed in separate comment

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+        return param_model._auth_timeout_get_parameter_ignoredurls()
+
+    def _auth_timeout_deadline_calculate(self):
+        """Pluggable method for calculating timeout deadline
+        Defaults to current time minus delay using delay stored as config param
+        """
+        param_model = self.pool['ir.config_parameter']
```
jmorgannz, line 34: `self.pool` usage: reasoning listed in separate comment

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+        Return:
+            True: session terminated
+            False: session timeout cancelled
+        """
+        if session.db and session.uid:
+            session.logout(keep_db=True)
+        return True
+
+    def _auth_timeout_request_get(self):
+        return request
```
jmorgannz: See separate comment on tests/mocking

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+
+    def _auth_timeout_session_filename_get(self, sid):
+        return root.session_store.get_session_filename(sid)
```
jmorgannz: See separate comment on tests/mocking

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+
+    def _auth_timeout_utime(self, path, dates=None):
+        return utime(path, dates)
```
jmorgannz: See separate comment on tests/mocking

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+
+    def _auth_timeout_getmtime(self, path):
+        return getmtime(path)
```
jmorgannz: See separate comment on tests/mocking

auth_session_timeout/models/res_users.py
```diff
@@ -13,31 +14,110 @@
...
+        # Check if past deadline
+        expired = False
+        if deadline is not False:
+            path = self._auth_timeout_session_filename_get(session.sid)
+            try:
+                expired = self._auth_timeout_getmtime(path) < deadline
+            except OSError as e:
+                _logger.warning(
+                    'Exception reading session file modified time: %s'
+                    % e
```
jmorgannz: I am using `warning` for these log entries.
Any reason I should use `debug` instead?

jmorgannz:
@lasley I have iterated the code a couple of times, and have got test coverage up, all tests passing.

There are a couple of things to note:
(see comments below)

jmorgannz:
self.pool usage
==========

Basically, if I try to use `self.env`, bad things happen to the entire Odoo session.

This code is invoked early in the pipeline - so early that `env.uid = None`
I believe Odoo's request pipeline would set up the proper `env` Environment later in the pipeline, after all this code has completed.

This seems like an easy case to work around. Just create a new `Environment` and use `SUPERUSER_ID` as the uid.
However if I try to do this, bad things happen to the user's session.
I get funny permission errors all throughout Odoo. Discussion would not load due to permissions (no permission for res_partner), and Settings would not load for other permission reasons.

I think what is happening is that the `env` that is available at this early stage is just a stand-in, to allow code like this to run. It is later replaced with the 'real' `env`, which is appended to each model as model.env.

By doing anything with `env` at this early stage, it seems to prevent Odoo from replacing the 'early stage env' with the proper one - leading to the permission problems.

The only fix I could find was to go back to using `self.pool` - but there must be some better alternative to that, without using `env`?

My first thought was to use the `Registry` - but again I don't know how I would get access to it at this stage in the request pipeline.

jmorgannz:
Tests / Mocking
==========

I have added some extra tests to improve test coverage.
I ran into an issue with making tests run, as this code is dependent on the `request` object - which doesn't exist in unit tests.

The only way I could find to work around this, was to remove all references to the `request` object, and replace them with a function call which returns it instead.

This then allowed me to `_patch_method` the method and return a mock `request` object instead.

The same is true for simulating exceptions in the `utime` and `getmtime` calls.
I couldn't find a way to mock these as they are base functions of Python. The only way I could get it to work was to do the same as I did for `request`

This creates several small methods in `ResUser` which are seemingly redundant - only there for the unit tests to be able to run.

Would love to know if there is a better way to deal with this.

MiquelRForgeFlow:
@lasley, you can already remove the "CLA Pending" label 😉

lasley:
@mreficent - Do we have a CLA here? Why didn't the bot say something?

MiquelRForgeFlow:
@lasley, I added the CLA to @jmorgannz just today 😄

lasley:
@mreficent - Weird, CLA bot usually instantaneously responds to the PR. Regardless, it all looks in order - woot!

lasley:
@jmorgannz - You're awesome; this looks great! 

Regarding the pool usage - ok cool, this is totally a good reason to use it. Can you please add some sort of notation in the code indicating the results of your investigation? I think maybe a class level docblock would be a good spot. Let's save future peoples some time during migrations and such 😉 

Regarding the tests - yeah I've run against the same wall and ended up with a similar solution.

Regarding built in - you can technically mock them. I've done it in special cases, but it's one of those situations where you should certainly take a step back and re-evaluate if you're doing thigns correctly. Take a look at this though, which should clue in on how to mock them:

``` python
>>> dir()
['__builtins__', '__doc__', '__name__', '__package__']
>>> dir(__builtins__)
['ArithmeticError', 'AssertionError', 'AttributeError', 'BaseException', 'BufferError', 'BytesWarning', 'DeprecationWarning', 'EOFError', 'Ellipsis', 'EnvironmentError', 'Exception', 'False', 'FloatingPointError', 'FutureWarning', 'GeneratorExit', 'IOError', 'ImportError', 'ImportWarning', 'IndentationError', 'IndexError', 'KeyError', 'KeyboardInterrupt', 'LookupError', 'MemoryError', 'NameError', 'None', 'NotImplemented', 'NotImplementedError', 'OSError', 'OverflowError', 'PendingDeprecationWarning', 'ReferenceError', 'RuntimeError', 'RuntimeWarning', 'StandardError', 'StopIteration', 'SyntaxError', 'SyntaxWarning', 'SystemError', 'SystemExit', 'TabError', 'True', 'TypeError', 'UnboundLocalError', 'UnicodeDecodeError', 'UnicodeEncodeError', 'UnicodeError', 'UnicodeTranslateError', 'UnicodeWarning', 'UserWarning', 'ValueError', 'Warning', 'ZeroDivisionError', '_', '__debug__', '__doc__', '__import__', '__name__', '__package__', 'abs', 'all', 'any', 'apply', 'basestring', 'bin', 'bool', 'buffer', 'bytearray', 'bytes', 'callable', 'chr', 'classmethod', 'cmp', 'coerce', 'compile', 'complex', 'copyright', 'credits', 'delattr', 'dict', 'dir', 'divmod', 'enumerate', 'eval', 'execfile', 'exit', 'file', 'filter', 'float', 'format', 'frozenset', 'getattr', 'globals', 'hasattr', 'hash', 'help', 'hex', 'id', 'input', 'int', 'intern', 'isinstance', 'issubclass', 'iter', 'len', 'license', 'list', 'locals', 'long', 'map', 'max', 'memoryview', 'min', 'next', 'object', 'oct', 'open', 'ord', 'pow', 'print', 'property', 'quit', 'range', 'raw_input', 'reduce', 'reload', 'repr', 'reversed', 'round', 'set', 'setattr', 'slice', 'sorted', 'staticmethod', 'str', 'sum', 'super', 'tuple', 'type', 'unichr', 'unicode', 'vars', 'xrange', 'zip']
>>> __builtins__.sorted
<built-in function sorted>
```

However in the case of `utime` and `getmtime`, aren't these library calls? `os` I think?

IMO you have performed significant changes to this module - I believe you should put yourself (or your company) as a co-author in the manifest. That's obviously up to you, but I see a lot of your code here.

lasley (approved):

Garamotte:
For the `request` attribute in tests, you can use Werkzeug's test client, as I did in the `oauth_provider` module.
I just had to mock the env of the request, not the full request object.

Base class to initialize the test client: https://github.com/OCA/server-tools/blob/9.0/oauth_provider/tests/common_test_controller.py
Exemple usage in test classes: https://github.com/OCA/server-tools/blob/9.0/oauth_provider/tests/test_oauth_provider_controller_web_application.py

lasley:
Oh whoah that's awesome @sylvain-garancher - I love it! How the hell did I miss this in the OAuth module?!

jmorgannz:
@lasley 
Appreciate the help and attitude, thanks!

I had the same thought with regard to leaving some breadcrumbs for the next guy (or girl)

As for the utime and getmtime, yep their in the `os` library.
I tried using mock to patch them, but it didn't appear to have any effect.
Likely I was doing something wrong - have only been using Python since March.
Will give it another go, I'd rather not have those silly little methods I included hanging around.

I also was going to squash my commits - had no intent of those 'WIP' commit messages staying.
Not sure what effect doing that will have at this stage. I guess it would mangle the PR a bit as I imagine all the review comments to date are pinned to specific git refs that will be goneburger.
Will ask uncle Google.

lasley:
> I tried using mock to patch them, but it didn't appear to have any effect.

You were probably mocking the wrong import. Happened to me all the time when I was learning. In the case of this, the mock would look a little something like:

```
@mock.patch('odoo.addons.auth_session_timeout.models.res_users.os')
```

Keep in mind that you have to patch the imported object, not the object itself. My go-to on thinking is imagine that you were trying to import that object specifically from the file, as if it were declared there. Then, instead of importing, mock.

> I also was going to squash my commits - had no intent of those 'WIP' commit messages staying.
Not sure what effect doing that will have at this stage

Squashing is amazing, and we highly recommend it. You're right about the mangling though. Typically what I do is squash until someone reviews, then work in a new commit until a new review, rinse, repeat. Someone re-reviewing your code will typically review changesets, so you definitely don't want to alter an already reviewed timeline & require a whole code review.

Alternatively, you can just leave them all and whoever merges this will just use the nifty Squash and Merge that Github gives us:

image

Downside of that is all your WIP messages still make it in the actual commit note. I tend to prefer squashing myself due to this.

jmorgannz:
@sylvain-garancher 
Awesome, thanks mate!

jmorgannz:
*facepalm*
I better get my linters sorted out

auth_session_timeout/__openerp__.py
```diff
@@ -9,7 +9,8 @@
     'summary': """
         This module disable all inactive sessions since a given delay""",
 
-    'author': "ACSONE SA/NV, Dhinesh D, Odoo Community Association (OCA)",
+    'author': "ACSONE SA/NV, Dhinesh D, Jesse Morgan, \
+Odoo Community Association (OCA)",
```
jmorgannz, line 13: Used an escaped newline to fix issue with new author string length.
Mangles the file layout somewhat.
Wasn't sure if un-escaped newlines using `"""` would do funny things in the author string.

Garamotte: If you don't like escaped new lines, you can also use multiple strings, like this:
```
    'author': 'ACSONE SA/NV, Dhinesh D, Jesse Morgan, '
              'Odoo Community Association (OCA)',
```

lasley: This is my preferred, although IMO it looks nicer one per line:

```
    'author': 'ACSONE SA/NV, '
              'Dhinesh D, '
              'Jesse Morgan, '
              'Odoo Community Association (OCA)',
```

lasley:
@jmorgannz - take a look [here](https://github.com/OCA/pylint-odoo#usage) for local lint

jmorgannz:
@lasley  - thanks - I just hadn't got around to it. I've done it now.
That was mainly because I had adjusted all my editor apps to have a ruler at 79 characters, but forgot to adjust pycharm.

jmorgannz:
Hi all,
Have been indisposed, but still plan to add some code comments to this PR to finish it off - it's otherwise complete.

auth_session_timeout/models/ir_config_parameter.py
```diff
@@ -5,15 +5,14 @@
...
 DELAY_KEY = 'inactive_session_time_out_delay'
 IGNORED_PATH_KEY = 'inactive_session_time_out_ignored_url'
 
 
 class IrConfigParameter(models.Model):
     _inherit = 'ir.config_parameter'
 
-    @tools.ormcache(skiparg=0)
+    @tools.ormcache('db')
     def get_session_parameters(self, db):
```
chienandalu, line 16: I followed @yajo's advice here (https://github.com/OCA/server-tools/pull/875#discussion_r124193147) and got rid of this whole file in favour of new api and method improvements (wich seem to be available in v9 as well: https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_config_parameter.py#L57).

chienandalu (commented):
Thanks @jmorgannz for the good job. I borrowed some ideas for v10 migration indeed :slightly_smiling_face: Some comments below:

auth_session_timeout/models/ir_config_parameter.py
jmorgannz: @chienandalu 
Glad you found my changes useful :)
Yes I struggled with the same thoughts myself - at first I couldn't see any reason why changes were required at all in ir_config_parameter.
The best I could come up with is that the original writer wanted to read both and cache once, rather than read twice and cache twice.
In any event it is indeed redundant, although I was advised against removing get_session_parameters (deceptively named for a public method :/) as its in the wild now and might be referenced by some unknown modules.

lasley:
Ok before we continue with the v10, let's knock this one. @chienandalu - have your comments been attended to here?

chienandalu:
@lasley Yes, sure :slightly_smiling_face:

chienandalu (approved):

lasley:
Thanks @jmorgannz - congrats on your first OCA submission🥇 ! 

Look forward to seeing you around 🚀

MiquelRForgeFlow (commented):
Line https://github.com/OCA/server-tools/pull/887/files#diff-42f066551d06ff9ed8b4c8c7fdc6c8f1R18 should have been changed to 9.0.1.0.1 or 9.0.1.1.0 or 9.0.2.0.0.

jmorgannz:
@lasley 
Thanks Dave.
I was supposed to add some comments to this pull, but for various reasons have been a bit distracted.
This will not impact on the function of the pull, however.

Thanks for the help!

PR #889, opened 2017-07-06
[10.0][MIG] oauth_provider

lasley:
This is a complete overhaul of the OAuth2 provider module, as well as a migration to v10. I added an abstract REST API as well, following are the docs:

REST API
========

This module also includes a basic REST-style CRUD interface backed by OAuth2 authentication.

List
----

Return allowed information from all records, with optional query.

Route: /api/rest/1.0/&lt;string:model&gt;

Accepts: GET

Args:  
access\_token (str): OAuth2 access token to utilize for the  
operation.

model (str): Name of model to operate on. domain (array, optional): Domain to apply to the search, in the standard Odoo format. This will be appended to the scope’s pre-existing filter.

Returns:  
array of objs: List of data mappings matching query.

Example:

Bash:

``` sourceCode
curl -H "Content-Type: application/json" \
   -X GET \
   http://localhost:8060/api/gogo/1.0/res.partner?access_token=2J391BkHipPmM9nXv2BF6V07fehWtM&domain=%5B%5B%22company_id%22%2C%20%22%3D%22%2C%201%5D%5D

{"jsonrpc": "2.0", "id": 5602, "result": [{"name": "Test Partner", "id": 5602}]}
```

Python:

``` sourceCode
json_data = {
    'access_token': token['access_token'],
    'domain': [('company_id', '=', 1)],
}
response = requests.get(
    'http://localhost:8060/api/gogo/1.0/res.partner/',
    params=json_data,
)
response_data = response.json()

{u'jsonrpc': u'2.0', u'id': 5602, u'result': [{u'name': u'Test Partner', u'id': 5602}]}
```

Read
----

Return allowed information from specific records.

Route: /api/rest/1.0/&lt;string:model&gt;/&lt;int:record\_id&gt;

Accepts: GET

Args:  
access\_token (str): OAuth2 access token to utilize for the  
operation.

model (str): Name of model to operate on. record\_id (int): ID of record to get.

Returns:  
obj: Record data.

Example:

Bash:

``` sourceCode
curl -H "Content-Type: application/json" \
   -X GET \
   http://localhost:8060/api/gogo/1.0/res.partner/5602?access_token=2J391BkHipPmM9nXv2BF6V07fehWtM

{"jsonrpc": "2.0", "id": 5602, "result": [{"name": "Test Partner", "id": 5602}]}
```

Python:

``` sourceCode
json_data = {
    'access_token': token['access_token'],
}
response = requests.post(
    'http://localhost:8060/api/gogo/1.0/res.partner/5602',
   params=json_data,
)
response_data = response.json()

{u'jsonrpc': u'2.0', u'id': 5602, u'result': [{u'name': u'Test Partner', u'id': 5602}]}
```

Create
------

Return allowed information from specific records.

Route: /api/rest/1.0/&lt;string:model&gt;

Accepts: POST

Args:  
access\_token (str): OAuth2 access token to utilize for the  
operation.

model (str): Name of model to operate on. kwargs (mixed): All other named arguments are used as the data for record mutation.

Returns:  
obj: Record data.

Example:

Bash:

``` sourceCode
curl -H "Content-Type: application/json" \
   -X POST \
   -d '{ "params": { "access_token": "2J391BkHipPmM9nXv2BF6V07fehWtM", "name": "Test Partner" }' \
   http://localhost:8060/api/gogo/1.0/res.partner

{"jsonrpc": "2.0", "id": 5602, "result": [{"name": "T
```

### DIFF
--- a/oauth_provider/README.rst
+++ b/oauth_provider/README.rst
@@ -1,0 +1,142 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+OAuth Provider
+==============
+
+This module allows you to turn Odoo into an OAuth 2 provider.
+
+It's meant to provide the basic authentication feature, and some data access routes.
+But you are encouraged to create custom routes, in other modules, to give structured data for any specific need.
+
+Installation
+============
+
+To install this module, you need to:
+
+#. Install the oauthlib python module
+#. Install the module like any other in Odoo
+#. For the token retrieval to work on a multi-database instance, you should add this module in the server_wide_modules list
+
+Configuration
+=============
+
+This module requires you to configure two things :
+
+#. The scopes are used to define restricted data access
+#. The clients are used to declare applications that will be allowed to request tokens and data
+
+To configure scopes, you need to:
+
+#. Go to Settings > Users > OAuth Provider Scopes
+#. Create some scopes:
+
+ - The scope name and description will be displayed to the user on the authorization page.
+ - The code is the value provided by the OAuth clients to request access to the scope.
+ - The model defines which model the scope is linked to (access to user data, partners, sales orders, etc.).
+ - The filter allows you to determine which records will be accessible through this scope. No filter means all records of the model are accessible.
+ - The field names allows you to define which fields will be provided to the clients. An empty list only returns the id of accessible records.
+
+To configure clients, you need to:
+
+#. Go to Settings > Users > OAuth Provider Clients
+#. Create at least one client:
+
+ - The name will be displayed to the user on the authorization page.
+ - The client identifier is the value provided by the OAuth clients to request authorizations/tokens.
+ - The application type adapts the process to four pre-defined profiles:
+
+   - Web Application : Authorization Code Grant
+   - Mobile Application : Implicit Grant
+   - Legacy Application : Resource Owner Password Credentials Grant
+   - Backend Application : User Credentials Grant (not implemented yet)
+
+ - The skip authorization checkbox allows the client to skip the authorization page, and directly deliver a token without prompting the user (useful when the application is trusted).
+ - The allowed scopes list defines which data will be accessible by this client applicaton.
+ - The allowed redirect URIs must match the URI sent by the client, to avoid redirecting users to an unauthorized service. The first value in the list is the default redirect URI.
+
+For example, to configure an Odoo's *auth_oauth* module compatible client, you will enter these values :
+
+- Name : Anything you want
+- Client identifier : The identifier you want to give to this client
+- Application Type : Mobile Application (Odoo uses the implicit grant mode, which corresponds to the mobile application profile)
+- Allowed Scopes : Nothing required, but allowing access to current user's email and name is used by Odoo to fill user's information on signup
+- Allowed Redirect URIs : http://odoo.example.com/auth_oauth/signin
+
+Usage
+=====
+
+This module will allow OAuth clients to use your Odoo instance as an OAuth provider.
+
+Once configured, you must give these information to your client application :
+
+#. Client identifier : Identifies the application (to be able to check allowed scopes and redirect URIs)
+#. Allowed scopes : The codes of scopes allowed for this client
+#. URLs for the requests :
+
+  - Authorization request : http://odoo.example.com/oauth2/authorize
+  - Token request : http://odoo.example.com/oauth2/token
+  - Token information request : http://odoo.example.com/oauth2/tokeninfo
+     Parameters : access_token
+  - User information request : http://odoo.example.com/oauth2/userinfo
+     Parameters : access_token
+  - Any other model information request (depending on the scopes) : http://odoo.example.com/oauth2/otherinfo
+     Parameters : access_token and model
+
+For example, to configure the *auth_oauth* Odoo module as a client, you will enter these values :
+
+- Provider name : Anything you want
+- Client ID : The identifier of the client configured in your Odoo Provider instance
+- Body : Text displayed on Odoo's login page link
+- Authentication URL : http://odoo.example.com/oauth2/authorize
+- Scope : A space separated list of scope codes allowed to the client in your Odoo Provider instance
+- Validation URL : http://odoo.example.com/oauth2/tokeninfo
+- Data URL : http://odoo.example.com/oauth2/userinfo
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+Known issues / Roadmap
+======================
+
+* Implement the backend application profile (client credentials grant type)
+* Add checkboxes on the authorization page to allow the user to disable some scopes for a token ? (I don't know if this is allowed in the OAuth protocol)
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Sylvain Garancher <sylvain.garancher@syleam.fr>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/oauth_provider/README.rst
+++ b/oauth_provider/README.rst
@@ -1,6 +1,6 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
-   :alt: License: AGPL-3
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
 
 ==============
 OAuth Provider
@@ -82,7 +82,7 @@ Once configured, you must give these information to your client application :
      Parameters : access_token
   - User information request : http://odoo.example.com/oauth2/userinfo
      Parameters : access_token
-  - Any other model information request (depending on the scopes) : http://odoo.example.com/oauth2/otherinfo
+  - Any other model information request (depending on the scopes) : http://odoo.example.com/oauth2/data
      Parameters : access_token and model
 
 For example, to configure the *auth_oauth* Odoo module as a client, you will enter these values :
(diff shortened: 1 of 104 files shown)


lasley:
@sylvain-garancher - I'm going to change some things (detailed below), which will give us a full featured CRUD API. I was wondering, however, if we could relicense this to LGPL? 

The reasoning for this is that I need to design an opinionated API based on some proprietary vendor data structures & I will not be able to release the final product as Open Source due to contractual obligations. I'd love to use this module as a base, and have planned a nice Open Source REST API module to sit on top of this one, that my proprietary one will sit on top of. I'm between a rock and a hard place with this vendor though - they're pretty tight about their IP.

Following are my plans for this module in order to support my requirements:

* I am planning on shifting the public facing API in the following ways:
  * Change `/oauth2/otherinfo` route to `/oauth2/data`
  * Add POST, PUT, and DELETE functionality to `/oauth2/data`, allowing for full CRUD API
  * Add a domain argument to the GET path, allowing for queries
* Add `create`, `read`, `update`, `delete` Boolean fields to `OAuthProviderScope`
* Break `OauthProviderScope.get_data_for_model` into a few helper methods. Public facing operation would be identical
* Add `create_record`, `write_record`, `delete_record` to OauthProviderScope, with passthrough methods in OauthProviderToken (similar to the `get_data_for_model` architecture)
   * The write and delete methods will validate scope by first trying to get the record. Validate found record fields returned against the fields intended to be modified
* Change the name of `get_data_for_model` to `get_records` to match the new method naming a bit better


Some interesting design considerations that I'm not too sure of yet:
* Should we validate scope against the written record? IE if the record does not match the scope after write (or create), should the transaction be rolled back
* I wonder what it would take to implement the client credentials grant type

Garamotte:
Hello,

I'm not against a licence change, if it allows the module to be improved instead of writing a new one to replace it. Maybe the OCA has to give an opinion, as the module is in OCA's repository ? @pedrobaeza @gurneyalex 

The current version of the module only includes what my customer needed : login from a mobile application and retrieving some information about a subscription. Providing a full CRUD API is a great improvement idea :D
For the URL change, maybe you could add a new one, instead of a simple replacement, and make the old one deprecated (with a warning somewhere ?), if this is easy to do.
I agree on the method name change from `get_data_for_model` to `get_records` and split it in several small helper functions, if it can help in terms of modularity :)

About the create/read/write/delete fields, I wonder if it can be managed using `ir.rule` instead of modifying the model ?

For the scope validation, would it be possible to add a selection field to let the user define if it has to be valid after/before/both ?
I don't think the client credentials grant type needs much work to be implemented, the hardest part should be to write the tests. A big part already exist (even tests), as there are many common parts with other grant types.

lasley:
> For the URL change, maybe you could add a new one, instead of a simple replacement, and make the old one deprecated (with a warning somewhere ?), if this is easy to do.

Easy enough, yeah.

> About the create/read/write/delete fields, I wonder if it can be managed using ir.rule instead of modifying the model ?

Technically `ir.rule`s should already be covering these operations, due to the `.sudo(user=oauth_provider_token.user_id)`. I figured that the extra granularity would be appreciated, but in hindsight I think you're right that we should just leave it at those. Otherwise the two layers could just cause confusion.

> For the scope validation, would it be possible to add a selection field to let the user define if it has to be valid after/before/both ?

Good idea - I'll add that the to `oauth.provider.scope` model

> I don't think the client credentials grant type needs much work to be implemented, the hardest part should be to write the tests. A big part already exist (even tests), as there are many common parts with other grant types.

Sweet - I thought it looked like a lot of it was already hooked up, glad you confirmed. It's not totally required for my project, but I would prefer to deploy with those vs. legacy tokens.

pedrobaeza:
No problem in the license change if all the code authors agree and the target license is OSI.

lasley:
Alright I've now made a dependent API & I would say that the structure of this is now finalized if anyone wants to take a look. I'll just be working on tests now 🚀

oauth_provider/README.rst
```diff
@@ -0,0 +1,147 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+Known issues / Roadmap
+======================
+
+* Implement the backend application profile (client credentials grant type)
+* Add checkboxes on the authorization page to allow the user to disable some scopes for a token ? (I don't know if this is allowed in the OAuth protocol)
```
lasley, line 389: @sylvain-garancher - can you elaborate on this part? Where would the user be selecting this, in their account? Or during the authorization phase?

Garamotte: I thought about checkboxes on the authorization page. This would add only selected scopes to the generated token, instead of the current "all or nothing" authorization page.

Like selecting only some permissions when installing an app on an iOS/Android device, instead of allowing any permission asked by the app.

Garamotte: With all scopes checked by default, to keep the same default behaviour.

lasley:
Added docs for the new REST API & copied into the PR comment

oauth_provider/README.rst
lasley: Oh cool yes, this is totally allowed in OAuth spec. I think though that this is actually implemented client side. Consider the following example:

```
   import os
   import requests

   from oauthlib.oauth2 import LegacyApplicationClient
   from requests_oauthlib import OAuth2Session

   # Allows for no HTTPS verification. DO NOT USE IN PRODUCTION!
   os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'

   client_id = '2ec860db-ce81-49c2-b4b7-c71bcbee481f'
   client_secret = 'Test'
   username = 'admin'
   password = 'admin'

   oauth = OAuth2Session(client=LegacyApplicationClient(client_id=client_id))
   token = oauth.fetch_token(
       token_url='http://localhost:8060/oauth2/token',
       username=username,
       password=password,
       client_id=client_id,
       client_secret=client_secret,
       scope=['public-pricelist', 'pricelist-item', 'medical-medicament'],
   )
```

Above, the client requests the scopes `['public-pricelist', 'pricelist-item', 'medical-medicament']`. It is actually authorized for quite a bit more though, but that authorization is not required in the context of this specific integration. 

In the case of a mobile or web application, there would be a checkbox in the application itself disallowing certain scopes. The disallowance would have to be subsequently handled in the application itself as well, because it would be dependent on the data that it will no longer be receiving due to the security restriction.

Given the above, I think this isn't necessary in the roadmap. Do you agree?

Garamotte: I added this line in the roadmap because this has to be managed in the module's authorization page, and I didn't took time to do that. In short:

- Add a checkbox here: https://github.com/OCA/server-tools/blob/9.0/oauth_provider/templates/authorization.xml#L19
- Take the checkboxes values in account here: https://github.com/OCA/server-tools/blob/9.0/oauth_provider/controllers/main.py#L142

lasley:
Note there is a timeout in Travis when this build is included as part of the main build, so there is some incompatibility between this module and something else. No idea how to identify it though, so I isolated the build.

oauth_provider/tests/test_rest_api_controller.py
```diff
@@ -0,0 +1,200 @@
...
+        super(TestRestApiController, self).setUp('legacy application')
+        self.logged_user = self.env.user
+        self.user = self.env.user
+        self.token = self.new_token().token
+        self.filter.domain = []
+        self.admin_user_data = {"city": "Scranton",
+                                "login": "admin",
+                                "id": self.env.uid,
+                                "name": "Administrator",
+                                "email": "admin@yourcompany.example.com"}
```
Garamotte: I prefer to use data from variables, like `self.user.email`, instead a fixed value, to avoid errors if demo data are modified.

lasley: That's what I'm doing here for the most part, although I think I forgot to reference a few times in the tests. I'll double check

lasley: Ohhhh I see what you mean now that I looked at your other tests. Good call!

oauth_provider/tests/common_test_oauth_provider_controller.py
```diff
@@ -0,0 +1,420 @@
...
+        self.login('demo', 'demo')
+        response = self.test_client.get(
+            '/oauth2/authorize', query_string=query_string,
+            environ_base=self.werkzeug_environ)
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(self.client.name in response.data)
+
+
+class TestOauthProviderRefreshTokenController(object):
+    def test_refresh_token_error_too_much_scopes(self):
```
lasley, line 133: @sylvain-garancher - this one is failing, and I'm actually kind of confused on the purpose of it. What is invalid about this request? These are valid scopes in the client, and multiple scopes should be supported at once. 

I feel like I'm missing something, particularly because there is a nearly identical request below in `test_refresh_token`, except that the codes are joined by whitespace

Garamotte: That's because you added `'scope_ids': [(6, 0, self.client.scope_ids.ids)],` in the `new_token` method.
The purpose of this test is: Check that trying to generate a new token by refrehsing an old token should not allow to add new scopes.

In the 9.0 branch, the `new_token` method doesn't add any scope to the generated token, so putting at least one scope in the refresh token request is wrong. In this PR, you added all scopes in the generated token, so this test fails (all scopes are already allowed).

Garamotte: But you are also true about the whitespace join... this should be the case in this test too (the scope argument should be a string, not a list).

lasley: Ahhh excellent call, thank you! Such a seemingly minor change 😆

lasley: Hrmmm so this still isn't working, even without the `scope_ids` in `new_records`.

I searched through v9 & I don't see the term "invalid_scope" anywhere other than tests. Do you perhaps remember where this validation logic is taking place?

oauth_provider/README.rst
```diff
@@ -0,0 +1,421 @@
...
+Returns:
+   array of objs: List of data mappings matching query.
+
+Example:
+
+Bash:
+
+.. code-block:: bash
+
+   curl -H "Content-Type: application/json" \
```
lasley, line 157: Add bearer token to the headers:
```
httpWebRequest.Headers.Add("Authorization", "Bearer " + provider.AccessToken);
```

oauth_provider/tests/common_test_oauth_provider_controller.py
Garamotte: The `invalid_scope` term comes from te RFC of the OAuth protocol.
This validation is done by the oauthlib python module (controlled by the `validator.py` file here).

oauth_provider/README.rst
lasley: NVM this was just someone that couldn't perform the OAuth dance correctly. Bearer token is not required past the OAuth authentication & token exchange phases.

That said, I think we're breaking RFC here (from [section 7](https://tools.ietf.org/html/rfc6749#section-7)):

>    The method in which the client utilizes the access token to
>   authenticate with the resource server depends on the type of access
>   token issued by the authorization server.  Typically, it involves
>   using the HTTP "Authorization" request header field [RFC2617] with an
>   authentication scheme defined by the specification of the access
>   token type used, such as [RFC6750].

So I think instead we need to scrape the token from the request headers for use, instead of requiring as call params. Thoughts?

lasley:
This is ready for review. Note that I had to leave the build split from the rest of the modules because there is some unknown incompatibility with another module that causes the build to time out. No idea how to find that, and it's not possible to functionally test this on runbot anyways due to needing only one db

Garamotte:
@lasley Could you test if the timeout issue on Travis could be solved by using this in the `OauthProviderControllerTransactionCase` class instead of excluding a module from the build?
https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L246-L248
https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L263-L265

cf https://github.com/OCA/report-print-send/pull/95

lasley:
Damn no dice unfortunately @sylvain-garancher. The only necessary change is switching to `HttpCase` right?

Garamotte:
Yes, but looking at the log on Travis, this is not the same issue.

The issue fixed by using `HttpCase` is a lock (on the registry) being waiting for the modules update to finish when loading an HTTP thread (by runing wkhtmltopdf during tests, for example). The `HttpCase` class replaces the lock by a dummy one to avoid blocking the HTTP threads loading.

The issue here may be due to the same lock, but is encountered when running the modules update, not the tests. The blocked module can be either `dbfilter_from_header` or the next module to be updated which has not yet been printed into the log (I don't know if the log line is printed before of after the update of the module). Comparing the logs of other builds, if the next module is not `oauth_provider` (which was absent), it could be `server_environment`, which seems to be installed just after `dbfilter_from_header` in other builds.

If I'm right, in the first case (`oauth_provider`'s update blocked), I think that `dbfilter_from_header` is in conflict with `oauth_provider`, in the other case, it could be `server_environment`.
I'll try to run some tests on my free time, you can remove the commit in the meantime. Thanks for testing this ;)

Garamotte:
OK, I reproduced the issue localy on my dev box.
The blocking appears during installation of the `oauth_provider` module.

The issue is a conflict between the pre-init hook of `oauth_provider`, used to initialize OAuth identifiers on users, and the `user_threshold`'s `__init__` method of `res.users`, which opens another database cursor, and thus is blocked because the *main* transaction has uncommited changes.
If we add a `cr.commit()` at the end of the pre-init hook of `oauth_provider`, everything works, but I'd prefer not to add *random* commits...

Do you know why the `user_threshold` module needs a separate cursor ?
Could we use a savepoint instead of a cursor ?

lasley:
> Do you know why the user_threshold module needs a separate cursor ?
Could we use a savepoint instead of a cursor ?

I honestly don't even think that is necessary. We'll find out - #955

lasley:
Good call @sylvain-garancher, looks like we have a 🍏  when based on #955.

I tested too and it seems the logic in that module works while operating in the same cursor, so I'm not sure what the commit was about TBH

lasley:
I'm curious - how did you figure out which module combination was required to replicate this?

Garamotte:
First, I reproduced the issue on my system (basically by copy/pasting TravisCI's commands).

Next step, while Odoo was blocked, I tried a `kill -3`, which didn't work (the process was fully blocked), so I checked running SQL queries using `SELECT * FROM pg_stat_activity WHERE datname = 'openerp_test'`, and found two queries (one idle, and the other blocked).

Then, the hardest part should have been to find the module from the SQL query, but I had the luck to find the `threshold` keyword in the query :)

Finally, knowing the line in the python code which was blocked, I easily found what was the real issue.

lasley:
Thanks for the breakdown @sylvain-garancher - this will help next time I run into something similar. I'm a bit embarrassed how horrible my strategy was for this TBH

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,343 @@
...
+
+        Returns:
+            dict:
+                This will be a dictionary of scoped record data, keyed by
+                record ID.
+                If `record_ids` is defined, and only one ID, the inner data
+                dictionary will be returned without nesting.
+        """
+
+        if isinstance(record_ids, int):
```
Garamotte: I'd prefer to avoid variables with several allowed types for an argument when possible.
Removing the if will require to change some calls to this method (basically replace `record.id` by `record.ids` or `record_id` by `[record_id]`), but I think this is better.

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,343 @@
...
+
+        Returns:
+            list of tuples: Domain of the scope, in standard Odoo format.
+        """
+
+        self.ensure_one()
+
+        if record_ids is None:
+            record_ids = []
+        elif isinstance(record_ids, int):
```
Garamotte: Same remark as the `get_data` method about several allowed types for an argument.

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,343 @@
...
+        """ Validate that the recordset is within the current scope.
+
+        Args:
+            records (BaseModel): Recordset to validate.
+
+        Returns:
+            bool: Indicating whether the records are within scope.
+        """
+
+        scoped_records = self.env[records._name].browse()
```
Garamotte: I didn't know calling `browse()` without any argument worked.
You can simply use `scoped_records = self.env['records._name']` to get an empty recordset.

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,343 @@
...
+        for scope in self:
+            scoped_records += scope._get_scoped_records(
```
Garamotte: Prefer using the `|=` operator, as it removes duplicates. The `+=` operator is not really helpful with Odoo's recordsets...
```
>>> partners = self.env['res.partner']
>>> partners += self.env['res.partner'].browse([1, 1])
>>> partners
res.partner(1, 1)
>>> partners |= self.env['res.partner'].browse([1, 1, 2])
>>> partners
res.partner(1, 2)
```

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,343 @@
...
+        Args:
+            model (BaseModel): Name of the model to operate on.
+            vals (dict): Values that should be checked against the current
+                scope, keyed by field name.
+
+        Returns:
+            bool: Whether the values are within the scope.
+        """
+        scopes = self.filter_by_model(model)
+        field_names = scopes.mapped('field_ids').mapped('name')
```
Garamotte: You can call mapped only once: `field_names = scopes.mapped('field_ids.name')`

Garamotte (changes requested):
Some small comments inline.

The CRUD API seems OK, but not fully tested yet on my side. I'm not sure about the URLs format, waiting for opinions from someone else (although this can be an endless discussion).

oauth_provider/models/oauth_provider_scope.py
lasley: Hmmm for some reason I thought there was a difference, but it seems you are correct:

```
>>> env['res.partner']
res.partner()
>>> env['res.partner'].browse()
res.partner()
```

lasley:
I'm open to changing the URI schema if some good arguments are exposed. We haven't designed anything specific on top of the REST API yet, so it would be an easy change. You are correct that this could be an endless discussion though 😆 🚀

lasley:
Comments handled @sylvain-garancher

simahawk:
@lasley just a ping in case you missed this https://github.com/OCA/server-tools/issues/1011

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,338 @@
...
+
+        Args:
+            model (str or BaseModel): Name or object of the model to operate
+                on.
+
+        Returns:
+            OauthProviderScope: Recordsets associated to the model.
+        """
+        if isinstance(model, models.BaseModel):
+            model = model._name
```
Garamotte, line 72: I find it cleaner to avoid using `isinstance` when it's not needed. Here, we can force the argument to be a string.

When we have a BaseModel instance, then we can simply call `filter_by_model(our_model._name)`.

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,338 @@
...
+        }
+        if matched_records:
+            if len(matched_records) == 1:
+                return matched_records.values()[0]
+            return matched_records
+
+        return data
+
+    @api.multi
+    def create_record(self, model_name, vals):
```
Garamotte, line 143: I don't understand why this method is in the `oauth.provider.scope` model, instead of calling the three methods (`_validate_scope_field`, `self.env['model_name'].create` and `_validate_scope_record`) directly in the `create_record` method of `oauth.provider.token` ?
What's the benefit of having this method on the scope ?

Same question about write and delete, obviously.

oauth_provider/models/oauth_provider_scope.py
```diff
@@ -0,0 +1,338 @@
...
+        Args:
+            records (BaseModel): Recordset to validate.
+
+        Returns:
+            bool: Indicating whether the records are within scope.
+        """
+
+        scoped_records = self.env[records._name]
+        for scope in self:
+            scoped_records |= scope._get_scoped_records(
```
Garamotte, line 310: I guess this will not work if you have scopes using different models on the same token.

Garamotte (commented):
I finally took time to carefully review the code, great work @lasley !
The docstrings are really good too, thanks for that (that's something I'm not yet good at).

Still not enough time to make some functional tests :(

lasley:
Thanks for the review @sylvain-garancher - I'll tend to the comments soon. I actually moved some of this code (the stuff to allow for proper receipt of JSON requests) into OCA/webhook#3 and will need to switch this code to use that. I think I'm about to make it a Python package instead though.

Regarding functional tests - I haven't used the REST API provided here much, but we do have quite a few proprietary APIs built on top of it that have been working quite well for some time. It's really helped us for automation.

eggu68:
Hello, what is the status on this module migration to newer versions? I see that it was in the 9.0 branch but not later? Any plans to have this oauth_provider in Odoo 13 or 14? Or are there other alternatives?

PR #893, opened 2017-07-08
[FIX] oauth_provider: Guard for instances of no redirects

lasley:
* Legacy application workflow does not have redirect_uris, so it was throwing an error on token generation. Return empty string in event of no redirect.

### DIFF
--- a/oauth_provider/oauth2/validator.py
+++ b/oauth_provider/oauth2/validator.py
@@ -186,7 +186,7 @@ class OdooValidator(RequestValidator):
                 datetime.now() + timedelta(seconds=token['expires_in'])),
         })
         redirect_uris = request.client.redirect_uri_ids
-        return redirect_uris and redirect_uris[0].name or ''
+        return redirect_uris[:1].name
 
     def validate_bearer_token(self, token, scopes, request):
         """ Ensure the supplied bearer token is valid, and allowed for the scopes
--- a/oauth_provider/oauth2/validator.py
+++ b/oauth_provider/oauth2/validator.py
@@ -185,7 +185,8 @@ class OdooValidator(RequestValidator):
             'expires_at': fields.Datetime.to_string(
                 datetime.now() + timedelta(seconds=token['expires_in'])),
         })
-        return request.client.redirect_uri_ids[0].name
+        redirect_uris = request.client.redirect_uri_ids
+        return redirect_uris and redirect_uris[0].name or ''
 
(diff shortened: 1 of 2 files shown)


oauth_provider/oauth2/validator.py
```diff
@@ -185,7 +185,8 @@ def save_bearer_token(self, token, request, *args, **kwargs):
             'expires_at': fields.Datetime.to_string(
                 datetime.now() + timedelta(seconds=token['expires_in'])),
         })
-        return request.client.redirect_uri_ids[0].name
+        redirect_uris = request.client.redirect_uri_ids
+        return redirect_uris and redirect_uris[0].name or ''
```
Garamotte: `return request.client.redirect_uri_ids[:1].name` seems easier to read to me, what's your opinion about that ?
This syntax is sometimes used in the standard code, and I find this trick really useful to avoid many if/else/and/or.

Garamotte (changes requested):
Good catch. I always had a value because the XML data file creates a redirect URI by default, so I never saw this mistake (even in tests). I think the data file should be fixed too.
This will need to adapt tests (that should be broken when removing the default redirect URI from XML data file).

Also, make sure that an error is raised in tests without your fix in the code :)

lasley:
Note: `web` is also a dependency of this module - `web.login_layout` is part of the OAuth dance from what I can tell. I'll add it here as well.

Garamotte:
Totally right :)

oauth_provider/oauth2/validator.py
lasley: Totally agree!

Garamotte (approved):
Could you add a small unit test for the fix ?
Otherwise :+1:

lasley:
@sylvain-garancher - I don't see a test covering the `save_bearer_token` method and I don't see it used anywhere in the code, so I'm not sure which of these tests is covering it and how to setup a test case for this specific item. Can you point me to the test here that covers the `save_bearer_token` method?

Garamotte:
All methods in the `validator.py` file are called by the `oauthlib` library.
The `save_bearer_token` method is called after a token has been created, to write it into the database: https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/tokens.py#L287

lasley:
Exactly the missing piece I needed, thank you @sylvain-garancher

PR #607, opened 2016-11-15
[MIG] database_cleanup: Migration to version 10.0

eantones:
[MIG] database_cleanup: Migration to version 10.0
Using new base model inheritance

PR #894, opened 2017-07-10
[MIG] 10.0 db cleanup

ecino:
Takes #607 with 9.0 commits for module database_cleanup.

### DIFF
--- a/database_cleanup/__manifest__.py
+++ b/database_cleanup/__manifest__.py
@@ -16,6 +16,8 @@
         'views/purge_columns.xml',
         'views/purge_tables.xml',
         'views/purge_data.xml',
+        "views/create_indexes.xml",
+        'views/purge_properties.xml',
         'views/menu.xml',
     ],
     'installable': False,
--- a/database_cleanup/__manifest__.py
+++ b/database_cleanup/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Database cleanup',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'author': "Therp BV,Odoo Community Association (OCA)",
     'depends': ['base'],
     'license': 'AGPL-3',
@@ -20,5 +20,5 @@
         'views/purge_properties.xml',
         'views/menu.xml',
     ],
-    'installable': False,
+    'installable': True,
 }
--- a/database_cleanup/models/create_indexes.py
+++ b/database_cleanup/models/create_indexes.py
@@ -2,7 +2,7 @@
 # © 2017 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from ..identifier_adapter import IdentifierAdapter
-from openerp import api, fields, models
+from odoo import api, fields, models
 
 
 class CreateIndexesLine(models.TransientModel):
@@ -48,6 +48,7 @@ class CreateIndexesWizard(models.TransientModel):
 
     @api.multi
     def find(self):
+        res = list()
         for field in self.env['ir.model.fields'].search([
                 ('index', '=', True),
         ]):
@@ -74,7 +75,8 @@ class CreateIndexesWizard(models.TransientModel):
             if not self.env.cr.rowcount:
                 continue
 
-            yield (0, 0, {
+            res.append((0, 0, {
                 'name': '%s.%s' % (field.model, field.name),
                 'field_id': field.id,
-            })
+            }))
+        return res
--- a/database_cleanup/models/create_indexes.py
+++ b/database_cleanup/models/create_indexes.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from ..identifier_adapter import IdentifierAdapter
+from openerp import api, fields, models
+
+
+class CreateIndexesLine(models.TransientModel):
+    _inherit = 'cleanup.purge.line'
+    _name = 'cleanup.create_indexes.line'
+
+    purged = fields.Boolean('Created')
+    wizard_id = fields.Many2one('cleanup.create_indexes.wizard')
+    field_id = fields.Many2one('ir.model.fields', required=True)
+
+    @api.multi
+    def purge(self):
+        tables = set()
+        for field in self.mapped('field_id'):
+            model = self.env[field.model]
+            name = '%s_%s_index' % (model._table, field.name)
+            self.env.cr.execute(
+                'create index %s ON %s (%s)',
+                (
+                    IdentifierAdapter(name, quote=False),
+                    IdentifierAdapter(model._table),
+                    IdentifierAdapter(field.name),
+                ),
+            )
+            tables.add(model._table)
+        for table in tables:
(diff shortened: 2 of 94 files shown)


pedrobaeza:
Please squash together "OCA Transbot" commits.

ecino:
Squashed some transbot commits (not all to preserve the order)

ecino:
Strange that no ir.property are present in the test database, tests are failing because of this. Any ideas?

pedrobaeza:
Create some of them in your setUp phase to see

ecino:
Ok now I have same error as in #607 which seems to come from field `threshold_exempt` from module `user_threshold` not being loaded when running tests and creating issue.

The class test is however using decorator `@post_install`, not sure what I can do more.

ecino:
Tests are fixed 😄

database_cleanup/__manifest__.py
```diff
@@ -16,7 +16,10 @@
         'views/purge_columns.xml',
         'views/purge_tables.xml',
         'views/purge_data.xml',
+        "views/create_indexes.xml",
+        'views/purge_properties.xml',
         'views/menu.xml',
     ],
-    'installable': False,
+    'installable': True,
+    "application": True,
```
StefanRijnhart: Not an application IMHO

ecino: You're right, I removed it.

database_cleanup/views/purge_properties.xml
```diff
@@ -0,0 +1,47 @@
...
+            <data />
+        </field>
+    </record>
+
+    <record id="action_purge_property" model="ir.actions.server">
+        <field name="name">Purge properties</field>
+        <field name="type">ir.actions.server</field>
+        <field name="state">code</field>
+        <field name="model_id" ref="database_cleanup.model_cleanup_purge_wizard_property" />
+        <field name="code">action = self.get_wizard_action(cr, uid, context=context)</field>
```
StefanRijnhart: I get spurious errors
```
ValueError: <type 'exceptions.NameError'>: "name 'self' is not defined" while evaluating u'action = self.get_wizard_action(cr, uid, context=context)'
```

when clicking on some of the menu items. It seems to help to replace the code contents of lines like this with

```action = env['cleanup.purge.wizard.property'].get_wizard_action()```

Comment (ecino) [reply database_cleanup/views/purge_properties.xml]: Corrected

Comment (StefanRijnhart) [reply database_cleanup/views/purge_properties.xml]: Thanks! You missed one, the one I referred to initially.

File: database_cleanup/views/purge_wizard.xml [hunk 3]
```diff
@@ -1,41 +1,42 @@
...
+                            <field name="purged" />
+                        </group>
+                        <footer>
+                            <button type="object" name="purge" class="oe_highlight"
+                                    string="Purge"
+                                    attrs="{'invisible': [('purged', '=', True)]}"/>
+                        </footer>
+                    </form>
+                    <div attrs="{'invisible': [('purge_line_ids', '!=', [])]}">
+                        Nothing found to clean up.
```
StefanRijnhart: This always shows up, even when there are result lines present. Maybe the way to check for an empty one2many widget has changed.

image

database_cleanup/views/purge_properties.xml
ecino: Sorry for that, it's now included

database_cleanup/views/purge_wizard.xml
ecino: I tried to figure out how to fix it but didn't came out with a working solution. By inspecting odoo source code I found code which is exactly the same as here. Any help from others would be appreciated.

StefanRijnhart: The problem is that the div is in the field tag itself. It works when the div is moved before the field tag.

ecino: Thanks, now it works.

database_cleanup/views/create_indexes.xml
```diff
@@ -0,0 +1,52 @@
...
+            </button>
+        </field>
+    </record>
+
+    <record id="cleanup_create_indexes_wizard_action" model="ir.actions.server">
+        <field name="name">Create missing indexes</field>
+        <field name="type">ir.actions.server</field>
+        <field name="state">code</field>
+        <field name="model_id" ref="database_cleanup.model_cleanup_create_indexes_wizard" />
+        <field name="code">action = env.get('cleanup.purge.wizard.data').get_wizard_action()</field>
```
StefanRijnhart: Wrong model. Should be 'cleanup.create_indexes.wizard'

database_cleanup/views/purge_properties.xml
```diff
@@ -0,0 +1,47 @@
...
+            <data />
+        </field>
+    </record>
+
+    <record id="action_purge_property" model="ir.actions.server">
+        <field name="name">Purge properties</field>
+        <field name="type">ir.actions.server</field>
+        <field name="state">code</field>
+        <field name="model_id" ref="database_cleanup.model_cleanup_purge_wizard_property" />
+        <field name="code">action = env.get('cleanup.purge.wizard.data').get_wizard_action()</field>
```
StefanRijnhart: Wrong model. Should be 'cleanup.purge.wizard.property'

ecino: Fixed. Sorry for the confusion.

StefanRijnhart (approved):
Thanks for all the quick fixes!

database_cleanup/models/create_indexes.py
```diff
@@ -74,7 +75,8 @@ def find(self):
             if not self.env.cr.rowcount:
                 continue
 
-            yield (0, 0, {
```
hbrunn: what was the problem with the iterator?

database_cleanup/models/purge_columns.py
```diff
@@ -20,7 +20,12 @@ def purge(self):
         """
         Unlink columns upon manual confirmation.
         """
-        for line in self:
+        if self:
+            objs = self
+        else:
```
hbrunn, line 25: if this happens, I think something went wrong in the UI. How can I reproduce this?

database_cleanup/models/purge_models.py
```diff
@@ -1,41 +1,38 @@
...
+class IrModelFields(models.Model):
+    _inherit = 'ir.model.fields'
+
+    @api.multi
+    def _prepare_update(self):
+        # Allow to skip this step during model unlink
+        # The super method crashes if the model cannot be instantiated
+        if self.env.context.get('no_prepare_update'):
+            return True
+        return super(IrModelFields, self)._prepare_update()
```
hbrunn: can't you filter `self` for existing models as done above? This also would make the extra context key unnecessary

hbrunn (changes requested):
small things. I also made CompassionCH/server-tools#1 for an issue when purging modules

database_cleanup/models/purge_columns.py
hbrunn: PS: If you really need this for whatever reason, just assign to `self` and leave the rest of the code as it is. It's just a variable like any other

alan196 (approved):

moylop260:
@ecino What about @hbrunn comments?

ecino:
@moylop260 I missed those comments, I will look into it when I have time.

hbrunn:
@ecino I addressed those in https://github.com/CompassionCH/server-tools/pull/2

ecino:
thanks @hbrunn, I merged your commits.

hbrunn (approved):

PR #896, opened 2017-07-11
[10.0] date_range: improve 'name' for generator wizard

alexis-via:
With the current code, when you run the wizard for monthly period over the year 2017 you get:
- 2017-1
- 2017-2
...
- 2017-11
- 2017-12

With this PR, you will get:
- 2017-01
- 2017-02
...
- 2017-11
- 2017-12

That way, the alphabetical order corresponds to the date order.

ir.rule should be active by default (it is very uncommon/strange to create inactive ir.rule).

### DIFF
--- a/date_range/wizard/date_range_generator.py
+++ b/date_range/wizard/date_range_generator.py
@@ -43,7 +43,7 @@ class DateRangeGenerator(models.TransientModel):
                      count=self.count+1)
         vals = list(vals)
         date_ranges = []
-        month_digits = len(unicode(self.count))
+        count_digits = len(unicode(self.count))
         for idx, dt_start in enumerate(vals[:-1]):
             date_start = fields.Date.to_string(dt_start.date())
             # always remove 1 day for the date_end since range limits are
@@ -51,9 +51,8 @@ class DateRangeGenerator(models.TransientModel):
             dt_end = vals[idx+1].date() - relativedelta(days=1)
             date_end = fields.Date.to_string(dt_end)
             date_ranges.append({
-                'name': '%s%s' % (
-                    self.name_prefix,
-                    unicode(idx + 1).zfill(month_digits)),
+                'name': '%s%0*d' % (
+                    self.name_prefix, count_digits, idx + 1),
                 'date_start': date_start,
                 'date_end': date_end,
                 'type_id': self.type_id.id,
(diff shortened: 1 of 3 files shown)


date_range/wizard/date_range_generator.py
```diff
@@ -43,14 +43,17 @@ def _compute_date_ranges(self):
...
+        month_digits = len(unicode(self.count))
         for idx, dt_start in enumerate(vals[:-1]):
             date_start = fields.Date.to_string(dt_start.date())
             # always remove 1 day for the date_end since range limits are
             # inclusive
             dt_end = vals[idx+1].date() - relativedelta(days=1)
             date_end = fields.Date.to_string(dt_end)
             date_ranges.append({
-                'name': '%s-%d' % (self.name_prefix, idx + 1),
+                'name': '%s-%s' % (
```
sbidoul: That was needed indeed. While you are at it could you remove the `-`. Sometimes we don't want a dash, or we want something else, so it's better to let the user add it in the prefix.

date_range/wizard/date_range_generator.py
```diff
@@ -43,14 +43,17 @@ def _compute_date_ranges(self):
...
+                    self.name_prefix,
+                    unicode(idx + 1).zfill(month_digits)),
```
Garamotte: Use some python magic instead: `'name': '%s-%02d' % (self.name_prefix, idx + 1),`

Garamotte (changes requested):

date_range/wizard/date_range_generator.py
sbidoul: @sylvain-garancher it's not necessarily 2 digits. One is quite common when you generate Quarters. BTW one could say `month_digits` is misnamed.

Garamotte: Oh ok, sorry, so: `'name': '%s-%0*d' % (self.name_prefix, month_digits, idx + 1)`

alexis-via:
@sbidoul I took into account your remark
@sylvain-garancher nice suggestion, but in my case the size is dynamic and I'm not sure it's possible to have a dynamic size with the "python magic" syntax.

Garamotte:
My second proposal allows a dynamic size : `'name': '%s-%0*d' % (self.name_prefix, month_digits, idx + 1)`

sbidoul (approved):

MiquelRForgeFlow (approved):

Garamotte (approved):

sebalix (approved):

PR #897, opened 2017-07-12
Migrate module auth_session_timeout from 9.0 to 10.0

dperaltab:
Migrate module auth_session_timeout from 9.0 to 10.0


pedrobaeza:
There is already another migration for this module (with incorrect name yet - https://github.com/OCA/server-tools/pull/875), and a big change in v9 that deserves to be considered in v10 (https://github.com/OCA/server-tools/pull/887). Please review them instead.

lasley:
Closing for #580

PR #903, opened 2017-07-19
[8.0] Add auditlog custom methods

lembregtse:
Add the possibility for a user to add logging to custom methods, next to read, create, write and unlink.

These custom_methods can be set per auditlog.rule, for each custom method a user can add a log message that will be used for the auditlog.log method field (for clear visibility). 

To allow support for wizards and other intermediate models we allow for the use of active_ids and active_model in the context variables when creating the auditlog.log line.


auditlog/tests/test_auditlog.py
```diff
@@ -114,3 +114,59 @@ def setUp(self):
...
+
+        # Create account period to test
+        self.partner = self.env['res.partner'].create({
+            'name': 'Test User'
+        })
+
+        self.partner_model = self.env['ir.model'].search([
+            ('model', '=', 'res.partner')])
+
+        # Setup rule for model "account.period.close" and method "data_save"
```
StefanRijnhart: Outdated comment ;-)

auditlog/models/rule.py
```diff
@@ -71,6 +72,11 @@ class AuditlogRule(models.Model):
         u"Log Creates", default=True,
         help=(u"Select this if you want to keep track of creation on any "
               u"record of the model of this rule"))
+    log_custom = fields.Boolean(
```
sebalix: I suggest `log_custom_method` to be more specific about the feature.

sebalix (changes requested):
Nice feature, just one comment about the field name.

auditlog/models/rule.py
```diff
@@ -364,11 +472,20 @@ def create_logs(self, uid, res_model, res_ids, method,
         http_session_model = self.env['auditlog.http.session']
         for res_id in res_ids:
             model_model = self.env[res_model]
+            # Do extra check for active_model situations where res_model
+            # is not preloaded in auditlog model_cache
+            model_id = self.pool._auditlog_model_cache.get(res_model, False)
```
StefanRijnhart: False as a default is redundant. If the model is not present, dict.get() returns None. Maybe make this a little more compact as
```
self.pool._auditlog_model_cache.get(
    res_model,
    self.env['ir.model'].search([('model', '=', res_model)]).id)
```
you will lose the check on whether the model exists but a missing model was going to error out in line 474 anyway.

lembregtse: True, while I do agree with the rewrite, I would still like to make use of the caching dict to prevent unnecessary lookups afterwards.

I would propose something like:

```
if not self.pool._auditlog_model_cache.get(res_model):
    self.pool._auditlog_model_cache[res_model] = \
        self.env['ir.model'].search([
            ('model', '=', res_model)]).id
```

StefanRijnhart: Yes, I fell for the trap that get()'s default is evaluated immediately. +1 for your suggestion.

auditlog/models/rule.py
lembregtse: Good suggestion, I've adapted it accordingly.

auditlog/models/rule.py
lembregtse: Refactored as such.

auditlog/models/methods.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class AuditlogMethods(models.Model):
+    _name = 'auditlog.methods'
+    _description = 'Auditlog custom methods'
+    _order = 'name'
+
+    name = fields.Char(required=True)
+    message = fields.Char(required=True)
+    rule_id = fields.Many2one('auditlog.rule', readonly=True)
```
StefanRijnhart, line 13: Maybe set to required + ondelete='cascade'?

StefanRijnhart (approved):
One small remaining suggestion, but approve otherwise.

StefanRijnhart:
Added a commit of general interest that helps Odoo loading in case of data model changes in auditlog rules.

sebalix (approved):

PR #906, opened 2017-07-21
[FIX] mass_editing : bug in view

fmdl:
Version : 10.0 and mayby other

Before :
image

After : 
image


pedrobaeza (approved):

Garamotte:
Your fix is OK for the button, but it makes the "Model" field very small.
Could you fix this ?

Garamotte (approved):

PR #909, opened 2017-07-25
[RFR] Split off trigram index module

StefanRijnhart:
Suggesting to split off the creation of trigram indexes from the fuzzy search module that comes with a monkey patch on expression.py.

cc @christophlsa


hparfr (approved):
👍

PR #918, opened 2017-08-01
[MIG][module_auto_update] Backport from v10

yajo:
Backporting #882 

@Tecnativa

### DIFF
--- a/module_auto_update/README.rst
+++ b/module_auto_update/README.rst
@@ -34,7 +34,7 @@ To perform upgrades manually, click the "Apply Scheduled Upgrades" menu item in 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/149/10.0
+   :target: https://runbot.odoo-community.org/runbot/149/9.0
 
 Bug Tracker
 ===========
@@ -57,6 +57,7 @@ Contributors
 
 * Brent Hughes <brent.hughes@laslabs.com>
 * Juan José Scarafía <jjs@adhoc.com.ar>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
 
 Do not contact contributors directly about support or help with technical issues.
 
--- a/module_auto_update/README.rst
+++ b/module_auto_update/README.rst
@@ -1,0 +1,76 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==================
+Module Auto Update
+==================
+
+This module will automatically check for and apply module upgrades on a schedule.
+
+Upgrade checking is accomplished by comparing the SHA1 checksums of currently-installed modules to the checksums of corresponding modules in the addons directories.
+
+Installation
+============
+
+Prior to installing this module, you need to:
+
+#. Install checksumdir with `pip install checksumdir`
+#. Ensure all installed modules are up-to-date. When installed, this module will assume the versions found in the addons directories are currently installed.
+
+Configuration
+=============
+
+The default time for checking and applying upgrades is 3:00 AM (UTC). To change this schedule, modify the "Perform Module Upgrades" scheduled action.
+
+This module will ignore .pyc and .pyo file extensions by default. To modify this, create a module_auto_update.checksum_excluded_extensions system parameter with the desired extensions listed as comma-separated values.
+
+Usage
+=====
+
+Modules scheduled for upgrade can be viewed by clicking the "Updates" menu item in the Apps sidebar.
+
+To perform upgrades manually, click the "Apply Scheduled Upgrades" menu item in the Apps sidebar.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
+
+Bug Tracker
+===========
+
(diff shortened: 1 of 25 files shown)


yajo:
Cool! all 💚

module_auto_update/models/module.py
```diff
@@ -30,11 +30,15 @@ def _compute_checksum_dir(self):
         ).split(",")
 
         for r in self:
-            r.checksum_dir = dirhash(
-                get_module_path(r.name),
-                'sha1',
-                excluded_extensions=exclude,
-            )
+            try:
```
pedrobaeza: Don't do this, but assign `dirhash = None` in import, and check here `if dirhash:`, but this should be declared as external dependency, so if the code get this, it's because you make sure the lib to be installed.

module_auto_update/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Usage
+=====
+
+Modules scheduled for upgrade can be viewed by clicking the "Updates" menu item in the Apps sidebar.
+
+To perform upgrades manually, click the "Apply Scheduled Upgrades" menu item in the Apps sidebar.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/10.0
```
pedrobaeza: 9.0

module_auto_update/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Module Auto Update',
+    'summary': 'Automatically update Odoo modules',
+    'version': '9.0.1.0.0',
+    'category': 'Extra Tools',
+    'website': 'https://odoo-community.org/',
+    'author': 'LasLabs, '
+              'Juan José Scarafía, '
```
pedrobaeza, line 12: Shouldn't this be Ingeniería AdHoc? cc @jjscarafia

module_auto_update/models/module.py
yajo: This is here because `get_module_path` can return `False` if an addon is installed bus it not found on disk. Then, `dirhash` raises a `TypeError` because it expects a `str`.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @api.model
+    def create(self, vals):
+        res = super(Module, self).create(vals)
+        res._store_checksum_installed(vals)
+        return res
+
+    @api.model
+    def update_list(self):
+        res = super(Module, self).update_list()
```
pedrobaeza: This should be removed, as it's totally anti-performance, specially on OCB, where installing or updating a module first trigger a module list update, which in case of very large installations will take a lot of time. You already have a method for updating pending modules (running cron job manually), so don't force it here.

pedrobaeza (changes requested):
The case we have mentioned in the Docker project (new dependency added to installed modules) should be treated also:

- Check which modules need to be updated (already done).
- Update module list (but removing the part in this method that call again to update).
- Check dependencies of these modules.
- If some of the dependencies are not installed, then marked them for being installed.

There's still the chance that your installation doesn't contain the new dependency, but at least we reduce the possibilities.

module_auto_update/models/module.py
pedrobaeza: OK, thanks for clarifying. Can you please put the same comment in code for later references?

module_auto_update/models/module.py
lasley: I didn't actually investigate, but I think this could be some of the issue behind the [deadlocks I mentioned](https://github.com/Tecnativa/docker-odoo-base/pull/64#discussion_r130646986)

lasley:
cc @hughesbm

module_auto_update/models/module.py
yajo: It's already there, above the `pass`

pedrobaeza: Ouch, got it

module_auto_update/models/module.py
yajo: I tested in a production database and indeed running `odoo --stop-after-init -u any_module` reloads dir hashes for all installed addons. Let's move all of this to the wizard method (which is also the one called from cron and https://github.com/Tecnativa/docker-odoo-base/pull/64) and I'll forward-port the patches.

yajo:
The refactor goes in no-fixup commits to be able to cherry-pick it later for v10. It required to change some tests.

module_auto_update/models/module.py
```diff
@@ -0,0 +1,80 @@
...
+        ).split(",")
+
+        for r in self:
+            try:
+                r.checksum_dir = dirhash(
+                    get_module_path(r.name),
+                    'sha1',
+                    excluded_extensions=exclude,
+                )
+                _logger.debug("Computed %s for %s", r.checksum_dir, r.name)
```
pedrobaeza: Should we keep this message?

pedrobaeza (approved):

module_auto_update/models/module.py
```diff
@@ -0,0 +1,80 @@
...
+            except TypeError:
+                # Module path not found
+                pass
```
lasley: Never pass silently

lasley (approved):

module_auto_update/models/module.py
lasley: Nah I think we should move it to the fail case instead (although with fail verbiage instead)

yajo:
Rebased, squashed for merge

rafaelbn (approved):
We are using it in production enviroments

PR #921, opened 2017-08-02
[FIX][9.0] date_range. Periods within the day should be allowed

AaronHForgeFlow:
In some cases it may be necessary to create periods for just one day. For example: when generating a payroll period for a single day of work: https://github.com/OCA/hr/pull/354

As there's an option for creating overlapping periods I think this change it's not harmful. Users still can set the date_range_type to don't allow overlapping periods.

### DIFF
--- a/date_range/__openerp__.py
+++ b/date_range/__openerp__.py
@@ -4,7 +4,7 @@
 {
     "name": "Date Range",
     "summary": "Manage all kind of date range",
-    "version": "9.0.1.0.0",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Change version number

MiquelRForgeFlow (approved):

sbidoul (approved):

pedrobaeza (approved):

pedrobaeza:
Please forward port it to v10

PR #922, opened 2017-08-03
[8.0] base_user_role: backport features from 9.0 + Fix from 10.0

sebalix:


sebalix:
WIP, The code doesn't work as expected on 8.0

sebalix:
Done, ready for review.

andhit-r (approved):
Functional test :+1:

sebalix:
Just added a backport of #934 .

StefanRijnhart (approved):

StefanRijnhart (commented):
I had a problem that removing the role did not remove the user's groups from that role. Of course, I was testing with just a single role and after removing that one, the user does not have any roles and the user's groups are left alone. Would you consider removing all the user's groups in this case?

sebalix:
@StefanRijnhart you mean just update the users' groups right after deleting the role? It should be fixed indeed (I'll update this PR). Regarding the group itself, I wonder if we should remove it as it can be linked to some views, menus, etc.

StefanRijnhart:
No, I mean that `users.set_groups_from_roles()` is called when a role line is removed, but it does not do anything if it was the last line. I think that set_groups_from_roles should have a `force` option to make it clean up after the last role line was removed instead of leaving the groups associated with the last role line.

Test scenario:
User has two role lines. Role line 1 adds financial manager group. Role line 2 adds sale manager group.
Role line 2 is removed first -> user is not in the sale manager group anymore.
Role line 1 is removed next -> user is not in the financial manager group anymore.

This last test currently fails.

sebalix:
@StefanRijnhart ok it's should be fixed now, I added 2 tests to check that the user's groups are cleaned up when its last line is removed either through its form, or indirectly by removing the related role.

StefanRijnhart:
Excellent, thank you!

StefanRijnhart (approved):

sebalix:
Rebased on top of #994

sebalix:
I think this one could be merged, do you agree?

StefanRijnhart:
Agreed!

sebalix:
@StefanRijnhart thank you!

sebalix:
Something weird happened, some commits were pushed on the PR branch (`osiell:8.0-base_user_role_backport_features`) but they are not listed here, there was some fixes and unit tests added.
I'll try to remake a PR because the current one added a regression when a role is unlinked.

PR #925, opened 2017-08-04
[8.0] Add module export_sql_mail

florian-dacosta:
The module depends on sql_export and allow to send the result files from the queries by mail to a list of users.
There is the possibility to receive the mail only if the query is not empty, so it may be used to monitor some key datas.


florian-dacosta:
@bealdav @aheficent

sql_export_mail/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
```
bealdav: Please a shorter header

sql_export_mail/models/sql_export.py
```diff
@@ -0,0 +1,155 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
```
bealdav: shorter header please

sql_export_mail/models/sql_export.py
```diff
@@ -0,0 +1,155 @@
...
+
+class SqlExport(models.Model):
+    _inherit = 'sql.export'
+
+    mail_user_ids = fields.Many2many(
+        'res.users',
+        'mail_user_sqlquery_rel',
+        'sql_id',
+        'user_id',
```
bealdav, line 19: The 3 args above are useless, I think, default generated ones  are sufficient

idem in the m2m below

sql_export_mail/models/sql_export.py
```diff
@@ -0,0 +1,155 @@
...
+            'user_id': SUPERUSER_ID,
+        }
+        cron = self.env['ir.cron'].create(cron_vals)
+        write_vals = {'args': '[[%s]]' % cron.id}
+        cron.write(write_vals)
+
+        self.write({'cron_ids': [(4, cron.id)]})
+
+    @api.one
+    def send_mail(self, params=None):
```
bealdav: not blocking, but you can use this lib https://github.com/OCA/oca-decorators to avoid deprectaed @api.one

bealdav (changes requested):
Except signaled typo, LGTM

sql_export_mail/models/sql_export.py
florian-dacosta: I thought it wasn't deprecated anymore, my bad.
Anyway, the api.one was useless in this case, I juste replaced by api.multi

sql_export_mail/models/sql_export.py
florian-dacosta: Indeed, theses 3 args are not very useful, changing it would break things for customers already using it though, so I would prefer let it this way.

bealdav (approved):

andhit-r (approved):

Issue #927, opened 2017-08-07
[9.0][module_auto_update] Remove hacks from module_auto_update


yajo:
Now that https://github.com/odoo/odoo/pull/18597 is merged, we can safely remove workarounds from https://github.com/OCA/server-tools/blob/9.0/module_auto_update/models/module.py#L54-L68

pedrobaeza:
Please make the PR and also forwardport changes made in 9.0 to 10.0

sbidoul:
@Yajo this one can thus be closed?

pedrobaeza:
Yes, closing.

PR #928, opened 2017-08-08
[REF][module_auto_update] Forward port v9 improvements

lasley:
Forward port of #918

* Recompute dir hashes only when needed - By removing the recomputation from `update_list` we get faster CLI module upgrades and it only performs the autoupdate when using the autoupdate wizard or cron.
* Pass tests if addon is in readonly directory
* Set dependencies to upgrade

[FIX][module_auto_update] Pass tests if addon is in readonly directory

[FIX][module_auto_update] Set dependencies to upgrade

### DIFF
--- a/module_auto_update/models/module.py
+++ b/module_auto_update/models/module.py
@@ -52,6 +52,13 @@ class Module(models.Model):
                 self.write({'checksum_installed': False})
 
     @api.multi
+    def button_uninstall(self):
+        return super(
+            Module,
+            self.with_context(module_uninstall=True),
+        ).button_uninstall()
+
+    @api.multi
     def button_uninstall_cancel(self):
         return super(
             Module,
--- a/module_auto_update/models/module.py
+++ b/module_auto_update/models/module.py
@@ -30,11 +30,16 @@ class Module(models.Model):
         ).split(",")
 
         for r in self:
-            r.checksum_dir = dirhash(
-                get_module_path(r.name),
-                'sha1',
-                excluded_extensions=exclude,
-            )
+            try:
+                r.checksum_dir = dirhash(
+                    get_module_path(r.name),
+                    'sha1',
+                    excluded_extensions=exclude,
+                )
+            except TypeError:
+                _logger.debug(
+                    "Cannot compute dir hash for %s, module not found",
+                    r.display_name)
 
     def _store_checksum_installed(self, vals):
         if self.env.context.get('retain_checksum_installed'):
@@ -64,16 +69,6 @@ class Module(models.Model):
     def create(self, vals):
         res = super(Module, self).create(vals)
         res._store_checksum_installed(vals)
-        return res
-
-    @api.model
-    def update_list(self):
-        res = super(Module, self).update_list()
-        installed_modules = self.search([('state', '=', 'installed')])
-        upgradeable_modules = installed_modules.filtered(
-            lambda r: r.checksum_dir != r.checksum_installed,
-        )
-        upgradeable_modules.write({'state': "to upgrade"})
         return res
 
     @api.multi
--- a/module_auto_update/wizards/module_upgrade.py
+++ b/module_auto_update/wizards/module_upgrade.py
@@ -8,6 +8,16 @@ from odoo import api, models
 class ModuleUpgrade(models.TransientModel):
     _inherit = 'base.module.upgrade'
 
+    @api.model
+    def get_module_list(self):
+        Module = self.env["ir.module.module"]
+        installed_modules = Module.search([('state', '=', 'installed')])
+        upgradeable_modules = installed_modules.filtered(
+            lambda r: r.checksum_dir != r.checksum_installed,
+        )
+        upgradeable_modules.button_upgrade()
+        return super(ModuleUpgrade, self).get_module_list()
+
     @api.multi
     def upgrade_module_cancel(self):
         return super(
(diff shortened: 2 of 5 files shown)


module_auto_update/models/module.py
```diff
@@ -48,34 +53,26 @@ def _store_checksum_installed(self, vals):
 
     @api.multi
     def button_uninstall_cancel(self):
-        return super(
-            Module,
-            self.with_context(retain_checksum_installed=True),
-        ).button_uninstall_cancel()
+        # TODO Use super() like in v10 after pull is merged
+        # HACK https://github.com/odoo/odoo/pull/18597
```
yajo: This was fixed (see #927) and besides it never affected v10, so leave this as it was please.

module_auto_update/models/module.py
```diff
@@ -48,34 +53,26 @@ def _store_checksum_installed(self, vals):
...
+        })
 
     @api.multi
     def button_upgrade_cancel(self):
-        return super(
-            Module,
-            self.with_context(retain_checksum_installed=True),
-        ).button_upgrade_cancel()
+        # TODO Use super() like in v10 after pull is merged
+        # HACK https://github.com/odoo/odoo/pull/18597
```
yajo: Same here

yajo (changes requested):
I think you forgot commit history 😿

lasley:
> I think you forgot commit history 😿

I did a format-patch of your PR branch, then squashed your three fixit commits into the one that you see here. I did slightly edit the message header to be all inclusive, with your original notes as the content.

Did I do something incorrectly in that? Sounds like I maybe didn't need to include all of the three commits?

_Edit: Actually I cherry-picked now that I think about it_

yajo:
Then I think I forgot to take a closer look at it instead 😆

OK, forget it, just please amend those 2 little details above 😉

lasley:
@Yajo - removed the hacks. Seeing as it was just removal, I went ahead and just squashed it into your commit so we have a cleaner history.

module_auto_update/models/module.py
```diff
@@ -46,36 +51,12 @@ def _store_checksum_installed(self, vals):
...
-            Module,
-            self.with_context(retain_checksum_installed=True),
-        ).button_uninstall_cancel()
-
-    @api.multi
-    def button_upgrade_cancel(self):
-        return super(
-            Module,
-            self.with_context(retain_checksum_installed=True),
-        ).button_upgrade_cancel()
```
yajo: I think you removed too much. The hack was to workaround https://github.com/odoo/odoo/pull/18597, but the code you had here was needed AFAIK...

module_auto_update/wizards/module_upgrade.py
```diff
@@ -8,6 +8,16 @@
 class ModuleUpgrade(models.TransientModel):
     _inherit = 'base.module.upgrade'
 
+    @api.model
+    def get_module_list(self):
```
yajo, line 12: We have a problem currently in this method. It gets triggered when user presses "Uninstall" button in any addon. That's a problem because you want to uninstall an addon and it turns out to ask you to upgrade a bunch of them sometimes. It happens in v9 too. I think it's a good chance to patch it too, what do you think?

yajo (changes requested):

module_auto_update/wizards/module_upgrade.py
lasley: Ahhhh so that's what caused this - I noticed it the other day then just kind of shrugged it off.

pedrobaeza:
Please add `checksumdir` to `requirements.txt`

lasley:
> Please add checksumdir to requirements.txt

[It's already there](https://github.com/LasLabs/server-tools/blob/aa2029ea6b278d0c3816ebc5b929f4ae0c505915/requirements.txt#L10). Am I missing something?

pedrobaeza:
Runbot is not recognizing it

lasley:
Hmmm it's not recognizing it on OCA/10.0 either, but it just happened after the merge of #930. That's weird..

image

lasley:
There's even a line where it installs:

```
Collecting checksumdir (from -r /home/odoo/build/OCA/server-tools/requirements.txt (line 10))
  Downloading checksumdir-1.1.4.tar.gz
```

lasley:
Looks like we're being affected by https://github.com/Vauxoo/docker-odoo-image/pull/263- will need to rebuild to test after https://hub.docker.com/r/vauxoo/odoo-80-image-shippable-auto/builds/ has finishes building

module_auto_update/wizards/module_upgrade.py
```diff
@@ -10,6 +10,8 @@ class ModuleUpgrade(models.TransientModel):
 
     @api.model
     def get_module_list(self):
+        if self.env.context.get('module_uninstall'):
+            return super(ModuleUpgrade, self).get_module_list()
```
yajo: Just to DRY, I'd prefer an `if not` and let the return at the end.

yajo (approved):

module_auto_update/wizards/module_upgrade.py
```diff
@@ -8,6 +8,18 @@
 class ModuleUpgrade(models.TransientModel):
     _inherit = 'base.module.upgrade'
 
+    @api.model
+    def get_module_list(self):
+        if self.env.context.get('module_uninstall'):
+            return super(ModuleUpgrade, self).get_module_list()
```
yajo: Oops my comment went in the wrong commit.

I was saying that here we duplicate the `super(...).get_module_list()` call, I'd prefer to have only one and just move all else inside the `if not` to get DRY.

yajo (approved):

pedrobaeza (changes requested):
Change module version and squash for merging.

lasley:
@pedrobaeza -done!

pedrobaeza (approved):

PR #929, opened 2017-08-10
[FIX] auth_admin_passkey: normal users are not able to read the admin user email

sebalix:
If a normal user and the administrator have the same password, an AccessError is raised (email of the admin account can not be read) and no security alert is sent by email.


legalsylvain (approved):
Thanks !

lasley (approved):

MiquelRForgeFlow (approved):

PR #930, opened 2017-08-10
[ADD] base_locale_uom_default: Add default unit of measure by lang

lasley:
* Create module to allow for the selection of default units of measure at the language level.

An implementation sample is available at https://github.com/OCA/vertical-medical/pull/191

Locale - Default UoM
====================

This module adds a concept of a default unit of measure on languages, unique by
unit category type.

It also provides a method that can be used in fields to work from said defaults.

Configuration
=============

Set default unit of measures in the `Languages` menu in settings.

Usage
=====

Fields that want to implement the language default should use the provided method,
such as in the below example:

```python
class MyModel(models.Model):
    _name = 'my.model'
    time_uom_id = fields.Many2one(
        string='Time Units',
        comodel_name='product.uom',
        default=lambda s: s.env['res.lang'].default_uom_by_category('Time'),
    )
```

### DIFF
--- a/base_locale_uom_default/models/res_lang.py
+++ b/base_locale_uom_default/models/res_lang.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class ResLang(models.Model):
+
+    _inherit = 'res.lang'
+
+    default_uom_ids = fields.Many2many(
+        string='Default Units',
+        comodel_name='product.uom',
+    )
+
+    @api.multi
+    @api.constrains('default_uom_ids')
+    def _check_default_uom_ids(self):
+        for record in self:
+            categories = set(record.default_uom_ids.mapped('category_id'))
+            if len(categories) != len(record.default_uom_ids):
+                raise ValidationError(_(
+                    'Only one default unit of measure per category may '
+                    'be selected.',
+                ))
+
+    @api.model
+    def default_uom_by_category(self, category_name, lang=None):
+        """Return the default UoM for language for the input UoM Category.
+
+        Args:
+            category_name (str): Name of the UoM category to get the default
+            for.
+            lang (ResLang or str, optional): Recordset or code of the language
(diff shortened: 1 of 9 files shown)


yajo (approved):

base_locale_uom_default/models/res_lang.py
```diff
@@ -0,0 +1,57 @@
...
+            omitted.
+
+        Returns:
+            ProductUom: Unit of measure representing the default, if set.
+            Empty recordset otherwise.
+        """
+        if lang is None:
+            lang = self.env.user.lang
+        if isinstance(lang, basestring):
+            lang = self.env['res.lang'].search([('code', '=', lang)])[:1]
```
simahawk: this could use ``limit=1`` instead of slicing and you could use ``('category_id.name`,'=', category_name)``

base_locale_uom_default/models/res_lang.py
```diff
@@ -0,0 +1,57 @@
...
+    default_uom_ids = fields.Many2many(
+        string='Default Units',
+        comodel_name='product.uom',
+    )
+
+    @api.multi
+    @api.constrains('default_uom_ids')
+    def _check_default_uom_ids(self):
+        for record in self:
+            uoms = {}
```
simahawk: can't you make this more readable? For instance:
```python
categs = record.default_uom_ids.mapped('category_id')
multiple = [x for x in categs if categs.count(x) > 1]
if multiple:
    raise ...
```

simahawk (changes requested):

max3903 (approved):

base_locale_uom_default/models/res_lang.py
lasley: I thought about that- but it's two iterations, two data structures, and one conditional vs. one iteration and one data structure and no conditional. It seemed more efficient to do this direction IMO, although I guess there's not many records so it probably doesn't make much of a difference.

lasley:
Thanks for the reviews everyone. @simahawk - comments attended to.

simahawk (approved):

lasley:
Woot we have enough reviews for a quick merge! I'm going to go ahead and do it then so we can start getting all the dependent PRs wrapped up too.

Thanks again for the reviews everyone - I'm hoping we all get some good usage out of this one. Backport inbound

PR #934, opened 2017-08-11
[10.0][FIX] User group not updating when removing a user from the role view

benwillig:
If you remove users from a role, users groups won't be updated. This PR fixes this behavior.


sebalix (approved):

adrienpeiffer (approved):
:+1:

sbidoul (approved):

lmignon (approved):

PR #936, opened 2017-08-12
[ADD] website_password_security: Glue fix module

lasley:
* Add conditional logic in the event that the user is not in the env due to password security failure; fixes OCA#914

cc @hmrodrigues


lasley:
Close in favor of #938

PR #937, opened 2017-08-13
[FIX] missing sudo in server_environment_ir_config_param

sbidoul:
Without this sudo get_param would fail when the first user reading a parameter that has changed in the configuration file does not have write access to system parameters.


lmignon (approved):
Good catch!

sbidoul:
This needs to be backported to 9.

sebalix (approved):

pedrobaeza (changes requested):
Please change version number

lmignon:
@pedrobaeza The version number is changed.... This one can be merged.

pedrobaeza (approved):

Issue #914, opened 2017-07-29
[10.0][password_security] Internal server error on wrong login


hmrodrigues:
Having the module password_security installed, if a login is incorrect, instead of showing the wrong login message, it shows a internal server error.

Steps to reproduce:
- On Odoo 10, install password_security
-  Log off and log in with a incorrect combination

Expected
- On a wrong login, show the  **Wrong login/password** message

Best Regards,
Hugo Rodrigues

lasley:
Hi @hmrodrigues - There is no `password_policy` module that I know of. Do you mean `password_security`?

hmrodrigues:
@lasley yes sorry my bad. I fixed the title and issue

lasley:
@hmrodrigues Hmmmm it's working fine for me on Runbot. On a hunch, I installed `website` & was then able to replicate the issue.
```
2017-07-30 17:33:04,969 167 WARNING openerp_test odoo.addons.auth_brute_force.controllers.main: Authentication failed from remote '172.17.42.1'. Login tried : 'asdfsd'. Attempt 1 / 10.
2017-07-30 17:33:05,221 167 ERROR openerp_test odoo.addons.website.models.ir_http: 500 Internal Server Error:

Traceback (most recent call last):
File "/home/odoo/odoo-10.0/addons/website/models/ir_http.py", line 273, in _handle_exception
response = super(Http, cls)._handle_exception(exception)
File "/.repo_requirements/odoo/odoo/addons/base/ir/ir_http.py", line 169, in _handle_exception
return request._handle_exception(exception)
File "/.repo_requirements/odoo/odoo/http.py", line 768, in _handle_exception
return super(HttpRequest, self)._handle_exception(exception)
File "/.repo_requirements/odoo/odoo/addons/base/ir/ir_http.py", line 195, in _dispatch
result = request.dispatch()
File "/.repo_requirements/odoo/odoo/http.py", line 827, in dispatch
r = self._call_function(**self.params)
File "/.repo_requirements/odoo/odoo/http.py", line 333, in _call_function
return checked_call(self.db, *args, **kwargs)
File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
return f(dbname, *args, **kwargs)
File "/.repo_requirements/odoo/odoo/http.py", line 329, in checked_call
result.flatten()
File "/.repo_requirements/odoo/odoo/http.py", line 1259, in flatten
self.response.append(self.render())
[...]

PR #938, opened 2017-08-16
[FIX][password_security] Fix internal error on wrong login

nguyenductamlhp:
Fix error when enter wrong password. `request.uid` is reset  to False and not reassign after authenticated.
[OCA#914](https://github.com/OCA/server-tools/issues/914)

### DIFF
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -45,8 +45,6 @@ class PasswordSecurityHome(AuthSignupHome):
             request.params['password']
         )
         if not uid:
-            request.uid = old_uid
-            request.uid = request.env.ref('base.public_user').id
             return response
         users_obj = request.env['res.users'].sudo()
         user_id = users_obj.browse(request.uid)
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -35,6 +35,7 @@ class PasswordSecurityHome(AuthSignupHome):
     @http.route()
     def web_login(self, *args, **kw):
         ensure_db()
+        old_uid = request.uid
         response = super(PasswordSecurityHome, self).web_login(*args, **kw)
         if not request.httprequest.method == 'POST':
             return response
@@ -44,6 +45,7 @@ class PasswordSecurityHome(AuthSignupHome):
             request.params['password']
         )
         if not uid:
+            request.uid = old_uid
             return response
         users_obj = request.env['res.users'].sudo()
         user_id = users_obj.browse(request.uid)
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -46,6 +46,7 @@ class PasswordSecurityHome(AuthSignupHome):
         )
         if not uid:
             request.uid = old_uid
+            request.uid = request.env.ref('base.public_user').id
             return response
         users_obj = request.env['res.users'].sudo()
         user_id = users_obj.browse(request.uid)
--- a/password_security/README.rst
+++ b/password_security/README.rst
@@ -83,6 +83,7 @@ Contributors
 
 * James Foster <jfoster@laslabs.com>
 * Dave Lasley <dave@laslabs.com>
+* Nguyen Duc Tam <nguyenductamlhp@gmail.com>
 
 Maintainer
 ----------


password_security/controllers/main.py
```diff
@@ -44,6 +45,7 @@ def web_login(self, *args, **kw):
             request.params['password']
         )
         if not uid:
+            request.uid = old_uid
```
lasley: This makes sense, because the public user is implicit when website comes in. I think this logic can be exploited though, so what if we maybe skip the `old_uid` part and simply:

```
request.uid = request.env.ref('base.public_user').id
```

lasley (changes requested):

password_security/controllers/main.py
nguyenductamlhp: Hello @lasley , can you explain more?
When website is loaded, current user will be public user, so `old_uid` will be `public_user`.
In the other hand, if last user  logged out, current user will be back to `public_user` again`.
Thank you!

lasley: @nguyenductamlhp - you're totally right, but in this code you're just assuming that's the case. We should instead just explicitly assign the `request.uid` to the public user, instead of removing it or assigning it to an unknown old uid. 

We must be very careful when dealing with authentication - this one variable is the key to the Odoo city. I haven't played around in it, but I can think of a few different ways that I would try to exploit this code as a foreign actor.

lasley: Oh and another good note - explicit is more Pythonic. Check out [PEP-20](https://www.python.org/dev/peps/pep-0020/):

> Explicit is better than implicit.

nguyenductamlhp: If we skip old_uid, we can not access `ir.model.data` to get public user because of security, it requires uid to get data. So i thank we should place `request.uid = request.env.ref('base.public_user').id` after `request.uid = old_uid`

lasley: Why not just `request.env.ref('base.public_user').sudo().id`?

nguyenductamlhp: How can we do like that? When we reference to `base.public_user`, it means we access to `ir.model.data`, and sql query will be  executed within access right checking. At that time, still no uid so it will cause error

lasley: @nguyenductamlhp - It looks to me like the code currently works, without the sudo. How can I replicate the security issues that you're referring to?

lasley: Oh I see, because the old/new UID trick is still in place as well. Ok so I checked out your branch and played - you're totally right. We can emulate the sudo using a similar trick as you, but just slightly more explicit:


```python
from odoo import http, SUPERUSER_ID
...
    @http.route()
    def web_login(self, *args, **kw):
        ensure_db()
        response = super(PasswordSecurityHome, self).web_login(*args, **kw)
        if not request.httprequest.method == 'POST':
            return response
        uid = request.session.authenticate(
            request.session.db,
            request.params['login'],
            request.params['password']
        )
        if not uid:
            # An intermediary uid is required to allow for lookup by reference
            request.uid = SUPERUSER_ID
            request.uid = request.env.ref('base.public_user').id
            return response
        users_obj = request.env['res.users'].sudo()
        user_id = users_obj.browse(request.uid)
        if not user_id._password_has_expired():
            return response
        user_id.action_expire_password()
        redirect = user_id.partner_id.signup_url
        return http.redirect_with_hash(redirect)
```

lasley:
@nguyenductamlhp - you should also add yourself to the contributors section in the ReadMe 😉

nguyenductamlhp:
There is  problem when we set `request.uid` as public user with following step:
1. Login with right user & password
2. Back to login page with back button of browser
3. Relogin again with wrong password
This will cause internal error.
After debuging, I saw that when we relogin (in step 3), user at that time is public user  because of wrong password. But it still redirect to menu and view of user in step 1, which public user is not allowed to access.
Can any one help me in this problem?

lasley:
@nguyenductamlhp - You're able to replicate this issue using the code I suggested in https://github.com/OCA/server-tools/pull/938#discussion_r134292878 ?

nguyenductamlhp:
@lasley  This error is not caused by intermediary uid. It occur because of the difference between uid in request and uid in website. This lead to ACLs error on model ir.ui.menu

lasley:
Hmmm yeah you're right @nguyenductamlhp. I just tested my solution from #936 & it seems that is still holding up. 

Unfortunately I deleted the branch which blocked me from reopening the PR, so I created #952 if you would like to test. I'm honestly not sure how to truly fix it without the glue module.

password_security/controllers/main.py
```diff
@@ -44,6 +45,8 @@ def web_login(self, *args, **kw):
...
+            request.uid = request.env.ref('base.public_user').id
```
lasley: @nguyenductamlhp - can you remove this line and we can just go with the original solution? I've performed an extensive audit and I think we're good here

nguyenductamlhp: @lasley, OK

password_security/controllers/main.py
```diff
@@ -45,8 +45,6 @@ def web_login(self, *args, **kw):
             request.params['password']
         )
         if not uid:
-            request.uid = old_uid
```
lasley: Isn't this part still needed for the fix?

Issue #908, opened 2017-07-25
[auth_totp][9.0] issue with firefox?


andhit-r:
I tried TOTP on runbot. Here are my steps:
1. I generate Authentication Apps/Devices for Demo user using FreeTOTP on android
2. I activate MFA enable
3. Log out
4. I log in. After confirming MFA code, internal server error shows.
5. I refresh the browser. Invalid MFA code warning showed.

I tried with firefox first because it is my default browser. Then i tried with chromium. It worked. Can someone confirm it?

Many thanks

dreispt:
Maybe the same issue as [this](https://github.com/OCA/server-tools/pull/703#issuecomment-315018876)?
cc @lasley @obulkin

lasley:
Hmmm it seems the error is coming from Werkzeug. Only reproducible in Firefox...

```
2017-08-16 15:10:39,554 167 INFO openerp_test werkzeug: 172.17.42.1 - - [16/Aug/2017 15:10:39] "GET /auth_totp/login?redirect=&mfa_login_token=LhN6KbeEGAUaz17XnRy9 HTTP/1.0" 200 -
2017-08-16 15:10:55,223 167 INFO openerp_test werkzeug: 172.17.42.1 - - [16/Aug/2017 15:10:55] "POST /auth_totp/login HTTP/1.0" 500 -
2017-08-16 15:10:55,245 167 ERROR openerp_test werkzeug: Error on request:
Traceback (most recent call last):
File "/usr/local/lib/python2.7/dist-packages/werkzeug/serving.py", line 177, in run_wsgi
execute(self.server.app)
File "/usr/local/lib/python2.7/dist-packages/werkzeug/serving.py", line 167, in execute
for data in application_iter:
File "/usr/local/lib/python2.7/dist-packages/werkzeug/wsgi.py", line 691, in __next__
return self._next()
File "/usr/local/lib/python2.7/dist-packages/werkzeug/wrappers.py", line 81, in _iter_encoded
for item in iterable:
TypeError: 'Response' object is not iterable
```

dreispt:
Well, my default browser is Firefox and my `auth_totp` tests were done with it.

grindtildeath:
I also get an Internal Server Error on Firefox, and get the same error in the logs using this module with odoo v10. However the login is successful as a refresh of server URL will display odoo main page.

obulkin:
[...]

PR #941, opened 2017-08-17
[9.0][FIX] auth_totp: Firefox support

obulkin:
* Fix Firefox MFA login error by adding logic that checks for the Firefox edge case where ``redirect_with_hash`` returns a ``Response`` object rather than a string
* Add test case for this scenario

### DIFF
--- a/auth_totp/controllers/main.py
+++ b/auth_totp/controllers/main.py
@@ -5,6 +5,7 @@
 from datetime import datetime, timedelta
 import json
 from werkzeug.contrib.securecookie import SecureCookie
+from werkzeug.wrappers import Response as WerkzeugResponse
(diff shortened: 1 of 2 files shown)


dreispt (approved):
Tested on Runbot - it works!

lasley (approved):
LGTM @obulkin thanks

Please amend your commit to include "fixes #908" so that the issue is properly closed with this one

https://help.github.com/articles/closing-issues-using-keywords/

obulkin:
@lasley - I've updated the commit message, but I don't think the feature will actually work here as the GitHub help page says that the target branch for the merge needs to be the default branch in the repo, and this is currently ``10.0``.

PR #945, opened 2017-08-22
[9.0][FIX] password_security: noupdate rule

chienandalu:
Record rules should be noupdate

cc @Tecnativa

### DIFF
--- a/password_security/__openerp__.py
+++ b/password_security/__openerp__.py
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '9.0.1.0.3',
+    'version': '9.0.1.1.0',
     'author': "LasLabs, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
Change module version

password_security/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '9.0.1.0.3',
+    'version': '9.0.1.0.4',
```
pedrobaeza: As this requires module update, it should be 9.0.1.1.0. You can squash already the commits as the history is very simple.

chienandalu:
@pedrobaeza Changes done

pedrobaeza (approved):

lasley (approved):

pedrobaeza:
Fast-tracking this.

PR #947, opened 2017-08-23
[10.0][FIX][date_range] Periods of one day should be allowed

MiquelRForgeFlow:
Forward port of https://github.com/OCA/server-tools/pull/921.

### DIFF
--- a/date_range/__manifest__.py
+++ b/date_range/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Date Range",
     "summary": "Manage all kind of date range",
-    "version": "10.0.1.0.1",
+    "version": "10.0.1.0.2",
     "category": "Uncategorized",
     "website": "https://odoo-community.org/",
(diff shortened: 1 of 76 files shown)


yajo (approved):

sebalix (approved):

pedrobaeza (commented):
Maybe you can change translations in mass in the same commit for avoiding to retranslate all in Transifex

MiquelRForgeFlow:
And what about translations in v9? 🤔

pedrobaeza:
Well, I didn't think it in that moment, but we have the occasion to make it right here in v10.

yajo:
I think those can be done in Transifex for v9 if you don't want to open a PR just for that

MiquelRForgeFlow:
Yeah, I won't do a PR for that in v9.

All :green_apple:

pedrobaeza (approved):

PR #948, opened 2017-08-23
[FIX][module_auto_update] Record base addon checksum

yajo:
Base addon is a special case: when getting updated, it is updated in a limited addons graph that only contains itself. Under such graph, no behavior modifications are respected, so no matter what we do in module_auto_update, it will get ignored in that step. This means that when we first install module_auto_update, it will record current base addon's dir hash, and it will never again be updated, so the next time your base addon is updated, the auto update will update all addons (as expected), and every next times too (not expected).

To fix that behavior, now we track the base addon status before and after running an upgrade. If it has changed, then we update its checksum.

Besides that, workarounds for https://github.com/odoo/odoo/pull/18597 are removed, and a new method for knowing if an addon is being installed, updated or uninstalled is used: check the existence of `latest_version` key in the create/write dict. This is because Odoo will always set that field to `False` when removing the addon, and it will also always write to it after installing or updating, so it provides basically all the information we need to know if we have to update or remove the checksum.

@Tecnativa

### DIFF
--- a/module_auto_update/models/module.py
+++ b/module_auto_update/models/module.py
@@ -41,31 +41,21 @@ class Module(models.Model):
                     "Cannot compute dir hash for %s, module not found",
                     r.display_name)
 
+    @api.multi
     def _store_checksum_installed(self, vals):
-        if self.env.context.get('retain_checksum_installed'):
-            return
+        """Store the right installed checksum, if addon is installed."""
         if 'checksum_installed' not in vals:
-            if vals.get('state') == 'installed':
-                for r in self:
-                    r.checksum_installed = r.checksum_dir
-            elif vals.get('state') == 'uninstalled':
+            try:
+                version = vals["latest_version"]
+            except KeyError:
+                return  # Not [un]installing/updating any addon
+            if version is False:
+                # Uninstalling
                 self.write({'checksum_installed': False})
-
-    @api.multi
-    def button_uninstall_cancel(self):
-        # TODO Use super() like in v10 after pull is merged
-        # HACK https://github.com/odoo/odoo/pull/18597
-        return self.with_context(retain_checksum_installed=True).write({
-            'state': 'installed',
-        })
-
-    @api.multi
-    def button_upgrade_cancel(self):
-        # TODO Use super() like in v10 after pull is merged
-        # HACK https://github.com/odoo/odoo/pull/18597
-        return self.with_context(retain_checksum_installed=True).write({
-            'state': 'installed',
-        })
+            else:
+                # Installing or updating
+                for one in self:
+                    one.checksum_installed = one.checksum_dir
 
     @api.model
     def create(self, vals):
--- a/module_auto_update/wizards/module_upgrade.py
+++ b/module_auto_update/wizards/module_upgrade.py
@@ -30,4 +30,12 @@ class ModuleUpgrade(models.TransientModel):
         # Compute updates by checksum when called in @api.model fashion
         if not self:
             self.get_module_list()
-        return super(ModuleUpgrade, self).upgrade_module()
+        # Get base adddon status before updating
+        base = self.env["ir.module.module"].search([("name", "=", "base")])
+        pre_state = base.state
+        result = super(ModuleUpgrade, self).upgrade_module()
+        # Update base addon checksum if its state changed
+        base.invalidate_cache()
+        if base.state != pre_state:
+            base.latest_version = base.latest_version
+        return result
--- a/module_auto_update/wizards/module_upgrade.py
+++ b/module_auto_update/wizards/module_upgrade.py
@@ -37,5 +37,9 @@ class ModuleUpgrade(models.TransientModel):
         # Update base addon checksum if its state changed
         base.invalidate_cache()
         if base.state != pre_state:
+            # This triggers the write hook that should have been triggered
+            # when the module was [un]installed/updated in the base-only
+            # module graph inside above call to super(), and updates its
+            # dir checksum as needed
             base.latest_version = base.latest_version
         return result
--- a/module_auto_update/__openerp__.py
+++ b/module_auto_update/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Module Auto Update',
     'summary': 'Automatically update Odoo modules',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'category': 'Extra Tools',
     'website': 'https://odoo-community.org/',
     'author': 'LasLabs, '
--- a/module_auto_update/tests/test_module.py
+++ b/module_auto_update/tests/test_module.py
@@ -10,6 +10,7 @@ import mock
 
 from openerp.modules import get_module_path
 from openerp.tests.common import TransactionCase
+from openerp.tools import mute_logger
 
 from .. import post_init_hook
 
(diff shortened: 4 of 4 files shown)


pedrobaeza:
Travis is failing

yajo:
Tests fixed

yajo:
💚 now

module_auto_update/wizards/module_upgrade.py
```diff
@@ -30,4 +30,12 @@ def upgrade_module(self):
...
             self.get_module_list()
-        return super(ModuleUpgrade, self).upgrade_module()
+        # Get base adddon status before updating
+        base = self.env["ir.module.module"].search([("name", "=", "base")])
+        pre_state = base.state
+        result = super(ModuleUpgrade, self).upgrade_module()
+        # Update base addon checksum if its state changed
+        base.invalidate_cache()
+        if base.state != pre_state:
+            base.latest_version = base.latest_version
```
pedrobaeza: Why don't you directly record the checksum instead of doing this weird reassignment?

module_auto_update/models/module.py
```diff
@@ -41,31 +41,21 @@ def _compute_checksum_dir(self):
...
     def _store_checksum_installed(self, vals):
-        if self.env.context.get('retain_checksum_installed'):
-            return
+        """Store the right installed checksum, if addon is installed."""
         if 'checksum_installed' not in vals:
-            if vals.get('state') == 'installed':
-                for r in self:
-                    r.checksum_installed = r.checksum_dir
-            elif vals.get('state') == 'uninstalled':
+            try:
```
pedrobaeza, line 48: I prefer the other approach:
```python
version = vals.get('latest_version')
if version is None:
    return
```

lasley: EAFP approach is more Pythonic in instances of key/attribute errors

module_auto_update/wizards/module_upgrade.py
lasley: Or maybe a comment indicating why weird assignment is necessary if this is some edge case?

lasley (approved):

module_auto_update/models/module.py
pedrobaeza: You need more lines and it's less readable (as you have to know what the exception means), so I prefer the other approach (and other Python gurus are saying the same lastly).

module_auto_update/wizards/module_upgrade.py
yajo: Explanation added.

module_auto_update/models/module.py
yajo: I also prefer EAFP

pedrobaeza: I don't and I won't put it on any of my code, but I won't block this time.

pedrobaeza (approved):

module_auto_update/models/module.py
lasley: Developer preference, but this exact use case is noted in the [Python docs](https://docs.python.org/2/glossary.html#term-eafp) so I don't feel that you can ever block for this without good reason.

I'm also pretty sure the py2to3 converts 

```
try:
    ...
except X:
    pass
```

to

```
with contextlib.suppress(X):
    ...
```

It will not do the same for the LBYL, which then becomes more lines and more resources.

pedrobaeza: The question is that I don't suggest `LBYL` (https://docs.python.org/2/glossary.html#term-lbyl), but this, which is not:
```python
version = vals.get('latest_version')
if version is None:
    return
```

lasley: Wouldn't this change the logic of the code? The `if version is False` looks pretty explicit, where a `None` would be falling into the `else`. Is this the intent @Yajo - were there any `None` when testing or?

pedrobaeza: No, the logic is the same: if no key is found, `None` is obtained in `version`, returning in the next condition, which is the same as in the `except:` part

lasley: This is true, however you are not considering there actually being a `None` in the `vals`. The logic as currently written will send us past the `KeyError` and into the `else` clause of the `if` because `None is not False`.

If we switch to the `.get` approach, we have no concept of whether the `None` came from the `vals` or was simply a missing key. In this instance, we immediately return instead of triggering the if/else below it.

pedrobaeza: That's not possible in ORM: it will never be a `None` as returned value. Any way, although you program bad your code, there's no problem with that.

lasley: Do we know where `vals` comes from off-hand? I personally don't, but am assuming it's from a create or write method. Why would ORM restrict what vals are entered into a dictionary?

pedrobaeza: Why are you trying to get an impossible case for my code? It works very well and doesn't mean any error even on the improbable case of somebody consciously wants to ruin it, but hey, if you have a big problem with that, just leave it!

lasley: I wasn't going for ruining it, I was just pointing out that this could have unintended side-effects outside of what we think. 

The `if version is False:` is an incredibly explicit if statement that makes me think @Yajo saw something in his testing where a False-type value can be in `vals` but not actually be a `False`. Maybe this is just a `"0"` version number; I'm not sure.

`None` is a False type value though - so while we may think it won't cause issues, this is why I asked @Yajo for clarification to the explicit if in https://github.com/OCA/server-tools/pull/948#discussion_r135314276

Honestly at this point I'd kind of like an inline comment to the edge case that the explicit if is covering. @Yajo can you clarify for us in a comment why we are being so explicit?

yajo: I wasn't trying to start a flamewar here. It's very easy:

- Odoo stores `False` for empty fields, **always**, so no need to check if there's `None` or whatever.
- If above point is true always, [it's even more in current case](https://github.com/odoo/odoo/blob/15cb84e0f77db350560aac9d32d947e919def128/openerp/addons/base/module/module.py#L472).
- What would happen if a module is defined with `"version": ""`? This is very unlikely, and if it ever happened, I'm pretty sure Odoo would fail or patch it with `"9.0"` or `"10.0"` automatically, but... _is it even worth it testing that? IMHO not._

So, if I wanted to use LBYL, the right code would be:

```python
# Option A, matching 100% current behavior
if "latest_version" not in vals:
    return
# Option B, not 100%, but side matches would not matter since they should theoretically never happen
if vals.get("latest_version") is None:
    return
```

...  but I didn't.

The fastest way to do it was as it is right now, and it uses EAFP because IMHO it's the pythonic way to do it.

However, there's one basic truth that we all have to acknowledge: **Sometimes there's more than 1 way to do the same thing right**. So, in case somebody else chooses a way that you wouldn't choose by yourself, but it's also a good way to get to the exact same result where the only difference is a matter of taste... why bother? **Hakuna matata!**

----

As a side note, kudos @lasley for finding out https://github.com/OCA/server-tools/pull/948#discussion_r135288685, a very nice addition to Python 3. I'm very eager to get it! ☺️

yajo:
5 days passed...

module_auto_update/models/module.py
pedrobaeza: https://github.com/OCA/server-tools/pull/948#discussion_r135288685 is another sign that they have found EAFP very cluttered in code :wink: but hey, as you have said, there are several ways of doing things. You won't see me a lot doing EAFP though, not at least on that form.

PR #949, opened 2017-08-24
[FIX] base_kanban_stage: fix wrong field referenced in base_kanban_abstract_view_kanban

tafaRU:


lasley (approved):

dreispt:
I disagree :-1:: 
The `user_id` is a special field that has special treatment is some places, such as the `mail` module.
It represents the responsible user, and that is a key concept for the kanban workflow.

tafaRU:
@dreispt, ok, however it's needed to change https://github.com/OCA/server-tools/blob/10.0/base_kanban_stage/views/base_kanban_abstract.xml#L53 and to modify slightly docstring at https://github.com/OCA/server-tools/blob/10.0/base_kanban_stage/models/base_kanban_abstract.py#L11
 
Are you ok with this proposal?

lasley:
TIL about the `user_id` special treatment. I wish there was some sort of documentation to these special fields.

@tafaRU - yeah I think the suggestion of modifying the view & updating the doc string is the best way to go then. Maybe even a note about the special treatment so someone can get a random TIL if they're in the code.

dreispt:
If I understood correctly, the current version references an unexisting field `kanban_user_id`.
Yes, it should be fixed to `user_id` and a mention added to the docstring.

dreispt:
@lasley Have a look at [`mail.thread`](https://github.com/odoo/odoo/blob/10.0/addons/mail/models/mail_thread.py) searching for `user_id`.

tafaRU:
@dreispt, @lasley, done.
I'll also modify the other PR, on 9.0 branch, once you'll have approved this.

Thank you in advance.

lasley (approved):
LGTM thanks @tafaRU

tafaRU:
Just made https://github.com/OCA/server-tools/pull/954

MiquelRForgeFlow (approved):

tafaRU:
@dreispt, could you please review?

PR #951, opened 2017-08-28
[7.0] ADD  base_readonly_user

eLBati:
This module allows to configure a user as 'readonly': the user will only be able to read data, without modifying


eLBati:
@Yajo fixed travis (for  base_readonly_user part)

PR #952, opened 2017-08-28
[10.0][ADD] website_password_security: Glue fix module

lasley:
* Add conditional logic in the event that the user is not in the env due to password security failure; fixes OCA#914

This is a resubmission of #936, which I can't reopen due to me deleting the branch. The proposed solution in #938 is running into issues, but the one here is still holding up. 

We'll need to decide which strategy to go with, because the glue module definitely isn't preferred. I'm not sure of how to fix https://github.com/OCA/server-tools/pull/938#issuecomment-324515293 though.

cc @nguyenductamlhp

### DIFF
--- a/website_password_security/README.rst
+++ b/website_password_security/README.rst
@@ -1,0 +1,51 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+=========================
+Website Password Security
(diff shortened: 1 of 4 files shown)


nhomar:
I am not totally against this module, but don't you think a well explained patch proposed to the core is better?

lasley:
I tested the shit out of this and #938 yesterday and I think we're good with the non-glue module approach. I was not able to circumvent security, although there is a minor issue that I won't talk about because nobody will probably notice. One of those bugs for under the rug.

PR #953, opened 2017-08-29
[FIX][module_auto_update] Always store changes in lower graphs

yajo:
The same problem that was fixed for the `base` addon in #948 happened with random addons that do not depend on `module_auto_update` (a.k.a. any addon) that Odoo decided to load before that one in the graph.

Now we always check for all addons if their state has changed, and make sure to trigger the udpate mechanism that stores the right value in `installed_checksum_dir` field.

@Tecnativa

### DIFF
--- a/module_auto_update/__openerp__.py
+++ b/module_auto_update/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Module Auto Update',
     'summary': 'Automatically update Odoo modules',
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.0.2',
     'category': 'Extra Tools',
     'website': 'https://odoo-community.org/',
     'author': 'LasLabs, '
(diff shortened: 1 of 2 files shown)


yajo:
This addon was being hit by several other problems as explained in last commit. Quoting myself from there:

> If you tried to uninstall the addon, you'd get a transaction lock that is now prevented by using autocommit cursor mode.
>
> If you installed and uninstalled the addon right away, you'd get a ProgrammingError saying that some columns exist no more. Checks are done now using `search_read`, which lets us limit the fields being fetched, and the environment is cleared to make sure nothing fails.
>
> Also we now guess if this own addon has been uninstalled and skip further logic if so, given it would hit broken triggers otherwise as it did before.

It should work as expected in every aspect now.

pedrobaeza (approved):
Tested in our own instance.

lasley (approved):
Awesome thanks @Yajo 

Interesting how many edge cases we ended up with here.

Tested in my own instance last night as well.

PR #955, opened 2017-08-29
[FIX] user_threshold: Don't create or commit new cursor

lasley:
* Remove new cursor creation/commit in favor of using current cursor

REF https://github.com/OCA/server-tools/pull/889#issuecomment-325610758

### DIFF
--- a/user_threshold/models/res_users.py
+++ b/user_threshold/models/res_users.py
@@ -21,7 +21,7 @@ class ResUsers(models.Model):
     )
 
     @api.model_cr
-    def _register_hook(self, pool, cr):
+    def _register_hook(self):
         """
         Override to check if env var to hide threshold configuration and
         reset the database state is set. If it is, run those actions
--- a/user_threshold/models/res_users.py
+++ b/user_threshold/models/res_users.py
@@ -6,7 +6,7 @@ import os
 from csv import reader
 from lxml import etree
 
-from odoo import SUPERUSER_ID, _, api, fields, models
+from odoo import api, fields, models, _
 from odoo.exceptions import AccessError, ValidationError
 
 from .ir_config_parameter import THRESHOLD_HIDE, MAX_DB_USER_PARAM
@@ -20,7 +20,7 @@ class ResUsers(models.Model):
         'Exempt User From User Count Thresholds',
     )
 
-    def __init__(self, pool, cr):
+    def _register_hook(self, pool, cr):
         """
         Override to check if env var to hide threshold configuration and
         reset the database state is set. If it is, run those actions
@@ -28,22 +28,15 @@ class ResUsers(models.Model):
         if THRESHOLD_HIDE:
             exempt_users_var = os.environ.get('USER_THRESHOLD_USER', '')
             exempt_users = reader([exempt_users_var])
-            with api.Environment.manage():
-                env = api.Environment(cr, SUPERUSER_ID, {})
-                installed = env['ir.module.module'].search_count([
-                    ('name', '=', 'user_threshold'),
-                    ('state', '=', 'installed'),
-                ])
-                if installed:
-                    users = env['res.users'].search([
-                        ('share', '=', False),
-                        ('threshold_exempt', '=', True),
-                    ])
-                    non_ex = users.filtered(
-                        lambda r: r.login not in exempt_users
-                    )
-                    for user in non_ex:
-                        user.threshold_exempt = False
+            users = env['res.users'].search([
+                ('share', '=', False),
+                ('threshold_exempt', '=', True),
+            ])
+            non_ex = users.filtered(
+                lambda r: r.login not in exempt_users
+            )
+            for user in non_ex:
+                user.threshold_exempt = False
 
     def _check_thresholds(self):
         """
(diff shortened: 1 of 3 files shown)


MiquelRForgeFlow (approved):

Garamotte:
Odoo has many hooks everywhere in the modules loading process, so that we can integrate where we want. The [`_register_hook`](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L4709-L4712) method is the [last called](https://github.com/odoo/odoo/blob/10.0/odoo/modules/loading.py#L415-L417) in the modules loading process, on a point where all modules are loaded, so the ORM is fully useable.

I think we should use the `_register_hook` method, instead of `__init__`. This method is too low-level, and should often not be overriden.
Another argument is that `__init__` is called once per module that inherits the model, and `_register_hook` only once per model. In both cases, this is done for each process.
We also don't need to check if the module is installed.

IMO, the right fix is:
```diff
diff --git a/user_threshold/models/res_users.py b/user_threshold/models/res_users.py
index b26151215b2..3f5a5bc7679 100644
--- a/user_threshold/models/res_users.py
+++ b/user_threshold/models/res_users.py
@@ -6,7 +6,7 @@ import os
 from csv import reader
 from lxml import etree
 
-from odoo import SUPERUSER_ID, _, api, fields, models, registry
+from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, ValidationError
 
 from .ir_config_parameter import THRESHOLD_HIDE, MAX_DB_USER_PARAM
@@ -20,7 +20,7 @@ class ResUsers(models.Model):
         'Exempt User From User Count Thresholds',
     )
 
-    def __init__(self, pool, cr):
+    def _register_hook(self):
         """
         Override to check if env var to hide threshold configuration and
         reset the database state is set. If it is, run those actions
@@ -28,24 +28,15 @@ class ResUsers(models.Model):
         if THRESHOLD_HIDE:
             exempt_users_var = os.environ.get('USER_THRESHOLD_USER', '')
             exempt_users = reader([exempt_users_var])
-            with api.Environment.manage():
-                with registry(cr.dbname).cursor() as new_cr:
-                    new_env = api.Environment(new_cr, SUPERUSER_ID, {})
-                    installed = new_env['ir.module.module'].search_count([
-                        ('name', '=', 'user_threshold'),
-                        ('state', '=', 'installed'),
-                    ])
-                    if installed:
-                        users = new_env['res.users'].search([
-                            ('share', '=', False),
-                            ('threshold_exempt', '=', True),
-                        ])
-                        non_ex = users.filtered(
-                            lambda r: r.login not in exempt_users
-                        )
-                        for user in non_ex:
-                            user.threshold_exempt = False
-                        new_cr.commit()
+            users = self.env['res.users'].search([
+                ('share', '=', False),
+                ('threshold_exempt', '=', True),
+            ])
+            non_ex = users.filtered(lambda r: r.login not in exempt_users)
+            for user in non_ex:
+                user.threshold_exempt = False
 
     def _check_thresholds(self):
         """
```

lasley:
Thanks @sylvain-garancher - I had no idea about `_register_hook` - good to know & updated!

Garamotte:
Oops, sorry, I forgot the decorator...
The `_register_hook` method must has an `@api.model_cr` decorator.

user_threshold/models/res_users.py
```diff
@@ -20,32 +20,24 @@ class ResUsers(models.Model):
         'Exempt User From User Count Thresholds',
     )
 
-    def __init__(self, pool, cr):
+    @api.model_cr
+    def _register_hook(self, pool, cr):
```
MiquelRForgeFlow: Why not simply `def _register_hook(self):`? We are in v10 api!

lasley: Annnnnnd now I have actually RTFM. This is why blindly copying code is probably a bad idea lmao

Garamotte: You should have copy-pasted my patch :D

Garamotte:
Travis still fails on `NameError: global name 'env' is not defined`.
Definitely, you should copy/paste my patch :)

Garamotte:
A PEP8 error you added in the last commit: `user_threshold/models/res_users.py:40:1: W293 blank line contains whitespace`.

And other lint errors not related to your changes:
```
user_threshold/README.rst:50: [E7901(rst-syntax-error), ]  Line block ends without a blank line.
user_threshold/README.rst:50: [E7901(rst-syntax-error), ]  Undefined substitution referenced: "------".
user_threshold/README.rst:62: [E7901(rst-syntax-error), ]  Unknown target name: "github issues `<https://github.com/oca/server-tools/issues>".
user_threshold/tests/test_res_company.py:18: [W7950(odoo-addons-relative-import), TestResCompany.test_fields_view_get] Same Odoo module absolute import. You should use relative import with "." instead of "openerp.addons.user_threshold"
user_threshold/tests/test_res_users.py:72: [W7950(odoo-addons-relative-import), TestResUsers.test_fields_view_get] Same Odoo module absolute import. You should use relative import with "." instead of "openerp.addons.user_threshold"
```

lasley:
Oops forgot about this. We should be 🍏  now!

pedrobaeza (approved):

PR #960, opened 2017-08-31
[10.0][IMP] module_auto_update: Update module list before getting

lasley:
* Update the module list before getting it in order to automatically resolve new dependencies

This is required if a pre-existing module is updated to depend on a module that does not exist in the database.

### DIFF
--- a/module_auto_update/wizards/module_upgrade.py
+++ b/module_auto_update/wizards/module_upgrade.py
@@ -22,10 +22,12 @@ class ModuleUpgrade(models.TransientModel):
     @api.multi
     def upgrade_module(self):
         """Make a fully automated addon upgrade."""
+        Module = self.env["ir.module.module"]
+        # Update the modules list to resolve new dependencies
+        Module.update_list()
         # Compute updates by checksum when called in @api.model fashion
         if not self:
             self.get_module_list()
-        Module = self.env["ir.module.module"]
         # Get every addon state before updating
         pre_states = {addon["name"]: addon["state"]
                       for addon in Module.search_read([], ["name", "state"])}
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Are you sure with this you are going to have auto-installed new dependencies?

lasley:
Normally they would just automatically install due to the dependency chain. The issue in the case of new modules is that they haven't been scanned into the Odoo database, so it instead of installing ignores every module in the dependency chain after it.

I'm admittedly planning on testing in production tonight, so we'll find out if it works!

pedrobaeza:
I think it's not that way if the dependent module is already installed. I think dependencies installation only happens when installing the module, not upgrading it, but let's see with a real test.

Garamotte:
The reality is between these two explanations :)
Sometimes, the new dependencies are installed, and sometimes, the new modules are not installed, and the update of the dependency chain is ignored.

I consider that as a bug in Odoo's core, but I never took the time to look deeper into that issue.
I think that this is a matter of "is the new dependency already known by Odoo before the update or not", but I haven't checked.

module_auto_update/wizards/module_upgrade.py
```diff
@@ -22,10 +22,12 @@ def get_module_list(self):
     @api.multi
     def upgrade_module(self):
         """Make a fully automated addon upgrade."""
+        Module = self.env["ir.module.module"]
+        # Update the modules list to resolve new dependencies
+        Module.update_list()
         # Compute updates by checksum when called in @api.model fashion
         if not self:
             self.get_module_list()
```
yajo, line 30: Shouldn't the addition go inside this method?

yajo (changes requested):

pedrobaeza:
Is this still applying?

yajo:
Indeed, and it's important.

yajo:
This makes no more sense since #1190

PR #962, opened 2017-09-05
[9.0][ADD] base_patch_models_mixin

LoisRForgeFlow:
Base Patch Models Mixin
=====================

This module adds a mixin class which is meant to simplify working with
patches on BaseModel or on abstract models like ``mail .thread``. If you just
change them, models created earlier will lack the attributes you add. Just
inherit from this mixin, it will check which existing models need your
changes and apply them.

---

Related to https://github.com/OCA/social/pull/161 discussion.

@hbrunn fixed the issue [here](https://github.com/OCA/social/blob/8.0/mail_follower_custom_notification/models/base_patch_models_mixin.py) and it came up again in the related module above. So, we decided to publish this module as a common fix for every current and possible future occurrence.

### DIFF
--- a/base_patch_models_mixin/models/base_patch_models_mixin.py
+++ b/base_patch_models_mixin/models/base_patch_models_mixin.py
@@ -9,8 +9,6 @@ from openerp import models
 _logger = logging.getLogger(__file__)
 
 
-# TODO: this should be a helper module to cetralize the point of failure
-# in case this introduces tacit bugs
 class BasePatchModelsMixin(models.AbstractModel):
     """
     This is a mixin class meant to simplify working with patches
--- a/base_patch_models_mixin/models/base_patch_models_mixin.py
+++ b/base_patch_models_mixin/models/base_patch_models_mixin.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from openerp import models
+
+
+_logger = logging.getLogger(__file__)
+
+
+# TODO: this should be a helper module to cetralize the point of failure
+# in case this introduces tacit bugs
+class BasePatchModelsMixin(models.AbstractModel):
+    """
+    This is a mixin class meant to simplify working with patches
+    on BaseModel or on abstract models like mail.thread.
+    If you just change them, models created earlier will lack the
+    attributes you add. Just inherit from this mixin, it will check
+    which existing models need your changes and apply them.
+
+    In your module, do something like
+
+    class MailThread(models.AbstractModel):
+        _name = 'mail.thread'
+        _inherit = ['base.patch.models.mixin', 'mail.thread']
+
+    in case you need to patch BaseModel, say
+
+    class BaseModel(models.BaseModel):
+        _name = 'my.unique.model.name'
+        _inherit = 'base.patch.models.mixin'
+
+    Your code will behave as if it was an inherited class of the class you pass
+    in the second parameter to _base_patch_models.
+    """
+    _name = 'base.patch.models.mixin'
+
+    def _base_patch_models(self, cr, our_class=None, parent_class=None):
+        """iterate through existing models to apply our changes there if
+        necessary"""
+        if self._name == BasePatchModelsMixin._name:
+            return
+        my_bases = self.__class__.__bases__
+        for i in range(max(len(my_bases) - 1, 0)):
+            if my_bases[i]._name == BasePatchModelsMixin._name:
+                our_class = my_bases[i - 1]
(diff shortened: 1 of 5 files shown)


base_patch_models_mixin/models/base_patch_models_mixin.py
```diff
@@ -0,0 +1,93 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from openerp import models
+
+
+_logger = logging.getLogger(__file__)
+
+
+# TODO: this should be a helper module to cetralize the point of failure
```
hbrunn: this can go away now

hbrunn (changes requested):
great you did it the complicated way!

please also add a simple test in the lines of https://github.com/OCA/server-tools/blob/8.0/field_rrule/tests/test_field_rrule.py#L42-L46

Just override mail.thread there and add some function, and after calling its register hook, this function should also exist on res.partner and the like

LoisRForgeFlow:
@hbrunn ok, I will look into those test whenever I have some extra spare time.

BTW, Thanks for the review/help in this topic!

JordiBForgeFlow:
Ok, this module almost caused a core meltdown on a production server. When this module is installed users got this error:

```
Odoo Server Error
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 650, in _handle_exception
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 687, in dispatch
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 323, in _call_function
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/service/model.py", line 118, in wrapper
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 316, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 966, in __call__
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/http.py", line 516, in response_wrap
    response = f(*args, **kw)
  File "/opt/ao-odoo-buildout/parts/odoo9/addons/calendar/controllers/main.py", line 67, in notify
    res = registry.get("calendar.alarm_manager").get_next_notif(cr, uid, context=context)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/api.py", line 250, in wrapper
  File "/opt/ao-odoo-buildout/parts/odoo9/addons/calendar/calendar.py", line 485, in get_next_notif
    partner = self.pool['res.users'].read(cr, SUPERUSER_ID, uid, ['partner_id', 'calendar_last_notif_ack'], context=context)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/api.py", line 250, in wrapper
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/addons/base/res/res_users.py", line 922, in read
    res = super(users_view, self).read(cr, uid, ids, other_fields, context=context, load=load)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/api.py", line 250, in wrapper
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/addons/base/res/res_users.py", line 314, in read
    return super(res_users, self).read(cr, uid, ids, fields=fields, context=context, load=load)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/api.py", line 250, in wrapper
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/models.py", line 3208, in read
    result = BaseModel.read(records, fields, load=load)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, **kwargs)
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/models.py", line 3237, in read
    if field.inherited and field.base_field.column:
  File "/opt/ao-odoo-buildout/parts/odoo9/openerp/fields.py", line 360, in __getattr__
    raise AttributeError(name)
AttributeError: base_field
```

_Edit by @lasley: multi-line traceback (sorry but I was having trouble reading it)_

hbrunn:
you use non-approved, non-tested code in a production instance? That's brave...

JordiBForgeFlow:
@hbrunn Yes, that was too brave in this case.... But we tested the module internally, and seemed to work properly.

LoisRForgeFlow:
The error seems to appear when using this module in conjunction with `view_sql_editor` and you try to *Generate BI View*. It is in some way related to this [line](https://github.com/OCA/reporting-engine/blob/9.0/bi_view_editor/models/ir_model.py#L323). However, this is issue is way over my level and I'm not able to find the real origin or the solution (if exists).

hbrunn:
see https://github.com/OCA/social/pull/161#issuecomment-665754284

PR #969, opened 2017-09-07
[FIX] equality

hbrunn:
Please
=======

read the thread. It's all there.

TLDR
====

summaries are harmful

### DIFF
--- a/base_suspend_security/README.rst
+++ b/base_suspend_security/README.rst
@@ -1,5 +1,6 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
     :alt: License: AGPL-3
+
 Suspend security
 =====================
 
--- a/base_suspend_security/__openerp__.py
+++ b/base_suspend_security/__openerp__.py
@@ -19,7 +19,7 @@
 ##############################################################################
 {
     "name": "Suspend security",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
     "author": "Therp BV,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Hidden/Dependency",
--- a/base_suspend_security/__openerp__.py
+++ b/base_suspend_security/__openerp__.py
@@ -20,19 +20,12 @@
 {
     "name": "Suspend security",
     "version": "8.0.1.0.0",
-    "author": "Therp BV",
+    "author": "Therp BV,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Hidden/Dependency",
     "summary": "Suspend security checks for a call",
     "depends": [
         'base',
     ],
-    "test": [
-    ],
-    "auto_install": False,
     "installable": True,
-    "application": False,
-    "external_dependencies": {
-        'python': [],
-    },
 }
--- a/base_suspend_security/base_suspend_security.py
+++ b/base_suspend_security/base_suspend_security.py
@@ -24,9 +24,14 @@ class BaseSuspendSecurityUid(int):
         return self
 
     def __eq__(self, other):
+        if isinstance(other, BaseSuspendSecurityUid):
+            return self * 1 == other * 1
         if isinstance(other, (int, long)):
             return False
         return super(BaseSuspendSecurityUid, self).__int__() == other
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
 
     def __iter__(self):
         yield super(BaseSuspendSecurityUid, self).__int__()
--- a/base_suspend_security/models/ir_model_access.py
+++ b/base_suspend_security/models/ir_model_access.py
@@ -27,6 +27,7 @@ class IrModelAccess(models.Model):
     @tools.ormcache_context(accepted_keys=('lang'))
     def check(self, cr, uid, model, mode='read', raise_exception=True,
               context=None):
+        # pylint: disable=old-api7-method-defined
         if isinstance(uid, BaseSuspendSecurityUid):
             return True
         return super(IrModelAccess, self).check(
--- a/base_suspend_security/tests/test_base_suspend_security.py
+++ b/base_suspend_security/tests/test_base_suspend_security.py
@@ -20,6 +20,7 @@
 from openerp import exceptions
 from openerp.tools import mute_logger
 from openerp.tests.common import TransactionCase
+from ..base_suspend_security import BaseSuspendSecurityUid
 
 
 class TestBaseSuspendSecurity(TransactionCase):
@@ -52,3 +53,11 @@ class TestBaseSuspendSecurity(TransactionCase):
         # this tests if _normalize_args conversion works
         self.env['res.users'].browse(
             self.env['res.users'].suspend_security().env.uid)
+        # check equality, that's relevant for picking the right environment
+        self.assertNotEqual(BaseSuspendSecurityUid(42), 42)
+        self.assertNotEqual(
+            BaseSuspendSecurityUid(42), BaseSuspendSecurityUid(43),
+        )
+        self.assertEqual(
+            BaseSuspendSecurityUid(42), BaseSuspendSecurityUid(42),
+        )


hbrunn:
(in case that's not obvious: This will improve speed of modules using this a *lot*)

pedrobaeza (changes requested):
Well, not very obvious... :wink: 

Please change module version number

hbrunn:
@pedrobaeza done. I made this a game for my colleagues until next week Sunday to come up with an explanation and pointers to code why this speeds up stuff, so please don't spoil it fellow reviewers :-)

pedrobaeza:
Hehe, OK, we can make this an OCA initiation ritual :stuck_out_tongue_closed_eyes:

pedrobaeza (approved):

hbrunn:
so my price is depending on preference a box of [probably the best chocolate pralines to be found in Amsterdam](https://puccinibomboni.com/?lang=en) or a bottle of the Pinot Noir they do in the region I grew up in. How about a company race? You too set out something for your colleagues until second Sunday, and maybe people like @gurneyalex @jbeficent @lasley are interested to join in?

hbrunn:
(I seriously just now realize that the title of this PR is `fix equality`. would it be so simple everywhere)

pedrobaeza:
Yes, why not? We love gamification!

lasley (approved):
Code LGTM. I'm in for the gamification too, good idea @hbrunn 

@bmw95 @hughesbm @obulkin - Bit difficult for me to get large items to you, but $50 USD Amazon gift card for the best answer to https://github.com/OCA/server-tools/pull/969#issuecomment-327859927 sent directly to my email so as to not spoil the fun for anyone else.

We can all compare our answers after the contest ends on Sunday the 17th

JordiBForgeFlow:
Ok, we can accept the challenge! We feed to the game an extra prize! A nice
bottle of best catalan wine to the winner! (To be delivered during OCA code
sprint).

cc @aheficent @lreficent @mreficent @rgarnau

On Sep 7, 2017 7:46 PM, "Holger Brunn" <notifications@github.com> wrote:

(I seriously just now realize that the title of this PR is fix equality.
would it be so simple everywhere)

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/OCA/server-tools/pull/969#issuecomment-327867452>, or mute
the thread

hbrunn:
...this seems to gain traction, awesome. How about @OSguard @nhomar @alexis-via @PCatinean @lmignon @legalsylvain @thomaspaulb joining too? I randomly ping github handles that come to my mind, so please spread the word to everyone I omitted purely for my own lack of spontaneousness. And please don't answer yourself, this is a riddle for your mid-level employees. Let's then the first person who receives a valid answer post here for the glory of the company (edit: just say first, don't give the answer), and all of us post the winning answers after the 17th 00:00 UTC?

hbrunn:
@astirpe and @elicoidal shouldn't miss the fun too. I hereby propose @StefanRijnhart as referee if he wants, which I hope

thomaspaulb:
@kiplangatdan @nzaterrence

StefanRijnhart:
@hbrunn it would be a pleasure. But do try the chocolates at http://www.pompadour.amsterdam though.

StefanRijnhart (approved):

rafaelbn:
Challenge accepted. Our price is a pack tasting of four varieties of artisan beer "[La Virgen](http://www.cervezaslavirgen.com/cervezas/)" from the village I grew up in.

lavirgen

Let's go @luismontalba @chienandalu @cubells @sergio-teruel @carlosdauden ! Take a look in https://github.com/OCA/server-tools/pull/969#issuecomment-327859927

thomaspaulb (approved):

hbrunn:
two hours to go

hbrunn:
So the link I expected was https://github.com/OCA/OCB/blob/8.0/openerp/api.py#L764

This class is what you get when you say `self.env`, and Environments [encapsulate](https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L5692) [the](https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L6078) [cache](https://github.com/OCA/OCB/blob/8.0/openerp/api.py#L771). I tried hard to have this comparison come out negative when we compare a [normal user ID](https://github.com/OCA/server-tools/pull/969/files#diff-fa12e01f01a0ef69243c1901999544e4L27) with our class that wraps a user ID. This is because we need the framework to read everything again with the wrapped `uid`, where the [other](https://github.com/hbrunn/server-tools/blob/02a4c86383906a64decc4661353129045cbe6664/base_suspend_security/models/ir_rule.py#L29) [manipulations](https://github.com/hbrunn/server-tools/blob/02a4c86383906a64decc4661353129045cbe6664/base_suspend_security/models/ir_model_access.py#L31) of the module take care that security rules are bypassed.

So a call like

```python
self.read()
self.suspend_security().read()
```

would read the data two times from the database, that's what we want.

But what happens when we say

```python
self.read()
self.suspend_security().read()
self.suspend_security().read()
```

?

Intuitively, the last `read` should be a no-op, because we've read this before with elevated permissions. But every time you say `.suspend_security`, you actually request a new environment (`.sudo` works the same way btw), and the constructor of the `Environment` class tries to recycle an existing one. And as the comparison in the very first link came out negative when comparing the same wrapped `uid`s (that's a bit hard to see, but in the end, we compared [instance references](https://github.com/OCA/server-tools/pull/969/files#diff-fa12e01f01a0ef69243c1901999544e4L29), and those differ when we wrap an `uid` multiple times, we always constructed a new environment. So actually, every call to `suspend_security` made the framework read everything again.

This turns out very badly if you have code like

```python
for this in self:
    this.suspend_security().do_something()
```

because then, everything is read anew for each iteration.

With this patch, we have made sure that [two instances of our class wrapping the same user ID compare as equal](https://github.com/OCA/server-tools/pull/969/files#diff-0f5ce34fb87d5a9d5e37242dd0e1adc8R61), so when the ORM looks for a matching environment, it will find one in the second call to `suspend_security` above.

How should you have approached this riddle? When a patch adds [another conditional](https://github.com/OCA/server-tools/pull/969/files#diff-fa12e01f01a0ef69243c1901999544e4R27), you know that exactly this condition is where things went south before. So the simplest way to understand what's going on is to put a breakpoint (`import pdb; pdb.set_trace()`) there and run some code. When the breakpoint hits, you say `bt` and then you'll see that the code is called by the comparison linked above. This will point you to reading [api.py](https://github.com/OCA/OCB/blob/8.0/openerp/api.py) and from there, you hop to [models.py](https://github.com/OCA/OCB/blob/8.0/openerp/models.py) where in the [_read_from_database](https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L3258) function, `Model`'s implementation of `$record['some_field']` is called, which in turn calls [the field's `__get__` function](https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L815) and that first tries to get the value from the cache, and reads it from the database otherwise. Programming is a very bookish reading based enterprise after all :-)

@StefanRijnhart, can you declare any winners? And can we merge this btw?

StefanRijnhart:
@hbrunn Merged! Thank you very much for the chrystal clear expanation. No need for a referee it seems, as I don't think I have received any entry from any contestant.

thomaspaulb:
I mailed mine to stefan@opener.am... Was that the wrong address?

Op 18 sep. 2017 07:23 schreef "Stefan Rijnhart (Opener)" <
notifications@github.com>:

sbidoul:
@hbrunn very nice explanation, thanks!

I've 2 questions:
- would it make sense to add a higher level test such as `assert self.suspend_security().env is self.suspend_security().env`?
- do you know if it applies to 9/10 too?

hbrunn:
1. definitely
2. yes, this should be forward ported

lasley:
>  No need for a referee it seems, as I don't think I have received any entry from any contestant.

Hah well none of my guys came up with an answer - seems maybe LasLabs needs some more training in the schedule!

pedrobaeza:
My colleagues have been as lazy as busy with other tasks, but they didn't dare to try an answer any way.

PR #973, opened 2017-09-10
Add new module DBF2CSV

felixvillafranca:
Add new module that allows convert a '.dbf' file to '.csv'


yajo:
Please review the bots messages and make them go 💚 😊

dbf2csv/controllers/controllers.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import http
+from odoo.http import request
+from odoo.addons.web.controllers.main import \
+    serialize_exception, content_disposition
+
+
+class Dbf2csvController(http.Controller):
+    @http.route('/web/binary/download_document', type='http', auth="public")
+    @serialize_exception
+    def download_document(self, filename, **kw):
```
naglis: This enables a [path traversal](https://www.owasp.org/index.php/Path_Traversal) vulnerability, where an attacker could gain access to files accessible to Odoo user by crafting a specific URL, eg. `/web/binary/download_document?filename=../etc/passwd`.

Maybe add an additional binary field for the converted file and allow user to download it once conversion is done?

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,48 @@
...
+    _description = 'File DBF to convert to CSV'
+
+    data_file = fields.Binary(
+        string='DBF file',
+        required=True,
+        help='Get you DBF file.')
+    filename = fields.Char(string='DBF file name')
+
+    @api.multi
+    def proccess_file(self):
```
naglis: [Typo](http://www.thesaurus.com/browse/process)

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,48 @@
...
+        # Process the DBF file chosen in the wizard,
+        # create and download CSV file.
+        self.ensure_one()
+        # Create DBF file in TMP
+        fp = open('/tmp/' + self.filename, 'w+')
```
naglis: I would suggest to avoid writting to external files (eg. by adding additional binary field, like I've mentioned before) as `self.filename` could be manipulate to overwrite system files writable by the Odoo user. 

At least this can be improved by using [`tempfile`](https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile), eg.:

```python
with tempfile.NamedTemporaryFile(mode='w+') as tmp_csv_file:
    # ...
```
It provides the benefit that it deletes the temporary file after it is closed (avoids spamming the file system) and is more failsafe - as it creates the file in a temporary directory (dependent on OS).

naglis: Also, IIUC `dbf.Dbf` [supports reading from a file-like object](http://dbfpy.bzr.sourceforge.net/bzr/dbfpy/annotate/head:/dbfpy/dbf.py#L96), so I would suggest to rewrite the module using in-memory file objects ([`io`](https://docs.python.org/2/library/io.html), [`(c)StringIO`](https://docs.python.org/2/library/stringio.html)).

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,48 @@
...
+        fp.write(base64.decodestring(self.data_file))
+        fp.close()
+
+        with open('/tmp/' + self.filename.replace('dbf', 'csv'),
```
naglis: Use [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join) to join path components and [`os.path.splitext`](https://docs.python.org/2/library/os.path.html#os.path.splitext) to change extension more safely ([example](https://stackoverflow.com/a/7363015/)).

dbf2csv/views/views.xml
```diff
@@ -0,0 +1,43 @@
...
+            <field name="res_model">dbftocsv.filedbf</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">form</field>
+            <field name="target">new</field>
+            <field name="view_id" ref="dbf_file_proccess_view"/>
+        </record>
+
+        <menuitem name="DBF to CSV"
+                  id="dbf2csv_parent_menu"
+                  parent = "base.menu_administration"
```
naglis: Extra spaces after `parent` and `=`.

dbf2csv/views/views.xml
```diff
@@ -0,0 +1,43 @@
...
+                  sequence="10"
+                  action="dbf_file_proccess"
+        />
+
+        <menuitem name="DBF to CSV"
+                  id="dbf2csv_child_menu"
+                  parent = "dbf2csv_parent_menu"
```
naglis: Extra spaces after `parent` and `=`.

dbf2csv/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    'summary': "Convert DBF file to CSV file",
+    'depends': [
+        'base',
+    ],
+    'data': [
+        'views/views.xml',
+    ],
+    'external_dependencies': {
+        'python': ['dbfpy']
```
naglis: Add `dbfpy` to [requirements.txt](https://github.com/OCA/server-tools/blob/10.0/requirements.txt)?

dbf2csv/controllers/controllers.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
naglis: Inconsistent copyright header - in other files you have used:
```
# Copyright 2017 INGETIVE (<https://ingetive.com>)
```

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,55 @@
...
+
+import csv
+import base64
+import logging
+from odoo import models, fields, api
+
+try:
+    from dbfpy import dbf
+except ImportError:
+    dbfpy = None
```
naglis: I suggest to log the import error as per [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror).

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
```
naglis: Transient models / views should go into the `wizards/` directory, as per [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#directories).

dbf2csv/views/views.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" ?>
```
naglis: Transient models / views should go into the `wizards/` directory, as per [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#directories).

dbf2csv/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 INGETIVE (<https://ingetive.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models,controllers
```
naglis: Missing space after comma.

felixvillafranca:
@Yajo Please, could you help me? what's wrong?

dbf2csv/controllers/controllers.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo import http
+from odoo.http import request
+from odoo.addons.web.controllers.main import \
+    serialize_exception, content_disposition
```
naglis: `serialize_exception` and `content_disposition` are defined in [_odoo/http.py_](https://github.com/odoo/odoo/blob/d23667262b3e0c14935f58cae793c0ed7b58bd08/odoo/http.py#L1639-L1649) [[2](https://github.com/odoo/odoo/blob/d23667262b3e0c14935f58cae793c0ed7b58bd08/odoo/http.py#L695-L732)] and not in `odoo.addons.web.controllers.main`.

👍 for use of `content_disposition`, nice helper!

dbf2csv/__manifest__.py
naglis: From [OCA idioms](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms):

> When making a comma separated list, dict, tuple, ... with one element per line, append a comma to the last element. This makes it so the next element added only changes one line in the changeset instead of changing the last element to simply add a comma.

dbf2csv/views/views.xml
```diff
@@ -0,0 +1,43 @@
...
+<odoo>
+        <record id="dbf_file_proccess_view" model="ir.ui.view">
+            <field name="name">Select DBF file to proccess</field>
+            <field name="model">dbftocsv.filedbf</field>
+            <field name="priority">1</field>
```
naglis: Is setting the priority necessary here?

dbf2csv/models/dbf2csv.py
```diff
@@ -0,0 +1,55 @@
...
+                names.append(field.name)
+            out_csv.writerow(names)
+            for rec in in_db:
+                out_csv.writerow(rec.fieldData)
+            in_db.close()
+
+        return {
+            'type': 'ir.actions.act_url',
+            'url': '/web/binary/download_document?filename=%s'
+            % (self.filename.replace('dbf', 'csv')),
```
naglis: Suggest to use [`urlencode`](https://docs.python.org/2/library/urllib.html#urllib.urlencode).

naglis (changes requested):
Nice work, although some improvements could be made.

naglis:
@felixvillafranca, I think you are missing `dbfpy` in [requirements.txt](https://github.com/OCA/server-tools/blob/10.0/requirements.txt).

dbf2csv/__init__.py
felixvillafranca: Ok, thanks.

dbf2csv/__manifest__.py
felixvillafranca: Ok, thanks

dbf2csv/controllers/controllers.py
felixvillafranca: Ok, thanks

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok, thanks

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok, thanks

dbf2csv/controllers/controllers.py
felixvillafranca: Ok, thanks

dbf2csv/controllers/controllers.py
felixvillafranca: Ok, thanks

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok, thanks

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok, thanks

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok, thanks

dbf2csv/views/views.xml
felixvillafranca: You're right; thanks

dbf2csv/views/views.xml
felixvillafranca: Ok, thnaks

dbf2csv/views/views.xml
felixvillafranca: Ok, thanks

dbf2csv/views/views.xml
felixvillafranca: Ok, thanks

dbf2csv/controllers/controllers.py
felixvillafranca: @nagils you're totally right, your way's better and safer so i'm going to redo it. Thanks.

dbf2csv/__manifest__.py
felixvillafranca: Ok, thanks.

felixvillafranca:
Hello;
Following instructions,
- The wizard has been changed for a model in order to store the files for downloading.
- Added Demo
- Added Tests

Now all is  :green_heart: :blush:

dbf2csv/demo/conversion_demo.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
+
+    <record id="demo_conversion_1" model="dbftocsv.conversion">
+        <field name="name">Demo conversion 1</field>
+        <field name="data_file" type="base64" file="dbf2csv/static/file/demo.dbf"/>
```
yajo: I guess the file should be in the `demo` directory, not in static...

dbf2csv/models/dbf2csv.py
```diff
@@ -2,45 +2,61 @@
...
+from odoo.exceptions import UserError
 
-class FileDBF(models.TransientModel):
-    _name = 'dbftocsv.filedbf'
+_logger = logging.getLogger(__name__)
+
+try:
+    from dbfpy import dbf
+except ImportError as err:
+    _logger.debug(err)
```
yajo: Debug logs usually pass unnoticed. Better `_logger.info("Cannot import dbfpy")`

dbf2csv/models/dbf2csv.py
```diff
@@ -2,45 +2,61 @@
...
+
+
+class Conversion(models.Model):
```
yajo: Does it really have sense for this to be a normal model? I think the source dbf file is useless once converted, so IMHO this should be an `AbstractModel` (A.K.A. wizard)

dbf2csv/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+"access_conversions_user","Manage dbf to csv conversions","model_dbftocsv_conversion","base.group_user",0,0,0,0
+"access_conversions_manager","Manage dbf to csv conversions","model_dbftocsv_conversion","base.group_system",1,1,1,1
```
yajo: Do you really need to be a sysadmin to  perform a file conversion?

dbf2csv/views/dbf2csv.xml
```diff
@@ -0,0 +1,67 @@
...
+        <field name="arch" type="xml">
+            <form string="Select DBF file">
+                <head>
+                    <button type="object"
+                            name="process_file"
+                            string="Convert to csv"
+                            class="oe_highlight"/>
+                </head>
+                <sheet>
+                    <h2> Upload a DBF file and convert to CSV</h2>
```
yajo: Remove the leading space

yajo (changes requested):

dbf2csv/views/dbf2csv.xml
felixvillafranca: Ok

dbf2csv/models/dbf2csv.py
felixvillafranca: Ok

dbf2csv/demo/conversion_demo.xml
felixvillafranca: Ok

dbf2csv/security/ir.model.access.csv
felixvillafranca: IMHO yes.
This module helps to convert SAGE .dbf files to .csv in order to import them later.
DBF files contain confidential information about the company: customers, products, pricelists, invoices, ... Therefore, from the moment they are stored in a model, it is necessary to protect them.

dbf2csv/models/dbf2csv.py
felixvillafranca: Initially it started being a wizard, but IMHO could be useful to have them stored:
* Files availables anytime, anywhere
* Relation between dbf and csv

yajo: But I see no use case on that: you convert it, download the csv, and then you store the csv in your computer wherever you want it. If you want to download it again, you just repeat the process, isn't it?

felixvillafranca: Ok, :pencil:

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,71 @@
...
+
+    data_file_csv = fields.Binary(
+        string='CSV file',
+        readonly=True,
+        help='Get you CSV file.')
+    filename_csv = fields.Char(string='CSV file')
+
+    def reopen_wizard(self):
+        return {
+            'type': 'ir.action.act_windows',
```
naglis: Typo, should be `ir.actions.act_window`.

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,71 @@
...
+            'name': 'Conversion',
+            'res_model': 'dbftocsv.conversion',
+            'res_id': self.id,
+            'view_type': 'form',
+            'view_mode': 'form',
+            'target': 'new'}
```
naglis: [OCA guideline](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#idioms): When making a comma separated list, dict, tuple, ... with one element per line, append a comma to the last element. This makes it so the next element added only changes one line in the changeset instead of changing the last element to simply add a comma.

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def process_file(self):
+        try:
```
naglis: I think this can be written without writing to the filesystem by using in memory buffers (eg. [cStringIO](https://docs.python.org/2/library/stringio.html)), for example:
```python
def process_file(self):
    dbf_buf = cStringIO.StringIO(base64.b64decode(self.data_file))
    csv_buf = cStringIO.StringIO()
    writer = csv.writer(csv_buf)
    with contextlib.closing(dbf.Dbf(dbf_buf)) as in_db:
        writer.writerow([f.name for f in in_db.header.fields])
        for rec in in_db:
            writer.writerow(rec.fieldData)
    self.write({
        'data_file_csv': base64.b64encode(csv_buf.getvalue()),
        'filename_csv': os.path.splitext(self.filename)[0] + '.csv',
    })
    return self.reopen_wizard()
```

dbf2csv/wizard/dbf2csv.xml
```diff
@@ -0,0 +1,57 @@
...
+                        </group>
+                        <group string="CSV">
+                            <field name="data_file_csv" filename="filename_csv" readonly="1"/>
+                            <field name="filename_csv" invisible="1"/>
+                        </group>
+                    </group>
+                    <footer>
+                        <button type="object"
+                            name="process_file"
+                            string="Convert to csv"
```
naglis: Nitpick: I think _CSV_ should be uppercased.

naglis (commented):
Looking better. What do you think about using in-memory buffers instead of writting the files to disk as part of processing?

dbf2csv/wizard/dbf2csv.xml
felixvillafranca: Ok, :memo:

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok, :memo:

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok, :memo:

dbf2csv/wizard/dbf2csv.py
felixvillafranca: :+1:  Cool

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class Conversion(models.TransientModel):
+    _name = 'dbftocsv.conversion'
+    _description = 'File DBF to convert to CSV'
+
+    data_file = fields.Binary(
+        string='DBF file',
+        required=True,
+        help='Get you DBF file.')
```
yajo: This is not correctly spelled, and also not very helpful. Try with *The DBF file that you want to transform* for example.

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+    filename = fields.Char(string='DBF file')
```
yajo: Better: `filename_dbf`

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+    from dbfpy import dbf
+except ImportError as err:
+    _logger.info(err)
+
+
+class Conversion(models.TransientModel):
+    _name = 'dbftocsv.conversion'
+    _description = 'File DBF to convert to CSV'
+
+    data_file = fields.Binary(
```
yajo: Following your own convention, I'd call this `data_dbf`

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+        string='DBF file',
+        required=True,
+        help='Get you DBF file.')
+    filename = fields.Char(string='DBF file')
+
+    data_file_csv = fields.Binary(
```
yajo: Better: `data_csv`

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+        string='CSV file',
+        readonly=True,
+        help='Get you CSV file.')
```
yajo: Again not correctly spelled. Maybe better *CSV file transformed and ready to download*

dbf2csv/wizard/dbf2csv.py
```diff
@@ -0,0 +1,65 @@
...
+                writer.writerow([f.name for f in in_db.header.fields])
+                for rec in in_db:
+                    writer.writerow(rec.fieldData)
+            self.write({
+                'data_file_csv': base64.b64encode(csv_buf.getvalue()),
+                'filename_csv': os.path.splitext(self.filename)[0] + '.csv',
+            })
+            return self.reopen_wizard()
+        except ValueError:
+            raise UserError(_("Unable to convert to csv"))
```
Yajo: It would be good if you add a call like `_logger.error("Could not convert to CSV", exc_info=True)` above this, just to make the admin be able to read the logs and know what actually failed.

yajo (changes requested):

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

dbf2csv/wizard/dbf2csv.py
felixvillafranca: Ok

yajo (changes requested):
Cool! Now could you please add some tests?

You can provide a dbf sample, run the converter, and check its output.

PS: No need to answer "OK" to every inline comment, just push the change and the comment disappears automatically 😉

PR #976, opened 2017-09-11
10.0 typo - base_search_fuzzy

gdgellatly:
@lasley satisfy my OCD please, can we squash the commit? in my excitement I forgot to pull before commit.

### DIFF
--- a/base_search_fuzzy/__manifest__.py
+++ b/base_search_fuzzy/__manifest__.py
@@ -6,7 +6,7 @@
     'name': "Fuzzy Search",
     'summary': "Fuzzy search with the PostgreSQL trigram extension",
     'category': 'Uncategorized',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.1.0',
     'website': 'https://odoo-community.org/',
     'author': 'bloopark systems GmbH & Co. KG, '
               'Eficent, '
--- a/base_search_fuzzy/models/trgm_index.py
+++ b/base_search_fuzzy/models/trgm_index.py
@@ -31,7 +31,7 @@ class TrgmIndex(models.Model):
         readonly=True,
         help='The index name is automatically generated like '
              'fieldname_indextype_idx. If the index already exists and the '
-             'index is located in the same table then this index is resused. '
+             'index is located in the same table then this index is reused. '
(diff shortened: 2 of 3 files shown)


lasley:
@gdgellatly - Go for it! Less commits, the better. Try not to amend/squash things that someone has reviewed though. Everything else, squashing is totally the best policy unless you're testing and expect to need to revert things.

Also, mergers will typically use the Github Squash & Merge feature as long as the PR is all one author and doesn't contain multiple migrations.

image

lasley:
Oh I see, best in this case would instead be to:
```
git reset --hard 379da5218824a45f3f15eb02fd840a9ae9f1224d
git rebase OCA/10.0
git push -f
```

gdgellatly:
@lasley thanks, I got the idea even if my remotes are named differently.  Indeed, perhaps I should use origin for my own branches, just I never expect to make a PR.

lasley (approved):
Thanks @gdgellatly

pedrobaeza:
Please change module version according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers (this requires a module update).

gdgellatly:
@pedrobaeza which one of x.y.z? , I'm just about to leave office for a couple of days so I'm going to guess y. A module upgrade is needed, although it could be z as it's not like anything breaks without an upgrade and it isn't a new feature.  Or it could be x as a view has changed.

Fun.  I'll just do y.  It can wait if I'm wrong.

pedrobaeza:
Indeed `y` is the correct one IMO.

gdgellatly:
@pedrobaeza OK I did it.  I suspect I might have messed it as the force push has changed the SHA when it usually doesn't.  If its a problem let me know and I'll sort on Friday.

EDIT: OK Refreshed screen.  Looks better

pedrobaeza:
No, it's OK for me. Thanks.

I'm remembering... how was your surgery? I suppose all goes well.

pedrobaeza (approved):

gdgellatly:
@pedrobaeza thanks, it's fine, all minor stuff. One to go and I'm done.

mFlayyan (approved):

PR #980, opened 2017-09-14
[FIX] deal with limit=0/None correctly

hbrunn:
this fixes string searches for many2one fields: https://github.com/OCA/OCB/blob/8.0/openerp/osv/expression.py#L1062

### DIFF
--- a/base_name_search_improved/__openerp__.py
+++ b/base_name_search_improved/__openerp__.py
@@ -4,7 +4,7 @@
 {
     'name': 'Improved Name Search',
     'summary': 'Friendlier search when typing in relation fields',
-    'version': '8.0.1.0.1',
+    'version': '8.0.1.0.2',
     'category': 'Uncategorized',
     'website': 'https://odoo-community.org/',
     'author': 'Daniel Reis, Odoo Community Association (OCA)',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change module version number, but OK with the patch. I had something similar in https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_partner/models/l10n_es_partner.py#L168

hbrunn:
@pedrobaeza done

pedrobaeza (approved):

sebalix (approved):

base_name_search_improved/__openerp__.py
```diff
@@ -4,7 +4,7 @@
 {
     'name': 'Improved Name Search',
     'summary': 'Friendlier search when typing in relation fields',
-    'version': '8.0.1.0.1',
```
gfcapalbo, line 7: We reviewed the module,  now in case of limit false behaves as one would expect.
Why change the version name? There are no migration scripts.

pedrobaeza: The module version number doesn't change because of migration scripts, but to note that this is something different, so you can say, version 8.0.1.0.1 fails because of that error, but check if you have version x that doesn't have that problem. See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers for reference.

gfcapalbo:
:+1:

mFlayyan:
:+1:

PR #988, opened 2017-09-17
[ADD] base_import_odoo

hbrunn:
Import from remote Odoo database
================================

This module was written to import data from another Odoo database. The idea is that you define which models to import from the other database, and add eventual mappings for records you don't want to import.

Use cases
=========

- merging databases
- one way sync (needs a bit polishing)
- aggregating management data from distributed systems

Configuration
=============

Go to Settings / Remote Odoo import / Import configurations and create a configuration.

After filling in your credentials, select models you want to import from the remote database. If you only want to import a subset of the records, add an appropriate domain.

The import will copy records of all models listed, and handle links to records of models which are not imported depending on the existing field mappings. Field mappings to local records also are a stopping condition. Without those, the import will have to create some record for all required x2x fields, which you probably don't want.

Probably you'll want to map records of model `res.company`, and at least the admin user.

The module doesn't import one2many fields, if you want to have those, add the model the field in question points to to the list of imported models, possibly with a domain.

If you don't fill in a remote ID, the addon will use the configure local ID for every record of the model (this way, you can for example map all users in the remote system to some import user in the current system).

For fields that have a uniqueness constraint (like `res.users#login`), set the flag `unique`, then the import will generate a unique value for this field.

Usage
=====

- go to an import configuration and hit the button ``Run import``
- be patient, this creates a cronjob which will start up to a minutes afterwards


andreschenkels:
I will test and review this

.travis.yml
```diff
@@ -33,6 +33,7 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
   - printf '[options]\n\nrunning_env = dev\n' > ${HOME}/.openerp_serverrc
+  - pip install erppeek
```
lasley: Why not requirements file?

Also, any thoughts to [`OdooRPC`](https://pypi.python.org/pypi/OdooRPC/) instead? It's much better IMO, and maintained by a community member (shout out @sebalix)

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+try:
+    from erppeek import Client
+except:
+    pass
```
lasley: Please add a logger indicating the lack of library (["Errors should never pass silently."](https://www.python.org/dev/peps/pep-0020/))\

Also can we catch only `ImportError` please

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+import psycopg2
+import traceback
+from openerp import _, api, exceptions, fields, models, tools
+from collections import namedtuple
+
+
+import_context_tuple = namedtuple(
```
lasley: Are these important at all for customization in child modules? I spent a bit trying to figure out if they are and needed docblocks, but I'm not totally sure?

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+            remote_counts[model.model] = len(remote_ids[model.model])
+        self.write({
+            'status_data': {
+                'counts': remote_counts,
+                'ids': remote_ids,
+                'error': None,
+                'done': {},
+            }
+        })
+        if commit and not tools.config['test_enable']:
```
lasley: > `tools.config['test_enable']`

Odoo uses [`threading.currentThread().testing`](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=%22threading.currentThread%28%29.testing+%3D%3D+%22&type=) for this usually. Is there a reason we're doing something different here? (real question)

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+                value = self._run_import_get_record(
+                    new_context,
+                    self.env[model._fields[name].comodel_name],
+                    {'id': record.get(name, [None])[0]},
+                )
+            elif field.type in ['selection'] and not callable(field.selection):
+                value = field.selection[0][0]
+            elif field.type in ['selection'] and callable(field.selection):
+                value = field.selection(model)[0][0]
+            # TODO: support more types, refactor to one function per type
```
lasley: Please remove comment and add to roadmap instead

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+        self, context, model, record, forcecreate=False,
+    ):
+        """Either misuse some existing record or create an empty one to satisfy
+        required links"""
+        dummy = model.search([
+            (
+                'id', 'not in',
+                [
+                    v for (model_name, remote_id), v
+                    in context.dummies.iteritems()
```
lasley: Please switch`.iteritems` to `.items`. I don't think that anything here would cause a performance impact between the two, and might as well get into the habit with Py3 coming to town.

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+        })
+        if commit and not tools.config['test_enable']:
+            # pylint: disable=invalid-commit
+            self.env.cr.commit()
+        for model_line in self.import_line_ids:
+            model = self.env[model_line.model_id.model]
+            done[model._name] = 0
+
+            for start_index in range(
+                    len(remote_ids[model._name]) / commit_threshold + 1
```
lasley: Might want to validate that `commit_threshold != -1` in order to avoid zero division errors. Not totally sure where it's coming from though, so maybe impossible case?

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+        remote = self._get_connection()
+        self.write({'password': False})
+        if commit and not tools.config['test_enable']:
+            # pylint: disable=invalid-commit
+            self.env.cr.commit()
+        for model_line in self.import_line_ids:
+            model = model_line.model_id
+            remote_ids[model.model] = remote.search(
+                model.model,
+                tools.safe_eval(model_line.domain) if model_line.domain else []
```
dreispt: A common need is to fetch only the records change since the last successful sync. It would be nice to add support for that.

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+                    error = traceback.format_exc()
+                    self.env.cr.rollback()
+                    self.write({
+                        'status_data': dict(self.status_data, error=error),
+                    })
+                    # pylint: disable=invalid-commit
+                    self.env.cr.commit()
+                    raise
+                done[model._name] += len(ids)
+                self.write({'status_data': dict(self.status_data, done=done)})
```
dreispt, line 162: Why not skip the inner `for start_index` loop, and instead add here an `if` with a checkpoint condition to enter the status data & commit section only when multiples of `commit_threshold` are reached? The effect is the same but (IMO) the code would be simpler and more readable.

base_import_odoo/models/import_odoo_database_model.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields, models
+
+
+class ImportOdooDatabaseModel(models.Model):
+    _name = 'import.odoo.database.model'
+    _description = 'A model to import from a remote database'
+    _order = 'sequence'
```
dreispt, line 10: Yes, this is very important.

base_import_odoo/models/import_odoo_database_model.py
```diff
@@ -0,0 +1,20 @@
...
+
+    sequence = fields.Integer()
+    model_id = fields.Many2one(
+        'ir.model', string='Model', required=True, ondelete='cascade',
+    )
+    database_id = fields.Many2one(
+        'import.odoo.database', string='Database', required=True,
+        ondelete='cascade',
+    )
+    domain = fields.Char(help='Optional filter to import only a subset')
```
dreispt, line 20: It would be nice to allow using a placeholder for the last sync date.

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+                    self.env.cr.execute(
+                        'select id from "%s" where id=%%s for update nowait' %
+                        self.env['ir.cron']._table,
+                        (this.cronjob_id.id,), log_exceptions=False,
+                    )
+            except psycopg2.OperationalError:
+                this.cronjob_running = True
+
+    @api.multi
+    def _create_cronjob(self):
```
dreispt, line 622: Yes, - one cron per remote.

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+                    data[field_name] = None
+            else:
+                data[field_name] = [(6, 0, data[field_name])]
+        for mapping in self.import_field_mappings:
+            if mapping.model_id.model != model._name or not mapping.fields_id:
+                continue
+            if mapping.unique:
+                value = data.get(mapping.fields_id.name, '')
+                counter = 1
+                while model.with_context(active_test=False).search([
```
dreispt: Yes - `active_test=False`

base_import_odoo/models/import_odoo_database.py
```diff
@@ -0,0 +1,472 @@
...
+                    )
+        if not _id:
+            xmlid = self.env['ir.model.data'].search([
+                ('import_database_id', '=', self.id),
+                ('import_database_record_id', '=', record['id']),
+                ('model', '=', model._name),
+            ], limit=1)
+            if xmlid:
+                _id = xmlid.res_id
+                context.idmap[(model._name, record['id'])] = _id
```
dreispt, line 279: Yes - XMLIds FTW

dreispt (approved):
Code review only.
It is an important feature and it looks excellent.
I thought about a few details but Holger seems to have got them all right.

dreispt:
Thank you for this feature Holger.

I'm considering using this for Odoo version migrations, for the cases where it is viable to start from an empty db and just migrate the data you need.

It also opens a door to the "federated Odoo instances" concept.

base_import_odoo/models/import_odoo_database.py
hbrunn: sure. but I'd also like a mode where we can update all again (should look at the modification dates then which one wins), then we can use this for two way sync

base_import_odoo/models/import_odoo_database.py
hbrunn: good point, will add that

hbrunn:
thanks for the review, I'll pick up your proposals later. for the migration part you'll probably have to massage the data a bit if the data model changed, but fortunately you can override `_run_import_map_values`, inspect the import context which model we're busy with at the moment, and change what needs to be changed. Should be quite straightforward I guess.

lasley (commented):
Just realized I did half a code review but didn't submit it. Submitting now, some may duplicate Dan's  stuff (I didn't check)

elicoidal (approved):
no test.

dreispt:
>  you'll probably have to massage the data a bit if the data model changed, but fortunately you can override _run_import_map_values,

Good point, thanks for the tip.

.travis.yml
hbrunn: since this is OCAized of course, when I had to made a choice I didn't really care one way or another, as both are just fine for what I need here

base_import_odoo/models/import_odoo_database.py
hbrunn: if some depending module want to do something radically different, they'll probably want to change those too. But then we'll have to refactor creation of those instances into some function too

base_import_odoo/models/import_odoo_database.py
hbrunn: now as I'm following up on the comments, no, not a good idea. The inner loop is also meant to do the import in blocks (as in: Read a block of data from remote), so we can't really get rid of this loop

base_import_odoo/models/import_odoo_database.py
hbrunn: those are pretty much equivalent, as this flag is only set in cases where `--test-enable` is passed: https://github.com/OCA/OCB/blob/10.0/odoo/modules/loading.py#L181

Other parts of the code (this is where I learned that in the first place) look at the config: https://github.com/OCA/OCB/blob/10.0/odoo/service/server.py#L272

hbrunn:
@dreispt @lasley comments attended or answered

The customer requested importing attachments too, for this I'll have to come up with some general treatment of a model + id field tuple, this will also be helpful for mail.message et al. I'm planning to add this tomorrow and also switch to https://github.com/OCA/odoorpc

lasley (approved):
Thanks @hbrunn

lasley:
I see some rebuttals to your comments @dreispt, so I'll hold for you to do the honors once confirmed

hbrunn:
travis is fixed in #1019

hbrunn:
@dreispt @elicoidal @lasley this turned out to be much more complex than I estimated, but finally got it right. The customer I wrote this for has quite elaborate dependencies between partners, employees, contracts and users, which by now can be imported without a single line of custom code, just configuration of the proper mappings and domains.
Maybe one of you can review #1019, merge it and trigger a rebuild here to verify that the tests now cover  the create and cleanup dummy functions

lasley:
>  merge it and trigger a rebuild here 

Done

hbrunn:
thanks! I feel haunted by [unrelated test failures](https://travis-ci.org/OCA/server-tools/jobs/289143707#L1519-L1544) here. should be fixed by https://github.com/OCA/server-tools/pull/1030

lasley (approved):
Changes LGTM. 

@hbrunn - you want to squash your own commits to preserve your GPG?

hbrunn:
yes. will do when a few days have passed without fixup commits...

PR #994, opened 2017-09-20
[IMP] Make groups readonly in the user form if roles are selected

StefanRijnhart:
As it says in this module's README: `you can not set groups manually on a user as long as there is roles configured on it`. This was only true in the backend, where manual modifications were ignored upon saving them. This change also makes the group selection fields and checkboxes on the user form readonly when there are roles configured.

@sebalix what do you think?

### DIFF
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -65,6 +65,7 @@ Contributors
 ------------
 
 * Sébastien Alix <sebastien.alix@osiell.com>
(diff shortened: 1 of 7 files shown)


sebalix:
@StefanRijnhart that makes sense :+1:

StefanRijnhart:
Tests are now fixed.

sebalix (approved):

lasley (approved):

Issue #939, opened 2017-08-16
auth_totp + password_security - TOTP needs to come first


lasley:
When `auth_totp` and `password_security` modules are both installed on a server and the user needs a password change, the password change is required before MFA has been validated. 

This is a potential security hole that would allow an unauthorized user to change the password of an account without access to the MFA device. It's no more of a hole than is already existing in core though, because the valid password is still required.

hbrunn:
this is one of the use cases I had in mind for https://github.com/OCA/server-tools/pull/876
You can simply depend on this module and conditionally on `auth_totp` in `password_security`, then you're done. Or, what's nicer, implement my proposal of conditional reverse dependencies in this PR, then you can add this to `auth_totp`'s manifest where it belongs imho

lasley:
Exactly what I needed, thanks @hbrunn! I agree `rdepends_if_installed` is the best way to go here, so I'll look into implementing. Were you planning on forward porting the module or should I also plan on that?

hbrunn:
glad it's useful. I fixed your comment and rebased, so you could do the honors if you like.

At the moment, I don't have plans for forward porting, so go ahead if you need it. And yes, given you guys always hammer on LGPL stuff, I won't complain if you change the license in the process ;-) Shouldn't be so much trouble, this part of the code didn't change a lot in the past years.

lasley:
> do the honors if you like.

Woot I'll hit the button once we get a CI 🍏 !

> At the moment, I don't have plans for forward porting, so go ahead if you need it.

Sweet, I'll tackle the 9 and 10 then. Doing the migrations should get me deep enough into the operation of the code so that I have a better idea of how to implement the reverse depends.

> And yes, given you guys always hammer on LGPL stuff, I won't complain if you change the license in the process ;-) 

Hooray! 😆  that was admittedly going to be my next question

> Shouldn't be so much trouble, this part of the code didn't change a lot in the past years.

This was my thought as well - here's to hoping for a 1:1 conversion 🍻

PR #996, opened 2017-09-20
[10.0][ADD][FIX] auth_totp and password_security compatibility

obulkin:
[FIX] ``password_security``: ``auth_totp`` compatibility
* Modify overloaded ``web_login`` action to not trigger a password expiration check if the login process is not complete yet (e.g. due to ``auth_totp``)
* Add unit test for new logic
* Fix warning caused by unrelated unit test

[ADD] ``auth_totp_password_security``: Compatibility
* Overload ``mfa_login_post`` action introduced by ``auth_totp`` so that password expiration check from ``password_security`` still happens during an MFA login
* Add unit tests for new logic

This PR addresses https://github.com/OCA/server-tools/issues/939

### DIFF
--- a/auth_totp_password_security/README.rst
+++ b/auth_totp_password_security/README.rst
@@ -1,0 +1,60 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+=======================================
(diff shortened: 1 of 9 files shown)


lasley (approved):
Thanks @obulkin

lasley:
@elicoidal - may want to look at this seeing as you use both modules (I think)

MiquelRForgeFlow (approved):

PR #998, opened 2017-09-27
[10.0][FIX] base_external_dbsource_firebird: add columns on execute_fdb

enriquemartin:
When making a query to dbsource, we need the column names for making an INSERT. Otherwise a BAD QUERY error is triggered.

### DIFF
--- a/base_external_dbsource_firebird/__manifest__.py
+++ b/base_external_dbsource_firebird/__manifest__.py
@@ -4,7 +4,7 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 {
     'name': 'External Database Source - Firebird',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Tools',
     'author': "Daniel Reis, "
               "LasLabs, "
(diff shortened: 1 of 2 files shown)


lasley (changes requested):
@enriquemartin - Thank you for the contribution!

There is a `metadata` parameter in the execute method that is meant to control whether the column names are returned. This aspect just wasn't implemented in the Firebird connector and I don't actually know Firebird so wasn't able to implement when I was breaking the modules apart.

Can we conform to this interface instead? [This docblock](https://github.com/OCA/server-tools/blob/230d855ee1d7ad5a6778989f1b4f2c1704f88619/base_external_dbsource/models/base_external_dbsource.py#L149) explains more, with an example implementation [here](https://github.com/OCA/server-tools/blob/230d855ee1d7ad5a6778989f1b4f2c1704f88619/base_external_dbsource/models/base_external_dbsource.py#L309). Basically it's just a conditional return instead of always, which I assume was to save resources.

We'll also need to bump the version number in the manifest by one patch version (far right most digit in the version raised by 1)

enriquemartin:
@lasley thanks for the info!
how about that?

lasley (approved):
Thanks @enriquemartin!

jonerikceberio (approved):
Working as expected

PR #1000, opened 2017-09-28
[ADD] Sequence check Digit

etobella:
This module adds a Check Digit at the end of the sequence in order to create a validable sequence.

It is useful when some manual checks are required or on integrations. The implemented codes can avoid the modification of one character and flip of two consecutive characters.

The implemented algorithms are:
* Luhn
* Damm
* Verhoeff
* ISO 7064 Mod 11, 2
* ISO 7064 Mod 11, 10
* ISO 7064 Mod 37, 2
* ISO 7064 Mod 37, 36
* ISO 7064 Mod 97, 10

It uses [stdnum](https://github.com/arthurdejong/python-stdnum) package.


JordiBForgeFlow:
@rgarnau, @mreficent can you review?

sequence_check_digit/README.rst
```diff
@@ -0,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Sequence Check Digit
```
MiquelRForgeFlow: add above `===================`

sequence_check_digit/README.rst
```diff
@@ -0,0 +1,61 @@
...
+No manual configuration is needed.
+
+Usage
+=====
+
+* Access sequences and configurate the model to use.
+* The model will check if the format of prefix, suffix and number is valid
+
+Implemented algorithms
+======================
```
MiquelRForgeFlow: remove `====================`. Instead, you can write `*Implemented algorithms:`

MiquelRForgeFlow (approved):
LGTM :+1:

sequence_check_digit/README.rst
```diff
@@ -0,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
rgarnau: If you finally decide to change the license to LGPL-3, then remember that you need to change this image and the header of almost every file.

rgarnau (commented):
Code review :+1: (just minor comments)

rgarnau (approved):

sequence_check_digit/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2017 Creu Blanca
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    "name": "Check Digit on Sequences",
+    "version": "10.0.1.0.0",
+    "category": "Reporting",
+    "website": "https://odoo-community.org",
```
lasley: Please point to this repo instead

sequence_check_digit/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+This module was written to configure check digits on sequences added on the end.
+It is useful as a control of the number on visual validation.
+
+It is useful when some manual checks are required or on integrations.
+The implemented codes can avoid modification of one character and flip of
+two consecutive characters.
+
+
+Configuration
```
lasley: Please remove this section

sequence_check_digit/models/ir_sequence.py
```diff
@@ -0,0 +1,69 @@
...
+        try:
+            self.get_next_char(0)
+        except Exception:
+            raise ValidationError(_('Format is not accepted'))
+
+    def get_check_digit(self, code):
+        if self.check_digit_formula == 'none':
+            return ''
+        try:
+            if self.check_digit_formula == 'luhn':
```
lasley: Instead of this block of conditionals, I recommend the use of a mapping instead. Something like:

```python
formula_map = {
    'none': lambda _: '',
    'luhn': luhn.calc_check_digit,
    'damm': damm.calc_check_digit,
...
}
try:
    return formula_map[self.check_digit_formula](code)
except KeyError:
    raise ValidationError(_('%s is not an implemented format') % self.check_digit_formula)
except Exception:
    raise ValidationError(_('There was an error parsing the code.'))
```

This also covers the instance when the format supplied is not implemented, instead of just returning `None`. That may be undesirable, but that's up to you to decide 😉

sequence_check_digit/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+    "summary": "Adds a check digit on sequences",
+    "depends": [
+        "base",
+    ],
+    "data": [
+        "views/sequence_views.xml",
+    ],
+    "external_dependencies": {
+        "python": [
+            "stdnum"
```
lasley: Please add trailing comma

sequence_check_digit/models/ir_sequence.py
```diff
@@ -0,0 +1,69 @@
...
+                return mod_97_10.calc_check_digits(code)
+            raise ValidationError(_('Function not found'))
+        except Exception:
+            raise ValidationError(_('Format is not accepted'))
+
+    def get_next_char(self, number_next):
+        code = super(IrSequence, self).get_next_char(number_next)
+        if not self.check_digit_formula:
+            return code
+        return code + self.get_check_digit(code)
```
lasley: It would be better to use string interpolation here. Above I noted that in the instance of an `code` being send in that is not defined, we would be returning `None`. In that instance, this line would yield an unexpected `TypeError`:
```python
>>> 'test' + None
  File "<console>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

Recommend:

```python
return '%d%d' % (code, self.get_check_digit(code))
```
The above would also validate for numeric types.

sequence_check_digit/tests/test_check_digit.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_check_digit(self):
+        sequence_obj = self.env['ir.sequence']
+        sequence = sequence_obj.create({
+            'name': 'Test sequence',
+            'implementation': 'standard',
+            'check_digit_formula': 'none',
+            'padding': '5'
+        })
+        sequence.check_digit_formula = 'luhn'
+        self.assertTrue(luhn.validate(sequence.next_by_id()))
```
lasley: It is better to break your tests up into small methods that only test one thing. In this way, failures down the line are more apparent and you get a more clear picture of things that are wrong. When one fails and they are all in the same method, none of the other tests run, which restricts your vision IMO.

lasley (changes requested):
Awesome contribution @etobella! Comments inline

sequence_check_digit/models/ir_sequence.py
etobella: Much cleaner in your way. Modified

sequence_check_digit/models/ir_sequence.py
etobella: I must use %s%s, because the code and the digit control could be alphanumeric (like in ISO 7064 mod 36, 35)

sequence_check_digit/models/ir_sequence.py
```diff
@@ -39,31 +40,29 @@ def check_check_digit_formula(self):
...
-                return mod_97_10.calc_check_digits(code)
-            raise ValidationError(_('Function not found'))
+            return self.get_formula_map()[self.check_digit_formula](code)
+        except KeyError:
+            raise ValidationError(_('%s is not an implemented function'
+                                    % self.check_digit_formula))
         except Exception:
             raise ValidationError(_('Format is not accepted'))
 
+    def get_formula_map(self):
```
lasley: Good call putting it in a method 👍

lasley (approved):
Thanks @etobella!

PR #1001, opened 2017-09-29
9.0

nadiaafa:
On the module **users_ldap_groups**  file **users_ldap_groups.xml** 
line 10, inheritance was using string as selector  :

`                <xpath expr="//form[@string='LDAP Configuration']" position="inside">
`

which result the following error;

> View inheritance may not use attribute 'string' as a selector.

we can't use string as selector in the xpath for odoo 9.0, instead we can do the following:

`<xpath expr="//field[@name='ldaps']/form" position="inside">`

Thank you


LoisRForgeFlow:
@nadiaafa I think your target branch is wrong

nadiaafa:
No @lreficent i am sure on the 9.0 branch, please check that

lasley:
Hi @nadiaafa - This PR is targeting OCA/10.0. I went to update the target to 9.0 and it seems that it's just a PR from OCA/9.0 => OCA/10.0. Is this intentional?

nadiaafa:
@lasley ah yeah sorry

PR #1009, opened 2017-10-03
[ADD] [database_cleanup] migrate to 11.0

hbrunn:


hbrunn:
this will however duplicate all of transbot's commits. Should we add an instruction to squash those commits to the migration guide?

StefanRijnhart:
These transbot commits are partial right? So they only include the changes to the migrated module? In that case it would be a good idea to include that instruction. Or maybe there is some obscure git command to do it automatically ;-)

database_cleanup/models/purge_columns.py
```diff
@@ -0,0 +1,130 @@
...
+            if model.model not in self.env:
+                continue
+            model_pool = self.env[model.model]
+            if not model_pool._auto:
+                continue
+            table2model.setdefault(
+                model_pool._table, (model.id, [])
+            )[1].append(model_pool)
+
+        for table, model_spec in table2model.iteritems():
```
StefanRijnhart: Remaining instance of iteritems

Garamotte: 2to3 normally takes care of this kind of modification, did you use it ?

StefanRijnhart:
Interesting new use case: in 11.0/base, the usage field from the inherited table ir_actions is uninherited and added to some of the child tables. Can you make purge_columns so that it removes the column from the parent table first (which removes the column from the inheriting tables as well), otherwise an error occurs when purging all columns.

```
  File "/home/bruno/buildout/11.0/parts/server-tools/database_cleanup/models/purge_columns.py", line 51, in purge
    IdentifierAdapter(line.name)
  File "/home/bruno/buildout/11.0/parts/odoo/odoo/sql_db.py", line 155, in wrapper
  File "/home/bruno/buildout/11.0/parts/odoo/odoo/sql_db.py", line 232, in execute
psycopg2.ProgrammingError: cannot drop inherited column "openupgrade_legacy_11_0_usage"
```

database_cleanup/models/purge_tables.py
```diff
@@ -0,0 +1,110 @@
...
+    _description = 'Purge tables'
+
+    @api.model
+    def find(self):
+        """
+        Search for tables that cannot be instantiated.
+        Ignore views for now.
+        """
+        # Start out with known tables with no model
+        known_tables = ['wkf_witm_trans']
```
StefanRijnhart: With the removal of the workflows, this table can and should now be purged as well.

StefanRijnhart (changes requested):
Thanks! See inline comments for requested changes.

hbrunn:
@StefanRijnhart thanks for the review, points taken. https://github.com/OCA/server-tools/pull/1009#issuecomment-334430872 is a feature request in my book that shouldn't be part of the migration PR

database_cleanup/models/purge_columns.py
hbrunn: no

StefanRijnhart (approved):
Fair enough! Thanks for the changes done.

pedrobaeza:
@hbrunn maybe you can take the occasion for squashing some of the same author commits that are not too relevant?

DelfosCloud (approved):

hbrunn:
@pedrobaeza done

pedrobaeza (approved):

PR #1013, opened 2017-10-06
[MIG][10.0] base_fontawesome

simahawk:


pedrobaeza:
Isn't 4.7 bundled in Odoo 10?

yajo (approved):
[No, it's 4.5](https://github.com/odoo/odoo/blob/10.0/addons/web/static/lib/fontawesome/css/font-awesome.css#L2). v11 is the one that includes 4.7.

pedrobaeza (approved):

Issue #1006, opened 2017-10-02
[RFC] Split Server-Tools for v11


lasley:
[Bringing this in from a mailing list thread for final discussion.](https://odoo-community.org/groups/tools-22/tools-44031?mode=thread)

From @dreispt (there's more in the thread, but this is the one that's still relevant right now and should be discussed):

> Hello,
>
> I believe that the repo can be subdivided in several more focused ones.
>
> I propose to keep OCA/serverl-tools mostly for features related to configuration and administration tasks.
> The other features can be moved to a few new spinf-off repos.
> 
> I can think of four new repos to create as spin-offs:
> - OCA/server-auth (10 modules): Authentication related
> - OCA/server-backend (5 modules): Server ORM extensions, new fields
> - OCA/server-brand (8 modules and counting): (De)branding related
> - OCA/server-ux (11 modules): Server side features for usability and user experience related
>
> Supporting the above division, and as a starting point for the discussion, I prepared this spreadsheet:
https://docs.google.com/spreadsheets/d/1Xg95cW4TFMf_Lo5i_CZC_qOOfN8RgxPRc0LJTLTkdUI/edit?usp=sharing

cc @pedrobaeza @hbrunn @OCA/board

dreispt:
Bringing up this again was on my list, thanks a lot for bringing it up Dave.
Experience tells us that focused repositories have better direction and more motivated maintainers.
This repository covers too much different topics for it to be effectively maintained.
It also has signs of being too large, looking at the number of existing modules and open PRs.

The list spinoffs proposed is quite old, so they might need an update.

JordiBForgeFlow:
@dreispt I do not see consensus, and at the same time I see that the PR's to v11 are occurring as we speak. I propose to defer this topic until v12, as it seems that this kind of structural changes need to be planned well in advance of a new release.

elicoidal:
Unfortunately I have to agree that this conversation is starting too late: we need to plan now the repository structure for v12 NOW.

JordiBForgeFlow:
OK, I have renamed the issue accordingly.

pedrobaeza:
No, it can be perfectly done in v11, and more as repo has been emptied. I'm creating the repos next day.

lasley:
I brought this topic up back in ~~June~~ August. My standpoint is that we should do it now or it will never get done.

JordiBForgeFlow:
[...]

PR #1014, opened 2017-10-06
11.0 mig sequence check digit

etobella:
No major changes with version 10.0.
Travis was modified because:
* Python module MySQL-python is not compatible with Python 3
* Module database_clean is not updated yet


lasley:
@etobella - would you mind submitting the Travis changes in a separate PR please?

etobella:
I understand that the changes I made on Travis will be solved on PR #1009, then, is it necessary to create the second PR when it will be solved in a few days?

lasley:
@etobella - Ok don't worry about it - really it should have been broken from that PR. This is why changes like this should be broken out - because a lot of people need them and they aren't obvious nested elsewhere. If it was a simple PR for the Travis change, I could just quick merge it right now and we could go on with life 😑

lasley (approved):
Code 👍

rgarnau (approved):
Code and functional review :+1:

JordiBForgeFlow (approved):
:+1: Functional review. But before merge we need to consider that the destination repo could change: https://github.com/OCA/server-tools/issues/1006

etobella:
Migrated to server-ux repository.

PR #1020, opened 2017-10-11
[FIX] warning about dynamic context

hbrunn:

PR #1021, opened 2017-10-11
[FIX] don't break when an attribute node in an xpath makes the

### LINKED ISSUE #885: 8.0 Incompatibility between base_view_inheritance_extension and stock_packaging_usability_ul
If you try to install the module base_view_inheritance_extension with the module stock_packaging_usability_ul from here : https://github.com/OCA/stock-logistics-tracking
The installation fail.
I do not have the time to look at here more (maybe latter) , but as I have detected the issue I think it worth to share it.

hbrunn:
expression not match the node any more. Fixes #885 @sebastienbeau @StefanRijnhart

### DIFF
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -55,6 +55,9 @@ class IrUiView(models.Model):
                     yield node, handler
             return
         if spec.get('position') == 'attributes':
+            if all(not c.get('operation') for c in spec):
+                yield spec, self._get_inheritance_handler(spec)
+                return
             for child in spec:
                 node = etree.Element(spec.tag, **spec.attrib)
                 node.insert(0, child)
--- a/base_view_inheritance_extension/__manifest__.py
+++ b/base_view_inheritance_extension/__manifest__.py
@@ -3,7 +3,7 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 {
     "name": "Extended view inheritance",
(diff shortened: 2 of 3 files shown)


hbrunn:
warning addressed in https://github.com/OCA/server-tools/pull/1020

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -55,6 +55,9 @@ def _iter_inheritance_specs(self, spec):
                     yield node, handler
             return
         if spec.get('position') == 'attributes':
+            if all(not c.get('operation') for c in spec):
```
yajo, line 58: What if one has operation and another doesn't?

yajo (changes requested):

base_view_inheritance_extension/models/ir_ui_view.py
hbrunn: then it's a view which uses this module and then we can expect devs to order the child nodes in a way it doesn't break. Might be a good idea to talk about that in the readme.
The important issue this PR is meant to fix is that just installing this module breaks completely fine code like https://github.com/OCA/OCB/blob/10.0/addons/product/views/product_views.xml#L285

StefanRijnhart (approved):

sebastienbeau (approved):

StefanRijnhart:
Oh, a bump of the version number is desired I believe?

pedrobaeza:
Can you please change module version?

hbrunn:
done

yajo (approved):

PR #1029, opened 2017-10-17
10.0 mig base_bypass_record_rules

SimoRubi:
Hi,
I have migrated to v10 the module developed in [this PR](https://github.com/OCA/server-tools/pull/773) for v8, I also have enhanced security of the module as proposed in the PR discussion.
Please check, thanks


pedrobaeza:
It already exists base_suspend_security. What is the difference?

hbrunn (changes requested):
what happens if i just pass a group here which I'm a member of? The only thing you improved with your change is that now your don't turn off rules globally, but only for the ones with the flag set.
Also note that the domains read are cached, so this will end badly if the same cache encounters a read with and without the context key.
I think you should abandon the idea doing anything security related with influence from the client side entirely. Whatever the problem is you want to solve with this, it should be possible to do this with group specific view overrides and extra rule assignments

eLBati:
> Whatever the problem is you want to solve with this, it should be possible to do this with group specific view overrides and extra rule assignments

@hbrunn the requirement in short is: user can't see records of the other company, except within a specific form.

You can find an example here https://github.com/eLBati/account-invoicing/blob/c94783ebbd9298650c8f16d41a764bbb4604b6ed/account_invoice_fiscal_agent/views/account_fiscal_position.xml : user should be able to configure the 'Fiscal Agent Account Mapping' selecting, within the ```account_dest_id``` field, accounts not belonging to their company.
But, in every other place, user should not be able to see accounts not belonging to their company.

Do you have any hint about how to do it?

hbrunn:
so you want to avoid lifting the rule globally because then you'd have to adapt all other selection fields? This means you actually want to use an ir.rule as a convenient default filter. Then you should be honest about it, deactivate the rule for reading (and only for reading), and override the model's `search` function such that it injects a `('company_id', '=', self.env.user.company_id.id)` term into the domain if there's no term that has `'company_id'` on the left hand side. That's basically what core does with the `active` flag.
I can imagine this to be useful as mixin class where you can specify a default domain, which will be imposed if none of the fields involved are used in the domain.
If you need to pretend to keep the rule intact for whatever reason, you could create an `AbstractModel` that inherits from the account model and passes through all `search` and `read` calls to the `sudo`ed account model, but skips writes, creates and unlinks. This way, you've created an unrestricted read only view.

SimoRubi:
Thanks, we are no more working for this so I'm closing the PR

PR #1030, opened 2017-10-17
[FIX] only purge our own table

hbrunn:


pedrobaeza:
The problem with this is that it can remove a table that you don't want to remove because it's not obsolete, but has a dependency.

hbrunn:
@pedrobaeza true, now a bit more cautiously

pedrobaeza:
I only see changes in the test

hbrunn:
yes, the test gets confused in #988. We probably won't ever run into a situation like this in the real world, so better to fix the test than code

lasley (approved):
Approved with a disclaimer that I don't actually understand what this code does, only that it seems safe enough and the reasoning for only adjusting a test instead of adjusting code seems sound.

hbrunn:
the thing is that up to now, the test has cleaned up tables from other modules, which are in the test case in a kind of limbo state. And this breaks now because `base_import_odoo` creates a table that has foreign key constraints to `ir_model_data`. So the test tested the wrong things before, and now we restrict it to what it should have been testing in the first case.

pedrobaeza (approved):
OK, merging then

PR #1033, opened 2017-10-20
[IMP] base_external_system: Add create bypass

lasley:
* In cases of deep inheritance, it may be required to create an adapter directly. Add an override in the create via the env context to support this.

I noticed this as working on OCA/partner-contact#500 - I had never considered the possibility of joining two adapters together like that.

This will also allow for things like connector backends to implement the system interface

### DIFF
--- a/base_external_system/__manifest__.py
+++ b/base_external_system/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Base External System",
     "summary": "Data models allowing for connection to external systems.",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Base",
     "website": "https://laslabs.com/",
     "author": "LasLabs, Odoo Community Association (OCA)",
--- a/base_external_system/models/external_system.py
+++ b/base_external_system/models/external_system.py
@@ -112,10 +112,11 @@ class ExternalSystem(models.Model):
     def create(self, vals):
         """Create the interface for the record and assign to ``interface``."""
         record = super(ExternalSystem, self).create(vals)
-        interface = self.env[vals['system_type']].create({
-            'system_id': record.id,
-        })
-        record.interface = interface
+        if not self.env.context.get('no_create_interface'):
+            interface = self.env[vals['system_type']].create({
+                'system_id': record.id,
+            })
+            record.interface = interface
         return record
 
     @api.multi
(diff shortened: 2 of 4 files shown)


lasley:
Note to self: interface should unlink the system when it is unlinked.

lasley:
As I'm implementing this into a few different types of systems, and creating adapter/interface mechanisms based on those (such as in OCA/partner-contact#500), I'm noticing that we need a way to define what interfaces the external systems implement.

A good example is with the Address Validation interface in OCA/partner-contact#500. Our Easypost connector is going to implement this interface, but it would be nice for it to also be able to implement some sort of unplanned/future abstract shipping interface as well.

Similar mindset exists with our redesign of the Avatax connector, which is also going to implement the address verification interface. It could also implement, say, a Tax verification interface.

I wonder the best approach to go about defining what is implemented on a system without requiring a bunch of boilerplate...

pedrobaeza:
Isn't this converting in something over-engineered?

lasley:
Possibly, yeah. But I really can't think of a better way to do it. The problem exists that multiple systems will implement the same thing in different ways, but we need a unified way of presenting it - or a pattern devised that works. Maybe we don't need to actually build anything, but just identify the best method for doing this.

Using the example of the address verifiers - every address verification system is going to have a similar workflow, and we would want to be able to abstractly link these verifiers into business processes without having to worry about what specific validator is in use. A validator is also an external system, so it makes sense to inherit it from that.

In my current design, I created an `address.validator` model that acts as a central inheritance point for systems that want to implement this. In order to implement, the system simply creates the proper methods prefixed by the system name - basically the same thing we have for shipping connectors.

This is a fine strategy I think, but we're going to end up with a lot of boilerplate (as you can see in [my implementation](https://github.com/LasLabs/partner-contact/blob/release/10.0/base_partner_address_validate/base_partner_validate_address/models/address_validate.py)). 

It would be nice if this could somehow be defined in the core, so that all a new interface needs to do is create a model for itself defining the methods to be implemented. Maybe the solution is just pulling some of this boilerplate into an abstract model, then writing some solid documentation around it.

lasley:
Ok yeah so I just kept coming up with over-engineered stuff. We'll leave it at this fix and I'll loop back around to something more abstract for system grouping once we've created a few more group types.

lasley:
Alright tests added & code finalized. This is ready for review!

lasley:
@obulkin - please take a look at this when you get a chance

obulkin (approved):
LGTM

obulkin:
Runbot is red to due to an issue with ``base_view_inheritance_extension``, which seems unrelated.

PR #1034, opened 2017-10-21
[FIX] oauth_provider: Fix tests when using last oauthlib versions

### LINKED ISSUE #1011: V9 broken
@lasley sounds like 9.0 is broken because of https://travis-ci.org/OCA/server-tools/jobs/282692962#L1764-L1780

Garamotte:
The last oauthlib versions changed the returned error code and description in some cases.
I adapted the tests to work with both old and new versions of the lib.

Fixes #1011


pedrobaeza (approved):
Fast-tracking this for greenifying branch

PR #1036, opened 2017-10-23
11.0 mig base user role

novawish:


pedrobaeza:
Maybe this is better suited now in server-backend instead of here. @lasley @sebalix  what do you think?

sebalix:
No problem with that. I suppose `auditlog` like others "user application" should be migrated there too.

jcdrubay (approved):

novawish:
As @pedrobaeza mentioned, i created a new PR of this module to `server-backend` repository: https://github.com/OCA/server-backend/pull/1

Therefore, I will close this PR since it was replaced by the new PR

pedrobaeza:
Thanks.

PR #1039, opened 2017-11-01
[ADD] support user_ids for user specific view overrides

hbrunn:
this adds the analogue to `groups_id` for users on views. Will be used in https://github.com/OCA/web/pull/787

### DIFF
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -2,7 +2,7 @@
 # © 2016 Therp BV <http://therp.nl>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 from lxml import etree
-from openerp import api, models, tools
+from openerp import api, fields, models, tools
 
 
 class UnquoteObject(str):
@@ -41,11 +41,23 @@ class UnquoteEvalObjectContext(tools.misc.UnquoteEvalContext):
 class IrUiView(models.Model):
     _inherit = 'ir.ui.view'
 
+    user_ids = fields.Many2many('res.users', string='Restrict to users')
+
     @api.model
     def apply_inheritance_specs(self, source, specs_tree, inherit_id):
         for specs, handled_by in self._iter_inheritance_specs(specs_tree):
             source = handled_by(source, specs, inherit_id)
         return source
+
+    @api.model
+    def get_inheriting_views_arch(self, view_id, model):
+        return [
+            (arch, view_id_)
+            for arch, view_id_ in
+            super(IrUiView, self).get_inheriting_views_arch(view_id, model)
+            if not self.sudo().browse(view_id_).user_ids or
+            self.sudo().browse(view_id_).user_ids & self.env.user
+        ]
 
     @api.model
     def _iter_inheritance_specs(self, spec):
--- a/base_view_inheritance_extension/README.rst
+++ b/base_view_inheritance_extension/README.rst
@@ -8,6 +8,7 @@ Extended view inheritance
 
 This module was written to make it simple to add custom operators for view
 inheritance.
+It also adds a field `user_ids` to restrict view inheritance to certain users.
 
 Usage
 =====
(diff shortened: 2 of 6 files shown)


StefanRijnhart:
I am thinking that there are no user level permissions anywhere in Odoo, so introducing one here seems a bit idiosyncratic to me. Why not work around this yourself by adding user-specific or function specific groups?

hbrunn:
this tries to do what custom views aim at and never delivered: User specific views

StefanRijnhart:
Sorry, I missed the reference to the web PR.

StefanRijnhart:
While unlinking users is unadvisable or usually even impossible, you still might want to delete custom views if someone manages to unlink a user that is the last user on such a view.

dreispt:
I agree with Stefan. Other than that you have my :+1:

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -41,12 +41,24 @@ def __missing__(self, key):
...
         return source
 
+    @api.model
+    def get_inheriting_views_arch(self, view_id, model):
+        return [
+            (arch, view_id_)
+            for arch, view_id_ in
+            super(IrUiView, self).get_inheriting_views_arch(view_id, model)
+            if not self.sudo().browse(view_id_).user_ids or
+            self.sudo().browse(view_id_).user_ids & self.env.user
```
gfcapalbo, line 59: means return this if does not have user_ids restriction or if the exclusive union (intersection) of user ids is and current user  is not null.  compact but hard to read . if there aren't computational advantages i do not know about ,  i believe this is less readable. gust a general opinion, MR looks good

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -41,12 +41,24 @@ def __missing__(self, key):
 class IrUiView(models.Model):
     _inherit = 'ir.ui.view'
 
+    user_ids = fields.Many2many('res.users', string='Restrict to users')
+
     @api.model
     def apply_inheritance_specs(self, source, specs_tree, inherit_id):
         for specs, handled_by in self._iter_inheritance_specs(specs_tree):
```
gfcapalbo: question/ explanation:   _iter_inheritance_specs  returns the node and the handler, 
 _iter_inheritance_specs then iterates and yields through the tree and computes node/attributes and finally returns, but i see no  new attribute user_id= to be handled. So is this intended not to be used in XML as a new attribute but just to impose via code some user_ids for certain XMLids, that will then be popped out get_inheriting_views_arch?

Or have i misunderstood?

gfcapalbo (commented):
I have a question to be sure i have understood the functioning of the module.

gfcapalbo (approved):

gfcapalbo:
:+1:

daramousk (approved):

StefanRijnhart:
Any thoughts on my suggestion (as supported by Daniel)?

daramousk:
> While unlinking users is unadvisable or usually even impossible, you still might want to delete custom views if someone manages to unlink a user that is the last user on such a view.

I went on to see whether the deletion of a group that has a view assigned to it also deletes the view, which is not the case. Maybe we should keep the logic consistent by not deleting the view?

StefanRijnhart:
> I went on to see whether the deletion of a group that has a view assigned to it also deletes the view, which is not the case. Maybe we should keep the logic consistent by not deleting the view?

Or it's a bug that needs fixing in Odoo.

I'll go as far as to say that the issue of deleting the last remaining user on a view is a *theoretical* one given given all the not-null FK constraints there already are on res_users so that users that have been active for some time are much more likely to be set to inactive instead of deleted, but then again the fix would be trivial so I'd like to hear @hbrunn's take on it.

hbrunn:
Sure that's something that should be fixed. I'd expect a bigger problem with deactivating a user that has a custom view, because in this case, this user won't show up in https://github.com/OCA/server-tools/pull/1039/files#diff-3275b498ea3880e3d85efa861d91ab17R58 and we apply the override for the deactivated user to all users. And this happens all the time. Please @daramousk fix both cases and PR to this branch

daramousk:
Pushed here https://github.com/hbrunn/server-tools/pull/7

StefanRijnhart (approved):

PR #1040, opened 2017-11-03
[11.0] [MIG] auditlog

etobella:
Migration to 11.0


auditlog/models/rule.py
```diff
@@ -566,10 +539,10 @@ def unsubscribe(self):
             if act_window:
                 value = 'ir.actions.act_window,%s' % act_window.id
                 act_window.unlink()
-                ir_value = ir_values_model.search(
+                ir_default = model_ir_default.search(
```
StefanRijnhart: Earlier on, ir.values was used for two purposes: registering default values, and exposing window actions under the More... button of certain record types. I believe that in this case, it was used for the latter purpose so after removing the window action of a model's logging you don't have to check for pending defaults so you can remove the modified lines.

auditlog/models/rule.py
```diff
@@ -527,34 +509,25 @@ def subscribe(self):
...
-        model_ir_values = self.env['ir.values']
         for rule in self:
             # Create a shortcut to view logs
             domain = "[('model_id', '=', %s), ('res_id', '=', active_id)]" % (
                 rule.model_id.id)
             vals = {
-                'name': _(u"View logs"),
+                'name': _("View logs"),
                 'res_model': 'auditlog.log',
                 'src_model': rule.model_id.model,
```
StefanRijnhart: When I test on runbot, I don't get the option to View Logs under the More... button after subscribing to the logs of a particular model. I believe you need to set binding_model_id to rule.model_id.id here.

StefanRijnhart:
Great work! I only have some minor comments about how action windows work in 11.

etobella:
@StefanRijnhart You are right. Once the field was added, it works as expected.
Thanks!

StefanRijnhart (approved):
Thanks for the update!

rgarnau (approved):
LGTM :+1:

pedrobaeza:
@dreispt next time it's good to ask for a squash of the commits that are meaningless like that one with only `https` change.

StefanRijnhart:
@pedrobaeza I respectfully disagree. I believe rebasing out the odd commit is not that important.

pedrobaeza:
It's not so important, but obviously a clean history helps a lot when browsing for logs.

dreispt:
It looks like we don't all have the same feelings regarding commit history.

pedrobaeza:
Well, when you have a module where in development phase, 100 commits has been done, you change your mind about cleaning commits... As said, in this case it's not so important, as the number if not very high, but remember that this scales with each version, and one of the criteria of the maintainability is the escalation.

PR #1044, opened 2017-11-09
fixed: search active exception rule

mqt2r:


pedrobaeza:
active flag is added by default in any search, so this doesn't add anything

PR #837, opened 2017-05-12
[WIP] 8.0: module_prototyper: Export data and demo data

mvaled:
Implements (to some degree) the option to export the data of a module when generating a module from the module_prototyper.

PR #1045, opened 2017-11-09
[10.0] module_prototyper: Export demo data.

mvaled:
Export model's demo data.

This is the same as in #837 but for Odoo 10.


fmdl:
Dear don't forget add the external_dependencies

```
    'external_dependencies': {
        'python': ['xoutil'],
    },
```

fmdl:
Dear

I try to export all account whit this filter:
image

It take 10 minutes, and the result is an empty file.
image

And lot of error of this kind:
`2018-01-14 13:59:43,443 59253 WARNING iguana_master odoo.addons.module_prototyper.utils.closure: Ignoring cycle detected: [<[(152, u'FR76 3000 4008 8900 0232 4727 616')]>, <[(347, u'SICOMIN')] (from: partner_id)>, <[(347, u'SICOMIN')] (from: commercial_partner_id)>]`

I have made this module specific to export data: https://github.com/OCA/server-tools/pull/1130

mvaled:
Hi @fmdl 

It's been a while since I don't use this addon.  Notice that the log issues a warning: it cannot find a topological order of the selected records.   This is quite common in some models.  That's why you have to select which relations to traverse.

PR #1049, opened 2017-11-13
[FIX] As an ERP manager, allow to reset users' passwords

StefanRijnhart:
As only SUPERUSER_ID can work with password history records for other users than the own user, users in the Settings group can no longer reset other users' passwords when `password_security` is installed. This fix makes it possible for them again to do so.

### DIFF
--- a/password_security/__openerp__.py
+++ b/password_security/__openerp__.py
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '8.0.1.0.0',
+    'version': '8.0.1.1.0',
     'author': "LasLabs, Odoo Community Association (OCA)",
     'category': 'Base',
     'depends': [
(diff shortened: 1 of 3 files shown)


lembregtse (approved):

lasley (approved):
Thanks @StefanRijnhart

pedrobaeza:
Please change module version number

StefanRijnhart:
As @hbrunn suggested recently, maybe make a CI option to signal this? Otherwise we will never learn.

StefanRijnhart:
Version number now updated!

elicoidal (approved):

pedrobaeza:
Yeah, it would be interested to have that automation, although there are other problems related to that automation (what to update, extra commits...).

Merging this one.

lasley:
I could have sworn we had an issue for version number automation somewhere, but I can't seem to find it

pedrobaeza:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

PR #1052, opened 2017-11-13
11.0 mig date range

TDu:


etobella:
Date range is migrated on https://github.com/OCA/server-ux/

PR #1056, opened 2017-11-14
[9.0] base_view_inheritance_extension: Implement target_position for position="move"

mehdi-ghezal:
By default, inheritance with spec **position="move"** append the node inside the target. 
The spec **target_position** let you choose what you want to do : 
- **inside**: append the node inside the target
- **before**: append the node before target
- **after**: append the node after target

By default, if no value is provide we use the standard behaviour **inside**

### DIFF
--- a/base_view_inheritance_extension/README.rst
+++ b/base_view_inheritance_extension/README.rst
@@ -35,8 +35,14 @@ Move an element in the view
 
     <xpath expr="$xpath" position="move" target="$targetxpath" />
 
-This can also be used to wrap some element into another, create the target
-element first, then move the node youwant to wrap there.
+Availables values for position attribute:
+
+* inside (default if omited)
+* after
+* before
+
+This can also be used to wrap some element into another, create the new target
+element first, then move the element you want inside the new element.
 
 Add to values in a list (states for example)
 --------------------------------------------
@@ -60,7 +66,6 @@ Known issues / Roadmap
 ======================
 
 * add ``<attribute operation="json_dict" key="$key">$value</attribute>``
-* support ``<xpath expr="$xpath" position="move" target="xpath" target_position="position" />``
 * support an ``eval`` attribute for our new node types
 
 Bug Tracker
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -2,7 +2,7 @@
 # © 2016 Therp BV <http://therp.nl>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 from lxml import etree
-from openerp import api, models, tools
+from openerp import api, models, tools, _
 
 
 class UnquoteObject(str):
@@ -120,7 +120,18 @@ class IrUiView(models.Model):
         target_node = self.locate_node(
             source, etree.Element(specs.tag, expr=specs.get('target'))
         )
-        target_node.append(node)
+        
+        target_position = specs.get('target_position') or 'inside'
+        
+        if target_position == 'after':
+            target_node.addnext(node)
+        elif target_position == 'before':
+            target_node.addprevious(node)
+        elif target_position == 'inside':
+            target_node.append(node)
+        else:
+            self.raise_view_error(_("Invalid target_position attribute: '%s'") % target_position, inherit_id, context=self.env.context)
+
         return source
 
     @api.model
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -120,9 +120,9 @@ class IrUiView(models.Model):
         target_node = self.locate_node(
             source, etree.Element(specs.tag, expr=specs.get('target'))
         )
-        
+
         target_position = specs.get('target_position') or 'inside'
-        
+
         if target_position == 'after':
             target_node.addnext(node)
         elif target_position == 'before':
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Thanks for this feature!

hbrunn (changes requested):
runbot has a few complaints, and please remove the line about this from the roadmap then too. Also mention in the readme that this works so that other devs don't have to study the code.

you should also extend the tests so that all your code branches are hit by testing

hbrunn (changes requested):
this is already great. Now just add your case to https://github.com/mehdi-ghezal/server-tools/blob/d5bcd02230124fd75bdfce1608d1930be9fd8018/base_view_inheritance_extension/demo/ir_ui_view.xml and assert something sensible about it in https://github.com/mehdi-ghezal/server-tools/blob/d5bcd02230124fd75bdfce1608d1930be9fd8018/base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py

StefanRijnhart:
Hi @mehdi-ghezal, can you please send your CLA to the OCA and add the extra tests? That way we can adopt this useful feature. Let us know if you need help.

FYI I'm now using this 10.0 version https://github.com/StefanRijnhart/server-tools/tree/feature/10.0/base_view_inheritance_extension/target_position_mehdi_ghezal

base_view_inheritance_extension/README.rst
```diff
@@ -35,8 +35,14 @@ Move an element in the view
...
 
-This can also be used to wrap some element into another, create the target
-element first, then move the node youwant to wrap there.
+Availables values for position attribute:
+
+* inside (default if omited)
+* after
+* before
+
+This can also be used to wrap some element into another, create the new target
```
Garamotte, line 44: If this means `simply create a new node around an existing node`, this is already possible with the standard `replace`, [using `$0`](https://www.odoo.com/documentation/11.0/reference/views.html#inheritance-specs).

If this means `create a new node anywhere and move another node in it in a single xpath node`, I think the same logic could be used on the `move` position to do it with a single xpath node.

hbrunn: we're in 9, `$0` is a feature from 10 upwards. Counterintuitive as this is, it might indeed make sense to backport this feature instead of cooking up another syntax

Garamotte: Oh, you're right, I didn't took attention to the target version :)

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -120,7 +120,22 @@ def inheritance_handler_xpath(self, source, specs, inherit_id):
...
         )
-        target_node.append(node)
+
+        target_position = specs.get('target_position') or 'inside'
+
+        if target_position == 'after':
+            target_node.addnext(node)
+        elif target_position == 'before':
+            target_node.addprevious(node)
+        elif target_position == 'inside':
```
NL66278, line 130: Please support 'move' as a (maybe deprecated) synonym for 'inside'. This module is already used a lot, with the move value. So better not break existing views.

StefanRijnhart: @NL66278 This is an additional attribute next to `position='move'` which remains necessary.

hbrunn:
@StefanRijnhart are you waiting for this to be merged to propose your port to 10?

hbrunn (approved):

StefanRijnhart:
@hbrunn I'm waiting for the CLA clearance before I propose my port.

hbrunn:
given the time this takes already it might not happen at all, maybe @NL66278 could reimplement this in a clean room fashion and propose himself?

StefanRijnhart:
That would be a solution. I share the fear that we are in a dead corner.

mehdi-ghezal:
Hi, I'm sorry but I am quite busy, I just sent the CLA by email :)

hbrunn:
@great. then the cla bot will kick in in a while and that's fine. But we still need to increase code coverage, please check https://github.com/OCA/server-tools/pull/1056#pullrequestreview-80411060 where I propose how to add tests

base_view_inheritance_extension/models/ir_ui_view.py
NL66278: Yes I see, I was put on a wrong footing by the README, which does not mention target_position.

hbrunn:
yes, please also make it clear in the readme that it's about attribute `target_position`

remihb:
LGTM, PR still pending due to missing tests ?

I really need this feature and I don't want to use my own fork anymore as the PR is working as expected since a while (tested in production in 9.0 and 10.0)

hbrunn:
missing tests and not enough approvals

remihb:
Having my approval is not enought ? :)
Ok, I will set up some tests and hope..

hbrunn:
you didn't use the review/approve mechanism, so you don't count yet

NL66278 (approved):

remihb (approved):

remihb:
Ok, I wasn't sure of the process. I thought that because of missing tests, the PR shouldn't be approved.

hbrunn:
it's fine to approve, the label `ready to merge` will only appear when all tests succeed, in this case, when coverage is sufficiently high

StefanRijnhart:
Did the cla-bot come back with a green light? I don't see it.

remihb:
Ok, what can I do if I want to add tests above mehdi's changes? Should I make my own PR based on this one?

StefanRijnhart:
OK, I checked the OCA database and there is a CLA! Yes @remihb, in that case please create PR to replace this one.

PR #1059, opened 2017-11-16
[ADD] create an xmlid for mapped records too

hbrunn:
this will fix cases where we have a mapping, but still want to overwrite values with duplicate handling overwrite*


StefanRijnhart (approved):

JayVora-SerpentCS (approved):

JayVora-SerpentCS:
Little squash and All good.

Issue #1041, opened 2017-11-05
[Security] auth_totp : can be bypass by xml_rpc


fmdl:
1- install instance with auth_totp
2- add device and enable MFA
3- connecte with your browser
4- disconnecte on your browser

Use xml_rpc to connecte with your login/password only, MFA no checked !!!

MFA doen't work with xml_rpc.

@obulkin @lasley

lasley:
Thanks for the report @fmdl - I never considered XML-RPC in this. Thinking about it, I'm actually not sure how it would be possible while still fitting within the framework of the RPC spec (or any API really).

I think the solution here is a module that allows for the restriction of XML-RPC based on some user options/ACL, then leverage that in `auth_totp` to allow the user to disable their XML-RPC or accept the risks.

Do you agree with this? Or maybe have a good idea on how to have MFA work with an API? I'm having a bit of trouble wrapping my head around the implementation for an API with MFA TBH.

fmdl:
​It is possible to solve this issue. You can not disable xml_rpc, because the UI web client use json_rpc (and you can do the same things).

​The solution is inside the odoo module : `auth_crypt`.
When you install this mode it copy the field `password` in `password_crypt` whis SHA cryptage. And it set to `NULL` the `password` field. Like that the standard methode (check the field password) return AccesDenied.

In this case create a new field `password_crypt_MFA`, when you active the MFA for an user, you need to copy password_crypt in `password_crypt_MFA`, and set to `NULL` `password_crypt`. Like that the methode to check of `auth_crypt` module return `AccesDenied`. 
And in methode check_credential, you need to do some thinks like that :

1- check_credential check if it is the good password in `password_crypt_FMA`, if it okay, it generate a token and a cookies, and return AccesDenied
2- the main.py know that the AccesDenied, but he see that there a token for this user and return to the auth_totp page.
3- the user set the confirmation code, and now the check_credential check cookie, token, and the confirmation code. If it is okay, set that is good.
4- when the system check credential (not for login), just check cookies and token lifetime.

Question, you system it is multi browser use in same time (iPhone, Home PC, Work PC) ? (At first view, I think no because there can be only one token per user).
Maybe create a new model of token with this fields to use in multi-brower:
- token (default = generate_token)
- confirmecode_validate (default = False)
- max date_time
- cookies_key (default = generate_cookies_key)


Nota 1 : to generate a token you can use `uuid5` or https://docs.python.org/3/library/secrets.html,  ?
[...]

PR #1065, opened 2017-11-21
[10.0][FIX] auth_totp: Various issues

obulkin:
* Restructure controller and ``res.users`` logic to prevent RPC authentication for users with MFA enabled and add support for multiple simultaneous MFA sessions
* Switch trusted device cookies from using the DB secret to user-level secret keys, thereby increasing security
* Remove MFA login tokens and trusted device model, which are now redundant
* Add migration logic that generates a trusted device cookie key for every user with MFA enabled and cleans up device model ``ir`` records to prevent warnings
* Update unit tests and remainder of module accordingly

This PR addresses the issues mentioned in https://github.com/OCA/server-tools/issues/1041.


auth_totp/controllers/main.py
```diff
@@ -81,12 +45,12 @@ def mfa_login_post(self, *args, **kwargs):
         """Process MFA login attempt
 
         Overview:
-            * Try to find a user based on the MFA login token. If this doesn't
+            * Identify current user based on login in session. If this doesn't
```
lasley: The `in` after `login` is superfluous. Or maybe there's a word missing?

auth_totp/migrations/10.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+import uuid
```
lasley: Typically you want to import only the namespace(s) you're using (with obvious exceptions if it starts getting out of hand). So in this case, `from uuid import uuid4` would be better.

auth_totp/models/res_users.py
```diff
@@ -2,12 +2,12 @@
 # Copyright 2016-2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-from datetime import datetime, timedelta
-import random
-import string
+import uuid
```
lasley: `from uuid import uuid4`

auth_totp/models/res_users.py
```diff
@@ -29,14 +29,20 @@ def _build_model(cls, pool, cr):
...
+        compute='_compute_trusted_device_cookie_key',
+        store=True,
     )
 
+    @api.multi
+    @api.depends('mfa_enabled')
+    def _compute_trusted_device_cookie_key(self):
+        for record in self:
+            if record.mfa_enabled:
+                record.trusted_device_cookie_key = uuid.uuid4()
```
lasley: `record.trusted_device_cookie_key = record.mfa_enabled and uuid.uuid4() or False`

^--I'm almost sure that'll fit on one line once the nesting is reduced. Otherwise, just use an intermediary var:

```
key = ...
record.tru... = ...
```

auth_totp/models/res_users.py
```diff
@@ -50,53 +56,38 @@ def _check_enabled_with_authenticator(self):
 
     @api.model
     def check_credentials(self, password):
-        try:
+        """Add MFA logic to core authentication process
```
lasley: Should end your docblocks with periods ([PEP-0257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) Single line docstring note 4). 

The bullet points should also have punctuation, because they are sentences that complete the introductory stem (`Overview:`)

auth_totp/migrations/10.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,36 @@
...
+from odoo import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    """Generate cookie keys for all users with MFA enabled and clean up"""
```
lasley: Period at end of docblock sentence

auth_totp/models/res_users.py
```diff
@@ -50,53 +56,38 @@ def _check_enabled_with_authenticator(self):
...
-                'Your MFA login token is not valid. Please try again.'
-            ))
 
-        token_exp = fields.Datetime.from_string(self.mfa_login_token_exp)
-        if token_exp < datetime.now():
-            raise MfaTokenExpiredError(_(
-                'Your MFA login token has expired. Please try again.'
-            ))
+        if request:
+            if request.session.get('mfa_login_active') == user.id:
```
lasley: `mfa_login_active` string key seems pretty integral to things, and is repeated in a lot of spots. Typically I try to make strings into constants once I've used them 3-4 times (not counting tests). Do we have a good place to do this? Maybe in MFA model?

lasley (changes requested):
Thanks @obulkin - I like this design much better. Comments inline

auth_totp/models/res_users.py
```diff
@@ -50,53 +56,38 @@ def _check_enabled_with_authenticator(self):
...
+        Overview:
+            * If user does not have MFA enabled, defer to parent logic
+            * If user has MFA enabled and has gone through MFA login process
+              this session or has correct device cookie, defer to parent logic
+            * If neither of these is true, call parent logic. If successful,
+              prevent auth while updating session to indicate that MFA login
+              process can now commence
+        """
+        user_model_sudo = self.sudo()
+        user = user_model_sudo.search([('id', '=', self.env.uid)])
```
fmdl: `user = self.sudo().browse(self.env.uid)`

fmdl:
It could great that a basic user cannot desactivate the MFA, for more security.
```
@multi
def write(self, vals):
    if 'mfa_enable' in vals.keys():
         if vals['mfa_enable'] == False and not self.user_has_group('base.group_user_root'):
              raise UserError(_('You have not allowed to desactivate MFA'))
```

lasley:
> It could great that a basic user cannot desactivate the MFA, for more security.

This would be a security consideration that a company administrator would need to make as an option IMO. At the moment, there aren't really any options for things like that, so this would expand the scope of this bug fix quite drastically I think. 

@obulkin - we have an internal ticket for it already I think, but would be good to add to roadmap as well.

auth_totp/models/res_users.py
lasley: Why not `self.env.user.sudo()`?

auth_totp/controllers/main.py
obulkin: The sentence is referring to the ``login`` value in the ``session`` dictionary.

auth_totp/models/res_users.py
obulkin: This doesn't really seem like a substantive improvement, especially since two lines will be needed, so I'm leaving it as is.

auth_totp/models/res_users.py
obulkin: Actually, it turns out that ``env.user`` already [bypasses access restrictions](https://github.com/odoo/odoo/blob/10.0/odoo/api.py#L796), so I'll just drop these lines entirely and use that as needed. Thanks for the push in the right direction.

auth_totp/models/res_users.py
obulkin: This key is only used in two places outside of tests, and any constant used to replace it will need to be just as verbose, so I don't think anything is gained by doing that.

obulkin:
I've updated this PR to address the comments made so far.

@lasley - There is already an item in the README roadmap that effectively covers this request to prevent MFA from being disabled:

> * Add company-level settings for forcing all users to enable MFA and disabling 
  the trusted device option

I agree that this would be a substantial new feature and doesn't really belong in a bugfix PR.

obulkin:
Travis is currently failing due to unrelated issues with the ``sentry`` module.

lasley (approved):
Changes look good, thanks @obulkin. Confirmed Travis unrelated

fmdl:
Hi guys, when it is merge ?

lasley:
@fmdl - [Two review approvals are required for merge](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#please-respect-a-few-basic-rules). There is only one here (mine). If you would like to see the merge faster, please review the work and approve if satisfactory.

fmdl:
Dear @lasley, 

Your update doesn't prevent an RPC acces.

This `uid = common.authenticate(db, username, password, {})` return `False`, for authentification it is works.

But if you know the uid (it not hard to find) and the password, RPC access is allowed:
`products = models.execute_kw(db, 1, password,'product.product', 'search', [[]], {'limit':1})`

The best way is to put the password in a new field

fmdl (changes requested):
Your update doesn't prevent an RPC acces.

This uid = common.authenticate(db, username, password, {}) return False, for authentification it is works.

But if you know the uid (it not hard to find) and the password, RPC access is allowed:
products = models.execute_kw(db, 1, password,'product.product', 'search', [[]], {'limit':1})

obulkin:
@fmdl - Nice find. This only happens if there is a successful web login for the MFA user prior to the RPC request and is caused by a caching mechanism that skips the call to ``check_credentials``: https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/res/res_users.py#L503. Changing the password field would not help, but I do have a solution in mind and will soon update this PR with the fix.

fmdl:
It seems there are a bug in odoo, see here : https://github.com/odoo/odoo/pull/22212

fmdl (approved):

obulkin:
@fmdl - I've pushed up a commit that mostly closes this RPC security hole, but it does have a limitation. Because the cache is implemented as a class attribute, it is not synced up across threads/processes. This means that when a user first enables MFA, there is no way to ensure that the cache states of every thread/process reflect this change until the server has been restarted (in the single-threaded, single-process scenario, just logging out and logging back in is enough). The only way around this involves accessing the DB in ``check`` and would offset the performance benefits of the caching, so we've decided to leave it as a known issue for now.

lasley (approved):
I'm good with this, thanks @obulkin

obulkin:
@pedrobaeza Just wanted to pop in here to recommend merging this PR. It closes some significant security holes in the module and simplifies a lot of the logic, which should make it easier for others to maintain and upgrade the code.

MiquelRForgeFlow (approved):
I'm interested in migrating this module to v11

PR #1072, opened 2017-11-27
[FIX] Underscore is a special character

StefanRijnhart:
For legacy reasons, \W does not match the underscore (see https://stackoverflow.com/questions/35942067/why-cant-the-underscore-be-matched-by-w-in-python). I would argue that the underscore is a special character and should be interpreted as such by this module.

### DIFF
--- a/password_security/__openerp__.py
+++ b/password_security/__openerp__.py
@@ -5,7 +5,7 @@
 
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
-    'version': '8.0.1.1.0',
+    'version': '8.0.1.1.1',
     'author': "LasLabs, Odoo Community Association (OCA)",
     'category': 'Base',
     'depends': [
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -78,7 +78,7 @@ class ResUsers(models.Model):
         if company_id.password_numeric:
             password_regex.append(r'(?=.*?\d)')
         if company_id.password_special:
-            password_regex.append(r'(?=.*?\W)')
+            password_regex.append(r'(?=.*?[\W_])')
         password_regex.append('.{%d,}$' % company_id.password_length)
         if not re.search(''.join(password_regex), password):
             raise PassError(_(self.password_match_message()))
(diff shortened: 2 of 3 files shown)


lasley (approved):
TIL thanks @StefanRijnhart. Can you please bump module by a fix version? (I forget all the time too - I swear it must be easier to notice as a reviewer or something)

StefanRijnhart:
Version number bumped. I see conflicts are built-in with constant version bumping. Let me rebase.

StefanRijnhart:
I'm beginning to think that version bumping is more a hassle than a boon to be honest. But then again, I pin my versions with git hashes, not these fancy pip installs.

StefanRijnhart:
To clarify: having to rebase means I cannot use this PR branch for my deployment.

StefanRijnhart:
It also means I cannot cherry pick more than one other person's PR branches for the same module :-(

lasley:
> I'm beginning to think that version bumping is more a hassle than a boon to be honest.

I agree personally. Realistically this only affects things when there are migrations that Odoo needs to run.

I'd +1 a PR to our contribution guidelines to remove version bumping unless migrations are involved. I dunno if the rest of the community would agree on that though, I think we might be nonconformist on that. Our updates hinge on hashing as well (although of the module contents). This is not semantic though, and only good for automation.

_Edit: words_

StefanRijnhart:
Version bumping continued here: https://github.com/OCA/maintainer-tools/issues/314

zamberjo (approved):

password_security/__openerp__.py
```diff
@@ -6,6 +6,7 @@
     'name': 'Password Security',
     "summary": "Allow admin to set password security requirements.",
     'version': '8.0.1.1.0',
```
lasley: Looks like the conflict resolution failed

StefanRijnhart: Thanks for noting! Fixed.

hparfr (approved):

lasley (approved):

lasley:
Red build is unrelated. Merging.

lasley:
@StefanRijnhart - do you plan on forward porting this?

StefanRijnhart:
Yes,will do.

PR #1079, opened 2017-11-30
[FIX] Adapt to newer versions of python-ldap and lxml

StefanRijnhart:
Also, mute error level log messages from false positives


pedrobaeza:
Ey, we have it! Green in codecov/patch also

StefanRijnhart:
Yes, thank you so much!

pedrobaeza (approved):

PR #1080, opened 2017-11-30
[8.0][sql_export] Allow to use many2one and many2many fields for query variables

florian-dacosta:
It is already possible to use variables into the sql queries, using simple fields (integer, float, char...).
I just improve a little to make possible the use of many2one and many2many.

@bguillot @aheficent


AaronHForgeFlow (approved):
:+1: Only code review

LRovira-PlanetaTIC (approved):
👍 It works perfectly

Issue #1083, opened 2017-12-03
[10]How to turn off the automatic reset password function


lonelyleaves:
After installing password security, will be forced to request a reset password after login, how to close, thank you！

pedrobaeza:
@lasley I'm a bit annoyed about this also. If I have a user with password, I don't want to reset my password next time I login

lasley:
Is this for all users? Basically all existing passwords are considered expired?

pedrobaeza:
Yes

lonelyleaves:
Can you explain the function of this reset password I did not understand

lasley:
The point of the password reset is to allow for the expiry of passwords based on criteria, which is intended functionality for this module and described in depth in the ReadMe. 

Fix sounds simple enough- will just need to add a post install hook to set the last reset date of all users to the current date. I’m assuming it’s just blank in the db, which is going to trigger the expiry. 


-Dave Lasley

pedrobaeza:
I understand the principle, but mixed with the fact that you can't repeat a password makes everyone to change the password, even if the old one is secured enough. I prefer to not reset it.

lasley:
@pedrobaeza - yeah I figured you got the principle, but @lonelyleaves was asking so I elaborated.

PR #1084

lasley:
I think this actually also affected new users. Not positive though & wasn't necessarily worth checking.

rven:
This will also fix the issue I've got.
When installing the module on a DB all users passwords are expired.
But when logging in, the login attempt fails and nothing happens, the pop-up is not shown but redirected to the login page again.
It is not possible to login to Odoo after installing this module in an existing environment.
I fixed it with the following update: `update res_users set password_write_date=current_date;`
Exactly the same what is done in this Bug Fix.
:+1:

lasley:
Thanks for the elaboration @rven. You were amazingly specific, so it's obvious that you looked at #1084 and approved it. In the future though, please add the approval comment to the PR itself so that we can count it easier 🚀

lasley:
* Merged in v10
* #1088 for v9
* https://github.com/OCA/server-auth/pull/3#issuecomment-349702117 for v11

PR #1084, opened 2017-12-04
[FIX] password_security: Default last write date

lasley:
* Add a default `password_write_date` to circumvent need for immediate password reset, fixing #1083

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -21,6 +21,7 @@ class ResUsers(models.Model):
 
     password_write_date = fields.Datetime(
         'Last password update',
+        default=fields.Datetime.now,
         readonly=True,
     )
     password_history_ids = fields.One2many(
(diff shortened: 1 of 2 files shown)


password_security/models/res_users.py
```diff
@@ -21,6 +21,7 @@ class ResUsers(models.Model):
 
     password_write_date = fields.Datetime(
         'Last password update',
+        default=lambda s: fields.Datetime.now(),
```
pedrobaeza: Shouldn't be this without final `()`?

lasley: Derp yes I totally didn't need the lambda here. Good call

pedrobaeza (approved):

LoisRForgeFlow (approved):
:+1:

lasley:
I'm going to count https://github.com/OCA/server-tools/issues/1083#issuecomment-349662556 from @rven as a review here, and quick merge.

PR #1085, opened 2017-12-04
[9.0][ADD] base_tier_validation

LoisRForgeFlow:
Base Tier Validation
====================

This module does not provide a functionality by itself but an abstract model
to implement a validation process based on tiers on other models (e.g.
purchase orders, sales orders...).

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -1,0 +1,73 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+====================
+Base Tier Validation
+====================
+
+This module does not provide a functionality by itself but an abstract model
+to implement a validation process based on tiers on other models (e.g.
+purchase orders, sales orders...).
+
+**Note:** To be able to use this module in a new model you will need some
+development.
+
+See `purchase_tier_validation <https://github
+.com/OCA/purchase-workflow>`_ as an example of implementation.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Settings > Technical > Tier Validations > Tier Definition*.
+#. Create as many tiers as you want for any model having tier validation
+   functionality.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/149/9.0
+
+Known issues / Roadmap
+======================
+
+* In odoo v11 it would be interesting to try to take advantage of ``mail.activity.mixin``.
+
+Bug Tracker
+===========
+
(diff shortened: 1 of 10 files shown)


lasley:
Interesting submission @lreficent. I think I'm missing a core principle, which I don't see on quick glance - what is a tier in this context?

LoisRForgeFlow:
@lasley The purpose is to provide a way to define several validation tiers or steps, e.g: Assuming we are validating a PO:

* Tier 1: You want to set a supervisor to check all PO.
* Tier 2: You want the CEO to review only POs over a certain amount. Note that this validation won't be always needed and (not sure yet) you will want to validate the tier 1 and then ask for the tier 2 validation.

Your thoughts and insights would be welcomed!

Edit: typo

lasley:
Just to make sure I'm understanding then, the tier is basically a workflow for a validation escalation on a specific model? 

So in the use case of the PO you described, you would maybe set a dollar amount limit that would provide for escalation of validation from the tier X agent to the tier X+1 agent?

_Edit: word_

LoisRForgeFlow:
@lasley yes, that's the idea, you can set `tier.definition` records with a python condition to be evaluated, if the condition is true in the context of a specific record to be validated (PO for example) that validation will be required. At the moment, `tier.definition` has a sequence field that will determinate the tier number within a specific PO.

I will submit now 'purchase_tier_validation' where you will be able to actually test the functionality.

yajo:
According to your description, it reminds me to [odoo v11's `base_automation`](https://github.com/odoo/odoo/blob/59729695680b33a1eec8e170777e07ee12cb3925/addons/base_automation/__manifest__.py#L9-L16).

Do you think it would fit your purpose? If so, a backport would be more maintentance-friendly, since you'd only have to support the addon in v9 and 10.

LoisRForgeFlow:
@yajo I think it wouldn't cover exactly my usecase. The purpose of this module is to make an operation (usually the confirmation, draft -> confirmed) in some specific document (like SO, PO, invoice...) pass through a validation process, during so, the operation and edition of the document would be locked (what I think is not supported by base_automation). This validation can be different depending on the status of the document and you would require a different validation type (maybe some specific user validates big POs and some specific group validates POs from some specific country). Also, the operation may need the validation of more than one person. With base_automation you would be likely able to set up similar rules, perhaps for sending an email to people involved in the validation process but then this process would be kind of manual, this module also tries to support the workflow of the validation process itself.

AaronHForgeFlow (approved):
:+1: It works fine with https://github.com/OCA/purchase-workflow/pull/481

andhit-r:
Give me an exception when i test it with https://github.com/OCA/purchase-workflow/pull/481. See https://github.com/OCA/purchase-workflow/pull/481#issuecomment-372547608

andhit-r (commented):
@lreficent no method to restart validation?

andhit-r (approved):

PR #1087, opened 2017-12-05
[10.0][FIX][IMP] base_kanban_stage: Default stage and tests

obulkin:
* Add default ``stage_id`` logic that returns lowest sequence stage associated with inheriting model and update unit tests accordingly
* Improve ``_read_group_stage_ids`` unit test
* Implement latest abstract model testing strategy


lasley (approved):
Thanks @obulkin

yajo (approved):

PR #1090, opened 2017-12-10
[FIX] import for safe_eval

hbrunn:
this works during tests because there `tools.safe_eval` is the function, but to have this work after Odoo has been fiddling with the imports (and during tests), we must use the canonical way to import this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - python-lxml # because pip installation is slow
       - unixodbc-dev
       - python-mysqldb
+      - postfix # some tests want a local mail server
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

hbrunn:
should we install a mailserver on travis, or should we rather not force_send during tests?

pedrobaeza:
It would be better to have a fake server I think.

cc @gurneyalex

hbrunn:
runbot is another issue, this is about travis. I'll have a look at the travis part today probably

pedrobaeza:
Oh, you're right. Again it would be great to have a Docker image for both for not having to dual think things...

hbrunn:
@pedrobaeza travis passes now, runbot can't be fixed from here. Did I earn a fast track by making the branch green with the other two commits? (rebase+merge is the way to go here imho)

hbrunn:
PS: I'm too old for the docker hipsterism...

pedrobaeza:
You got it!

P.S.: Come to the Docker side, young Padawan!

PR #1098, opened 2017-12-18
[RFC|ADD] infrastructure: Connect to infra services

lasley:
This module will provide the base models and (if necessary) methods that are required to connect to an infrastructure service. I'm trying to be as generic as possible here, with the basic breakdown being:

* Environment - Grouping for hosts
* Hosts - Servers (virtualized or physical)
* Volumes, Mounts, Filesystems - self-explanatory
* Stack - Grouping for Services
* Service - A configuration for instance deploy/scaling
* Instance - Application instance or Docker container

I'm building a Rancher connector on top of this, so we should see some data structures shifting as I identify things I missed. I'm submitting this early to primarily serve as an RFC in case someone has some thoughts on my overall design. I don't think I have any specific questions at the moment.


lasley:
After some thought, I renamed this to `infrastructure` & made this an application. While it doesn't provide functionality on its own, I think this is probably a good move that would allow us to also add config settings for other module installations (infrastructure-dns, etc.)

lasley:
It might also make sense to move all of the current models, other than `infrastructure_environment`, into some other module meant for the management of the server & application level infrastructure. We could then start the config paradigm by providing a "Manage Server Infrastructure" config option

lasley:
As I'm building this, I also realize that a secondary usage is without connectors. This module will also allow you to build an inventory of your infrastructure, such as for use with license audits.

PR #1103, opened 2017-12-22
[9.0][FIX] Auth_totp : ensure no special chars are passed to totp.provisioning_uri

grindtildeath:
When a user name has special character, opening the wizard will fail with such an error :
```
tools/auth_totp/wizards/res_users_authenticator_create.py", line 78, in _compute_qr_code_tag
    issuer_name=record.user_id.company_id.display_name,
  File "/usr/local/lib/python2.7/dist-packages/pyotp/totp.py", line 88, in provisioning_uri
    digits=self.digits, period=self.interval)
  File "/usr/local/lib/python2.7/dist-packages/pyotp/utils.py", line 61, in build_uri
    label = quote(name)
  File "/usr/lib/python2.7/urllib.py", line 1294, in quote
    return ''.join(map(quoter, s))
KeyError: u'\xee'
```
As described here : https://github.com/pyotp/pyotp/issues/21

With this fix, any special character will be escaped and the QR code will be generated.

### DIFF
--- a/auth_totp/wizards/res_users_authenticator_create.py
+++ b/auth_totp/wizards/res_users_authenticator_create.py
@@ -74,7 +74,7 @@ class ResUsersAuthenticatorCreate(models.TransientModel):
 
             totp = pyotp.TOTP(record.secret_key)
             provisioning_uri = totp.provisioning_uri(
-                record.user_id.display_name,
+                record.user_id.display_name.encode('utf-8'),
                 issuer_name=record.user_id.company_id.display_name,
             )
             provisioning_uri = urllib.quote(provisioning_uri)
--- a/auth_totp/__openerp__.py
+++ b/auth_totp/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'MFA Support',
     'summary': 'Allows users to enable MFA and add optional trusted devices',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'category': 'Extra Tools',
     'website': 'https://laslabs.com/',
     'author': 'LasLabs, Odoo Community Association (OCA)',


grindtildeath:
@obulkin Would be great to have your feedback on this.

ghost (commented):
LGTM!

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -73,8 +73,10 @@ def _compute_qr_code_tag(self):
                 continue
 
             totp = pyotp.TOTP(record.secret_key)
+            user_dname = urllib.quote(
```
obulkin: This URL encoding step is not needed because PyOTP already [calls ``quote``](https://github.com/pyotp/pyotp/blob/master/src/pyotp/utils.py#L61) as part of the URI creation process (this is why UTF-8 is not supported in the first place). Calling it twice causes the name to show up incorrectly in my MFA app because it is not fully decoded.

obulkin (changes requested):
Thanks for the fix, @grindtildeath. Please see my comment in line.

auth_totp/wizards/res_users_authenticator_create.py
lasley: Ahh good ol' Unicode.

Wouldn't we just want to pass `record.user_id.display_name.encode('utf-8')` to `totp.provisioning_url`?

lasley: That said, this encoding (`utf-8`) would be specific to the underlying encoding defined for the PSQL database, right? I wonder if we have an easy way to determine that from Odoo.

obulkin: @lasley - I don't think the DB encoding matters since the relevant Odoo field should always return a ``unicode`` object or raise an exception, and a ``unicode`` object can always be encoded using UTF-8 (in v9, the logic for ensuring the former is [here](https://github.com/odoo/odoo/blob/9.0/openerp/loglevels.py)).

A bigger question is whether the user's MFA device/app will be able to properly decode UTF-8. I tried this with LastPass Authenticator and Duo Mobile, which both worked fine, and I suspect that most will. However, if we're really worried about compatibility, we could encode to ASCII or Latin1 and only fall back to UTF-8 if that fails.

obulkin (approved):
This looks good to go now. Please note that you'll need to squash down to a single commit before this can be merged.

lasley (approved):
Thanks @grindtildeath 

Do you plan on forward porting this to v10?

(Note it's not applicable on v11 because Python3 uses Unicode internally)

grindtildeath:
@obulkin @lasley Commits squashed and I directly forward-ported it to 10.0.

lasley:
THanks @grindtildeath

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -74,7 +74,7 @@ def _compute_qr_code_tag(self):
 
             totp = pyotp.TOTP(record.secret_key)
             provisioning_uri = totp.provisioning_uri(
-                record.user_id.display_name,
+                record.user_id.display_name.encode('utf-8'),
                 issuer_name=record.user_id.company_id.display_name,
```
NothingCtrl, line 78: The company name should encode too

Issue #1110, opened 2017-12-28
[11.0][base_exception] Gives errors when depending modules are proposed


JordiBForgeFlow:
When proposing module 'purchase_exception' in https://github.com/OCA/purchase-workflow/pull/467, the testing process fails with the errors: 

`2017-12-28 08:53:08,294 5408 ERROR openerp_test odoo.modules.registry: Model base.exception.test.purchase.line has no table.
2017-12-28 08:53:08,294 5408 ERROR openerp_test odoo.modules.registry: Model base.exception.test.purchase has no table.`

We already detected during the review of PR related to base_exception:
https://github.com/OCA/server-tools/pull/1025#discussion_r155192855

But somehow it was silently missed during the review process.

PR #1111, opened 2017-12-28
[11.0][base_exception] refactor tests

JordiBForgeFlow:
Fixes https://github.com/OCA/server-tools/issues/1110


JordiBForgeFlow:
@mourad-ehm can you review?

base_exception/tests/test_base_exception.py
```diff
@@ -2,43 +2,116 @@
...
 
-        self.base_exception = self.env['base.exception']
-        self.exception_rule = self.env['exception.rule']
-        self.exception_confirm = self.env['exception.rule.confirm']
+    @classmethod
+    def setUpClass(cls):
+        super(TestBaseException, cls).setUpClass()
 
-        self.exception_rule._fields['rule_group'].selection.append(
+        class PurchaseTest(models.Model):
```
simahawk: I'd move this to a separated file anyway and here run only `init_test_models`

base_exception/tests/test_base_exception.py
```diff
@@ -47,18 +120,18 @@ def setUp(self):
...
             'name': "Qty > 0",
             'sequence': 10,
             'rule_group': "test_base",
             'model': "base.exception.test.purchase.line",
             'code': """if test_base_line.qty <= 0:
         failed=True"""})
 
-    def test_sale_order_exception(self):
-        partner = self.env.ref('base.res_partner_1')
+    def test_sale_order_exception(cls):
```
simahawk: why is this wrapped into setUpClass?

base_exception/tests/test_base_exception.py
```diff
@@ -2,43 +2,116 @@
...
 from odoo.tests import common
+from odoo import api, fields, models
+from .common import init_test_models
 
 import logging
 
 _logger = logging.getLogger(__name__)
 
 
-# @common.at_install(False)
```
simahawk, line 11: so this trick plays nice only if you run the whole test suite at install I guess?

simahawk (changes requested):

base_exception/tests/test_base_exception.py
JordiBForgeFlow: This is the way I've seen it being done in https://github.com/OCA/mis-builder/blob/10.0/mis_builder/tests/test_kpi_data.py#L19, and seems to be no issue testing during install, which is the default method and easiest to verify.

JordiBForgeFlow:
@simahawk Many thanks for your review. I have attended to your comments.

base_exception/tests/test_base_exception.py
JordiBForgeFlow: No need. I fixed

base_exception/tests/test_base_exception.py
JordiBForgeFlow: Done.

base_exception/tests/test_base_exception.py
simahawk: hmm then I guess that if for any reason we'll need to run a test after installation we could use a different class and make sure this one is loaded before

simahawk (approved):
@jbeficent can you squash?

simahawk:
@mourad-ehm ping :)

mourad-ehm (approved):

mourad-ehm:
Hi @jbeficent,  Thanks for this PR

JordiBForgeFlow:
@mourad-ehm @simahawk I have now cleaned up in e3658cc715b132643af2b65c6286d3c4f6178a48 the way to create and tear down test classes, following the proposal made by @sbidoul in https://github.com/OCA/mis-builder/pull/67

mourad-ehm:
OK :+1:

JordiBForgeFlow:
@simahawk squashed commits. Will you merge now?

simahawk:
@jbeficent I wish I could do it :)
Nevertheless, pls, make the commit more explicit by including the module name (as you did for the PR title), tnx :)

Issue #1096, opened 2017-12-14
[10.0] sentry: Tests failing


obulkin:
Two of the unit tests in the ``sentry`` module are currently failing, as can be seen here: https://travis-ci.org/OCA/server-tools/jobs/316634324. I've replicated this issue locally using the 10.0 branch and confirmed that it is not related to ``auth_totp``.

@lasley

naglis:
I think this is due to the fact that the [`FIELDS` class attribute](https://github.com/OCA/server-tools/blob/3bb12bd80cb9c6409521eb54025d12da9aa210cc/sentry/logutils.py#L104) has been [renamed to `KEYS`](https://github.com/getsentry/raven-python/blob/9f56a51ebfd67ee094d47a93fb049fe8b40ac0a4/raven/processors.py#L154) in [raven 6.4.0](https://github.com/getsentry/raven-python/releases/tag/6.4.0).

I think it might be enough to add a `KEYS` class attribute on [`SanitizeOdooCookiesProcessor`](https://github.com/OCA/server-tools/blob/3bb12bd80cb9c6409521eb54025d12da9aa210cc/sentry/logutils.py#L97) pointing to `FIELDS` (to maintain backwards compatibility with older versions of Raven).

I will test it and hopefully make a fix tomorrow.

This should affect #1023 as well.

naglis:
@pedrobaeza, OK to close?

PR #1117, opened 2018-01-05
[10.0][FIX] sentry: fix tests after raven 6.4.0

naglis:
This should fix https://github.com/OCA/server-tools/issues/1096.

### DIFF
--- a/sentry/__manifest__.py
+++ b/sentry/__manifest__.py
@@ -4,7 +4,7 @@
 {
     'name': 'Sentry',
     'summary': 'Report Odoo errors to Sentry',
-    'version': '10.0.1.0.0',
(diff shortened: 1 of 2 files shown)


lasley (approved):
Thanks @naglis 

I'm quick merging this one to fix our build

lasley:
@naglis - I notice you use GPG. Would you like to squash this yourself before merge so I don't destroy your sig?

naglis:
@lasley I've squased the commits, however as only my work email has OCA CLA signed I would only be able to sign it on Monday (d'oh!). I don't mind if you'd merge it before then.

lasley:
Works for me, thanks @naglis!

PR #1121, opened 2018-01-08
[10.0][Fix] Auth_totp: Ensure no special chars are passed to totp.provisioning_uri

grindtildeath:
Forward port of https://github.com/OCA/server-tools/pull/1103


lasley (approved):
Thanks @grindtildeath 

I'm quick merging this one - it's a bug fix and identical to the v9 one that went through proper process.

Issue #1008, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Repository split
As talked in [this mailing list thread](https://odoo-community.org/groups/tools-22/tools-44031?mode=thread) this repository is very crowded, so for 11.0, an split has been done across several repositories. Please check current spreadsheet for the initial distribution plan. If there's a module not included here or that you have doubts, please ask in this issue:

https://docs.google.com/spreadsheets/d/1Xg95cW4TFMf_Lo5i_CZC_qOOfN8RgxPRc0LJTLTkdUI/edit?usp=sharing

# Modules to migrate

- [ ] admin_technical_features
- [ ] attachment_base_synchronize
- [x] auditlog - By @etobella - #1040 
- [ ] auth_admin_passkey
- [x] auth_brute_force - By @Yajo - https://github.com/OCA/server-auth/pull/19
- [x] auth_dynamic_groups - Moved to OCA/server-auth - By @NL66278 - OCA/server-auth#98
- [ ] auth_from_http_basic
- [ ] auth_from_http_basic_logout
- [x] auth_from_http_remote_user - By @TDu - OCA/server-auth#7
- [x] auth_session_timeout - By @zaoral - OCA/server-auth#32
- [x] auth_signup_verify_email - @simahawk OCA/server-auth#8
- [ ] auth_supplier
- [x] auth_totp - By @mreficent - OCA/server-auth#17
- [x] auth_totp_password_security - By @mreficent - OCA/server-auth#18
- [x] auth_user_case_insensitive - By @mgosai - OCA/server-auth#35
- [x] auto_backup - By @astirpe - #1158 
- [x] base_cron_exclusion - By @bodedra - in #1156 
- [ ] base_custom_info - Improve usability with polymorphic widget?
- [x] base_exception - By @mourad-ehm - #1025 
- [x] base_export_manager - By @Yajo - https://github.com/OCA/server-ux/pull/22
- [x] base_external_dbsource - By @sergio-teruel - OCA/server-backend#9
- [ ] base_external_dbsource_firebird
- [x] base_external_dbsource_mssql - By @sergio-teruel - OCA/server-backend#6
- [x] base_external_dbsource_mysql - By @sergio-teruel - OCA/server-backend#7
- [ ] base_external_dbsource_odbc
- [ ] base_external_dbsource_oracle
[...]

PR #1124, opened 2018-01-09
[11.0] [MIG] Base technical features

ventilooo:


pedrobaeza:
Module is already migrated in https://github.com/OCA/server-ux/pull/2.

Please follow migration issue #1008 where everything is tracked.

PR #1129, opened 2018-01-14
[DEV]mass_editing: Migrated 10.0 to 11.0

codewareuae:


pedrobaeza:
There's already a migration in progress in https://github.com/OCA/server-ux/pull/5. Please review that one instead, and also check migration issue for tracking the progress of each module: https://github.com/OCA/server-tools/issues/1008

PR #1130, opened 2018-01-14
[NEW] base_export_xml

fmdl:
Dear Guy

It a module to export record in XML format, to be use to hardcoded data (for customer module, for a migration, ...)

What do you think ?

image

image

@pedrobaeza 
@sylvain-garancher


pedrobaeza:
This is already done in #1045, as part of the module_prototyper.

fmdl:
@alnslang what do you think ?

PR #1142, opened 2018-02-01
[11.0] [mig] [WIP] base custom info 

fanha99:
WIP
checking odoo bug


base_custom_info/models/custom_info.py
```diff
@@ -30,21 +29,6 @@ class CustomInfo(models.AbstractModel):
         auto_join=True, string='Custom Properties',
     )
 
-    # HACK: Until https://github.com/odoo/odoo/pull/10557 is merged
```
fanha99: with or without this block of code, still the same result.

base_custom_info/models/custom_info.py
```diff
@@ -60,9 +44,6 @@ def _onchange_custom_info_template_id(self):
                 "res_id": self.id,
                 "value": prop.default_value,
             })
-            # HACK https://github.com/odoo/odoo/issues/13076
```
fanha99: is this issue fixed in v11?

pedrobaeza:
I think this patch is related, but for v12: https://github.com/odoo/odoo/commit/fff578fa22530aa66081cc27e72f70f996fe9c6a

pedrobaeza:
Why do you close it?

PR #1144, opened 2018-02-05
[10.0][MIG] users_ldap_groups

hbrunn:


users_ldap_groups/tests/test_users_ldap_groups.py
```diff
@@ -0,0 +1,65 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from mock import Mock, patch
+from odoo.tests.common import TransactionCase
+
+
+@patch('ldap.initialize', return_value=Mock(
+    search_st=Mock(return_value=[
+        ('cn=hello', {'name': ['hello', 'hello2']})
+    ]),
+))
```
sebalix, line 12: I'm not sure but this patch call will affect the whole Python runtime, no? Would not be better to use the `start()` method in `setUp()` and `stop()` in `tearDown()`?

hbrunn: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

sebalix: Oups sorry, didn't see it was a decorator on the test class.

sebalix (approved):

users_ldap_groups/tests/test_users_ldap_groups.py
hbrunn: thanks for the review, nothing to apologize!

NL66278 (approved):
:smile: LGTM and tested in practice

PR #1145, opened 2018-02-07
[ADD] Added new module module_upgrade_link

wpichler:


pedrobaeza:
There's already another module for this: https://github.com/OCA/server-tools/tree/10.0/module_auto_update.

You can't duplicate functionality inside OCA. Please check the other one and you can propose enhancements to that one.

PR #1146, opened 2018-02-07
[FIX] base_technical_features: Test in post-install

yajo:
Creating a user in a test leads to errors in integration tests. Moving test to post install mode.

@Tecnativa


pedrobaeza (approved):
We can put all tests in post without problem

sebalix (approved):

PR #983, opened 2017-09-14
[FIX][auth_signup_verify_email] Remove error from logs

yajo:
This fixes #981.

@Tecnativa

PR #1148, opened 2018-02-08
[FIX] html_image_url_extractor: Do not test the specific exception

yajo:
Newer versions of LXML raise different exceptions.

It doesn't matter, as long as there's an exception, so we move to less explicit tests.

Besides, I mute the logger to avoid expected failure error logging.

@Tecnativa


pedrobaeza:
This is given red Travis

yajo:
Rebased to fix conflicts from #983

rafaelbn (approved):

pedrobaeza (approved):

PR #1149, opened 2018-02-08
[FIX] html_text: Do not test the specific exception

yajo:
Mostly the same as #1148 

Newer versions of LXML raise different exceptions.

It doesn't matter, as long as there's an exception, so we move to less explicit tests.

Besides, I mute the logger to avoid expected failure error logging.

@Tecnativa

### DIFF
--- a/html_text/tests/test_extractor.py
+++ b/html_text/tests/test_extractor.py
@@ -53,10 +53,3 @@ class ExtractorCase(TransactionCase):
         self.assertEqual(self.text_from_html(False), "")
         with self.assertRaises(Exception):
             self.text_from_html(False, fail=True)
-
-    @mute_logger("openerp.addons.html_text.models.ir_fields_converter")
-    def test_bad_html(self):
-        """Bad HTML handled correctly."""
-        self.assertEqual(self.text_from_html("<<bad>"), "")
-        with self.assertRaises(Exception):
-            self.text_from_html("<<bad>", fail=True)
--- a/html_text/tests/test_extractor.py
+++ b/html_text/tests/test_extractor.py
@@ -2,8 +2,8 @@
 # © 2016 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from lxml import etree
 from openerp.tests.common import TransactionCase
+from openerp.tools import mute_logger
 
 
 class ExtractorCase(TransactionCase):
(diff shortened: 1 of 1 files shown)


html_text/tests/test_extractor.py
```diff
@@ -40,20 +40,16 @@ def test_excerpts(self):
...
 
+    @mute_logger("openerp.addons.html_text.models.ir_fields_converter")
     def test_false_html(self):
         """``False`` HTML handled correctly."""
         self.assertEqual(self.text_from_html(False), "")
-        with self.assertRaises(TypeError):
+        with self.assertRaises(Exception):
             self.text_from_html(False, fail=True)
-
-    def test_bad_html(self):
```
pedrobaeza, line 55: Why removing this test?

yajo: Because on newer LXML versions, that code is considered "good" HTML (remember that those rules are less strict than XML ones) and thus it doesn't raise an error.

pedrobaeza: But isn't better to provide a bad HTML?

yajo: I guess yes, but not sure on what is now considered bad exactly... I opted for this quicker solution.

pedrobaeza: Well, we are reducing test scope. For sure multiple unclosed tags will be considered illegal.

yajo: Nope, it seems now it eats anything:

```py
>>> html.fromstring("<<bad></p><a=neneeñ/>").xpath("//text()")
['<']
>>> html.fromstring("<<bad></p><a=neneeñ<ñ>/>").xpath("//text()")
['<', '/>']
>>> html.fromstring("<<bad></p><a=neneeñ<ñ>/>{}&al").xpath("//text()")
['<', '/>{}&al']
>>> html.fromstring("<<bad></p><a=neneeñ<ñ>/>{}&al;").xpath("//text()")
['<', '/>{}&al;']
>>> html.fromstring("<<bad></p><a=neneeñ<ñ>/>{}&x00185;").xpath("//text()")
['<', '/>{}&x00185;']
>>> html.fromstring("<<bad></p><a=neneeñ<ñ>/>{}&nbsp;").xpath("//text()")
['<', '/>{}\xa0']
```

It's a good thing, since HTML parsing should never raise exceptions, but then yes, we have to reduce the test scope now

pedrobaeza: OK, let's accept that

pedrobaeza (approved):
Fast-tracking this

PR #1152, opened 2018-02-12
[ADD] mass editing support for v11

ronozoro:


pedrobaeza:
There's already another migration in https://github.com/OCA/server-ux/pull/5. Please review that one instead.

PR #1153, opened 2018-02-14
[MIG] Date Range

open-net-sarl:
Remove useless encoding definition
Fixed tests by removing Unicode cast in wizard/date_range_generator.py at line 45


feketemihai:
Hi there, the module has been moved to server-ux repo, https://github.com/OCA/server-ux and the module is already migrated.

pedrobaeza:
Please always check issue for migration where we try to track everything.

PR #1160, opened 2018-02-15
[10.0][ADD] request_auditor

andreparames:
This module is used to log the requests from the browser, already having been pre-parsed by the HTTP layer of Odoo. It uses the Python logging infrastructure, so it can log to anywhere it can (e.g. files, logstash, etc).

An example of the configuration for logging as JSON-lines to a file (requires `pythonjsonlogger`):

    [audit_log]
    class = logging.FileHandler
    filename = /tmp/audit.log
    formatter_class = pythonjsonlogger.jsonlogger.JsonFormatter

An example for logstash (requires `python-logstash`):

    [audit_log]
    class = logstash.LogstashHandler
    host = localhost
    port = 5000

And here's an example of the output (pretty-printed):

```json
	{
	  "message": "AUDIT",
	  "uid": 1,
	  "params": {
	    "context_id": 1,
	    "model": "account.invoice",
	    "args": [
	      [
		3
	      ],
	      {
		"lang": "en_US",
		"tz": false,
		"uid": 1,
		"journal_type": "sale",
		"params": {
		  "menu_id": 106,
		  "view_type": "form",
		  "_push_me": false,
		  "action": 202,
		  "model": "account.invoice",
		  "id": 3
		},
		"type": "out_invoice"
	      }
	    ],
	    "method": "action_invoice_cancel",
	    "domain_id": null
	  },
	  "method": "call_button",
	  "timestamp": "2018-02-15T13:47:13.449715",
	  "dbname": "mydb",
	  "exception": null
	}
```

Suggestions are welcome!

### DIFF
--- a/request_auditor/__init__.py
+++ b/request_auditor/__init__.py
@@ -1,0 +1,50 @@
+# coding: utf-8
+from datetime import datetime
+
+import odoo
+
+from . import config
+from .config import MAX_VALUE_SIZE
+
+audit_logger = config.logger()
+filter_methods = config.filter_methods()
+
+
+def restrict_dict(o):
+    new = {}
+    for key, value in o.iteritems():
+        if isinstance(value, basestring) and 'passw' in key:
+            new[key] = 'PASSWORD_CENSORED'
+        elif isinstance(value, basestring) and len(value) > MAX_VALUE_SIZE:
+            new[key] = '%s...' % value[:MAX_VALUE_SIZE]
+        elif isinstance(value, dict):
+            new[key] = restrict_dict(value)
+        else:
+            new[key] = value
+    return new
+
+
+if audit_logger:
+    old_call_function = odoo.http.WebRequest._call_function
+
+    def call_function(self, *args, **kw):
+        try:
+            exception = None
+            return old_call_function(self, *args, **kw)
+        except Exception as e:
+            exception = repr(e)
+            raise
+        finally:
+            method = self.endpoint.method.__name__
+            if method in filter_methods:
+                log_record = {
+                    'dbname': self.db,
+                    'method': method,
+                    'params': restrict_dict(self.params),
+                    'uid': self.session.uid,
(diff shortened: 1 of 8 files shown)


andreparames:
Seems like the runbot is complaining because the module warns if there's no configuration; I think the warning should be kept, though, since the module is otherwise invisible.

request_auditor/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+from .config import MAX_VALUE_SIZE
+
+audit_logger = config.logger()
+filter_methods = config.filter_methods()
+
+
+def restrict_dict(o):
+    new = {}
+    for key, value in o.iteritems():
+        if isinstance(value, str) and len(value) > MAX_VALUE_SIZE:
```
sbidoul: unicode too?

request_auditor/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+            new[key] = '%s...' % value[:MAX_VALUE_SIZE]
+        elif isinstance(value, dict):
```
sbidoul: Hide values of keys containing `passw` like it is done in https://github.com/OCA/server-env/tree/11.0/server_environment#server-environment?

sbidoul (commented):
@andreparames something has to be done about runbot.

Could you disable the warning in test mode?

andreparames:
@sbidoul good suggestions, thanks. I've downgraded the log level when in test mode, should fix the runbot.

andreparames:
Apparently the runbot does not run in test mode, so I'm a bit at a loss. Should I check the database name for `*test*`?  Seems like an hack, but I don't see other way to detect it.

andreparames:
I think I got it: if the server is running with `stop-after-init`, then no requests will be logged anyway.

lmignon (approved):
@andreparames Thank you for this useful addon. LGTM (Code review no test)

PR #1162, opened 2018-02-16
[11.0][ADD] Converted mass_editing

Manuel-Dellisse:
[UPD]Replaced "basestring" with "str" (being removed from Python 3)
[UPD]Reworked "create_action" method ("ir.values" being removed from Odoo 11.0)
[UPD]Reworked "unlink_action" based on the same named method from "/mail/mail_template.py"


pedrobaeza:
mass_editing is migrated in https://github.com/OCA/server-ux/pull/5. Please review that one.

PR #1165, opened 2018-02-20
Backport improvements made to Keychain in 10.0

andreparames:
After the migration there has been some work done on the Keychain module in branch 10.0, this PR ports it back to branch 9.0


keychain/tests/test_keychain.py
```diff
@@ -21,6 +24,7 @@ def setUp(self):
 
         self.keychain = self.env['keychain.account']
         config['keychain_key'] = Fernet.generate_key()
+        config['running_env'] = ""
```
lmignon: @andreparames IMO this commit must be removed... Indeed, you reset the running_env before before to save its previous value at the next line.

lmignon (changes requested):

keychain/tests/test_keychain.py
andreparames: You're right, I took the "Mig V10" commit and removed the actual migration parts, without realizing what was left was no longer relevant. Will remove the commit.

lmignon (approved):
LGTM (Code review no test)

lmignon:
@andreparames Thank you for this backport!

ThomasBinsfeld (approved):
Code review and functional tests :+1:

PR #1168, opened 2018-02-22
[9.0] Keychain account caching

andreparames:
This commit creates a computed field on the keychain backend, so that repeated calls to `_get_keychain_account()` will return the value from the ORM cache rather than repeating the query each time.

This PR depends on the [backport of features from 10.0](https://github.com/OCA/server-tools/pull/1165).


keychain/models/keychain_backend.py
```diff
@@ -47,7 +54,7 @@ def _prepare_keychain(self):
     @api.multi
     def _get_keychain_account(self):
         self.ensure_one()
-        account = self._get_existing_keychain()
+        account = self.keychain_account_id
         if not account:
             vals = self._prepare_keychain()
             account = self.env['keychain.account'].create(vals)
```
lmignon: @andreparames you should assign the created account to `self.keychain_account_id`

lmignon (changes requested):

ThomasBinsfeld (approved):

lmignon (approved):
LGTM: (Code review only)

ThomasBinsfeld (approved):
Code and functional :+1:

PR #1169, opened 2018-02-22
[9.0][IMP] Load key from server_environment

andreparames:
Checks whether server_environment is installed and if so, tries to load the key from:
```
[keychain]
<env>=...
```

If it doesn't exist, it will fallback to the previous process.

### DIFF
--- a/keychain/models/keychain.py
+++ b/keychain/models/keychain.py
@@ -11,6 +11,12 @@ from openerp.exceptions import ValidationError
 from openerp.tools.config import config
 from openerp.tools.translate import _
 
+try:
+    from openerp.addons.server_environment import serv_config
+except ImportError:  # server_environment not installed or configured
+    serv_config = None
+
+
 _logger = logging.getLogger(__name__)
 
 try:
@@ -172,7 +178,7 @@ class KeychainAccount(models.Model):
         force_env = name of the env key.
         Useful for encoding against one precise env
         """
-        def _get_keys(envs):
+        def _get_keys_main_config(envs):
             suffixes = [
                 '_%s' % env if env else ''
                 for env in envs]  # ('_dev', '')
@@ -187,11 +193,30 @@ class KeychainAccount(models.Model):
                 if key and len(key) > 0  # remove False values
             ]
 
+        def _get_keys_serv_config(envs):
+            keys_name = [
+                env for env in envs if env]  # ignores empty env
+            keys_str = [
+                serv_config.get('keychain', key)
+                for key in keys_name]  # fetch from config
+            return [
+                Fernet(key) for key in keys_str  # build Fernet object
+                if key and len(key) > 0  # remove False values
+            ]
+
+        keys = []
+
         if force_env:
             envs = [force_env]
         else:
             envs = cls._retrieve_env()  # ex: ('dev', False)
-        keys = _get_keys(envs)
+
+        if serv_config and serv_config.has_section('keychain'):
+            keys = _get_keys_serv_config(envs)
+
+        if not keys:
+            keys = _get_keys_main_config(envs)
+
         if len(keys) == 0:
             raise Warning(_(
                 "No 'keychain_key_%s' entries found in config file. "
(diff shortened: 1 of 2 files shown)


keychain/models/keychain.py
```diff
@@ -187,11 +193,14 @@ def _get_keys(envs):
                 if key and len(key) > 0  # remove False values
             ]
 
-        if force_env:
-            envs = [force_env]
+        if serv_config and serv_config.has_option('keychain', 'key'):
+            keys = [Fernet(serv_config.get('keychain', 'key'))]
```
hparfr: why not using MultiFernet like in the last instruction of this method ?

keychain/models/keychain.py
```diff
@@ -187,11 +193,14 @@ def _get_keys(envs):
                 if key and len(key) > 0  # remove False values
             ]
 
-        if force_env:
```
hparfr: I think this should be put in _retrieve_env() instead

hparfr (changes requested):
Hi @andreparames 

Thanks for this PR.
You can also remove the Known issue / Roadmap entry about server_environnement

keychain/models/keychain.py
andreparames: How would that work? `_retrieve_env()` only returns the environment name so that it can be used when looking in the main config. This code ignores all that, since `serv_config` already has the correct environment loaded.

keychain/models/keychain.py
andreparames: These `keys` will be passed to that MultiFernet, which is supposed to received a list of Fernets: https://cryptography.io/en/latest/fernet/#cryptography.fernet.MultiFernet

andreparames:
@hparfr thanks for the review and for the notice about the Roadmap!

keychain/models/keychain.py
hparfr: It's been a while I didn't use server_env, 
So I guess you have something like this : 

```
[prod]
keychain_key: 
   - prod: aabbcc...
   - all: eeffgg..
[dev]
keychain_key:
    - dev: xxyyzz... 
    - all: eeffgg...
```
and `server_config.get(keychain_key)` will return ` ((prod: aabbcc...), (all: eeffgg...))`

So `_retrieve_env()` should  return `('prod', 'all')`
And `config.get(keychain_key_prod)` should return `'aabbcc...'`

keychain/models/keychain.py
hparfr: Yes, I read to quickly :raised_hand_with_fingers_splayed:  :hammer:

keychain/models/keychain.py
andreparames: Why would you have a `prod:` key inside the prod configuration? And what's the purpose of `all:`?

The way I assume people would use it would be to have something like:

  [prod]
  keychain_key: rrttggcc..

  [dev]
  keychain_key: ooeeaa...

And so `server_config.get(keychain_key)` will just return the key of the loaded environment.

hparfr: You can have multiple key for one environment.
Like `prod` to encode stmp credentials and `all` to encode shipping carriers credentials. 
In that way, you ensure mails will not go out in dev and preprod env. But you can ship in all envs (usually carriers won't charge if they don't scan the parcel).

andreparames: Thanks, I think that makes sense. I've updated the code to allow for setting multiple keys in the same environment. `server_environment` actually uses different files, not sections, for each environment, so the format would now be:

  // File dev.conf
  [keychain]
  dev = eerrttee..
  all = oowerrww..

  // File prod.conf
  [keychain]
  prod = qqweertt..
  all = oowerrww..

hparfr (approved):
Thanks

ThomasBinsfeld (approved):
:+1:

lmignon (approved):
great!

ThomasBinsfeld (approved):
Code and functional :+1:

ThomasBinsfeld:
Could we merge this one? @pedrobaeza

pedrobaeza:
Of course!

PR #1170, opened 2018-02-22
[10.0][UPD] base_user_role: Disabled Administrator user selection in notebo…

artux404:
If Administrator user is selected for error in the notebook users of the rules form, and the current rule has no admin settings, administrator can lose all the privileges and will be problematic to fix this without direct editing database records.

### DIFF
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -68,6 +68,7 @@ Contributors
 ------------
 
 * Sébastien Alix <sebastien.alix@osiell.com>
+* Antonio Russo <antonio.r@rwsdigital.com>
 
 Maintainer
 ----------
--- a/base_user_role/__manifest__.py
+++ b/base_user_role/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'User roles',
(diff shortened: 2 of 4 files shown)


sbidoul (approved):

sebalix (commented):
I would prefer a domain defined in Python on the data model, using the `SUPERUSER_ID` constant to be explicit (instead of hardcoding the ID `1`).
What do you think?

sebalix (approved):
I prefer to not block this PR to get this merged :+1:

andhit-r (approved):

andhit-r (commented):
I think you should bump version

moylop260:
> administrator can lose all the privileges and will be problematic to fix this without direct editing database records.

This looks like as dangerous then Shouldn't be used a constraint instead?
Because if a "web service" is used (where a view domain is not computed) could generate the same issue or a importation from csv file.

artux404:
I've updated the code with domain in the models and SUPERUSER_ID constant

sebalix:
This PR can be merged.

sebalix:
@Yajo thanks!

PR #1171, opened 2018-02-23
[MIG] 11.0 mig date range

ovnicraft:


emagdalenaC2i:
I think that this module is already migrated, it was moved to https://github.com/OCA/server-ux/pull/3

ovnicraft:
Could be good if comment this on server-tools repo 👍

pedrobaeza:
It's already on the migration issue. You should check that.

Issue #1151, opened 2018-02-08
[module_auto_update] Server is not restarted after update?


BT-astauder:
IMHO the server must be restarted to read apply changes in python code. Or am I wrong?

This would be in fact quite easy

from odoo.service.server import restart
restart()

We have a similar approach in our company. But there are a few things different:
- We first load all translations of a changes module. We do that also for all modules that depend on the changed module. So modules for our customers always win in the "translation battle".
- Then we just mark the changed modules in the DB as to upgrade.
- Then we restart the server and the upgrades are made as it would be called by -u

The upgrade is done in done before spanning the workers. So no concurrent problems with that.
The restart waits for cron jobs to finish first. Can be seen as advantages, but also disadvantage...

@Yajo @pedrobaeza

pedrobaeza:
Thanks for the advises, Andreas.

About restarting, I see it feasible for the module indeed. TIL that method.

About the general method, what we want to avoid is server downtime, so there can be of course concurrency problems, but repeating is for now our approach, as we consider more important to keep the system running. And as we are running over Docker, there's little chance that the system gets unusable during the repetitions.

astirpe:
What about enabling the restart with a flag in configuration? Restarting the server in the night would be acceptable by many companies I guess.

pedrobaeza:
I think it's a hard change in the module, but @Yajo can say more accurately.

yajo:
Seems legit, feel free to open PR. I guess we'd have to use mock for testing though... Not much of a problem.

BT-astauder:
Thank you all for your responses. I just tested the behaviour again in 11.0:
New fields are created without any restart, but changes in code are not loaded. So I do not understand how this module is supposed to work. How do you get new code running?

For better understanding I like to clarify some differences in the approches.

module_auto_update: 
Updates in a single worker. Other workers are informed right after the update to invalidate the cache. But during the update process concurrency problems may occur. I just can guess that the update during the night works mostly because no other worker is doing something. Otherwise I guess the other workers runs quite soon in concurrency errors.
So the big advantage, that all other workers continue is more theoretically in my eyes. When they really try to do something, they are not very reliable anymore and will fail quite soon.

Our module:
Marks all modules as 'to_update' and restarts the server then. The update then is done before the workers are spanned up again and only a single process is running:
https://github.com/odoo/odoo/blob/2aabdc00c353aa0b2d4e98e66094cee9f149c000/odoo/service/server.py#L619-L629
[...]

PR #1176, opened 2018-03-01
[9.0] add module_checksum_upgrade and refactor module_auto_update

sbidoul:
Following the very interesting conversation in #1151 I've had a deeper look into module_auto_update. The idea is indeed brilliant. However I came to the conclusion that `module_auto_update` is not adequate for our needs for the following reasons:

* it does too many things automatically and is intrusive in the regular Odoo upgrade process
* in v9, it goes into autocommit mode during the upgrade process
* once installed, it cannot be uninstalled (tested in v9), due to a deadlock when removing the stored field on ir.module.module

In general by looking at the history of `module_auto_update`, we can see that it started with a simple and elegant idea, but the resulting implementation taking into account of the edge cases has become very complex and difficult to understand and looks fragile. I believe this comes from the fact that it tries to immediately update the installed hash whatever the mean of installing a module or running the upgrade. 

So I came up with this new module `module_checksum_upgrade` that provides the basic scriptable building blocks (ie computing hashes, and saving them, and upgrading the changed modules only). 

The approach is a little bit different from module_auto_update as it saves the checksums in a ir.config_parameter instead as a field on the module (this is necessary for the uninstall to work, at least in 9). 

We intend to port this module from 8 to 11.

If there is interest, it is easy to adapt module_auto_update to depend on this new module, therefore fixing the uninstall issue in module_auto_update, and avoiding code duplication. I've a [commit](https://github.com/acsone/server-tools/commit/63b5fc961475a63eeb8033f6dd720cec081f05de) ready that does just that.


module_checksum_upgrade/__openerp__.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Module Auto Update Base',
```
legalsylvain: name and summary does'nt seems to fit with the objective of the module. don't you think ?

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,75 @@
...
+            self.skipTest("Own directory not writeable")
+        with tempfile.NamedTemporaryFile(suffix='.py', dir=self.own_dir_path):
+            self.assertNotEqual(
+                self.own_module._get_checksum_dir(), self.own_checksum,
+                'SHA1 checksum not recomputed',
+            )
+
+    def test_saved_checksums(self):
+        Imm = self.env['ir.module.module']
+        base_module = self.env['ir.module.module'].search([
```
sebalix: You can use `Imm`

module_checksum_upgrade/__openerp__.py
sbidoul: @legalsylvain I fixed the name. How would you suggest to change the summary?

module_checksum_upgrade/tests/test_module.py
sbidoul: @sebalix indeed. Done.

pedrobaeza:
The problem with this is that in OCA we can't have several things for the same goal. We should work to have a common ground, not to duplicate modules.

sbidoul:
@pedrobaeza that is why I propose to make module_auto_update depend on this one, so there is no code duplication. This would look like this: https://github.com/acsone/server-tools/commit/63b5fc961475a63eeb8033f6dd720cec081f05de.

This way people can choose to use `module_checksum_upgrade` and script it, or have the automatic `module_auto_update`.

pedrobaeza:
But I don't want to have 2 modules for performing the same that I'm doing now with one. Integrate your solution in the other.

sbidoul:
I don't see how to integrate them without breaking backward compatibility. The point is that we don't want the `auto` part of `module_auto_update`. 

What is the problem of having one module that provides a base functionality, and another one that extends it and make it automatic?

pedrobaeza:
The problem is that I'm rising my complexity, number of modules and so on for no benefit in my approach. The auto part in module_auto_update can be disabled if you want or making parametric.

sbidoul:
Let's see what other say. I can maintain this outside of OCA. That would be a pity.

module_checksum_upgrade/README.rst
```diff
@@ -0,0 +1,83 @@
...
+An easy way to invoke this upgrade mechnism is by issuing the following
+in an Odoo shell session::
+
+  env['ir.module.module'].upgrade_changed_checksum()
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
ThomasBinsfeld: @sbidoul s/{project_repo}/server-tools

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @api.model
+    def _is_install_complete(self):
+        return not bool(self.search([
+            '!',
+            ('state', 'in', ('installed', 'uninstalled', 'uninstallable')),
+        ], limit=1))
+
+    @api.model
+    def _modules_with_changed_checksum(self):
```
ThomasBinsfeld: @sbidoul s/`_modules_with_changed_checksum`/`_get_modules_with_changed_checksum`?

sbidoul:
Regarding the complexity argument, I argue that a two modules approach is less complex and much cleaner than littering `module_auto_update` with tests on a parameter that completely disables it's function.

setup/module_checksum_upgrade/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
andreparames: Is this file needed?

andreparames:
The code LGTM :+1:

setup/module_checksum_upgrade/.eggs/README.txt
sbidoul: Oops. Well spotted. Removed.

ThomasBinsfeld (approved):
(Nitpicking)
Code and functional reviews :+1:

andreparames (approved):

module_checksum_upgrade/models/module.py
sbidoul: @ThomasBinsfeld, done!

pedrobaeza:
Another thing to have in mind is that currently the other module doesn't work in v9. For v10 and v11 it does. You have to check that ones.

sbidoul:
I looked at the code of 10 and 11 before coding this. It is very similar and does not change my reasoning.

module_checksum_upgrade/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Module Checksum Upgrade',
+    'summary': 'Base mechanism to detect changes to '
+               'installed Odoo modules and upgrade them '
+               'automatically',
+    'version': '9.0.1.0.0',
+    'category': 'Extra Tools',
+    'website': 'https://odoo-community.org/',
```
yajo: server-tools

module_checksum_upgrade/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+    'external_dependencies': {
+        'python': [
+            'checksumdir',
+        ],
+    },
+    'depends': [
+        'base',
+    ],
+    'data': [
+    ],
```
yajo: :fire:

module_checksum_upgrade/README.rst
```diff
@@ -0,0 +1,83 @@
...
+
+This addon provides mechanisms to compute sha1 hashes of installed addons,
+and save them in the database. It also provides a method that exploits these
+mechanisms to update a database by upgrading only the modules for which the
+hash has changed since the last successful upgrade.
+
+Installation
+============
+
+There are no specific installation instructions.
```
yajo: :fire:

module_checksum_upgrade/README.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Configuration
+=============
+
+There is nothing to configure.
```
yajo: :fire:

module_checksum_upgrade/README.rst
```diff
@@ -0,0 +1,83 @@
...
+on ``ir.module.module``. It runs a database upgrade for all installed
+modules for which the hash has changed since the last successful
+run of this method. On success it saves the hashes in the database.
+
+The first time this method is invoked after installing the module, it
+runs an upgrade of all modules, because it has not saved the hashes yet.
+This is by design, priviledging safety. Should this be an issue,
+the method ``_save_installed_checksums`` can be invoked in a situation
+where one is sure all modules on disk are installed and up-to-date in the
+database.
```
yajo, line 27: How can you call the method before installing the addon? :thinking:

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,134 @@
...
+        saved.
+
+        In case of error during the upgrade, an exception is raised.
+        If any module remains to upgrade or to uninstall after the upgrade
+        process, this method returns False. It returns True only in case
+        of complete success.
+        """
+        _logger.info("Upgrade starting...")
+
+        _logger.info("Updating modules list...")
```
yajo: 2 logs for the same thing?

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,134 @@
...
+        self.env['base.module.upgrade'].upgrade_module()
+        self.env.cr.commit()  # pylint: disable=invalid-commit
+
+        _logger.info("Upgrade successful, updating checksums...")
+        self._save_installed_checksums()
+        self.env.cr.commit()  # pylint: disable=invalid-commit
+
+        partial_modules = self._get_modules_partially_installed()
+        if partial_modules:
+            _logger.info("Upgrade successful "
```
yajo: This is a warn.

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,134 @@
...
+        changed_modules.button_upgrade()
+        self.env.cr.commit()  # pylint: disable=invalid-commit
+
+        _logger.info("Upgrading...")
+        self.env['base.module.upgrade'].upgrade_module()
+        self.env.cr.commit()  # pylint: disable=invalid-commit
+
+        _logger.info("Upgrade successful, updating checksums...")
+        self._save_installed_checksums()
+        self.env.cr.commit()  # pylint: disable=invalid-commit
```
yajo, line 140: Could you explain why the commits instead of savepoints? I'm pretty sure it makes sense, but a comment would be quite useful. :wink:

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertEqual(
+            self.own_module._get_checksum_dir(), self.own_checksum,
+            'Module directory checksum not computed properly',
+        )
+
+    def test_compute_checksum_dir_ignore_excluded(self):
+        """It should exclude .pyc/.pyo extensions from checksum
+        calculations"""
+        if not self.own_writeable:
+            self.skipTest("Own directory not writeable")
```
yajo, line 48: The original addon test does it differently.

Instead, it writes a false checksum before checking the upgrade behavior: https://github.com/OCA/server-tools/blob/83bce19b454041344ed79de7b7374a9f01f8e98f/module_auto_update/tests/test_module_upgrade.py#L24

It makes addon testable everywhere; please do that instead. :blush:

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,73 @@
...
+            self.assertEqual(
+                self.own_module._get_checksum_dir(), self.own_checksum,
+                'SHA1 checksum does not ignore excluded extensions',
+            )
+
+    def test_compute_checksum_dir_recomputes_when_file_added(self):
+        """It should return a different value when a non-.pyc/.pyo file is
+        added to the module directory"""
+        if not self.own_writeable:
+            self.skipTest("Own directory not writeable")
```
yajo, line 59: Same here

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,73 @@
...
+        self.own_module = self.env['ir.module.module'].search([
+            ('name', '=', module_name),
+        ])
+        self.own_dir_path = get_module_path(module_name)
+        self.own_checksum = dirhash(
+            self.own_dir_path,
+            'sha1',
+            excluded_extensions=['pyc', 'pyo'],
+        )
+        self.own_writeable = os.access(self.own_dir_path, os.W_OK)
```
yajo: Not needed, see below

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,134 @@
...
+class Module(models.Model):
+    _inherit = 'ir.module.module'
+
+    @api.multi
+    def _get_checksum_dir(self):
+        self.ensure_one()
+
+        excluded_extensions = self.env["ir.config_parameter"].get_param(
+            PARAM_EXCLUDED_EXTENSIONS,
+            'pyc,pyo',
```
yajo: could we add `.pot` too?

yajo (changes requested):
I like this new design.

To be honest, at the end of the day we were not using the cron because of its problems. Just manual upgrading. Yikes! odoo is so anti-HA by design... :man_facepalming: 

Well, I added some comments, and before we continue I think we really need to test the `upgrade_changed_checksum` method. As you said, we've found so many corner cases that I really feel we need that, please.

About having 2 addons for the same thing... That part I don't like it TBH. This could become an upgrade to the original one.

I'd appreciate having some UI exposing this to the user too. Not a blocker.

pedrobaeza:
@Yajo @sbidoul that's what I'm saying since beginning: if there's something incorrect in current module, let's change it, but not to create an alternative module.

sbidoul:
@Yajo thanks for the detailed review.

module_checksum_upgrade/models/module.py
sbidoul: What would be the benefit of savepoints?

module_checksum_upgrade/models/module.py
sbidoul: These are not the same: the first is to log the start of the whole checksum upgrade process, then each step is logged (updating module list, marking to upgrade, etc)

module_checksum_upgrade/models/module.py
sbidoul: Yes, makes sense. Would be nice to have a mechanism to exclude/include languages too.

module_checksum_upgrade/README.rst
sbidoul: You don't. Installing this module does nothing until  `upgrade_changed_checksum` is called.

yajo: Oh sorry I understood bad; it's OK then

module_checksum_upgrade/models/module.py
yajo: Savepoints produce the same effect as a commit, but just for the current cursor. At the end, the cursor is committed automatically by Odoo following its usual rules.

Please read this section: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction

sbidoul: Savepoint are not the same as commits: https://www.postgresql.org/docs/9.6/static/sql-savepoint.html.

I'm aware of the risks of commits in normal business logic. Here I really want to commit at each step (to be sure also to have the same behaviour with each Odoo version). There is no risk of corruption because this method is not intended to be invoked as part of a larger transaction. I'll update the docstring on this topic.

sbidoul:
Regarding the duplicate, may I suggest deprecating module_auto_update. Again I don't see the benefit of upgrading it. Fixing while keeping it backward compatible is impossible. 

Actuallly, I'm more and more convinced module_auto_update is dangerous. If you update the code then run the upgrade without restarting the server, I can imagine situations where a migration runs with a previous version of the code in memory.

module_checksum_upgrade/tests/test_module.py
sbidoul: This particular test comes unchanged from the original module:

https://github.com/OCA/server-tools/blob/3ed2a554b0465614ed75be4d4991333e7e2f98a5/module_auto_update/tests/test_module.py#L58-L59

module_checksum_upgrade/models/module.py
sbidoul: Docstring updated

pedrobaeza:
@sbidoul please tell me which compatibility do you break doing this on the existing module. I think that can be reasonable if the new method is better as said.

sbidoul:
@pedrobaeza 

> which compatibility do you break doing this on the existing module 

As pointed out #1151 a restart is needed to have the new python code. Since python code maybe needed in upgrades involving `migrations` scripts, the risk is too high to run an incorrect upgrade.

So I want to get rid of everything that is potentially unsafe in the existing module.
* the cron
* the UI that triggers the checksum upgrade
* all overrides of the standard Odoo methods
* the mechanisme to store checksums (the fields added on ir.module.module prevent uninstall)

To sum up: everything. Except the fundamental idea which is brilliant as I said, but needs to be implemented differently to be safe.

yajo:
> To sum up: everything. Except the fundamental idea which is brilliant as I said, but needs to be implemented differently to be safe.

Lol, quite right! :laughing: OK, let me see if we can get to an agreement :blush: 

> - the cron

The cron job is currently buggy, as you specified, because it doesn't restart the server.

However it's also true that some people might find useful run unattended upgrades. Just upgrade code and, when you know the upgrade is not dangerous, keep it running until it gets upgraded. In fact, that was my initial thought, although we "rolled back" the idea internally due to the obvious problems it has.

So we have 2 paths:

1. Delete the cron data. It's buggy, and it's buggy by design, so we drop it.
2. Disable the cron by default. It's under a `noupdate=1` block, so old installs will keep it as it was, and new ones will have it disabled.

I think option 1 is better. It's a breaking change, but after all it wasn't working fine in the first place. I'd merge that happily.

> - the UI that triggers the checksum upgrade
> - all overrides of the standard Odoo methods

The real problem is not having a UI, but instead:

- The UI is not obvious, because it reuses standard Odoo methods.
- The UI doesn't work perfectly fine because it doesn't force a server restart.

However, overriding the standard Odoo methods has its advantage: it forces consistency. If you're installing or upgrading an addon, it makes sure all other addons are currently up to date before doing that. That's a good thing IMHO, although in some situations it seems a little bit excessive. For instance, if you're installing a new addon and have to upgrade all others, it could take quite a bit of time.

Also, providing an UI also empowers less technical users to use this good feature. Flattening the learn curve is always a good thing.

Then the best solution would be:

1. Yes, we provide a wizard, but it's explicit and obvious.
2. Yes, we override standard Odoo methods, but do not do that by default. Just a system parameter that enables or not the feature. The same autoupdate wizard should have a checkbox that allows the sysadmin to *force module consistency when performing module changes* that should enable the standard odoo methods overriding.
3. We restart the server after upgrades. Odoo does so when hitting any upgrade button, so there must be a way to do it.

> - the mechanisme to store checksums (the fields added on ir.module.module prevent uninstall)

This is also true, although it's something that can be easily achieved with a premigration script.

---

So, after looking at this with perspective, I agree with both:

- @sbidoul, you're right: the design needs to change dramatically.
- @pedrobaeza, you're also right: the module should be upgraded instead of adding a new one.

@sbidoul In case you were worried about breaking changes not being merged, don't worry, I won't block those if they make sense. :wink: 

So, do we all agree now?

pedrobaeza:
Yes, of course.

sbidoul:
@pedrobaeza @Yajo I think you missed my main point.

What I say is that the server must be restarted BEFORE the upgrade, not after. Indeed, when a migration runs, it may invoke methods that already loaded in memory and therefore can potentially run with the previous version of the code.

That is why I say this method must not have a UI.

Moreover, if we agree the current module is broken in subtle ways, and we don't need a UI why taking the burden of maintaining it?

pedrobaeza:
What I don't want is to change the module name to ease the transition to this new method. That's why I insist to change current module.

yajo:
Yes, @sbidoul, I didn't get that point :wink: 

However, I also think you didn't get ours: you're right, the addon was broken, so go on and break backwards compatibility, we won't block that. We don't want to maintain a broken addon, but we also don't want to maintain 2; it's just that.

lmignon:
@Yajo @pedrobaeza the name `module_auto_update` doesn't reflect the scope of the implementation proposed by @sbidoul. Into this PR there is nothing automatic anymore. Why do you want to keep a meaningless name compared to the proposed implementation here?

pedrobaeza:
As said, we don't want 2 modules in the repository.

sbidoul:
@pedrobaeza I understand your need and for me the module name does not matter that much (although I agree with @lmignon). Question is: is it ok for us to radically change the behaviour of an existing module? To me it's not.

The [commit](https://github.com/acsone/server-tools/commit/63b5fc961475a63eeb8033f6dd720cec081f05de) I proposed above is still the best approach IMO:
* we keep backward compatibility so it's transparent for you and other at the beginning (at the minimal cost of adding an new module)
* we inform the community that module_auto_update is deprecated and will not be migrated to future versions
* whenever it's convenient, you add a call upgrade_changed_checksum in your deploy script, and you can uninstall module_auto_update

module_checksum_upgrade/__openerp__.py
```diff
@@ -0,0 +1,28 @@
...
+    'name': 'Module Checksum Upgrade',
+    'summary': 'Base mechanism to detect changes to '
+               'installed Odoo modules and upgrade them '
+               'automatically',
+    'version': '9.0.1.0.0',
+    'category': 'Extra Tools',
+    'website': 'https://github.com/OCA/server-tools/',
+    'author': 'ACSONE SA/NV, '
+              'LasLabs, '
+              'Juan José Scarafía, '
```
pedrobaeza, line 14: This should Ingenería ADHOC I think cc @jjscarafia

pedrobaeza:
@sbidoul you know that we can't uninstall the module, so that's why I want to preserve the same for letting Odoo to remove the field, but OK, we will rename the module with `openupgradelib`.

sbidoul:
@pedrobaeza I just pushed the refactoring of module_auto_update on this branch. I think you can uninstall it. Give it a try.

pedrobaeza:
OK, thank you very much for your efforts. We'll try tomorrow and tell you.

sbidoul:
I added a note in the README of `module_auto_update`.

I also increased the test coverage. I don't see a way to test `upgrade_changed_checksum` but it's as simple as possible and all the smaller methods it calls are tested.

module_checksum_upgrade/__openerp__.py
sbidoul: I don't know, I copied this from module_auto_update.

sbidoul:
Pffew, now with full test coverage.

rvalyi:
Guys, you may be interested by this little bash script too: basically it grep in the Odoo branch diff to check if some XML is changed or if something related to classes or fields is changed and extract a list of changed modules to upgrade (indeed most python changes require only a restart). False positive doesn't matter too much, what is important is not missing updates. You could use it in complement may be.
https://github.com/akretion/modulesdiff

module_checksum_upgrade/tests/test_module.py
```diff
@@ -102,3 +103,30 @@ def test_get_modules_partially_installed(self):
         self.own_module.button_upgrade_cancel()
         self.assertTrue(
             self.own_module not in Imm._get_modules_partially_installed())
+
+    def test_upgrade_changed_checksum(self):
+        Imm = self.env['ir.module.module']
+        Bmu = self.env['base.module.upgrade']
+
+        # upgrade_changed_checksum commits, so mock that
+        self.env.cr.commit = mock.MagicMock()
```
yajo: Oops, wrong patching, do this instead please:

```python
with mock.patch.object(self.env, "cr") as patched_cr:
    ...
```

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Imm._save_checksums({})
+        changed_modules = Imm._get_modules_with_changed_checksum()
+
+        def upgrade_module_mock(self):
+            changed_modules.write({'state': 'installed'})
+
+        # we simulate an install by setting module states
+        Bmu._patch_method('upgrade_module', upgrade_module_mock)
```
yajo: If you mock it like the above, you can then do `upgrade_module_mock.assert_called_once()` which gives bigger sense to the mock.

module_checksum_upgrade/tests/test_module.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def test_upgrade_changed_checksum(self):
+        Imm = self.env['ir.module.module']
+        Bmu = self.env['base.module.upgrade']
+
+        # upgrade_changed_checksum commits, so mock that
+        self.env.cr.commit = mock.MagicMock()
+
+        Imm._save_checksums({})
+        changed_modules = Imm._get_modules_with_changed_checksum()
```
yajo: You can change you own module checksum before calling this and then make sure it appears there. Pretty similar to https://github.com/OCA/server-tools/pull/1176/files#r172818652

yajo (changes requested):
Thanks for the last changes and the tests! I'm giving you some suggestions on those.

Also please move everything into `module_auto_update`.

We were not asking you to support things that were not working fine before, but after you've already done that, then it makes even less sense to split in 2 addons: you can put all in the same one, and add a simple migration script and a simple note in *Known Issues / Roadmap* that declares what methods are deprecated now. You can even log a deprecation warning when calling them.

Thanks!

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,138 @@
...
+class Module(models.Model):
+    _inherit = 'ir.module.module'
+
+    @api.multi
+    def _get_checksum_dir(self):
+        self.ensure_one()
+
+        excluded_extensions = self.env["ir.config_parameter"].get_param(
+            PARAM_EXCLUDED_EXTENSIONS,
+            'pyc,pyo,pot',
```
yajo: From https://github.com/OCA/server-tools/pull/1176#issuecomment-371625731 I got the idea that we should add here some more: `js,css,less,sass,scss`. Upgrades in those files land with a simple F5, and renames on them would land in an XML after all, so we can skip them safely too.

sbidoul: These should be restricted to the static directory. I update the roadmap.

yajo: Can they actually be found elsewhere?

sbidoul: you never know what developers can invent, safety first

yajo: Come on... This is just a sane default, if you install an addon that implements Odoo in JS, you can add this system parameter for sure :laughing:

sbidoul:
> Also please move everything into module_auto_update.

No, there is no remaining good reason to do that, and OCA cannot completely change the behaviour of an existing module. See also https://github.com/OCA/server-tools/pull/1176#issuecomment-371486590. There is a deprecation warning in module_auto_update readme.

I improved the integration test and added a migration in module_auto_update.

Now I'm done with this (except bugs of course).

module_checksum_upgrade/tests/test_module.py
sbidoul: @Yajo `mock.patch.object(self.env['ir.module.module'], 'upgrade_module', upgrade_module_mock)` does not seem to work. Do you have an example where an Odoo model is patched with mock?

sbidoul:
@rvalyi thanks for the pointer. We were also experimenting a `git diff` approach. The main drawback is that it requires git and the database to be available in the same environment. Which is not always the case. The approach here is a bit less precise (some false positives) but works everywhere whatever the origin of the addons installed. In most practical situations, it's a huge time saver.

sbidoul:
I implemented a more general mechanism to exclude files (eg `static/*`), as well as the mechanism to ignore the `.po` of languages that are not installed.

In the process, I discovered that the checksumdir module on which module_auto_update depended has no tests at all. Moreover it was hard to add the new exclusion mechanism using the checksumdir api. So I implemented a local version of the algorithm, with tests (it's very simple, no need for an unreliable dependency that we don't control).

module_checksum_upgrade/__openerp__.py
```diff
@@ -0,0 +1,28 @@
...
+              'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'license': 'LGPL-3',
+    'application': False,
+    'installable': True,
+    'external_dependencies': {
+        'python': [
+            'checksumdir',
```
lmignon: @sbidoul According to your last comment this dependency should be removed

module_checksum_upgrade/addon_hash.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def addon_hash(top, exclude_patterns, keep_langs):
+    """ Compute a sha1 digest of file contents """
+    m = hashlib.sha1()
+    for filepath in _walk(top, exclude_patterns, keep_langs):
+        # hash filename so empty files influence the hash
+        m.update(filepath.encode('utf-8'))
+        # hash file content
+        m.update(open(os.path.join(top, filepath), 'rb').read())
```
andreparames: I think it's good practice to close your file handles. Might run into problems when running Odoo in PyPy or other non-referenced-counted implementations.

sbidoul: @andreparames indeed, thanks, done.

module_checksum_upgrade/__openerp__.py
sbidoul: @lmignon indeed. done.

yajo:
Wow, really good contribution! :raised_hands: 

I still don't get why can't we move all into the preexisting module.

Your strategy on the matter so far has been:

1. Deprecate `module_auto_update` by adding a message to its README.
1. Make it depend on `module_checksum_upgrade`, which provides a different API to do the same.
1. Add a migration script to perform such upgrade.

But if you just follow this alternative strategy, we don't need 2 addons:

1. Keep a similar migration script. It just lacks a new `ir.config_parameter` (if absent) that we could call `module_auto_update.deprecated_enabled` so we know that deprecated sections are kept working as before.
1. Deprecate just the problematic sections of `module_auto_update` by logging a warning when calling such methods and adding a note in known issues.

   In new installations, where the above parameter is not created by the migration scripts, it defaults to `False` and skips such behaviours.
1. Add the new API into `module_auto_update` instead of into `module_checksum_upgrade`. Since it's completely different, they won't conflict.
1. Set `active=False` by default on the cron. Since it's under a `noupdate=1` section, existing installations will not be affected, but new ones will not get the buggy behaviour out of the box, although still available: https://github.com/OCA/server-tools/blob/3ed2a554b0465614ed75be4d4991333e7e2f98a5/module_auto_update/data/cron_data.xml#L2-L5

The hardest part (backwards compatibility) is already done, so anything else is little more than copy & paste from here to there.

sbidoul:
One last thing I want to add is an argument to upgrade_changed_checksum to control `--i18-override`.

sbidoul:
@Yajo I don't understand why you want me to go all the trouble of adding all those deprecation options in the existing module, while the solution I have here is simple and clean.

yajo:
Want me to do it?

sbidoul:
> Want me to do it?

That's not the point. What you propose is just more complex. I mean Odoo gives us a modularity mechanism. Why adding messy tests to deprecate all features of a module when the module can be simply uninstalled.

module_checksum_upgrade/models/module.py
```diff
@@ -0,0 +1,149 @@
...
+        of complete success.
+
+        Note: this method commits the current transaction at each important
+        step, it is therefore not intended to be run as part of a
+        larger transaction.
+        """
+        _logger.info("Checksum upgrade starting...")
+
+        config['overwrite_existing_translations'] = \
+            overwrite_existing_translations
```
ThomasBinsfeld: @sbidoul Looks nice! Should we log that? :+1:

yajo:
Of course it seems not much useful to you because your target is to not install `module_auto_update`, but we are using it since long ago and my target is to fix it.

I understand you don't want to work for me, so I did it here: https://github.com/OCA/server-tools/pull/1190. Basically, all of your job in a single addon, with a clear deprecation path: the best of both worlds :blush:

module_checksum_upgrade/README.rst
```diff
@@ -68,6 +68,8 @@ Contributors
 * Juan José Scarafía <jjs@adhoc.com.ar>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
 
+The main idea of this module comes from `module_auto_update` by LasLabs Inc.
```
yajo: Please reword to avoid confusing readers: you can credit LasLabs for their [initial implementation](https://github.com/OCA/server-tools/pull/882), but [the idea comes from somebody else](https://github.com/OCA/server-tools/issues/790#issuecomment-298837926). :wink:

sbidoul: I did not know the idea was yours. Credits updated.

sbidoul:
Moved to #1190

PR #1158, opened 2018-02-15
[11.0][MIG] auto_backup

astirpe:
Porting to v11 + code review

PR #1177, opened 2018-03-01
11.0 mig auto backup

luistriana28:
Migrate auto_backup module to v11

@alan196 @lreficent @lasley 

Regards

Hope your help!


pedrobaeza:
There's already another migration in #1158. You should check before for not duplicating efforts

PR #1180, opened 2018-03-05
[11.0] module_checksum_upgrade

sbidoul:
Same as #1176


sbidoul:
Will need to port #1190

PR #1181, opened 2018-03-07
[FIX] manifest version

rgarnau:


pedrobaeza:
Fast-tracking this, but it's better to provide a bit of information in the PR itself.

JordiBForgeFlow:
Thanks!

PR #1184, opened 2018-03-09
[FIX] auth_signup_verify_email: Better integration tests

yajo:
- Remove commits in tests.
- Bigger timeout.
- Run at post-install, since partners are created, which easily leads to `ProgrammingError`s.

@Tecnativa


pedrobaeza (approved):
Make a commit in tests is very dangerous. Let's fast track this!

PR #1186, opened 2018-03-11
[IMP] allow to save passwords in attempted logins

ruifranco:
This creates a new parameter that will enable to save the passwords used for failed logins.

This feature is important as knowing what possible intruders are trying to use as a password may actually tell us a lot about who they might be.

### DIFF
--- a/auth_brute_force/controllers/main.py
+++ b/auth_brute_force/controllers/main.py
@@ -32,7 +32,9 @@ class LoginController(Home):
                 # Test if remote user is banned
                 banned = banned_remote_obj.search([('remote', '=', remote)])
                 if banned:
-                    request.params['password'] = ''
+                    if not int(config_obj.get_param(
+                            'auth_brute_force.save_passwords')):
+                        request.params['password'] = ''
                     _logger.warning(
                         "Authentication tried from remote '%s'. The request "
                         "has been ignored because the remote has been banned "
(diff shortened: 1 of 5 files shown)


auth_brute_force/controllers/main.py
```diff
@@ -45,13 +47,13 @@ def web_login(self, redirect=None, **kw):
                         request.session.db, request.params['login'],
                         request.params['password'])
                 # Log attempt
-                attempt_obj.create({
+                attempt_id = attempt_obj.create({
```
legalsylvain: minor changes, but the name is better attempt, than attempt_id, because it is an object.

auth_brute_force/controllers/main.py
```diff
@@ -32,7 +32,9 @@ def web_login(self, redirect=None, **kw):
                 # Test if remote user is banned
                 banned = banned_remote_obj.search([('remote', '=', remote)])
                 if banned:
-                    request.params['password'] = ''
```
legalsylvain: Maybe I'm wrong, but it you make this sentence, conditionable, you'll change the algorithm. you should write better : 

```
password = request.params['password'] 
request.params['password'] = ''
```
and later
```
attempt_id.write({'password': password})
```

legalsylvain (changes requested):

auth_brute_force/controllers/main.py
ruifranco: It wouldn't but I've changed it anyway.

auth_brute_force/controllers/main.py
ruifranco: Done

JayVora-SerpentCS:
Little travis fix and we are good to go! 
for v11 , can we add an idea like...we allow users to recover password only if the password entered matches the last 3 TRUE passwords?

ruifranco:
@JayVora-SerpentCS

The problem is that those errors that show on Travis have nothing to do with the changes I did on the module I'm dealing with:

- auth_user_case_insensitive
- base_export_manager
- base_kanban_stage_state
- base_multi_image
- base_suspend_security
- sql_request_abstract
- user_immutable
- webhook

I forked from the latest release so, I don't see why it should be throwing errors. Am I to fix those?

;)

JayVora-SerpentCS:
If those errors are coming from other modules, you need not to do!

ruifranco:
@JayVora-SerpentCS

Well, yes, they do come from other modules.

PR #1187, opened 2018-03-11
[10.0] add module_checksum_upgrade and refactor module_auto_update

sbidoul:
Same as #1176


sbidoul:
Will need to port #1190

PR #1189, opened 2018-03-11
Update ru.po

nek-v:


pedrobaeza:
Well, the usual way to do this is through Transifex, but we can admit it this time. Please do it next times through the other way. https://odoo-community.org/page/translate

PR #1190, opened 2018-03-14
[REF] module_auto_update: Deprecate buggy features, add new API

yajo:
The previous implementation of this addon proved being extremely buggy:

- It supplied out of the box a enabled cron to update Odoo that didn't restart the server, which possibly meant that upgrades broke things.
- It overloaded standard Odoo upgrade methods that made i.e. installing an addon sometimes forced to upgrade all other addons in the database.
- The checksum system wasn't smart enough, and some files that didn't need a module upgrade triggered the upgrade.
- It was based on a dirhash library that was untested.
- Some updates were not detected properly.
- Storing a column into `ir.module.module` sometimes forbids uninstalling the addon.

For all of these reasons, the addon has been refactored:

- All buggy features are now deprecated and moved into `*_deprecated.*` files.
- A new update system has been added.
- More tests.
- New checksum system, tested too.
- Smarter diff detection system that triggers only on really needed upgrades.
- Cron is disabled by default.
- Old installations should keep most functionality intact.

Most of the job is copied from @sbidoul's #1176, so credits are added where needed too.

@Tecnativa


lmignon (changes requested):
@Yajo This PR MUST BE CLOSED!!!!!! It doesn't respect at all the ownership of the author and is a lack of courtesy regarding the work done by @sbidoul in #1176.

pedrobaeza:
@lmignon why it doesn't respect ownership? Everything is properly credited, and he wants to save work to @sbidoul converging both needs, ours and yours. I don't see this as a discourtesy sincerely.

lmignon:
@pedrobaeza In a lot of comments, @sbidoul explained his motivations for keeping the work he did in a new module. I fully share his point of view on this topic. This PR looks like an attempt to impose your way of seeing things regardless of the arguments expressed. Moreover, in #1176 the commit history is very clean and each commit is signed while here all the work done is summarized into one commit with @Yajo as author. For me, this PR is disrespectful and does not respect the ownership of the work done. The name `module_auto_update` has nothing to do with the logic implemented in #1176.

pedrobaeza:
Sorry, but you are the ones that want to make the things their way. This module is already in all existing versions, and you are forcing to use 2 modules and break possible installations imposing a new dependency. As you know, other things (some of mine for example) have been blocked because the same reason. The only thing for not forcing this was that we don't want you to load extra work, but now that we have made it, there's no excuse. I agree to rename the module in newer versions (v12), or even to try to put a link to make some kind of alias.

About the commits, if you insist, we can bring here that commits, and the final one will remove all and put it like it is now.

sbidoul:
@Yajo what you are doing here is extremely rude.

pedrobaeza:
I don't agree, Stéphane. Check my previous comment.

sbidoul:
Yes, commits matter, of course they do.

It's not you vs us. I was just expecting an answer to my [last comment](https://github.com/OCA/server-tools/pull/1176#issuecomment-372672718). Frankly all this deprecated stuff is so messy, for what benefit? Saving the uninstallation of one deprecated module. I just don't understand.

As said before, I don't care about the module name. So if you are ready to port and maintain all this from 8 to 11...

sbidoul (changes requested):
Please, at the very least, preserve my commits.

sbidoul:
Can't get this thing out of my head :)

I try to put myself in your shoes, so:
- with #1176 you will need to uninstall `module_auto_update`
- with this PR you will need to remove system parameter `module_auto_update.enable_deprecated`

Is that correct?

yajo:
> Please, at the very least, preserve my commits.

Hi there pals, sorry if I offended somebody, :disappointed: that wasn't my intention at all! I just wanted to stop discussing and just get the job done.

When doing the job, I found myself on the situation on what to do with the commit history:

- It didn't make sense to me to keep all the commits from #1176, since after all, if merged, we'd squash them in 1 or 2.
- I could have had all those commits squashed and then one of mine over it moving all things into this addon, but then when following the next migration to other branches, @sbidoul's history would be lost also, because our git recommended method strips out all history from other folders, and `module_checksum_upgrade` would be one of them. At the end we'd have the same problem but on other branches, so I feel this wasn't a good solution.
- I could have squashed all of this into the first @sbidoul's commit, and I have nothing against doing that (I can if you want), although I quickly remembered https://github.com/OCA/web/pull/647#issuecomment-305475943 and thought that attributing to @sbidoul some job he wasn't willing to do could be seen as a personal harm. As said above I didn't want to harm anybody here.

So, considering all points above, I decided to own the commit but credit @sbidoul in the README, [in the commit message](https://github.com/OCA/server-tools/pull/1190/commits/f836adf6037ea5e1947ee35a59e670765a4d103e) and in the copyright of the files. This is all open source software, so it's not actually *wrong* to copy and paste with the same license, but please keep in mind that my intention on the git history has been to find a good balance between being pragmatic and respectful. If you still want, I can squash all into a @sbidoul's commit; I'm open to suggestions since I consider @sbidoul the original designer of this code (I just did the duck-taping).

When thinking on this matter, please also consider these points:

- Some code from `module_checksum_upgrade` was copy-pasted from `module_auto_update` too.
- If we ever dumped `module_auto_update` as you suggested, all commit history from that addon would be lost in newer versions too.
- The idea was completely borrowed from `module_auto_update`.
- Having all of this in mind, @sbidoul also did his own commits in #1176 and added authorship and copyright wherever needed.
- We didn't become angry because any of that :wink: 

> Saving the uninstallation of one deprecated module. I just don't understand.

It would require quite a bit of manual job in our side that we want to avoid.

Besides, if the addon was broken, we think that it's better to fix it than to dump it.

At the end of the day, once the job is done, there's no more reason to split things.

> As said before, I don't care about the module name. So if you are ready to port and maintain all this from 8 to 11...

Since Laslabs went away, we are almost the only maintainer of this code from 9-11, so obviously yes, we'll maintain it. Here we are deprecating some code that now will have reduced (if any) maintenance from us from now on, of course. Most of the new, undeprecated, code is yours, so I'm pretty sure you'll be ready to co-maintain it too, no matter the addon name. :blush: 

When backporting to v8, all deprecated options can be dropped, since there's no chance to break existing installations, and we just have to make sure it creates a `module_auto_update.enable_deprecated` parameter set to `0`, so such instances don't need to care about the deprecated stuff when forward-migrating.

When forward-porting to future v12, it would be the same, except that we don't need to care about such parameter anymore.

> with this PR you will need to remove system parameter module_auto_update.enable_deprecated

With this patch, this parameter is only set to `1` in the migration script (when upgrading to `9.0.2.0.0` or newer), which means that for new deployments, you simply don't need to care about the deprecated features.

In case you had an old deployment and want to disable them, you can set the parameter to `0` either before or after updating the addon. It will be respected.

lmignon:
@Yajo I'm very tired with your continuous squash by default of the commit histroy. If the history is proper and the author ask you to keep it, just keep it. I agree that sometimes we can squash some useless commits but if we take the time to write and organise our dev into proper and well structured commit, there's no reason to squash the history.

sbidoul:
No hard feelings either, trust me. 

Commits are important because it's all that remains over time. When we look at contribution history, it's easy to analyse commits, while it's hard to analyse unstructured code headers. And a well crafted commit history provides valuable information on the reason why things end up being how they are.

> If we ever dumped module_auto_update as you suggested, all commit history from that addon would be lost in newer versions too.

That is false: the commits remain in the history of server-tools.

> Some code from module_checksum_upgrade was copy-pasted from module_auto_update too.

That is true: a few tests and one method name. A tiny fraction.

I now give more prominent credit to laslabs for the idea, though.

> Besides, if the addon was broken, we think that it's better to fix it than to dump it.

Absolutely, that's what I always do too. In this case however, it is not a fix, it is a full deprecation. 

> It would require quite a bit of manual job in our side that we want to avoid.

That is the only argument I see so far in favor of this PR over #1176. And IMO it is dwarfed by the additional complexity. If I understood it better I could change my mind. @Yajo would you like to explain why uninstalling `module_auto_update` is more manual work for you, compared to changing  a system parameter (`module_auto_update.enable_deprecated`)? In my understanding both are scriptable in a couple of lines of code.

yajo:
`module_auto_update` is itself the addon that takes care of updating other addons. For that reason, a good design decision is to make it depend only on `base`. Otherwise, you easily get to the chicken-and-egg problem.

Basically with the next update, our automated systems would be calling `env["base.module.upgrade"].upgrade_module()` as usual, which lower on the stack would be calling some methods that wouldn't exist because they would be found in an uninstalled addon.

It's similar to some systems that provide a very basic base for others and need to vendorize libraries instead of relying on automatic dependency resolution systems. [Pipenv is an example](https://github.com/pypa/pipenv/tree/master/pipenv/vendor) that comes to my mind.

Also, I know that's not your case, but some people still install addons with the git clone and symlink method. This update would break their flow too.

This is a very important piece in automated deployments and has to be treated with the care it deserves.

The deprecation process, however (adding the parameter), makes the addon work as it did before out of the box, breaking nothing. It lets the operators decide at which point and in which projects to stop using the deprecated features.

The deprecation diff wouldn't be so big were this the original mindset when proposing the patch, but since you fully rewrote the addon (and honestly it looks better), I think it's fair to use your code instead.

About the idea attribution, I'll comment in the other PR.

sbidoul:
@Yajo 

It is true that `module_auto_update` will not upgrade itself automatically the first time when adding the dependency on `module_checksum_upgrade`. Its upgrade needs to be forced. That I can see as a real problem of #1176, although a very small one compared to the complexity of the deprecation approach, IMO.

Anyway if that's the only way to progress let's go.

Could you clarify the problem you see with clone+symlinks deployments?

Do you prefer to rebase this on top of my commits, or do you want me to do it?

yajo:
> Could you clarify the problem you see with clone+symlinks deployments?

Yep, the problem would be that, even if you updated `module_auto_update` before anything else, its update would fail due to depending on a new addon that is not in the addons path. You'd have to create the symlink before, and updates would get broken.

> Do you prefer to rebase this on top of my commits, or do you want me to do it?

I don't want your commits to be lost in migrations due to what I explained in https://github.com/OCA/server-tools/pull/1190#issuecomment-373314683 :confused:, let me squash all into your 1st commit so you appear as the author. You deserve such recognition indeed! :raised_hands:

sbidoul:
Why don't you make 2 commits: one with my work and one with yours?

yajo:
Hmm yes, that'd be better indeed, I'll do ASAP. :wink:

yajo:
OK, I have done it in 3 commits so everything is clear and still we have all history in the same addon, so we avoid losing your commits by accident in the future:

1. Moving all stuff into deprecated files.
2. Moving all your code into `module_auto_update`, pretty much verbatim. Your patches that were originally for this addon land now in the deprecated files.
3. My deprecation things.

lmignon:
@Yajo Is it possible to preserve all the commits from #1176 ?  As @sbidoul says this history provides valuable information on the reason why things end up being how they are and it's really valuable to be able to analyse each diff according the commit message to understand the development path.
Regards, 
lmi

sbidoul:
> Is it possible to preserve all the commits from #1176 

What you say is true but it's no big deal in this case, Laurent. Let's move on.

yajo:
Oops, butterfinger me.

Just pointing out that as you can see @sbidoul's commit message includes a link to that PR, so we can see the full history on how that commit became to exist.

Glad we have an agreement finally! :blush:

Now I'm gonna get into the technical stuff and check all of this really works in some real instance, setting as WIP until then.

pedrobaeza:
@Yajo please check the 2 problems we found:

* if the changed module is base, Odoo treats update other way. Maybe we can ignore these changes as they for sure won't introduce changes that require to update?
* if a module changes, make sure dependent modules are upgraded too.

sbidoul:
@pedrobaeza I've tested the two cases you mention, but it's good if you can check it too.

sbidoul:
@ThomasBinsfeld can you test this version?

yajo:
> if the changed module is base, Odoo treats update other way.

Not sure what you mean by "other way", AFAIK it just updates base and all of its dependencies as usual... Will check it though. :confused: 

> Maybe we can ignore these changes as they for sure won't introduce changes that require to update?

The improved ignored paths system introduced by @sbidoul prevents most unnecesary updates that were producing such problems, so it seems better to keep it as it is IMHO.

> if a module changes, make sure dependent modules are upgraded too.

AFAIK that was happening before; will double check.

pedrobaeza:
>  if the changed module is base, Odoo treats update other way.

> > Not sure what you mean by "other way", AFAIK it just updates base and all of its dependencies as usual... Will check it though. confused

There was a commit implied when updating it.

module_auto_update/migrations/9.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,24 @@
...
+# TODO Remove when deprecated features are removed; see README's known issues
+@openupgradelib.migrate(no_version=True, use_env=True)
+def migrate(env):
+    """Move stored checksums to new JSON format."""
+    # Store module checksums in the new JSON format
+    env.cr.execute("""SELECT name, checksum_installed
+                      FROM ir_module_module
+                      WHERE checksum_installed""")
+    checksums = dict(env.cr.fetchall())
+    env["ir.module.module"]._save_checksums(checksums)
```
sbidoul: note the checksum algorithm has changed so this not help.

module_auto_update/models/module.py
```diff
@@ -1,70 +1,148 @@
...
+        saved.
+
+        In case of error during the upgrade, an exception is raised.
+        If any module remains to upgrade or to uninstall after the upgrade
+        process, an exception is raised as well.
+
+        Note: this method commits the current transaction at each important
+        step, it is therefore not intended to be run as part of a
+        larger transaction.
+        """
```
sbidoul, line 110: Raise an error here if enable_deprecated is set? I'm not sure the two methods work well together.

yajo:
I added a patch that makes the 1st autoupgrade since the module update know it needs to update all.

Since the sha algorightm has changed, this addon upgrade will mean the upgrade of every addons in the database. This seems a little bit bad, but after all it's not such a really big problem since it should happen only once, and it's not that uncommon.

module_auto_update/wizards/module_upgrade_deprecated.py
```diff
@@ -10,10 +10,25 @@
...
+            self.env["ir.config_parameter"].get_param(PARAM_DEPRECATED))
+        if deprecated is False:
+            # Enable deprecated features if this is the 1st automated update
+            # after the version that deprecated them (X.Y.2.0.0)
+            own_module = self.env["ir.module.module"].search([
+                ("name", "=", "module_auto_update"),
+            ])
+            if int(own_module.latest_version.split(".")[2]) < 2:
+                deprecated = "1"
+        return deprecated == "1"
```
sbidoul: headache :smile:

module_auto_update/migrations/9.0.2.0.0/pre-migrate.py
yajo: You are correct, I'll do a couple of upgrade tests just to be sure and remove it, thanks :wink:

module_auto_update/models/module.py
yajo: According to my tests, they work fine together.

yajo:
Cool, this one should be ready to review. 

- Deprecated features get autoenabled only when upgrading
- A warning is logged when you use deprecated thing
- The deprecated system is based on the new one, so boty upgrade mechanisms work together flawlessly
- Addon can now be uninstalled
- I double-checked that updating an addon updates its dependencies
- `base` is treated normally since the new sha algorithm should prevent many unneeded updates to it

module_auto_update/models/module.py
```diff
@@ -1,70 +1,148 @@
...
+        _logger.info(
+            "Checksum upgrade starting (i18n-overwrite=%s)...",
+            overwrite_existing_translations
+        )
+
+        tools.config['overwrite_existing_translations'] = \
+            overwrite_existing_translations
```
sbidoul, line 117: This translation option may not work as expected. We are investigating.

yajo: OK, I'm gonna rebase all my commits in one, with another fix. Do you have push access to this PR? If so, you can push here the fix for that part after that :wink:

yajo:
I squashed my commits so you can continue work on top of this; AFAICS my part should be ready to merge.

I patched another little thing: when uninstalling the addon, the stored checksums now disappear.

module_auto_update/models/module.py
ThomasBinsfeld: My last functional tests show that `overwrite_existing_translations` works well. Our doubts came from an external factor. :smiley:

sbidoul (approved):

module_auto_update/models/module.py
```diff
@@ -1,70 +1,148 @@
...
+    @api.model
+    def _get_modules_with_changed_checksum(self):
+        saved_checksums = self._get_saved_checksums()
+        installed_modules = self.search([('state', '=', 'installed')])
+        return installed_modules.filtered(
+            lambda r: r._get_checksum_dir() != saved_checksums.get(r.name),
+        )
+
+    @api.model
+    def upgrade_changed_checksum(self, overwrite_existing_translations=False):
```
yajo, line 88: Do you think it would be a good idea to add here a different default?: `overwrite_existing_translations=os.environ.get("ODOO_I18N_OVERWRITE", False)`, so we can set this directly through an env variable in automated environments? :thinking:

sbidoul: Yes. Not 100% sure about the `ODOO_` prefix, but otherwise :+1: .

carlosdauden (approved):

ThomasBinsfeld (approved):

yajo:
This one seems ready to merge.

pedrobaeza:
Thanks to all!

PR #1192, opened 2018-03-16
Quick create avoidance per model transfered from Python backend to JS…

MasterBlaster479:
… frontend. Reason for doing this is the fact that there is unconsistent behaviour in multi-processing OpenERP instances


MasterBlaster479:
Tests have failed, but I think it's not related to changes made in this pull request.

PR #1196, opened 2018-03-21
[MIG][11.0] base_user_gravatar migration

acsonefho:
About translations:
- I let existing translation as in v10.0


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

acsonefho:
@pedrobaeza Sorry but this feature is into the Odoo standard now =/
So I close this MR.
Thank you for the link :)

PR #1198, opened 2018-03-22
[MIG][8.0] module_auto_update backport from 9

sbidoul:
Last commit is migration, removing the deprecated features that never existed in 8.


JayVora-SerpentCS:
Better to merge OCA transbot commits into 1

pedrobaeza:
In this case they are not adjacent, so they can be let as is.

module_auto_update/models/module_deprecated.py
```diff
@@ -1,69 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from openerp import api, fields, models
-
 PARAM_DEPRECATED = "module_auto_update.enable_deprecated"
```
yajo: I think you can remove this file and declare the param name directly into the hooks file; it's now used only there.

yajo (approved):

module_auto_update/models/module_deprecated.py
pedrobaeza: We don't need deprecated features in v8 as there was no module there, isn't it? But if it's more work, it can be keep.

sbidoul: @pedrobaeza the deprecated features are not here. The parameter is required here in 8, to avoid the deprecated features are enabled automatically when migrating to 9.

sbidoul: @yajo is that blocking for you?

yajo: Well... not blocking, I approved already. However, are there any reasons to keep this file just for setting a variable?

sbidoul: Well it's just in the same place as in the other versions.

yajo:
Please include fix from #1205

sbidoul:
@Yajo done

yajo (approved):

pedrobaeza (approved):

PR #1200, opened 2018-03-24
[10.0][FIX] server_env_base_external_referentials. Make it pass mqt lint.

NL66278:
Part of an effort to make commits to server-env repo pass mqt pre-commit hook.

First try for this particular module was to solve the errors.

On closer inspection the module is bases on another module not maintained since 6.1.

Therefore deletion of the moduel seems to me the proper course of action.


server_env_base_external_referentials/__manifest__.py
```diff
@@ -21,16 +21,20 @@
...
     'license': 'AGPL-3',
-    "description": """This module is based on the server_environment module to use files for configuration.
-Thus we can have a different file for each environment (dev, test, staging, prod).
-This module define the config variables for the base_external_referential module.
-In the configuration file, you can configure the url, login and password of the referentials
+    "description": """This module is based on the server_environment module
+to use files for configuration.
+Thus we can have a different file for each environment
+(dev, test, staging, prod).
+This module define the config variables for the base_external_referential
```
JayVora-SerpentCS: This module defines!

server_env_base_external_referentials/__manifest__.py
```diff
@@ -21,16 +21,20 @@
...
+module.
+In the configuration file, you can configure the url, login and password
+of the referentials
```
JayVora-SerpentCS: DOT after referentials

server_env_base_external_referentials/base_external_referentials.py
```diff
@@ -18,37 +18,40 @@
...
-                   method=True, string='Password', multi='connection_config'),
+        'location': fields.function(
+            _get_environment_config_by_name, type='char', size=200,
+            method=True, string='Location', multi='connection_config'),
+        'apiusername': fields.function(
+            _get_environment_config_by_name, type='char', size=64,
+            method=True, string='User Name', multi='connection_config'),
+        'apipass': fields.function(
+            _get_environment_config_by_name, type='char', size=64,
+            method=True, string='Password', multi='connection_config'),
```
JayVora-SerpentCS: method and size are not necessary params IMHO

NL66278:
@JayVora-SerpentCS Thanks for looking into this. I considered doing the actual migration, but then found out it is based on a module that is no longer maintained since 6.1. Sorry for the work you also out in the first review, but I now think it is better to get rid of the whole thing.

NL66278:
@JayVora-SerpentCS Do you agree with me that the module can better be removed? Then please approve in review.

JayVora-SerpentCS (approved):
Better to be removed yes

yvaucher (approved):
LGTM seems not ported in version 9.0

@guewen reading the manifest it seems you created it in relation to magento connector? Can we drop it?

AFAIK we have no more use of it in later versions.

PR #1201, opened 2018-03-25
[8.0][MIG] oauth provider backport from 9.0

hhgabelgaard:
oauth_provider backport from 9.0:

- `<odoo> -> <openerp><data>`
- crsf handling removed
- website added to http routes
- Exceptions: UserInfo -> Warning

Scope preserved from when swicthing authentification code to token
Extension point for additional userinfo response


pedrobaeza:
Please squash together adjacent "OCA Transbot..." commits using this guide: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

oauth_provider/__openerp__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 SYLEAM
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'OAuth Provider',
+    'summary': 'Allows to use Odoo as an OAuth2 provider',
+    'version': '9.0.1.0.0',
```
JayVora-SerpentCS: 8.0.1.0.0 please

hhgabelgaard (commented):
` cr.execute('ALTER TABLE res_users ADD COLUMN oauth_identifier varchar')`
in oauth_provider/__init.py__
Why is Travis failing over this line? Is it installing the module twice?
Neither OCA's Runbot or my own fails over this

pedrobaeza:
Because the column already exists. Is it created by other module? Maybe you should check if the column exists.

There's a flake8 error too:
```
oauth_provider/models/res_users.py:26:18: W291 trailing whitespace
```

Garamotte:
The obvious issue is that the init hook must check if the column exists before trying to create it.
But there is also [another issue](https://github.com/OCA/server-tools/issues/1202), IMO due to the database_cleanup module (to be fixed in another PR).

hhgabelgaard:
Hi @sylvain-garancher 
Thanks for the analysis in #1202 - That's explain it, and yes, the obvious solution is to check for existence in the pre_init hook. That's now implemented, and the log lines (to be removed)  shows that pre_init is called twice, but the column adding is only done once. So far so good.

But now I'm getting errors from auth_brute_force: [KeyError: 'res.authentication.attempt'](https://travis-ci.org/OCA/server-tools/jobs/358531175#L1710) in line 1710 in the Travis log - But auth_brute_force is installed earlier - in line 1228: [2018-03-26 18:10:00,539 5101 INFO openerp_test openerp.modules.module: module auth_brute_force: creating or updating database tables](https://travis-ci.org/OCA/server-tools/jobs/358531175#L1228)

Could that be due to the same issue? Are the tests being run on "half a registry"?
I'm not familiar with the auth_brute_force module, and it doesn't exist in 9.0 so I can't exclude some kind of "incompabilty" between them.

And why is it only on Travis? Isn't the same tests been run under the Runbot? 
Should I add EXCLUDE="database_cleanup" to .travis.yml ?

Transbot commits is squashed, flake/lint tests is green

PR #1203, opened 2018-03-26
[10.0][MIG] base_tier_validation

LoisRForgeFlow:
Migration to v10.

Base Tier Validation
====================

This module does not provide a functionality by itself but an abstract model
to implement a validation process based on tiers on other models (e.g.
purchase orders, sales orders...).

**Note:** To be able to use this module in a new model you will need some
development.

See `purchase_tier_validation <https://github
.com/OCA/purchase-workflow>`_ as an example of implementation.

Configuration
=============

#. Go to *Settings > Technical > Tier Validations > Tier Definition*.
#. Create as many tiers as you want for any model having tier validation
   functionality.

### DIFF
--- a/base_tier_validation/README.rst
+++ b/base_tier_validation/README.rst
@@ -27,7 +27,7 @@ To configure this module, you need to:
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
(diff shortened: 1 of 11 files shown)


JordiBForgeFlow (approved):

JordiBForgeFlow:
Merging as it is a trivial migration

PR #1194, opened 2018-03-20
[WIP] Use acme official client

daramousk:
This is a work in progress, Mostly I am concerted with the usage of acme client libraries, I seem to have difficulties with their usage. There is a TODO on the code which I was not able to resolve so far, input over this will be greatly appreciated

cc @hbrunn

PR #1204, opened 2018-03-26
update README with ACME2 and wildcard certificate

zeroheure:
update README with parameters path and a note about ACME v2 compatibility and wildcard names
(acme-tiny is compatible with ACME v2 but will not implement wildcard certificate)


hbrunn:
there is a very far developed plan to support wildcard certificates (just not yet working): https://github.com/OCA/server-tools/pull/1194, so please scrap this paragraph

hbrunn (approved):

PR #1210, opened 2018-03-30
[10.0] add fetchmail inbox

NL66278:
Move fetchmail_inbox and fetchmail_inbox_invoice modules from therp/therp-addons to server-tools.

While usefull in itself, this is also in preparation for merging the strong points of fetchmail_attach_from_folder and fetchmail_inbox.


StefanRijnhart:
Nice! It's probably a good idea to squash all of it into a smaller number of commits. Personally I would not mind if you squash my minor contributions into oblivion.

NL66278:
I was able to reduce the number of commits somewhat. Strangely enough if I try to squash some more, I encounter the problem that the commits are not presented in chronological order in ```it rebase -i ...```. No doubt due to history in different branches, but I hope we can live with a history of 15 commits.

PR #1215, opened 2018-04-09
[ADD] copy flag in context to have other code do the right thing (tm)

hbrunn:
When creating users from ldap with a template user, we basically clone the template user with our default values.

In cases where partner_firstname is involved, this goes wrong because this module will overwrite the name from the template user's first, lastname again. Given this is a copy operation in principle anyways, I think it's safe to pass the flag partner_firstname reacts to here: https://github.com/OCA/partner-contact/blob/10.0/partner_firstname/models/res_partner.py#L45 (which it does because the copy context is set in cases some record is copied)

### DIFF
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -69,7 +69,7 @@ class CompanyLDAP(models.Model):
                     conf['ldap_filter'])
             results = self.get_ldap_entry_dicts(conf)
             for result in results:
-                user_id = self.get_or_create_user(
+                user_id = self.with_context(copy=True).get_or_create_user(
                     conf, result[1][login_attr][0], result)
                 # this happens if something goes wrong while creating the user
                 # or fetching information from ldap


sebalix (changes requested):
Some lints (but lint errors are not related to this module) to fix:
```
************* Module users_ldap_populate.__manifest__
users_ldap_populate/__manifest__.py:5: [C8103(manifest-deprecated-key), ] Deprecated key "description" in manifest file
************* Module users_ldap_populate.tests.test_users_ldap_populate
users_ldap_populate/tests/test_users_ldap_populate.py:8: [C8104(class-camelcase), patch_ldap_connection] Use `CamelCase` "PatchLdapConnection" in class name "patch_ldap_connection". You can use oca-autopep8 of https://github.com/OCA/maintainer-tools to auto fix it.
```
BTW, is the scheduled action mentioned in the `description` field in the manifest is always relevant? If so it should be added to the readme.

Issue #1125, opened 2018-01-10
auth_brute_force doesn't consider xmlrpc/jsonrpc login attempts


LeartS:
It's possible to circumvent `auth_brute_force` security by trying logins via xmlrpc/jsonrpc. The login attempts done this way are not checked by the module and do not count towards the failed attempts limit towards the ban.

I would argue that the jsonrpc interface is the preferred one for bruteforcing tools, so IMO this limitation is non-negligible.

legalsylvain:
:+1: When I wrote this module, I just tested with Front UI attempts. 
this could be a good addition to log tryes from xmlrpc. Do you think you could make a PR ? 

regards.

legalsylvain:
Hi @lasley. I allowed myself to change the label you set. I think that "bug" is more appropriated because what @LeartS is talking make this module useless.

yajo:
The fix is in https://github.com/OCA/server-tools/pull/1219, let's continue there.

PR #1219, opened 2018-04-12
[REF] auth_brute_force: Cover all auth entrypoints

yajo:
To fix https://github.com/OCA/server-tools/issues/1125 I needed to refactor the addon.

The fix is affected by https://github.com/odoo/odoo/issues/24183 and will not work until it gets fixed upstream due to the technical limitations implied. ~~This PR is implemented assuming https://github.com/odoo/odoo/pull/24187 will be merged and backported.~~

@Tecnativa


yajo:
According to https://github.com/odoo/odoo/pull/24187#issuecomment-381657198, I monkey-patched Odoo to get this safely working. As you might know, monkey patches affect all databases in the same instance, which can be bad in a SaaS environment. However, the patch is completely harmless in such case.

pedrobaeza:
Using `_register_hook` y `_apply_patch` you do it only for the current DB.

yajo:
Yikes, tricky tests! It's all :heavy_check_mark: now, ready to merge IMHO.

rafaelbn (approved):
Yes! Very important in terms of GDPR :+1: :

chienandalu (commented):
With users behind the same IP (or bad configured proxies): if a user gets banned, would it put the other's sessions (already logged) down?

yajo:
@chienandalu Those already logged in would retain their sessions, although new login would get locked, yes.

This is a side effect of this addon, docummented in Known Issues. Do you think there would be a better approach on the matter?

chienandalu (approved):
No, it's ok like that :)

yajo:
I'm gonna think some more on the subject. It could be really problematic. Imagine:

- One office with 1 single public IP
- 200 users in that office
- At the same time, 5 of them fail passwords or users twice
- **Nobody else from the office will be able to log in**

So, the problem is quite big. I think it should be blocked just by user. Then maybe a notification could be sent to him.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -2,35 +2,80 @@
...
+        return result
+
+    @api.model
+    def search_last_failed(self, remote, until=None):
+        if not until:
+            until = fields.Datetime.now()
         last_ok = self.search(
-            [('result', '=', 'successfull'), ('remote', '=', remote)],
+            [('result', '=', 'successfull'), ('remote', '=', remote),
+             ("attempt_date", "<=", until)],
```
LeartS: Maybe at least some of the fields used in this search should be setted as `index=True` ? On a medium/big installation running for years there could be potentially thousands (or even tens or hundreds of thousands) of login attempts to search in..

LeartS (approved):
Just code review, great code!

If someone tested this functionally, considering also the unittest, for me it's ok to merge.

legalsylvain:
> So, the problem is quite big. I think it should be blocked just by user. Then maybe a notification could be sent to him.

I fully understand your use case. But if what you say is implemented, it should be optional. When I wrote this module, it was not designed to block user that doesn't remember their password, but to prevent attack and the only thing we have is the IP. (it's not perfect, it doesn't prevent Distributed attack, but it's a first step).

If it is only blocked by user and not by IP (as you suggest), it will generate the following use case : 

I want to attack a company that use odoo.
I know that the login is made with name@mycompany.com
I try to log in with all the login I know. (easy social attack).
Result : I can block all the login because my IP is not banned.

I see the implementation of the "block by login" as a big regression. 

**Solution 0 (no dev)**
- set a max_try to 30 (for exemple)
- improve password quality (to avoid attack by dictionnary)

-> I don't see any users trying to log in 30 times. After some tries, he will contact the administrator.


**solution 1**
- add a option "block by login" and / or "block by IP". The admin user can choose the policy, that depends on the network implementation of Odoo.

**solution 2**
- set hard_max_try to 10 (for exemple)
- add a soft_max_try to 5
make a improvment (by JS) to block user to attempt to login after 5 tries during 10 minutes, then during 1 hours, etc...

-> only one browser is blocked.
- maybe quite hard to develop

**solution 3**
- set hard_max_try to 10 (for exemple)
- add a soft_max_try to 5
make a improvment (by JS) to block users to log after 5 tries and redirect the user to a page to reinit the password, via a form. (classical token process).

what do you think ?

kind regards.

LeartS:
Yes, that could be a problem. Adding the user on login attempts is in my opinion an ok solution; it would just mean that a potential attacker could try 10 passwords per user instead of 10 in general; which shouldn't be a problem.

auth_brute_force/models/res_authentication_attempt.py
legalsylvain: In the other hand, I don't see any interest to keep attempts with a date > X monthes. So maybe a cron to clean this table should be great. (I think there is a generic module that does that task for any table in server-tools.)

But Indeed, set an index=True is a good idea !

LeartS:
> If it is only blocked by user and not by IP (as you suggest)

I think @Yajo meant to block just login attempts for that specific user to that IP, not to block the user in general. So the IP is still considered, and you can't use this to block access to odoo (unless you have the same IP as the real user, but in that case it's an internal problem and can't really be avoided..)

yajo:
Yep, I think I'm gonna go that path: just block the user+IP combination instead of just the IP.

legalsylvain:
> I think @Yajo meant to block just login attempts for that specific user to that IP, not to block the user in general. So the IP is still considered.

How ! I didn't understood ! Yes indeed. An option still is a good implementation. like 

```block_user_policy``` : 
* option 1 : block the IP ;
* option 2 : block the IP for the current login only.


> it would just mean that a potential attacker could try 10 passwords per user instead of 10 in general; which shouldn't be a problem.

It can be a problem. if you have a company with 1000 users, you let an attacker to make an attack with up to 10.000 attempts. he has more chance to succeed, statisticaly.

**Comment by LeartS** [comment]
> It can be a problem. if you have a company with 1000 users, you let an attacker to make an attack with up to 10.000 attempts. he has more chance to succeed, statisticaly.

Yes, 10000 in total, but still 10 per user. But I get what you mean: if an attacker tries the 10 most common passwords, the probability of at least 1 user in 1000 using it is much higher than the probability of a single specific user using it.
Maybe we could solve it this way, with two config values:

* Block remote-user after x invalid attempts (the param already implemented)
* Block remote generally after y invalid attempts

So, for example, block ip 1.1.1.1 for user 'admin' after 5 incorrect attempts for user 'admin', and block 1.1.1.1 in general after 50 consecutive invalid attempts between all users.
Setting x = y would mean to always block the IP, setting y to -1 could mean to never block the entire IP.

**Comment by LeartS** [comment]
@Yajo I was thinking it could be useful to save and show the login attempt channel (web - xmlrpc -jsonrpc). But it's ok to do it in a future PR as to not overload this one.

**Comment by yajo** [comment]
You'd be interested in v12 approach, currently in RD phase: https://github.com/odoo/odoo/pull/22212

**Comment by legalsylvain** [comment]
> So, for example, block ip 1.1.1.1 for user 'admin' after 5 incorrect attempts for user 'admin', and block 1.1.1.1 in general after 50 consecutive invalid attempts between all users.
Setting x = y would mean to always block the IP, setting y to -1 could mean to never block the entire IP.

Good idea ! very modular. :+1:

**Comment by yajo** [comment]
This has to change quite dramatically current implementation or it would be too hard to implement. Changing, though.

**Comment by yajo** [comment]
It changed now, take a look please!

**Comment by pedrobaeza** [comment]
Check Travis

**Comment by yajo** [comment]
It's weird, it seems some kinda conflict with other addon, i did a blind commit, let's see if it works.

**Comment by yajo** [comment]
OK Travis is :heavy_check_mark: now. I digged a lot and couldn't discover the conflict, so I just triggered separate test suites.

Shall we squash+merge now?

File: auth_brute_force/models/res_authentication_attempt.py [hunk 2]
```diff
@@ -2,35 +2,160 @@
...
+
+    @api.multi
+    @api.depends('remote')
+    def _compute_metadata(self):
+        for item in self:
+            url = GEOLOCALISATION_URL.format(item.remote)
+            try:
+                res = json.loads(urlopen(url, timeout=5).read())
+            except Exception:
+                _logger.error(
```
naglis: I think you can use `_logger.exception` instead of `_logger.error` with `exc_info=True`.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -2,35 +2,160 @@
...
+        domain_good = [
+            ("result", "=", "successful"),
+            ("remote", "=", remote),
+        ]
+        domain_bad = [
+            ("result", "!=", "successful"),
+            ("remote", "=", remote),
+        ]
+        last_ok = self.search(domain_good, order='create_date desc', limit=1)
+        recent_failures = self.search(
```
naglis: Nitpick: you may also use the [`.search_count()`](https://github.com/odoo/odoo/blob/14f59f4b780fa1e757cfce6a46c62c7a6b00d13e/odoo/models.py#L1486-L1494) shortcut.

auth_brute_force/models/res_users.py
```diff
@@ -0,0 +1,155 @@
...
+from odoo.service import wsgi_server
+
+_logger = logging.getLogger(__name__)
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    # HACK https://github.com/odoo/odoo/issues/24183
+    # TODO Remvove when fixed, and use normal odoo.http.request to get details
```
naglis: Typo: *Remvove*

auth_brute_force/models/res_users.py
```diff
@@ -0,0 +1,155 @@
...
+                "remote": remote_addr,
+            })
+            return attempt.id
+
+    @classmethod
+    def _auth_attempt_update(cls, values):
+        """Update a given auth attempt if we still ignore its result."""
+        auth_id = getattr(current_thread(), "auth_attempt_id", False)
+        if not auth_id:
+            return dict()  # No running auth attempt; nothing to do
```
naglis: Nitpick: `{}` is [faster](https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/) :trollface:

auth_brute_force/tests/test_brute_force.py
```diff
@@ -0,0 +1,361 @@
...
+@patch(http.__name__ + ".redirect_with_hash", side_effect=redirect)
+# Faster tests without calls to geolocation API
+@patch(res_authentication_attempt.__name__ + ".urlopen", return_value="")
+class BruteForceCase(HttpCase):
+    def setUp(self):
+        super(BruteForceCase, self).setUp()
+        # Some tests could retain environ from last test and produce fake
+        # results without this patch
+        # HACK https://github.com/odoo/odoo/issues/24183
+        # TODO Remvove when fixed
```
naglis: Typo: *Remvove*.

auth_brute_force/tests/test_brute_force.py
```diff
@@ -0,0 +1,361 @@
...
+    @skipUnless(can_import("odoo.addons.web"), "Needs web addon")
+    @mute_logger(*GARBAGE_LOGGERS)
+    def test_web_login_existing(self, *args):
+        """Remote is banned with real user on web login form."""
+        data1 = {
+            "login": "admin",
+            "password": "1234",  # Wrong
+        }
+        # Make sure user is logged out
+        self.url_open("/web/session/logout", timeout=30)
```
naglis, line 74: Nitpick: might be worth storing the `timeout` as a constant, seeing that it is used multiple times.

auth_brute_force/tests/test_brute_force.py
```diff
@@ -0,0 +1,361 @@
...
+        # Fail 3 times
+        for n in range(3):
+            response = self.url_open("/web/login", bytes(urlencode(data1)), 30)
```
naglis, line 77: Nitpick / Consistency: here (and below) timeout is passed without the argument name where before the argument name is specified.

auth_brute_force/views/view.xml
```diff
@@ -3,78 +3,71 @@
...
-                   <filter name="filter_failed" string="Failed" domain="[('result','=', 'failed')]"/>
-                   <filter name="filter_successful" string="Successful" domain="[('result','=', 'successfull')]"/>
-                </search>
-            </field>
-        </record>
+    <record id="view_res_authentication_attempt_graph" model="ir.ui.view">
+        <field name="model">res.authentication.attempt</field>
+        <field name="arch" type="xml">
+            <graph>
+                <field name="create_date" />
```
naglis: Nitpick: extra space before `/>`.

auth_brute_force/views/view.xml
```diff
@@ -3,78 +3,71 @@
...
-                    <field name="login" />
-                    <field name="result" />
-                </tree>
-            </field>
-        </record>
+    <!-- Model: res.authentication.attempt -->
+    <record id="view_res_authentication_attempt_tree" model="ir.ui.view">
+        <field name="model">res.authentication.attempt</field>
+        <field name="arch" type="xml">
+            <tree colors="orange: result == 'failed';red: result == 'banned'">
```
naglis: Nitpick: `colors` attribute [has been deprecated ](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#views) since v9.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -2,35 +2,160 @@
...
+            try:
+                res = json.loads(urlopen(url, timeout=5).read())
+            except Exception:
+                _logger.error(
+                    "Couldn't fetch details from %s",
+                    url,
+                    exc_info=True,
+                )
+            else:
+                item.metadata = "\n".join(
```
naglis: Shouldn't it be `remote_metadata`?

rafaelbn:
We need here the last approval of a PSC :+1:

auth_brute_force/README.rst
```diff
@@ -69,14 +69,15 @@ For further information, please visit:
...
+* Remove 🐒 patch for https://github.com/odoo/odoo/issues/24183 when fixed.
+
 * Depending of server and / or user network configuration, the idenfication
   of the user can be wrong, and mainly in the following cases:
-* If the Odoo server is behind an Apache / NGinx proxy without redirection,
-  all the request will be have the value '127.0.0.1' for the REMOTE_ADDR key;
-* If some users are behind the same Internet Service Provider, if a user is
-  banned, all the other users will be banned too;
+
+  * If the Odoo server is behind an Apache / NGinx proxy and it is not properly
```
pedrobaeza, line 79: Incorrect indentation

auth_brute_force/data/ir_config_parameter.xml
```diff
@@ -1,15 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright 2015 GRAP -Sylvain LE GAL
-     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
-<odoo noupdate="1">
-
-    <data>
-
-        <record id="max_attempt_qty" model="ir.config_parameter">
-            <field name="key">auth_brute_force.max_attempt_qty</field>
```
pedrobaeza, line 9: How do we control this now? No change in README?

pedrobaeza (changes requested):

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -2,35 +2,160 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, fields, models
+import json
+import logging
+from urllib2 import urlopen
+from odoo import api, fields, models
+from odoo.osv.expression import TRUE_LEAF
+
+GEOLOCALISATION_URL = u"http://ip-api.com/json/{}"
```
naglis: Not sure about pros/cons of existing online service alternatives, but maybe it would be better to use a service which has HTTPS in their free plan, eg. https://ipapi.co/json?

Also, I have tested http://ip-api.com/json/ locally and after a couple of hundreds of requests I got banned (I guess timeout based), so in a real world case I believe this could potentially break if the limit is reached within the time window, resulting in empty metadata?

IIUC this would also be stored upon successful logins? In that case, each new `res.authentication.attempt` record could potentially take ~5 seconds? What if we try to fill the metadata asynchronously afterwards, eg. with Cron or connector? This way you could also group records by remote address and reduce the number of requests in this way  (now it is repeated for the same IP).

Also, I am wondering why an external service is used here instead of eg. [GeoIP](https://github.com/odoo/odoo/blob/14f59f4b780fa1e757cfce6a46c62c7a6b00d13e/addons/website/geoipresolver.py#L16)?

auth_brute_force/tests/test_brute_force.py
```diff
@@ -0,0 +1,361 @@
...
+        with self.cursor() as cr:
+            env = self.env(cr)
+            self.assertFalse(
+                env["res.authentication.attempt"]._trusted(
+                    "127.0.0.1",
+                    "demo",
+                ),
+            )
+        # Attempts recorded
+        with self.cursor() as cr:
```
naglis, line 115: Readabillty: These test cases seem quite repetitive, maybe it would be possible to factor out the common parts to reduce code repetition?

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -2,35 +2,160 @@
...
+        if recent_failures >= limit:
+            _logger.warning(
+                "Authentication failed from remote '%s'. "
+                "The remote has been banned. "
+                "Login tried: %r.",
+                remote,
+                login,
+            )
+            return False
+        # Check if remote + login combination is banned
```
naglis, line 128: Readability: The `# Check if remote is banned` and `# Check if remote + login combination is banned` parts share much of the code, might be worth factoring out into separate helper function.

naglis (approved):
@Yajo thank you so much for the effort, can't wait to test it out!

Added my 2¢, hopefully some of it will be helpful.

auth_brute_force/__manifest__.py
```diff
@@ -3,22 +3,21 @@
...
+    'version': '10.0.2.0.0',
     'category': 'Tools',
-    'summary': "Tracks Authentication Attempts and Prevents Brute-force"
-               " Attacks module",
+    'summary': "Track Authentication Attempts and Prevent Brute-force Attacks",
     'author': "GRAP, "
               "Tecnativa, "
               "Odoo Community Association (OCA)",
-    'website': 'http://www.grap.coop',
+    'website': 'https://github.com/OCA/server-auth',
```
MiquelRForgeFlow: But this PR is in serve-tools 🤔

auth_brute_force/README.rst
```diff
@@ -69,14 +69,15 @@ For further information, please visit:
 Known issues / Roadmap
 ======================
 
-* The ID used to identify a remote request is the IP provided in the request
-  (key 'REMOTE_ADDR').
+* Remove 🐒 patch for https://github.com/odoo/odoo/issues/24183 when fixed.
```
MiquelRForgeFlow: This issue was closed. Solved for v12

rafaelbn:
@Yajo which is the state of this? OCA need this great work!

auth_brute_force/models/res_authentication_attempt.py
legalsylvain: Hi. I implemented the call of ip-api some years ago, but I did'nt made a benchmark about all the solutions, so maybe it's not the best solution.
Fill the metadata via a cron could be a great improvment indeed. And using GeoIP too ! (but maybe not the objective of this PR).

auth_brute_force/README.rst
yajo: It's because it's a sub-point for previous point.

auth_brute_force/data/ir_config_parameter.xml
yajo: I updated the README.

auth_brute_force/models/res_authentication_attempt.py
yajo: I agree with both: this should be improved, but this PR targets other purpose. I added it to known issues.

auth_brute_force/models/res_authentication_attempt.py
yajo: I didn't want to give so much importance to this failure, which is not such a big problem. In fact, I think I'm gonna lower it to a warning instead :thinking:

auth_brute_force/tests/test_brute_force.py
yajo: Yes, that's because before we didn't have a 2nd arg.

auth_brute_force/tests/test_brute_force.py
yajo: Possibly, but the little details make it quite difficult to refactor, and after all repeating code in test cases is not so bad as in other parts of the code.

auth_brute_force/models/res_authentication_attempt.py
yajo: Done

yajo:
Updated with all remarks attended.

pedrobaeza (approved):

PR #1221, opened 2018-04-16
Migrated date_range to 11.0

michael-villamar:


pedrobaeza:
Module is already migrated in https://github.com/OCA/server-ux/tree/11.0/date_range

PR #1224, opened 2018-04-20
[MIG][11.0] fetchmail_notify_error_to_sender

nedaszilinskas:


nedaszilinskas:
@pedrobaeza would you be so kind giving me a hint on how to solve the issues above (Missing base report)?

pedrobaeza:
Which issue?

nedaszilinskas:
The checks - codecov/patch and codecov/project say "Missing base report.". From docs I read:
_Best Practice: Always fork a branch from a successful CI commit that has coverage uploaded, as illustrated below._
Not sure I understand what its about since the previous commit on my fork was passing.

pedrobaeza:
I don't see such message, but the red crosses is due to the lack of tests on the modules that lower the test coverage. It can be acceptable in some cases where the function is not considered critical.

nedaszilinskas:
I do not see that there were tests for this module on any of the previous versions so I take its the case when it can be accepted?

pedrobaeza:
Yeah, that's a pending topic in several modules. It's a question of quality, for avoiding problems, but if you don't feel qualified for making them, I think reviewers can accept the migration itself without them.

nedaszilinskas:
At the moment I just do not have time to make improvements, I just want to contribute a migration.

pedrobaeza:
OK, as you prefer.

pedrobaeza:
This PR doesn't preserve commit history. Closing in favor of #1238, that already includes it.

nedaszilinskas:
@pedrobaeza my interest is to have this module on v11 as soon as possible. What we can do to help contributing here and fast-forwarding it?

pedrobaeza:
But the migration is not correct. Review the other PR.

nedaszilinskas:
@pedrobaeza what would be the process of contributing to that other PR (https://github.com/OCA/server-tools/pull/1238)?
From my understanding - I would need to fork "haidd:11.0-mig-fetchmail_notify_error_to_sender", make changes and create PR first to "haidd:11.0-mig-fetchmail_notify_error_to_sender"? That looks like too much of the process.

pedrobaeza:
@nedaszilinskas there are 2 possibilities:

* The one you mention, which is a bit more work, but it's a courtesy to the author of the other PR if he/she is still active, but is not able to perform the changes.
* If the author didn't answer in a time, which is the case, then you fork from the PR branch, but start a new PR you control, and then we superseed the PR. This is also not discourteous at all, as we are keeping his/her initial migration commit.

The second one is the suitable IMO for this case.

Issue #1227, opened 2018-04-22
auth_admin_passkey error when opening general settings with missing PR from 8, 9


mohamedhagag:
Dears,
why below merged commit in 8 and 9 by akretion avoided in v10 ?
https://github.com/OCA/server-tools/pull/519

The auth_admin_passkey is having bug in v10 since 4-5 months as below:
when opening general settings menu the passkey module through below error:
```
 Traceback (most recent call last):
  File "/opt/O10/odoo/odoo/odoo/http.py", line 641, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/O10/odoo/odoo/odoo/http.py", line 683, in dispatch
    result = self._call_function(**self.params)
  File "/opt/O10/odoo/odoo/odoo/http.py", line 333, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/O10/odoo/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/O10/odoo/odoo/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/O10/odoo/odoo/odoo/http.py", line 941, in __call__
    return self.method(*args, **kw)
  File "/opt/O10/odoo/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/opt/O10/odoo/odoo/addons/web/controllers/main.py", line 891, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/O10/odoo/odoo/addons/web/controllers/main.py", line 883, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/O10/odoo/odoo/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/opt/O10/odoo/odoo/odoo/api.py", line 672, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/opt/O10/odoo/odoo/odoo/addons/base/res/res_config.py", line 508, in default_get
    res.update(getattr(self, method)(fields))
  File "/opt/O10/.enabled/auth_admin_passkey/models/res_config.py", line 18, in get_default_auth_admin_passkey_send_to_admin
    'auth_admin_passkey.send_to_admin', 'True')),
TypeError: 'module' object is not callable
```

updating with the mentioned PR fixes the problem.

[...]

PR #1228, opened 2018-04-22
[10] auth_admin_passkey Fix Issue #1227

mohamedhagag:
this is a fix for issue:
https://github.com/OCA/server-tools/issues/1227 
based on the PR by akretion for 8 and 9 here:
https://github.com/OCA/server-tools/pull/519

### DIFF
--- a/auth_admin_passkey/models/res_config.py
+++ b/auth_admin_passkey/models/res_config.py
@@ -4,55 +4,49 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo import api, fields, models
-from odoo.tools import safe_eval
-
 
 class BaseConfigSettings(models.TransientModel):
     _inherit = 'base.config.settings'
 
+    # Getter / Setter Section
     @api.model
     def get_default_auth_admin_passkey_send_to_admin(self, fields):
-        icp = self.env['ir.config_parameter']
         return {
-            'auth_admin_passkey_send_to_admin': safe_eval(icp.get_param(
-                'auth_admin_passkey.send_to_admin', 'True')),
+            'auth_admin_passkey_send_to_admin':
+            self.env["ir.config_parameter"].get_param(
+                "auth_admin_passkey.send_to_admin")
         }
-
-    @api.model
-    def get_default_auth_admin_passkey_send_to_user(self, fields):
-        icp = self.env['ir.config_parameter']
-        return {
-            'auth_admin_passkey_send_to_user': safe_eval(icp.get_param(
-                'auth_admin_passkey.send_to_user', 'True')),
-        }
-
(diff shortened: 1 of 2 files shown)


pedrobaeza:

And check travis status

mohamedhagag:
Version number change will create another PR, is this OK ?

Regarding the travis status I can't find errors related to this module, it complains about other modules like auth_brute_force , auth_user_case_insensitive and other modules.
What should I do ?

Sorry but I'm new in contributing to OSS projects using github.

pedrobaeza:
@mohamedhagag no, adding version number doesn't create other PR if you do it through git, not via GitHub interface.

And the Travis errores are yours:
```
auth_admin_passkey/models/res_config.py:8:1: E302 expected 2 blank lines, found 1
auth_admin_passkey/models/res_config.py:17:63: E231 missing whitespace after ','
auth_admin_passkey/models/res_config.py:32:62: E231 missing whitespace after ','
```

mohamedhagag:
@pedrobaeza would you plz chk and merge.

pedrobaeza:
This is not how it works. Check https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md for the general rules. It requires at least 2 reviews, and a way to get reviews is to make another reviews and ask in return to review yours.

auth_admin_passkey/models/res_config.py
```diff
@@ -4,55 +4,50 @@
...
-            repr(self.auth_admin_passkey_send_to_user))
+    auth_admin_passkey_send_to_admin = fields.Boolean(
+        string='Send email to admin user.',
+        help="""When the administrator use his password to login in """
+             """with a different account, System will send an email """
+             """to the admin user.""")
+
+    auth_admin_passkey_send_to_user = fields.Boolean(
+        string='Send email to user.',
+        help="""When the administrator use his password to login in """
```
legalsylvain, line 51: Hi, 
- you moved a lot of code, that make the diff quite hard to review. I don't see the reason of that ?
- I think adding triple quote is unnecessary.

Other it looks great for me !

legalsylvain (changes requested):

NL66278:
I think the latest two commits, using  the string 'False' instead of an empty string when setting the parameters, and then testing for string 'True' or '1' when getting the parameters is not needed. The second commit fixes an error introduced by the first commit. Setting the parameter to an empty string effectively sets it to a False or rather falsey value. So original code as now existing in 8.0 just should work IMHO.

auth_admin_passkey/models/res_config.py
NL66278: I agree it is a good idea to get rid of the triple quotes. They were introduced in the fix for 8.0.

dek-odoo:
are there any updates ? The issue is still there.

gurneyalex:
superseeded by #1720

PR #1230, opened 2018-04-23
New module: smtp_connection_reuse

andreparames:
This module caches the SMTP connection, rather than creating a new one *for each email* as is Odoo's default behavior. 

Specifically, it keeps a connection cached for each Odoo thread (using thread locals), reusing them when the host, port and username are the same. It also sends a RSET command to make sure the connection is still usable; this is the [same behavior as Postfix](http://www.postfix.org/CONNECTION_CACHE_README.html).


smtp_connection_reuse/models/ir_mail_server.py
```diff
@@ -0,0 +1,55 @@
...
+        key = (host, port, user)
+        connection = connections_cache.connections.get(key)
+        if connection:
+            try:
+                # reset session, also serves to test connection
+                connection.rset()
+                return connection
+            except smtplib.SMTPException:
+                _logger.info("Existing SMTP connection is unusable")
+                pass
```
lmignon: @andreparames `pass`is useless

smtp_connection_reuse/README.rst
```diff
@@ -0,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Smtp Connection Reuse
+=====================
+
+Allows an SMTP connection to be reused between emails, increasing the
+throughput.
```
lmignon, line 10: @andreparames Do you have some metrics on how it improves the performances?

smtp_connection_reuse/models/ir_mail_server.py
```diff
@@ -0,0 +1,55 @@
...
+        pass
+
+
+class IrMailServer(models.Model):
+
+    _inherit = 'ir.mail_server'
+
+    def connect(self, host, port, user=None, password=None,
+                encryption=False, smtp_debug=False):
+        if getattr(connections_cache, 'connections', None) is None:
```
lmignon, line 33: @andreparames Why `getattr` and not `if not connection_cache.connections`?

smtp_connection_reuse/models/ir_mail_server.py
```diff
@@ -0,0 +1,55 @@
...
+                return connection
+            except smtplib.SMTPException:
+                _logger.info("Existing SMTP connection is unusable")
+                pass
+        _logger.info("Creating new SMTP connection")
+        connection = super(IrMailServer, self).connect(
+            host, port, user=user, password=password, encryption=encryption,
+            smtp_debug=True)
+        if connection.__class__ is smtplib.SMTP:
+            connection.__class__ = SMTPNoQuit
```
lmignon, line 49: @andreparames I've never seen this way of patching an object instance. That would be less intrusive to only patch the method `quit` by a direct assignment of an empty method:  `connection.quit = lambda a: pass`

andreparames: It's for a planned future improvement: I want to implement `__del__` in the new classes, to close the connection when the thread ends. I'm not sure if patching `connection.__del__` would work correctly.

smtp_connection_reuse/models/ir_mail_server.py
andreparames: Each thread has its own attributes, so `connections` might not exist for the current thread. The problematic line is actually the 11th, since it gives the idea it's creating `connections` for all threads, when in fact it only creates it for the current. I'll remove the line.

smtp_connection_reuse/README.rst
andreparames: Not yet, I'll prepare a dataset with a few hundred emails to do a benchmark.

PR #1231, opened 2018-04-24
FIX remove old api parameters in base_exception

bealdav:
ping @simahawk @mourad-ehm @jbeficent 
thanks for a review

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -56,7 +56,9 @@ class ExceptionRule(models.Model):
 #       base.exception line (ex rule_group = sale for sale order)
 #  - object: same as order or line, browse_record of the base.exception or
 #    base.exception line
-#  - pool: ORM model pool (i.e. self.pool)
+#  - pool: ORM model pool (i.e. self.pool, deprecated in new api)
+#  - obj: same as object
+#  - env: ORM model pool (i.e. self.env)
 #  - time: Python time module
 #  - cr: database cursor
 #  - uid: current user id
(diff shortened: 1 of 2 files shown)


base_exception/models/base_exception.py
```diff
@@ -162,16 +163,16 @@ def detect_exceptions(self):
     def _exception_rule_eval_context(self, obj_name, rec):
         user = self.env['res.users'].browse(self._uid)
         return {obj_name: rec,
-                'self': self.pool.get(rec._name),
```
simahawk: this is not the same thing
```python
>>> self.pool.get('res.users')
<class 'odoo.api.res.users'>
>>> self.env.get('res.users')
res.users()
```
Make sure rules are not impacted. An option can be to keep `self` and deprecate it until v12 and use `model` key for the new one (which seems a better name too IMO).

simahawk (changes requested):

base_exception/models/base_exception.py
```diff
@@ -162,16 +163,16 @@ def detect_exceptions(self):
...
+                'self': self.env.get(rec._name),
                 'object': rec,
                 'obj': rec,
-                'pool': self.pool,
-                'cr': self._cr,
-                'uid': self._uid,
+                'env': self.env,
+                'cr': self.env.cr,
+                'uid': self.env.uid,
```
yvaucher, line 172: Are you keeping self.env.cr and self.env.uid for backward compatibility? Could it be deprecated to use only env ? Maybe add a note for version 12.0 if we can simplify it.

Or is it to have helpers in `safe_eval`

base_exception/models/base_exception.py
```diff
@@ -162,16 +163,16 @@ def detect_exceptions(self):
     def _exception_rule_eval_context(self, obj_name, rec):
         user = self.env['res.users'].browse(self._uid)
```
yvaucher: This can be replaced by `self.env.user` no need for a browse.

base_exception/models/base_exception.py
bealdav: DONE

bealdav: ping @yvaucher

yvaucher: Ok self.env.user is now used. Thanks.

yvaucher (approved):

mourad-ehm (approved):

bealdav:
ping @simahawk please could you check ?

bealdav:
ping @pedrobaeza could you merge, please. Thanks

pedrobaeza:
I suppose this should be forward-ported to v11.

bealdav:
Done there https://github.com/OCA/server-tools/pull/1433

PR #1232, opened 2018-04-26
[9.0][ADD] cron_daylight_saving_time_resistant

hparfr:
Because some times you want to set a cron to run everyday at 0:00 and not each 24hours.


cron_daylight_saving_time_resistant/models/ir_cron.py
```diff
@@ -0,0 +1,56 @@
...
+from openerp.addons.base.ir.ir_cron import _intervalTypes
+_logger = logging.getLogger(__name__)
+
+
+class IrCron(models.Model):
+    _inherit = "ir.cron"
+
+    daylight_saving_time_resistant = fields.Boolean(
+        help="Adjust interval to run at the same hour after and before"
+        "daylight saving time change. It's used twice a year")
```
legalsylvain, line 21: I don't understand why you say that it is used twice a year. if it run at the same hour, it will change the hour of the run, during 6 monthes ? 
Or maybe I totally missed the aim of this module ?

thanks for the clarification !

florian-dacosta: Sylvain,
The 'nextcall' datetime of crons, like every datetime field in odoo is stored in UTC.
So if you have a cron, planned for 10H00 UTC time, which runs once a day, in the french context, it will be executed at 10H+2H=12H during summer and 10H+1H=11H during winter.
The aim of this module is to avoid this difference.

legalsylvain: OK. understood correctly the objective of the module. I was just confused wit the helper of this field, that says that the value is used twice a year. In fact, the impact is during the winter zone, so during 6 monthes.

:+1: for this module. no time to make a deep review. sorry.

PR #1239, opened 2018-05-02
[FIX] Avoid crash when using auth_session_timeout module with XML-RPC api

xtof-osd:
When auth_session_timeout is installed and you try to use the XML-RPC api, any request will failed because http.request is not set,  with the following stack:

```
  File "/usr/lib/python2.7/dist-packages/odoo/http.py", line 118, in dispatch_rpc
    result = dispatch(method, params)
  File "/usr/lib/python2.7/dist-packages/odoo/service/model.py", line 35, in dispatch
    security.check(db,uid,passwd)
  File "/usr/lib/python2.7/dist-packages/odoo/service/security.py", line 13, in check
    return res_users.check(db, uid, passwd)
  File "/home/odootest/addons/10.0/server-tools/auth_session_timeout/models/res_users.py", line 108, in check
    _logger.warning("Testing: %s -- %s "%(str(res),str(http.request)))
  File "/usr/lib/python2.7/dist-packages/werkzeug/local.py", line 357, in <lambda>
    __str__ = lambda x: str(x._get_current_object())
  File "/usr/lib/python2.7/dist-packages/werkzeug/local.py", line 297, in _get_current_object
    return self.__local()
  File "/usr/lib/python2.7/dist-packages/werkzeug/local.py", line 132, in _lookup
    raise RuntimeError('object unbound')
RuntimeError: object unbound
```


auth_session_timeout/models/res_users.py
```diff
@@ -104,6 +104,10 @@ def _auth_timeout_check(self):
 
     @classmethod
     def check(cls, *args, **kwargs):
+
```
pedrobaeza, line 107: Don't add extra empty lines

auth_session_timeout/models/res_users.py
```diff
@@ -4,13 +4,13 @@
...
 import logging
-
+from os import utime
 from os.path import getmtime
 from time import time
-from os import utime
 
 from odoo import api, http, models
 
+
```
pedrobaeza, line 13: Don't add extra empty lines

pedrobaeza (changes requested):

auth_session_timeout/__manifest__.py
```diff
@@ -14,7 +14,7 @@
     'maintainer': 'Odoo Community Association (OCA)',
     'website': "http://acsone.eu",
     'category': 'Tools',
-    'version': '10.0.1.0.1',
+    'version': '10.0.2.0.1',
```
pedrobaeza, line 17: This is very little change, so it should be `10.0.1.0.2`

pedrobaeza:
You have messed up the PR with unrelated commits.

xtof-osd:
I need to cleanup the mess before re-submitting => close PR

PR #1240, opened 2018-05-04
[10.0][IMP] Keychain : Add default validation logic

hparfr:
For each keychain implementation (namespace), you have to define what data you expect and validate it.

Before :
If you don't use any data in your namespace: you still have to define these methods (initalize and validate)

After:
If you don't use any data in your namespace, no initialize and no validate method to write (there is default methods)

More info : 
The goal was to force developers to write validation logic.
Now it's becoming useless with keychain.backend
And it adds boilerplate code.

In this commit, we set a permissive default validation logic.
It will reduce code in consumers module and still allow fine grained validation.

### DIFF
--- a/keychain/tests/test_keychain.py
+++ b/keychain/tests/test_keychain.py
@@ -79,14 +79,14 @@ class TestKeychain(TransactionCase):
         config['keychain_key'] = Fernet.generate_key()
         try:
             account._get_password()
-            self.assertTrue(False, 'It should not work with another key')
+            self.fail('It should not work with another key')
         except UserError as err:
             self.assertTrue(True, 'It should raise a UserError')
             self.assertTrue(
                 'has been encrypted with a diff' in str(err),
                 'It should display the right msg')
         else:
-            self.assertTrue(False, 'It should raise a UserError')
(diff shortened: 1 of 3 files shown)


hparfr:
ping @sebastienbeau

keychain/tests/test_keychain.py
```diff
@@ -221,3 +221,24 @@ def test_valid_json(self):
...
+        account = self.keychain.create({
+            "name": "test",
+            "namespace": "keychain_test_default",
+            "login": "test",
+            "technical_name": "keychain.test"
+        })
+        try:
+            account.write({"login": "test default"})
+        except ValidationError:
+            self.assertTrue(False, 'It should validate any json in default')
```
lmignon: @hparfr you should use ` self.fail(msg)¶` in place of `self.assertTrue(False, msg)`

lmignon (approved):
@hparfr Good idea! A little comment otherwise LGTM. (code review only)

bealdav (approved):

sebastienbeau (approved):
code review ok for me

PR #1243, opened 2018-05-11
[FIX] Avoid crash when using auth_session_timeout module with XML-RPC API

xtof-osd:
Made some cleanup before re-submitting PR (needed to create a brand new 10.0branch).
Thanks @pedrobaeza  for all the comments

### DIFF
--- a/auth_session_timeout/models/res_users.py
+++ b/auth_session_timeout/models/res_users.py
@@ -105,5 +105,6 @@ class ResUsers(models.Model):
     @classmethod
     def check(cls, *args, **kwargs):
         res = super(ResUsers, cls).check(*args, **kwargs)
-        http.request.env.user._auth_timeout_check()
+        if http.request:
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

auth_session_timeout/models/res_users.py
```diff
@@ -104,6 +104,8 @@ def _auth_timeout_check(self):
 
     @classmethod
     def check(cls, *args, **kwargs):
```
fmdl: This could be better, alway return the super(), no ? @pedrobaeza 

```
    def check(cls, *args, **kwargs):
        res = super(ResUsers, cls).check(*args, **kwargs)
        if http.request:
             http.request.env.user._auth_timeout_check()
        return res
```

xtof-osd: Did make the change according to @fmdl  comment

bealdav (approved):

PR #1244, opened 2018-05-13
11.0 mig letsencrypt

acysos:


acysos:
The travis test fails becuase can't load acme_tiny library, despite of it's in __manifest__.py, can some one tell me what happens?

Greetings

letsencrypt/data/ir_cron.xml
```diff
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
     <data noupdate="1">
-        <record id="cronjob" model="ir.cron">
+        <record id="letsencrypt_cron" model="ir.cron">
```
hbrunn: if you insist on renaming the xmlid, we'll need a migration script for updates to go smoothly. I don't see a reason to do so to be honest, because the full xmlid will be `letsencrypt.cronjob`, which sound perfectly intelligible to me

hbrunn (changes requested):
with the dependency in the manifest, you only tell Odoo to check if the module exists on the target system. For CI, you'll have to take care it's there by adding it to `requirements.txt`: https://github.com/OCA/server-tools/blob/10.0/requirements.txt#L3 (both acme_tiny and iPy)

letsencrypt/data/ir_cron.xml
acysos: I rename it, to don't use a too general name. I can return it. It's no problem.

acysos:
Ok, thank you. 
I have added in requirements the dependencies.

Greetings

letsencrypt/demo/ir_cron.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <openerp>
     <data>
-        <record id="cronjob" model="ir.cron">
+        <record id="letsencrypt_cron" model="ir.cron">
```
hbrunn: don't forget to undo the renaming everywhere

acysos: I see it now, when I check the differences. Thank you.

acysos:
Ok, all test passed. Thank you.

hbrunn (approved):

pedrobaeza (approved):

pedrobaeza:
I recommend you to check https://gitlab.tecnativa.com/Tecnativa/docker-odoo-base for a new integrated system with LE certificate auto-download, but any way, merging this as trivial migration and with tests passing.

PR #1247, opened 2018-05-16
[REF] auth_signup_verify_email: Skip mail send in tests and use email_validator

yajo:
This addon introduced an integration conflict when tested in a database that had `mail_tracking_mass_mailing` installed, producing this failure:

    File "/opt/odoo/auto/addons/auth_signup_verify_email/controllers/main.py", line 44, in passwordless_signup
        sudo_users.reset_password(values.get("login"))
    File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
        self.gen.next()
    File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 419, in savepoint
        self.execute('RELEASE SAVEPOINT "%s"' % name)
    File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 154, in wrapper
    File "/opt/odoo/custom/src/odoo/odoo/sql_db.py", line 231, in execute
    InternalError: no such savepoint

Which in turn produced the following in the next test:

    InternalError: current transaction is aborted, commands ignored until end of transaction block

The problem comes from the fact that one cannot rollback to a nested savepoint if the parent savepoint was released. It became a problem because that's the strategy that both this addon and upstream's `TestCursor` follow.

To avoid that, tests now mock the `send_mail` method. This results also in having a predictable outcome from the test `test_good_email`, so it is more meaningful now.

Besides, previously we were using the `validate_email` package, which is currently a dead project that can silently fail under certain environments, as seen in syrusakbary/validate_email#80.

There's the `email_validator` package, freely available, supported, and it provides a human-readable error message whenever some format from the email fails.

As such, here I'm switching the dependency, while still adding a backwards compatibility layer for preexisting installations

@Tecnativa

### DIFF
--- a/auth_signup_verify_email/README.rst
+++ b/auth_signup_verify_email/README.rst
@@ -45,6 +45,8 @@ Known issues / Roadmap
 * Remove calls to ``cr.commit()`` in tests when
   https://github.com/odoo/odoo/issues/12237 gets fixed.
 
+* Drop support for ``validate_email`` in favor of ``email_validator``.
+
 Bug Tracker
 ===========
 
--- a/auth_signup_verify_email/__manifest__.py
+++ b/auth_signup_verify_email/__manifest__.py
@@ -1,14 +1,12 @@
 # -*- coding: utf-8 -*-
-# © 2015 Antiun Ingeniería, S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Verify email at signup",
     "summary": "Force uninvited users to use a good email for signup",
-    "version": "10.0.1.0.0",
+    "version": "10.0.2.0.0",
     "category": "Authentication",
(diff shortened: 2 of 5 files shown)


yajo:
This evolved from a fix to a little refactoring. I updated the description. The underlying validation library was having some side effects and it wasn't working in optimized environments, as one can see in https://github.com/syrusakbary/validate_email/pull/80.

However, I doubt that gets ever merged, due to being a dead project nowadays, so I switched to a newer and maintained alternative.

The original fix is included also in the commit, so tests are now more reliable.

rafaelbn (approved):

rafaelbn:
Please @Yajo could you correct my email in Contributors section? Thanks

chienandalu (approved):

yajo:
Please review the forward-port of this PR to v11: https://github.com/OCA/server-auth/pull/29

PR #1250, opened 2018-05-21
[FIX] auth_brute_force: Small typos

yajo:
- The `whitelisted` field needs to exist in view to be usable.
- The correct class is `decoration-danger` for tree views.

@Tecnativa


angelmoya (approved):
👍

pedrobaeza (approved):
Fast-tracking this

PR #1251, opened 2018-05-21
[FIX] auth_brute_force: Fix addon requirement computation

yajo:
Include HACK for https://github.com/odoo/odoo/pull/24833, which explains the false positive problem we were having here: an addon being importable doesn't mean it is installed.

@Tecnativa

### DIFF
--- a/auth_brute_force/__manifest__.py
+++ b/auth_brute_force/__manifest__.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Authentification - Brute-Force Filter',
-    'version': '10.0.2.1.0',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I don't get it. `web` module is the base for any installation. Why is not going to be installed?

rafaelbn (approved):
:+1:  Trust

yajo:
It's a very corner case, but it's actually not in the dependency chain.

pedrobaeza:
But where are you using web resources? If so, you have to add the dependency.

yajo:
Just in the test, but adding a dependency just to run a test is unnecessary, when you can just skip the test if it is not installed.

pedrobaeza (approved):
I see this very convoluted for only that, but OK, let's move on...

PR #1253, opened 2018-05-22
[IMP] make posible to call the onchange on an existing record and avoid returning computed value

sebastienbeau:
Make play onchange working on existing record, see documentation.


acsonefho (commented):
LGTM (code, not tested)

acsonefho (approved):

legalsylvain (approved):
:+1:

onchange_helper/models/ir_rule.py
```diff
@@ -26,7 +26,10 @@ def play_onchanges(self, values, onchange_fields):
     all_values = values.copy()
     for field in self._fields:
         if field not in all_values:
-            all_values[field] = False
+            # If self is a record (play onchange on existing record)
+            # we take the value of the field
+            # If self is an empty record we will have an empty value
+            all_values[field] = self[field]
```
lmignon: you should use convert to write to transform the record's properties into the format expected by 'new'

onchange_helper/README.rst
```diff
@@ -18,12 +18,21 @@ To use this module, you need to:
...
 
+You can also use it on existing record for example:
+
+    `vals = {'partner_shipping_id': 1}`
+
+    `vals = sale.play_onchanges(vals, ['partner_shipping_id'])`
+
+Then the onchange will be played with the vals passed and the existing vals of the sale. `vals` will be updated with partner_invoice_id, pricelist_id, etc...
+
+
```
lmignon: @sebastienbeau IMO we should explain why this way to trigger the `onchanges` is safe and generic....

"Behind the scene, `play_onchanges` will execute all the methods registered for the list of changed fields. To avoid performance issue when the method is called on a record , the record will be transformed into a memory record before calling the registered methods to avoid to trigger SQL updates command when values are assigned to the record by the onchange"

lmignon (approved):

PR #1255, opened 2018-05-24
[MIG] auth_brute_force: Backport from v10

yajo:
Odoo v9 auth methods usually violate both old and new api; they just only work in old api, so I had to adapt some tests to use it instead.

Normal backport outside of that.

@Tecnativa


rafaelbn (approved):

pedrobaeza (approved):
Great work

PR #1256, opened 2018-05-24
[FIX] password_security: Default last write date

yajo:
* Add a default `password_write_date` to circumvent need for immediate password reset, fixing #1083 

Backport of #1084 

@Tecnativa


pedrobaeza (approved):
Fast-tracking as direct backport

PR #1258, opened 2018-05-25
[FIX] auth_brute_force: Avoid storing false login attempts

yajo:
In Odoo v9, every request calls `res.users.check()`, which stores one authentication attempt per request, which is false.

Besides, it easily leads to hitting ip-api.com rate limits, so now that API is only asked when seeing in form view (simply by setting the computed field as not stored).

Also, form view was hidden, so it's now visible.

@Tecnativa


pedrobaeza (approved):

rafaelbn (approved):

pedrobaeza:
Merging as critical for avoiding bans.

Issue #1179, opened 2018-03-05
Traceback in audit log read rule


CasVissers-360ERP:
Hi,

When adding a read rule we get a traceback when opening a record:
```
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/customer/odoo/odoo/http.py", line 640, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/customer/odoo/odoo/http.py", line 677, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/customer/odoo/odoo/http.py", line 333, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/customer/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/customer/odoo/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/customer/odoo/odoo/http.py", line 935, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/customer/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/customer/odoo/addons/web/controllers/main.py", line 885, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/customer/odoo/addons/web/controllers/main.py", line 877, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/customer/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/customer/odoo/odoo/api.py", line 678, in call_kw_multi
    recs = self.with_context(context or {}).browse(ids)
  File "/opt/odoo/customer/odoo/odoo/models.py", line 4889, in with_context
    context = dict(args[0] if args else self._context, **kwargs)
ValueError: dictionary update sequence element #0 has length 20; 2 is required
```

Thank you.
Cas

rockcesar:
I have the same problem. My error shows:



Odoo Server Error

Traceback (most recent call last):
  File "/odoo/odoo-10-server/odoo/http.py", line 641, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/odoo/odoo-10-server/odoo/http.py", line 683, in dispatch
    result = self._call_function(**self.params)
[...]

PR #1259, opened 2018-05-28
Bugfixed, published in this page:

GlobalManager:
https://github.com/OCA/server-tools/issues/1179

And Chilean translation added.

### DIFF
--- a/auditlog/models/http_request.py
+++ b/auditlog/models/http_request.py
@@ -2,6 +2,8 @@
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from psycopg2.extensions import AsIs
+
 from odoo import models, fields, api
 from odoo.http import request
 
@@ -45,14 +47,20 @@ class AuditlogHTTPRequest(models.Model):
         first call.
         If no HTTP request is available, returns `False`.
         """
-
         if not request:
             return False
         http_session_model = self.env['auditlog.http.session']
         httprequest = request.httprequest
         if httprequest:
             if hasattr(httprequest, 'auditlog_http_request_id'):
-                return httprequest.auditlog_http_request_id
+                # Verify existence. Could have been rolled back after a
+                # concurrency error
+                self.env.cr.execute(
+                    "SELECT id FROM %s WHERE id = %s", (
+                        AsIs(self._table),
+                        httprequest.auditlog_http_request_id))
+                if self.env.cr.fetchone():
+                    return httprequest.auditlog_http_request_id
             vals = {
                 'name': httprequest.path,
                 'root_url': httprequest.url_root,
@@ -62,5 +70,4 @@ class AuditlogHTTPRequest(models.Model):
             }
             httprequest.auditlog_http_request_id = self.create(vals).id
             return httprequest.auditlog_http_request_id
-
         return False
--- a/auditlog/models/http_request.py
+++ b/auditlog/models/http_request.py
@@ -2,8 +2,6 @@
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from psycopg2.extensions import AsIs
-
 from odoo import models, fields, api
 from odoo.http import request
 
@@ -47,20 +45,14 @@ class AuditlogHTTPRequest(models.Model):
         first call.
         If no HTTP request is available, returns `False`.
         """
+
         if not request:
             return False
         http_session_model = self.env['auditlog.http.session']
         httprequest = request.httprequest
         if httprequest:
             if hasattr(httprequest, 'auditlog_http_request_id'):
-                # Verify existence. Could have been rolled back after a
-                # concurrency error
-                self.env.cr.execute(
-                    "SELECT id FROM %s WHERE id = %s", (
-                        AsIs(self._table),
-                        httprequest.auditlog_http_request_id))
-                if self.env.cr.fetchone():
-                    return httprequest.auditlog_http_request_id
+                return httprequest.auditlog_http_request_id
             vals = {
                 'name': httprequest.path,
                 'root_url': httprequest.url_root,
@@ -70,4 +62,5 @@ class AuditlogHTTPRequest(models.Model):
             }
             httprequest.auditlog_http_request_id = self.create(vals).id
             return httprequest.auditlog_http_request_id
+
         return False
--- a/auditlog/models/http_session.py
+++ b/auditlog/models/http_session.py
@@ -41,6 +41,7 @@ class AuditlogtHTTPSession(models.Model):
         first call.
         If no HTTP user session is available, returns `False`.
         """
+
         if not request:
             return False
         httpsession = request.session
@@ -57,4 +58,5 @@ class AuditlogtHTTPSession(models.Model):
             }
             httpsession.auditlog_http_session_id = self.create(vals).id
             return httpsession.auditlog_http_session_id
+
         return False
--- a/auditlog/models/http_session.py
+++ b/auditlog/models/http_session.py
@@ -41,7 +41,6 @@ class AuditlogtHTTPSession(models.Model):
         first call.
         If no HTTP user session is available, returns `False`.
         """
-
         if not request:
             return False
         httpsession = request.session
@@ -58,5 +57,4 @@ class AuditlogtHTTPSession(models.Model):
             }
             httpsession.auditlog_http_session_id = self.create(vals).id
             return httpsession.auditlog_http_session_id
-
         return False
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -263,7 +263,7 @@ class AuditlogRule(models.Model):
                 env = api.Environment(cr, uid, {'auditlog_disabled': True})
                 rule_model = env['auditlog.rule']
                 rule_model.sudo().with_context({'auditlog_disabled': True}).\
-                create_logs(
+                    create_logs(
                     env.uid, self._name, ids,
                     'read', read_values, None, {'log_type': log_type})
             # New API
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -2,7 +2,7 @@
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models, fields, api, modules, _, sql_db
+from odoo import models, fields, api, modules, _
 
 FIELDS_BLACKLIST = [
     'id', 'create_uid', 'create_date', 'write_uid', 'write_date',
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -257,12 +257,12 @@ class AuditlogRule(models.Model):
                 # avoid logs on `read` produced by auditlog during internal
                 # processing: read data of relevant records, 'ir.model',
                 # 'ir.model.fields'... (no interest in logging such operations)
-                if self.env.context.get('context', {}).\
+                if self.env.context.get('context', {}).
                     get('auditlog_disabled'):
                     return result
                 env = api.Environment(cr, uid, {'auditlog_disabled': True})
                 rule_model = env['auditlog.rule']
-                rule_model.sudo().with_context({'auditlog_disabled': True}).\
+                rule_model.sudo().with_context({'auditlog_disabled': True}).
                     create_logs(
                     env.uid, self._name, ids,
                     'read', read_values, None, {'log_type': log_type})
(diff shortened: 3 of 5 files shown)


pedrobaeza:
You are mixing several things in the same patch. Please change only the needed for solving the issue, and you can change translations in Transifex: https://odoo-community.org/page/translate

GlobalManager:
Some errors were encountered, related to the version of npm in the server, for example:

/home/travis/.npm/_logs/2018-05-28T01_33_46_147Z-debug.log

GlobalManager:
@gurneyalex 
@pedrobaeza 

Watch this. Some errors were encountered, related to the version of npm in the server, for example:

/home/travis/.npm/_logs/2018-05-28T01_33_46_147Z-debug.log

pedrobaeza:
The error seems to be due to too many requests.

Any way, I see that you are changing too much things and removing features of the module, so please keep the minimal diff for fixing the problem. If there are several things to fix, only one per PR.

GlobalManager:
All fixes are uploaded.

GlobalManager:
You can merge whenever you can. Thanks.

GlobalManager:
Can you check this?. @pedrobaeza

Thanks

auditlog/models/http_request.py
```diff
@@ -47,20 +45,14 @@ def current_http_request(self):
         first call.
         If no HTTP request is available, returns `False`.
         """
+
```
sebalix: Is this empty line required? It introduces a useless change which can create trouble when backporting patches between branches.

auditlog/models/http_request.py
```diff
@@ -70,4 +62,5 @@ def current_http_request(self):
             }
             httprequest.auditlog_http_request_id = self.create(vals).id
             return httprequest.auditlog_http_request_id
+
```
sebalix: Same remark, is this empty line required?

auditlog/models/http_request.py
```diff
@@ -47,20 +45,14 @@ def current_http_request(self):
...
             if hasattr(httprequest, 'auditlog_http_request_id'):
-                # Verify existence. Could have been rolled back after a
-                # concurrency error
-                self.env.cr.execute(
-                    "SELECT id FROM %s WHERE id = %s", (
-                        AsIs(self._table),
-                        httprequest.auditlog_http_request_id))
-                if self.env.cr.fetchone():
-                    return httprequest.auditlog_http_request_id
+                return httprequest.auditlog_http_request_id
```
sebalix: This piece of code was added maybe in 8.0 or 9.0 to avoid some error, are you sure it is not required anymore? You got some problem with it?

auditlog/models/http_session.py
```diff
@@ -41,6 +41,7 @@ def current_http_session(self):
         first call.
         If no HTTP user session is available, returns `False`.
         """
+
```
sebalix: Is this empty line required?

auditlog/models/http_session.py
```diff
@@ -57,4 +58,5 @@ def current_http_session(self):
             }
             httpsession.auditlog_http_session_id = self.create(vals).id
             return httpsession.auditlog_http_session_id
+
```
sebalix: Is this empty line required?

auditlog/models/rule.py
```diff
@@ -4,6 +4,9 @@
 
 from odoo import models, fields, api, modules, _, sql_db
 
+import logging
+_logger = logging.getLogger(__name__)
+
```
sebalix: `_logger` does not seem to be used in this file.

auditlog/models/rule.py
```diff
@@ -237,28 +239,31 @@ def _make_read(self):
...
             # Also, we can not modify the current result as it will break calls
             result2 = result
             if not isinstance(result2, list):
                 result2 = [result]
             read_values = dict((d['id'], d) for d in result2)
             # Old API
-            if args and isinstance(args[0], sql_db.Cursor):
-                cr, uid, ids = args[0], args[1], args[2]
+            if self and isinstance(self.env.cr, sql_db.Cursor):
+                cr, uid, ids = self.env.cr, self.env.uid, self.ids
```
sebalix: This code is to be compatible with old API which can always be present in a Odoo 10.0 project (modules not standard nor in OCA), that's why the signature of the method was `*args, **kwargs` to inspect the content of `args`.
If your call to `read(...)` is made with the new API, the rule creation is already satisfied below line 264 ("New API", https://github.com/OCA/server-tools/pull/1259/files#diff-e2cdd2f4b9da04f93936a15fa53fef79R269).

auditlog/models/rule.py
```diff
@@ -362,30 +367,36 @@ def create_logs(self, uid, res_model, res_ids, method,
...
+                if method is 'create':
+                    self._create_log_line_on_create(
+                        log, diff.added(), new_values)
+                elif method is 'read':
+                    self._create_log_line_on_read(
+                        log, old_values.get(res_id, EMPTY_DICT).keys(),
+                        old_values)
+                elif method is 'write':
+                    self._create_log_line_on_write(
+                        log, diff.changed(), old_values, new_values)
```
sebalix: If the only change is the condition `if res_name and res_name != "False"`, limit the diff by rewriting it to use the `continue` statement as we are in a `for` loop.
But this change also need some clarification, have you an example of how you encountered a "False" string value?

sebalix (changes requested):
For sure some bugs need to be fixed in `auditlog`. Some remarks, but in general we need some explanation to understand the different fixes proposed here.

auditlog/models/rule.py
GlobalManager: There is, always, a case when creates a register of the action performed, but creates two of them each time, one with (example): res.partned(1,), and the other the same res.partned(1,), but the first the attributes in False, like the name, although it's the object with the correct ID: res.partned(1,).

auditlog/models/rule.py
GlobalManager: I can let the signature as *args, but the signature *args, **kwargs, was what was generating the error, fixed by this patch. The actual signature for read in Odoo 10.0, is:

def read(self, fields=None, load='_classic_read'): #Two parameters

I can let the signature like:
def read(self, *args):

So the next sentence doesn't make sense in this case: cr, uid, ids = args[0], args[1], args[2]

I guess read in Odoo 10.0, isn't compatible with old api anymore. The module is working.

auditlog/models/rule.py
sebalix: Can you write
```python
if not res_name or res_name == "False":
   continue
```
to avoid an indentation level? It makes the diff very big (and harder to backport too).

sebalix (changes requested):

GlobalManager:
@sebalix We did some changes and fix the feature uploaded.

The reads, writes, creates and unlinks were going to be registered even if the read, create, write, unlink, weren't checked. We fixed this issue.

GlobalManager:
To reproduce the issue with the last code of auditlog: Set the rule for res.users and subscribe it, in the rules view, without marking any checkbox. Then perform one read of res.users, that can be made reloading the Odoo's webpage. The read will be registered, even if the read checkbox wasn't checked.

StefanRijnhart:
@GlobalManager can you create a separate PR for that last bugfix? BTW. the idea is that if the rule is not enabled for an operation, the model is not patched for that method: https://github.com/OCA/server-tools/blob/10.0/auditlog/models/rule.py#L131 Can you fix the problem there instead?

GlobalManager:
But there are 4 checkbox to be checked, if the checkbox aren't checked, the rules shouldn't be executed. Is the normal flow of a understandable module.

GlobalManager:
Create patching method is working in the current patch to be merged: https://github.com/OCA/server-tools/blob/10.0/auditlog/models/rule.py#L131

StefanRijnhart:
Thanks for isolating that one! Let's continue the discussion on the 11.0 port that you already created here: https://github.com/OCA/server-tools/pull/1264

GlobalManager:
Once it is merged, I will apply the others changes, you told me to delete.

GlobalManager:
Because, the rest of the changes depend of the changes in this PR.

auditlog/models/rule.py
sebalix: Okay, I'm testing a piece of code with an `read` overload in old API, and it seems it's not compatible anymore with the new API in Odoo 10.0.
So OK to drop the support for old API. But regarding the fix, the one proposed here is wrong. The `_make_read` method should be written like this:
```python

        @api.multi
        def read(self, fields=None, load='_classic_read'):
            result = read.origin(self, fields, load)
            read_values = dict((d['id'], d) for d in result)
            # If the call came from auditlog itself, skip logging:
            # avoid logs on `read` produced by auditlog during internal
            # processing: read data of relevant records, 'ir.model',
            # 'ir.model.fields'... (no interest in logging such operations)
            if self.env.context.get('auditlog_disabled'):
                return result
            rule_model = self.env['auditlog.rule']
                self.env.uid, self._name, self.ids,
                'read', read_values, None, {'log_type': log_type})
            return result

        return read
```

auditlog/models/rule.py
```diff
@@ -237,28 +237,31 @@ def _make_read(self):
...
                 # 'ir.model.fields'... (no interest in logging such operations)
-                if kwargs.get('context', {}).get('auditlog_disabled'):
+                if self.env.context.get('context', {}).\
+                   get('auditlog_disabled'):
                     return result
                 env = api.Environment(cr, uid, {'auditlog_disabled': True})
                 rule_model = env['auditlog.rule']
-                rule_model.sudo().create_logs(
+                rule_model.sudo().with_context({'auditlog_disabled': True}).\
+                    create_logs(
```
sebalix: Why changing this? the environment has already the `auditlog_disabled` set to True (two lines above).

auditlog/models/rule.py
```diff
@@ -362,27 +365,34 @@ def create_logs(self, uid, res_model, res_ids, method,
...
                 'model_id': self.pool._auditlog_model_cache[res_model],
                 'res_id': res_id,
                 'method': method,
                 'user_id': uid,
-                'http_request_id': http_request_model.current_http_request(),
-                'http_session_id': http_session_model.current_http_session(),
+                'http_request_id': http_request_model.
+                current_http_request(),
+                'http_session_id': http_session_model.
+                current_http_session(),
```
sebalix: Do not change these lines, this is not necessary.

auditlog/models/rule.py
```diff
@@ -362,27 +365,34 @@ def create_logs(self, uid, res_model, res_ids, method,
...
             }
             vals.update(additional_log_values or {})
             log = log_model.create(vals)
             diff = DictDiffer(
                 new_values.get(res_id, EMPTY_DICT),
                 old_values.get(res_id, EMPTY_DICT))
             if method is 'create':
-                self._create_log_line_on_create(log, diff.added(), new_values)
+                self._create_log_line_on_create(
+                    log, diff.added(), new_values)
```
sebalix: Do not change these lines, this is not necessary.

auditlog/models/rule.py
```diff
@@ -362,27 +365,34 @@ def create_logs(self, uid, res_model, res_ids, method,
...
             elif method is 'read':
                 self._create_log_line_on_read(
-                    log, old_values.get(res_id, EMPTY_DICT).keys(), old_values)
+                    log, old_values.get(res_id, EMPTY_DICT).keys(),
+                    old_values)
```
sebalix: Do not change these lines, this is not necessary.

sebalix (changes requested):
Comment done bellow.
Please avoid to change lines of code not related to the fix, it makes the diff bigger (hard to read, and possibly hard to backport in other branches).

GlobalManager:
The test_pylint applied during the tests, is giving the error, but there is not have to be with auditlog.

GlobalManager:
What we have to

GlobalManager:
If the pylint failed, what can we do?.

sebalix:
The lint failed but it's not related to auditlog, so do nothing about it. If a lint was related to auditlog, just fix it :)

GlobalManager:
That's nice.

auditlog/models/rule.py
```diff
@@ -237,44 +237,26 @@ def _make_read(self):
...
-            result = read.origin(self, *args, **kwargs)
-            # Sometimes the result is not a list but a dictionary
-            # Also, we can not modify the current result as it will break calls
+        @api.multi
+        def read(self, fields=None, load='_classic_read'):
+            result = read.origin(self, fields, load)
             result2 = result
             if not isinstance(result2, list):
                 result2 = [result]
             read_values = dict((d['id'], d) for d in result2)
```
sebalix, line 246: `result`/`result2` trick isn't necessary anymore as the `read` method is only compatible with the new API. Check the piece of code I sent in the other comment.

sebalix (changes requested):

Issue #1260, opened 2018-05-29
auditlog module: field deletion triggers log removal


PieterPaulussen:
I noticed the following cascade deletes [here](https://github.com/OCA/server-tools/blob/8.0/auditlog/models/log.py#L36-L39).

Is it desired that removing a model / field (f.i. during a migration) removes the auditlog lines as well?
What if we keep the auditlog line and use a related field to store the field name (and the log for that matter). Would that not be preferable to losing record in the auditlog?
@sebalix @hbrunn

sebalix:
@PieterPaulussen indeed if you need to preserve the logs even if the corresponding field is removed, you can change the `ondelete` to `set null` on `field_id` (however we should not change the `log_id` IMO). Feel free to open a PR for that, I don't see any problem with that feature.

pedrobaeza:
Isn't that field required?

PieterPaulussen:
@pedrobaeza Indeed, the field is required. If we raise an exception during create/write on the value of the field, we could cover that. 
@sebalix I will keep the cascade on log_id. Thanks for the suggestion. 
PR coming up ;-)

PR #1262, opened 2018-05-29
[FIX] remove ondelete cascade on auditlog_log_line

PieterPaulussen:
Related to https://github.com/OCA/server-tools/issues/1260

### DIFF
--- a/auditlog/migrations/8.0.1.4.0/pre-migration.py
+++ b/auditlog/migrations/8.0.1.4.0/pre-migration.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+# © 2018 Pieter Paulussen <pieter_paulussen@me.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    logger = logging.getLogger(__name__)
+    logger.info("Creating columns: auditlog_log (model_name, model_model) "
+                "and auditlog_log_line (field_name, field_description).")
+    cr.execute("""
+    ALTER TABLE auditlog_log
+    ADD COLUMN IF NOT EXISTS model_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS model_model VARCHAR;
+    ALTER TABLE auditlog_log_line
+    ADD COLUMN IF NOT EXISTS field_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS field_description VARCHAR;
+    """)
+
+    logger.info("Creating indexes on auditlog_log column 'model_id' and "
+                "auditlog_log_line column 'field_id'.")
+    cr.execute("""
+    DO $BLOCK$
+    BEGIN
+      BEGIN
+        CREATE INDEX auditlog_log_model_id_index
+          ON auditlog_log (model_id);
+        EXCEPTION
+        WHEN duplicate_table
+          THEN RAISE NOTICE
+          'index ''auditlog_log_model_id_index'' on auditlog_log already '
+          'exists, skipping';
+      END;
+      BEGIN
+        CREATE INDEX auditlog_log_line_field_id_index
+          ON auditlog_log_line (field_id);
+        EXCEPTION
+        WHEN duplicate_table
+          THEN RAISE NOTICE
+          'index ''auditlog_log_line_field_id_index'' on auditlog_log_line '
+          'already exists, skipping';
+      END;
+    END;
+    $BLOCK$;
+    """)
+
+    logger.info("Preemptive fill auditlog_log columns: 'model_name' and "
+                "'model_model'.")
+    cr.execute("""
+    UPDATE auditlog_log al
+    SET model_name = im.name, model_model = im.model
+    FROM ir_model im
+    WHERE im.id = al.model_id
+    """)
+
+    logger.info("Preemtive fill of auditlog_log_line columns: 'field_name' and"
+                " 'field_description'.")
+    cr.execute("""
+    UPDATE auditlog_log_line al
+    SET field_name = imf.name, field_description = imf.field_description
+    FROM ir_model_fields imf
+    WHERE imf.id = al.field_id
+    """)
+    logger.info("Successfully updated auditlog_log and auditlog_log_line "
+                "tables.")
--- a/auditlog/models/log.py
+++ b/auditlog/models/log.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from openerp import models, fields
+from openerp import api, models, fields
+from openerp.tools.translate import _
+from openerp.exceptions import Warning as UserError
 
 
 class AuditlogLog(models.Model):
@@ -11,7 +13,9 @@ class AuditlogLog(models.Model):
 
     name = fields.Char("Resource Name", size=64)
     model_id = fields.Many2one(
-        'ir.model', string=u"Model")
+        'ir.model', string=u"Model", index=True, on_delete='set null')
+    model_name = fields.Char(string=u'Model Name')
+    model_model = fields.Char(string=u'Technical Model Name')
     res_id = fields.Integer(u"Resource ID")
     user_id = fields.Many2one(
         'res.users', string=u"User")
@@ -28,19 +32,60 @@ class AuditlogLog(models.Model):
          ],
         string=u"Type")
 
+    @api.model
+    def create(self, vals):
+        """ Insert model_name and model_model field values upon creation. """
+        if 'model_id' in vals and vals['model_id']:
+            model = self.env['ir.model'].browse(vals['model_id'])
+            vals.update({'model_name': model.name,
+                         'model_model': model.model})
+        return super(AuditlogLog, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        """ Update model_name and model_model field values to reflect model_id
+        changes. """
+        if 'model_id' in vals and vals['model_id']:
+            model = self.env['ir.model'].browse(vals['model_id'])
+            vals.update({'model_name': model.name,
+                         'model_model': model.model})
+        return super(AuditlogLog, self).write(vals)
+
 
 class AuditlogLogLine(models.Model):
     _name = 'auditlog.log.line'
     _description = "Auditlog - Log details (fields updated)"
 
     field_id = fields.Many2one(
-        'ir.model.fields', ondelete='cascade', string=u"Field", required=True)
+        'ir.model.fields', ondelete='set null', string=u"Field", index=True)
     log_id = fields.Many2one(
         'auditlog.log', string=u"Log", ondelete='cascade', index=True)
     old_value = fields.Text(u"Old Value")
     new_value = fields.Text(u"New Value")
     old_value_text = fields.Text(u"Old value Text")
     new_value_text = fields.Text(u"New value Text")
-    field_name = fields.Char(u"Technical name", related='field_id.name')
-    field_description = fields.Char(
-        u"Description", related='field_id.field_description')
+    field_name = fields.Char(u"Technical name")
+    field_description = fields.Char(u"Description")
+
+    @api.model
+    def create(self, vals):
+        """ Ensure field_id is not empty on creation and store field_name and
+        field_description. """
+        if 'field_id' not in vals or not vals['field_id']:
+            raise UserError(_("No field defined to create line."))
+        field = self.env['ir.model.fields'].browse(vals['field_id'])
+        vals.update({'field_name': field.name,
+                     'field_description': field.field_description})
+        return super(AuditlogLogLine, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        """ Ensure field_id is set during write and update field_name and
+        field_description values. """
+        if 'field_id' in vals:
+            if not vals['field_id']:
+                raise UserError(_("The field 'field_id' cannot be empty."))
+            field = self.env['ir.model.fields'].browse(vals['field_id'])
+            vals.update({'field_name': field.name,
+                         'field_description': field.field_description})
+        return super(AuditlogLogLine, self).write(vals)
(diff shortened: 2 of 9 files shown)


auditlog/models/log.py
```diff
@@ -34,13 +38,31 @@ class AuditlogLogLine(models.Model):
...
         'auditlog.log', string=u"Log", ondelete='cascade', index=True)
     old_value = fields.Text(u"Old Value")
     new_value = fields.Text(u"New Value")
     old_value_text = fields.Text(u"Old value Text")
     new_value_text = fields.Text(u"New value Text")
-    field_name = fields.Char(u"Technical name", related='field_id.name')
+    field_name = fields.Char(
+        string=u"Technical name", related='field_id.name', store=True)
     field_description = fields.Char(
         u"Description", related='field_id.field_description')
```
StefanRijnhart: store=True?

auditlog/views/auditlog_view.xml
```diff
@@ -133,6 +133,7 @@
                             </group>
                             <group colspan="1">
                                 <field name="model_id" readonly="1"/>
+                                <field name="model_name"/>
```
StefanRijnhart: Maybe store model.model as well, and show both model_name and model.model iff model_id is False?

auditlog/models/log.py
```diff
@@ -34,13 +38,31 @@ class AuditlogLogLine(models.Model):
...
+        if not res.field_id or res.field_id is None:
+            raise UserError(_("No field defined to create line."))
+        return res
+
+    @api.multi
+    def write(self, vals):
+        """ Ensure field_id is set during write. """
+        res = super(AuditlogLogLine, self).write(vals)
+        empty_field = self.filtered(lambda x: not x.field_id)
+        if empty_field:
```
StefanRijnhart: I'd prefer
`if 'field_id' in vals and not vals['field_id']:`

StefanRijnhart:
Thanks! Can you add a premigration script to preemptively creates the columns and fills them with an UPDATE...FROM query?

auditlog/views/auditlog_view.xml
StefanRijnhart: Similarly, can you show the field name and description below iff field_id is False?

PieterPaulussen: model_name is a related field to model_id.model. But I will also store the model_id.name in a separate field and update with your request.

auditlog/models/log.py
PieterPaulussen: Isn't a related field non-stored?

PieterPaulussen: Indeed you are correct, my bad. 
```
:param store: whether the field is stored in database (boolean, by
            default ``False`` on computed fields)
```

auditlog/models/log.py
```diff
@@ -44,3 +49,19 @@ class AuditlogLogLine(models.Model):
     field_name = fields.Char(u"Technical name", related='field_id.name')
     field_description = fields.Char(
         u"Description", related='field_id.field_description')
+
+    @api.model
+    def create(self, vals):
+        """ Ensure field_id is not empty on creation. """
+        res = super(AuditlogLogLine, self).create(vals)
+        if not res.field_id or res.field_id is None:
```
StefanRijnhart: I think the first clause covers the second clause already.

auditlog/models/log.py
StefanRijnhart: I think we misunderstood each other. My suggestion was to add store=True to the definition of field_description like you did for field_name, but now I believe field_name has lost its `store` parameter. This would call for migration too of course.

StefanRijnhart:
BTW the migration alone calls for adding an index on model_id and field_id but if you add them now you will want to precreate them before you start filling the values for model_name etc.

auditlog/models/log.py
PieterPaulussen: I always get confused when None is compared equally to False.

StefanRijnhart: `None` and `False` are both falsy but sometimes you need to distinguish `None` from `False` in which case you use `is None`

auditlog/migrations/8.0.1.4.0/pre-migration.py
```diff
@@ -0,0 +1,60 @@
...
+    ALTER TABLE auditlog_log
+    ADD COLUMN IF NOT EXISTS model_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS model_model VARCHAR;
+    ALTER TABLE auditlog_log_line
+    ADD COLUMN IF NOT EXISTS field_description VARCHAR
+    """)
+    logger.info("Creating indexes on auditlog_log column 'model_id' and "
+                "auditlog_log_line column 'field_id'.")
+    cr.execute("""
+    CREATE INDEX auditlog_log_model_id_index ON auditlog_log (model_id);
```
StefanRijnhart: Check if the indexes don't exist already.

auditlog/migrations/8.0.1.4.0/pre-migration.py
```diff
@@ -0,0 +1,60 @@
...
+    SET model_name = (
+      SELECT im.name
+      FROM ir_model im
+      WHERE im.id = al.model_id
+    );
+    UPDATE auditlog_log al
+    SET model_model = (
+      SELECT im.model
+      FROM ir_model im
+      WHERE im.id = al.model_id
```
StefanRijnhart: You can combine the statements above with 
```
UPDATE auditlog_log al
SET model_model = im.model, model_name = im.name
FROM ir_model im WHERE im.id = al.model_id
```

auditlog/migrations/8.0.1.4.0/pre-migration.py
```diff
@@ -0,0 +1,60 @@
...
+    SET field_name = (
+      SELECT imf.name
+      FROM ir_model_fields imf
+      WHERE imf.id = al.field_id
+    );
+    UPDATE auditlog_log_line al
+    SET field_description = (
+      SELECT imf.field_description
+      FROM ir_model_fields imf
+      WHERE imf.id = al.field_id
```
StefanRijnhart: similarly

auditlog/models/log.py
```diff
@@ -11,7 +13,10 @@ class AuditlogLog(models.Model):
 
     name = fields.Char("Resource Name", size=64)
     model_id = fields.Many2one(
-        'ir.model', string=u"Model")
+        'ir.model', string=u"Model", index=True)
+    model_name = fields.Char(u'Model Name', related="model_id.name")
+    model_model = fields.Char(
+        string=u'Technical Model Name', related="model_id.model")
```
StefanRijnhart: add `store=True`

StefanRijnhart (approved):
Thanks! Can you rebase onto 8.0 to fix flake8 from the other modules?

PieterPaulussen:
Rebased on 8.0

auditlog/models/log.py
```diff
@@ -34,13 +40,31 @@ class AuditlogLogLine(models.Model):
     _description = "Auditlog - Log details (fields updated)"
 
     field_id = fields.Many2one(
-        'ir.model.fields', ondelete='cascade', string=u"Field", required=True)
+        'ir.model.fields', ondelete='set null', string=u"Field", required=True,
+        index=True)
```
sebalix: So there is no problem having a `NOT NULL` constraint (required=True) and `ondelete='set null'`? Is it OK for PostgreSQL?

PieterPaulussen: No, thanks for noticing. The required should be dropped (I moved it to check on create and write [here](https://github.com/OCA/server-tools/pull/1262/files#diff-076172f5c3697c2178f79ca37dee1446R60))

sebalix: Ok :) I saw the `create`/`write` overloads too and I did not understand the point while preserving the `required=True`.

sebalix:
Lint fails on the compute method names, but it's not related to your dev: https://travis-ci.org/OCA/server-tools/jobs/387711522#L593
Would you like to fix them please?
Regarding `old-api7-method-defined` Pylint warnings, you can disable them with `# pylint: disable=old-api7-method-defined` where these methods are defined.

PieterPaulussen:
@StefanRijnhart I made some significant changes to this PR. Could you please review this again from the bottom up? Thanks!
@sebalix Updated the function definitions and added pylint ignores. Thanks for the tip!

StefanRijnhart:
Not sure what is wrong with the actual tests in travis, but you can fix flake8
```
auditlog/tests/test_auditlog.py:239:14: E225 missing whitespace around operator
auditlog/tests/test_auditlog.py:275:14: E225 missing whitespace around operator
```

auditlog/migrations/8.0.1.4.0/pre-migration.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+# © 2015 Therp BV <http://therp.nl>
```
StefanRijnhart: You're being too modest ;-)

StefanRijnhart:
I fully agree with your latest changes. Let's see what the tests do after fixing flake8.

PieterPaulussen:
@StefanRijnhart there are only 2 errors I can see that possibly trigger Travis to think they are errors:
```
2018-06-06 08:18:23,914 4559 [1;32m[1;49mINFO[0m openerp_test openerp.modules.module: module sentry: creating or updating database tables
2018-06-06 08:18:23,923 4559 [1;32m[1;49mINFO[0m openerp_test openerp.modules.module: openerp.addons.sentry.tests.test_client running tests.
2018-06-06 08:18:23,923 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client: test_capture_event (openerp.addons.sentry.tests.test_client.TestClientSetup)
2018-06-06 08:18:23,927 4559 [1;33m[1;49mWARNING[0m openerp_test openerp.addons.sentry.tests.test_client: Test event, can be ignored
2018-06-06 08:18:23,929 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client: test_ignore_exceptions (openerp.addons.sentry.tests.test_client.TestClientSetup)
2018-06-06 08:18:23,931 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client.InMemoryClient: Not capturing exception due to filters: <class 'openerp.exceptions.ValidationError'>
File "/home/travis/build/OCA/server-tools/sentry/tests/test_client.py", line 119, in test_ignore_exceptions
raise exceptions.ValidationError(msg)
ValidationError: ('ValidateError', 'Test ValidationError')
2018-06-06 08:18:23,931 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client: test_initialize_raven_sets_dsn (openerp.addons.sentry.tests.test_client.TestClientSetup)
2018-06-06 08:18:23,933 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client: Ran 3 tests in 0.009s
2018-06-06 08:18:23,933 4559 [1;32m[1;49mINFO[0m openerp_test openerp.addons.sentry.tests.test_client: OK
```

StefanRijnhart:
These lines are present in https://travis-ci.org/OCA/server-tools/jobs/388640262 as well. That is a run from the last 8.0 build and it's green.

StefanRijnhart:
Ah, the travis scripts helpfully list the problematic lines ;-)
```
The model test.model has no access rules, consider adding one. E.g. access_test_model,access_test_model,model_test_model,,1,0,0,0
The model test.model has no access rules, consider adding one. E.g. access_test_model,access_test_model,model_test_model,,1,0,0,0
```

StefanRijnhart:
Can you use this? https://github.com/OCA/server-tools/blob/8.0/database_cleanup/tests/test_database_cleanup.py#L46

PieterPaulussen:
@StefanRijnhart I completely rewrote the test so that the test model is removed from the DB. This should solve the ACL warning. I think it's difficult to really simulate the removal of a complete module within a testcase but I believe this should be sufficient to validate the result.

sebalix (approved):

auditlog/tests/test_auditlog.py
```diff
@@ -181,3 +182,101 @@ def test_03_copy_res_partner_logging_old_api(self):
...
+    def tearDown(self):
+        """ Clean up after testcase. """
+        self.auditlog_rule.unsubscribe()
+        self.auditlog_rule.unlink()
+        self.registry.models.pop('x_test.model')
+        self.registry._pure_function_fields.pop('x_test.model')
+
+        # Force reset of environment and clear cursor actions
+        self.env.reset()
+        self.cr.rollback()
```
StefanRijnhart: Thy do you need to perform all these SQL actions? Is that not why this runs in a test case?

PieterPaulussen: I believe during the model creation https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_model.py#L222 the database cursor is commited. During the testcase, I pass MODULE_UNINSTALL_FLAG to the unlink function to avoid a registry reload which in turn triggers a complete rerun of the previous testcases (including this one) in this module and resulting in a duplicate model name constraint error.
My assumption here was that I manually needed to rollback all other cursor actions before I deleted the ir_model record. If you don't delete the ir_model record, the ACL warning pops up because the registry reloads everything from scratch on other test modules.
The testcase works fine without the above two statement, but I think this cleans up the auditlog_log_lines? Am I wrong?

StefanRijnhart: No, I think you are correct in your analysis but your solution looks void to me. A rollback only rolls back to the last commit so I don't think it helps you to circumvent the commits in _auto_init.

Try this before you create the model: https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/modules/loading.py#L140-L141, then reassign the original commit method after the model creation.

PieterPaulussen: Indeed, that is the correct fix to solve the problem. Thanks for the tip!

StefanRijnhart (approved):
Great work, thanks!

sebalix:
@StefanRijnhart can you merge this one?

thomaspaulb (approved):

thomaspaulb:
@pedrobaeza Could you merge please?

StefanRijnhart:
@sebalix @thomaspaulb @PieterPaulussen apologies for the delay

PR #1264, opened 2018-05-29
The methods read, create, write and unlink were going to be executed,

GlobalManager:
Even if they weren't checked in the rule related to the model.

It is fixed with this patch.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -207,8 +207,8 @@ class AuditlogRule(models.Model):
             model_obj = self.env['ir.model']
             model_result = model_obj.search([('model', '=', self._name)])
             auditlog_obj = self.env['auditlog.rule']
-            auditlog_result = auditlog_obj.search([('model_id', '=',
-                model_result.id)])
+            auditlog_result = auditlog_obj.search(
+                [('model_id', '=', model_result.id)])
             if not auditlog_result.log_create:
                 return new_record
             new_values = dict(
@@ -229,8 +229,8 @@ class AuditlogRule(models.Model):
             model_obj = self.env['ir.model']
             model_result = model_obj.search([('model', '=', self._name)])
             auditlog_obj = self.env['auditlog.rule']
-            auditlog_result = auditlog_obj.search([('model_id', '=',
-                model_result.id)])
+            auditlog_result = auditlog_obj.search(
+                [('model_id', '=', model_result.id)])
             if not auditlog_result.log_create:
                 return new_record
             new_values = {new_record.id: vals2}
@@ -252,8 +252,8 @@ class AuditlogRule(models.Model):
             model_obj = self.env['ir.model']
             model_result = model_obj.search([('model', '=', self._name)])
             auditlog_obj = self.env['auditlog.rule']
-            auditlog_result = auditlog_obj.search([('model_id', '=',
-                model_result.id)])
+            auditlog_result = auditlog_obj.search(
+                [('model_id', '=', model_result.id)])
             if not auditlog_result.log_read:
                 return result
             # Sometimes the result is not a list but a dictionary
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Can I ask you how to reproduce this?

GlobalManager:
Set the rule for res.users and subscribe it, in the rules view, without marking any checkbox. Then perform one read of res.users, that can be made reloading the Odoo's webpage. The read will be registered, even if the read checkbox wasn't checked.

StefanRijnhart:
Can't reproduce it like that but it is the case that if you subscribed to a rule with a checkbox on one of the methods and you then uncheck the box the rule for that method is not deactivated. You should try and make it so that when a checkbox is set to False, a patched method is reverted (starting from https://github.com/OCA/server-tools/blob/11.0/auditlog/models/rule.py#L185)

GlobalManager:
The current patch is to ask if the checkbox is set True or False. If it is False, it returns the result of the action performed (read, create, write, unlink) and don't register anything in the log. But if it is True, it registers in the log and returns the result of the action performed.

sebalix:
The bug could be reproduced with a unit test I think, and the one pointed by @StefanRijnhart is there from the beginning. And regarding this bug, I was thinking to fix it by setting the fields readonly when a rule is subscribed, this way we force the user to unsubscribe the rule before editing the checkbox, then subscribe it again.
Cons: records are not audited between the unsubscribe/subscribe actions.
Pros: methods are unpatched/patched accordingly and we avoid boilerplate code, which is cleaner.

GlobalManager:
Readonly on subscribe will give better results, because it can't be edited meanwhile, a rule is being in the middle of an execution. Even though, if the checkbox isn't checked, the logs are still being registered. I guess we can do both, these corrections I'm applying, and what you said, to add security to the records.

sebalix:
@GlobalManager 

> Even though, if the checkbox isn't checked, the logs are still being registered.

Okay so we are not talking about the same bug. You are fixing another one... We need tests if logs are generated even if a checkbox is not enabled, it should not work like that.

StefanRijnhart:
I think you *are* talking about the same bug. @GlobalManager like I said, I cannot reproduce the bug if I follow your instructions of creating a new rule for res.users and leaving the read checkbox unchecked all the time. Read operations are not logged in that case.

StefanRijnhart:
So yeah, a test reproducing the problem would be good.

GlobalManager:
You have to create the res.users record and log in as that user.

GlobalManager:
In that way you can do that use case.

StefanRijnhart:
Sorry, I cannot reproduce. I execute the following acitons:

On a newly created database, I install auditlog. I enter debug mode. I configure an audit rule for Users, uncheck all log options and subscribe.

image

I log in again (again as admin). I go to my user account and click View logs. No logs are created.

image

StefanRijnhart:
That is on Odoo 11.0 commit a30d9aa560244cf7c and server-tools commit 5279365c745

GlobalManager:
Right. It's working, it was tested again and worked. But it still registering a False value of the name of the resource. So I'm uploading the change to add these two sentences, to the line 347, in auditing/model/rule.py:

            if not res_name or res_name == "False":
                continue

StefanRijnhart:
@GlobalManager I understand that you agree that the bug cannot be reproduced in the way you described. You are not interested in fixing the bug that @sebalix and me agree exists, when a checkbox is deactivated on a rule after subscribing. And now you want to reuse this PR to stop creating logs for records that don't have a name (yet). I'm afraid I don't agree with your suggestion. It can be perfectly acceptable for records of certain types not to have a name (especially at creation time). 

Also, I don't agree with reusing PRs in this way. Please open a new issue or PR if you insist on this new change so that it can be freshly triaged.

GlobalManager:
What I said is that I tried in the same way and I couldn't reproduce the error neither.

PR #1265, opened 2018-05-30
[ADD] Migration scripts for mass_editing

bealdav:
Put migrations in 10.0 to allow migrate from 8.0 or older to 10.0 directly

ping @mreficent 
@bguillot 
@JayVora-SerpentCS

### DIFF
--- a/mass_editing/migrations/9.0.1.0.0/openupgrade_analysis_work.txt
+++ b/mass_editing/migrations/9.0.1.0.0/openupgrade_analysis_work.txt
@@ -1,0 +1,14 @@
+---Fields in module 'mass_editing'---
+mass_editing / mass.object              / model_ids (many2many)         : DEL relation: ir.model
+mass_editing / mass.object              / model_list (char)             : NEW
+# NOTHING TO DO
+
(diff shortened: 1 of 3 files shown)


bguillot (approved):

mass_editing/migrations/9.0.1.0.0/openupgrade_analysis_work.txt
```diff
@@ -0,0 +1,14 @@
+---Fields in module 'mass_editing'---
+mass_editing / mass.object              / model_ids (many2many)         : DEL relation: ir.model
+mass_editing / mass.object              / model_list (char)             : NEW
+# NOTHING TO DO
```
legalsylvain, line 4: Hi, 
I'm not expert with that modules, but it seems that model_list and model_ids have the same meaning. I guess that postmigration script should fill model_list.
Don't you think ?

legalsylvain (changes requested):
Thanks for this great improvment.
Code review / no test.
One question. Otherwise. LGTM. Thanks

PR #1267, opened 2018-05-30
auditlog is printing False in the Name of the log created

GlobalManager:
auditlog is printing False in the Name of the log created.

To recreate the log failed, just activate res.users checkbox and subscribe the rule.


sebalix:
@GlobalManager can you give some more info? On which data model did you experience the issue?

PR #1279, opened 2018-06-06
[ADD] nagios_passive_check

etobella:
Nagios Passive check
====================

This module allows to send data to nagios through passive checks.
Initially, it will send cpu, ram and connected users, but it could be extended.

**Configuration**

It requires to have a nagios server in order to test it.
- Open development mode
- Access `Automation / Nagios`
- Create a nagios server connection
- Nagios connection can be tested.
- Every five minutes, data will be sent to nagios


sebalix:
Hi,

We have developed this one which has almost the same purpose: https://github.com/OCA/server-tools/pull/312
It is used to send data to any NSCA daemon (Nagios, Icinga, Shinken...), but it must be updated to use nsca-ng. It does nothing by default, it's up to the developer to code the checks (to check mail queue, job queue... whatever).
Pros: not limited to Nagios
Cons: use a dedicated protocol (and port), it's not HTTP

etobella:
You are right, but currently, no nsca python plugins works with Python3. For that reason I used nagios api. I could try to use send_nsca directly.

sebalix:
Yes there is no NSCA Python module well maintained, that's why I chose to use the `send_nsca` binary directly.
But it was just to mention my PR, if it does not fit your needs feel free to continue your work with yours!

etobella:
Following @sebalix comments, I have made a migration of #312 PR in #1280, so I close this PR because it would be duplicating functionality.

PR #1280, opened 2018-06-07
[11.0] nsca client

etobella:
NSCA Client
===========

This is a technical module to send passive alerts to your favorite NSCA daemon
(Nagios, Shinken...).
This module is based on the Odoo cron system and requires a NSCA client
installed on the system to satisfy the ``/usr/sbin/send_nsca`` command.

Installation
============

To use this module, you need to install a NSCA client.

On Debian/Ubuntu::

    $ sudo apt-get install nsca-client

Configuration
=============

* Configure your server and a passive service in your monitoring tool
  (e.g service ``Odoo Mail Queue`` on host ``MY-SERVER``).
* Declare your NSCA server in the menu Configuration / Technical / NSCA Client / Servers
* Create NSCA checks in the menu Configuration / Technical / NSCA Client / Checks
* Code the methods which will be called by the NSCA checks.

Such methods must return a tuple ``(RC, MESSAGE, PERFORMANCE_DATA)`` where ``RC`` is an integer,
``MESSAGE`` a unicode string AND ``PERFOMANCE_DATA`` is a dictionary.
``RC`` values and the corresponding status are:

- 0: OK
- 1: WARNING
- 2: CRITICAL
- 3: UNKNOWN

each element of ``PERFORMANCE_DATA`` will be a dictionary with the following values:

- value: value of the data (required)
- max: Max value on the chart
- min: Minimum value on the chart
- warn: Warning value on the chart
- crit: Critical value on the chart
- uom: Unit of Mesure on the chart (s - Seconds, % - Percentage, B - Bytes, c - Continuous)


etobella:
It is the migration of #312. Now it allows to send performance_date on the call and I have added a default function in order to send server status.

nsca_client/README.rst
```diff
@@ -40,14 +40,24 @@ To configure this module, you need to:
 
 * Code the methods which will be called by the NSCA checks.
 
-Such methods must return a tuple ``(RC, MESSAGE)`` where ``RC`` is an integer,
-and ``MESSAGE`` a unicode string. ``RC`` values and the corresponding status are:
+Such methods must return a tuple ``(RC, MESSAGE, PERFORMANCE_DATA)`` where ``RC`` is an integer,
+``MESSAGE`` a unicode string AND ``PERFOMANCE_DATA`` is a dictionary.
```
sebalix, line 44: Typo: `PERFORMANCE_DATA` (a `R` is missing)

nsca_client/README.rst
```diff
@@ -59,32 +69,25 @@ E.g:
         def nsca_check_mails(self):
             mails = self.search([('state', '=', 'exception')])
             if mails:
-                return (1, u"%s mails not sent" % len(mails))
-            return (0, u"OK")
+                return (1, u"%s mails not sent" % len(mails), {
+                  'exceptions': {'value': mails}})
+            return (0, u"OK", {'exceptions': {'value': mails}})
```
sebalix: The README does not mention the `exceptions` key, is it required? If so, what are the others keys?

nsca_client/__manifest__.py
```diff
@@ -1,24 +1,23 @@
...
 # © 2015 ABF OSIELL <http://osiell.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "NSCA Client",
     "summary": "Send passive alerts to monitor your Odoo application.",
-    "version": "8.0.1.0.0",
+    "version": "11.0.1.0.0",
     "category": "Tools",
-    "website": "http://osiell.com/",
+    "website": "http://github.com/oca/server-tools",
```
sebalix: OCA (don't know if there is an aggregation somewhere on the `website` key, but it could matter I think).

nsca_client/__manifest__.py
```diff
@@ -1,24 +1,23 @@
...
     "installable": True,
     "data": [
         "security/ir.model.access.csv",
-        "data/nsca_server.xml",
         "views/nsca_menu.xml",
         "views/nsca_check.xml",
         "views/nsca_server.xml",
     ],
     "demo": [
+        "data/nsca_server.xml",
```
sebalix: If the default NSCA server should be a demo data (maybe), then it should be moved under the `demo` directory. In this case, maybe renamed the XML-ID to `nsca_server_demo`.

nsca_client/README.rst
sebalix: After reading the code, I saw that performance data is not mandatory (which is a good thing), but the README suggests the opposite. Can you reformulate the description?

sebalix (changes requested):

sebalix:
Nice work! Thank you, you even add tests.
I wanted to migrate to [nsca-ng](http://www.nsca-ng.org/) as the old `nsca-client` is not secure enough. On their website they claim that the send_nsca binary is 100% compatible.

nsca_client/README.rst
etobella: It is the label of the performance data sent. I will upgrade the README.

sebalix (approved):

HviorForgeFlow (approved):
Functional Review, useful module working like a charm :+1:

nsca_client/README.rst
etobella: https://github.com/OCA/server-tools/pull/1280/files#diff-0ea33317e255ff360217de4784a9bffdR52

PR #1283, opened 2018-06-11
10 migr custom attributes

bguillot:
Hi, 

Migration in v10 of the module base_custom_attributes.

Thanks for the reviews.


pedrobaeza:
- The README doesn't reflect the module goal
- There can't be plurals on module name
- Why migrating this one having base_custom_info?

sebastienbeau:
Hi Pedro, the aim of the module is not the same as product_variant_configurator. The aim is to introduce the magento concept of "attribute_set". We have split this concept in two module, a generic one and an implementation for the product.

The concept of "attribute set" is something really important for huge catalog, the idea is to create an attribute set for a typology of product (fridge, camera, car...) and for each of this typology you define a set of field that are optionnal or mandatory. We have a customer who have around 1600 custom fields on the product object, this is not possible to show all of this field on the product form, he really need to see only the required field for the typology of the product.

If you want, I can show you a demo ;)

pedrobaeza:
Already talked about this in person and the more similar one is base_custom_info.

sebastienbeau:
Hi after thinking twice and testing base_custom_info, I think this both module look similare but the behavior is different. For managing e-commerce attribute like Magento I prefer the way we did it in base_custom_attribute (mostly inspired from Magento). As we plan to improve a lot that module as we are in a process of migrating our customer from Magento to Shopinvader, I think it simplier for now to not merge this module in OCA and maintain it in our repo for the 10 version.
For 12 version we will open an issue to move back that module to OCA, I thinks that both module base_custom_info and base_custom_attribute, have a place in OCA and do not served the same need.
The official repo is now here : https://github.com/akretion/ak-odoo-incubator

PR #1284, opened 2018-06-11
[MIG] base_search_fuzzy: Migration to 11.0

cubells:
cc @Tecnativa


base_search_fuzzy/tests/test_query_generation.py
```diff
@@ -37,8 +36,8 @@ def test_fuzzy_where_generation(self):
             where_clause_params)
         self.assertEqual(
             complete_where,
-            'SELECT FROM "res_partner" WHERE '
-            '("res_partner"."name" % \'test\')')
+            b'SELECT FROM "res_partner" WHERE '
```
pedrobaeza: I think the `b` (binary) is not needed

base_search_fuzzy/models/ir_model.py
```diff
@@ -43,7 +43,7 @@ def decorate_leaf_to_sql(self, eleaf):
             if left in model._fields:
                 params = str(right)
 
-            if isinstance(params, basestring):
+            if isinstance(params, pycompat.string_types):
```
pedrobaeza: Is this something made by `2to3`? I think `isinstance(params, str)` is better

pedrobaeza (changes requested):

base_search_fuzzy/models/ir_model.py
yajo: Indeed, there's no need to support py2 for v11.

yajo (approved):

base_search_fuzzy/tests/test_query_generation.py
cubells: It's necessary because complet_where is binary.

pedrobaeza: Uhm, I don't think so. @Yajo what do you think?

yajo: I have no reasons to believe he's wrong, and Travis was green, so it's OK for me.

pedrobaeza (approved):

pedrobaeza:
Please squash before merging

cubells:
@pedrobaeza done!

PR #1285, opened 2018-06-12
11.0 mig base multi image

qtheuret:


base_multi_image/views/image_view.xml
```diff
@@ -0,0 +1,151 @@
...
+
+<record id="image_kanban_view" model="ir.ui.view">
+    <field name="name">Product multi image kanban</field>
+    <field name="model">base_multi_image.image</field>
+    <field name="arch" type="xml">
+        <kanban string="Product Images">
+            <field name="name"/>
+            <field name="storage"/>
+	    <field name="sequence"/>
+	    <field name="id" />
```
yajo: Why do you add this?

yajo (approved):
One minor detail, and also it would be great if you can squash commit history by logical author/changeset. But not a requirement :wink:

base_multi_image/views/image_view.xml
qtheuret: In the v9 of the module, there is no "id" field in the view, but in v11, the kanban-box widget requires this field to avoid a JS error when displaying product form with 'product_multi_image' module.

yajo: Thanks, I didn't know. However, if that's the case, shouldn't it be invisible? ID tend to be not much useful for users.

BTW, wrong indentation here.

pedrobaeza:
Travis is red. Meanwhile, it has appeared #1286 for handling correctly the proper uninstallation of the module. Can you please add it here?

base_multi_image/models/image.py
```diff
@@ -0,0 +1,239 @@
...
+                    return base64.b64encode(f.read())
+            except:
+                _logger.error("URL %s cannot be fetched", url,
+                              exc_info=True)
+
+        return False
+
+    @api.multi
+    @api.depends('image_main')
+    def _get_image_sizes(self):
```
jesusVMayor, line 178:
```suggestion
    def _compute_image_sizes(self):
```

base_multi_image/models/image.py
```diff
@@ -0,0 +1,239 @@
...
+        """Get a reference field based on the split model and id fields."""
+        for s in self:
+            s.owner_ref_id = "{0.owner_model},{0.owner_id}".format(s)
+
+    @api.multi
+    @api.depends('storage', 'path', 'file_db_store', 'url')
+    def _compute_image(self):
+        """Get image data from the right storage type."""
+        for s in self:
+            s.image_main = getattr(s, "_get_image_from_%s" % s.storage)()
```
jesusVMayor, line 125: You changed the name of the function _get_image_from_filestore to _compute_image_from_filestore so don't work in that case.

base_multi_image/models/image.py
```diff
@@ -0,0 +1,239 @@
...
+             "image exceeds one of those sizes. Use this field in form views "
+             "or kanban views.")
+    image_small = fields.Binary(
+        "Small-sized image",
+        compute="_compute_image_sizes",
+        help="Small-sized image. It is automatically resized as a 64 x 64 px "
+             "image, with aspect ratio preserved. Use this field anywhere a "
+             "small image is required.")
+    comments = fields.Text(
+        'Comments',
```
jesusVMayor, line 100: unnecesary string

base_multi_image/models/image.py
```diff
@@ -0,0 +1,239 @@
...
+            ('filestore', 'Filestore'),
+        ],
+        required=True,
+    )
+    name = fields.Char(
+        string='Image title',
+        translate=True,
+    )
+    filename = fields.Char(
+        string='Filename',
```
jesusVMayor, line 60: unnecesary string

jesusVMayor (changes requested):

PR #1286, opened 2018-06-13
[10.0] [IMP] base_multi_image: improve uninstall hook

FFernandez-PlanetaTIC:
Hi,

Improve uninstall hook to move images from multi to single mode.

Thanks!

### DIFF
--- a/base_multi_image/hooks.py
+++ b/base_multi_image/hooks.py
@@ -63,8 +63,8 @@ def pre_init_hook_for_submodules(cr, model, field):
         )
 
 
-def uninstall_hook_for_submodules(cr, registry, model):
-    """Remove multi-images for a given model.
+def uninstall_hook_for_submodules(cr, registry, model, field=None, field_medium=None, field_small=None):
+    """Moves images from multi to single mode and remove multi-images for a given model.
 
     :param odoo.sql_db.Cursor cr:
         Database cursor.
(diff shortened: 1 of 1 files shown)


base_multi_image/hooks.py
```diff
@@ -75,11 +75,73 @@ def uninstall_hook_for_submodules(cr, registry, model):
...
+        if images and (field or field_medium or field_small):
+            main_images = {}
+            for image in images:
+                if image.owner_id not in main_images:
+                    main_images[image.owner_id] = image
+            main_images = main_images.values()
+            Model = env[model]
+            Field = field and Model._fields[field]
+            FieldMedium = field_medium and Model._fields[field_medium]
+            FieldSmall = field_small and Model._fields[field_small]
```
yajo, line 107: one of these 3 can fail with `KeyError` easily

yajo (changes requested):

yajo:
Wasn't it supposed that this addon lets the original image in place, stored in database?

base_multi_image/hooks.py
FFernandez-PlanetaTIC: Hi @Yajo , if a KeyError is raised it means that the uninstall hook hasn't been used correctly by the module that depends on it. It would be a programming error.

FFernandez-PlanetaTIC:
@Yajo , no, it moves the main of multiple images to the original field, stored in a database column or as an attachment depending of the field declaration.

base_multi_image/hooks.py
yajo: In such case, the arguments shouldn't be optional

FFernandez-PlanetaTIC: If the new parameters would be required then it will break existing modules using the hook.

They should be optional because the introduced behavior isn't the desired in some cases as product variants in OCA/product-attribute#359

FFernandez-PlanetaTIC:
Hi @Yajo ,

Is there something more to change? Could you change you review?

Thanks!

yajo (approved):
Yikes, the code is very hard to read :confused: 

However it seems to do the trick.

pedrobaeza (approved):
Indeed, this is very convoluted, but it's what Odoo forces us to do for preserving this. Thanks for the hard work

PR #1288, opened 2018-06-14
Remove *.pyc and *.pyo.

mvaled:


Garamotte (approved):

yvaucher (approved):

yajo:
Sorry, these files are used specifically for this addons' tests, because they are configured to skip certain file extensions, and then they are included so we can check they are correctly skipped.

It's an exception to the good rule of not including them. 

Thanks anyway :wink:

PR #1289, opened 2018-06-18
[10.0][IMP] users_ldap_populate. Log failing filter if populate error...

NL66278:
It can be difficult to see what the failing filter (if any) is if any error occurs during ldap population. 

This PR adds the failing filter to the server log, then reraises the original exception.

### DIFF
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
-# © 2012 Therp BV (<http://therp.nl>)
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+# Copyright 2012 Therp BV (<https://therp.nl>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/gpl.html).
 
 import re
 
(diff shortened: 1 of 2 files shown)


users_ldap_populate/models/users_ldap.py
```diff
@@ -109,9 +109,14 @@ def get_ldap_entry_dicts(self, conf, user_name='*'):
...
         conn.simple_bind_s(conf['ldap_binddn'] or '',
                            conf['ldap_password'] or '')
-        results = conn.search_st(conf['ldap_base'], ldap.SCOPE_SUBTREE,
-                                 ldap_filter.encode('utf8'), None,
-                                 timeout=60)
+        try:
+            results = conn.search_st(
+                conf['ldap_base'], ldap.SCOPE_SUBTREE,
+                ldap_filter.encode('utf8'), None, timeout=60)
+        except Exception:
```
sebalix, line 132: `except:` alone should work. Or is it required for the logger?

sebalix (approved):
One remark/question, but :+1:

users_ldap_populate/models/users_ldap.py
NL66278: @sebalix A bare except results in a pylint warning. So not required functionally but by code conventions.

sebalix: @NL66278 ok, thanks!

gurneyalex (approved):

NL66278:
Closed in favour of https://github.com/OCA/server-tools/pull/1565/commits that contains all changes in this PR, but also makes it really safe to delete unknown users, when there are many users not from ldap.

PR #1291, opened 2018-06-18
[ADD] remote_base

etobella:
Remote Base
===========

This module allows to store all the connected remotes (external ip addresses) to odoo.
It should be used with other modules in order to check remote's configurations.

Usage
=====

When installed, all remotes will be stored by `hostname` on `res.remote`.
They can be viewed on `Settings / Users & Companies / Remotes`.
The last Ip of the remote will be stored.

It could be interesting to integrate this module with [report-print-send,](https://github.com/OCA/report-print-send) because we could configure the printers by remote, not by user.

### DIFF
--- a/remote_base/models/res_users.py
+++ b/remote_base/models/res_users.py
@@ -1,0 +1,53 @@
+# Copyright 2018 Creu Blanca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from threading import current_thread
+from odoo import api, models, SUPERUSER_ID
+from odoo.exceptions import AccessDenied
+from odoo.service import wsgi_server
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    # HACK https://github.com/odoo/odoo/issues/24183
+    # TODO Remove in v12, and use normal odoo.http.request to get details
+    @api.model_cr
+    def _register_hook(self):
+        """🐒-patch XML-RPC controller to know remote address."""
+        original_fn = wsgi_server.application_unproxied
+
+        def _patch(environ, start_response):
+            current_thread().environ = environ
+            return original_fn(environ, start_response)
+
+        wsgi_server.application_unproxied = _patch
+
+    @classmethod
+    def _auth_check_remote(cls, login, method):
+        """Force a method to raise an AccessDenied on falsey return."""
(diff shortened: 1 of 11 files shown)


JordiBForgeFlow (approved):
:+1: Very nice!!

remote_base/models/res_users.py
```diff
@@ -0,0 +1,53 @@
...
+from threading import current_thread
+from odoo import api, models, SUPERUSER_ID
+from odoo.exceptions import AccessDenied
+from odoo.service import wsgi_server
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    # HACK https://github.com/odoo/odoo/issues/24183
```
AaronHForgeFlow, line 13: Odoo will not fix that so I guess the hack will remain.

AaronHForgeFlow (approved):
:+1: functional test + human readable code review :smile:

remote_base/models/res_users.py
```diff
@@ -0,0 +1,53 @@
...
+
+        wsgi_server.application_unproxied = _patch
+
+    @classmethod
+    def _auth_check_remote(cls, login, method):
+        """Force a method to raise an AccessDenied on falsey return."""
+        with cls.pool.cursor() as cr:
+            env = api.Environment(cr, SUPERUSER_ID, {})
+            remote = env["res.users"].remote
+            remote.ensure_one()
```
tarteo, line 32: Is this necessary? The variable is not used anywhere. 
I don't really understand the intention of this file. Why do you need to override the normal login method? Can you please explain? Thank you :+1:

tarteo (changes requested):

remote_base/models/res_users.py
etobella: This way, the remote is generated on login (even on failures). These way, we can configure them.
Moreover, remote may not be used in this modules, but it is interesting on printing. remote-print-send is configured by user, but some companies need it by computer (remote).

tarteo: Alright I understand now, thanks!

tarteo (approved):

pedrobaeza:
Please rename this to `base_remote` now this is just merged, as the nomenclature is not correct.

etobella:
Ok, I will make a fast PR

PR #1292, opened 2018-06-20
[FIX] module name from remote_base to base_remote

etobella:
As commented on the closure of the PR #1291, the name should be changed.


pedrobaeza:
Thanks for the quick reaction. Fast-tracking this.

PR #1293, opened 2018-06-20
[FIX] missing super declaration

etobella:
The super declaration is missing and it gives problems with other modules.

### DIFF
--- a/base_remote/models/res_users.py
+++ b/base_remote/models/res_users.py
@@ -15,6 +15,7 @@ class ResUsers(models.Model):
     @api.model_cr
     def _register_hook(self):
         """🐒-patch XML-RPC controller to know remote address."""
+        super()._register_hook()
         original_fn = wsgi_server.application_unproxied
 
         def _patch(environ, start_response):


JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:

tafaRU (approved):

moylop260 (approved):

base_remote/models/res_users.py
```diff
@@ -15,6 +15,7 @@ class ResUsers(models.Model):
     @api.model_cr
     def _register_hook(self):
         """🐒-patch XML-RPC controller to know remote address."""
+        super()._register_hook()
         original_fn = wsgi_server.application_unproxied
 
         def _patch(environ, start_response):
```
pedrobaeza, line 21: Why don't you use the safest `_patch_method` that is bundled with Odoo: https://github.com/odoo/odoo/blob/ab85e9594651fd33ffd2747f4d3c5b360e7c7922/odoo/models.py#L4255?

etobella: I reused the code from auth_brute_force from OCA/server-auth in [res_users](https://github.com/OCA/server-auth/blob/11.0/auth_brute_force/models/res_users.py#L20-L29)

pedrobaeza: Well, that one also doesn't use the right method also :wink:

JordiBForgeFlow: @pedrobaeza as far as I understand the '_patch_method' is to be used for another purpose that this one. The method you propose is used to monkey_patch a specific method of a model.

In this case the method '_register_hook' is called after the registry is built on a given model, and allows to change the env, so that we can get to know the remote address of the request: https://github.com/etobella/server-tools/blob/3cf905989c58a9c9b01c34b688843daa0bbe675a/base_remote/models/base.py#L14

As far as I understand we can use '_patch_method' as a better alternative to patches like https://github.com/OCA/sale-workflow/blob/11.0/sale_order_action_invoice_create_hook/hooks.py#L12

JordiBForgeFlow: I suggest that we merge this PR now, because it fixes a problem. And we can discuss an improvement in 'auth_brute_force' and here.

pedrobaeza:
Ok, I'm out right now and can't expand my comment, but it's overriding base model.  Merging this.

PR #1294, opened 2018-06-20
[ADD] [base_import_odoo] don't write falsy values on required fields with a default value

hbrunn:


base_import_odoo/models/import_odoo_database.py
```diff
@@ -162,6 +164,8 @@ def _run_import(self, commit=True, commit_threshold=100):
                     self.env.cr.commit()
                     raise
                 done[model._name] += len(ids)
+                # this write will trigger a recomputation of imported records
+                self.env.invalidate_all()
                 self.write({'status_data': dict(self.status_data, done=done)})
```
thomaspaulb, line 169: @hbrunn I've encountered a situation where this line causes a `MemoryError`, where the recomputation of a field (completeness grade) of "chunk" number of records (employees) at once references several long binary fields (employee document attachments), which calls [this binary sanitation function](https://github.com/OCA/OCB/blob/8.0/openerp/osv/fields.py#L1090) which loads the full attachment record in memory and after doing that several times, runs out of memory.

Is the correct way of tackling that:

1. Lowering the chunk size
2. Recomputing/writing records one by one
3. Making our field an [xmlrpclib.Binary](https://github.com/OCA/OCB/blob/8.0/openerp/osv/fields.py#L1077) so that the sanitation does not trigger
4. Changing the computation of our 'completeness' field to not query the contents of very big fields but just whether it is empty or non-empty

On consideration, I think the latter one is good to do anyways, but wondering if something should change here as well.

hbrunn: 1. will be detrimental if you have to lower it a lot, as the whole norecompute/invaidate_all dance is a performance improvement (which is mainly caused by Odoo v8 doing a lot of stuff double and thrice and X-times when computing fields, this is fixed in later versions)
2. will work, but slowly
3. nonononono
4. yes, much better

PR #1047, opened 2017-11-10
[MIG][10.0] sql_export

AaronHForgeFlow:
Tasks done:
- Bump code to new version. Minor changes.

No extra features added/removed from v9

PR #1301, opened 2018-06-25
[9.0][FIX] sql_request_abstract: mogrify doesn't allow dicts

MiquelRForgeFlow:
A fix cheery-picked from https://github.com/OCA/server-tools/pull/1047/commits/733364656d4b1f74db2cdce15bcc37aa43e6432a (https://github.com/OCA/server-tools/pull/1047). `Mogrify` allows a _list_ as parameter but not a _dict_, and a list is not useful here. Thus, `mogrify` is substituted by something more suited.

### DIFF
--- a/sql_request_abstract/models/sql_request_mixin.py
+++ b/sql_request_abstract/models/sql_request_mixin.py
@@ -144,8 +144,10 @@ class SQLRequestMixin(models.AbstractModel):
         if mode in ('view', 'materialized_view'):
             rollback = False
 
-        params = params and params or {}
-        query = self.env.cr.mogrify(self.query, params).decode('utf-8')
+        params = params or {}
+        # pylint: disable=sql-injection
+        query = self.query % params
+        query = query.decode('utf-8')
 
         if mode in ('fetchone', 'fetchall'):
             pass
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
:+1: It works now with parameters

sql_request_abstract/models/sql_request_mixin.py
```diff
@@ -144,8 +144,10 @@ def _execute_sql_request(
         if mode in ('view', 'materialized_view'):
             rollback = False
 
-        params = params and params or {}
-        query = self.env.cr.mogrify(self.query, params).decode('utf-8')
+        params = params or {}
+        # pylint: disable=E8103
```
moylop260: Use the name of the lint instead of the number.

MiquelRForgeFlow: and what is the name?

moylop260: sql-injection
https://github.com/OCA/pylint-odoo/blob/master/README.rst

JordiBForgeFlow (approved):

PR #1305, opened 2018-06-28
[IMP] date_range: new README structure

lmignon:
once approved and merged I'll forward port the changes to https://github.com/OCA/server-ux/tree/11.0/date_range


lmignon:
@mreficent Can you check your email into the contributors list?

yajo (approved):

date_range/readme/INSTALL.rst
```diff
@@ -0,0 +1,9 @@
+
```
pedrobaeza: I would remove this file as there are a lot of install methods and this doesn't make sense on apps.odoo.com. This section is for specifying pre-requirements or possible computation times on install.

lmignon:
@sbidoul A requirement to get 'MIS builder' mature  :smirk:

date_range/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Laurent Mignon <laurent.mignon@acsone.eu>
+* Alexis de Lattre <alexis.delattre@akretion.com>
+* Miquel Raïch <miquel.raïch@eficent.com>
```
pedrobaeza: Incorrect e-mail. It should be `miquel.raich@eficent.com`

pedrobaeza (changes requested):

date_range/readme/INSTALL.rst
sbidoul: I agree, we must prepare a generic page somewhere explaining installation methods, there is no need to repeat what should be common knowledge is each addon.

lmignon: @pedrobaeza I propose the Odoo and the Pythonic way to install this addon.... For sure there are other ways to install this addon but I d'ont see why I should remove this section.

sbidoul (approved):

date_range/readme/CONTRIBUTORS.rst
lmignon: done and rebased

date_range/readme/INSTALL.rst
pedrobaeza: I prefer the link to the possible installation methods, as said by @sbidoul, and it can be added by the generator. We already have https://odoo-community.org/blog/our-blog-1/post/how-to-install-oca-modules-79

lmignon: OK makes sense. 
Done

pedrobaeza (approved):

MiquelRForgeFlow:
@sbidoul my email is correct

pedrobaeza:
@mreficent it is now, but not before :wink:

PR #1312, opened 2018-07-14
[FIX] doesn't try to replace params if they are no args

legalsylvain:
In some cases, the query contains the char ```%```. (with ilike request for exemple).
this will fail, when trying to replace with params.
This trivial patch fixes this bug if they are no params. (if not, query should be escaped).

Sample

```
SELECT
    name as x_name,
    case
        when author ilike '%OpenERP SA%' THEN 'Odoo SA'
        when author ilike '%Odoo Community Association (OCA)%' THEN 'OCA'
       else 'Undefined Author' END as x_author_type
FROM ir_module_module
```


HviorForgeFlow (approved):
LGTM :+1

legalsylvain:
fastracking this fix.

PR #1314, opened 2018-07-16
[WIP][ADD] letsencrypt_transip_nl module

daramousk:
@hbrunn This supplements `letsencrypt`


letsencrypt_transip_nl/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+    <record id="transip" model="ir.config_parameter">
```
hbrunn, line 3: this will conflict with other dns providers. just don't set anything there, the user has to configure this anyways

letsencrypt_transip_nl/models/base_config_settings.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class BaseConfigSettings(models.TransientModel):
+    _inherit = 'base.config.settings'
+
+    letsencrypt_dns_provider = fields.Selection(
+        selection_add=[('transip', 'TransIP')],
+    )
+
+    def letsencrypt_transip_support(self, challenge, domain):
```
hbrunn, line 25: I still don't understand why this is a function of the config wizard and not of the letsencrypt abstract model

letsencrypt_transip_nl/README.rst
```diff
@@ -0,0 +1,80 @@
...
+Follow the usuall installation procedure for a module in your Odoo version.
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+Go to Settings -> General Settings and fill the required fields under
+LetsEncrypt.
```
hbrunn, line 30: we do expect people to know what they're doing, but here i think you can add a few words on what to fill in and where to acquire the values

hbrunn (changes requested):
thanks, but this should go into https://github.com/OCA/l10n-netherlands

PR #1316, opened 2018-07-17
[IMP] purge uninstalled unavailable modules in bulk

hbrunn:
without this, you'll have to wait unnecessarily long for the wizard to start in case there are a lot of uninstallable modules to purge, because purging causes a full registry reload, which costs a lot of time. Doing it in bulk will cause only one registry reload for all modules.

### DIFF
--- a/database_cleanup/models/purge_modules.py
+++ b/database_cleanup/models/purge_modules.py
@@ -63,15 +63,18 @@ class CleanupPurgeWizardModule(models.TransientModel):
     @api.model
     def find(self):
         res = []
+        purge_lines = self.env['cleanup.purge.line.module']
         for module in self.env['ir.module.module'].search([]):
             if get_module_path(module.name):
                 continue
             if module.state == 'uninstalled':
-                self.env['cleanup.purge.line.module'].create({
+                purge_lines += self.env['cleanup.purge.line.module'].create({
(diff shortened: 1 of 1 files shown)


database_cleanup/models/purge_modules.py
```diff
@@ -63,16 +63,19 @@ class CleanupPurgeWizardModule(models.TransientModel):
     @api.model
     def find(self):
         res = []
+        purge_lines = self.env['cleanup.purge.line.module']
```
sebalix, line 66: Is it worth adding `.browse()` for clarity (empty recordset)?
:+1: I have experienced slow processes when cleaning databases, maybe it was the cause.

sebalix (approved):

database_cleanup/models/purge_modules.py
hbrunn: I used to add the `.browse([])`, but meanwhile I see this used as idiom for an empty recordset very often, so if you don't insist, I'd like to keep it this way

sebalix: I don't, it was a coding-style question!

CasVissers-360ERP (approved):
LGTM functional test

pedrobaeza (approved):

PR #1326, opened 2018-07-26
[8.0][FIX] dbfilter_from_header: prevent autoloading

nilshamerlinck:


sbidoul (approved):

RoelAdriaans (changes requested):
No change of the description to mention that proxy_mode should be true?

nilshamerlinck:
@RoelAdriaans I've just changed the description to take into account your suggestion. Thanks!

RoelAdriaans (approved):

PR #1330, opened 2018-07-27
[10.0][FIX] base_exception: add active_model in returned action

Cedric-Pigeon:
The active modle is not propagated at the same time than active_id and active_ids.
In some particular cases it causes some troubles.


lmignon (approved):
Code tested

bealdav (approved):

pedrobaeza (approved):

PR #1333, opened 2018-08-01
auto_backup: allow to change the format of backup

foutoucour:
Add the option to do the backup without the filestore.
We found ourselves the need for hourly backups.


auto_backup/models/db_backup.py
```diff
@@ -151,21 +158,23 @@ def action_backup(self):
...
                 successful |= rec
 
         # Ensure a local backup exists if we are going to write it remotely
         sftp = self.filtered(lambda r: r.method == "sftp")
         if sftp:
-            if backup:
-                cached = open(backup)
-            else:
-                cached = db.dump_db(self.env.cr.dbname, None)
+            for rec in sftp:
```
ddufresne, line 174: @foutoucour the if statement can be removed since you loop on the for.

ddufresne (changes requested):
Minor issue

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: '3'
```
astirpe: Could you remove file `docker-compose.yml` from this PR?

auto_backup/models/db_backup.py
```diff
@@ -83,6 +83,13 @@ class DbBackup(models.Model):
              "read permissions for that file.",
     )
 
+    backup_format = fields.Selection(
+        [("dump.zip", "zip (includes filestore)"), ("dump", "pg_dump custom format (without filestore)")],
```
astirpe: Shouldn't be "`zip`" here? I'm wondering how this functionality could work fine with "`dump.zip`" passed to `dump_db()` :thinking:

astirpe (changes requested):

auto_backup/models/db_backup.py
foutoucour: True!

foutoucour: @astirpe fixed!

auto_backup/tests/test_db_backup.py
```diff
@@ -175,18 +175,6 @@ def test_action_backup_sftp_remote_open(self):
                     'wb'
                 )
 
-    def test_action_backup_sftp_remote_open(self):
```
astirpe, line 178: Why this test was removed?

auto_backup/models/db_backup.py
```diff
@@ -83,6 +83,16 @@ class DbBackup(models.Model):
              "read permissions for that file.",
     )
 
+    backup_format = fields.Selection(
+        [
+            ("zip", "zip (includes filestore)"),
+            ("dump", "pg_dump custom format (without filestore)")
+        ],
+        string="Backup Format",
```
astirpe: No need to write `string="Backup Format",`

auto_backup/tests/test_db_backup.py
foutoucour: Hello, 
I found that it is the same test than the previous test

auto_backup/models/db_backup.py
foutoucour: Question of style. I prefer explicit declaration that implicit declaration.
Has OCA guide on this that I didn't see?

astirpe: See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields

auto_backup/tests/test_db_backup.py
astirpe: Your're right

auto_backup/models/db_backup.py
foutoucour: 👍

foutoucour:
@astirpe could you approve the PR? Thank you.

astirpe (approved):
Code review ok

ddufresne (approved):

PR #1340, opened 2018-08-10
[REF] dbfilter_from_header: Use post_load hook to apply patch

moylop260:
It avoids apply patch if the module is not installed


pedrobaeza (commented):
Why don't you use Odoo `patch` method?

moylop260:
Hi @pedrobaeza 
Thanks for review

`http.db_filter` is not a `models` method.
More info about: 
 - https://github.com/OCA/maintainer-tools/pull/367#issuecomment-412388869

pedrobaeza (approved):

PR #1347, opened 2018-08-22
scheduler_error_mailer: fix version in manifest

sbidoul:
I thought we had a lint check for that.


pedrobaeza:
Well, it should be not working...

cc @moylop260

pedrobaeza (approved):

PR #1350, opened 2018-08-28
[ADD] field_dependant_default

etobella:
This module allows to define default function dependant of the sent values.
It is useful to create fields dependant of the first sent values.
The code will only be triggered on creation if the value is False or it is the
same than the default value

Configure
------
Fields must be defined with the function. For example::

        from odoo import api, fields, models

            _inherit = 'res.partner'

            ref = fields.Char(dependant_default='default_ref')

            def default_ref(self, vals):
                # Return the exepcted result from your vals
                return 'AAA'


etobella:
This addon could be useful in order to not edit the create function.
For example: https://github.com/OCA/stock-logistics-warehouse/blob/11.0/stock_request/models/stock_request_order.py#L245
@jbeficent

LoisRForgeFlow:
What about proposing this change directly to Odoo? Better if they integrate it than maintaining a module that will add a dependency to any module using the feature.

etobella:
@lreficent I created the Issue on odoo, but it will take some time to review it. Moreover, I think they will add this improvement on odoo 12 or odoo 13 if they approve the idea.
So, I think it will be a good idea to review it in order to improve the current modules. What do you think?

JordiBForgeFlow (approved):
:+1: Tested. Works as advertised

pedrobaeza:
The name for this shouldn't be this one, as `default` values are populated at the beginning of the lifecycle of the record (in cache). However, you are computing these "defaults" on record creation, after everything has interacted at cache level.

This can have also functionally some problems, with incoherences between values populated by onchanges and these defaults.

I don't recommend thus to have this as it's dangerous.

A better option was suggested by my colleague @Yajo some time ago: `@api.onwrite(fields)`. @Yajo, can you explain or link here about this?

yajo:
It's all here: https://github.com/odoo/odoo/pull/11042

etobella:
Well, the function should only be called on create, so `onwrite` will not work, because it is called on create and on write.
You are right with the lifecycle of the record, but it will not overwrite the value if it has changed from the default value.
If you think that the best option is to keep rewriting the `create` function we will continue this way.

pedrobaeza:
You can create `api.oncreate` decorator then.

pedrobaeza:
This is going to be luckily handled on v13 through several decorators: https://github.com/odoo/odoo/pull/31059

PR #1348, opened 2018-08-23
[WIP][11.0][ADD] base_cron_oneshot: Add the module

ghost:
### TODO:
- [ ] provide README
- [ ] finalize API method that should schedule new oneshot crons

PR #1352, opened 2018-08-29
[11.0][ADD] base_cron_oneshot

simahawk:
As per module's description:
"""
This module extends the functionality of Odoo crons to allow you to create single-purpose crons without any further setup or modules such as queue_job.

The typical use case is: you have an expensive task to run on demand and only once.

Once the cron has been executed it will be deleted.
"""
Replaces #1348


pedrobaeza:
I still prefer `queue_job` for persistence and control possibilities, but I'm not blocking this.

base_cron_oneshot/models/ir_cron.py
```diff
@@ -0,0 +1,69 @@
...
+        res = super()._process_job(job_cr, job, cron_cr)
+        if job.get('oneshot'):
+            try:
+                with api.Environment.manage():
+                    env = api.Environment(job_cr, job['user_id'], {})
+                    env[cls._name].browse(job.get('id')).unlink()
+                    _logger.info('One shot cron %s deleted' % job['cron_name'])
+            finally:
+                job_cr.commit()
+                cron_cr.commit()
```
grindtildeath: Do you really need to commit this cursor ? Doesn't seem to be used in this extension and should have been already committed in the call to super, and considering https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/ir/ir_cron.py#L120

base_cron_oneshot/tests/test_cron.py
```diff
@@ -0,0 +1,84 @@
...
+        cron_cr.rollback()
+        cron_cr.close()
+
+    def test_schedule_oneshot_unlink(self):
+        cron = self.env['ir.cron'].schedule_oneshot(
+            'res.partner', method='read')
+        cron_data = cron.read(load='_classic_write')[0]
+        with self._cron_fake_cursors() as (job_cr, cron_cr):
+            self.env['ir.cron']._process_job(job_cr, cron_data, cron_cr)
+        self.assertIn(cron.id, self.deleted)
```
grindtildeath: Wouldn't it be simpler to just search on `ir.cron` and make sure `cron` is not there anymore ?

grindtildeath (commented):
Shouldn't you set a new default value for `nextcall` on  `ir.cron` ? Or ask a delay to the user in `schedule_oneshot` ? Because it won't be really scheduled, as I understand it, the newly created ir.cron will be executed as soon as it's created and then deleted as soon as it's been executed with no other trace than a server log to keep a track of it. Then what happens if it fails, it will be deleted anyway ?
IMO would make sense to set these inactive instead of deleting, and have another `ir.cron` to delete these.

grindtildeath (approved):

tschanzt (approved):

sebalix (approved):

sbidoul:
@simahawk I fixed a title level issue in the README fragments at e818c57975a144ec24ab890017f6f8bccb28c247.

Out of curiosity, why do you prefer this over anthem or click-odoo or a plain odoo shell command?

PR #1354, opened 2018-09-10
[BACKPORT][8.0] base_tier_validation

andhit-r:


LoisRForgeFlow (approved):
No major changes, LGTM

andhit-r:
@lreficent could you point me to documentation or example on how to test abstract model. Many thanks.

LoisRForgeFlow:
@andhit-r I did it for this [module's v11](https://github.com/OCA/server-ux/blob/11.0/base_tier_validation/tests/test_tier_validation.py) and you can see here another example in v11: [base_exception tests](https://github.com/OCA/server-tools/tree/11.0/base_exception/tests). 

However, I tried at the time to add similar tests in base_tier_validation's v9 but didn't succeed :disappointed:. TBH, I don't exactly remember my blocking point, I saw tests for abstract modules in v9 and even 8 (there is also a template for it the [OCA template module](https://github.com/OCA/maintainer-tools/blob/master/template/module/tests/test_abstract_something.py)) and got very close in this particular case, but I eventually ended giving up.

andhit-r:
@lreficent thanks. I will look those sources. Codecov still need love.

PR #1360, opened 2018-09-19
[10.0][ADD]  Exception type as warning

TDu:
Allow to tag an exception as a warning. Contrary to a regular exception
a warning will be raised but not be blocking.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -6,6 +6,7 @@ import time
 
 from odoo import api, models, fields, _
 from odoo.exceptions import UserError, ValidationError
+from odoo.osv.expression import AND
 from odoo.tools.safe_eval import safe_eval
 from odoo import osv
 
@@ -52,6 +53,16 @@ class ExceptionRule(models.Model):
 #  - uid: current user id
 #  - context: current context
 """)
+    warning_only = fields.Boolean(
+        string='Warning only',
+        default=False
+    )
+    warning_text = fields.Char(
+        string='Warning message',
+        translate=True,
+        help='This message can be added to the record that raised a warning.'
+             'For a sale order line it could be the product name sold.'
+    )
 
     @api.constrains('exception_type', 'domain', 'code')
     def check_exception_type_consistency(self):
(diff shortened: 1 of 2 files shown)


base_exception/models/base_exception.py
```diff
@@ -62,6 +63,16 @@ class ExceptionRule(models.Model):
 #  - uid: current user id
 #  - context: current context
 """)
+    warning_only = fields.Boolean(
+        string='Warning only',
```
Tonow-c2c, line 57: Translation already added ?

grindtildeath (commented):
LGTM :+1: 
Can you please rebase and increase module's version number ?

bealdav:
ping @hparfr

hparfr (commented):
Hi,

I don't see why you need this.
You can already check a subset of rules in your own module with `_rule_domain`.
Blocking or not if there is some exceptions can be done in your own `test_exceptions`.

TDu:
Hello @hparfr, indeed this maybe achievable with a cusom `test_exceptions` method. But it is much more user friendly to be able to flag/see which exception are only warnings or not.
There is also an additional text for the warning which is handy.
One of the use case for this (which is live on a production instance) is on sale order lines. Warning only would be displayed on the line as additional information to not forget but they are not blocking to confirm the sale order.

bealdav:
@TDu thanks for this contribution. On my side I think it's a feature that can be implemented in another module (base_exception <> extended_exception ;-) ). 
Then, you can explain more precisely the use case and how implement it in sale for example.
Thanks a lot

base_exception/models/base_exception.py
```diff
@@ -73,6 +74,16 @@ class ExceptionRule(models.Model):
...
+        default=False
+    )
+    warning_text = fields.Char(
+        string='Warning message',
+        translate=True,
+        help='This message can be added to the record that raised a warning.'
+             'For a sale order line it could be the product name sold.'
```
yvaucher, line 64: Missing a space before `For`

grindtildeath (approved):

i-vyshnevska (approved):

hparfr (changes requested):
Please create another module in order to keep base_exception module away from the warning logic.
Most of the dependencies (purchase, sale, product?) of base_exception don't have to deal with warnings.

PR #1361, opened 2018-09-19
[MIG] base_fontawesome: Migration to 11.0

etobella:
This module upgrades the fontawesome font to 5.3.1.
It allows to use new icons.
Following the instructions of OCA/web#1042


AaronHForgeFlow:
Got this on runbot when logging in, no idea if it is due this module:
image

etobella:
The gitignore was removing the lib folder. I removed it, following the configuration of OCA/web

JordiBForgeFlow (approved):
:+1: Functional review. Works great!

base_fontawesome/README.rst
```diff
@@ -0,0 +1,59 @@
...
+=====
+
+Now, all free icons from `Font Awesome <https://fontawesome.com/icons?d=gallery&m=free>`_
+can be used on odoo. It also adds three options in order to add the icons:
+* solid_icon
+* regular_icon
+* brand_icon
+The usage depends on the configuration of the icon.
+
+For example, brand_icon is necessary if we are using an icon of a branch::
```
JordiBForgeFlow: brand, not branch

LoisRForgeFlow (approved):
Tested on runbot, looks great!

AaronHForgeFlow (approved):
:+1: Functional test

JordiBForgeFlow:
@simahawk we can wait for your final review, as you authored the module in v10, and then this one can be merged.

base_fontawesome/templates/assets.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+Copyright 2017 Simone Orsi
+License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+<odoo>
+
+	<template id="assets_backend" inherit_id="web.assets_backend">
+        <xpath expr="." position="inside">
+			<link rel="stylesheet" type="text/css" href="/base_fontawesome/static/lib/fontawesome-5.3.1/css/all.css" />
```
simahawk: can we get rid of unused resources? :)

base_fontawesome/templates/assets.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+Copyright 2017 Simone Orsi
+License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+<odoo>
+
+	<template id="assets_backend" inherit_id="web.assets_backend">
```
simahawk: AFAIK you should use assets_common to match both back and frontend

base_fontawesome/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2017 Simone Orsi
+# Copyright 2018 Creu Blanca
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Base Fontawesome',
+    'summary': """Up to date Fontawesome resources.""",
+    'version': '11.0.1.0.0',
```
simahawk: module version, as per README, should match the version of FA ;)

simahawk (changes requested):

base_fontawesome/templates/assets.xml
etobella: If I replace the elements it is giving problems, related to the usage of FontAwesome font directly on some code by odoo. I will try to review if it is possible to fix the problem.

etobella:
@simahawk I finally applied all the changes that you suggested. There was an issue with the configuration of some buttons, but it seems right now. I also changed the version.

Just in order to leave it prepared for merge, I rebased and squashed all the fixup comits.

base_fontawesome/static/lib/fontawesome-5.3.1/css/brands.css
```diff
@@ -0,0 +1,13 @@
+/*!
```
simahawk: already mentioned in the old PR: do we really have to keep both compressed and uncompressed?
Our repos are already heavy... I think it's better to drop useless files especially if big :)

base_fontawesome/templates/assets.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+Copyright 2017 Simone Orsi
+License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+<odoo>
+
+	<template id="assets_common" inherit_id="web.assets_common">
+		<xpath expr="link[@href='/web/static/lib/fontawesome/css/font-awesome.css']" position="replace">
```
simahawk: can you check indentation? seems messed up :/

base_fontawesome/static/lib/fontawesome-5.3.1/css/brands.css
etobella: I will remove all the unused files.

simahawk (approved):
LG thanks!

simahawk:
you can squash & merge sir :)

etobella:
Well, I found an small issue on sortable columns. I will fix it first :wink:

etobella:
imagen
Sorting icons are not shown correctly, because It tries to find FontAwesome font but it no longer exists. 
We cannot replace fontawesome original css...

simahawk:
loading them all does not sound like a good idea, sorry :)
Who's causing the issue? Who tries to find what?
Where is the line adding that action? Can't we make it use new FA?

etobella:
The problem is that the old font is referenced on some lessc files:
https://github.com/odoo/odoo/search?l=Less&q=FontAwesome
As they are using it directly it gives problems.

simahawk:
ahh they've changed the name....... `font-family: "Font Awesome 5 Free";  /* updated font-family */`

simahawk:
@etobella do you think we can simply re-define the font-face w/ the old name?
Like here https://github.com/OCA/server-tools/pull/1361/files#diff-5163cade13025e35cf7b4df7687f8697R3787
We could split it to a different compat-only file

etobella:
I finally made it!!
imagen
The problem was related to the link of the font :)

PR #1363, opened 2018-09-21
11.0 mass editing

CGenie:
This seems to work for us since April/May 2018 so I guess it's worth merging :)


pedrobaeza:
It's already migrated since a lot in https://github.com/OCA/server-ux/tree/11.0/mass_editing. You should see migration tracking issue: https://github.com/OCA/server-tools/issues/1008

PR #1369, opened 2018-09-27
12.0 - [ADD] profiler: Integration of python cprofile and postgresql logging collector for Odoo

moylop260:
# TODO
- [ ] Fix new conventions
- [ ] Fix if the user doesn't have permission of superuser for psql
- [ ] Add default filters to remove from profiling lines tree view the following unused output:
```bash
odoo/odoo/http.py:513 (response_wrap)
odoo/addons/bus/controllers/main.py:32 (poll)
odoo/odoo/http.py:328 (checked_call)
odoo/odoo/http.py:936 (__call__)
odoo/addons/mail/controllers/bus.py:21 (_poll)
odoo/addons/bus/controllers/main.py:24 (_poll)
odoo/addons/bus/models/bus.py:113 (poll)
/usr/lib/python3.5/threading.py:531 (wait)
```

### DIFF
--- a/profiler/hooks.py
+++ b/profiler/hooks.py
@@ -1,0 +1,45 @@
+
+import logging
+
+from openerp.http import WebRequest
+from openerp.sql_db import Cursor
+
+from .models.profiler_profile import ProfilerProfile
+
+_logger = logging.getLogger(__name__)
+
+
+def patch_web_request_call_function():
+    """Modify Odoo entry points so that profile can record.
+
+    Odoo is a multi-threaded program. Therefore, the :data:`profile` object
+    needs to be enabled/disabled each in each thread to capture all the
+    execution.
+
+    For instance, Odoo spawns a new thread for each request.
+    """
+    _logger.info('Patching http.WebRequest._call_function')
+    webreq_f_origin = WebRequest._call_function
+
+    def webreq_f(*args, **kwargs):
+        with ProfilerProfile.profiling():
+            return webreq_f_origin(*args, **kwargs)
+    WebRequest._call_function = webreq_f
+
+
+def patch_cursor_init():
+    _logger.info('Patching sql_dp.Cursor.__init__')
+    cursor_f_origin = Cursor.__init__
+
+    def init_f(self, *args, **kwargs):
+        cursor_f_origin(self, *args, **kwargs)
+        enable = ProfilerProfile.activate_deactivate_pglogs
+        if enable is not None:
+            self._obj.execute('SET log_min_duration_statement TO "%s"' %
+                              ((not enable) * -1,))
+    Cursor.__init__ = init_f
+
+
+def post_load():
+    patch_web_request_call_function()
+    patch_cursor_init()
--- a/profiler/hooks.py
+++ b/profiler/hooks.py
@@ -1,8 +1,9 @@
 
 import logging
 
-from openerp.http import WebRequest
-from openerp.sql_db import Cursor
+from odoo.http import WebRequest
+from odoo.sql_db import Cursor
+from odoo import api
 
 from .models.profiler_profile import ProfilerProfile
 
@@ -27,6 +28,16 @@ def patch_web_request_call_function():
     WebRequest._call_function = webreq_f
 
 
+def patch_call_kw_function():
+    _logger.info('Patching odoo.api.call_kw')
+    call_kw_f_origin = api.call_kw
+
+    def call_kw_f(*args, **kwargs):
+        with ProfilerProfile.profiling():
+            return call_kw_f_origin(*args, **kwargs)
+    api.call_kw = call_kw_f
+
+
 def patch_cursor_init():
     _logger.info('Patching sql_dp.Cursor.__init__')
     cursor_f_origin = Cursor.__init__
@@ -42,4 +53,5 @@ def patch_cursor_init():
 
 def post_load():
     patch_web_request_call_function()
+    patch_call_kw_function()
     patch_cursor_init()
--- a/profiler/hooks.py
+++ b/profiler/hooks.py
@@ -1,9 +1,8 @@
 
 import logging
 
-from odoo.http import WebRequest
-from odoo.sql_db import Cursor
-from odoo import api
+from openerp.http import WebRequest
+from openerp.sql_db import Cursor
 
 from .models.profiler_profile import ProfilerProfile
 
@@ -28,16 +27,6 @@ def patch_web_request_call_function():
     WebRequest._call_function = webreq_f
 
 
-def patch_call_kw_function():
-    _logger.info('Patching odoo.api.call_kw')
-    call_kw_f_origin = api.call_kw
-
-    def call_kw_f(*args, **kwargs):
-        with ProfilerProfile.profiling():
-            return call_kw_f_origin(*args, **kwargs)
-    api.call_kw = call_kw_f
-
-
 def patch_cursor_init():
     _logger.info('Patching sql_dp.Cursor.__init__')
     cursor_f_origin = Cursor.__init__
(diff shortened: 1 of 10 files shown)


moylop260:
@carlosdauden 
This is the module that I talked with you

etobella:
Amazing!!! Will it be ported to 11.0?

I tested it, and it worked like a charm!

MiquelRForgeFlow:
What is the difference between this profiler module and the `odoo.tools.profiler`? Because if they are not the same, maybe they should not have the same name.

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+    enable_postgresql = fields.Boolean(
+        default=False,
+        help="It requires postgresql server logs seudo-enabled")
+    use_py_index = fields.Boolean(
+        name="Get cProfiling report", default=False,
+        help="Index human-readable cProfile attachment."
+        "\nTo access this report, you must open the cprofile attachment view "
+        "using debug mode.\nWarning: Uses more resources.")
+    date_started = fields.Char(readonly=True)
+    date_finished = fields.Char(readonly=True)
```
yajo, line 100: Why not `fields.Datetime()`?

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+    state = fields.Selection([
+        ('enabled', 'Enabled'),
+        ('disabled', 'Disabled'),
+    ], default='disabled', readonly=True, required=True)
+    description = fields.Text(readonly=True)
+    attachment_count = fields.Integer(compute="_compute_attachment_count")
+    pg_log_path = fields.Char(help="Getting the path to the logger",
+                              default=_find_loggers_path)
```
yajo, line 108: Use a lambda please. This allows to override the method without overriding the field, and allows to sort default methods, putting them below fields.

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+        pgbadger_cmd = self._get_pgbadger_command()
+        if pgbadger_cmd is None:
+            return
+        pgbadger_cmd_str = subprocess.list2cmdline(pgbadger_cmd)
+        self.description += (
+            '\nRunning the command: %s') % pgbadger_cmd_str
+        result = tools.exec_command_pipe(*pgbadger_cmd)
+        datas = result[1].read()
+        if not datas:
+            self.description += "\nPgbadger output is empty!"
```
yajo, line 278: AFAIK each `self.description +=` will issue a `self.write()` from the ORM. Isn't it better to just use a normal variable and assign only once at the end?

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+            result.append(
+                tools.html_sanitize(lxml.html.tostring(this_result[0])))
+        return result
+
+    @api.multi
+    def _get_pgbadger_command(self):
+        self.ensure_one()
+        # TODO: Catch early the following errors.
+        try:
+            pgbadger_bin = tools.find_in_path('pgbadger')
```
yajo, line 314: This is missing in the manifest.

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+        action['domain'] = [('id', 'in', attachments.ids)]
+        return action
+
+    @api.model
+    def set_pgoptions_enabled(self):
+        """Verify if postgresql has configured the parameters for logging"""
+        ProfilerProfile.pglogs_enabled = True
+        pgoptions_enabled = bool(os.environ.get('PGOPTIONS'))
+        _logger.info('Logging enabled from environment '
+                     'variable PGOPTIONS? %s', pgoptions_enabled)
```
yajo, line 457: Does this make sense? :thinking: What if Postgres is in a different server?

To check enabled options, shouldn't you get those from the database itself?

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+        pg_config_file = self.env.cr.fetchone()[0]
+        db_host = tools.config.get('db_host')
+        if db_host == 'localhost' or db_host == '127.0.0.1':
+            db_host = False
+        if db_host:
+            pg_config_file = 'postgres@%s:%s' % (db_host, pg_config_file)
+            self.pg_remote = db_host
+
+        self.description = (
+            "You need seudo-enable logs from your "
```
yajo, line 143: "pseudo"

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+            "postgresql-server configuration file.\n"
```
yajo: Any chance for this very addon to enable those options? Using https://www.postgresql.org/docs/current/static/sql-set.html or similar?

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+        pgoptions_enabled = bool(os.environ.get('PGOPTIONS'))
+        _logger.info('Logging enabled from environment '
+                     'variable PGOPTIONS? %s', pgoptions_enabled)
+        if pgoptions_enabled:
+            return
+        pgparams_required = {
+            'log_min_duration_statement': '0',
+        }
+        for param, value in pgparams_required.items():
+            # pylint: disable=sql-injection
```
yajo, line 464: You can use [`AsIs`](http://initd.org/psycopg/docs/extensions.html#psycopg2.extensions.AsIs) in the params tuple and avoid this line.

profiler/models/profiler_profile.py
```diff
@@ -0,0 +1,470 @@
...
+            self.env.cr, PGOPTIONS.keys())
+        _logger.info('Logging enabled from postgresql.conf? %s',
+                     ProfilerProfile.pglogs_enabled)
+
+    @staticmethod
+    def get_psql_params(cr, params):
+        result = {}
+        for param in set(params):
+            # pylint: disable=sql-injection
+            cr.execute('SHOW %s' % param)
```
yajo, line 480: Possible SQL injection? :thinking: :warning:

profiler/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2018 Vauxoo (https://www.vauxoo.com) <info@vauxoo.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
yajo, line 2: I wouldn't put these lines if they amount more than the code itself :grin:

yajo (changes requested):

profiler/hooks.py
```diff
@@ -0,0 +1,45 @@
+
+import logging
+
+from openerp.http import WebRequest
+from openerp.sql_db import Cursor
```
bealdav: Oops openerp in v12 ?

etobella:
I backported the module to v11 using your original code, but I discovered that it is not checking the external xmlrpc calls. https://github.com/etobella/server-tools/tree/11.0-mig-profiler (I will propose the  PR to OCA once it is approved on v12).
On our case, odoo is integrated with another system through xmlrpc calls. However, profiler is not checking this calls. Do you know why?

etobella:
I found the problem, currently, the system is only patching the WebRequest. I have made some changes on my branch in order to review also xmlrpc calls. I also added a filter for the profile lines.
https://github.com/etobella/server-tools/commit/8b6ff7dbfd5b552403762f95c31d3a059ed8b893

etobella:
@moylop260 If you are reusing my commit, you should maintain me as the author...

moylop260:
Hi @etobella 
I hope you are doing well!

Thanks for your contribution.

For next time, you can create a  PR to my dev-branch.
Regards!

If you have change to fix another feedback it is welcome, since that I didn't find time yet.
Thank you

PR #1372, opened 2018-09-30
12.0 - [ADD] company_country: Set company.country value early to use account hook to install l10n_* modules and avoid installing l10n_generic_coa

moylop260:
Allow to use [account._auto_install_l10n](https://github.com/odoo/odoo/blob/a01c1bf5bf377471c1c0fb31a20b59fc116fd545/addons/account/__init__.py#L13) without install `l10n_generic_coa`


luistorresm (approved):
Technical  & functional review 👍

dreispt:
This sounds like an issue in core Odoo, no? Or is this for CI/Automatic database provisioning scenarios?

moylop260:
Hi @dreispt 

> This sounds like an issue in core Odoo, no? Or is this for CI/Automatic database provisioning scenarios?

It is not really a Odoo issue, it is an expected behaviour for odoo since that there is a hook auto-installing `l10n_**` modules depending of the `company_id.country_id`:
 - https://github.com/odoo/odoo/blob/660cebb4faaebd64da8fc502f5d42fd7f3196513/addons/account/__init__.py#L13

So, you can:
1) Create a `/web/database/manager´
2) Choose the Country. e.g. México.
3) Install `account` module.
4) You will see auto-install `l10n_mx` module.

It is great!

The main issue is when you are not using the `/web/database/manager` because if you have a module where the country is changed but it module depends of `account` so the hook run first before to install your custom module so before to change country.

I mean,

1) odoo-bin -d db -i custom_module
  - custom_module has manifest depends: account and set company.country to México.
2) You will see `l10n_generic_coa` module installed instead of `l10n_mx`.
Since that your custom module depends of account, and it is installed before custom module. So account run a hook.

This module makes a workaround in order to force use a specific country before to install account module.

You can test this behaviour using the database `-base` and change the country and install account.

Screen Shot 2019-11-12 at 22 21 35

bealdav (approved):

dreispt:
@moylop260 Thank for your time writing a detailed explanation, I wasn't aware of this mechanism.
It still sound like a design issue that should be solved in Odoo core.
Please go ahead with the PR, but I think it might be a good idea to open a discussion with Odoo people.

dreispt:
PS: `_auto_install_l10n()` looks quite hacky, with all those hardcoded codes and rule. It is begging to be refactored...

Idea: have a many2many "Localization modules" between res.country and ir.modules. Installation would be triggered onchange of company.country_id or on adding new record to the list.
I know, I'm off topic for the discussion here...

moylop260:
@dreispt 
I totally agree with you
Trust me, I discussed this hacky method with Quentin of Odoo but this method is used for 13.0 version too.
So, revert this for stable will be a hard war
Maybe, we should think a better way for 14.0?

Looking for an Odoo and OCA contribution in order to avoid hacky methods so hacky modules

max3903 (approved):

moylop260:
/ocabot merge

Second attempt

PR #1373, opened 2018-10-01
[MIG] mail_enviroment: Migrated from v10

luistorresm:


sbidoul:
This should be migrated from 11 to 12 in https://github.com/oca/server-env. Therefore closing.

PR #1375, opened 2018-10-01
[Fix] dot in report name is not mandatory

diggy128:
Create requires that report name has at least one dot in it, however in core Odoo there are reports that do not have any and thus fail migration eg. mrp_bom_cost in mrp module.

### DIFF
--- a/base_report_auto_create_qweb/models/report_xml.py
+++ b/base_report_auto_create_qweb/models/report_xml.py
@@ -59,11 +59,11 @@ class IrActionsReport(models.Model):
     def create(self, values):
         values['report_name'] = self._format_template_name(
             values.get('report_name', ''))
-        if (values.get('report_type') in ['qweb-pdf', 'qweb-html'] and
-                values.get('report_name') and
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@oihane can you confirm this?

Any way, @diggy128 you should remove code, not comment it.

diggy128:
> Any way, @diggy128 you should remove code, not remote it.

I know, I just commented it because I don't know for sure it doesn't create implications when normally running the server.
If you confirm it can be removed, I'll push another commit removing it completely.

oihane:
It is version 9.0, I don't know, but it usually puts ids with the module name and it is based on those ids. Isn't it mrp_bom_cost taken as mrp.mrp_bom_cost??

diggy128:
That's what I thought too, but it seems not.
In any case, if upstream doesn't impose any restriction on the name wouldn't it be best if we don't add any extra restrictions?

pedrobaeza:
@oihane what is the reason for the restriction?

PR #1377, opened 2018-10-02
[ADD] Add module base_conditional_image

p-tombez:
Addition of a custom image module.
As soon as it is accepted, I'll port it to V10+


tschanzt (approved):
code review only. but looks good.

custom_image/models/image.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class Image(models.Model):
+    _name = 'image'
+
+    name = fields.Char('Name', required=True)
+    model_name = fields.Char('Model Name', required=True)
+    selector = fields.Text(
+        'Selector',
+        help='Python expression used to distinguish the images '
```
simahawk: distinguish images for what?

custom_image/models/abstract_image.py
```diff
@@ -0,0 +1,56 @@
...
+    image_small = fields.Binary(
+        compute='_compute_image', string="Image",
+        store=False, readonly=True
+    )
+
+    def _compute_image(self):
+        cls = type(self)
+
+        images = self.env['image'].search(
+            [('model_name', '=', cls._name)],
```
simahawk: self._name is enough

custom_image/models/abstract_image.py
```diff
@@ -0,0 +1,56 @@
...
+            order='company_id, selector'
+        )
+
+        def test_method(im, r):
```
simahawk: ?

custom_image/models/abstract_image.py
```diff
@@ -0,0 +1,56 @@
...
+            return ((im.company_id == r.company_id or
+                     r.company_id and not im.company_id) and
+                    safe_eval(im.selector or "True", {'object': r}))
+
+        if not hasattr(cls, 'company_id'):
```
simahawk:
```python
if 'company_id' in self:
```
or, to be more explicit
```python
if 'company_id' in self._fields:
```

custom_image/models/abstract_image.py
```diff
@@ -0,0 +1,56 @@
...
+            # If inherited object doesn't have a `company_id` field,
+            # remove the items with a company defined and the related checks
+            images = images.filtered(lambda r: not r.company_id)
```
simahawk: you are already searching, do this check in the search above instead

custom_image/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Custom Image
+============
+
+This module intents to add images to any object, or to override the standard Odoo behavior on existing ones.
```
simahawk: to achieve what?
I'll put review on hold until you clarify what are the main goals and - for instance - how is this different from base_multi_image and/or if it's compatible w/ it :wink: 

A real use case might help

simahawk (changes requested):

p-tombez:
I fixed the code according to the comments and also rephrased the module description as well as renaming it to better match its actual purpose

base_conditional_image/models/abstract_image.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    image_small = fields.Binary(
+        compute='_compute_image', string="Image",
+        store=False, readonly=True
+    )
+
+    def _compute_image(self):
+        if 'company_id' in self._fields:
+            images = self.env['image'].search(
+                [('model_name', '=', cls._name)],
```
acsonefho: `cls` is not defined, you can use `self`.

base_conditional_image/models/abstract_image.py
```diff
@@ -0,0 +1,57 @@
...
+                order='company_id, selector'
+            )
+            def test_method(image, rec):
+                return ((image.company_id == rec.company_id or
```
acsonefho: It's quite hard to read/understand, can you rewrite it please?

base_conditional_image/models/image.py
```diff
@@ -0,0 +1,58 @@
...
+        help="Medium-sized image. It is automatically "\
+             "resized as a 128x128px image, with aspect ratio preserved. "\
+             "Use this field in form views or some kanban views.")
+    image_small = fields.Binary("Small-sized image", attachment=True,
+        help="Small-sized image. It is automatically "\
+             "resized as a 64x64px image, with aspect ratio preserved. "\
+             "Use this field anywhere a small image is required.")
+
+    @api.multi
+    def _process_images(self, vals, required=False):
```
acsonefho, line 48: This function should be `api.model`

base_conditional_image/views/image_view.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+  <data>
```
acsonefho, line 3: Entire file is bad intended (should be 4 spaces)

base_conditional_image/views/image_view.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
acsonefho, line 2: I think it should be `odoo` and not `openerp`

base_conditional_image/README.rst
```diff
@@ -0,0 +1,71 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
acsonefho: This file should be generated automatically.
You have to create a readme directory with every sections of the `README.rst`.

acsonefho (changes requested):

base_conditional_image/views/image_view.xml
leemannd: @acsonefho AFAIR, there is no common agreement on the tab space into OCA repos.

base_conditional_image/views/image_view.xml
p-tombez: It's a V9 branch, so it's ```openerp``` not ```odoo```

p-tombez: It's a V9 branch, so it's ```openerp``` not ```odoo```

base_conditional_image/models/abstract_image.py
p-tombez: You're right, I missed it during refactoring

base_conditional_image/models/abstract_image.py
p-tombez: I don't really see how to rewrite that, do you have a proposal ?

acsonefho: No I don't, sorry =/ But I think it could be more readable maybe with 2 `if`?

base_conditional_image/views/image_view.xml
acsonefho: Ok ok no problem, I thought there was

base_conditional_image/views/image_view.xml
acsonefho: Ok ok, I think I work too much on v10.0 :)

base_conditional_image/README.rst
p-tombez: It's a V9 branch, no module are using this pattern. I'll switch to it when I migrate the module to V10 and later

acsonefho: I think it's not related to the version.
No module in v9.0 using this pattern because this pattern is quite recent

leemannd: AFAIK it's a recommendation, and it's not mandatory. And the readme is really explicit.
@acsonefho Do you think your comment on this point can be dismissed?

acsonefho: @leemannd Some say that's required and others just a recommendation.
On the documentation (on maintainer-tools repo), the readme file is mandatory and the way to generate it is with the `readme` directory and some files.
I think that someone should take a final decision about that.
I'll not "block" the MR for this of course!

base_conditional_image/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    'name': 'Image Storage',
+    'version': '9.0.1.0.0',
+    'author': 'Odoo Community Association (OCA), Camptocamp',
+    'license': 'AGPL-3',
+    'category': 'Misc',
+    'depends': [
+        'base',
+    ],
+    'website': 'http://www.camptocamp.com',
```
simahawk: you know... :)

base_conditional_image/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    'name': 'Image Storage',
+    'version': '9.0.1.0.0',
+    'author': 'Odoo Community Association (OCA), Camptocamp',
```
simahawk: c2c 1st, OCA always at last place

base_conditional_image/README.rst
simahawk: Usually is not considered mandatory but for NEW modules I'd argue it would be :man_teacher: 
Plus, splitting this to readme folder is a 5mins operation :wink:

base_conditional_image/models/abstract_image.py
simahawk: I'd do it like this more or less:
```python
    [...]

    @staticmethod
    def _compute_image_test_without_company(image, rec):
        return bool(safe_eval(image.selector or "True", {'object': rec})))

    @staticmethod
    def _compute_image_test_with_company(image, rec):
        if (image.company_id == rec.company_id or 
                rec.company_id and not image.company_id):
            return self._compute_image_test_without_company(image, rec)
        return False

    def _compute_image(self):
        domain = [('model_name', '=', self._name)]
        if 'company_id' in self._fields:
            test_func = self._compute_image_test_with_company
        else:
            # If inherited object doesn't have a `company_id` field,
            # remove the items with a company defined and the related checks
            domain.append(('company_id', '=', False))
            test_func = self._compute_image_test_without_company

        images = self.env['image'].search(domain, order='company_id, selector')
        for rec in self:
            found = None
            for image in images:
                if test_func(image, rec):
                    found = image
                    break
            if found:
                rec.update({
                    'image': found.image,
                    'image_medium': found.image_medium,
                    'image_small': found.image_small,
                })
```

simahawk (changes requested):

base_conditional_image/models/abstract_image.py
```diff
@@ -0,0 +1,57 @@
...
+                [('model_name', '=', self._name),
+                 ('company_id', '=', False)],
+                order='company_id, selector'
+            )
+            def test_method(image, rec):
+                return safe_eval(image.selector or "True", {'object': rec})
+
+        for rec in self:
+            found = None
+            for image in images:
```
simahawk: @p-tombez this thing can be reaaaaally expensive. Are we sure there's no way to limit this?

p-tombez: I don't see what could be expensive here nor any way of limiting anything more. There is already an exit condition as soon as a result is found...

simahawk: @p-tombez an I wrong or potentially you can have gazillion of images here?

p-tombez: I think you're wrong ;)
The recordset ```images``` will only hold the image "selectors" for the current object.
Indeed, the size will depends on how the user configured these selectors, but I think it will not go over a dozen

guewen (approved):

base_conditional_image/models/abstract_image.py
simahawk: @p-tombez ok, then pls add a comment of top of it saying that it should be fine in most of the cases because bla bla bla ;)

simahawk:
@p-tombez build is red. Also, pls, add the comment mentioned in https://github.com/OCA/server-tools/pull/1377#discussion_r248267223

p-tombez:
@simahawk In my previous commit (https://github.com/OCA/server-tools/pull/1377/commits/7c7bcf46b9741479be2d34e74cc3b0636e1cff1d) I changed the names to avoid confusion. I think the code is now pretty self-explanatory and does not require a comment anymore (unless you don't know how nested for loop works).
For the red build, I will not fix the lint issues in the other modules (at least, not in this PR) and the errors in the module ```oauth_provider``` were already there before I made this PR (https://travis-ci.org/OCA/server-tools/builds/435521532?utm_source=github_status&utm_medium=notification)

simahawk:
@p-tombez fine, naming works better. Build is still red tho.

PR #1378, opened 2018-10-02
[12.0][MIG] server environment ir config parameter

GillesTephaneMeyomesse:
#1368


server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -0,0 +1,54 @@
...
+SECTION = 'ir.config_parameter'
+
+
+class IrConfigParameter(models.Model):
+
+    _inherit = 'ir.config_parameter'
+
+    @api.model
+    def get_param(self, key, default=False):
+        value = super(IrConfigParameter, self).get_param(key, default=None)
```
acsonefho, line 18: In Python 3 you can let the super() empty:
`value = super().get_param(key, default=None)`
It's certainly not an issue (about Python) but it's not necessary.

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -0,0 +1,54 @@
...
+            return default
+        return value
+
+    @api.model
+    def create(self, vals):
+        key = vals.get('key')
+        if serv_config.has_option(SECTION, key):
+            # enforce value from config file
+            vals = dict(vals, value=serv_config.get(SECTION, key))
+        return super(IrConfigParameter, self).create(vals)
```
acsonefho, line 42: Same here (about the `super()`)

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -0,0 +1,54 @@
...
+    @api.multi
+    def write(self, vals):
+        for rec in self:
+            key = vals.get('key') or rec.key
+            if serv_config.has_option(SECTION, key):
+                # enforce value from config file
+                newvals = dict(vals, value=serv_config.get(SECTION, key))
+            else:
+                newvals = vals
+            super(IrConfigParameter, rec).write(newvals)
```
acsonefho, line 53: But here you have to keep it because you use `rec` and not `self`.

server_environment_ir_config_parameter/tests/test_server_environment_ircp.py
```diff
@@ -0,0 +1,80 @@
...
+
+from odoo.exceptions import UserError
+from odoo.tests import common
+from odoo.tools import convert
+
+
+class TestEnv(common.TransactionCase):
+
+    def setUp(self):
+        super(TestEnv, self).setUp()
```
acsonefho, line 14: Same comment here (about `super()`)

acsonefho (changes requested):
Just minor changes but seems good (about code)

ThomasBinsfeld (approved):
Fix the super() and it's ok

sbidoul:
This should go in OCA/server-env, can you move it there?

sbidoul (changes requested):
This should go to OCA/server-env

PR #1379, opened 2018-10-02
[ADD] 10.0 base_jsonify

PierrickBrun:
Add module base_jsonify to repository.

This module add the jsonify method to the ORM. This method take as argument
the browse record and the "parser" that specify the field to extract.

Example of parser:

    parser = [
        'name',
        'number',
        'create_date',
        ('partner_id', ['id', 'display_name', 'ref'])
        ('line_id', ['id', ('product_id', ['name']), 'price_unit'])
    ]

In order to be consitent with the odoo api the jsonify method always
return a list of object even if there is only one element in input

By default the key into the json is the name of the field extracted
from the model. If you need to specify an alternate name to use as key, you
can define your mapping as follow into the parser definition:

    parser = [
         'field_name:json_key'
    ]

    parser = [
        'name',
        'number',
        'create_date:creationDate',
        ('partner_id:partners', ['id', 'display_name', 'ref'])
        ('line_id:lines', ['id', ('product_id', ['name']), 'price_unit'])
    ]

Also the module provide a method "get_json_parser" on the ir.exports object
that generate a parser from an ir.exports configuration


hparfr (approved):

sbidoul (commented):
Could you use the readme fragments mechanism?

PierrickBrun:
Hi @sbidoul , The module uses the readme fragments mechanism now, can you re-check and approve the PR ?

lmignon (changes requested):
@PierrickBrun It seems that the proposed sources is not the last one available on the original repo...  A least the know issue has been resolved by https://github.com/akretion/ak-odoo-incubator/commit/7297f5418830e01cbc6e6a5adc4724801448e689#diff-06d91370052ae6bd8a15b06f6c613182

Moreover, IMO we must keep the commit history from the original repo. You can find different ways to keep the history when moving one folder from one repo to an other one.
I've experienced yesterday a way to do it based on `git format-patch`

1. into the original repo:
    ```
    git format-patch --keep-subject  --root 10.0 -- base_jsonify
    ```
    The command will generate a patch file for each commit involving base_jsonify. (0001-*.patch,0002-*.patch,0003-*.patch............)

2. Checkout the 10.0 branch of https://github.com/OCA/server-tools and create a new branch
3. Into server-tools, apply each patch generated into you orginal repository with
    ```
    git am ../ak-odoo-incubator/*.patch
    ```

PierrickBrun:
Thank you for your detailed answer @lmignon , I've updated the PR with the forgotten fix and kept the history.

I was experimenting the same thing right now, I've found a similar but slightly different solution from this StackOverflow response : https://stackoverflow.com/a/11426261/7042184 

This is how I updated the PR. I don't know if this way is more adapted than yours but I think that edge cases have been well tested, judging from the discussion about it.

base_jsonify/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+Make a distinction between float 0.0L and no value
```
lmignon: @PierrickBrun IMO this issue has been solved by 3ca15ceada6c48ac9e6a87f41c41719dcfe4ca89

lmignon (changes requested):
The roadmap is no more valid and should be removed...

base_jsonify/readme/ROADMAP.rst
PierrickBrun: indeed

lmignon (approved):

base_jsonify/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (http://www.akretion.com)
```
simahawk: Copyright

base_jsonify/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+# Sébastien BEAU <sebastien.beau@akretion.com>
+# Raphaël Reverdy <raphael.reverdy@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Jsonify",
+    "summary": "Base module that provide the jsonify method on all object",
+    "version": "10.0.1.0.0",
+    "category": "Uncategorized",
+    "website": "https://odoo-community.org/",
```
simahawk: https://github.com/OCA/server-tools

base_jsonify/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (http://www.akretion.com)
+# Sébastien BEAU <sebastien.beau@akretion.com>
+# Raphaël Reverdy <raphael.reverdy@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Jsonify",
+    "summary": "Base module that provide the jsonify method on all object",
```
simahawk: on all models

base_jsonify/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "version": "10.0.1.0.0",
+    "category": "Uncategorized",
+    "website": "https://odoo-community.org/",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
```
simahawk: empty, remove it

base_jsonify/models/ir_export.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (http://www.akretion.com)
+# Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from collections import OrderedDict
+from odoo import api, models
+
+
+def update_dict(data, fields):
```
simahawk, line 10: docstring

base_jsonify/models/ir_export.py
```diff
@@ -0,0 +1,44 @@
...
+        if field == '.id':
+            field = 'id'
+        data[field] = True
+    else:
+        if field not in data:
+            data[field] = OrderedDict()
+        update_dict(data[field], fields[1:])
+
+
+def convert_dict(dict_parser):
```
simahawk, line 30: docstring

base_jsonify/models/ir_exports_line.py
```diff
@@ -0,0 +1,35 @@
...
+    _order = 'name'
+
+    alias = fields.Char(
+        'Alias',
+        help='The complete path to the field where you can specify an '
+             'alias on the a step as field:alias'
+    )
+
+    @api.constrains('alias', 'name')
+    def _check_alias(self):
```
simahawk, line 20: docstring

base_jsonify/models/ir_exports_line.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import api, fields, models, _
+
+
+class IrExportsLine(models.Model):
+    _inherit = 'ir.exports.line'
+    _order = 'name'
+
+    alias = fields.Char(
+        'Alias',
+        help='The complete path to the field where you can specify an '
```
simahawk, line 15: please rephrase

base_jsonify/models/models.py
```diff
@@ -0,0 +1,75 @@
...
+from odoo.exceptions import UserError
+from odoo.tools.translate import _
+
+
+class Base(models.AbstractModel):
+
+    _inherit = 'base'
+
+    @api.model
+    def __parse_field(self, parser_field):
```
simahawk, line 17: any reason for double underscore?

simahawk, line 17: docstring

base_jsonify/models/models.py
```diff
@@ -0,0 +1,75 @@
...
+        if isinstance(parser_field, tuple):
+            field_name, subparser = parser_field
+        json_key = field_name
+        if ':' in field_name:
+            field_name, json_key = field_name.split(':')
+        return field_name, json_key, subparser
+
+    @api.multi
+    def jsonify(self, parser):
+        """ Convert the record according to the parser given
```
simahawk, line 32:
```"""Convert current recordset to JSON data using given parser.

[...] example [...]
put the example output right after
[...]
"""
```

base_jsonify/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,42 @@
+This module add the jsonify method to the ORM. This method take as argument
```
simahawk:
```suggestion
This module adds the a `jsonify` method to every model in the ORM.
It works on the current recordset and requires a single argument `parser` that specifies the fields to extract.
```

base_jsonify/models/models.py
```diff
@@ -0,0 +1,75 @@
...
+        Example of parser:
+            parser = [
+                'name',
+                'number',
+                'create_date',
+                ('partner_id', ['id', 'display_name', 'ref'])
+                ('line_id', ['id', ('product_id', ['name']), 'price_unit'])
+            ]
+
+        In order to be consitent with the odoo api the jsonify method always
```
simahawk: *consistent

base_jsonify/models/ir_exports_line.py
simahawk: it seems that `alias` is used only for the json parser conversion so why don't we call it `json_config` or `json_path` or something similar?

simahawk (changes requested):

simahawk (approved):
@sebastienbeau can we merge? :)

simahawk:
@lmignon @sebastienbeau @sbidoul piiiiiing :smile:

lmignon:
@simahawk merged! Thank you for the review

PR #1380, opened 2018-10-04
[IMP][10.0] Added hierarchical feature for date_range, via parent_id

ntsirintanis:
@gfcapalbo, @lfreeke , @NL66278

### DIFF
--- a/date_range/models/date_range.py
+++ b/date_range/models/date_range.py
@@ -28,8 +28,15 @@ class DateRange(models.Model):
     active = fields.Boolean(
         help="The active field allows you to hide the date range without "
         "removing it.", default=True)
+    parent_type_id = fields.Many2one(
+        related='type_id.parent_type_id',
+        store=True,
+        readonly=True)
     parent_id = fields.Many2one(
-        comodel_name='date.range', string="Parent", index=1)
+        comodel_name='date.range', string="Parent",
+        domain="['|', ('type_id.parent_type_id', '!=', parent_type_id), "
+        "('parent_type_id', '=', False)]",
+        index=1)
 
     _sql_constraints = [
         ('date_range_uniq', 'unique (name,type_id, company_id)',
--- a/date_range/models/date_range.py
+++ b/date_range/models/date_range.py
@@ -35,6 +35,31 @@ class DateRange(models.Model):
         ('date_range_uniq', 'unique (name,type_id, company_id)',
          'A date range must be unique per company !')]
 
+    @api.constrains('parent_id', 'date_start', 'date_end')
+    def _validate_child_range(self):
+        for this in self:
+            if not this.parent_id:
+                continue
+            start = this.parent_id.date_start <= this.date_start
+            end = this.parent_id.date_end >= this.date_end
+            child_range = start and end
+            if not child_range:
+                if (not start) and end:
+                    text = _("Start dates are not compatible (%s < %s)") % (
+                        this.date_start, this.parent_id.date_start)
+                elif (not end) and start:
+                    text = _("End dates are not compatible (%s > %s)") % (
+                        this.date_end, this.parent_id.date_end)
+                else:
+                    text = _("%s range not in %s - %s") % (
+                        this.name,
+                        this.parent_id.date_start,
+                        this.parent_id.date_end,
+                    )
+                raise ValidationError(
+                    _("%s not a subrange of %s: " + text) % (
+                        this.name, this.parent_id.name))
+
     @api.constrains('type_id', 'date_start', 'date_end', 'company_id')
     def _validate_range(self):
         for this in self:
--- a/date_range/models/date_range.py
+++ b/date_range/models/date_range.py
@@ -34,8 +34,6 @@ class DateRange(models.Model):
         readonly=True)
     parent_id = fields.Many2one(
         comodel_name='date.range', string="Parent",
-        domain="['|', ('type_id.parent_type_id', '!=', parent_type_id), "
-        "('parent_type_id', '=', False)]",
         index=1)
 
     _sql_constraints = [
@@ -51,21 +49,35 @@ class DateRange(models.Model):
             end = this.parent_id.date_end >= this.date_end
             child_range = start and end
             if not child_range:
+                text_dict = {
+                    'name': this.name,
+                    'start': this.date_start,
+                    'end': this.date_end,
+                    'parent_name': this.parent_id.name,
+                    'parent_start': this.parent_id.date_start,
+                    'parent_end': this.parent_id.date_end,
+                }
                 if (not start) and end:
-                    text = _("Start dates are not compatible (%s < %s)") % (
-                        this.date_start, this.parent_id.date_start)
+                    text = _(
+                        "Start date %(start)s of %(name)s must be greater than"
+                        " or equal to "
+                        "start date %(parent_start)s of %(parent_name)s"
+                    ) % text_dict
                 elif (not end) and start:
-                    text = _("End dates are not compatible (%s > %s)") % (
-                        this.date_end, this.parent_id.date_end)
+                    text = _(
+                        "End date %(end)s of %(name)s must be smaller than"
+                        " or equal to "
+                        "end date %(parent_end)s of %(parent_name)s"
+                    ) % text_dict
                 else:
-                    text = _("%s range not in %s - %s") % (
-                        this.name,
-                        this.parent_id.date_start,
-                        this.parent_id.date_end,
-                    )
+                    text = _(
+                        "%(name)s range not in "
+                        "%(parent_start)s - %(parent_end)s"
+                    ) % text_dict
                 raise ValidationError(
-                    _("%s not a subrange of %s: " + text) % (
-                        this.name, this.parent_id.name))
+                    _("%(name)s not a subrange of"
+                        " %(parent_name)s: " % text_dict) + text
+                    )
 
     @api.constrains('type_id', 'date_start', 'date_end', 'company_id')
     def _validate_range(self):
--- a/date_range/models/date_range.py
+++ b/date_range/models/date_range.py
@@ -122,14 +122,27 @@ class DateRange(models.Model):
                 (field_name, '<=', self.date_end)]
 
     @api.multi
-    @api.onchange('type_id')
+    @api.onchange('company_id', 'type_id', 'date_start', 'date_end')
     def onchange_type_id(self):
-        return {
-            'domain': {
-                'parent_id': [
-                    '|',
-                    ('parent_type_id', '=', False),
-                    ('type_id', '=', self.type_id.parent_type_id.id),
-                ]
-            }
-        }
+        """The type_id and the dates determine the choices for parent."""
+        domain = []
+        if self.company_id:
+            domain.append(('company_id', '=', self.company_id.id))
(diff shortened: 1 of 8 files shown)


gfcapalbo:
looks good +1

gfcapalbo:
Add the view

NL66278:
@ntsirintanis Maybe I am missing something, but you test a constraint that parent and child should be of a different type. But I do not see the actual constraint.

ntsirintanis:
@NL66278 So there must be a constraint in the sense that the parent date range (a century) should contain all child date ranges (e.g. decades, if parent is a century). What really differentiates a parent from a child  then, is the date.range.type, which is tested here https://github.com/OCA/server-tools/pull/1380/files#diff-cf005ed2488dad0cd047c8a33432db1bR113 . I aim into adding this feature in the most general way possible, allowing the user to consider if further constraints should be imposed on the parent-child relation. Or perhaps I'm missing something too.

lmignon (changes requested):
@ntsirintanis Thank you for this improvement.

@NL66278 @ntsirintanis There is no explicit constraint that parent and child should be of a different type. But by default overlap is not allowed for date range of the same time. That means that it's not possible to use the same constraint for parent and child since parent will always overlap children.... (except if you explicitly allow overlap on the date range type)  

Nevertheless IMO we should add a check that children date ranges  are part of the parent range.

ntsirintanis:
@lmignon @NL66278 a constraint on date range of children with respect to parent has been added. Another thing that we could add, would be a general parent-type, and the appropriate domain for parent_id when creating a new date.range object. Can you please share any insight that you may have on that?

date_range/models/date_range.py
```diff
@@ -28,11 +28,27 @@ def _default_company(self):
...
+        for this in self:
+            if not this.parent_id:
+                continue
+            child_range = (
+                this.parent_id.date_start <= this.date_start and
+                this.parent_id.date_end >= this.date_end
+            )
+            if not child_range:
+                raise ValidationError(
+                    _("%s can not be child of %s") % (
```
NL66278: In your messages you should always be explicit. Otherwise the user will wonder what is wrong. Tell that the new range has a start date before, or an end date after the start of the parent range.

NL66278:
As for the idea to have a certain range type connected to a parent range type:
Add a field parent_type_id to date.range.type.
Add the field as related field to date.range.
Then if you create a range for the child type you can have a domain for the parent type:
```['|', ('parent_id.type_id', '=', parent_type_id), ('parent_type_id'), '=', False)]```
(Or something similar to this).

File: date_range/models/date_range.py (hunk 1 above)
Comment (lmignon) [reply date_range/models/date_range.py]: :+1: with @NL66278

**Comment by lmignon** [comment]
> As for the idea to have a certain range type connected to a parent range type:
> Add a field parent_type_id to date.range.type.
> Add the field as related field to date.range.
> Then if you create a range for the child type you can have a domain for the parent type:
> `['|', ('parent_id.type_id', '=', parent_type_id), ('parent_type_id'), '=', False)]`
> (Or something similar to this).

Good Idea. If you add a related field, don't forget to declare the field as `read_only=True`

File: date_range/models/date_range.py [hunk 2]
```diff
@@ -28,11 +28,44 @@ def _default_company(self):
     active = fields.Boolean(
         help="The active field allows you to hide the date range without "
         "removing it.", default=True)
+    parent_type_id = fields.Many2one(
+        related='type_id.parent_type_id',
+        readonly=True)
```
lmignon, line 34: This field is part of the tree view... may be `store=True`

date_range/models/date_range_type.py
```diff
@@ -22,6 +22,9 @@ def _default_company(self):
     company_id = fields.Many2one(
         comodel_name='res.company', string='Company', index=1,
         default=_default_company)
+    parent_type_id = fields.Many2one(
```
lmignon, line 27: We should declare a constrains to avoid cycle into the parent_type_id relation...

lmignon (changes requested):
Thank you @ntsirintanis for these improvements. Some minor comments...

date_range/models/date_range_type.py
ntsirintanis: Please take a look at the constraint method I implemented that checks if a date.range.type has itself as parent, and returns a validation error. Isn't this what you suggested?

lmignon: @ntsirintanis The implementation is not complete. It will not detect Type A `parent_of` Type B `parent_of` Type C `parent_of` Type A

date_range/models/date_range_type.py
```diff
@@ -22,7 +24,19 @@ def _default_company(self):
...
+        index=1)
 
     _sql_constraints = [
         ('date_range_type_uniq', 'unique (name,company_id)',
          'A date range type must be unique per company !')]
+
+    @api.constrains('parent_type_id')
+    def _validate_parent_type_id(self):
+        for this in self:
+            if not self.parent_type_id:
```
lmignon: /s/self/this

date_range/models/date_range_type.py
```diff
@@ -22,7 +24,19 @@ def _default_company(self):
...
+                continue
+            if self == self.parent_type_id:
```
lmignon: /s/self/this

date_range/models/date_range_type.py
lmignon: @ntsirintanis something like:

```python
    @api.constrains('parent_type_id')
    def _validate_parent_type_id(self):
        for record in self:
            parent = record
            while parent:
                if not parent.parent_type_id:
                    break
                if record.parent_type_id == parent:
                    raise ValidationError()
                parent = parent.parent_type_id  
```

ntsirintanis: @lmignon thanks so much for this tip! We are really really close to finishing this.

ntsirintanis:
@NL66278  , @lfreeke  , @gfcapalbo  please take a look.

date_range/models/date_range_type.py
```diff
@@ -22,7 +24,24 @@ def _default_company(self):
...
+    @api.constrains('parent_type_id')
+    def _validate_parent_type_id(self):
+        for record in self:
+            parent = record
+            while parent:
+                if not parent.parent_type_id:
+                    break
+                if record.parent_type_id == parent:
+                    raise ValidationError(
+                        _("A type parent  can not have a parent:"
```
NL66278: The message is wrong. A type with a parent can have a parent. Like months can belong to years and years can belong to centuries. So the error should be something like: type can not have itself as parent or ancestor. No need to have parent.name and record.name both in the message. The moment the error occurs, they will be the same! Actually no need to have variables at all in the message, the moment the error occurs the user is on the record, so needs no help in identifying the record.

date_range/models/date_range.py
```diff
@@ -28,11 +28,45 @@ def _default_company(self):
...
+    @api.constrains('parent_id', 'date_start', 'date_end')
+    def _validate_child_range(self):
+        for this in self:
+            if not this.parent_id:
+                continue
+            start = this.parent_id.date_start <= this.date_start
+            end = this.parent_id.date_end >= this.date_end
+            child_range = start and end
+            if not child_range:
+                if (not start) and end:
```
NL66278, line 59: These messages will be seen by normal users. The word 'compatible' is really tech lingo. Better say something like:
```
_("Start date %(start) of %(name)s must be greater than or equal to"
" start date %(parent_start)s of %(parent_name)s") %
{'start': this.date_start,
 'name': this.name,
 'parent_name': this.parent_id.name,
 'parent_start': this.parent_id.date_start}
```

date_range/models/date_range.py
```diff
@@ -28,11 +28,45 @@ def _default_company(self):
     active = fields.Boolean(
         help="The active field allows you to hide the date range without "
         "removing it.", default=True)
+    parent_type_id = fields.Many2one(
+        related='type_id.parent_type_id',
+        store=True,
+        readonly=True)
+    parent_id = fields.Many2one(
+        comodel_name='date.range', string="Parent",
+        domain="['|', ('type_id.parent_type_id', '!=', parent_type_id), "
```
NL66278: No need to have the domain as a string. String domains are evaluated client side, if possible it is better to evaluate server side. Also I think the type_id of the parent should be equal (if filled) as the parent_type_id, not unequal.

NL66278 (changes requested):
Still a few things to attent to. See my other comments.

date_range/models/date_range.py
ntsirintanis: Shall this domain be re-introduced in the view, then?

NL66278: Maybe we have a misunderstanding. I did not suggest to remove the domain, I said you can add it as a proper python array, so without the enclosing double quotes.

date_range/views/date_range_view.xml
```diff
@@ -11,7 +11,7 @@
                 <field name="date_end"/>
                 <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
                 <field name="parent_type_id" invisible="1"/>
-                <field name="parent_id"/>
+                <field name="parent_id" on_change="onchange_type_id()"/>
```
lmignon: @ntsirintanis What's the purpose of these "onchange" attributes????

ntsirintanis: @lmignon it serves no purpose, it's a mistake, will be removed right away. Thanks!

lmignon:
@ntsirintanis We're close to the finish line. :smirk: 
IMO A last feature is missing. We should be able to specify a parent when generating a date-range to avoid to have to modify generated items one by one....(ie you create a year range and you use the wizard to generate quarter. In this case you must modify each generated quarter to select the year...)
image

NL66278:
@lmignon Adding a parent is definitely a good idea. Actually in many cases the parent can be determined from the combination of type (and therefore parent type) and start date.

lmignon:
@NL66278 

> Actually in many cases the parent can be determined from the combination of type (and therefore parent type) and start date.

We could go one step further: the parent_id could be computed by the wizard.

@ntsirintanis You've already done a great job. If you don't have time to add this feature, we can add it in the roadmap and finalize this PR.

ntsirintanis:
@lmignon  @NL66278  thanks for the comments and recommendations. I think this is close to what we want for the wizard. I will correct any errors caused by sloppiness, but first I want to make sure that I understood the issue correctly.

NL66278 (approved):
:+1: LGTM and tested

lfreeke:
@sbidoul can you remove the work in progress label please? I think this can be merged now.

lmignon (approved):
Thank you @ntsirintanis 
(code + functional)

lmignon:
I plan to merge this PR after https://github.com/OCA/server-tools/pull/1430 to ease the forward port to 11.0 and 12.0.

lmignon:
@ntsirintanis I close this one since all your changes are included into #1431 where I've rebased your code on top of changes backported from 11.0 and where I'll prepare the next release of date_range.

lmignon:
@ntsirintanis Thank you for your contrib. Merged in https://github.com/OCA/server-tools/commit/b644616228900ea426f9e8709b07d5d664fec863

PR #1381, opened 2018-10-05
[11.0][MIG] record_archiver

ghost:
based on #409 opened against branch **9.0**


record_archiver/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2015 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Records Archiver',
+    'version': '11.0.1.0.0',
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'category': 'misc',
+    'depends': ['base'],
+    'website': 'www.camptocamp.com',
```
simahawk: https://github.com/OCA/server-tools

record_archiver/data/cron.xml
```diff
@@ -0,0 +1,16 @@
...
+  <record forcecreate="True" id="ir_cron_record_archiver" model="ir.cron">
+    <field name="name">Records Archiver</field>
+    <field name="active" eval="True"/>
+    <field name="user_id" ref="base.user_root"/>
+    <field name="interval_number">1</field>
+    <field name="interval_type">months</field>
+    <field name="numbercall">-1</field>
+    <field name="doall" eval="False"/>
+    <field name="model_id" ref="record_archiver.model_record_lifespan"/>
+    <field name="code">model._scheduler_archive_records()</field>
```
simahawk, line 14: :warning: this won't work w/out  `<field name="state">code</field>`

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
+
+# Copyright 2015-2016 Yannick Vaucher (Camptocamp SA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+
+from openerp import api, exceptions, fields, models
```
simahawk: odoo

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
...
+    model_id = fields.Many2one(
+        'ir.model',
+        string='Model',
+        required=True,
+        domain=[('has_an_active_field', '=', True)],
+    )
+    model = fields.Char(
+        related='model_id.model',
+        string='Model Name',
+        store=True,
```
simahawk: readonly

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    months = fields.Integer(
+        required=True,
+        help="Number of month after which the records will be set to inactive "
+             "based on their write date"
+    )
+
+    _sql_constraints = [
+        ('months_gt_0', 'check (months > 0)',
+         "Months must be a value greater than 0"),
```
simahawk, line 45: translate missing

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,114 @@
...
+
+        delta = relativedelta(months=self.months)
+        expiration_date = fields.Datetime.to_string(today - delta)
+
+        domain = self._archive_domain(expiration_date)
+        recs = model.search(domain)
+        if not recs:
+            return
+
+        # use a SQL query to bypass tracking always messages on write for
```
simahawk: if you use `tracking_disable = True` this should not happen

record_archiver/tests/test_archive.py
```diff
@@ -0,0 +1,49 @@
...
+from datetime import datetime, timedelta
+import openerp.tests.common as common
+
+
+class TestArchive(common.TransactionCase):
+
+    def setUp(self):
+        super(TestArchive, self).setUp()
+        Partner = self.env['res.partner']
+        self.partner1 = Partner.create(
```
simahawk: pls rewrite to:
1. use SavepointCase
2. setUp -> setUpClass
3. use tracking_disable on create

record_archiver/tests/test_archive.py
```diff
@@ -0,0 +1,49 @@
...
+        old_date = datetime.now() - timedelta(days=365)
+        self.env.cr.execute(
+            'UPDATE res_partner SET write_date = %s '
+            'WHERE id IN %s', (old_date, tuple([self.partner2.id,
+                                                self.partner3.id]))
+        )
+        self.Lifespan = self.env['record.lifespan']
+        self.model_id = self.ref('base.model_res_partner')
+
+    @common.at_install(False)
```
simahawk: this flags can be set at class level (either as decorators or as class attributes)

simahawk (changes requested):

ghost:
@simahawk addressed

record_archiver/models/record_lifespan.py
ghost: u sure it's needed here?

grindtildeath (approved):
LGTM :+1:

tschanzt (approved):

record_archiver/tests/test_archive.py
```diff
@@ -0,0 +1,53 @@
...
+            'UPDATE res_partner SET write_date = %s '
+            'WHERE id IN %s', (old_date, tuple([cls.partner2.id,
+                                                cls.partner3.id]))
+        )
+        cls.Lifespan = cls.env['record.lifespan']
+        cls.model = cls.env.ref('base.model_res_partner')
+
+    def test_lifespan(self):
+        lifespan = self.Lifespan.create({
+            'model': self.model.id,
```
simahawk: isn't this wrong? we should have `model_id`

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,111 @@
...
+        'ir.model',
+        string='Model',
+        required=True,
+        domain=[('has_an_active_field', '=', True)],
+    )
+    model = fields.Char(
+        related='model_id.model',
+        readonly=True,
+        string='Model Name',
+        store=True,
```
simahawk: passing by a second time... I don't think we need to store this field. 
1st of all we don't use it, 2nd we can easily search records using `('model_id.model', '=', 'foo.bar')`.
Am I wrong?

ghost: Could be that initial purpose was to index `record.lifespan` records by model
nevertheless, I agree, and it looks just fine w/o that field

ghost:
@simahawk should be fine now :wink:

record_archiver/models/record_lifespan.py
simahawk: `months` is editable via UI right? If the user enters a bad value you gonna bump that error on her/his face, right?
If these assumptions are true: pls, translate.

record_archiver/tests/test_active_search.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2015 Guewen Baconnier (Camptocamp SA)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import openerp.tests.common as common
```
simahawk: odoo

record_archiver/models/record_lifespan.py
simahawk: totally forgot that odoo handle this automatically, bah

simahawk (approved):

simahawk:
@arkostyuk pls squash

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @api.multi
+    def _archive_domain(self, expiration_date):
+        """Returns the domain used to find the records to archive.
+
+        Can be inherited to change the archived records for a model.
+        """
+        model = self.env[self.model_id.model]
+        domain = [('write_date', '<', expiration_date)]
+        if 'state' in model.fields_get_keys():
```
etobella, line 101: Mmmh, wouldn't be better to add this kind of configuration directly as a field?

I mean, sometimes, you have states, and none of them have this values. However, If you define it this way, it will be impossible to manage this conditions. You should probably need to manage it with a safe_eval in order to ensure that no issues are raised, but it will be more configurable.

record_archiver/models/record_lifespan.py
```diff
@@ -0,0 +1,109 @@
...
+
+        delta = relativedelta(months=self.months)
+        expiration_date = fields.Datetime.to_string(today - delta)
+
+        domain = self._archive_domain(expiration_date)
+        recs = model.search(domain)
+        if not recs:
+            return
+
+        recs.with_context(tracking_disable=True).write({
```
etobella: Why are you not using toggle_active?

etobella (commented):
Just a few comments, I don't know if they are issues, but I think it is good to ask.

record_archiver/models/record_lifespan.py
ghost: cause this piece is meant to be generic :slightly_smiling_face: 
we're checking the presence [here](https://github.com/arkostyuk/server-tools/blob/ff9e69b9a9f678fe2019ecc3b34aee126a57c8d9/record_archiver/models/record_lifespan.py#L85-L87), but we can't tell whether there is such thing as `toggle_active` method defined at the target model

etobella: `toggle_active` is a generic model function
https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L4248

record_archiver/models/record_lifespan.py
ghost: sorry, I'm not sure I follow: how could `safe_eval` help in this case?

etobella: You might be able to define it directly using other options, safe_eval it's just an option.
However, the interesting part is to define a dynamic domain.

ghost:
@etobella regarding dynamic archive state resolution: check out https://github.com/OCA/server-tools/pull/1381/commits/804297b3d7f5aa9b00910be260afb4e7034134ff

etobella (approved):
LGTM :+1:

yvaucher:
Thanks

PR #1374, opened 2018-10-01
[MIG] module_auto_update: Migrated from v12.0

Angelfentanez:
The `module_auto_update` module has been migrated from 11.0 to 12.0, I performed a functional test to check the behavior and it worked as expected:   

- checksums created the first time the `upgrade_changed_checksum()` method was called:  
image   
- Changes added to the `module_auto_update` module and checksums generated the second time the method was executed:   
image  
image

PR #1382, opened 2018-10-06
[MIG][12.0] module auto update from 11 to 12

sbidoul:
- migration of module_auto_update
- some repo-level maintenance in .travis.yml and README.md

#1368 

Superseedes #1374 to fix broken tests and remove deprecated code.


lmignon (approved):

module_auto_update/hooks.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import SUPERUSER_ID, api
+
+from .models.module import PARAM_INSTALLED_CHECKSUMS
+
+
+def uninstall_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env["ir.config_parameter"].set_param(PARAM_INSTALLED_CHECKSUMS, False)
+    # TODO Remove from here when removing deprecated features
```
astirpe: Shouldn't the entire block of code be removed, starting from this line?

sbidoul: @astirpe you are right. Done.

module_auto_update/tests/test_module.py
```diff
@@ -79,8 +79,7 @@ def test_get_modules_with_changed_checksum(self):
         self.assertFalse(Imm._get_modules_with_changed_checksum())
 
 
-@common.at_install(False)
-@common.post_install(True)
+@odoo.tests.tagged('post_install', '-at_install')
```
astirpe, line 82: Is this fix on tests generic? I mean, is it required in general for every other module? In that case, it would be good to add this fix in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

module_auto_update/readme/USAGE.rst
```diff
@@ -0,0 +1,16 @@
...
+runs an upgrade of all modules, because it has not saved the hashes yet.
+This is by design, priviledging safety. Should this be an issue,
+the method ``_save_installed_checksums`` can be invoked in a situation
+where one is sure all modules on disk are installed and up-to-date in the
+database.
+
+An easy way to invoke this upgrade mechanism is by issuing the following
+in an Odoo shell session::
+
+  env['ir.module.module'].upgrade_changed_checksum()
```
astirpe, line 18: I would prefer to have the python code inside a `.. code-block:: python`:

```
An easy way to invoke this upgrade mechanism is by issuing the following
in an Odoo shell session:

.. code-block:: python

  env['ir.module.module'].upgrade_changed_checksum()
```

module_auto_update/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'version': '12.0.2.0.4',
+    'category': 'Extra Tools',
+    'website': 'https://github.com/OCA/server-tools',
+    'author': 'LasLabs, '
+              'Juan José Scarafía, '
+              'Tecnativa, '
+              'ACSONE SA/NV, '
+              'Odoo Community Association (OCA)',
+    'license': 'LGPL-3',
+    'application': False,
```
astirpe: I would remove the `application` line, since it takes the default value: False

astirpe (changes requested):

module_auto_update/tests/test_module.py
sbidoul: I've not investigated in depth, but at first glance yes, it looks like an Odoo backward compatibility bug.

astirpe (approved):
Thanks! Code review OK.

module_auto_update/tests/test_module.py
sbidoul: Reported to Odoo: https://github.com/odoo/odoo/issues/27471

astirpe: Thank you!

yajo (approved):

sbidoul:
Squashed

PR #1383, opened 2018-10-08
[9.0] [REF] Server Environment: secret keys

ThomasBinsfeld:
This PR:
 - restricts access to server configuration to a new security group (Admin is part of it by default)
 - allows to define which keys of the server configuration must be hidden in other environments than DEV

### DIFF
--- a/server_environment/security/res_groups.xml
+++ b/server_environment/security/res_groups.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+
+<odoo>
+
+    <record model="res.groups" id="has_server_configuration_access">
+        <field name="name">View Server Environment Configuration</field>
+        <field name="users" eval="[(4, ref('base.user_root'))]"/>
+    </record>
+
+</odoo>
--- a/server_environment/serv_config.py
+++ b/server_environment/serv_config.py
@@ -25,7 +25,7 @@ import ConfigParser
 from lxml import etree
 from itertools import chain
 
-from openerp import models, fields
+from openerp import api, models, fields
 from openerp.tools.config import config as system_base_config
 
 from .system_info import get_server_environment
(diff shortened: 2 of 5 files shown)


server_environment/serv_config.py
```diff
@@ -246,10 +246,26 @@ def fields_view_get(self, cr, uid, view_id=None, view_type='form',
...
         return res
 
+    @api.model
+    def _get_secret_keys(self):
+        """
+        This method is intended to be inherited to defined which keywords
+        should be secret.
+        :return: list of secret keywords
+        """
+        return ['passw', 'key', 'secret', 'token']
```
sbidoul: I suggest `_is_secret(self, key)` and doing the `if ... in key` in this methods.

server_environment/security/res_groups.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+
+<odoo>
+
+    <record model="res.groups" id="has_server_configuration_access">
+        <field name="name">Access to server configuration</field>
```
sbidoul: View server environment configuration?

server_environment/serv_config.py
ThomasBinsfeld: Done

server_environment/security/res_groups.xml
ThomasBinsfeld: Done

yvaucher:
@ThomasBinsfeld Thanks for the PR, could you also update the README.rst ? I think there is a mention of accepted as secret key 'passw'.

yvaucher (approved):
Approving code change, just missing an update in README.rst

ThomasBinsfeld:
@yvaucher Thanks for your review. I updated the README.rst.

yvaucher:
@ThomasBinsfeld thanks, as this can be a security issue I will fast track it.

yvaucher:
Forward ported here
- 10.0 https://github.com/OCA/server-tools/pull/1404
- 11.0 https://github.com/OCA/server-env/pull/16

I don't have the time yet to port it to version 12.0 seems it had some more refactors.

PR #1389, opened 2018-10-10
[FIX][8.0] Hard default True, dont break default name_search for models added after this module

thomaspaulb:
…earch for some models

### DIFF
--- a/base_name_search_improved/tests/test_name_search.py
+++ b/base_name_search_improved/tests/test_name_search.py
@@ -28,6 +28,22 @@ class NameSearchCase(TransactionCase):
              'supplier': True,
              'phone': '+351 777 333 555'})
 
+    def test_default_name_search(self):
+        """ Check if default name_search is preserved """
+        # pre-8.0.1.0.2 name_search_use_standard was set to NULL instead of
+        # True for any models initialized after this module, thereby breaking
+        # default name_search functionality (#1389).
+        # - Test for no NULL values
+        self.env.cr.execute(
+            'select count(*) from ir_model where '
+            'name_search_use_standard is null'
+        )
+        self.assertEquals(self.env.cr.fetchone()[0], 0)
+        # - report (auto)installs after base, so test if default name_search
+        # works well for one of its models
+        self.assertEquals(len(self.env['report.paperformat'].name_search(
+            'US Letter', operator='=')), 1)
+
     def test_RelevanceOrderedResults(self):
         """Return results ordered by relevance"""
         res = self.Partner.name_search('555 777')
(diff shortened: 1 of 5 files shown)


thomaspaulb:
@hbrunn Found this with @nzaterrence and made a unit test to prove the point. We ran into this while trying to figure out why importing a CSV file with names for `hr.location` and `sale.order` was not linking to the right values (base_import calls name_search and thus we ended up here). Turns out `name_search_use_standard` was False for a select set of models.

Am a little surprised this new test runs well in Travis, because on my local with a fresh db containing only this module, the paperformat was one of the models proving the point.

Maybe this bug only shows as an interaction with something else on our specific blend of Odoo which makes the defaults of fields not being set properly? In that case, maybe we need to dig deeper.

I have not come around to fixing it yet, ideas are:

1. A pre-installation hook that already creates the database field with all True values, before now exceptions can be made by the rest of modules in the system - although this would not solve things for already existing installations
2. Inverting the value of the field, with a one time pre-migration script that sets all the values to False again before now upgrading dependent modules and making exceptions
3. ...

hbrunn:
the problem is that Odoo [uses sql](https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L377) to create most models. So our ORM defaults won't help anything. Given that Odoo updates existing rows with the default when adding a field, models that existed before installation of the module will have the flag correctly set.

Fortunately, the value is not `False` but `NULL` in such a case, so you can write a migration script that updates nulls to `True` (edit), and calls an init hook which sets an sql level default to `True` on the column.

thomaspaulb:
@hbrunn This is ready for review.

base_name_search_improved/tests/test_name_search.py
```diff
@@ -28,6 +28,22 @@ def setUp(self):
...
+        # - Test for no NULL values
+        self.env.cr.execute(
+            'select count(*) from ir_model where '
+            'name_search_use_standard is null'
+        )
+        self.assertEquals(self.env.cr.fetchone()[0], 0)
+        # - report (auto)installs after base, so test if default name_search
+        # works well for one of its models
+        self.assertEquals(len(self.env['report.paperformat'].name_search(
+            'US Letter', operator='=')), 1)
```
hbrunn: this might break depending of the module available or not. Better just create a new `ir.model` record via SQL and assert the value of the field

thomaspaulb:
@hbrunn fixed tests

hbrunn (approved):

thomaspaulb:
@gcapalbo Could we do a review trade on this one? :-)

fkantelberg (approved):
Code review and tests. :+1:

thomaspaulb:
@pedrobaeza This seems ready

pedrobaeza:
Please squash commits by author.

PR #1401, opened 2018-10-23
[11.0][MIG] mail_cleanup: Migrate to 11.0

ghost:


mail_cleanup/models/fetchmail_server.py
```diff
@@ -1,64 +1,49 @@
...
-                config_vals['purge_days'] = int(config_vals['purge_days'])
-
-            for field in ['cleanup_days', 'purge_days', 'cleanup_folder']:
-                fetchmail[field] = config_vals[field]
+        help="Number of days before removing an e-mail",
+    )
+
+    @property
+    def _server_env_fields(self):
+        base_fields = super()._server_env_fields
```
grindtildeath, line 35: missing `()` ?

mail_cleanup/models/fetchmail_server.py
```diff
@@ -99,25 +86,29 @@ def _purge_fetchmail_server(self, server, imap_server):
...
+        # Called before the fetch, in order to clean up right before
+        # retrieving emails.
         context = self.env.context.copy()
         context['fetchmail_cron_running'] = True
         for server in self:
             _logger.info('start cleaning up emails on %s server %s',
                          server.type, server.name)
-            context.update({'fetchmail_server_id': server.id,
-                            'server_type': server.type})
+            context.update({
```
grindtildeath, line 108: Are all these context update required ? It seems it's never used...

grindtildeath (commented):
Questions

mail_cleanup/models/fetchmail_server.py
ghost: that's a property :wink:

mail_cleanup/models/fetchmail_server.py
ghost: I've found a number of hits on an instance w/ multiple repos in scope, some of them are reading this key from a context, so I'm not sure if it is useless

image

regarding `server_type` - I've found only two hit that set it, but nothing was reading that key

PR #1402, opened 2018-10-24
[11.0] [MIG] [REN] mail_log_message_to_process -> fetchmail_incoming_log

HviorForgeFlow:


pedrobaeza:
This should be called `fetchmail_incoming_log` or similar. Any way, this is something that now Odoo does, isn't it?

HviorForgeFlow:
Yes, odoo throw some logs if email is correctly recieved. This modules aims to launch logs before odoo process incoming email, so even if it has an error you can see preprocessed message in logs.

@jbeficent What do you think about proposed module name?

LoisRForgeFlow (approved):
LTGM :+1:  Thank for adding tests!

MiquelRForgeFlow (approved):
LGTM

PR #1406, opened 2018-10-25
[IMP] Update base_exception views

ivantodorovich:
- Show menu only to Exception Rule Managers
- Use sequence and active widgets on tree view
- Updated form to use sheet


ivantodorovich:
@pedrobaeza this is the another commit

pedrobaeza (approved):

lmignon (approved):

PR #1408, opened 2018-10-26
[12.0] [WIP] database_cleanup: migration to 12.0

mart-e:
As part of the [hacktoberfest](http://hacktoberfest.digitalocean.com/), I am starting to migrate the module database_cleanup to 12.0 (hopefully nobody is working on it yet).

- [x] update version number
- [x] apply new README rules
- [x] squash OCA Transbot
- [x] migration scripts
- [x] date(time)
- [x] label/filter
- [x] SASS
- [x] test


pedrobaeza:
Great! No, nobody is working on this yet, so I assign you in #1368

mart-e:
@pedrobaeza should be good now !

gustavovalverde:
Teste with a module migration use case from V11, which was also done in v12. Almost all seems like working OK, except for a weird behavior which I couldn't reproduce again when Purging all columns; after doing that it started complaining that my user (admin 1) didn't have enough priviliges when clicking on all menus. After a refresh a *403 Forbidden* was raised.

After restoring a snapshot and doing the process again, this did not happened; and everything seems working as expected.

gustavovalverde:
Found the culprit; it's the `password` column from `res.users`. If purged it locks you out of the system. cc: @mart-e @pedrobaeza

paulius-sladkevicius:
I've tested it out on the v12, works well. However I found `password` and `password_crypt` in the list of columns that wants to purge, so @gustavovalverde is right.

pedrobaeza:
#1507 should be included here.

bealdav:
@pedrobaeza  Don't you think we can merge and add a comment in roadmap to cherry pick v11 changes .

gustavovalverde:
@bealdav this shouldn't be merged until the passwords issue is solved (at least): https://github.com/OCA/server-tools/pull/1408#issuecomment-443593754 

P.D.: I'm using it already, with this PR, knowing that I have to remove the password table from being purged. But someone can have a really bad moment not knowing this.

pedrobaeza:
Well, that comment means to not do it ever, and that changes are very important, even possibly solving @gustavovalverde's issue. If you want to retake the migration and add that PR, it's very straight-forward.

pedrobaeza:
@mart-e has just pushed an updated branch. Please check your issue, @gustavovalverde

mart-e:
Hi,

Sorry for the lack of replies. I will try to fix the remaining issue.

pedrobaeza:
Thanks!

mart-e:
I have also picked the commits from #1507, I hope it was the only ones.

gustavovalverde (approved):

gustavovalverde:
@bealdav and @paulius-sladkevicius could you guys check and approve if everything is working as expected?

mart-e:
By the way, the translation files from the previous version are outdated. I have updated en.po at mart-e/server-tools@465c6e4561b1e136842e7 for the example but should I apply it everywhere or let the translation memory do its magic?

pedrobaeza:
@mart-e it seems latest changes have broken Travis

mart-e:
@pedrobaeza got it, the culprit was odoo/odoo@54238c8 which is problematic when combined with a `NOT IN` in SQL
https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_NOT_IN

pedrobaeza:
@mart-e still no luck I'm afraid, but seeing logs I'm not sure why it's failing.

mart-e:
> but seeing logs I'm not sure why it's failing.

Because we had the good idea to introduce fake modules with a `to_buy` flag in 12.0 :roll_eyes: 
However, what I am still not sure is about this one:

https://github.com/mart-e/server-tools/blob/8fb4995ce35ff8080c899c7af14fbd5a2da56ccc/database_cleanup/models/purge_modules.py#L50-L52

when purging modules, you try to uninstall it first but this will produce a warning as the created module is not a real one.
I am not really sure what should be the right behaviour when purging modules.

pedrobaeza:
> However, what I am still not sure is about this one:
> https://github.com/mart-e/server-tools/blob/8fb4995ce35ff8080c899c7af14fbd5a2da56ccc/database_cleanup/models/purge_modules.py#L50-L52
> when purging modules, you try to uninstall it first but this will produce a warning as the created module is not a real one.
> I am not really sure what should be the right behaviour when purging modules.

This was introduced in 33727c5ae709cb3e9d68ec9ba1168a0c83978126 by @hbrunn.

I think the warning is not a problem, but these 2 errors:
```
2019-05-09 15:33:41,645 6556 ERROR openerp_test odoo.modules.registry: Model base.exception.test.purchase has no table.
2019-05-09 15:33:41,645 6556 ERROR openerp_test odoo.modules.registry: Model base.exception.test.purchase.line has no table.
```
They are not in the main branch, so it should be somehow provoked due to this module. However, I have restarted build on main branch to see if something new since last build due to any change upstream:

https://travis-ci.org/OCA/server-tools/jobs/529353188

pedrobaeza:
Main server-tools branch is green after testing again, so definitively something here is provoking that errors.

mart-e:
@hbrunn do you recall the context of d464fb2c596225f91b3968e4643138a449cf92db ? I think I got the same issue with 12.0 migration, incompatibility between tests of base_exception and database_cleanup

hbrunn:
the thing is that some of the actions this module calls involve opening a new cursor and committing transactions, so the database will contain garbage after the tests

.travis.yml
```diff
@@ -17,9 +17,11 @@ env:
 
   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="database_cleanup"
+  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="database_cleanup"
   - TESTS="1" ODOO_REPO="OCA/OCB" MAKEPOT="1"
```
pedrobaeza: I see that we are not testing against "odoo/odoo" and this line is also including both together, which makes this to fail. Please put:
```
- TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="database_cleanup" MAKEPOT="1"
- TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="database_cleanup" MAKEPOT="1"
```

The MAKEPOT part is for synchronizing translations with weblate.

pedrobaeza:
Only one flake8 error to go:
```
database_cleanup/models/purge_modules.py:66:80: E501 line too long (84 > 79 characters)
```
We use strict PEP8 80 cols limit.

And with this and the other change, everything will be finally ready. Thanks for the efforts.

As you are in holidays this week, I'll do it for you for finishing the PR.

pedrobaeza:
Merging after having green. Thanks for the efforts done!

mart-e:
Thanks for the final cleanup, glad this was finally merged !

PR #1409, opened 2018-10-29
[9.0][IMP] auth_totp: Secret key visible in wizard

fkantelberg:
Hello everyone,

this change should allow the usage of the module for more categories of devices to generate the confirmation codes. I'm currently working with an USB stick to generate the confirmation codes and therefore I have to enter the secret key in the configuration. This change shouldn't decrease the security because the secret key is already visible in the QR code.

### DIFF
--- a/auth_totp/__openerp__.py
+++ b/auth_totp/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'MFA Support',
     'summary': 'Allows users to enable MFA and add optional trusted devices',
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.1.0',
     'category': 'Extra Tools',
     'website': 'https://laslabs.com/',
     'author': 'LasLabs, Odoo Community Association (OCA)',
--- a/auth_totp/wizards/res_users_authenticator_create.py
+++ b/auth_totp/wizards/res_users_authenticator_create.py
@@ -29,6 +29,7 @@ class ResUsersAuthenticatorCreate(models.TransientModel):
         index=True,
     )
     secret_key = fields.Char(
+        string='Secret Code',
         default=lambda s: pyotp.random_base32(),
         required=True,
     )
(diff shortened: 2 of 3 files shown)


fkantelberg:
@obulkin @mreficent Would you approve this change as the creator of this module?

OSevangelist:
Hi @obulkin @mreficent may i kindly ask you again to approve this change and merge it? Its just one line and well justified. Thanks and happy X-mas to all.

OSevangelist:
@pedrobaeza would you please have a look on this single line merge request. I was not sure whether it'll be ok if i do it my own (now that i have the the rights to do so) so that's why i am asking. @obulkin and @mreficent didn't reply on this one so far

pedrobaeza:
I don't know about the module, but if you think it's correct, please proceed.

MiquelRForgeFlow:
@lasley @grindtildeath your opinions are welcome

OSevangelist:
ok let's wait on the comments of at least one of either @lasley or @grindtildeath but i think that @lasley is no longer active in the community afaik (as we moved full time to AWS)

grindtildeath (commented):
IMO not showing the secret key on the form has more to do with UX than anything else, but I'm not sure it really makes sense to display this value all the time. :man_shrugging: 

@fkantelberg I didn't test, but if it's only to be used in special cases (i.e on devices not able to scan the QR code) wouldn't it make more sense to set the group `base.group_no_one` on the view so it's easily displayed but in debug mode only ?

BTW, please fix travis and increase module version number :wink:

fkantelberg:
@grindtildeath 
Thanks for the reply. I want to make the change to allow people to use this kind of special devices with odoo too. Therefore I'm not sure if debug mode only is the right choice here. I would propose to change the text in the view to increase the UX because most apps are also allowing to enter the secret key.

I'm absolutely not sure what's wrong with travis because the tests are pointing to other modules.

grindtildeath (approved):
@fkantelberg IMO that's good enough, thank you. :+1:

OSevangelist:
@pedrobaeza can we merge this. i think all reviews and changes have been done accordingly but i would still not use my powers here without your senior consent ;-) Apart from that TravisCI occasionally seems to have errors with other modules as @fkantelberg reported

pedrobaeza:
OK, let's ignore for now CI error as this is an old branch.

PR #1410, opened 2018-10-29
[10.0][IMP] base_exception add rules by domain and improve perfs

hparfr:
Rules can now be written in python or a domain syntax.

Domain based rules can be applied rule by rule instead of record per record in order to achieve better perfs. It reduces the number of queries significantly because writes are done in rules instead of records. 
For small recordset there is no gain. But for large recordset, there is usually less records than rules, so the gain is more visible.

For instance with a modest recordset: 44 products, 4 rules. Before ~ 485 requests, after: ~220 requests.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -41,6 +41,11 @@ class ExceptionRule(models.Model):
     model = fields.Selection(
         selection=[],
         string='Apply on', required=True)
+    type = fields.Selection(
+        selection=[('by_domain', 'By domain'),
+                   ('by_py_code', 'By Python Code')],
+        string='Exception Type', required=True)
+    domain = fields.Char('Domain')
     active = fields.Boolean('Active')
     code = fields.Text(
         'Python Code',
@@ -64,6 +69,19 @@ class ExceptionRule(models.Model):
 #  - uid: current user id
 #  - context: current context
 """)
+
+    @api.multi
+    def _get_domain(self):
+        """ override me to customize domains according exceptions cases """
+        self.ensure_one()
+        return safe_eval(self.domain)
+
+    @api.onchange('type',)
+    def onchange_type(self):
+        if self.type == 'by_domain':
+            self.code = False
+        elif self.type == 'by_py_code':
+            self.domain = False
 
 
 class BaseException(models.AbstractModel):
@@ -141,6 +159,7 @@ class BaseException(models.AbstractModel):
     def detect_exceptions(self):
         """returns the list of exception_ids for all the considered base.exceptions
         """
+        import pdb; pdb.set_trace()
         if not self:
             return []
         exception_obj = self.env['exception.rule']
@@ -195,10 +214,17 @@ class BaseException(models.AbstractModel):
     def _detect_exceptions(self, model_exceptions,
                            sub_exceptions):
         self.ensure_one()
+        import pdb; pdb.set_trace()
         exception_ids = []
         for rule in model_exceptions:
-            if self._rule_eval(rule, self.rule_group, self):
+            if rule.type == 'by_py_code' and self._rule_eval(
+                    rule, self.rule_group, self):
                 exception_ids.append(rule.id)
+            elif rule.type == 'by_domain' and rule.domain:
+                domain = rule._get_domain()
+                domain.append(('id', '=', self.id))
+                if self.search(domain):
+                    exception_ids.append(rule.id)
         if sub_exceptions:
             for obj_line in self._get_lines():
                 for rule in sub_exceptions:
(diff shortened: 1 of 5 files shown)


mourad-ehm (approved):
Code review, not tested

bealdav (approved):
Code review, no test

base_exception/models/base_exception.py
```diff
@@ -63,6 +72,19 @@ class ExceptionRule(models.Model):
 #  - context: current context
 """)
 
+    @api.multi
+    def _get_domain(self):
+        """ override me to customize domains according exceptions cases """
+        self.ensure_one()
+        return safe_eval(self.domain)
```
pedrobaeza, line 81: I don't think the domain should be evaled

base_exception/models/base_exception.py
```diff
@@ -63,6 +72,19 @@ class ExceptionRule(models.Model):
...
+
+    @api.onchange('exception_type',)
+    def onchange_exception_type(self):
+        if self.exception_type == 'by_domain':
```
pedrobaeza: You shouldn't empty fields, or you will lose possible store things. You have to check the type to know which field to use, but not clear it IMO.

base_exception/models/base_exception.py
hparfr: I don't like safe_eval, but it seams to fit with its docstring 

>     This can be used to e.g. evaluate an OpenERP domain expression from an untrusted source.

https://github.com/odoo/odoo/blob/10.0/odoo/tools/safe_eval.py#L267
And there is some use of it across the sources

hparfr:
Hi @pedrobaeza is it ok for you ?

Your remark about empty field has been taken in account.

chafique-delli (approved):
Code review, no test

base_exception/models/base_exception.py
pedrobaeza: Yes, but this is introducing a great security risk, enabling everyone that access these records to do what they want with the DB. You can write a domain like: `self.sudo().password = 'x'` and gain admin access with the specific password you have changed

Always remember the dear BOBBY TABLES:

imagen

Or you find another way of representing safely domains, or for me this is :-1:

hparfr: I agree about how dangerous is to allow user execute code, even in a "safe_eval".
But, 
* I don't see if is more dangerous than allowing user write python code (default version of the module) 
* I don't know a safer way

pedrobaeza: Yes, indeed this is the same hole as with python code. What about this:

* Put a constraint over this field.
* The constraint will add under a savepoint that:

  * There's no `sudo` word inside the text
  * When evaled, it's a valid domain (I'm not sure if there's a method for checking this), but maybe we can perform the search with it to see it.

I know it's limited, but it can be a good first step.

hparfr: It's too much work to fix : `getattr(self, 'odus'[::-1])() `

IMO, the only way to fix it, is to require python or domain rules to be hard coded in class files. It's secure by default. User input rules should be allowed by another optional module. 

It's out of scope of the current PR.

pedrobaeza: OK, at least add this debate in the known issues / roadmap section.

rvalyi:
Notice that this save_eval security issue is present in many places in Odoo. For instance you can use sudo to rewrite a user password everywhere a user can edit a qweb template. So let's say doors are wide open at least if you have hackers within the Odoo advanced users. I am all for fixing this but I think it requires security decisions such as not use an evaling template technology at the Odoo SA level. Otherwise the safe_eval things we will fight for in advanced module will have peanut effect in the overall security.

pedrobaeza:
@rvalyi well, all depends on the given permissions. For QWeb, you need high permissions like being system admin or Website administrator.

I have checked for this one and there's no extra rights when you install `sale_exception`: you still need the extra permission for being explicitly granted to change these rules.

Any way, I think that we should iron this (both domain and eval expression) in a simple way: as the evaluation shouldn't write anything, only to read, then we can enclose the eval in a savepoint, and do rollback after the evaluation, so malicious expressions that changes anything will be discarded. What do you think? We have to evaluate though what happened doing a commit inside the evaled code.

hparfr:
I added a note about safe_eval in the readme.

rvalyi:
Thank @hparfr for the notice.
@pedrobaeza well honestly I think your idea is a good idea but I'm afraid I will not trust safe_eval to be safe even with such counter measures. I think Tryton lead is right when he mention that safe_eval is broken by design https://lwn.net/Articles/574215/

Even if you manage to prevent writing in the DB within the current transaction, it is still likely a motivated hacker find a way to read secret server files (one could read the Odoo session tokens just a fez years ago), monkey patch some code to do harmful stuff in a later request, like write into the db later, try to upload malicious code on the server...

On the contrary I think Tryton were absolutely right to build a domain system (filter) called Pyson that don't use safe_eval at all: https://tryton-readthedocs.readthedocs.io/projects/server/en/latest/topics/pyson.html
Eventually literal_eval can be used at some places.

As for the templating language, I think experienced people were right to design jinja2 or liquid to be "non evaling" since the start vs qweb for which design language ignorance lead to base everything on eval and now safe_eval.

Well this doesn't change the outcome of this PR which IMHO is not making things worse. But I just gave my 2 cts about the save_eval issue.

pedrobaeza:
@rvalyi indeed it's not adding too much, except your typical rants, hehe. The workaround I propose can be easily implemented currently and add an extra layer, although none is perfect of course.

PierrickBrun (approved):

PR #1416, opened 2018-11-03
12.0 mig database cleanup

xhair:


elicoidal:
looks like it is duplicated with #1408

xhair:
> looks like it is duplicated with #1408

Eric , I  change module!! hehe~~

pedrobaeza:
@xhair sorry but there's other migration in progress for this module, so that one has priority. Please review it instead.

PR #1417, opened 2018-11-03
12.0 mig html text

Dike90:


bigteethk:
LGTM

html_text/tests/test_extractor.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2016-2017 Jairo Llopis <jairo.llopis@tecnativa.com>
```
elicoidal, line 1: everywhere update to 2018

elicoidal (commented):
LGTM

elicoidal (approved):

yajo (approved):

html_text/tests/test_extractor.py
pedrobaeza: Not for this contributor. He hasn't done anything in this year.

pedrobaeza:
@Dike90 have you signed the CLA?

html_text/models/ir_fields_converter.py
```diff
@@ -0,0 +1,72 @@
...
+        except (TypeError, etree.XMLSyntaxError, etree.ParserError):
+            if fail:
+                raise
+            else:
+                _logger.exception("Failure parsing this HTML:\n%s",
+                                  html_content)
+                return ""
+
+        # Get words
+        words = u"".join(doc.xpath("//text()")).split()
```
fkantelberg, line 53: This will cause a problem if the html isn't using spaces around tags. e.g. `<a>Test</a><b>Hello</b>`. Best would be to use a space for joining.

fkantelberg (changes requested):
Code review

cristinamartinrod (approved):
LGTM

html_text/models/ir_fields_converter.py
pedrobaeza: @Yajo any thoughts about this?

yajo: I don't agree, because i.e. `This is a <b>test</b>` still has an ending space after "a". On the other hand `My<b>Long</b>Word` is still one word.

pedrobaeza: Then is it correct for all cases?

yajo: Yes, it's ok as it is.

fkantelberg: You get problems when you use something like `<div>`, `<p>` or `<td>` tags. e.g. `<div>Hello,</div><div>Ok</div>`. But of course it's probably a edge case and can be fixed with html formatting.

PR #1421, opened 2018-11-07
[FIX] add and remove modules to/from graph one by one to avoid race conditions

hbrunn:
if we remove a big amount of modules, chances are they won't be added to the graph because we already removed some dependency of it. Removing modules and adding them one by one solves this issue


fkantelberg (approved):
LGTM

PR #1423, opened 2018-11-12
[11.0][MIG] kpi

AdriaGForgeFlow:
Migration from version 9.0 to 11.0


oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+server-backend
```
JordiBForgeFlow: Runbot fails. Instead of this line, try 
`server-backend https://github.com/OCA/server-backend`

AdriaGForgeFlow: Done. Thanks :+1:

JordiBForgeFlow:
@moylop260 @pedrobaeza not sure why runbot fails on this PR.. perhap you have found this issue before?

pedrobaeza:
It seems a glitch. I have restarted the build to see.

pedrobaeza:
Well, same problem... so no clue. @moylop260 your turn, hehe

kpi/README.rst
```diff
@@ -0,0 +1,78 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
etobella: Can you divide the readme following the new rules?
https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

etobella:
It is interesting, but you could change the view from tree to kanban. This way it is easier to view it and you can make it prettier (also, you can add much more options).
If you want a pretty option in order to view the KPI, you can use http://dashing.io. It is interesting to have the last update in order to be sure when the KPI was computed.

pedrobaeza:
There's already a similar dashboard widget in https://github.com/OCA/web/tree/8.0/web_dashboard_tile, although I think it's enough with kanban with proper styling.

JordiBForgeFlow:
@ageficent can you replace the images in kpi/images with ones from v11? The ones shown now are from a prehistoric version :)

etobella:
I have checked the runbot and the problem seems related to the inclusion of server-backend dependency.

I checked that mysqlclient-dev is being installed  on server-backend and it is necessary on the runbot in order to install all python dependencies (mysqlclient). However, mysqlclient-dev is not installed on server-tools runbot.  I don't know how to add this dependency, but I think that it will solve the problem.
Maybe it is enough to add the information to .travis.yml (https://github.com/OCA/server-backend/blob/11.0/.travis.yml#L15)
@ageficent @jbeficent

etobella:
Missing dependency to web

JordiBForgeFlow:
@etobella fixed!

kpi/views/kpi_views.xml
```diff
@@ -0,0 +1,154 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record model="ir.actions.act_window" id="open_kpi_dashboard">
+        <field name="name">KPI Dashboard</field>
+        <field name="res_model">kpi</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">kanban,form</field>
```
etobella, line 133: Just kanban, if you add form, you are may enter to the configuration, but the usual user only needs to access to the kanban

etobella (commented):
LGTM 👍 Just a small comment.

JordiBForgeFlow:
@etobella we discussed about this. But then the user cannot see the history of past kpi values. Which is interesting to a user.

You will notice that as a user without permissions you cannot create/edit/delete anything, and you cannot see the sql code. Just the history and the thresholds.

etobella:
I see, then it's fine for me 😃

etobella (approved):

JordiBForgeFlow:
@dreispt You are listed as author of this module. Are you good with the improvements described?

kpi/security/ir.model.access.csv
```diff
@@ -4,8 +4,8 @@
 "access_kpi_category_user","kpi.category.user","model_kpi_category","base.group_user",1,0,0,0
 "access_kpi_threshold_user","kpi.threshold.user","model_kpi_threshold","base.group_user",1,0,0,0
 "access_kpi_threshold_range_user","kpi.threshold.range.user","model_kpi_threshold_range","base.group_user",1,0,0,0
-"access_kpi_manager","kpi.manager","model_kpi","base.group_user",1,1,1,1
-"access_kpi_category_manager","kpi.category.manager","model_kpi_category","base.group_user",1,1,1,1
-"access_kpi_threshold_manager","kpi.threshold.manager","model_kpi_threshold","base.group_user",1,1,1,1
-"access_kpi_threshold_range_manager","kpi.threshold.range.manager","model_kpi_threshold_range","base.group_user",1,1,1,1
+"access_kpi_manager","kpi.manager","model_kpi","kpi.group_kpi_manager",1,1,1,1
+"access_kpi_category_manager","kpi.category.manager","model_kpi_category","kpi.group_kpi_manager",1,1,1,1
+"access_kpi_threshold_manager","kpi.threshold.manager","model_kpi_threshold","kpi.group_kpi_manager",1,1,1,1
```
AdriaGForgeFlow, line 9: I think there's a missing line for access_kpi_history_manager:
"access_kpi_history_manager","kpi.history.user","model_kpi_history","kpi.group_kpi_manager",1,1,1,1

AdriaGForgeFlow (commented):
Observation

kpi/security/ir.model.access.csv
etobella: Sure!, but maybe the user should be able to create also the history if he can refresh the data.
Just a question, all users should be able to access the KPIs? Or a group should be created?

kpi/__manifest__.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2012 - Now Savoir-faire Linux <https://www.savoirfairelinux.com/>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Key Performance Indicator",
+    "version": "11.0.1.1.0",
```
etobella: That should be 11.0.1.0.0

kpi/models/kpi.py
```diff
@@ -0,0 +1,196 @@
...
+                obj.value = 0
+                obj.color = '#FFFFFF'
+                obj.last_execution = False
+
+    @api.multi
+    def compute_kpi_value(self):
+        for obj in self:
+            kpi_value = 0
+            if obj.kpi_code:
+                if obj.kpi_type == 'local' and is_sql_or_ddl_statement(
```
etobella: Maybe, it should be better to add a hook if someone wants to make another system of Kpis.

etobella: The function should return the dict with value, color and kpi_id.

This way, we could use the python code to compute the color dynamically (compare with Previous year for example)

kpi/models/kpi.py
```diff
@@ -0,0 +1,205 @@
...
+        for obj in self:
+            history_vals = obj._get_kpi_value()
+            history_obj = self.env['kpi.history']
+            history_obj.create(history_vals)
+        return True
+
+    @api.multi
+    def update_next_execution_date(self):
+        for obj in self:
+            if obj.periodicity_uom == 'hour':
```
etobella: It makes sense to add the option of minutes periodicity?

JordiBForgeFlow: Now added!

kpi/models/kpi.py
JordiBForgeFlow: @etobella Done!

JordiBForgeFlow:
@ageficent Can you review the latest commits and also port them to the v12 PR?

kpi/models/kpi.py
```diff
@@ -0,0 +1,208 @@
...
+            }
+        res.update({'kpi_id': self.id})
+        return res
+
+    @api.multi
+    def compute_kpi_value(self):
+        for obj in self:
+            history_vals = obj._get_kpi_value()
+            history_obj = self.env['kpi.history']
+            history_obj.sudo().create(history_vals)
```
etobella, line 189: Maybe should be better to give permissions to the KPI Manager user. Isn't it?

pedrobaeza:
You can move this the same to OCA/reporting-engine.

AdriaGForgeFlow:
> @ageficent Can you review the latest commits and also port them to the v12 PR?

Changes LGTM :+1: and have been ported to v12 here: [OCA/reporting-engine#267](https://github.com/OCA/reporting-engine/pull/267)

kpi/models/kpi.py
```diff
@@ -0,0 +1,208 @@
...
+
+            obj.next_execution_date = new_date.strftime(DATETIME_FORMAT)
+
+        return True
+
+    # Method called by the scheduler
+    @api.model
+    def update_kpi_value(self):
+        filters = [
+            '&',
```
etobella: The query is not correct. It should be:
[
 ('active', '=', True),
'|',
('next_execution_date', '<=', datetime.now().strftime(DATETIME_FORMAT)),
('next_execution_date', '=', False),
]

etobella (changes requested):
The cron update is not working

kpi/models/kpi.py
JordiBForgeFlow: This is now fixed!

JordiBForgeFlow:
@etobella in the last commit we introduce a small migration to new values for select field 'periodicity_uom' so that the values are in line with ir.cron 'interval_type'.

We also introduce a separate cron that will update the 'Update KPI values' frequency if there exist KPIs with a different max frequency.

So, if you have a KPI that is to be run every minute, the cron to update kpis will be updated to be executed every minute for all kpi's.

JordiBForgeFlow:
We found that the option to create a cron that updates the first cron is not good, and we'll revert it.

etobella (approved):
LGTM :+1:

PR #1431, opened 2018-11-19
[10.0] date_range 10.0.3.0.0: Added hierarchical feature for date_range, via parent_id

lmignon:
includes and supersedes #1380 (already approved)
rebased on 10.0 and prepare next release.


lmignon:
Fast tracking since #1380 is already approved

PR #1433, opened 2018-11-26
FIX remove old api parameters in base_exception

bealdav:
ping @pedrobaeza

### DIFF
--- a/base_exception/README.rst
+++ b/base_exception/README.rst
@@ -29,6 +29,11 @@ Bugs are tracked on `GitHub Issues
 check there if your issue has already been reported. If you spotted it first,
 help us smash it by providing detailed and welcomed feedback.
 
+Roadmap
+-------
+Terms used in old api like `pool`, `cr`, `uid` must be removed porting this module in version 12.
+
+
 Images
 ------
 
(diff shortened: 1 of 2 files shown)


mourad-ehm (approved):

bealdav:
Ping @yvaucher a really small review please.

base_exception/README.rst
```diff
@@ -29,6 +29,11 @@ Bugs are tracked on `GitHub Issues
 check there if your issue has already been reported. If you spotted it first,
 help us smash it by providing detailed and welcomed feedback.
 
+Roadmap
+-------
+Terms used in old api like `pool`, `cr`, `uid` must be removed porting this module in version 12.
```
pedrobaeza, line 34: Is this comment still needed?

yvaucher (approved):

base_exception/README.rst
yvaucher: This is about `'cr': self.env.cr,` and questioning the need of it in the `rule_eval_context`.

For backward compatibility we probably want to keep it if very complexe exceptions were created, otherwise it can be deprecated in v 12 as you can reach `cr` and `uid` using `env` which is already in the context.

Issue #1422, opened 2018-11-09
[RFC] Auto Export: powerful tool to export data from any model


ThomasBinsfeld:
# Auto Export

## Description
This new module would introduce a new model allowing to export data of a selected model with a saved export (ir.exports). In the first time, the CSV file would be saved on the filesystem but additional protocols could be added. Export would be triggered from a button on the form and would be asynchronous (queue.job).

The model would be accessible from a new menu and access restricted to a new security group.

## The new auto.export model
* Name
* User processing the export (default: Admin)
* Technical domain for data filtering (optional)
* Saved export template (ir.exports)
* Saved file prefix
* Protocol (default: save on filesystem)
* Destination path on the filesystem

pedrobaeza:
There were something similar on module `module_prototyper`

ThomasBinsfeld:
@pedrobaeza I don't think so. The idea here is to export the same CSV file as you would get by going to the list view of invoices (e.g.) and export them with a selected export template.

But here, the `auto.export` model would allow to filter on the records (with a domain) and save the CSV file to a selected path on the filesystem.

pedrobaeza:
Oh, right, that was XML, not CSV.

rvalyi:
Hello @ThomasBinsfeld. Are you aware of this module https://github.com/OCA/connector-interfaces/tree/10.0/base_import_async ?
Is there no overlap? I pretend to migrate base_import_async to v11 and v12 in the very next days. If you see possible synergies, please suggest design changes to fit your needs as soon as you can. But I tend to see your module could be built atop of base_import_async or use a common design and may be share some parts.

cc @lmignon @sbidoul @sebastienbeau @bealdav

lmignon:
@rvalyi The proposed module is to support scheduled export of data not to import data asynchronously. Therefore there is no overlap.

ThomasBinsfeld:
@rvalyi @lmignon Indeed. I think the only common thing would be `queue.job`.

rvalyi:
My bad, I wasn't aware @ThomasBinsfeld was from Acsone too. Then you know the import module of course. I also just had a meeting with our Akretion customer BTW and it turns out his CSV import will keep being a legacy PHP external service +Odoo RPC in his project, so I'm afraid I'll have no budget to migrate base_import_async to v11/v12 contrarily to what I said earlier.

zakiuu:
[...]

PR #1434, opened 2018-11-27
[9.0] [ADD] Auto Export: powerful tool to export data

ThomasBinsfeld:
From #1422


auto_export/models/auto_export.py
```diff
@@ -0,0 +1,429 @@
...
+    )
+    user_id = fields.Many2one(
+        string="User",
+        help="The export job will be triggered using the selected user's "
+             "access rights. It means that the export result may differ "
+             "depending on the user's groups.",
+        comodel_name='res.users',
+        required=True,
+        ondelete='restrict',
+        default=_get_default_user_id,
```
lmignon: You should use a lambda to call the default method so the method can be override...

ThomasBinsfeld: Done

PR #1440, opened 2018-11-30
[12.0][MIG] kpi

AdriaGForgeFlow:
Migration of kpi module from #1423 to v12.0


pedrobaeza:
Please preserve commit history.

pedrobaeza:
Shouldn't this be in other repo? `kpi` is not a server tool, neither an UX server improvement.

JordiBForgeFlow:
Probably reporting-engine would be better. We can move it there

PR #1441, opened 2018-11-30
[FIX] base_exception: Migration scripts

pedrobaeza:
Current method doesn't work, as this module is not being migrated, but fresh-installed by dependencies of old `sale_exception`, so we should use the pre_init_hook and detect if we should run the migration code according the existence of the table.

cc @Tecnativa


pedrobaeza:
@mourad-ehm please review, as you were the original author.

pedrobaeza:
@bealdav maybe you can review this one.

bealdav (approved):
of course

base_exception/hooks.py
```diff
@@ -0,0 +1,17 @@
...
+
+def pre_init_hook(cr):
+    """If coming from v9 where only `sale_exception` is installed, when
+    auto-installing this one by dependencies, we should take care of
+    renaming previous table before the new model is loaded.
+    """
+    cr.execute("SELECT 1 FROM pg_class WHERE relname = 'sale_exception'")
+    if not cr.fetchone():
+        return  # fresh install, so exiting
+    from openupgradelib import openupgrade
```
mourad-ehm, line 15: As we are not in the context of openupgrade,  penupgradelib is an external lib. Should not be nice to add and external dependency in manifest ? And manage import exception like: 
try:
   from openupgradelib import openupgrade
except ImportError:
    ....
?

pedrobaeza: Not needed. That's why I import after the checking of the column: if you are fresh installing the module, no need of this lib. But if you come from previous version, then you have already openupgradelib due to OpenUpgrade, so also not a problem.

mourad-ehm: Ok

mourad-ehm (approved):

pedrobaeza:
Thanks for reviewing

PR #1454, opened 2018-12-10
Fix avoid_quick_create with multiprocess

guewen:
When starting Odoo in multiprocess, a change to the field
IrModel.avoid_quick_create would be active only on the current worker.
Otherwise, it would be active after a restart.

The proposed change is to signal a registry change, forcing the other
workers to reload a registry and then be aware of the change.

### DIFF
--- a/base_optional_quick_create/models/ir_model.py
+++ b/base_optional_quick_create/models/ir_model.py
@@ -51,4 +51,6 @@ class IrModel(models.Model):
     def write(self, vals):
         res = super(IrModel, self).write(vals)
         self._patch_quick_create()
+        if 'avoid_quick_create' in vals:
+            self.pool.signal_registry_change()
         return res


sbidoul (approved):
Makes sense to me.

This module should be rewritten to become a client-side check.

pedrobaeza:
I have never understood the duality of this, as `web_m2x_options` cover this option as client side part since a lot of time before this one.

leemannd (approved):

lmignon (approved):

sbidoul:
@pedrobaeza a global, per model disabling of quick create is useful, no? m2x options requires to override everywhere it is used.

pedrobaeza:
There's currently some system config parameters for preventing quick create globally (for all models). This is the option I use, but if you want that feature, it's a question of expanding the module for adding the possibility of defining system parameter per model (current one is `web_m2x_options.create`, so for example it can be `web_m2x_options.model_x.create`) for disabling the same, and it's a bit of JS lines for processing this, but I still think this has to be JS side.

sbidoul:
>  this has to be JS side.

I agree

pedrobaeza:
Then do you like my proposition?

lmignon:
We could choice to merge both addons in v12 or v13. Nevertheless, the current one is in production in a lot of places IMO we must therefore maintain it. I want to avoid to have to migrate all my projects using this addon to use an other one...

pedrobaeza:
Of course, we are talking about not maintaining both modules in the future, but also to switch those that you want to this client side module for avoiding problems like this. Any way. I'm merging this one.

sbidoul:
@lmignon yes, of course. This PR is still approved, just chatting here :)

@pedrobaeza your proposition makes sense to me, alternatively (or additionally) web_m2x_option could use the model flag defined by this module.

BTW, is web_m2x_option in good shape? Last time I checked (for v9) it was quite messy with heavy overwrites, and I had the impression it had options that were redundant with native Odoo options.

pedrobaeza:
I would say yes. We are maintaining it since that version, and was rewritten for the new JS framework in v11, but always an external audit can be good :smiley:

PR #1455, opened 2018-12-11
[FIX][auth_dynamic_groups] don't break when there are no dynamic groups

hbrunn:
other commit fixes some linting issues with this module


pedrobaeza:
There's a pylint error, but I suppose we can fast-track this.

hbrunn:
thanks. Author problem fixed in https://github.com/OCA/server-tools/pull/1456

PR #1461, opened 2018-12-17
[12.0][MIG] onchange helper

sbejaoui:


onchange_helper/models/base.py
```diff
@@ -31,7 +31,12 @@ def play_onchanges(self, values, onchange_fields):
         # If self is an empty record we will have an empty value
         if self:
             self.ensure_one()
-            record_values = self._convert_to_write(self.read()[0])
+            record_values = self._convert_to_write(
+                {
+                    field_name: self[field_name]
+                    for field_name, field in self._fields.items()
+                }
+            )
```
sbejaoui: @lmignon , What do you think about this?

onchange_helper/models/base.py
```diff
@@ -31,7 +31,12 @@ def play_onchanges(self, values, onchange_fields):
         # If self is an empty record we will have an empty value
         if self:
             self.ensure_one()
-            record_values = self._convert_to_write(self.read()[0])
```
sbejaoui: @lmignon , What do you think about this?

onchange_helper/models/base.py
```diff
@@ -31,7 +31,12 @@ def play_onchanges(self, values, onchange_fields):
...
+            record_values = self._convert_to_write(
```
sbejaoui, line 45: @lmignon , What do you think about this?

lmignon: @sbejaoui  LGTM and will avoid 1 useless SQL query on each call to the method

onchange_helper/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Florian da Costa <florian.dacosta@akretion.com>
+* Andrea Stirpe <a.stirpe@onestein.nl>
```
lmignon, line 3: @sbejaoui You can add yourself into the contributors list

lmignon (approved):
Thank you @sbejaoui LGTM (Code review only)

onchange_helper/readme/CONTRIBUTORS.rst
sbejaoui: Done

sbejaoui (commented):
@lmignon , I pushed some changes. can you check them please.

lmignon:
@astirpe @sebastienbeau @florian-dacosta Could you take a look to the two last commits PLZ?
https://github.com/OCA/server-tools/pull/1461/commits/a22f10e2ad4df24d585284742491a2edcce9955a
https://github.com/OCA/server-tools/pull/1461/commits/0f4f8bbd74769a782710e66769d202702f3a99b5
These changes looks to me very useful and should be back ported into previous versions IMO.

onchange_helper/models/base.py
```diff
@@ -0,0 +1,69 @@
...
+        new_values = {}
+        for field in onchange_fields:
+            onchange_values = self.onchange(all_values, field, onchange_specs)
+            new_values.update(self._get_new_values(values, onchange_values))
+            all_values.update(new_values)
+
+        return {
+            f: v
+            for f, v in all_values.items()
+            if not self._fields[f].compute
```
lmignon: @sbejaoui You should take a look at https://github.com/OCA/server-tools/pull/1484/files

sbejaoui: done, thanks :)

onchange_helper/models/base.py
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    field_name: self[field_name]
+                    for field_name, field in self._fields.items()
+                }
+            )
+        else:
+            # We get default values, they may be used in onchange
+            record_values = self.default_get(self._fields.keys())
+        for field in self._fields:
+            if field not in all_values:
```
lmignon, line 55: @sbejaoui Do you remember the motivations behind the lines from 55 to 56? IMO we must exclude non stored fields. I observe that if I put all the fields of my model, my non stored fields are recomputed even if not accessed by an onchange...

sbejaoui: We try to get values from existing record or defaults one to fill missing data in `all_values` passed to `play_onchange`.
https://github.com/OCA/server-tools/blob/176916f536793047ed7f9355f2ee56d9872735c9/onchange_helper/models/base.py#L43-L50
IMO, `field_name: self[field_name] ` trigger the compute process.

lmignon: But do we really need all values.... We should let the ORM compute the one required once accessed. In my case, these lines trigger the compute of useless field and cause performance issue.

sbejaoui: I think we can exclude computed fields.

lmignon: I've the same feeling.

bealdav:
Hi, thanks for your work.

I have a use case that doesn't work with this code.

**Float fields with a value of 0.0 is evaluated to False and then is not returned to vals**

It's is not easy to reproduce with unit tests on res.partner.

My case :

I import csv of purchase line with only specified product_id and qty

I want this onchange played called by onchange_product_id()
https://github.com/odoo/odoo/blob/339bf145ced6fb30f10d6d2941cfdaadf594d426/addons/purchase/models/purchase.py#L630

```python
vals = self.env['purchase.order.line'].play_onchanges(vals, ['product_id'])
```
`price_unit` is equal to 0.0, and don't reach vals but is mandatory for purchase lines.

I'll investigate deeper.

lmignon:
@bealdav I'm refactoring this module in V10 to improve the performances... In the same time I'm adding a lot of tests. I've a corner case that doesn't pass the tests (recursive onchange on on2many relation). I'll also add a test for this UC to see if the new code is more robust. https://github.com/OCA/server-tools/pull/1510/files

bealdav:
@lmignon thanks a lot

njeudy (approved):
Just install and test on a copy of a prod database.

Ok for me it works as needed !

ThomasBinsfeld (approved):

rousseldenis:
@OCA/core-maintainers

PR #1469, opened 2018-12-29
Switch AGPL to LGPL license

elicoidal:
Following public and private conversations with the contributors, I propose this PR to switch the license between AGPL to LGPL.
The change is motivated by the fact that when using this kind of module in a project, any kind of module (private/open source) could be a dependency. With AGPL license, this would not be possible, hence the request to switch to LGPL.

@sebalix @tuxintheshell 

@pedrobaeza I am not sure I am doing it right starting from the oldest version v8.0. I might need guidance here.

### DIFF
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -1,6 +1,6 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
-   :alt: License: AGPL-3
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.png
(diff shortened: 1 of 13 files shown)


hbrunn (approved):
still I personally think it's wrong to move to a less freedom enforcing license, but that's the thing with freedom, people pick what they think fits best. All relevant files changed, so the PR is technically fine.

hbrunn:
concerning merge order: Probably this patch won't apply flawlessly to subsequent versions, so I guess you'll have to do some tedious work there. But to deal with the license change swiftly, it'll be optimal if you prepare the PRs now already for all versions so that the person who merges can do it in one go.

pedrobaeza:
I wouldn't change it for versions that you are not interested in. Just switch in the version 10 or 11. And more being v8, where Odoo is still AGPL, and thus, this can't be done.

PR #1470, opened 2018-12-29
Switch AGPL to LGPL license

elicoidal:
Following public and private conversations with the contributors, I propose this PR to switch the license between AGPL to LGPL.
The change is motivated by the fact that when using this kind of module in a project, any kind of module (private/open source) could be a dependency. With AGPL license, this would not be possible, hence the request to switch to LGPL.

@sebalix @tuxintheshell

@pedrobaeza / @hbrunn following up on https://github.com/OCA/server-tools/pull/1469#issuecomment-450472003


sebalix (approved):
Some builds are failing but it's not related to this PR.

hbrunn (approved):

PR #1477, opened 2019-01-10
[FIX] strange behavior of users_ldap_populate

gurneyalex:
when the LDAP search returns values for a user existing in Odoo but
deactivated, the 'deactivate unknown users' feature was silently disabled.

This commit changes the behavior to reactivate the user in Odoo

### DIFF
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -69,13 +69,28 @@ class CompanyLDAP(models.Model):
                     conf['ldap_filter'])
             results = self.get_ldap_entry_dicts(conf)
             for result in results:
+                login = result[1][login_attr][0].lower().strip()
                 user_id = self.with_context(
                     no_reset_password=True
-                ).get_or_create_user(conf, result[1][login_attr][0], result)
-                # this happens if something goes wrong while creating the user
-                # or fetching information from ldap
+                ).get_or_create_user(conf, login, result)
                 if not user_id:
-                    deactivate_unknown = False
+                    # this happens if the user exists but is active = False
+                    # -> fetch the user again and reactivate it
+                    self.env.cr.execute(
+                        "SELECT id FROM res_users "
+                        "WHERE lower(login)=%s",
+                        (login,))
+                    res = self.env.cr.fetchone()
+                    if res:
+                        self.env['res.users'].sudo().browse(
+                            res[0]
+                        ).write(
+                            {'active': True}
+                        )
+                    else:
+                        raise UserError(
+                            'Unable to process user with login %s' % login
+                        )
                 known_user_ids.append(user_id)
 
         users_created = users_model.search_count([]) - users_count_before


users_ldap_populate/models/users_ldap.py
```diff
@@ -69,13 +69,29 @@ def action_populate(self):
...
                     no_reset_password=True
-                ).get_or_create_user(conf, result[1][login_attr][0], result)
-                # this happens if something goes wrong while creating the user
-                # or fetching information from ldap
+                ).get_or_create_user(conf, login, result)
                 if not user_id:
-                    deactivate_unknown = False
+                    # this happens if the user exists but is active = False
+                    # -> fetch the user again and reactivate it
+                    self.env.cr.execute(
```
simahawk, line 79: [get_or_create_user](https://github.com/odoo/odoo/blob/10.0/addons/auth_ldap/models/res_company_ldap.py#L174) in core module uses the same query and should retrieve the user already no matter the state, no?

users_ldap_populate/models/users_ldap.py
```diff
@@ -69,13 +69,29 @@ def action_populate(self):
...
+                        "SELECT id, active FROM res_users "
+                        "WHERE lower(login)=%s",
+                        (login,))
+                    res = self.env.cr.fetchone()
+                    if res:
+                        user_id, active = res
+                        self.env['res.users'].sudo().browse(
```
simahawk, line 85: am I wrong or you are not setting anything here? "active" value stays the same that you've just queried.

users_ldap_populate/models/users_ldap.py
simahawk: nevermind, I missed `if res[1]:` which means `if not active`

users_ldap_populate/models/users_ldap.py
hbrunn: indeed. I'd also suggest to call `get_or_create_user` again in this case, to update the user with the newest values from ldap. And thanks, this was my mistake from v8 I think :-)

gurneyalex: @hbrunn get_or_create_user does not update the user if it exists

gurneyalex: @simahawk fixed, good catch!

hbrunn: ah, that's just something I do: https://github.com/OCA/server-tools/blob/10.0/users_ldap_groups/models/res_company_ldap.py

Ignore then, as this is called periodically it will be called anyways some time later. It might be worthwhile to write the result of `map_ldap_values` on the user for cases where the user came from another company or has other values changed in ldap. Maybe we should do this anyways for all users during populate? (that's more a general question, not a feature request on your PR)

hbrunn (approved):

simahawk (approved):

users_ldap_populate/models/users_ldap.py
simahawk: @hbrunn Ideally we should keep track of "last ldap update date" and compare it w/ ldap write date and update them only if something has changed otherwise we gonna write on all the users all the values every time, no?

hbrunn: indeed. Totally out of scope for this PR, but seems like a useful extension of this module to me

leemannd (approved):
Thanks

damdam-s (approved):

PR #1480, opened 2019-01-15
[10][IMP] base_exception : manage subset of rules or records

hparfr:
Allow the possibility to manage a subset of rules and records
and keep previous (untouched) values

Rules can be a subset of all the rules applied to records.
Self can be a subset of all the records concerned by rules.

This commit ensure detect_exception() to not interfere
with previous exception found which may be still in exception.

Refactor :
 - Add comments and rename some variables
 - Improve perfs by setting values once per record or per rule


florian-dacosta (approved):
Functional tests + technical review
It solves an important bug presently in OCA, where checking a "domain optimized" rule on a simple record will erase the exceptions of all other records concerned by the rule.

base_exception/models/base_exception.py
```diff
@@ -201,32 +201,58 @@ def detect_exceptions(self):
         """List all exception_ids applied on self
         Exception ids are also written on records
         """
+        all_exception_ids = []  # return of the current def()
```
sebastienbeau, line 314: After thinking twice we should refactor this code.

He is the proposition

in base_exception module

```

detect_exception(self):
    rules = self.env['exception.rule'].search([('model', '=', self._name)])
    for rule in rules:
        records = self._dect_exception(rule)
        reverse_field = self._reverse_exception_field()
        if self:
            commons = self and rule[reverse_field]
            to_remove =  commons - records
            to_add = records - commons
            # TODO contruct 3, id, 4, id
            rule.write({reverse_field: [(3, id, _), (4, id, _)]})
        else:
            rule.write({reverse_field: [(6, 0, records.ids)]})


       # ALTERNATIVE TEST PERF
        if self:
            commons = self and rule[reverse_field]
            to_remove =  commons - records
            to_add = records - commons
        else:
            to_remove = rule[reverse_field] - records.ids
            to_add =   records.ids - rule[reverse_field]
       # TODO contruct 3, id, 4, id
       rule.write({reverse_field: [(3, id, _), (4, id, _)]})

```


```
def _detect_exception(self, rule):
    if rule.exception_type == 'by_py_code':
        return self._detect_exception_by_py_code(rule)
    elif rule.exception_type == 'by_domain':
        return self._detect_exception_by_py_domain(rule)

def _detect_exception_by_py_code(self, rule):
    ....

def _detect_exception_by_domain(self, rule):
    ....

 ```

Then in sale_exception module

```
class ExceptionRule
    sale_ids = fields.many2many(...)

class SaleOrder

    def _reverse_exception_field()
        return 'sale_ids'

    def detect_exception()
        super(...)
        self.env['sale.order.line'].detect_all_record_exception()

class SaleOrderLine

    def _reverse_exception_field()
        return 'sale_ids'
    
    def _detect_exception()
        records = super(...)
```

With that solution we can remove the code "optimize" as it always optimize.
We can remove the sub_exception logic

sebastienbeau (changes requested):
We need to refactor the code as it's start to be too much complexe

hparfr:
I strongly disagree your request for change in this PR. This PR fixes a nasty bug so it should be merged for every one to benefit from it. it blocks other PR for dependent modules (because tests are failing waiting this PR).

The refactoring you asks improves the quality of this module, not the functionality nor the outcome of the detection. It should be done in a separate PR.

florian-dacosta:
@hparfr You can close this as the refactore has been merged.
It could be nice if you took a look though

PR #1487, opened 2019-01-25
[8.0] auth_ldap_populate: backport fix of #1477

damdam-s:
same as #1477

### DIFF
--- a/users_ldap_populate/model/users_ldap.py
+++ b/users_ldap_populate/model/users_ldap.py
@@ -22,6 +22,7 @@
 import re
 from openerp.osv import orm, fields  # pylint: disable=W0402
 from openerp import SUPERUSER_ID
+from openerp.tools.translate import _
 import logging
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sebalix (approved):

leemannd (approved):
Thanks

users_ldap_populate/model/users_ldap.py
```diff
@@ -100,12 +101,29 @@ def action_populate(self, cr, uid, ids, context=None):
                     conf['ldap_filter'])
             results = self.get_ldap_entry_dicts(conf)
             for result in results:
+                login = result[1][login_attr][0].lower().strip()
```
tschanzt, line 104: afaik odoo login name is case sensitive wouldn't it be weird to have to log into odoo with lowercase and the rest of the apps with uppercase or am i missing something here? /cc @gurneyalex

tschanzt (commented):
looks good. just one question.

users_ldap_populate/model/users_ldap.py
leemannd: @tschanzt AFAIK ldap is case insensitive, so my guess is that we may receive them in lowercase and want to make sure it stays that way.
EDIT: sidenote, you're totally right on the odoo side

gurneyalex: in LDAP, the login field is typically case insensitive, and people are expecting this -> if we need to mimick this in odoo. 

The code here is stolen from auth_ldap, and we need to keep this consistent.

gurneyalex (approved):

tschanzt (approved):

users_ldap_populate/model/users_ldap.py
tschanzt: @leemannd @gurneyalex thanks for the explanation. Makes sense that way.

PR #1489, opened 2019-01-29
[10.0] base user role multicompany adapting

PierrickBrun:
With this PR, a user role is now company dependent, that means you don't have the same roles depending on the active company.

A user can then switch company and have different roles applied.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -66,9 +66,10 @@ class ResUsersRoleLine(models.Model):
     _description = 'Users associated to a role'
 
     role_id = fields.Many2one(
-        'res.users.role', string=u"Role", ondelete='cascade')
+        'res.users.role', required=True, string=u"Role", ondelete='cascade')
     user_id = fields.Many2one(
-        'res.users', string=u"User", domain=[('id', '!=', SUPERUSER_ID)])
+        'res.users', required=True, string=u"User",
+        domain=[('id', '!=', SUPERUSER_ID)], ondelete='cascade')
     date_from = fields.Date(u"From")
     date_to = fields.Date(u"To")
     is_enabled = fields.Boolean(u"Enabled", compute='_compute_is_enabled')
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -6,8 +6,9 @@
 import datetime
 import logging
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 from odoo import SUPERUSER_ID
+from odoo.exceptions import ValidationError
 
 
 _logger = logging.getLogger(__name__)
@@ -73,6 +74,20 @@ class ResUsersRoleLine(models.Model):
     date_from = fields.Date(u"From")
     date_to = fields.Date(u"To")
     is_enabled = fields.Boolean(u"Enabled", compute='_compute_is_enabled')
+    company_id = fields.Many2one(
+        'res.company', 'Company',
+        default=lambda self: self.env.user.company_id)
+
+    @api.multi
+    @api.constrains('user_id', 'company_id')
+    def _check_company(self):
+        for record in self:
+            if (self.company_id and
+                    self.company_id != self.user_id.company_id and
+                    self.company_id not in self.user_id.company_ids):
+                raise ValidationError(
+                    _('User "{}" does not have access to the company "{}"')
+                    .format(self.user_id.name, self.company_id.name))
 
     @api.multi
     @api.depends('date_from', 'date_to')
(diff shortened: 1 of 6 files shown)


base_user_role/models/user.py
```diff
@@ -58,13 +63,11 @@ def set_groups_from_roles(self):
             if not user.role_line_ids:
                 continue
             group_ids = []
-            role_lines = user.role_line_ids.filtered(
-                lambda rec: rec.is_enabled)
+            role_lines = user._get_enabled_roles()
             for role_line in role_lines:
```
sebastienbeau: You can just set this code on the same line for "role_line in user._get_enabled_roles()"

base_user_role/models/user.py
```diff
@@ -49,6 +49,11 @@ def write(self, vals):
         self.sudo().set_groups_from_roles()
         return res
 
+    def _get_enabled_roles(self):
+        return self.role_line_ids.filtered(
+            lambda rec: rec.is_enabled and
+            rec.company_id == rec.user_id.company_id)
```
sebastienbeau: In case that the role is not restricted to one specific company we should apply it to all company

sebastienbeau (changes requested):
Hi Pierrick can you
- add a python constraint that check that the company on the role line is one of the company available on the customer?
- add python unit test

PierrickBrun:
@sebastienbeau all requested changes are addressed

base_user_role/models/role.py
```diff
@@ -66,12 +67,25 @@ class ResUsersRoleLine(models.Model):
...
+        'res.users', required=True, string=u"User",
+        domain=[('id', '!=', SUPERUSER_ID)], ondelete='cascade')
     date_from = fields.Date(u"From")
     date_to = fields.Date(u"To")
     is_enabled = fields.Boolean(u"Enabled", compute='_compute_is_enabled')
+    company_id = fields.Many2one(
+        'res.company', 'Company',
+        default=lambda self: self.env.user.company_id)
+
+    @api.one
```
sebastienbeau: @api.one is deprecated

base_user_role/models/role.py
```diff
@@ -66,12 +67,25 @@ class ResUsersRoleLine(models.Model):
...
+    @api.constrains('user_id', 'company_id')
+    def _check_company(self):
+        if self.company_id and self.company_id != self.user_id.company_id and(
```
sebastienbeau: no need to make this double check, user.company_ids include user.company_id

base_user_role/models/role.py
```diff
@@ -66,12 +67,25 @@ class ResUsersRoleLine(models.Model):
...
+                self.company_id not in self.user_id.company_ids):
+            raise ValidationError(
+                'User "{}" does not have access to the company "{}"'.format(
```
sebastienbeau: Translation "_(" is missing

bealdav:
ping @sebalix, could you have a look to this PR ?
I think it adds a really good value to the module.

sebalix (approved):
LGTM, nice addition!

base_user_role/models/role.py
PierrickBrun: In a non-multicompany context, company_ids can be empty but company_id isn't

bealdav:
Please @benwillig could you have a look please ?

benwillig (commented):
Nice feature @PierrickBrun.
Code LGTM.

But don't we need record rules for ```res.users.role.line``` as it is company dependent now ?

base_user_role/models/role.py
sebastienbeau: you right

base_user_role/models/role.py
```diff
@@ -66,12 +67,27 @@ class ResUsersRoleLine(models.Model):
...
     is_enabled = fields.Boolean(u"Enabled", compute='_compute_is_enabled')
+    company_id = fields.Many2one(
+        'res.company', 'Company',
+        default=lambda self: self.env.user.company_id)
+
+    @api.multi
+    @api.constrains('user_id', 'company_id')
+    def _check_company(self):
+        for record in self:
+            if (self.company_id and
```
sebastienbeau: replace "self" by "record" as you are in the loop. Current code will not work 

Maybe the code can be simplify (more explicit) with
allowed_companies = record.user_id.company_id + record.user_id.company_ids
if record.company_id and record.company_id not in allowed_companies:
   raise ...

sebastienbeau (changes requested):

PierrickBrun:
@sebastienbeau can you look at it again ?

elicoidal (approved):
LGTM
Excellent feature addition

sebastienbeau (approved):

PR #1493, opened 2019-01-30
Bad syntax in table exist

flotho:


pedrobaeza (approved):
Indeed, I didn't check that. Fast-tracking this.

PR #1494, opened 2019-01-30
[IMP] Allow the administator to forbid passwords that contain the login.

daramousk:
@NL66278

### DIFF
--- a/password_security/demo/res_company.xml
+++ b/password_security/demo/res_company.xml
@@ -1,0 +1,7 @@
+<odoo>
+
+    <record id="base.main_company" model="res.company">
+        <field name="password_no_login">False</field>
+    </record>
+
+</odoo>
--- a/password_security/tests/test_res_users.py
+++ b/password_security/tests/test_res_users.py
@@ -13,35 +13,35 @@ from ..exceptions import PassError
 @post_install(True)
 class TestResUsers(SavepointCase):
 
-    def setUp(cls):
-        super(TestResUsers, cls).setUp()
-        cls.main_comp = cls.env.ref('base.main_company')
+    def setUp(self):
+        super(TestResUsers, self).setUp()
+        self.main_comp = self.env.ref('base.main_company')
         # Modify users as privileged, but non-root user
-        cls.privileged_user = cls.env['res.users'].create({
+        self.privileged_user = self.env['res.users'].create({
             'name': 'Privileged User',
             'login': 'privileged_user@example.com',
-            'company_id': cls.main_comp.id,
+            'company_id': self.main_comp.id,
             'groups_id': [
-                (4, cls.env.ref('base.group_erp_manager').id),
-                (4, cls.env.ref('base.group_partner_manager').id),
-                (4, cls.env.ref('base.group_user').id),
+                (4, self.env.ref('base.group_erp_manager').id),
+                (4, self.env.ref('base.group_partner_manager').id),
+                (4, self.env.ref('base.group_user').id),
             ],
         })
-        cls.privileged_user.email = cls.privileged_user.login
-        cls.login = 'foslabs@example.com'
-        cls.partner_vals = {
+        self.privileged_user.email = self.privileged_user.login
+        self.login = 'foslabs@example.com'
+        self.partner_vals = {
             'name': 'Partner',
             'is_company': False,
-            'email': cls.login,
+            'email': self.login,
         }
-        cls.password = 'asdQWE123$%^'
-        cls.vals = {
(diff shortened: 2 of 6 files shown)


yajo (approved):

NL66278 (approved):
:+1: Tested and LGTM

pedrobaeza:
Why Travis is red?

NL66278:
It seems there are now some tests that fail that assume login's could be the same. Ais is true of course by default for admin and demo in demo. Probably some additional code is needed that suspends these checks for the relevant users, but of course only when testing (not to re-introduce the security hole, this PR intents to stop).

yajo:
You could add a demo data file that disables that check in demo envs.

NL66278:
@daramousk I think the most easy thing to do is either:
1. Make False the default for the option to check passwords, and enable it for the test
2. Add demo data (I think that was meant by @Yajo) to set the option to False for the main company. (And then enable it in your test method). That seems actually the most "clean" method.

yajo:
Yes, I meant that. I think option 2 is better because `True` seems saner as a default value.

password_security/tests/test_res_users.py
```diff
@@ -16,6 +16,7 @@ class TestResUsers(SavepointCase):
     def setUp(cls):
         super(TestResUsers, cls).setUp()
         cls.main_comp = cls.env.ref('base.main_company')
+        cls.main_comp.password_no_login = True
```
NL66278: Put this (suitable adapted, use self instead of cls) in the test method instead of in the setup. In stead of the assertion on the value. No need to reset it in the teardown, because change will automatically be rolled back.

daramousk:
@NL66278 Ok I tried figuring out whether this module breaks the tests on `auth_brute_force`. I installed and run the `auth_brute_force` in an empty database and that test still fails. Am I missing something here?

yajo:
Odoo broke that test, and possibly lots of others, with https://github.com/odoo/odoo/pull/30768. I'm investigating its fix.

NL66278:
@daramousk the module auth_brute_force tests with passwords that contain the login. That should be rectified to be fully compatible with password_security. I had made a fix that I wanted to submit as a PR on your branch, but unfortunately I can not make PR's on your repo anymore. See patch.
```
diff --git a/auth_brute_force/tests/test_brute_force.py b/auth_brute_force/tests/test_brute_force.py
index 2e21b53..a3deb15 100644
--- a/auth_brute_force/tests/test_brute_force.py
+++ b/auth_brute_force/tests/test_brute_force.py
@@ -66,10 +66,10 @@ class BruteForceCase(HttpCase):
         except AttributeError:
             pass
         # Complex password to avoid conflicts with `password_security`
-        self.good_password = "Admin$%02584"
+        self.good_password = "A_dmin$%02584"
         self.data_demo = {
             "login": "demo",
-            "password": "Demo%&/(908409**",
+            "password": "D_emo%&/(908409**",
         }
         with self.cursor() as cr:
             env = self.env(cr)
```

daramousk:
@hbrunn Maybe I could get a review here so that this can be merged as well.

password_security/tests/test_res_users.py
```diff
@@ -13,35 +13,35 @@
 @post_install(True)
 class TestResUsers(SavepointCase):
 
-    def setUp(cls):
-        super(TestResUsers, cls).setUp()
-        cls.main_comp = cls.env.ref('base.main_company')
+    def setUp(self):
```
yajo, line 16: I think that the fix is wrong. Instead of changing all `cls` for `self`, you should change `setUp` for `setUpClass`.

This benefits from `SavepointCase` and speeds up tests.

Another option would be to switch for a `TransactionCase` instead.

yajo (changes requested):

password_security/demo/res_company.xml
```diff
@@ -0,0 +1,7 @@
+<odoo>
+
+    <record id="base.main_company" model="res.company">
+        <field name="password_no_login">False</field>
```
sebalix, line 4:
```suggestion
        <field name="password_no_login" eval="False"/>
```

sebalix (changes requested):

daramousk:
@NL66278 Yes indeed, I have reverted that

daramousk:
@hbrunn Ping for this one that has not been merged yet.

daramousk:
hello @pedrobaeza ,
Perhaps this can be merged?

pedrobaeza:
Please check comment from @Yajo

PR #1495, opened 2019-01-31
[FIX] Use _compute_session_token to invalidate user sessions on auth_session_timeout

daramousk:
I am not sure if that worked after commit https://github.com/OCA/OCB/commit/29c00a56daa377e40f4d28e86a10f105617fd243#diff-5e2f434047c379642786a87195c806f9L1126 since the `check` function is called only if https://github.com/OCA/OCB/blob/9.0/openerp/http.py#L1158

### DIFF
--- a/auth_session_timeout/models/res_users.py
+++ b/auth_session_timeout/models/res_users.py
@@ -102,9 +102,8 @@ class ResUsers(models.Model):
                     'Exception updating session file access/modified times.',
                 )
 
-    @classmethod
-    def check(cls, *args, **kwargs):
-        res = super(ResUsers, cls).check(*args, **kwargs)
+    def _compute_session_token(self, sid):
+        res = super(ResUsers, self)._compute_session_token(sid)
         if http.request:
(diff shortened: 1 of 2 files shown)


daramousk:
@NL66278 A review on this would be awesome, according to @lfreeke this has already been deployed and working

NL66278 (approved):
:+1: Tested by coworker and looks good to me.

daramousk:
@hbrunn Going to need an extra review for this one to be merged.

hbrunn (changes requested):
please adapt tests so that your code is actually hit by them

daramousk:
@hbrunn Done, also took the liberty of adding some docstrings on the functions.

hbrunn (approved):

anassahmed:
The method `_compute_session_token` isn't called with each request like `check` was.

In `11.0` version of `auth_session_timeout`; it uses `_authenticate` in `ir.http` model, which is called with each request (we've been using the `11.0` version with our Odoo `10.0` instance for some time now).

PR #1496, opened 2019-02-01
12.0 mig datetime_formatter

ernestotejeda:
Cc @Tecnativa


yajo (approved):
Perfect!

datetime_formatter/models/res_lang.py
```diff
@@ -105,7 +105,7 @@ def datetime_formatter(self, value, lang=None, template=MODE_DATETIME,
         # Convert str to datetime objects
         if isinstance(value, str):
             try:
-                value = fields.Datetime.from_string(value)
+                value = fields.Datetime.to_datetime(value)
```
pedrobaeza, line 108: Why is this change needed? Now Odoo handles natively datetime

datetime_formatter/models/res_lang.py
```diff
@@ -105,7 +105,7 @@ def datetime_formatter(self, value, lang=None, template=MODE_DATETIME,
...
             except ValueError:
                 # Probably failed due to value being only time
                 value = datetime.strptime(value, DEFAULT_SERVER_TIME_FORMAT)
```
pedrobaeza, line 111: I'm not sure this is still needed.

pedrobaeza (changes requested):
Now that Odoo handles datetimes natively, is this module needed at all?

datetime_formatter/models/res_lang.py
yajo: Look [here](https://github.com/OCA/server-tools/pull/1496/files#diff-3ef157e45961029ea12b27a92d575e19R54). The main usage of this addon, if I don't remember bad, is formatting dates in specific formats for qweb templates, which is sometimes a requirement for website/report design.

In Qweb, you cannot import `fields`, and if an incoming value is a string, you cannot also format it to a datetime object in some contexts, at least not in an easy-to-reuse way.

So, I think that to answer your question, you should answer: when Odoo 12 fetches a date/datetime/time field from database, is it always returned as the corresponding python object? Keep in mind that time fields are usually floats AFAIK, but I haven't had very deep hands-on experience on v12 development yet so as to answer this myself.

pedrobaeza: OK, let's keep it this way as safeguard.

datetime_formatter/models/res_lang.py
pedrobaeza: @ernestotejeda can you check this?

ernestotejeda: Well, I checked it now and it's true Odoo 11 returns date/datetime fields as string in Qweb, but Odoo 12 returns date/datetime fields as the corresponding python date/datetime object. Anyway, we could keep it this way as safeguard, as a tool in case someone gets the value of a field in date/datetime directly from DB using a select query and wants to format that field according to a certain language

pedrobaeza (approved):

PR #1504, opened 2019-02-06
[FIX] field_rrule: respect rrule timezone when serializing

thomaspaulb:
Ran into some issues where serialized and subsequently deserialized rrules suddenly have shifted 1 or 2 hours forward. This normally does not lead to problems, unless eg. your weekly rrule starts on `Monday February 4 2019 00:00 AM` and now suddenly starts at `Monday February 4 2019 02:00 AM`, thereby now missing its first occurrence on `Monday February 4 2019 00:00 AM` and skipping directly to February 11.

### DIFF
--- a/field_rrule/field_rrule.py
+++ b/field_rrule/field_rrule.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 # © 2016 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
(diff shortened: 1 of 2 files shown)


thomaspaulb:
@hbrunn

hbrunn:
are you sure you're not working around bugs you introduced in https://github.com/OCA/hr/pull/472 or possibly the handling of `stable_times` here? All of this *should* come from the tz parameter of the rruleset.

PR #1505, opened 2019-02-08
[12.0][IMP] module_auto_update: UI button

alexey-pelykh:
Adds UI button in developer mode under _Apps_ menu.

### DIFF
--- a/module_auto_update/README.rst
+++ b/module_auto_update/README.rst
@@ -63,7 +63,11 @@ the method ``_save_installed_checksums`` can be invoked in a situation
 where one is sure all modules on disk are installed and up-to-date in the
 database.
 
-An easy way to invoke this upgrade mechanism is by issuing the following
+To invoke the upgrade mechanism, navigate to *Apps* menu and use the
+*Auto-Upgrade Modules* button, available only in developer mode. Restarting
+the Odoo instance is highly recommended to minify risk of any possible issues.
+
+Another easy way to invoke this upgrade mechanism is by issuing the following
 in an Odoo shell session:
 
 .. code-block:: python
--- a/module_auto_update/__manifest__.py
+++ b/module_auto_update/__manifest__.py
@@ -1,10 +1,11 @@
 # Copyright 2017 LasLabs Inc.
-# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 {
     'name': 'Module Auto Update',
     'summary': 'Automatically update Odoo modules',
-    'version': '12.0.2.0.4',
+    'version': '12.0.2.0.5',
     'category': 'Extra Tools',
     'website': 'https://github.com/OCA/server-tools',
     'author': 'LasLabs, '
@@ -17,6 +18,9 @@
     'uninstall_hook': 'uninstall_hook',
     'depends': [
         'base',
+    ],
+    'data': [
+        'views/ir_module_module.xml',
     ],
     'development_status': 'Production/Stable',
     'maintainers': ['sbidoul'],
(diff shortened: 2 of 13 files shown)


module_auto_update/__manifest__.py
```diff
@@ -1,23 +1,28 @@
...
     'summary': 'Automatically update Odoo modules',
-    'version': '12.0.2.0.4',
+    'version': '12.0.2.0.5',
     'category': 'Extra Tools',
     'website': 'https://github.com/OCA/server-tools',
     'author': 'LasLabs, '
               'Juan José Scarafía, '
               'Tecnativa, '
               'ACSONE SA/NV, '
+              'Brainbean Apps, '
```
pedrobaeza: This doesn't bring you as co-author.

module_auto_update/addon_hash.py
```diff
@@ -1,5 +1,5 @@
 # Copyright 2018 ACSONE SA/NV.
-# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
pedrobaeza: License is LGPL in this case. Why you change it?

pedrobaeza (changes requested):

module_auto_update/addon_hash.py
alexey-pelykh: Replace tricked me :) `http` => `https` it should be

module_auto_update/__manifest__.py
alexey-pelykh: Righty

alexey-pelykh:
@pedrobaeza ready for re-review

pedrobaeza (approved):

tafaRU (approved):

yajo (commented):
This is not recommended because, after an upgrade, the only safe thing to do is to reboot Odoo. Also, to upgrade Odoo, the only safe way to do it is trough CLI, or the server could hang on boot if there are i.e. new fields declared.

That being said, this is not actually breaking anything. It's just opening a path for errors. My recommendation is to close this. :-1:

alexey-pelykh:
Nothing would stop user in debug mode to go in module form and Upgrade it manually w/o restarting, which this button automates. IMHO, concern is valid, yet irrelevant for the use-case

alexey-pelykh:
@pedrobaeza shall this be merged?

pedrobaeza:
@Yajo's arguments are strong...

alexey-pelykh:
If to extrapolate those arguments, by default Upgrade button should be disabled on modules UI? What harm is automating clicking on that button using code?

yajo:
I already explained the harm and casted my vote. However, if you get enough thumbs up from others you can still merge, but I cannot vote something I think is wrong, sorry. 🤷🏻‍♂️

alexey-pelykh:
@Yajo absolutely, I totally understand your position!

@pedrobaeza what is the procedure in this case? How many approvals and from whom are needed?

pedrobaeza:
Why don't you add a warning about @Yajo's concerns in the README and go this way?

module_auto_update/README.rst
```diff
@@ -63,7 +63,10 @@ the method ``_save_installed_checksums`` can be invoked in a situation
 where one is sure all modules on disk are installed and up-to-date in the
 database.
 
-An easy way to invoke this upgrade mechanism is by issuing the following
+To invoke the upgrade mechanism, navigate to *Apps* menu and use the
+*Auto-Upgrade Modules* button, available only in developer mode.
+
+Another easy way to invoke this upgrade mechanism is by issuing the following
```
sbidoul, line 70: Please add a note saying that a server restart is highly recommended to take new python code into account.

alexey-pelykh: Done

alexey-pelykh:
@pedrobaeza added, please check if the wording is proper.

PR #1508, opened 2019-02-12
[12.0][MIG] html_image_url_extractor

cristinamartinrod:
cc @Tecnativa


legalsylvain (approved):
Could you set in the readme file, some references to OCA modules that are using this module. (if any). It could be usefull to understand better that tools and how it works.

otherwise, LGTM :+1: Thanks.

yajo (approved):

PR #1468, opened 2018-12-27
[FIX] Fields o2m need to be integer not tuple in return dictionary.

mbcosta:
Try to solve https://github.com/OCA/server-tools/issues/1457 

Unfortunately I didn't find a way to create tests, the dependencies of the module don't install any complex model to test. I appreciate any ideia/tip to find a way to do.

cc @rvalyi @renatonlima

Issue #1457, opened 2018-12-11
[10.0] Module onchange_helper - dict returned have wrong values to o2m fields 


mbcosta:
Hi all
I am working in the PR https://github.com/OCA/account-invoicing/pull/320 , to create invoice from picking and would like to use the module onchange_helper to filled the invoice lines, but I noticed that the dictionary return have wrong values to o2m fields. When called the method:

                vals = invoice_id.play_onchanges(
                    {'picking_id': move.picking_id.id},
                    ['picking_id'],
                )

the dict returned is

{'tax_line_ids': [(5,), (0, 0, {'account_id': (14, u'111200 Tax Received'), 'sequence': 1, 'manual': False, 'name': u'Tax 15.00%', 'currency_id': (3, u'USD'), 'amount': 296.40000000000003, 'account_analytic_id': False, 'tax_id': (1, u'Tax 15.00%')})], 'invoice_line_ids': [(5,), (0, 0, {'uom_id': (1, u'Unit(s)'), 'account_id': (5, u'101120 Stock Interim Account (Received)'), 'analytic_tag_ids': [(5,)], 'move_line_id': (38, u'Test - account_invoice_from_picking/E-COM05: Stock>Customers'), 'price_unit': 247.0, 'sequence': 0, 'price_subtotal': 1976.0, 'company_id': (1, u'YourCompany'), 'currency_id': (3, u'USD'), 'discount': 0.0, 'product_id': (14, u'[E-COM05] Bose Mini Bluetooth Speaker'), 'account_analytic_id': False, 'invoice_line_tax_ids': [(5,), (4, 1)], 'quantity': 8.0, 'picking_id': (28, u'WH/OUT/00014'), 'name': u'Bose Mini Bluetooth Speaker'})], 'picking_id': 28}

The fields o2m need to be id not a list (  'currency_id': (3, u'USD')  ), if you try to write the values in this way you get error message, the dict should be like this 

[...]

PR #1510, opened 2019-02-14
[10.0][IMP] onchange_helper: Improve performance when called on an existing record

lmignon:
Before this change, the inmemory record was populated from values copied from the exisitng record.
It was no predictable to know in advance which values were required to correctly play onchange methods. Therefore all the values defined on the model were copied from the exising record to populate the inmemory record. The side effect of this approach was that in a lot of cases, a lot of useless values was copied leading to performance issue with computed fields.
With this change, we use the current record to call the onchange methods in an onchange context to avoid direct write to the database each time a new value is assigned by an onchange. At the end of the process, we restore the current record to its originale state and return only a dictionary with the fields modified by the onchange.


pedrobaeza:
@lmignon are you aware of the new feature in 12.0 called SSF (Server-Side Form) that imitates a form and play all the onchanges and so on?

https://github.com/odoo/odoo/blob/7ecd750a777a8c021786c0ea152de395daa5e565/odoo/tests/common.py#L1013

It's weird that it's located in tests tools, but it's used on some parts of Odoo code, like in

https://github.com/odoo/odoo/blob/7ecd750a777a8c021786c0ea152de395daa5e565/addons/account_facturx/models/account_invoice.py#L64

I think you can check how they do it for comparing your method. And for v12, this module should be deprecated of wrapped by the other mechanism.

lmignon:
@pedrobaeza The main difference here is that we don't want to rely on a default form. This approach works in every cases. Even if a module add a new onchange method, it will be called since we don't rely on the fom to determine which fields are displayed or could be changed. The method is fully generic and provide an easy way to play onchange methods on the server side without having to know which methods exist or giving a form definition.

bealdav:
@pedrobaeza v11 and before new api depends only on fields. 

With the new Odoo implementation, your onchange behavior depends on view(s). What happens if there are several views on the same object ? Seems dangerous don't you think ?

pedrobaeza:
OK, I would tend to prefer the restricted scope through the proper form (see for example invoice duality), but it's legit if you want to continue with this addon.

pedrobaeza:
@bealdav that's the great thing IMO: you mimic exactly the behavior in UI than in code, as in UI you can have several restrictions that want to apply also in your code (as the mentioned supplier/customer invoice).

The first application I have in my mind right now is to do this for invoice generation in the `contract` module: I want the exact behavior like if I am introducing manually the customer invoice. I don't want any side effect because I have a readonly field or similar in the view, and then my invoice generation through code is not exactly the same as the one done on UI (I had in the past cases like this with payment modes and similar extension modules).

bealdav:
I've just seen view is not mandatory in class Form

sbidoul:
`<rant>`When I was young(er?) I always thought business logic should not rely on the UI... ah well I guess pragmatism needs to beat purity sometimes.`</rant>`

lmignon:
@pedrobaeza 
Personally I find the code is awful and I can't even imagine the performance of such an implementation. (parsing xml, xpath request to access the values).
Moreover, such an approach is not maintainable and the code will no longer work as soon as the basic form is modified.  IMO The logic should not be constrained by data defined into the views otherwise we have a problem of design.

IMO we should propose this implementation to Odoo in place of the Server-Side Form :smirk:

pedrobaeza:
@sbidoul indeed they should be isolated, but at the end, the reality is that one...

@lmignon I'm not saying that one is better, but that it is more homogeneous and "standard" (as it's in Odoo core).

rvalyi:
`"@lmignon I'm not saying that one is better, but that it is more homogeneous and "standard" (as it's in Odoo core)."`

@pedrobaeza you lost @lmignon already, they are already heretic at Acsone as they even touched ShopInvader.

pedrobaeza:
Hehe, OK, then this song is for you:

https://www.youtube.com/watch?v=6ul-cZyuYq4

lmignon:
@sbejaoui This one might interest you.

lmignon:
The errors into travis are not related to the current changes. Travis is failing for the same tests into others PR https://github.com/OCA/server-tools/pull/1489

pedrobaeza:
Yeah, the errors are in `auth_brute_force` due to this change in upstream:

https://github.com/odoo/odoo/pull/30768

We are investigating a solution.

lmignon:
Thank you for the info @pedrobaeza

onchange_helper/models/models.py
```diff
@@ -1,51 +1,170 @@
...
+        field_onchange = {field_name: "1" for field_name in self._fields}
+
+        if not all(name in self._fields for name in onchange_fields):
+            return {}
+
+        # determine subfields for field.convert_to_onchange() below
+        secondary = []
+        subfields = defaultdict(set)
+        for dotname in field_onchange:
+            if "." in dotname:
```
mourad-ehm: To avoid bug of nested one2many's (https://github.com/odoo/odoo/issues/17618)  ex (module sale_commission or  sale_line_option), Its not recommanded to use this code  'if "." in dotname: ...'.
 I syggest to replace it like  the part of code in this commit  I back ported form 12 to fix this bug  https://github.com/akretion/odoo/commit/538e206dd8328088ff42ab2f42a24647f082ea23#diff-9af8b31d921dcccb7bcf230271fd3ef4R5476
https://github.com/OCA/OCB/pull/805

lmignon: @mourad-ehm I'll remove all the code related to subfields since the code doesn't rely on a FormView. We'll never have a dot into a `field_onchange` key ...

mourad-ehm: OK

bealdav:
Hi @lmignon thanks for your work here.

Have you any news on this topic because I have several unmergeable PR v12 waiting for a stable version of onchange_helper

Do you need any resource from akretion on this topic ? for unit test or other ?

lmignon:
@bealdav IMO it's ready. I've added a lot of tests to validate some corner cases....

mbcosta:
Hi @lmignon , thanks for your PR.

I use this module in implementation of the possibility to create Invoice from Picking ( https://github.com/OCA/account-invoicing/pull/320 I run play_onchanges on field product_id to get the dictonary to make a Write ), and by testing I found an issue https://github.com/OCA/server-tools/issues/1457 in actual code, tried to solve here https://github.com/OCA/server-tools/pull/1468, but your code seems far better and solve the initial problem.

However testing your PR I noticed another problem, some fields get values like this:

 'invoice_line_tax_ids': [(5,), (4, 1)]

but in your code it's changed to 

'invoice_line_tax_ids': [(5,), (1, 1, {'amount_type': u'percent', 'description': u'Test Invoice From Picking 20%', 'name': u'Test Invoice From Picking 20%', 'sequence': 1, 'tax_adjustment': False, 'company_id': 1, 'type_tax_use': u'sale', 'refund_account_id': False, 'analytic': False, 'amount': 20.0, 'include_base_amount': False, 'tax_group_id': 1, 'active': True, 'price_include': False, 'account_id': False})]

For me this is an error, because when we have the number 4 it should be used ID instead of change for 1 and include the dictionary

- (1, ID, { values })    update the linked record with id = ID (write *values* on it)
- (4, ID)                link to existing record with id = ID (adds a relationship) 

*  Reference about numbers meaning https://www.odoo.com/forum/help-1/question/how-to-insert-value-to-a-one2many-field-in-table-with-create-method-28714 

In my case of use when this change occurs the field 'invoice_line_tax_ids' is not filled, but maybe it's the correct behaviour by your view or use, if it's the case, please let me know.

pedrobaeza:
FYI, if https://github.com/odoo/odoo/pull/32323 is merged, we probably don't need anymore this kind of things (well, if the model is adapted to the "new way").

lmignon:
@pedrobaeza Indeed if the proposed changes find their way into master, https://github.com/odoo/odoo/pull/32784 will be a big step forward... and all this code would be no more required

onchange_helper/models/models.py
```diff
@@ -1,51 +1,219 @@
...
-            onchange_values = self.onchange(all_values, field, onchange_specs)
-            new_values.update(self._get_new_values(values, onchange_values))
-            all_values.update(new_values)
-
-        return {
-            f: v for f, v in all_values.iteritems()
-            if not (self._fields[f].compute and not self._fields[f].inverse)
-            and (f in values or f in new_values)}
+
+        if not isinstance(onchange_fields, list):
```
apineux: https://github.com/acsone/server-tools/blob/e466d297fa4a524242a7559164072e95e65ce2f0/onchange_helper/models/models.py#L135-L139
From line 135 to 139 the conditions should be inverted. If onchange_fields is None, it will be transformed into [None] and therefore will not be evaluated as False
```suggestion
        if not onchange_fields:
            onchange_fields = values.keys()

        elif not isinstance(onchange_fields, list):
            onchange_fields = [onchange_fields]
```

lmignon: good catch!

sebastienbeau (approved):
Test are green let's merge it !
/ocabot merge major

PR #1517, opened 2019-02-20
[11.0][resource_calendar_schedule_iterations]

JordiBForgeFlow:
When you want to plan a number of days in the past or in the future consiering
a working calendar, Odoo limits the number of days/hours that you can plan
ahead.

In case that you want to plan days forward/backward, it currently restricts
to 100 iterations.

In case that you want to plan hours forward/backward, it currently restricts
to 100 iterations.

This module allows you to increase the iteration limit used in the resource
calendar to schedule days or hours by means of a system parameter defined
by the administrator.

Configuration
=============

#. Go to *Ŝettings* and activate the developer mode.

#. Go to *Settings / Technical / Parameters / System Parameters* and define,
   new values for the parameters *resource.calendar.schedule.days.iteration.limit*
   and/or *resource.calendar.schedule.hours.iteration.limit*


pedrobaeza:
Please avoid plurals on module name.

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
Just added test to this, hoping for a merge when :green_apple: (:crossed_fingers: )

resource_calendar_schedule_iteration/hooks.py
```diff
@@ -0,0 +1,126 @@
...
+            # avoid infinite loops
+            iterations += 1
+
+        return intervals
+
+    if not hasattr(ResourceCalendar, '_schedule_days_original'):
+        ResourceCalendar._schedule_days_original = \
+            ResourceCalendar._schedule_days
+
+    ResourceCalendar._schedule_days = _new_schedule_days
```
lmignon: @jbeficent you should use `BaseModel._patch_method("_schedule_days", _new_schedule_days)` https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L4281

lmignon: @jbeficent you should use `ResourceCalendar._patch_method("_schedule_days", _new_schedule_days)` https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L4281

resource_calendar_schedule_iteration/hooks.py
```diff
@@ -0,0 +1,126 @@
...
+import pytz
+import datetime
+
+from datetime import timedelta
+
+from odoo.addons.resource.models.resource import ResourceCalendar
+from odoo.tools.float_utils import float_compare
+
+
+def to_naive_user_tz(datetime, record):
```
lmignon: @jbeficent Why do you redefine this method already defined into `odoo.addons.resource.models.resource` ?

Why not
`from odoo.addons.resource.models.resource import to_naive_user_tz`

lmignon (changes requested):
@jbeficent a little comment otherwise LGTM. 
Do you plan to propose a change on Odoo master to make this limit configurable by default?

resource_calendar_schedule_iteration/hooks.py
LoisRForgeFlow: Don.e

resource_calendar_schedule_iteration/hooks.py
LoisRForgeFlow: Agree

LoisRForgeFlow:
@lmignon Added your suggestions. Maybe it can be proposed for v12 or master, which one do you think has the more chances to get accepted?

lmignon:
> @lmignon Added your suggestions. Maybe it can be proposed for v12 or master, which one do you think has the more chances to get accepted?

@lreficent IMO only master

lmignon (approved):

resource_calendar_schedule_iteration/README.rst
```diff
@@ -0,0 +1,97 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/server-tools-11-0/server-tools-11-0-resource_calendar_schedule_iteration
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/149/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+When you want to plan a number of days in the past or in the future consiering
```
pedrobaeza: s/consiering/considering

resource_calendar_schedule_iteration/README.rst
```diff
@@ -0,0 +1,97 @@
...
+a working calendar, Odoo limits the number of days/hours that you can plan
+ahead.
+
+In case that you want to plan days forward/backward, it currently restricts
+to 100 iterations.
+
+In case that you want to plan hours forward/backward, it currently restricts
```
pedrobaeza: Repeated sentence

pedrobaeza (changes requested):

LoisRForgeFlow:
@pedrobaeza Attended your comments.

@lmignon Resources were refactored in v12, so I will first migrate this to v12 when this PR gets merged and then propose a current fix to master.

pedrobaeza (approved):

PR #1520, opened 2019-02-21
[10.0][FIX] auth_ldap_populate

gurneyalex:
would not always archive users, esp. when run from a cron

on the way:

* add some logs to help check what's going on
* use sets to remove duplicates
* don't query all ldap configurations if one has said the user must be archived


gurneyalex:
forward port of #1518

CharlineDumontet (approved):

simahawk (approved):
LG

PR #1521, opened 2019-02-26
[11.0][ADD] mail_server_relay_disallowed

AdriaGForgeFlow:
Back port of module mail_server_relay_disallowed from v12.0 (#1463 ) to v11.0

This module fixes the issue of using a mail client where relaying is disallowed.
See:
https://www.odoo.com/es_ES/forum/help-1/question/zoho-com-connection-unexpectedly-closed-51162


cristinamartinrod (approved):

mail_server_relay_disallowed/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Mail server relay disallowed',
+    'version': '11.0.1.0.0',
+    'author':   'Eficent, '
```
pedrobaeza, line 6: Incorrect spacing

pedrobaeza (changes requested):
There's a general main problem: override totally the method. We shouldn't do that in any case. The technique can be:

* Call super (this is a must always). We can do it as first intend, and capture the specific error for doing specific stuff
* Try to pass other argument values that avoid the error?

Issue #1237, opened 2018-04-30
[10.0][auth_session_timeout]Doesn work anymore after Odoo commit da1f153d61d747d9357694382fe04f96c0ca886a


sferket:
The function check in res.users simply doesn't get called anymore after this patch because password is not stored in the session anymore. 

See Odoo commit: 
https://github.com/odoo/odoo/commit/da1f153d61d747d9357694382fe04f96c0ca886a#diff-03542f2676ca18278ea62cb40f5a2261

yelizariev:
This update should be applied for branch 10: https://github.com/OCA/server-auth/commit/f19bd6ef199e284cbbfd002ba0b40ef3bd67b598

lfreeke:
Isn't fixed by this merge request? https://github.com/OCA/server-tools/pull/1495

yelizariev:
It doesn't look so

anassahmed:
The method `_compute_session_token` isn't called with each request like `check` was.

In `11.0` version of `auth_session_timeout`; it uses `_authenticate` in `ir.http` model, which is called with each request (we've been using the `11.0` version with our Odoo `10.0` instance for some time now).

PR #1523, opened 2019-03-02
auth_session_timeout hasn't been working in 10.0 for about 1 year

yelizariev:
See https://github.com/OCA/server-tools/issues/1237

### DIFF
--- a/auth_session_timeout/README.rst
+++ b/auth_session_timeout/README.rst
@@ -32,6 +32,7 @@ Setup the session parameters as described above.
 Known issues / Roadmap
 ======================
 
+* The module doesn't work after `this commit <https://github.com/odoo/odoo/commit/da1f153d61d747d9357694382fe04f96c0ca886a#diff-03542f2676ca18278ea62cb40f5a2261>`__. Backport of `update from 11.0 <https://github.com/OCA/server-auth/commit/f19bd6ef199e284cbbfd002ba0b40ef3bd67b598>`__ has to be done. See `#1237 <https://github.com/OCA/server-tools/issues/1237>`__
 
 Bug Tracker
 ===========


pedrobaeza:
Haven't you got any solution instead of disabling it?

yelizariev:
We just need to apply this: https://github.com/OCA/server-auth/commit/f19bd6ef199e284cbbfd002ba0b40ef3bd67b598
But I don't plan to do it in coming months. Disabling module would clearly mark module as not working and there is chance that someone else do it before me

pedrobaeza:
Then at least add that text to the README or in any place for being identifiable what to do.

yelizariev:
I've added a note to Roadmap section

pedrobaeza:
I'm thinking about those that have the module installed. Isn't better to only not working than having an error reporting missing module?

yelizariev:
I don't really feel a difference. Both are errors

> Isn't better to only not working than having an error reporting missing module?

Existing users who use the old version of odoo will get a problem if they update to latest version of the module at any case (whether we fix module or mark as non-installable).

What is more important for me here is whether we allow non-working module in OCA repo or not.

pedrobaeza:
Why they will have an error? I think they will only have a non time-out system, isn't it?

yelizariev:
Correct. Non time-out is a bug, not an error. User will not get any error messsages. 
With an alternative option, user will get "module is not installable" warning. I'm not sure whether the update will break such odoo database or not.
Ok, I think we can just update README then

pedrobaeza:
Yeah, I prefer that option, as the other one have chances of breaking things.

yelizariev:
I've updated PR

pedrobaeza (approved):
Thanks. Travis error is unrelated. Fast-tracking this

PR #1526, opened 2019-03-07
[FIX] Monkeypatch TestCursor in auth_brute_force's tests

StefanRijnhart:
By implementing nestable savepoints in the TestCursor (courtesy of
Holger Brunn), the failed login attempts are preserved during the test.

See https://github.com/odoo/odoo/pull/20033 for @hbrunn's original proposal

### DIFF
--- a/auth_brute_force/tests/test_brute_force.py
+++ b/auth_brute_force/tests/test_brute_force.py
@@ -46,6 +46,65 @@ def skip_unless_addons_installed(*addons):
     return _wrapper
 
 
+def patch_cursor(func):
+    """ Decorator that patches the current TestCursor for nested savepoint
+    support """
+    def acquire(cursor):
+        cursor._depth += 1
+        cursor._lock.acquire()
+        cursor.execute("SAVEPOINT test_cursor%d" % cursor._depth)
+
+    def release(cursor):
+        cursor.execute("RELEASE SAVEPOINT test_cursor%d" % cursor._depth)
+        cursor._depth -= 1
+        cursor._lock.release()
+
+    def close(cursor):
+        cursor.release()
+
+    def commit(cursor):
+        cursor.execute("RELEASE SAVEPOINT test_cursor%d" % cursor._depth)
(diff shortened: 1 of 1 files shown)


yajo (approved):
Thank you very much for taking the time to debug and fix this.

The main problem for these tests is that, since Odoo merged https://github.com/odoo/odoo/pull/30768, the workaround we had<sup>1</sup>  for using the same cursor as the HTTP requests stopped working.

This means that, unless we add this boliterplate to any test that needs to alter the records exposed to an HTTP request<sup>2</sup>, or needs to review the alterations made to records by an HTTP request, the test will not pass.

Thus, I think it would be good to add the decorator to https://github.com/OCA/oca-decorators and instantiate it here, so we have it available for other v11- tests.

This would be an important improvement, although it's actually not blocking this PR, which could be merged and then later improved when the need arises, so I'm approving here.

Thanks!

---

<sup>1</sup> The workaround is https://github.com/odoo/odoo/issues/12237#issuecomment-254781169 and in Odoo 8-11 we had no other way to manipulate/review the records exposed for a test. In Odoo v12 https://github.com/odoo/odoo/pull/22928 is merged and it allows us to use always the same cursor, removing the need for that workaround.

<sup>2</sup> This involves calls via XMLRPC or front end tests, such as tours. E.g., if you check out in e-commerce and you want to check later in the test that the SO was properly created in the back end.

StefanRijnhart:
Thanks for the pointer! I made this into a decorator rather than integrate it into setUp/tearDown with something like that in mind.

yajo:
I think that we can merge and, if https://github.com/odoo/odoo/pull/20033 is rejected, then move this decorator to oca-decorators.

I wouldn't like to block this PR further as it fixes the main branch.

pedrobaeza (approved):
Let's greenify this

pedrobaeza:
Is this applicable as is in v11?

yajo:
Possibly, but better to open a PR just in case.

StefanRijnhart:
Thanks guys. Let me port to server-auth/11.0

PR #1529, opened 2019-03-14
[12.0][MIG] base_view_inheritance_extension: Migration to 12.0

rvalyi:
standard migration: nothing to be done: same code as version 11 (except version number)

cc @hbrunn @NL66278 @pedrobaeza @sergio-teruel @alexis-via


alexis-via:
It seems you forgot the update of the version number.

rvalyi:
@alexis-via sorry. I now amended the commit with the right version

rvalyi:
@alexis-via sorry, I did it wrong as I was in a meeting with Renato. I should not have changed the version in the Transiflex pot commit, so I undo that and I added a new commit for it https://github.com/OCA/server-tools/pull/1529/commits/61a83bb372dc83d6344993034ad0ee6123846d53
Should be good now, sorry again.

hbrunn:
v12 has its own version of `position="move"`, cf https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/ir_ui_view.py#L650 ff and https://www.odoo.com/documentation/12.0/reference/views.html#inheritance-specs (note at the end).

It seems to me both can coexist, and I find the way they have chosen a little bit clumsy, but it serves the same purpose. Should we drop support for `position="move"` here for consistency or keep it for simpler migrations of modules using it and a imho nicer syntax?

NL66278:
@hbrunn I would people encourage to use the standard Odoo way as much as possible. So keep the position="move" for now, but clearly mark/document it as deprecated.

pedrobaeza:
But are they going to conflict having the same name?

hbrunn:
no, because for Odoo standard, the attribute is supposed to be added to a child of the element you match, in our version it's on the element you match

pedrobaeza:
Pretty confusing, isn't it? I prefer to remove it...

bealdav:
Any decision here ?

tbaden:
I would prefer to use odoo core behaviour, but just my opinion. nothing against this feature, but to have both will be quite confusing.

astirpe:
I would prefer the position="move" removed from this module. Just a preference, not blocking for this.

ernestotejeda:
@rvalyi, what conclusion do we get here? What can we do to conclude with this migration?
maybe drop support for position="move" here is good idea, what do you think? ;-)

hbrunn:
note that since version 9, also the `list_{add,remove}` parts can be done with Odoo core: https://github.com/OCA/OCB/commit/f962b0f62831614137a9e1e509803bfa39be1110 (actually in a nicer way because you can also select a separator) - so if we deprecate the moving, we also should deprecate that

ernestotejeda:
> note that since version 9, also the `list_{add,remove}` parts can be done with Odoo core: [OCA/OCB@f962b0f](https://github.com/OCA/OCB/commit/f962b0f62831614137a9e1e509803bfa39be1110) (actually in a nicer way because you can also select a separator) - so if we deprecate the moving, we also should deprecate that

Yes, i agree with you @hbrunn  :ok_hand:

tsabi:
Hi, when will you merge this? A lot of module depends from this.
Thanks, Csaba

base_view_inheritance_extension/__init__.py
```diff
@@ -0,0 +1,2 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```
NL66278: Make license lines https:

NL66278 (changes requested):
I agree that all the methods that are now in standard Odoo should be removed. Either completely, or at least by giving a Deprecated message when still used (to be turned into an Exception in 13.0).

rvalyi:
Hello folks, I dont really use this module myself. So if somebody wants to redo the migration in another PR/another way please feel free to go ahead as I cannot dedicate time to this one these days unfortunately...

alexis-via:
I continued the work of @rvalyi and removed the features which are now native. The readme of the module has been updated. Tests works fine. I also tested it with my PR https://github.com/OCA/account-invoicing/pull/570 which makes use of this module in v12 and it works well. So it's ready to merge now.

bealdav (approved):
LGTM

base_view_inheritance_extension/__init__.py
alexis-via: done

alexis-via:
As said by Pedro in the PR for v13, the list_add and list_remove feature is NOT native. So I cancelled my previous commit and push a new commit with only the removal of the move feature.
I took into account all the remarks of the reviewers. So it's now ready to merge.

pedrobaeza (approved):

PR #1519, opened 2019-02-21
[8.0][FIX] auth ldap populate

gurneyalex:
backport of #1518 to 8.0

PR #1531, opened 2019-03-22
[FIX] users_ldap_populate: Wrong attribute usage

yajo:
The variable is a dict, and the desired value must be accessed via key, not attribute.

This fixes a bug introduced in https://github.com/OCA/server-tools/pull/1519.

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as straight-forward.

PR #1539, opened 2019-04-02
[10.0][FIX] fuzzy search when accents are used in the search values

ecino:
When we use fuzzy search and user inputs a value with accented characters, the search fails because of encoding issues in the strings. This should fix the issue.


pedrobaeza (approved):

tbaden:
thanks for the work 👍 
Code lgtm

tbaden (approved):
thanks for the work 👍
Code lgtm

PR #1540, opened 2019-04-02
[10.0] [FIX] auth_session_timeout: use `_authenticate` from `ir.http`

anassahmed:
This commit kinda uses the logic in `11.0` module counterpart.

I've done some initial tests and looks like it works, but I will do more extensive tests in the following days to ensure it works just fine.

I've modified the tests, but I'm unsatisfied with the current format (in `11.0` there's no test for `ir.http` at all). also, `patch.multiple` is unfamiliar to me, so I did my own `patch` for now.

If you have any suggestions, please comment and leave your review.

### DIFF
--- a/auth_session_timeout/models/__init__.py
+++ b/auth_session_timeout/models/__init__.py
@@ -4,3 +4,4 @@
 
 from . import res_users
 from . import ir_config_parameter
+from . import ir_http
--- a/auth_session_timeout/models/ir_http.py
+++ b/auth_session_timeout/models/ir_http.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# (c) 2019 Anass Ahmed, Katherine Zaoral
+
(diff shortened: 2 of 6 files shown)


anassahmed:
The last modifications on this pull request are for the following scenario:
1. A user logins into an active session.
2. The user closes the browser (or tab) while the session was active.
3. The user re-visits the Odoo instance again after the timeout has been reached.
4. The initial requests to `/` or `/web` for some reason don't pass the `request.env.user` but `request.session.uid` is passed. Then there's a mysterious code somewhere (in Odoo or Werkzeug) that modifies or updates the modification time of the session file, before our code is accessed with a defined `request.env.user` which invalidates the deadline condition and the user is authenticated again despite the timeout!!

If you have a better solution for this issue, I'm all ears.

Thanks.

anassahmed:
Is there any other steps I've to take to get this PR reviewed and merged?

PR #1543, opened 2019-04-03
[9.0] [REF] Keychain: add index on keychain.account

ThomasBinsfeld:
On my database, the following query took ~11ms before this change and ~0.3ms now.

``` sql
SELECT "keychain_account".id FROM "keychain_account" WHERE ((("keychain_account"."namespace" = ?) AND ("keychain_account"."technical_name" = ?)) AND (("keychain_account"."environment" in (?)) OR "keychain_account"."environment" IS NULL)) ORDER BY "keychain_account"."id";
```

Forward port to 10.0 : https://github.com/OCA/server-tools/pull/1748
Forward port to 11.0 : https://github.com/OCA/server-auth/pull/153

### DIFF
--- a/keychain/models/keychain.py
+++ b/keychain/models/keychain.py
@@ -45,7 +45,8 @@ class KeychainAccount(models.Model):
     name = fields.Char(required=True, help="Humain readable label")
     technical_name = fields.Char(
         required=True,
-        help="Technical name. Must be unique")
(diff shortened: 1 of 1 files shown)


sbidoul (commented):
I'd say the index on technical_name is enough, since the number of namespaces and environment should be small? Unless there are queries on namespace/environment without specifying the technical name?

PR #1545, opened 2019-04-03
[10.0][FIX] base_techical_features: Don't load all users if not needed

rousseldenis:
Odoo ORM will load all users contained in loaded group in cache.
That has huge performances impact if there is a lot of users
in that group (especially in computed fields).


pedrobaeza (approved):

LoisRForgeFlow (approved):
LGTM

Issue #1546, opened 2019-04-04
[8.0] ImportError: No module named server_environment_files


sergiocorato:
The last commits on repo re-create the issue on server_environment_files import, the problem is not present at 5ec40c3e47e8ab0fa090b4f51237a8673ce11d85

Error is ```ImportError: No module named server_environment_files``` as #89

sergiocorato:
@pedrobaeza I didn't find a fast solution, some tips?

pedrobaeza:
Sorry, but no, but why you don't simply put needed addons and not full directory to your addons path?

thomaspaulb:
Am also getting the same issue here, will submit a fix.

thomaspaulb:
@pedrobaeza For us that is difficult, we have not moved to git-aggregator yet but are still on Anybox buildout

pedrobaeza:
Ok, @thomaspaulb that hurts to be heard :stuck_out_tongue_closed_eyes:

Move directly to Doodba (https://github.com/Tecnativa/doodba). You won't regret!

thomaspaulb:
Yes yes, will think about it, but you know how the README says it's highly opinionated, so it affects my stubborn character to suddenly move to it and get Docker as a free present :-)

pedrobaeza:
Well, opinionated doesn't mean inflexible. But the opinionated part is our 8 years accumulated experience implementing Odoo/OpenERP, so you will appreciate the design decisions we have taken.

sergiocorato:
@pedrobaeza very good, this is for old installation that I don't touch

pedrobaeza:
Yes, it was just an off-topic/advertisement, hehe

thomaspaulb:
I appreciate all of it, if there was an "odba" version without Docker i would have probably already switched. Anyway, we are polluting this thread, you are utterly right, but for legacy reasons maybe OCA should also try to support including the whole `server_tools` repo.

Funnily, I just realized [one of your own commits](https://github.com/OCA/server-tools/commit/f0b86820008c8c3bdb2c177eb0a8bb9b3a8c9e11) did the same for 9.0.

So i backported that one and its parent to 8.0: https://github.com/OCA/server-tools/pull/1548

pedrobaeza:
[...]

PR #1542, opened 2019-04-03
[FIX] logging_json ImportError

thomaspaulb:
Fixes #1541

PR #1547, opened 2019-04-04
Revert "Merge pull request #1542 from sunflowerit/8.0-logging_json-do…

sergiocorato:
…nt-bork-on-missing-package"

This reverts commit 4e22ed02f6ba36e070df043138bab3f934f99981, reversing
changes made to eacac19f0fe31d6b3d5755aa2724136b3b2f6de5.
#1546

### DIFF
--- a/logging_json/json_log.py
+++ b/logging_json/json_log.py
@@ -17,9 +17,9 @@ from openerp.netsvc import ColoredFormatter
 _logger = logging.getLogger(__name__)
 
 try:
-    from pythonjsonlogger.jsonlogger import JsonFormatter, RESERVED_ATTRS
+    from pythonjsonlogger import jsonlogger
 except ImportError:
-    JsonFormatter = object
+    jsonlogger = None  # noqa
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Sorry, but this has nothing to do with the issue you mention, and the fix is correct.

sergiocorato:
> Sorry, but this has nothing to do with the issue you mention, and the fix is correct.

I don't find a relation too, but removing the fix, the bug disappear...

pedrobaeza:
Then the problem is in the other module that is not handling correctly the situation. Probably the previous error makes this one to not appear, but the correct fix is to solve the problem in `server_environment` module.

sergiocorato:
:+1: but i don't use that module

pedrobaeza:
You haven't put what is the error in the issue, so it's difficult then to deduce the problem.

sergiocorato:
> You haven't put what is the error in the issue, so it's difficult then to deduce the problem.

sorry, I wrote it in the title, I put more info now

PR #1392, opened 2018-10-14
12.0 mig auditlog

GoodERPJeff:

PR #1551, opened 2019-04-05
12.0 mig auditlog

shepilov-vladislav:


pedrobaeza:
There's already other migration in #1392. Please review that one instead.

PR #1564, opened 2019-04-24
[10.0] [FIX] users_ldap_populate special characters

r-hede:
Hi,

Currently, you can not use special characters in the username, password, or LDAP base.

To fix that, you just have to modify the lines 126 to 130 of the file _users_ldap.py_ like this:
```
conn.simple_bind_s(conf['ldap_binddn'].encode('utf-8') or '',
                   conf['ldap_password'].encode('utf-8') or '')
results = conn.search_st(conf['ldap_base'].encode('utf-8'), ldap.SCOPE_SUBTREE,
                         ldap_filter.encode('utf8'), None,
                         timeout=60)
```

Best regards,

Robin Hède.

### DIFF
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -125,10 +125,17 @@ class CompanyLDAP(models.Model):
         conn = self.connect(conf)
         ldap_password = conf['ldap_password'] or ''
         ldap_binddn = conf['ldap_binddn'] or ''
-        conn.simple_bind_s(ldap_binddn.encode('utf-8'), ldap_password.encode('utf-8'))
-        results = conn.search_st(conf['ldap_base'].encode('utf-8'), ldap.SCOPE_SUBTREE,
-                                 ldap_filter.encode('utf8'), None,
-                                 timeout=timeout)
+        conn.simple_bind_s(
+            ldap_binddn.encode('utf-8'),
+            ldap_password.encode('utf-8')
+        )
+        results = conn.search_st(
+            conf['ldap_base'].encode('utf-8'),
+            ldap.SCOPE_SUBTREE,
+            ldap_filter.encode('utf8'),
+            None,
+            timeout=timeout
+        )
         conn.unbind()
         return results
 
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -123,9 +123,9 @@ class CompanyLDAP(models.Model):
         """
         ldap_filter = filter_format(conf['ldap_filter'] % user_name, ())
         conn = self.connect(conf)
-        conn.simple_bind_s(conf['ldap_binddn'] or '',
-                           conf['ldap_password'] or '')
-        results = conn.search_st(conf['ldap_base'], ldap.SCOPE_SUBTREE,
+        conn.simple_bind_s(conf['ldap_binddn'].encode('utf-8') or '',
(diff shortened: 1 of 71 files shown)


NL66278:
Please fix this error in travis (not caused by your PR, but apparently existing issue):
```
************* Module users_ldap_populate.models.users_ldap
users_ldap_populate/models/users_ldap.py:90: [C8107(translation-required), CompanyLDAP.action_populate] String parameter on "UserError" requires translation. Use _('Unable to process user with login %s')
```

NL66278 (changes requested):
:+1: LGTM, but needs one extra change to satisfy travis.

users_ldap_populate/models/users_ldap.py
```diff
@@ -123,9 +123,10 @@ def get_ldap_entry_dicts(self, conf, user_name='*', timeout=60):
         """
         ldap_filter = filter_format(conf['ldap_filter'] % user_name, ())
         conn = self.connect(conf)
-        conn.simple_bind_s(conf['ldap_binddn'] or '',
-                           conf['ldap_password'] or '')
-        results = conn.search_st(conf['ldap_base'], ldap.SCOPE_SUBTREE,
+        ldap_password = conf['ldap_password'] or ''
+        ldap_binddn = conf['ldap_binddn'] or ''
+        conn.simple_bind_s(ldap_binddn.encode('utf-8'), ldap_password.encode('utf-8'))
```
NL66278: Another travis error, sorry for missing that on previous comment. Lines 128 and 129 are too long.

Best to have function arguments on separate lines with indent of four spaces. Most readable...

NL66278 (approved):
:+1: LGTM

r-hede:
Is there any other steps I've to take to get this pull request reviewed and merged?

NL66278:
I will ping some people I know. I do not have any special authority in this repository.

@StefanRijnhart @hbrunn

hbrunn (approved):
@r-hede the best way to get your stuff reviewed is reviewing yourself in the same repository

StefanRijnhart:
Couple of questions: Why are there changes in all these translation files? Don't we leave those to the automatic Weblate integration? And, why are there arbitrary whitespace changes? Between those two, I'm having trouble to spot the actual fix. So if you could give me a pointer to that, that would be great.

StefanRijnhart:
OK, found the actual fix (after reading this PR's description, thanks!) and they warrant the whitespace changes. That leaves the question about the translation files.

pedrobaeza (changes requested):
Here there's a remaining question, so I mark for waiting this way

NL66278:
@StefanRijnhart I looked into the translation files and they contained some wierdness before the changes, mainly having first an empty msgid ```""``` on one line, and then the actual string on the next. So I can imagine @r-hede wanting to rectify this. But maybe he can answer this better himself.

StefanRijnhart:
@pedrobaeza Can you confirm that translation files should be left to the weblate integration?

r-hede:
Indeed, @NL66278 is right. I deleted the empty spaces at the beginning of many translations which seemed to me useless.

pedrobaeza:
That's automatically done by OCA Git Bot when merging the PR, but it's not a critical problem to have it here. In the worst case, we'll have an extra commit undoing what is done here.

StefanRijnhart (approved):

StefanRijnhart:
OK, then it's allright!

PR #1565, opened 2019-04-26
[10.0] ldap populate safety

NL66278:
Register ldap configuration from which user added, to safely remove those later.


NL66278:
Now contained in https://github.com/OCA/server-tools/pull/1584

PR #1572, opened 2019-05-08
[IMP] base_exception: Only write if a change is necessary

etobella:
Without this change, the base_exception was unable to integrate with server-ux/base_tier_validation without an integration module for each affected model. Also, no writes will be launched if no differences are found.

### DIFF
--- a/base_exception/__manifest__.py
+++ b/base_exception/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Exception Rule',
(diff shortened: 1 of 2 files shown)


etobella:
ping @jbeficent

JordiBForgeFlow (approved):
:+1: Tested + code review.

JordiBForgeFlow:
Fast tracking this one.

PR #1576, opened 2019-05-14
[11.0][ADD] base_basic_employee_group

MiquelRForgeFlow:
NOTE: I am not sure if this module should go in this repository or in another.

Base Basic Employee Group
=========================

This module adds a new basic group in the Employee's category. This group
only has the most basic right accesses to work, the indispensable ones.
It will not have all the secondary accesses that the Employee group receives
throughout many other modules. Obviously, an user with this group alone
cannot do nothing. Ideally, this group should serve as a basis for new groups,
ones more specific and role oriented.


pedrobaeza:
This already exists in core.. It gives basic permissions to standard objects, but not to areas. I think this is a bit over-engineer for specific purposes.

MiquelRForgeFlow:
@pedrobaeza What do you mean? In odoo, if you are not an employee you can't see the menus for example. So, each employee must to have the employee group (`base.group_user`). But happens that the employee group has access rights to too many models. For example, you may want an employee without access rights to read products. Thus, this module provides another king of employee, very basic, and then after you can create other groups inheriting this basic employee instead the normal employee.

pedrobaeza:
Yes, base user must have that rights, so `base.group_user` is the base one you need. If you remove some, bad things will happen.

MiquelRForgeFlow:
Or for example, you may want users that cannot access to "Discuss" menu.

pedrobaeza:
But that should work on contrary, forbidding that through a new authorization group, but not rebuilding everything from the beginning. I already talk about this with @jbeficent. Check https://github.com/OCA/web/tree/11.0/web_disable_export_group as an example.

JordiBForgeFlow:
@pedrobaeza the would would be very grey if we always have to rely on the stock user and stock manager to design the stock related job reponsabilities.

What we have thought is a tool (in the other PR from Miquel) that allows to easily build security groups that are customized for the company roles.

This kind of tools are to be found in SAP, and very much appreciated by consultants. Because not all companies will accept the defult security groups.

pedrobaeza:
That tool already exist. It's called `base_user_role`, and repeating again about the approach, your specific restrictions must be additional security layers, not the contrary. Doing this is an error for me as this will require a lot of overhead for getting the same result.

hbrunn:
for the implementation of our internal helpdesk, we also needed a backend user that's not a base_user, so I totally agree with this module. It might actually even make the base role more restricted. This is our version (which will probably be based on this eventually after merge, and it does a little bit more than this module tailored to our use case): https://github.com/Therp/Therp-Addons/tree/10.0/trp_external_user
the argument for the different group is that whitelisting is much more controllable than blacklisting

JordiBForgeFlow:
@hbrunn the idea is that with this basic group + https://github.com/OCA/server-tools/pull/1577 you can fully design your groups based on your custom requirements.

pedrobaeza:
Well, I still agree that the needed whitelist is very huge for starting with a minimum working base so it's not worth to build it, and building the security whitelisting through a logger like #1577 will be a huge work, as needs to be repeated on each version you migrate the DB.

PR #1578, opened 2019-05-16
[9.0] server_environement: server_environment_file optional

lmignon:


lmignon:
travis errors seems to be into an other addons

PR #1581, opened 2019-05-22
IMP sql_export: render field as require when matching field def

bealdav:
With this change required fields are required in the UI

render

cc @bguillot @florian-dacosta


bguillot (approved):

bealdav:
seems Runbot is on rasberry field

florian-dacosta (approved):

PR #1584, opened 2019-05-24
[10.0] [IMP] users ldap reorganise

NL66278:
Improve usability and reliability of users_ldap_* modules by:
----------------------------------------------------------------------------------
* Put all configuration in its own view instead of general settings
* Add a cron job for scheduled retrieval of users from LDAP server
* Prevent users from setting password, if user has been created from LDAP
* Keep account of which users have been created from LDAP
* It is now, because of this, safe and usable to delete unknown users
* Increase test coverage
* Log more information on exceptions


NL66278:
Now succesfully used for nearly a year in production

users_ldap_menu/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "LDAP menu",
+    "version": "10.0.0.0.0",
```
thomaspaulb, line 6: Shouldnt this be 10.0.1.0.0?

users_ldap_menu/README.rst
```diff
@@ -0,0 +1,98 @@
...
+Usage
+=====
+
+Edit LDAP settings through Settings ==> Users ==> LDAP.
+
+Changelog
+=========
+
+
+10.0.0.0.0 (2019-05-24)
```
thomaspaulb, line 48: And here

users_ldap_menu/models/res_company_ldap.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @api.model
+    def create(self, vals):
+        """Make sure vals contains a company_id.
+
+        The link in standard Odoo between company and ldap server settings
+        lacks any logic. But the company must still be filled, so we default
+        it to the company of the user.
+
+        It might be usefull in a multi-company situation to enable imported
```
thomaspaulb: useful

users_ldap_no_reset/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "LDAP Populate",
+    "version": "10.0.0.0.0",
```
thomaspaulb, line 6: As well as here

thomaspaulb (commented):
It's very difficult to review this because you are changing so many things at the same time. Is it possible to do separate pull requests?

NL66278:
@thomaspaulb The fact that many things change is due to the need for reorganisation. The basic step was to create a separate menu for the ldap configuration, then base the other ldap modules on that one.

I tried to keep a logical division in commits for the reorganisation.

I gave the new modules a 10.0.0.0 version. The pre-existing modules, I incremented the version number. If thing are merged, there should also automatically be some increment, unless merge is done with nobump.

users_ldap_menu/models/res_company_ldap.py
NL66278: Corrected

thomaspaulb (approved):

thomaspaulb:
I've gone through the commits one by one, which makes it easier to review. It looks good to me.

PR #1586, opened 2019-05-28
[10.0][base_exception] Simplifies code and fix rules by domain

florian-dacosta:
Important note : When merging this PR, you will also need to merge PR on sale-workflow for sale_exception module, because the present sale_exception module won't work with this PR.
Sale_exception PR : https://github.com/OCA/sale-workflow/pull/855

The main goal of this PR is to fix the module and also simplify it a lot because it was becoming too complicated. (see https://github.com/OCA/server-tools/pull/1480).

It replaces https://github.com/OCA/server-tools/pull/1480

@sebastienbeau @hparfr @mourad-ehm

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -3,23 +3,11 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import time
-from functools import wraps
 
 from odoo import api, models, fields, _
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.safe_eval import safe_eval
-
-
-def implemented_by_base_exception(func):
-    """Call a prefixed function based on 'namespace'."""
-    @wraps(func)
-    def wrapper(cls, *args, **kwargs):
-        fun_name = func.__name__
-        fun = '_%s%s' % (cls.rule_group, fun_name)
-        if not hasattr(cls, fun):
-            fun = '_default%s' % (fun_name)
-        return getattr(cls, fun)(*args, **kwargs)
-    return wrapper
+from odoo import osv
 
 
 class ExceptionRule(models.Model):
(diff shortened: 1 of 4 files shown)


sebastienbeau (approved):
Code review
This new version is ok with the Specification done here  in my comment : https://github.com/OCA/server-tools/pull/1480
The code is cleanner and the domain on sale order and sale order line now work correctly without bug.

Thanks

rousseldenis:
@Cedric-Pigeon

base_exception/models/base_exception.py
```diff
@@ -74,14 +52,132 @@ class ExceptionRule(models.Model):
...
+
+    @api.multi
+    def _detect_exceptions_by_py_code(self, rule):
+        """
+            Find exceptions found on self.
+            If self is empty, check on all records.
+        """
+        domain = [('ignore_exception', '=', False)]
+        if self:
+            domain.append(('id', 'in', self.ids))
```
rousseldenis: It's better to use 

```suggestion
            domain = osv.expression.AND([domain, [('id', 'in', self.ids)]])
```

base_exception/models/base_exception.py
```diff
@@ -74,14 +52,132 @@ class ExceptionRule(models.Model):
...
+    @api.multi
+    def _detect_exceptions_by_domain(self, rule):
+        """
+            Find exceptions found on self.
+            If self is empty, check on all records.
+        """
+        domain = rule._get_domain()
+        domain.append(('ignore_exception', '=', False))
+        if self:
+            domain.append(('id', 'in', self.ids))
```
rousseldenis: same as above

rousseldenis (approved):
Code review. LGTM

base_exception/models/base_exception.py
bguillot: @florian-dacosta You can maybe avoid duplicate code for the domain of the search by adding a method to get the base domain

bguillot (approved):
Code review
My comment is not mandatory, otherwise it looks good.

Thanks for the work !

florian-dacosta:
comments attended @bguillot @rousseldenis

PR #1589, opened 2019-05-29
[12.0][base_exception] Refactore base_exception in order to drastically simplify it

florian-dacosta:
Hello,
Same refactore that if version 10. The module became overcomplicated and the goal is to make it simple again.
Doing this, I removed one feature, that I discovered doing it...
The next state id feature.

I remove it for several reasons : 
1) It is not generic enought and really should go to a separated module. Most people using this module won't use this feature.
2) This feature seems extremly dangerous, and like I said, quite specific. Dangerous because you can break workflows really easily. The state of a purchase order, a sale order or worse, an invoice, has a meaning. With this feature, an exception can put a sale order from draft to done, or an invoice from paid to cancel.... Of course, that is a matter of configuration, but I think it is really too permisive.
3) It actually breaks the module... I think nobody notices that because the state feature is used only in a specific case/specific module (Used only here AFAIK : purchase_order_approval_block and in version 11).
You can easily reproduce the bug, with sale_exception, put a state for an exception, which will apply for several draft sale orders.
Run cron test all draft orders => It will fail with a stack trace because `_detect_exceptions` is a multi record method and not mono record.
Also it tries to write on `self` which contains all record tested, not all records with exceptions...
4) It is not mature at all (see 3) ) and it is a pity because base_exception is an old module that make us thing it is really mature but it becomes quite buggy.

The goal of this PR is really to make the module simple again, to avoid nasty complicated bugs or performance issue.
Additional features should probably go in separated modules when possible.
This block https://github.com/OCA/purchase-workflow/pull/638

When merging this, we should also merge PR sale_exception PR or it will break it : https://github.com/OCA/sale-workflow/pull/860
(and purchase_exception if it is merged before)

@jbeficent I'd like your opinion seems you seems involved with purchase_order_approval_block module, which will need refactoring before going to version 12. (or Depend on a new module like base_exception_state or something.)
@rvalyi @sebastienbeau @bealdav @guewen @mourad-ehm @bguillot @hparfr @ivantodorovich

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -4,22 +4,10 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import time
-from functools import wraps
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.safe_eval import safe_eval
-
-
-def implemented_by_base_exception(func):
-    """Call a prefixed function based on 'namespace'."""
-    @wraps(func)
-    def wrapper(cls, *args, **kwargs):
-        fun_name = func.__name__
-        fun = '_%s%s' % (cls.rule_group, fun_name)
-        if not hasattr(cls, fun):
-            fun = '_default%s' % (fun_name)
-        return getattr(cls, fun)(*args, **kwargs)
-    return wrapper
+from odoo import osv
 
 
 class ExceptionRule(models.Model):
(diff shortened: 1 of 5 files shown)


base_exception/models/base_exception.py
```diff
@@ -94,7 +57,118 @@ def _get_domain(self):
...
+                'object': rec,
+                'obj': rec,
+                'pool': self.pool,
+                'env': self.env,
+                'cr': self.env.cr,
+                'uid': self.env.uid,
+                'user': self.env.user,
+                'time': time,
+                # copy context to prevent side-effects of eval
+                'context': self.env.context.copy()}
```
bealdav: Why do you change the method code here ? The old code must stay there, I thinks

bealdav (changes requested):
Thanks a lot for huge work. The code is very clean now and make easy to dive into

Thanks to fix my small remark

base_exception/models/base_exception.py
florian-dacosta: My mistake!
Good catch, I have fixed it

bealdav (approved):
Thanks a lot

sebastienbeau (approved):
As discuss here, this is the port of the refactor merged in 10.0 https://github.com/OCA/server-tools/pull/1586
Code review ok for me

PR #1592, opened 2019-06-03
[FIX] fix security issue due to wrong cache invalidation

sebastienbeau:
When we do an inherit with ormcache the cache is not clear correctly.
So if you try to remove the access right on a user, it will not remove until a restart of Odoo!
Here is a patch and a test that show the bug and fix it with a hack.

@StefanRijnhart @sbidoul @guewen @gurneyalex 

you can reproduce the bug by just removing the code of the method call_cache_clearing_methods

### DIFF
--- a/base_suspend_security/models/ir_model_access.py
+++ b/base_suspend_security/models/ir_model_access.py
@@ -33,3 +33,8 @@ class IrModelAccess(models.Model):
         return super(IrModelAccess, self).check(
             cr, uid, model, mode=mode, raise_exception=raise_exception,
             context=context)
+
+    def call_cache_clearing_methods(self, cr):
+        # we need to clear manually the original cache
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Interesting! Cc. @hbrunn

hbrunn (approved):
why do you think your solution is hacky? I think that's fine as it is, out of habit I'd return the result of super, but that's about it

StefanRijnhart (approved):
Thanks!

sebastienbeau:
> why do you think your solution is hacky? I think that's fine as it is, out of habit I'd return the result of super, but that's about it

I said it's hacky because the real perfect solution will be to fix directly Odoo ORM and support inheriting decorated function with cache. But this is not an easy task

sebastienbeau:
I force push to fix travis and return super call

PR #1598, opened 2019-06-07
[10.0][FIX] base_exception fix intersection operator

hparfr:
intersection operator is `&` instead of `and`


florian-dacosta (approved):

florian-dacosta:
@bealdav

bealdav (approved):

PR #1601, opened 2019-06-12
[10.0][base_exception] Fix union between recordset and parent recordset in base_exception

florian-dacosta:
There is still a bug due to last big refactore on the module.
A test has been added on sale_exception to detect it.

In the code, tested with sale_exception, we try to get the intersection between sale.order.line and sale.order, which obviously fail.

In this PR, we implement a way to manage this properly in submodules.

@hparfr @bealdav @mourad-ehm @sebastienbeau


bealdav (approved):
LGTM

sebalix (approved):

mourad-ehm (approved):

PR #1603, opened 2019-06-13
[9.0][sql_export_mail] Migration from v8

florian-dacosta:
Allow to send some file generated by sql_export by mail regularly (by cron).

@aheficent @bealdav 
I think you use sql_export, you may be interested.


florian-dacosta:
Lint test fails because of missing readme file. Is it really mandatory, the splited readme file does not work for v9?

PR #1605, opened 2019-06-13
[FIX][10.0] add safe_eval context

mourad-ehm:
Add safe_eval context to allow evaluating domain with python params ex(use ref method to get object by xml_id).

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -63,11 +63,23 @@ class ExceptionRule(models.Model):
                       "domain is missing to match the exception type.")
                 )
 
+    def _get_eval_context(self):
+        """ Prepare the context used when evaluating python code
+            :returns: dict -- evaluation context given to safe_eval
+        """
+        return {
+            'uid': self.env.uid,
+            'user': self.env.user,
+            'env': self.env,
+            'self': self,
+        }
+
     @api.multi
     def _get_domain(self):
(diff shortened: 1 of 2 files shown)


mourad-ehm:
ping @hparfr , @florian-dacosta, @bealdav , @pedrobaeza

base_exception/models/base_exception.py
```diff
@@ -63,11 +63,23 @@ def check_exception_type_consistency(self):
...
+        """ Prepare the context used when evaluating python code
+            :returns: dict -- evaluation context given to safe_eval
+        """
+        return {
+            'uid': self.env.uid,
+            'user': self.env.user,
+            'env': self.env,
+            'self': self,
+        }
+
```
hparfr, line 73: I think it should be only `''ref'": self.env.ref` to limit the risk of evading.

hparfr (approved):
I suggest to limit the scope at a maximum by default (only ref for xml id lookup).
And missing update minor version number.

This feature is really interesting.

base_exception/models/base_exception.py
sbidoul: Even `ref` returns a recordset which in turn exposes `sudo`.

thomaspaulb: Does that risk also hold for https://github.com/OCA/server-tools/pull/1606?

PR #1608, opened 2019-06-17
[ADD] Adds ir_rule_restrictions

daramousk:


daramousk:
@gfcapalbo @ntsirintanis @NL66278 We have discussed about this some time back, perhaps you could take a look?

ir_rule_restriction/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,14 @@
+Right now, Odoo defines to types of record rules:
+Global: All of them must have allow the action to be performed in order for
+the operation a user is performing to succeed. Therefore, an AND is
+performed in all of them.
+Group-specific: Any of them can allow the action to be performed in order for
+the operation a user is performing to succeed. Therefore, an OR is
+performed in all of them.
+
+Which means that if we have userA of groupA which global rules allow them to
```
NL66278, line 9: To have this effect it might be a better idea to use the concept of an exclusion rule. it means do NOT give the asked for permission to the records that apply, no matter if other rules would grant those permissions.

ir_rule_restriction/tests/test_ir_rule_restriction.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo.tests.common import TransactionCase
+from odoo.osv.expression import TRUE_LEAF, FALSE_LEAF
+
```
NL66278, line 6: This is quite an addition/change of the record rules, therefore your tests should me more comprehensive.

Let us say we have a rule X that grants users of group A access to partners in Amsterdam. Let us say we have a rule Y that grants users access to partners in group B that are companies. In the "normal" implementation, a user that is in group A and in group B would have access to partners that are either companies, or in Amsterdam. You want to limit the user to companies that are in Amsterdam. To properly test this, you will need a few test partners [person in Amsterdam, person outside Amsterdam, company in Amsterdam, company outside Amsterdam], a user that is in two groups, a user that is in one of your test groups, a user in none of the groups... And then see wether the partners they can read are limited to those you expect.

PR #1609, opened 2019-06-18
[IMP][12.0][base_exception] Improvements on base_exception

ivantodorovich:
This is what makes this possible in `sale_exception` (https://github.com/OCA/sale-workflow/pull/878):

image

If modules like `sale_exception` implement this approach for displaying errors, main_exception_id field wouldn't be necessary anymore.

---

Also added `copy=False` to `exception_ids` because they should be reevaluated, not copied.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import time
+import html
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.safe_eval import safe_eval
@@ -191,8 +192,21 @@ class BaseException(models.AbstractModel):
         string='Main Exception',
         store=True,
     )
-    exception_ids = fields.Many2many('exception.rule', string='Exceptions')
+    exceptions_summary = fields.Html(
+        'Exceptions Summary',
+        compute='_compute_exceptions_summary',
+    )
+    exception_ids = fields.Many2many(
+        'exception.rule',
+        string='Exceptions',
+        copy=False,
+    )
     ignore_exception = fields.Boolean('Ignore Exceptions', copy=False)
+
+    @api.multi
+    def action_ignore_exceptions(self):
+        self.write({'ignore_exception': True})
+        return True
 
     @api.depends('exception_ids', 'ignore_exception')
     def _compute_main_error(self):
@@ -201,6 +215,14 @@ class BaseException(models.AbstractModel):
                 rec.main_exception_id = rec.exception_ids[0]
             else:
                 rec.main_exception_id = False
+
+    @api.depends('exception_ids', 'ignore_exception')
+    def _compute_exceptions_summary(self):
+        for rec in self:
+            if rec.exception_ids and not rec.ignore_exception:
+                rec.exceptions_summary = '<ul>%s</ul>' % ''.join([
+                    '<li>%s: <i>%s</i></li>' % tuple(map(html.escape, (
+                        e.name, e.description))) for e in rec.exception_ids])
 
     @api.multi
     def _popup_exceptions(self):
--- a/base_exception/README.rst
+++ b/base_exception/README.rst
@@ -75,7 +75,7 @@ Contributors
 * SodexisTeam <dev@sodexis.com>
 * Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
 * Raphaël Reverdy <raphael.reverdy@akretion.com>
-* Iván Todorovich <ivan.todorovich@gmail.com>
+* Iván Todorovich <ivan.todorovich@druidoo.io>
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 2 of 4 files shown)


hparfr (approved):
* main_exception can be still be good in tree view, where you don't have much space
* copy=False is a good improvement.

florian-dacosta:
LGTM
But do not delete main_exception_id it still can be very usefull.
The exceptions have a sequence and sometimes you really just want to see the main exception, you can filter on this exception, group by this exception, etc...

ivantodorovich:
@florian-dacosta great!
Don't worry, not deleting anything here. 
Anyways, if we were to delete something like this it shuold be done on the next migration, IMO.

base_exception/models/base_exception.py
```diff
@@ -202,6 +218,16 @@ def _compute_main_error(self):
             else:
                 rec.main_exception_id = False
 
+    @api.depends('exception_ids', 'ignore_exception')
+    def _compute_exception_ids_formatted(self):
+        for rec in self:
+            if rec.ignore_exception or not rec.exception_ids:
+                rec.exception_ids_formatted = False
```
rousseldenis: IMHO this line is not necessary as before compute, values are void

ivantodorovich: Absolutely right. Just removed them

rousseldenis (approved):
Code review

rafaelbn:
Please @pedrobaeza @Yajo or any @OCA/tools-maintainers could review here and merge? this PR is blocking other PR, thanks

florian-dacosta (approved):

base_exception/models/base_exception.py
```diff
@@ -191,9 +191,25 @@ class BaseException(models.AbstractModel):
         string='Main Exception',
         store=True,
     )
-    exception_ids = fields.Many2many('exception.rule', string='Exceptions')
+
+    exception_ids_formatted = fields.Text(
```
yajo: I don't see this field being used anywhere... why?

Also I think the name is a bit confusing, as `_ids` suggests it's a x2many. You could change to:

```suggestion
    exceptions_description = fields.Text(
```

base_exception/models/base_exception.py
```diff
@@ -202,6 +218,14 @@ def _compute_main_error(self):
             else:
                 rec.main_exception_id = False
 
+    @api.depends('exception_ids', 'ignore_exception')
+    def _compute_exception_ids_formatted(self):
+        for rec in self:
+            if rec.exception_ids and not rec.ignore_exception:
+                rec.exception_ids_formatted = '\r\n'.join([
```
yajo: Would it make sense to make this field of type `Html` and join it in a `<ul><li>...</li>` way? IMHO it would look better.

base_exception/models/base_exception.py
ivantodorovich: It's not intended to be used in this module.
The purpose is to be used in the model that implements the exception. ie: https://github.com/OCA/sale-workflow/pull/878/files#diff-e3019ca8759cdf323e38ee35b80b147cR33

base_exception/models/base_exception.py
ivantodorovich: Yes, I thought about that but in fact I couldn't find a way to render the raw html on the message. (https://github.com/OCA/sale-workflow/pull/878/files#diff-e3019ca8759cdf323e38ee35b80b147cR33)
If you know an easy and clean way let me know

yajo: Yes, just use an html field as I said :sweat_smile:

rousseldenis (approved):

base_exception/models/base_exception.py
yajo: Are you going to apply this, or do we merge?

ivantodorovich: I'll try it on Monday ;) I made this PR a while ago but I remember trying to use html at the time, I just don't remember what the issue was. Anyways it's worth trying again

ivantodorovich: Well.. I honestly don't remember what I tried 😅 but it's working

base_exception/models/base_exception.py
```diff
@@ -191,9 +192,25 @@ class BaseException(models.AbstractModel):
         string='Main Exception',
         store=True,
     )
-    exception_ids = fields.Many2many('exception.rule', string='Exceptions')
+
```
yajo:
```suggestion
```

base_exception/models/base_exception.py
```diff
@@ -191,9 +192,25 @@ class BaseException(models.AbstractModel):
...
+    exceptions_summary = fields.Html(
+        'Exceptions Summary',
+        compute='_compute_exceptions_summary',
+    )
+
```
yajo:
```suggestion
```

base_exception/models/base_exception.py
```diff
@@ -191,9 +192,25 @@ class BaseException(models.AbstractModel):
...
+    exception_ids = fields.Many2many(
+        'exception.rule',
+        string='Exceptions',
+        copy=False,
+    )
+
```
yajo:
```suggestion
```

yajo (changes requested):
This looks mostly fine now, thanks.

Just some whitespace to remove, as a matter of code convention.

Also please squash all commits in one, to do the final merge.

ivantodorovich:
done @Yajo !
Thanks for reviewing this :)

yajo (approved):

PR #1610, opened 2019-06-18
[DEL] pyc file

thomaspaulb:
Saw a stray .pyc file, deleting it


sbidoul:
IIRC the presence of that file is important for testing the hashing algorithm.

thomaspaulb:
Isn't it excluded?

https://github.com/OCA/server-tools/blob/10.0/module_auto_update/tests/test_addon_hash.py#L65

sbidoul:
As you can see tests fail because the file must be present.

thomaspaulb:
Thanks, closing then.

Issue #1534, opened 2019-03-30
v11.0 database_cleanup Odoo server error !


m3asmi:
```
Error:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo11/odoo/tools/safe_eval.py", line 350, in safe_eval
    return unsafe_eval(c, globals_dict, locals_dict)
  File "", line 1, in <module>
  File "/opt/odoo/qs/custom/database_cleanup/models/purge_wizard.py", line 57, in get_wizard_action
    wizard = self.create({})
  File "/opt/odoo/qs/custom/database_cleanup/models/purge_wizard.py", line 93, in create
    return super(PurgeWizard, self).create(values)
  File "/opt/odoo/odoo11/odoo/models.py", line 3371, in create
    vals = self._add_missing_default_values(vals)
  File "/opt/odoo/odoo11/odoo/models.py", line 1602, in _add_missing_default_values
    defaults = self.default_get(list(missing_defaults))
  File "/opt/odoo/qs/custom/database_cleanup/models/purge_wizard.py", line 43, in default_get
    res['purge_line_ids'] = self.find()
  File "/opt/odoo/qs/custom/database_cleanup/models/purge_modules.py", line 77, in find
    purge_lines.purge()
  File "/opt/odoo/qs/custom/database_cleanup/models/purge_modules.py", line 54, in purge
    modules.unlink()
  File "/opt/odoo/odoo11/odoo/addons/base/module/module.py", line 317, in unlink
    return super(Module, self).unlink()
  File "/opt/odoo/odoo11/odoo/models.py", line 2952, in unlink
    cr.execute(query, (sub_ids,))
  File "/opt/odoo/odoo11/odoo/sql_db.py", line 155, in wrapper
    return f(self, *args, **kwargs)
  File "/opt/odoo/odoo11/odoo/sql_db.py", line 232, in execute
    res = self._obj.execute(query, params)
psycopg2.IntegrityError: null value in column "module" violates not-null constraint
DETAIL:  Failing row contains (1182, hr_payslip_input_salary_rule_rel_hr_salary_rule_id_fkey, null, 362, null, f, 2019-01-26 15:46:34.198773, 2019-01-26 15:46:34.198773, null, null, null, null).
CONTEXT:  SQL statement "UPDATE ONLY "public"."ir_model_constraint" SET "module" = NULL WHERE $1 OPERATOR(pg_catalog.=) "module""


During handling of the above exception, another exception occurred:

[...]

PR #1611, opened 2019-06-18
[FIX][10.0] database_cleanup should also cleanup stuff, eg. constraints, from uninstalled modules

thomaspaulb:
This solves https://github.com/OCA/server-tools/issues/1534 for me, but I don't know if it will break other things.


thomaspaulb:
@hbrunn Could you review this?

hbrunn:
Odoo used to crash with one of both, but https://github.com/CompassionCH/server-tools/pull/2/commits/95e0b9c96f62cb9e85fc6e2304a5af0145a048e8 is not very helpful about which one. Did you try out both cases?

PR #1613, opened 2019-06-21
[9.0][sql_export] Fix uses of params in sql query

florian-dacosta:
Hello,

The use of variable in this module is broken since https://github.com/OCA/server-tools/pull/1301 I think.
Indeed, variable of type date, char don't work...
Variable of type integer or boolean work, but not the rest...

I don't understand what that was changed, also because in the present way, sql injection is possible.
So I have reverted the change in sql_request_abstract.
Also, I fixed the variable view for sql_export and I have added test to cover the bug (testing a sql_query with variables : integer/date/many2many.

@mreficent @bguillot @bealdav @legalsylvain

### DIFF
--- a/sql_export/demo/sql_export.xml
+++ b/sql_export/demo/sql_export.xml
@@ -44,8 +44,8 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
     <record id="sql_export_partner_with_variables" model="sql.export">
         <field name="name">Export Partners With Variables (Demo Data)</field>
-	<field name="query">SELECT p.id FROM res_partner p LEFT JOIN res_partner_res_partner_category_rel rel ON rel.partner_id = p.id WHERE create_date &lt; %(x_date)s AND id = %(x_id)s AND rel.category_id in %(x_partner_categ_ids)s</field>
-	<field eval="[(6, 0, [ref('date_field_variable_sql'), ref('integer_field_variable_sql'), ref('m2m_field_variable_sql')])]" name="field_ids"/>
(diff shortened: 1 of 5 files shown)


MiquelRForgeFlow:
What about https://github.com/OCA/server-tools/pull/1047#issuecomment-391664598?

florian-dacosta:
Well, we don't have details about it.
I can't reproduce it.
Thing is, this way, it does not work at all for some params.
@Abranes Can you tell us more about this : https://github.com/OCA/server-tools/pull/1047#issuecomment-391664598

The tests I add show this working for different type of field.
If another case fail, we should absolutely add a test case about it else, it is probable it will fail again one day or another.

florian-dacosta:
No answer from @Abranes for now.
@mreficent Would you mind to test this PR with the date (there is demo data with variable)
And then test current merged version with date?
You will see then the problem. Or playing the test I have added in this PR with the current sql_export code, we can see easily it fails.
I'd like to go forward with this PR and the one in v12 because right now the variable feature is totally broken... Even the view to add the variable fields is broken!

MiquelRForgeFlow:
Ok. Could you fix the lint check errors?

florian-dacosta:
@mreficent 
I fixed pylint erros in the concerned modules, but Travis is still failing.
It seems it comes from the numerous pylint error or the other modules of the branch...No sure we can really make it green...
But anyway, I think the most important is the v12 as it is the last version, so, I would prefer to priorize this one https://github.com/OCA/server-tools/pull/1614

MiquelRForgeFlow (approved):

PR #1615, opened 2019-06-29
[12.0][autovacuum_mail_message] Standard migration

florian-dacosta:
@bealdav @bguillot


pedrobaeza:
This should go to OCA/social IMO

bealdav (approved):

florian-dacosta:
@pedrobaeza 
Why not... I am not sure as this module have more a technical goal of database cleaning.
It was made to avoid the mail_message table to become too huge overtime.
So even if it concerns messages, I find more logic if the module is together with database_cleanup and similar modules.
I am not sure someone would search for such a module in social repo?

pedrobaeza:
Well, as the involved table is mail, that was why I said that, and for not crowding a lot this generic repo, but you decide.

legalsylvain (approved):
Thanks a lot for porting this module. 
About the repos. Both are ok imho.
It's a social module because it's about mail, and it's a server-tools module, as it will be used by admin user to maintain a database small.
So ok to merge this PR. 

Regards.

florian-dacosta:
@legalsylvain Could you merge this one then?

florian-dacosta:
@legalsylvain 
On second thought, I'd like to wait a little since I may do a big refactore to extract the logic and share for a similar module about attachments...

PR #1618, opened 2019-07-07
[ADD] new module module_analysis

legalsylvain:
This module provides features to analyze the code installed on your instance distinguishing modules that comes from Odoo SA, the OCA, etc...

For each module installed, it compute the quantity of lines of code for Python, Javascript, XML and CSS.

analysis

analysis_pie

For a complete description of the module, see the [readme file](https://github.com/legalsylvain/server-tools/blob/12.0-ADD-module_analysis/module_analysis/README.rst).


etobella:
Interesting idea, but it is not working on runbot.

legalsylvain:
> Interesting idea, but it is not working on runbot.

Indeed. @etobella. Thanks for testing. The problem came from runbot / travis environment, regarding the call of ``cloc``.
- In runbot, cloc does'nt seems installed, and so it fails.
- On Travis cloc is installed, but not the good version, so we have some errors, regarding some extra settings. ([Log](https://travis-ci.org/OCA/server-tools/jobs/555717694#L1356))

I refactored the module, removing the call of the system lib ``cloc``and using the python lib ``pygount``.

* travis is now green :ok_hand: 

* Runbot is now working. (even it is red, but I don't know how to fix it, as log are not available (404 error) @gurneyalex, @pedrobaeza : any idea ?) 

image

Note to @bealdav : ``pygount`` has a follow-links option, so it should work in your docky context. Could you test it ? 

Thanks for your review.

legalsylvain:
and the winner is ``excel_import_export``, with 1316 lines of python code ;-)

pedrobaeza:
I was thinking to do this, but with `github_connector` for getting statistics of OCA modules. Are you planning this also (without the need of having them installed)?

legalsylvain:
> I was thinking to do this, but with github_connector for getting statistics of OCA modules. Are you planning this also (without the need of having them installed)?

yes I began this work, and it's work in progress since monthes (in github_connector) I forcast to finish this work in the next Open OCA Days, in October. Please ping me if you want to work before on it.

but well this module has other objectives, as the main one is to know the quantity of custom / community / Core code installed to have an evaluation of your technical debt on each instance of your customers.
(Or maybe, if we change rules, to know how many EE modules a customer is using).

pedrobaeza:
OK, I think this can be used as basecode for the other work, so I will check to see how can it be achieved and tell you back.

legalsylvain:
> OK, I think this can be used as basecode for the other work, so I will check to see how can it be achieved and tell you back.

The code is here, ~but for the time being, it analyses per repo, and not per module~
https://github.com/legalsylvain/interface-github/blob/11.0_FIX_github_modules/github_connector_odoo/models/github_repository_branch.py#L163
Edited : the analyses is done per module.

bealdav:
Nice I will test tomorrow. Thanks

bealdav:
Just for info

Pygounty with these settings interprets no encoding marker as cp1252

https://github.com/roskakori/pygount#source-code-encoding

`
INFO db pygount: /odoo/links/web_environment_ribbon/view/base_view.xml: analyze as XML using encoding utf-8 
INFO db pygount: /odoo/links/web_environment_ribbon/models/web_environment_ribbon_backend.py: analyze as Python using encoding cp1252 
INFO db pygount: /odoo/links/web_environment_ribbon/models/__init__.py: analyze as Python using encoding cp1252 
INFO db pygount: /odoo/links/web_environment_ribbon/static/src/css/ribbon.css: analyze as CSS+Lasso using encoding cp1252 
INFO db pygount: /odoo/links/web_environment_ribbon/static/src/js/ribbon.js: analyze as JavaScript using encoding cp1252 
`

probably need chardet here

but no importance for the purpose of the module

module_analysis/views/menu.xml
```diff
@@ -0,0 +1,24 @@
...
+<!-- 
+Copyright (C) 2019-Today: GRAP (<http://www.grap.coop/>)
+@author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ -->
+
+<odoo>
+
+    <menuitem id="menu_module_analysis"
+        name="Module Analysis"
```
bealdav: To be more mobile friendly it may be nice to use shorter name.
Here we are in module menu then no confusion

```suggestion
        name="Analysis"
```

module_analysis/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,20 @@
...
+Copyright (C) 2019-Today: GRAP (<http://www.grap.coop/>)
+@author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ -->
+
+<odoo noupdate="1">
+
+    <record id="parameter_exclude_directories" model="ir.config_parameter">
+        <field name="key">module_analysis.exclude_directories</field>
+        <field name="value">lib,demo,test,tests,doc,description</field>
```
bealdav, line 12: Just a personal opinion.
Tests should be taken account in lines count by default
They ensure the right behavior of the module

module_analysis/README.rst
```diff
@@ -0,0 +1,166 @@
...
+
+  class IrModuleModule(models.Model):
+     _inherit = 'ir.module.module'
+
+     xml_documentation_qty = fields.Integer(
+        string='Quantity of Comments in XML Files')
+
+    @api.model
+    def _get_analyse_settings(self):
+        res = super(IrModuleModule, self)._get_analyse_settings()
```
bealdav:
```suggestion
        res = super()._get_analyse_settings()
```

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,183 @@
+# Copyright (C) 2019-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import os
+# import json
```
bealdav:
```suggestion
```

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,183 @@
...
+import pygount
+from pathlib import Path
+import logging
+# import os
```
bealdav:
```suggestion
```

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,183 @@
...
+# import subprocess
```
bealdav:
```suggestion
```

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,183 @@
...
+            'python_code_qty': False,
+            'xml_code_qty': 0,
+            'js_code_qty': 0,
+            'css_code_qty': 0,
+        }
+
+    # Overload Section
+    @api.model
+    def update_list(self):
+        res = super(IrModuleModule, self).update_list()
```
bealdav:
```suggestion
        res = super().update_list()
```

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.env.context.get('analyse_installed_modules', False):
+            self.search([('state', '=', 'installed')]).button_analyse_code()
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super(IrModuleModule, self).write(vals)
```
bealdav:
```suggestion
        res = super().write(vals)
```

module_analysis/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,61 @@
...
+
+  class IrModuleModule(models.Model):
+     _inherit = 'ir.module.module'
+
+     xml_documentation_qty = fields.Integer(
+        string='Quantity of Comments in XML Files')
+
+    @api.model
+    def _get_analyse_settings(self):
+        res = super(IrModuleModule, self)._get_analyse_settings()
```
bealdav:
```suggestion
        res = super()._get_analyse_settings()
```

module_analysis/tests/test_module.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase, post_install
+
+
+@post_install(True)
+class TestModule(TransactionCase):
+
+    def setUp(self):
+        super(TestModule, self).setUp()
```
bealdav:
```suggestion
        super().setUp()
```

bealdav (changes requested):

module_analysis/data/ir_config_parameter.xml
bealdav, line 12:
```suggestion
        <field name="value">lib,demo,doc,description</field>
```

bealdav:
Thanks a lot for this contribution. just small typos found.
I have not found a real difference in performance comparing to cloc

legalsylvain:
Hi @bealdav. Thanks a lot for all the suggestions. I didn't know the possibility to user ``super()``. Do you know when it has been introduced ? It was not possible in V8, right ? 
All suggestion applyed. Except about the inclusion of test folders in the analysis. I think that in first approximation, we need to know the code running on the instance. (xml, js, python). test are like demo datas, extra code set to ensure the module is running, and available with data on runbot.

Any way, this entry is ``noupdate=1``, I think that the better approach is to be exhaustive, and let then the user removing the folder they want to include.

bealdav:
For super it's a python3 feature
No problem for the rest.

bealdav:
test result is a false negative.
Should be considered as Ok

ping @etobella 

----------
Module test summary
[1;32mSuccess[0;m auditlog,auto_backup,base_exception,base_fontawesome,base_jsonify,base_search_fuzzy,base_technical_user,datetime_formatter,excel_import_export,excel_import_export_demo,fetchmail_notify_error_to_sender,html_image_url_extractor,html_text,module_analysis,module_auto_update,onchange_helper,scheduler_error_mailer,sentry,sql_export,sql_request_abstract

+=======================================
|  Tests summary:
|---------------------------------------
| test_server.py              [1;32mSuccess[0;m
+=======================================

bealdav (approved):

mourad-ehm (approved):

mourad-ehm:
Code review. no test

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,180 @@
...
+
+            # Update Module Type, based on rules
+            module_type_id = rules._get_module_type_id_from_module(module)
+            module.module_type_id = module_type_id
+
+            # Get Path of module folder and parse the code
+            module_path = get_module_path(module.name)
+
+            # Get Files
+            analysed_datas = self._get_analyse_data_dict()
```
rvalyi, line 141: it's wrong to write "datas" with s; it looks like Odoo SA code. So like many latin words, data is already the plural form. Would you want to use the singular form, you would write datum. There are also 4 other occurences of "datas" in the subsequent 15 lines. https://www.learnenglish.de/mistakes/data.html

rvalyi (changes requested):

bealdav:
OK only pep Latin to fix

legalsylvain:
Hi. @rvalyi, @bealdav 

Sorry, but using datum / data doesn't seem to be an OCA convention. 
On some popular OCA 12 repositories (account-invoicing, server-tools, queue, etc...) locally : 
- ``rgrep "datas" ./ --include=*.py`` return a lot of lines (queue_job, excel_import_export, ...)
- ``rgrep "datum" ./ --include=*.py`` return nothing. (Well, in fact, some results in Odoo SA code, you know ;-) )

Please point OCA guidelines that mentions such points. Otherwise, please mark your comment as resolved.

kind regards.

bealdav:
@rvalyi an opportunity to contribute to guidelines (or to travis tests) to avoid to repeat yourself #jediscajedisrien #jeveuxpasfoutrelamerdemaissijepeuxycontribuercpasmalnonplus

legalsylvain:
@rvalyi : could you unblock this PR, as your comment about Latin-Pep8 is not part of OCA guidelines for the time being ?  thanks !

@all : 

- Runbot display this error. 
```
ERROR: pylint-odoo 0.0.1.dev459 has requirement docutils==0.14, but you'll have docutils 0.12 which is incompatible.
ERROR: pylint-odoo 0.0.1.dev459 has requirement lxml>=4.2.3, but you'll have lxml 3.7.1 which is incompatible.
```
- Runbot is green on 12.0 master branch
- This PR install a new python lib ``pygount``, but I don't understand why it could generate troubles, because pygount  requires only pygments lib : 
https://github.com/roskakori/pygount/blob/master/setup.py#L53

I dont' know how to fix it. Any idea ? @pedrobaeza, @sbidoul ?

sbidoul:
@legalsylvain I don't think these errors are the reason runbot is red. I rather suspect this line:

```2019-07-09 19:15:40,472 160 WARNING openerp_test pygount: cannot read /home/odoo/OCB-12.0/addons/mail/static/src/js/emojis.js using encoding cp1252: 'charmap' codec can't decode byte 0x9d in position 1713: character maps to <undefined>```

legalsylvain:
Thanks a lot @sbidoul ! I'll investigate on that point.

rafaelbn (approved):
Thanks Sylvain. This is very helpful to make quick audit in rescue projects

legalsylvain:
Thanks @sbidoul, it fixed the problem, forcing encoding.

3 approvals, and a Latin blocker. Could we consider merging this PR ? ;-)

Regards.

PR #1620, opened 2019-07-10
[10.0]base_exception Remove empty self feature.

hparfr:
This recently added feature is counter intuitive, error prone and is
already causing bugs in sale_workflow.

Like this:
https://github.com/OCA/sale-workflow/blob/10.0/sale_exception/models/sale_order_line.py#L28
```py
(sale_order_line)
    @api.multi
    def _detect_exceptions(self, rule):
        records = super(SaleOrderLine, self)._detect_exceptions(rule)
```

I think it's the correct and obvious way to write this code. But the actual implementation, will fail if the sale order has no lines. It will test exceptions on all the sale order line (empty self = search([])) and return exceptions tied to other sale orders.

Please merge it with a major version number bump.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -99,7 +99,6 @@ class BaseExceptionMethod(models.AbstractModel):
     def detect_exceptions(self):
         """List all exception_ids applied on self
         Exception ids are also written on records
-        If self is empty, check exceptions on all active records.
         """
         rules = self.env['exception.rule'].sudo().search(
             self._rule_domain())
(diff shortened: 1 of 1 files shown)


hparfr:
@sebastienbeau @florian-dacosta

florian-dacosta (approved):
LGTM

yvaucher (approved):
I agree you want to check the exception on a selected set of object. When you start to have thousands of objects this feature is meaningless and resource greedy.

Just got some MemoryError due to this lastly being run while preparing data with onchanges with addition of a api.constraints which was making the check on all the 3.1 M sale order lines in the system. That was not easy to spot. Thanks for the fix!

guewen (approved):

gurneyalex (approved):

PR #1621, opened 2019-07-10
10.0 fix dependencies in fetchmail attach from folder

ntsirintanis:
Replace `simplejson` dependency in `fetchmail_attach_from_folder` with `json`. 
`simplejson` could not be loaded and this caused all kinds of problems in a running odoo instance. 

@hbrunn @NL66278 @gfcapalbo


fetchmail_attach_from_folder/__manifest__.py
```diff
@@ -19,7 +19,7 @@
     'auto_install': False,
     'external_dependencies': {
         'python': [
-            'simplejson',
+            'json',
```
hbrunn: remove all of this, `json` is a standard python library

hbrunn (changes requested):

hbrunn (approved):

NL66278 (approved):
:+1:  LGTM

PR #1623, opened 2019-07-17
[IMP] auto_backup: Count files instead of looking at dates

hbrunn:
this solves two problems: If the oldest file to delete is some seconds older/newer than now, we'll either delete it when it shouldn't be deleted or the other way around.

Also, by making this not time dependent any more, there's no danger to delete all backups if some backups in between failed, it will always keep the last N backups.

### DIFF
--- a/auto_backup/models/db_backup.py
+++ b/auto_backup/models/db_backup.py
@@ -9,7 +9,7 @@ import shutil
 import tempfile
 import traceback
 from contextlib import contextmanager
-from datetime import datetime, timedelta
+from datetime import datetime
 from glob import iglob
 from openerp import exceptions, models, fields, api, _, tools
 from openerp.service import db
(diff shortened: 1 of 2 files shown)


foutoucour (approved):

auto_backup/models/db_backup.py
```diff
@@ -240,10 +240,10 @@ def cleanup(self):
                 oldest = self.filename(now - timedelta(days=rec.days_to_keep))
```
foutoucour: I think the new philosophy messes with this approach. What do you think ?

auto_backup/models/db_backup.py
```diff
@@ -240,10 +240,10 @@ def cleanup(self):
...
-                                                   "*.dump.zip")):
-                        if os.path.basename(name) < oldest:
-                            os.unlink(name)
+                    for name in sorted(
+                            iglob(os.path.join(rec.folder, "*.dump.zip"))
+                    )[:-rec.days_to_keep]:
+                        os.unlink(name)
 
                 elif rec.method == "sftp":
                     with rec.sftp_connection() as remote:
```
foutoucour, line 247: The update affects also remote backups. Could you work on this part too ?

foutoucour (changes requested):

auto_backup/models/db_backup.py
hbrunn: I kept the remote part as it is because I don't use it myself, but I think you're right, it doesn't make sense to do A remotely and B locally. Will add this to my TODO list

auto_backup/models/db_backup.py
hbrunn: done, I also added some (incomplete) tests for the remote case

foutoucour (approved):

thomaspaulb (approved):
Code and functional review: LGTM and battle-tested for more than a year in production.

thomaspaulb:
@dreispt Could you merge

andhit-r (approved):

thomaspaulb:
@hbrunn This is ready for merge.

marcelsavegnago (approved):

PR #1625, opened 2019-07-25
 [12.0]base_exception Remove empty self feature. #1620 

hparfr:
This recently added feature is counter intuitive, error prone and is
already causing bugs in sale_workflow.

Like this:
https://github.com/OCA/sale-workflow/blob/10.0/sale_exception/models/sale_order_line.py#L28

(sale_order_line)
    @api.multi
    def _detect_exceptions(self, rule):
        records = super(SaleOrderLine, self)._detect_exceptions(rule)

I think it's the correct and obvious way to write this code. But the actual implementation, will fail if the sale order has no lines. It will test exceptions on all the sale order line (empty self = search([])) and return exceptions tied to other sale orders.

Please merge it with a major version number bump.

same as #1620

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -86,7 +86,6 @@ class BaseExceptionMethod(models.AbstractModel):
     def detect_exceptions(self):
         """List all exception_ids applied on self
         Exception ids are also written on records
-        If self is empty, check exceptions on all active records.
         """
         rules = self.env['exception.rule'].sudo().search(
             self._rule_domain())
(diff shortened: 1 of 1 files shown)


hparfr:
@florian-dacosta

florian-dacosta (approved):
thanks

hparfr:
@yvaucher @guewen you can approve this one too

gurneyalex (approved):

yvaucher (approved):

yvaucher:
damn the merge failed here a while ago.

Will handle the rebase manually.

yvaucher:
Seems it has already been merged there: https://github.com/OCA/server-tools/commit/60d3d6241df96256768638194adeff43d32998cc

PR #1627, opened 2019-07-29
auditlog: Fix fast-logging exception on create with multiple records

### LINKED ISSUE #1619: [12.0] Auditlog traceback
Traceback when creating an object (invoice in this case);

```
  File "/home/odoo/src/odoo/odoo/http.py", line 656, in _handle_exception
  File "/home/odoo/src/odoo/odoo/http.py", line 314, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/odoo/src/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/home/odoo/src/odoo/odoo/http.py", line 698, in dispatch
  File "/home/odoo/src/odoo/odoo/http.py", line 346, in _call_function
  File "/home/odoo/src/odoo/odoo/service/model.py", line 97, in wrapper
  File "/home/odoo/src/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo/src/odoo/odoo/http.py", line 941, in __call__
  File "/home/odoo/src/odoo/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 966, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 954, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/odoo/src/odoo/odoo/api.py", line 749, in call_kw
    return _call_kw_multi(method, model, args, kwargs)
  File "/home/odoo/src/odoo/odoo/api.py", line 736, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/src/odoo/addons/sale/wizard/sale_make_invoice_advance.py", line 135, in create_invoices
    sale_orders.action_invoice_create()
  File "/home/odoo/src/odoo/addons/sale/models/sale.py", line 542, in action_invoice_create
    invoice = inv_obj.create(inv_data)
  File "<decorator-gen-407>", line 2, in create
  File "/home/odoo/src/odoo/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/home/odoo/src/odoo/addons/base_automation/models/base_automation.py", line 228, in create
    records = create.origin(self.with_env(actions.env), vals_list, **kw)
  File "/home/odoo/src/user/community/auditlog/models/rule.py", line 230, in create_fast
    vals2 = dict(vals)
ValueError: dictionary update sequence element #0 has length 16; 2 is required
```

mtelahun:
When fast-logging create calls the code assumed the old-style
calling convention where only one dictionary value is passed
into the create() function. When it received a list of
dictionaries instead it blew up trying to manipulate a
single dict object instead of a list.

o Fix the code to degrade gracefully to full logging, instead.
o Add a unit test to catch this case in the future
o Fixes #1619

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import models, fields, api, modules, _
+from collections import Mapping
 
 FIELDS_BLACKLIST = [
     'id', 'create_uid', 'create_date', 'write_uid', 'write_date',
@@ -227,12 +228,25 @@ class AuditlogRule(models.Model):
         def create_fast(self, vals, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env['auditlog.rule']
-            vals2 = dict(vals)
             new_record = create_fast.origin(self, vals, **kwargs)
-            new_values = {new_record.id: vals2}
-            rule_model.sudo().create_logs(
-                self.env.uid, self._name, new_record.ids,
-                'create', None, new_values, {'log_type': log_type})
+
+            # If only a single new record is being created do a fast log.
+            # Otherwise, degrade to a full log.
+            #
+            if isinstance(vals, Mapping):
+                vals2 = dict(vals)
+                new_values = {new_record.id: vals2}
+                rule_model.sudo().create_logs(
+                    self.env.uid, self._name, new_record.ids,
+                    'create', None, new_values, {'log_type': log_type})
+            else:
+                new_values = dict(
+                    (d['id'], d) for d in new_record.sudo()
+                    .with_context(prefetch_fields=False)
+                    .read(list(self._fields)))
+                rule_model.sudo().create_logs(
+                    self.env.uid, self._name, new_record.ids,
+                    'create', None, new_values, {'log_type': log_type})
             return new_record
 
         return create_full if self.log_type == 'full' else create_fast
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -119,3 +119,22 @@ class TestAuditlogFast(TransactionCase, AuditlogCommon):
     def tearDown(self):
         self.groups_rule.unlink()
         super(TestAuditlogFast, self).tearDown()
+
+    def test_LogCreateMulti(self):
+        """Test creation of multiple objects at once."""
+
+        auditlog_log = self.env['auditlog.log']
+        self.groups_rule.subscribe()
+
+        vals = [
+            {'name': 'Test Group 01'},
+            {'name': 'Test Group 02'},
+        ]
+        groups = self.env['res.groups'].create(vals)
+        self.assertTrue(len(groups) == 2)
+        for grp in groups:
+            self.assertTrue(auditlog_log.search([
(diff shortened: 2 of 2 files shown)


auditlog/models/rule.py
```diff
@@ -227,12 +228,25 @@ def create_full(self, vals, **kwargs):
         def create_fast(self, vals, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env['auditlog.rule']
-            vals2 = dict(vals)
```
sebalix, line 230: Hi @mtelahun 

Thanks for the PR!
Instead of generating a full log when we get a list of dictionaries, what about this:
- replace the above line by making a deep copy of `vals`: `vals2 = copy.deepcopy(vals)`
- as the `create` method keep the order (record IDs should fit the order of dictionaries provided), set `new_values` with something like:
```python
for new_record_id, record_vals in zip(new_record.ids, vals2):
    new_values = {new_record_id: vals2}
        self.env.uid, self._name, [new_record_id],
        'create', None, new_values, {'log_type': log_type})
````

What do you think?

mtelahun: Hi @sebalix 
I thought about that, but decided against it because there is no guarantee (at least in the ORM reference) that the returned recordset is in the same order as the list passed to the create() method. Specifically, I wasn't sure what would happen if the '_order' field of the model specified an ordering other than the 'id' of the record. Given that this is supposed to be an auditing tool I thought it would be better to opt for correctness rather than performance.

sebalix: I understand, and before proposing that I read a bit the code in `odoo/models.py` and the assumption that the order is the same is right (unless I misread the code). But I didn't test it myself... the `_order` should occur only on a search operation, here we are more in a `browse` usage where the order of IDs and recordset returned should be the same. To confirm.

The fast log was created to reduce the performance penalty of the full log, and I don't know how much the create multi is used by Odoo v12, but it could lead to big issues from a UX experience depending on the data model.

EDIT: the tests could be adapted to ensure that this assumption is right BTW

CasVissers-360ERP (approved):
Fixes the issue for me!

sebalix:
@mtelahun I open another PR #1688 which fixes the same issue, using `@api.model_create_multi` on patch methods to ensure to get a list of values instead of checking the type of `vals`. It fixes both fast and full log types, and the unit test ensure that the recordset corresponds to the list of values (`res.group` has an _order on `name`).

PieterPaulussen (approved):

PieterPaulussen:
Going to implement this fix for a customer to see the results and will update with validation! Thanks for the fix @sebalix

sebalix:
@PieterPaulussen no my advise is to not merge the fix proposed in this PR. Instead we should merge https://github.com/OCA/server-tools/pull/1688 (and https://github.com/OCA/server-tools/pull/1686 for the 11.0 branch) which is cleaner.

PieterPaulussen:
@sebalix indeed, I would even suggest to close the PR since the fix is not a complete fix.

StefanRijnhart:
Thank you for your contribution! I'm closing this because the superseded PR was merged.

PR #1628, opened 2019-08-02
[ADD] record_merge: Technical module to merge records from any table

jonerikceberio:
The README should be extensive enough, but the TL;DR would be:
A tool to merge records like the base_partner_merge, but in any table and configurable. Also, group and merge many records automatically.


pedrobaeza:
Why aren't you using existing library foundation openupgradelib, https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py, that implements all the logic without interface?

jonerikceberio:
You are absolutely right, did not think of that one. At a quick glance is not exactly what I have done, but pretty much the same idea. I suppose this calls for a comparition of both logics and see how to use the openupgradelib in my module. I will look into it.

pedrobaeza:
Thanks any way for the module! It was in my TODO since I developed the methods in openupgradelib. If you are going to come to OCA Days in Belgium, we can meet for setting the common ground and finish this module.

PR #1638, opened 2019-08-07
[10.0] base_exception: fix concurrency issue

gurneyalex:
When several connections are creating sale orders, the exceptions are checked parallel transactions. 

Without this fix:

* the write() can be  a noop but this will nevertheless trigger a UPDATE of the write_uid and write_date automatic fields, which in turn will cause a Rollback of the transaction because of concurrent access. When using the web client, there is a retry, but this causes a slow down. When scripting the creation of sale orders from an external system (e.g. Magento) this causes failed queue.jobs
* if the write() actually writes something, then we need to serialize the database updates / insertions, otherwise we run into the same problem as above. We do this by locking the rows of the transactions in SQL with SELECT ... FOR UPDATE.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -112,7 +112,12 @@ class BaseExceptionMethod(models.AbstractModel):
             to_add = records_with_exception - commons
             to_remove_list = [(3, x.id, _) for x in to_remove]
             to_add_list = [(4, x.id, _) for x in to_add]
-            rule.write({reverse_field: to_remove_list + to_add_list})
+            if to_remove_list or to_add_list:
+                self.env.cr.execute(
+                    'SELECT id FROM exception_rule '
+                    'WHERE id = %s FOR UPDATE', (rule.id,)
+                )  # no deadlock because the order is enforced by sequence
+                rule.write({reverse_field: to_remove_list + to_add_list})
             if records_with_exception:
                 all_exception_ids.append(rule.id)
         return all_exception_ids


yvaucher (approved):
LGTM + tested locally in a setup that was raising concurrency errors. (queue job creating sale orders with 2 workers in parallel)

guewen:
It reads `exception_rule` before locking the table, which means even if the table is locked by transaction A, transaction B will throw a concurrent update error when transaction A releases the lock; the values read by B will be invalid at this point.

Queries:
```
2019-08-13 11:46:04,134 1 DEBUG prod odoo.sql_db: query: SELECT "exception_rule".id FROM "exception_rule" WHERE (("exception_rule"."active" = %s)  AND  (("exception_rule"."model" = %s)  AND  ("exception_rule"."warning_only" IS NULL or "exception_rule"."warning_only" = fals
e ))) ORDER BY COALESCE("exception_rule"."active", false) DESC,"exception_rule"."sequence" ASC
2019-08-13 11:46:04,135 1 DEBUG prod odoo.sql_db: query:  SELECT "exception_rule"."create_uid" as "create_uid","exception_rule"."sequence" as "sequence",COALESCE("exception_rule__description"."value", "exception_rule"."description") as "description",COALESCE("exception_rul
e__warning_text"."value", "exception_rule"."warning_text") as "warning_text","exception_rule"."exception_type" as "exception_type","exception_rule"."write_date" as "write_date","exception_rule"."active" as "active",COALESCE("exception_rule__name"."value", "exception_rule".
"name") as "name","exception_rule"."model" as "model","exception_rule"."code" as "code","exception_rule"."create_date" as "create_date","exception_rule"."domain" as "domain","exception_rule"."warning_only" as "warning_only","exception_rule"."write_uid" as "write_uid","exce
ption_rule"."id" as "id" FROM "exception_rule" LEFT JOIN
                (SELECT DISTINCT ON (res_id) res_id, value
                 FROM "ir_translation"
                 WHERE name=%s AND lang=%s AND value!=%s
                 ORDER BY res_id, id DESC)
             as "exception_rule__description" ON ("exception_rule"."id" = "exception_rule__description"."res_id") LEFT JOIN
                (SELECT DISTINCT ON (res_id) res_id, value
                 FROM "ir_translation"
                 WHERE name=%s AND lang=%s AND value!=%s
                 ORDER BY res_id, id DESC)
             as "exception_rule__warning_text" ON ("exception_rule"."id" = "exception_rule__warning_text"."res_id") LEFT JOIN
                (SELECT DISTINCT ON (res_id) res_id, value
                 FROM "ir_translation"
                 WHERE name=%s AND lang=%s AND value!=%s
                 ORDER BY res_id, id DESC)
             as "exception_rule__name" ON ("exception_rule"."id" = "exception_rule__name"."res_id")
                        WHERE "exception_rule".id IN %s  ORDER BY COALESCE("exception_rule"."active", false) DESC,"exception_rule"."sequence" ASC

2019-08-13 11:46:04,137 1 DEBUG prod odoo.sql_db: query: SELECT "sale_order".id FROM "sale_order" WHERE (("sale_order"."ignore_exception" IS NULL or "sale_order"."ignore_exception" = false )  AND  ("sale_order"."id" in (%s))) ORDER BY "sale_order"."date_order" DESC,"sale_o
rder"."id" DESC
2019-08-13 11:46:04,138 1 DEBUG prod odoo.sql_db: query:  SELECT "exception_rule"."id" as "id" FROM "exception_rule"
                        WHERE "exception_rule".id IN %s  ORDER BY COALESCE("exception_rule"."active", false) DESC,"exception_rule"."sequence" ASC

2019-08-13 11:46:04,139 1 DEBUG prod odoo.sql_db: query:  SELECT exception_rule_sale_order_rel.exception_rule_id, exception_rule_sale_order_rel.sale_order_id FROM exception_rule_sale_order_rel, "sale_order"
                    WHERE 1=1 AND exception_rule_sale_order_rel.exception_rule_id IN %s AND exception_rule_sale_order_rel.sale_order_id = sale_order.id
                     ORDER BY "sale_order"."date_order" DESC,"sale_order"."id" DESC   OFFSET 0

2019-08-13 11:46:04,141 1 DEBUG prod odoo.sql_db: query: SELECT id FROM exception_rule WHERE id = %s FOR UPDATE
```

guewen:
Maybe the patch is valid but incomplete, see https://github.com/OCA/server-tools/issues/1642#issuecomment-521168334
I'll update the PR once I find...

guewen:
I propose an alternative correction at https://github.com/OCA/server-tools/pull/1647 that corrects the same lock + an issue with unrelated sale orders being locked as well.

guewen (changes requested):
I propose an alternative correction at https://github.com/OCA/server-tools/pull/1647 that corrects the same lock + an issue with unrelated sale orders being locked as well.

gurneyalex:
superseeded

PR #1639, opened 2019-08-07
[12.0] base_exception: fix concurrency issue

gurneyalex:
When several connections are creating sale orders, the exceptions are checked parallel transactions.

Without this fix:

* the write() can be a noop but this will nevertheless trigger a UPDATE of the write_uid and write_date automatic fields, which in turn will cause a Rollback of the transaction because of concurrent access. When using the web client, there is a retry, but this causes a slow down. When scripting the creation of sale orders from an external system (e.g. Magento) this causes failed queue.jobs
* if the write() actually writes something, then we need to serialize the database updates / insertions, otherwise we run into the same problem as above. We do this by locking the rows of the transactions in SQL with SELECT ... FOR UPDATE.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -86,7 +86,6 @@ class BaseExceptionMethod(models.AbstractModel):
     def detect_exceptions(self):
         """List all exception_ids applied on self
         Exception ids are also written on records
-        If self is empty, check exceptions on all active records.
         """
         rules = self.env['exception.rule'].sudo().search(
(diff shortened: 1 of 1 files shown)


gurneyalex:
forward port of #1638

yvaucher (approved):

guewen (changes requested):
#1648 is more complete

yvaucher:
@guewen let's close this one in favor of #1648

PR #1643, opened 2019-08-13
[ADD] base_test_chrome

hbrunn:
This addon makes Odoo v12's chrome based tests available for lower versions in order not to constantly have to work around phantomjs quirks.

Should be very simple to forward port to 9, 10, 11


hbrunn:
@mart-e is the attribution in https://github.com/OCA/server-tools/pull/1643/files#diff-99dd4eacb4161ba180bcbc5f42831f55R71 and https://github.com/OCA/server-tools/pull/1643/files#diff-727577c155200751bc6491d6d9108c86R7 sufficient or do you prefer some other way of doing it?

mart-e:
@hbrunn no that's good, thanks for checking

base_test_chrome/__openerp__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Run JS tests using chrome",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA),Odoo SA",
+    "license": "LGPL-3",
+    "category": "Tools",
+    "summary": "Use a modern browser for running client side tests",
+    "depends": [
+        'base',
```
yajo:
```suggestion
        'web',
```

hbrunn: thanks, good one

Issue #1642, opened 2019-08-12
base_exception: side-effect in @api.constrains lead to no retry


guewen:
# Introduction

The `base_exception` module states that we have to call `_check_exception` in an `@api.constrains` method.

https://github.com/OCA/server-tools/blob/5157720da42cccfd3bbda0072384d61d6162cd5d/base_exception/models/base_exception.py#L221-L236

And it's what's been done in the `sale_exception` module:

https://github.com/OCA/sale-workflow/blob/8428446c79981aee281eb1773ff84c5a32d02067/sale_exception/models/sale.py#L49-L53

```python
    @api.constrains('ignore_exception', 'order_line', 'state')
    def sale_check_exception(self):
        orders = self.filtered(lambda s: s.state == 'sale')
        if orders:
            orders._check_exception()
```

# So what?

One of the method called in `_check_exception`, the method `detect_exceptions`, has side-effects on the database; it does a write on `exception.rule` on a field which must be a `Many2many` with the record being checked. For instance, [sale_ids](https://github.com/OCA/sale-workflow/blob/8428446c79981aee281eb1773ff84c5a32d02067/sale_exception/models/sale.py#L16-L18) for the sales. the write will issue an UPDATE on `exception.rule` (for the `write_date`) and an INSERT in the `Many2many` relation.

[A PR](https://github.com/OCA/server-tools/issues/1647) has already been created to limit the effect of concurrency here, as 2+ orders with an exception won't be able to call the method at the same time due to the `write_date` change. 

The subject of this issue is about the usage of `@api.constrains` with side-effects, so what's the problem?
* **any** exception thrown in an `@api.constrains` method will be caught and re-raised as `ValidationError` (https://github.com/odoo/odoo/blob/af3f6e596df6a558de8b465019028ff8cc2d7439/odoo/models.py#L1056)
* "*any* exception" includes `psycopg2.OperationalError`
* Some kinds of `OperationalError` are normally retried (https://github.com/odoo/odoo/blob/af3f6e596df6a558de8b465019028ff8cc2d7439/odoo/service/model.py#L110)

Summarizing:

Any create or write done in a method decorated by `@api.constrains` will never be retried when an `OperationalError` that could be retried (such as "could not serialize access due to concurrent update") occurs as the `OperationalError` is shadowed by a `ValidationError`.

# Solutions

[...]

PR #1644, opened 2019-08-13
[10.0] base_exception: Remove side effect from api.constrains

guewen:
In the documentation.

The method called by '_check_exception' has a side effect, it writes
on 'exception.rule' + on the Many2many relation between it and
the related model (such as sale.order). When decorated by
@api.constrains, any error during the method will be caught and
re-raised as "ValidationError".  This part of code is very prone to
concurrent updates as 2 sales having the same exception will both write
on the same 'exception.rule'.  A concurrent update (OperationalError) is
re-raised as ValidationError, and then is not retried properly.

Calling the same method in create/write has the same effect than
@api.constrains without shadowing the exception type.

Full explanation:
OCA/server-tools#1642

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -237,14 +237,40 @@ class BaseException(models.AbstractModel):
     @api.multi
     def _check_exception(self):
         """
-        This method must be used in a constraint that must be created in the
-        object that inherits for base.exception.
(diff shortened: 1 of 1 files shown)


florian-dacosta (approved):
Thanks

sebalix (approved):

PR #1647, opened 2019-08-14
[10.0] Remove RowExclusiveLock on exception_rule

guewen:
The goal of the modified method is to create or remove the relationship
(in the M2m relation tabel) between the tested model (such as
sale_order) and the exception rules. When the ORM writes on
ExceptionRule.sale_ids (using the example of sale_exception), it will
first proceeds with these updates:

* an UPDATE on exception_rule to set the write_date
* INSERT or DELETE on the relation table
* but then, as "write" is called on the exception rule, the ORM will
  trigger the api.depends to recompute all the "main_exception_ids"
  of the records (sales, ...) related to it, leading to an UPDATE
  for each sale order

We end up with RowExclusiveLock on such records:

* All the records of the relation table added / deleted for the current
sale order
* All the records of exception_rule matching the current sale order
* All the records of sale_order related to the exception rules matching
the current sale order

The first one is expected, the next 2 are not. We can remove the lock on
the exception_rule table by removing `_log_access`, however in any case,
the main_exception_ids computed field will continue to lock many sale
orders, effectively preventing 2 sales orders with the same exception
to be confirmed at the same time.

Reversing the write by writing on SaleOrder instead of ExceptionRule
fixes the 2 unexpected locks. It should not result in more queries: the
"to remove" part generates one DELETE on the relation table for the rule
to remove and the "to add" part generates one INSERT (with unnest) for the rule to add,
both will be exactly the same in both cases.

Related to #1642
Replaces #1638

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -103,6 +103,8 @@ class BaseExceptionMethod(models.AbstractModel):
         rules = self.env['exception.rule'].sudo().search(
             self._rule_domain())
         all_exception_ids = []
+        rules_to_remove = {}
+        rules_to_add = {}
         for rule in rules:
             records_with_exception = self._detect_exceptions(rule)
             reverse_field = self._reverse_field()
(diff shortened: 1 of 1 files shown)


base_exception/models/base_exception.py
```diff
@@ -103,18 +103,28 @@ def detect_exceptions(self):
...
+            # we expect to always work on the same model type
+            rules_to_remove.setdefault(
+                rule.id, main_records.browse()
+            ).update(to_remove)
+            rules_to_add.setdefault(
+                rule.id, main_records.browse()
+            ).update(to_add)
             if records_with_exception:
                 all_exception_ids.append(rule.id)
+        for rule_id, records in rules_to_remove.iteritems():
```
sebalix, line 137: It deserves a comment (the last paragraph of your PR description maybe) for the future.

sebalix (approved):
Small comment, but approving.

base_exception/models/base_exception.py
guewen: done!

florian-dacosta (approved):
Thanks a lot for the fix @guewen 
It seems fine to me.
I agree these concurrent update issues have to be fixed and we'll eventually fix some other performance issue in the future if necessary, having this issue in mind.

guewen:
Squashed the fixup

PR #1653, opened 2019-08-26
10.0 base test chrome

gfcapalbo:
New commit for the porting of v8 base_test_chrome to v10.
Preserved original commit history of @hbrunn and added a XMLRPC import used in v12


base_test_chrome/common.py
```diff
@@ -0,0 +1,566 @@
...
+import requests
+import shutil
+import signal
+import subprocess
+import tempfile
+import threading
+import time
+import unittest
+import werkzeug.urls
+import xmlrpclib
```
thomaspaulb, line 20: Pylint is complaining that you're importing it here, but also below.

thomaspaulb (changes requested):
LGTM except for one Pylint. I think test failure will be gone after rebase?

PR #1654, opened 2019-08-26
[ADD] base_server_action_parse: To work with base64 files like xmls

hugho-ad:


hugho-ad:
@moylop260 

could you review please

PR #1661, opened 2019-09-09
[10.0][FIX]base_jsonify: Preserve initial order on ir.export.line

lmignon:
We can't change the order on ir.export.line since it's used to preserve the order speicifed by the user when creating a export filter into the UI

### DIFF
--- a/base_custom_info/models/custom_info_template.py
+++ b/base_custom_info/models/custom_info_template.py
@@ -53,18 +53,18 @@ class CustomInfoTemplate(models.Model):
         self._inverse_model()
 
     @api.multi
-    @api.constrains("model_id")
-    def _check_model(self):
-        """Avoid error when updating base module and a submodule extends a
-        model that falls out of this one's dependency graph.
+    def _check_model_update_allowed(self, model_id):
+        """Check if the template's model can be updated.
+
+        Template can be updated only if no property values already exists for
+        this template
         """
         for record in self:
-            with self.env.norecompute():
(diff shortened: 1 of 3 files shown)


foutoucour (approved):

sebastienbeau (approved):
Ok for me

simahawk (approved):
I wonder how important is this order anyway. By default is ordered by ID which means that to control it you have to delete/create items. Am I wrong?

lmignon:
Tests for `base_custom_info` are failing...... :cry: (Not related to my changes even the 10.0 branch is red :tired_face: )

lmignon:
Tecnativa https://github.com/OCA/server-tools/pull/1661/commits/f499379c02ce24be64aa8538540c48ec0a153c91 fixes a bug into base_custom_info due to a change into the ORM

lmignon:
> I wonder how important is this order anyway. By default is ordered by ID which means that to control it you have to delete/create items. Am I wrong?

You are right. But when the user create a custom export for a model, he can specify the order of the column into the exported document by ordering the selected fields with the arrows up and down. This order is preserved since the ir.export.line are created into the same order than the list of selected fields. That's why we can't change the order of ir.export.line otherwise the functionality is lost and the exported documents no more respect the expected order for columns.

pedrobaeza:
> Tecnativa f499379 fixes a bug into base_custom_info due to a change into the ORM

Thanks for that.

PR #1662, opened 2019-09-09
[12.0][FIX]base_jsonify: Preserve initial order on ir.export.line

lmignon:
We can't change the order on ir.export.line since it's used to preserve the order speicifed by the user when creating a export filter into the UI

forward port of #1661


foutoucour (approved):

lmignon:
@sebastienbeau #1661 is approved.. This one is the formard port and could be merged

sebastienbeau (approved):

PR #1665, opened 2019-09-10
[12.0][ADD] base_deterministic_session_gc

nilshamerlinck:
The purpose of this module is to provide a deterministic session garbage collection instead of the default random one.

Please have a look at the [rendered documentation here](https://htmlpreview.github.io/?https://raw.githubusercontent.com/OCA/server-tools/c18b3df64cfa2e7cdcd514483111bfe332521b95/base_deterministic_session_gc/static/description/index.html).


base_deterministic_session_gc/http.py
```diff
@@ -0,0 +1,42 @@
...
+    _logger.debug('Deleting all sessions inactive since %s',
+                  datetime.fromtimestamp(expired_time)
+                  .strftime(DEFAULT_SERVER_DATETIME_FORMAT))
+    for fname in os.listdir(session_store.path):
+        path = os.path.join(session_store.path, fname)
+        try:
+            if os.path.getmtime(path) < expired_time:
+                os.unlink(path)
+        except OSError:
+            pass
```
hparfr: why adding this except case ?
at least it should be logged no ?

nilshamerlinck: This comes straight from Odoo's original code. I guess that it was to mitigate race conditions when 2 processes would be doing the garbage collection in the same time.

2 concurrent garbage collections should not happen with this module if using the scheduled action. But nevertheless, adding logs cannot hurt, and might even help to detect access rights/disk issues in some contexts. Added! Thanks for the suggestion and the review!

hparfr (approved):
LGTM

hkapatel-initos (approved):
did code review and functional test

hkapatel-initos:
it would be great if some one add reviews here

dsolanki-initos:
@dreispt  Can we merge this?

dsolanki-initos:
@dreispt Can we merge this?

PR #1669, opened 2019-09-13
[10.0][FIX] onchange_helper: Values recieved as input must be preserved.

lmignon:
The goal of this lib is to play the onchange methods to get values for the fields not yet filled into the received list of values used to createor update a record. Therefore, the values from the user must be preserved. If we don't process on this way and you play the onchange methods on a sale.order for a new line with 'product_id' and 'product_uom_qty', the product_uom_qty will be reset to 1 when the onchange for product_id will be played and we'll lost the initial value for product_uom_qty

### DIFF
--- a/onchange_helper/models/models.py
+++ b/onchange_helper/models/models.py
@@ -128,6 +128,7 @@ class Base(models.AbstractModel):
         This method reimplement the onchange method to be able to work on the
         current recordset if provided.
         """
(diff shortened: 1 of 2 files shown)


lmignon:
@sebastienbeau Get it!

sebastienbeau (approved):
youhou ! \o/ let's merge it

PR #1671, opened 2019-09-14
[11.0][ADD] excel_import_export, excel_import_export_demo

kittiu:
Back port from 12.0


kittiu:
Redo as [11.0][MIG] from [12.0]

kittiu:
@smartapp-ep would you test this?

PR #1674, opened 2019-09-19
[12.0][autovacuum_mail_message] Migration and renaming

florian-dacosta:
Migration and rename of autovacuum_mail_message.
I have added the possibility to also delete attachments, following a similar logic.
I don't think we should create an abstract module and 2 sub modules because both dependencies (mail and base) are almost always installed and we should not have other modules of this kind with such a similar logic.

Replace previous simple migration of autovacuum_maill_message : 
https://github.com/OCA/server-tools/pull/1615


florian-dacosta:
@bealdav @legalsylvain 
I replaced the old PR by this one, adding the possibility to clean attachments. 
Could you please make a new review here?
@bguillot

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+        for rule in self:
+            if not rule.retention_time:
+                raise exceptions.ValidationError(
+                    _("The Retention Time can't be 0 days"))
+
+    @api.multi
+    def _get_message_domain(self):
+        today = date.today()
+        limit_date = today - timedelta(days=self.retention_time)
+        limit_date = limit_date.strftime(DEFAULT_SERVER_DATE_FORMAT)
```
MeetKD: Can we make limit_date one liner?

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+        elif not subtype_ids and not self.empty_subtype:
+            message_domain += [('subtype_id', '!=', False)]
+        return message_domain
+
+    @api.multi
+    def _get_attachment_domain(self):
+        today = date.today()
+        limit_date = today - timedelta(days=self.retention_time)
+        limit_date = limit_date.strftime(DEFAULT_SERVER_DATE_FORMAT)
+        attachment_domain = [('create_date', '<', limit_date)]
```
MeetKD: Can we make limit_date one liner?

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+            domain += self._get_message_domain()
+        elif self.ttype == 'attachment':
+            domain += self._get_attachment_domain()
+
+        # Case we want a condition on linked model records
+        if self.model_id and self.model_filter_domain:
+            record_domain = safe_eval(self.model_filter_domain,
+                                      locals_dict={'datetime': datetime})
+
+            res_model = self.model_id.model
```
MeetKD:
```suggestion
```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+            res_records = self.env[res_model].with_context(
```
MeetKD:
```suggestion
            res_ids = self.env[self.model_id.model].with_context(
```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+                active_test=False).search(record_domain)
```
MeetKD:
```suggestion
                active_test=False).search(record_domain).ids
```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,137 @@
...
+            res_ids = res_records.ids
```
MeetKD:
```suggestion
```

MeetKD (changes requested):

florian-dacosta:
@MeetKD Comments attended!
Feels free to review if all is ok for you now.

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,135 @@
...
+    @api.constrains('retention_time')
+    def retention_time_not_null(self):
+        for rule in self:
+            if not rule.retention_time:
+                raise exceptions.ValidationError(
+                    _("The Retention Time can't be 0 days"))
+
+    @api.multi
+    def _get_message_domain(self):
+        today = date.today()
```
MeetKD:
```suggestion

```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,135 @@
...
+        limit_date = (today - timedelta(days=self.retention_time)).strftime(
```
MeetKD:
```suggestion
        limit_date = (date.today() - timedelta(days=self.retention_time)).strftime(
```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,135 @@
...
+                              ('subtype_id', '=', False)]
+        elif subtype_ids and not self.empty_subtype:
+            message_domain += [('subtype_id', 'in', subtype_ids)]
+        elif not subtype_ids and not self.empty_subtype:
+            message_domain += [('subtype_id', '!=', False)]
+        return message_domain
+
+    @api.multi
+    def _get_attachment_domain(self):
+        today = date.today()
```
MeetKD:
```suggestion

```

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,135 @@
...
+        limit_date = (today - timedelta(days=self.retention_time)).strftime(
```
MeetKD:
```suggestion
        limit_date = (date.today() - timedelta(days=self.retention_time)).strftime(
```

MeetKD (changes requested):
Minor Suggestions. Overall Looks good

florian-dacosta:
@MeetKD Actually the last suggestions make the line too long (83 char).
I prefer to let it this way as it is in my opinion more readable.

MeetKD:
You are right. I missed it. Let it be more readable. 👍

jarroyomorales (approved):
LGTM! Very useful module! :+1:

jarroyomorales:
Just one comment, there are a couple of files with `# -*- coding: utf-8 -*-`, which can be removed

florian-dacosta:
Thanks for the review @jarroyomorales I just removed the useless lines.

@MeetKD Could you update your review since you seemed to agree with the PR?

etobella (approved):
LGTM :+1:

MeetKD (approved):

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,134 @@
...
+            domain += self._get_attachment_domain()
+
+        # Case we want a condition on linked model records
+        if self.model_id and self.model_filter_domain:
+            record_domain = safe_eval(self.model_filter_domain,
+                                      locals_dict={'datetime': datetime})
+
+            res_ids = self.env[self.model_id.model].with_context(
+                active_test=False).search(record_domain).ids
+            domain += ['|', ('res_id', 'in', res_ids),
```
etobella: This solution is extremely inefficient, if your list of elements is big, this query will be crazy. I will try to propose you a different approach.

florian-dacosta: @etobella 
Actually, before installing this module, you shoud cleanup yourself the messages/attachments according to your rules. Because anyway, the deletion will take for ever and the cron won't ever pass...

For example, if you have 30 millions mail message over 5 years and you decide to keep one year history and install the module, the cron won't be able to delete 24 millions mail message...
In this cases, you should delete your 4 years history as you wish (a script with direct sql queries seems more appropriates, maybe adding some indexes for the occasion...).
And even this way, it will take hours (or days)

And, then, install and parameter the module, so it deletes one day history...each night!

I've done this way for some customers and I have no performance issue AFAIK (with the cron running at night)

I should probably say this in the README. 

Anyway, I am not against another proposal, just need to keep the feature.

etobella: I know, but that's not the problem, Assume you want to clean messages only on sale.orders that are cancelled, then your system will get all the sale.orders cancelled and add this on the query. Every night, you will check all the cancelled orders. Assume you have a million cancelled orders, then imagine how the query will be...

florian-dacosta: Yes, that's right...
I did kind of get around the issue, but I know it is not ideal.
For one customer, I do delete messages for done/canceled pickings, procurements (version 8), etc.. (huge tables) doing something like this : 
`[('state', 'in', ('done', 'cancel')), ('write_date', '<', ((datetime.datetime.now()-datetime.timedelta(days=365)).strftime('%Y-%m-%d'))), ('write_date', '>', ((datetime.datetime.now()-datetime.timedelta(days=375)).strftime('%Y-%m-%d')))]`

Far from perfect, and kind of complicated, but working well.
Once again, if it can be improved, that is better of course.

etobella: I made a refactoring on my branch:
https://github.com/tegin/server-tools/commits/12-autovacuum_message_attachment

Can you check it please :wink:

florian-dacosta: Thanks for the proposal.
I am not familiar with the class `Base`, where do we see that every other models depend on this one?

Overall it looks good, but I don't really see why it would be faster.
I'd need to do some test on a big database about it. I can backport this to v8 to make tests with old and new version and compare.

etobella: Ok! Thanks, the main difference is that we only find the records from the model that has messages with issues.
With the case I checked before: We will find all the sale.orders that are cancelled with messages older than a year, then we search only this messages.

Base is parent from all: https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L516

florian-dacosta: @etobella well, I did some tests with a huge table in v8 and it seems really faster indeed.
Can you make a PR in my branch so I can merge it?
Thanks

etobella: Sure!! Akretion#22

autovacuum_message_attachment/models/base.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    attachment_ids = fields.One2many(
+        'ir.attachment', 'res_id', string='Attachments',
+        domain=lambda self: [('model', '=', self._name)], auto_join=True
```
etobella: It's res_model, not model, that's the main reason for travis failing.

autovacuum_message_attachment/models/vacuum_rule.py
```diff
@@ -0,0 +1,85 @@
...
+             "models into account")
+    model_id = fields.Many2one(
+        'ir.model', readonly=True,
+        compute='_get_model_id',
+        help="Technical field used to set attributes (invisible/required, "
+             "domain, etc...for other fields, like the domain filter")
+    model_filter_domain = fields.Text(
+        string='Model Filter Domain')
+    model = fields.Char(
+        readonly=True,
```
etobella, line 54: set a string in order to not confuse `model_id` and `model` (runtbot is failing for this)

etobella (commented):
There are some issues due to my commit. Sorry :cry:

autovacuum_message_attachment/models/base.py
florian-dacosta: Well strange the test `test_attachment_vacuum_rule` did not fail...

autovacuum_message_attachment/models/autovacuum_mixin.py
```diff
@@ -0,0 +1,71 @@
...
+        for leaf in domain:
+            if not isinstance(leaf, (tuple, list)):
+                record_domain.append(leaf)
+                continue
+            field, operator, value = leaf
+            record_domain.append(
+                ('%s.%s' % (autovacuum_relation, field), operator, value))
+        records = self.env[rule.model_id.model].search(record_domain)
+        return self.search(
+            domain + [('res_id', 'in', records.ids)]
```
florian-dacosta, line 70: I think we have some issues here, because I guess (did not test) this code only work when a rule is link to 1 specific model.
An it is not always the case. The rule coud concerns multiple models, or none. Then, we don't have `model_filter_domain`.
Then `record_domain` is empty., so I guess Odoo will fail in these case.

Anyway I think we have to avoid this logic if we don't have any `model_filter_domain`.
We need to refactore this part.
Something like 
```
   def _get_autovacuum_records_model(self, rule):
        domain = self._get_autovacuum_domain(rule)
        if rule.model_id and rule.model_filter_domain:
            record_domain = safe_eval(rule.model_filter_domain,
                                  locals_dict={'datetime': datetime})
            autovacuum_relation = self._autovacuum_relation
            for leaf in domain:
                if not isinstance(leaf, (tuple, list)):
                    record_domain.append(leaf)
                    continue
                field, operator, value = leaf
                record_domain.append(
                    ('%s.%s' % (autovacuum_relation, field), operator, value))
            records = self.env[rule.model_id.model].search(record_domain)
            domain +=  [('res_id', 'in', records.ids)]
        return self.search(
            domain)
```

autovacuum_message_attachment/models/base.py
florian-dacosta: We probably should add test for rule with no linked model..

autovacuum_message_attachment/models/autovacuum_mixin.py
etobella: _get_autovacuum_records_model is only called if we have a domain and one model, You can check on `_get_autovacuum_records`.

I made the changes locally and the tests from other modules are executed properly. I will test on travis.

florian-dacosta: Oh right, should be good then

etobella: https://travis-ci.org/tegin/server-tools/builds/640488894?utm_medium=notification&utm_source=email

Here you can check that with only my change travis is fixed :wink:

florian-dacosta: Great, can you add the string to the `model_id` field and make a pull request?
Thanks for the work

etobella: akretion#23

etobella:
Well, now all is :green_apple::green_apple::green_apple::green_apple::green_apple::green_apple:

@JordiBForgeFlow

bealdav (approved):

bealdav:
Please @pedrobaeza could you merge, please, thanks

pedrobaeza:
Sorry, I'm not anymore in charge of this and I want to let PSC/module maintainers this responsibility.

bealdav:
Sorry but i've seen you here as PSC 
https://odoo-community.org/psc-teams/tools-30
May be I have missed smth ?

pedrobaeza:
Yeah, I'm PSC of this repo, but not interested in this specific module. Starting with, I will ask for squashing commits, but I let the interested reviewers to impose their preferences. @bealdav I think you should apply for some PSC, as you have worked a lot on some of them.

bealdav:
OK thanks

etobella:
@florian-dacosta can you squash the commits so everything is simpler and easier to manage?

etobella:
@florian-dacosta Can you rebase in order to fix runbot. There was an issue (#1770)

LoisRForgeFlow (approved):
Code review, LGTM :+1:

florian-dacosta:
@LoisRForgeFlow 
Can you merge then? Thanks

LoisRForgeFlow:
sure!

/ocabot merge

PR #1678, opened 2019-09-25
[10.0][FIX] base_exception: browse records were not updated

rousseldenis:
As recordset are immutables, update() does not work on dict
that contain such recordsets.


rousseldenis:
@lmignon @guewen

pedrobaeza (approved):
Fast-tracking this to fix the branch

lmignon (approved):

Issue #1676, opened 2019-09-20
[11.0] BUG: Auditlog - related fields are not stored (with store=True)


vbezl:
If I create a rule with Type="Full log" - some related fields are not stored to database (with store=True of course!) - they have null values instead.
If I use "Fast log" instead - they are stored just fine.
Anyone can help? (it took me a lot of time while I found that it was because of auditlog module!)
Thanks!

sebalix:
It is the same issue that https://github.com/OCA/server-tools/issues/1134 (not clear here but the field in this issue is a related/computed one with `store=True`).
You can test this PR https://github.com/OCA/server-tools/pull/1686, it should fix your issue.

PR #1399, opened 2018-10-17
[11.0][auditlog] Logging `create` events on `account.move.line` breaks invoice validation and Journal Entry validation seperately

smangukiya:
similar ref issue #1134

Issue #1134, opened 2018-01-24
[11.0][auditlog] Logging `create` events on `account.move.line` breaks invoice validation


MaxyMoos:
## Minimum setup to reproduce the issue:

* Create a new Odoo V11 database
* Install `account_invoicing` and `auditlog`
* Create an auditlog rule, logging Create/Write/Delete events on `account.move.line` items
* Create a Customer Invoice
* Try to validate the invoice

## Problem

A popup appears with the error message:
"Cannot create moves for different companies."

## Analysis

It appears that the account.move.line instances that are created during the invoice validation process are missing the `company_id` field value, hence the error message displayed.

From what I've gathered from debugging, the problem is coming from the `create_full` function in the auditlog module:

```python
@api.model
@api.returns('self', lambda value: value.id)
def create_full(self, vals, **kwargs):
    self = self.with_context(auditlog_disabled=True)
    rule_model = self.env['auditlog.rule']
    new_record = create_full.origin(self, vals, **kwargs)
    new_values = dict(
        (d['id'], d) for d in new_record.sudo()
        .with_context(prefetch_fields=False).read(list(self._fields)))
    rule_model.sudo().create_logs(
        self.env.uid, self._name, new_record.ids,
        'create', None, new_values, {'log_type': log_type})
    return new_record
```

When we call the `origin` method to create `new_record`, everything is OK: printing `new_record.company_id` gives out the correct value.

The problem is that the generation of the `new_values` dict, and especially the call to `new_record.read` does remove the `company_id` value from `new_record`.
Maybe because it's not yet commited in the database but we still try to read data from it. I've stopped my investigations here since I've started to get in a bit over my head.

MaxyMoos:
After experimenting a bit more, it would appear that all the fields using both the `related=<class>` option and the `store=True` option are afflicted by this issue.
Another example being the fields `journal_id`, `user_type_id` or `date` on the account.move.line model.

sebalix:
Maybe this is due to the `.with_context(prefetch_fields=False)`, try to remove it just to see if it works?

MaxyMoos:
No, unfortunately it seems like the problem is still there whether I drop the call to `with_context` altogether or if I just remove the `prefetch_fields` parameter from the call.

MaxyMoos:
As a quick-and-dirty workaround, here's what I've done so far - feedback appreciated by Odoo's inner workings experts as to what can be done to improve this:

```python
@api.model
[...]

PR #1686, opened 2019-10-01
[11.0][FIX] auditlog: log computed fields stored in db as expected

sebalix:
Fixing #1134 and #1676 + supersedes #1399.
Odoo stores values of computed fields at the end of the transaction only, as such performing a 'read()' to make a data snapshot on the record created in the current transaction doesn't return the expected result regarding these fields. Also as a side-effect 'read()' alters the environment cache and break the values on the record inducing issues in the whole user transaction/workflow.

This fix replaces the use of 'read()' to do the data snapshot directly from the cache of the record (computed values are already there).

I also backported from 12.0 the split of README and a fix regarding the name of the common test case to be compatible with pytest, nosetest, etc.


StephanRozendaal (approved):
LGTM (tested functionally and code review)

jarroyomorales (approved):
LGTM!

PieterPaulussen (approved):

PR #1687, opened 2019-10-01
[ADD] search function supporting case insensitive login

gfcapalbo:
Case insensitive module allows to create/write lowercase logins in DB.
After that the input is sanitized to be allways lowercase.

This extension adds search, so that it is not necessary anymore to lowercase all inputs, but the input sanitization is built-into the search function if needed.

Very useful for projects for with several login methods/interfaces.

### DIFF
--- a/auth_user_case_insensitive/models/res_users.py
+++ b/auth_user_case_insensitive/models/res_users.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2015-2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
-
+import collections
 from odoo import api, fields, models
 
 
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

auth_user_case_insensitive/models/res_users.py
```diff
@@ -32,3 +32,24 @@ def write(self, vals):
         if vals.get('login'):
             vals['login'] = vals['login'].lower()
         return super(ResUsers, self).write(vals)
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        """
```
NL66278: Format this according to PEP rules for doc comments (using full sentences, If possible one line, etc.).

gfcapalbo: now compliant with https://www.python.org/dev/peps/pep-0257/

NL66278 (approved):
:+1:

Issue #1429, opened 2018-11-17
refactoring module_auto_update


legalsylvain:
Hi, 

For some projects, I have to generate a hash of modules, to ensure that code did not changed. (It's about certification process, like, for exemple, in France, the connexion between point of sale and scale).

For that purpose, I have to compute hash, based on source code (.py, .xml, but .js too). All the code to compute this hash is in the module ``module_auto_update``, so I propose : 

- to create a new generic module in server tools named module_hash, that provides all the algorigthm, currently in the file ``addon_hash``
https://github.com/OCA/server-tools/blob/12.0/module_auto_update/addon_hash.py
- to make ``module_auto_update`` depends on this new module.

Are you agre with that proposal ?

CC : maintainer : @sbidoul

legalsylvain:
Extra question : for the time being, seems that 

- if you add / change test files. (``./tests/*.py`` or ``demo/*.xml``), the module_to_update consider the module as 'to be updated'. is it a desired behaviour ?

https://github.com/OCA/server-tools/blob/12.0/module_auto_update/models/module.py#L19

pedrobaeza:
The problem here is the same problem when the other time it was factorized: you are going to add an extra dependency for current systems, and this is going to break them. Why not simply depend on this module? It's harmless if you don't trigger the update through one of the mechanism.

legalsylvain:
Hi @pedrobaeza : thanks for your answer.

> Why not simply depend on this module? It's harmless if you don't trigger the update through one of the mechanism.

because it's an other feature. (computing hash for certification process). So, I don't see as a good idea to force users that want the certification module, to install a module they (maybe) don't want. (even if it is disabled by default and harmless).

> you are going to add an extra dependency for current systems, and this is going to break them

I don't understand why adding dependency will break current systems. Could you be more precise ? what is the "other time" ? Maybe I missed something.
But anyway, if it's not possible in stable version, we can imagine to create the new module for 12 version, and to mention in the roadmap of the module_auto_update that it should depends on the new generic module in V13.

pedrobaeza:
Your idea is to split `module_auto_update` in 2 modules: `addon_hash` (I prefer `module_hash` any way), and `module_auto_update`, and make `module_auto_update` dependent on the first one. Current systems will break when updating as the new module is not installed (and even maybe not available).

Although I don't like to have to install 2 modules for the feature I use (auto-update), something tolerable can be:

[...]

PR #1689, opened 2019-10-01
[13.0][MIG][WIP] module_auto_update migration to 13

schout-it:
@sbidoul I'm migrating this module, but found #1429 
should the module still be split up?

Also, is there a reason the hashes are stored in the system parameters and not on ir.module.module?
If a lot of modules are installed, this could get messy.


sbidoul:
@schout-it I remember trying to split the module or storing the hashes on ir.module.module created several reliability issues. Especially when module_auto_update updates itself, but not only.

In the end I don't use this module anymore and warmly recommend using `click-odoo-update` (in [click-odoo-contrib](https://pypi.org/project/click-odoo-contrib/)) which is much more reliable and battle tested.

gustavovalverde:
Hi @schout-it @sbidoul was this PR dropped? Just to know if I can take over it

sbidoul:
@gustavovalverde I don't maintain this module anymore and I don't recommend using it. Use [click-odoo-update](https://github.com/acsone/click-odoo-contrib#click-odoo-update-beta) instead: it's the same principle but much more robust.

gustavovalverde:
@sbidoul Oh, my bad. I actually thought that click-odoo-update used this one as a dependency.

sbidoul:
@gustavovalverde at some point it did depend on this module, but it had stability issues and now it's a standalone tool.

PR #1695, opened 2019-10-04
12 add base substate

mourad-ehm:
This module provide an abstract models to manage customizable
substates to be applied on different models (sale order, purchase, ...).

It is not useful for itself. You can see an example of implementation
in the 'sale_substate' module. (sale-workflow repository).
Applied to sale
https://github.com/OCA/sale-workflow/pull/955


mourad-ehm:
Hi @bealdav, @pedrobaeza , @hparfr , @rvalyi,
Could you review this PR please ? There is an error on temporary test model "base.substate.test.sale" and "base.substate.test.sale.line" in travis. Have you an idea to solve it ?

base_substate/models/base_substate.py
```diff
@@ -0,0 +1,71 @@
...
+from odoo import fields, models
+
+
+class BaseSubstateType(models.Model):
+    _name = "base.substate.type"
+    _description = 'Base Substate Type'
+    _order = 'name asc, model asc'
+
+    # Data in this object should be created by import as technical data
+    # in specific module ex : sale_subsatate
```
hparfr, line 13: s/subastate/substate

base_substate/models/base_substate.py
```diff
@@ -0,0 +1,71 @@
...
+    name = fields.Char('Substate Name', required=True, translate=True)
+    model = fields.Selection(selection=[], string='Apply on', required=True)
+    target_state_field = fields.Char(
+        required=True, help='Technical target state field name.'
+        ' Ex fore sale order "state" for other "status" ... ')
```
hparfr, line 18: s/fore/for

base_substate/models/base_substate.py
```diff
@@ -0,0 +1,71 @@
...
+    sequence = fields.Integer(
+        index=True,
+        help="Gives the sequence order when applying the default substate",
+    )
+    target_state_value_id = fields.Many2one(
+        'target.state.value',
+        string='Target State Value',
+        ondelete='restrict')
+    active = fields.Boolean(default=True)
+    mail_template_id = fields.Many2one(
```
hparfr, line 63: It's seams not generic for me to put a dependency on mail here

mourad-ehm:
Hi, @jgrandguillaume, @gurneyalex, @kittiu 
What do you think about this feature ?

hparfr (changes requested):

rvalyi:
Hello @mourad-ehm, I've goot a question: did you consider compatbility with Acsone state machine https://github.com/acsone/scobidoo I feel like it's a nice way to manage sub-states so I want to make sure you are aware of this module. How is your substate module better or complementary to Scobidoo?

kittiu:
@mourad-ehm I think this is interesting concept.
May be can use also with base_tier_validation ?
https://github.com/OCA/server-ux/tree/12.0/base_tier_validation
cc @ps-tubtim @LoisRForgeFlow 
By the way, should this module be in server-ux ?

mourad-ehm:
> Hello @mourad-ehm, I've goot a question: did you consider compatbility with Acsone state machine https://github.com/acsone/scobidoo I feel like it's a nice way to manage sub-states so I want to make sure you are aware of this module. How is your substate module better or complementary to Scobidoo?

Hi @rvalyi, I did not know this module. But when I reviewed de code I think that most goal of scobidoo module is to add workflow management (replace the old openerp workflow) and accessoritly manage sub states. base_substate module is a very simple to configure and use (no need to define a yaml spec of workflow). But it not cover a  workflow management.

mourad-ehm:
Hi @kittiu, 
Ok to move this module to server-ux.
Could some one give me solution for this error ? https://travis-ci.org/OCA/server-tools/jobs/632221691#L6529
cc @ps-tubtim @LoisRForgeFlow

kittiu:
@mourad-ehm may be just a GUESS, but as I tested around, I found the problem is with the test of `database_cleanup`
And it happen with any module that did create testing models on its test script, i.e.,
* Run test of database_cleanup and base_exception together
* Run test of database_cleanup and base_substate together

Seem like, the @post_install(True) (just guess), make it run after every module in the repo is installed. And while database_cleanup is cleaning the table, i.e., `base.substate.test.sale.line` is not there...

I am not sure how to fix it too. Sorry...

May be, moving your module out of this server-tools  will already fix it? :)

mourad-ehm:
thanks @kittiu. 
Module moved to https://github.com/OCA/server-ux/pull/121

PR #1700, opened 2019-10-13
Disable the hook for fresh install

zeroheure:
Purpose of hook.py is to upgrade from a previous version, but it block a fresh install.


pedrobaeza:
The hooks is only activated if you have the table coming from v9:

https://github.com/OCA/server-tools/blob/c7b8902ea2386fa720b18bcd7408d51934cc2f14/base_exception/hooks.py#L12

so there's no sense in this affirmation.

zeroheure:
True, but you need openupgrade libs to install.

zeroheure:
Sorry, on a totaly new clean fresh install it work as expected.

PR #1701, opened 2019-10-14
[10.0][IMP] base_view_inheritance_extension : Add new feature 'target_position'

remihb:
By default, inheritance with spec **position="move"** append the node inside the target.
The spec **target_position** let you choose what you want to do :

- **inside**: append the node inside the target
- **before**: append the node before target
- **after**: append the node after target

By default, if no value is provide we use the standard behavior **inside**.

### DIFF
--- a/base_view_inheritance_extension/README.rst
+++ b/base_view_inheritance_extension/README.rst
@@ -33,10 +33,18 @@ Move an element in the view
 
 .. code-block:: xml
 
-    <xpath expr="$xpath" position="move" target="$targetxpath" />
+    <xpath expr="$xpath" position="move" target="$targetxpath" target_position="$target_position"/>
 
-This can also be used to wrap some element into another, create the target
-element first, then move the node youwant to wrap there.
+Availables values for target_position attribute:
+
+* inside (default if omitted)
+* after
+* before
+
+
+This can also be used to wrap some element into another, create the new target
+element first, then move the element you want inside (or before/after) the new
+element.
 
 Add to values in a list (states for example)
 --------------------------------------------
@@ -60,7 +68,6 @@ Known issues / Roadmap
 ======================
 
 * add ``<attribute operation="json_dict" key="$key">$value</attribute>``
-* support ``<xpath expr="$xpath" position="move" target="xpath" target_position="position" />``
 * support an ``eval`` attribute for our new node types
 
 Bug Tracker
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -2,7 +2,7 @@
 # © 2016 Therp BV <http://therp.nl>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 from lxml import etree
-from odoo import api, models, tools
+from odoo import api, models, tools, _
 from odoo.tools.safe_eval import safe_eval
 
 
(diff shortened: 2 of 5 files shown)


remihb:
As #1056 in 9.0 but for version 10.0 with adding tests.

base_view_inheritance_extension/README.rst
```diff
@@ -33,10 +33,18 @@ Move an element in the view
...
 .. code-block:: xml
 
-    <xpath expr="$xpath" position="move" target="$targetxpath" />
+    <xpath expr="$xpath" position="move" target="$targetxpath" target_position="$target_position"/>
 
-This can also be used to wrap some element into another, create the target
-element first, then move the node youwant to wrap there.
+Availables values for target_position attribute:
+
+* inside (default if omited)
```
hbrunn: omitTed

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -125,7 +125,21 @@ def inheritance_handler_xpath(self, source, specs, inherit_id):
         target_node = self.locate_node(
             source, etree.Element(specs.tag, expr=specs.get('target'))
         )
-        target_node.append(node)
+
+        target_position = specs.get('target_position') or 'inside'
```
hbrunn: pass `'inside'` as second parameter to `get`

hbrunn (changes requested):
thanks, only small things found

base_view_inheritance_extension/README.rst
remihb: > omitted

 OK

base_view_inheritance_extension/models/ir_ui_view.py
remihb: > target_position = specs.get('target_position', 'inside')

OK

hbrunn (approved):

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -125,7 +125,21 @@ def inheritance_handler_xpath(self, source, specs, inherit_id):
...
+            target_node.addnext(node)
+        elif target_position == 'before':
+            target_node.addprevious(node)
+        elif target_position == 'inside':
+            target_node.append(node)
+        else:
+            self.raise_view_error(
+                _("Invalid target_position attribute: '%s'") % target_position,
+                inherit_id,
+                context=self.env.context
```
hbrunn: this needs to go away: https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/ir/ir_ui_view.py#L460

hbrunn (changes requested):

base_view_inheritance_extension/models/ir_ui_view.py
remihb: I don't understand, I have to remove the **raise** itself or to raise in another way ?

hbrunn: I mean the context parameter, that's not in the function signature

remihb: Ah sure, it was late, I misread the code ^^'

hbrunn (approved):

remihb:
Any news for this PR ?

thomaspaulb (approved):
Really well done, including the demo data and the tests. Tested this in an installation and it works.

You could still mention your name in the 'contributors' section of the README.

hbrunn:
/ocabot merge

thanks! Note however that we've deprecated this in favor of the Odoo way of doing it for later versions: https://github.com/OCA/server-tools/blob/12.0/base_view_inheritance_extension/README.rst#usage

remihb:
> This PR looks fantastic, let's merge it!

Cool :+1: 

> thanks! Note however that we've deprecated this in favor of the Odoo way of doing it for later versions: https://github.com/OCA/server-tools/blob/12.0/base_view_inheritance_extension/README.rst#usage

Sure, but still needed for older versions :wink:

PR #1705, opened 2019-10-18
[13.0][MIG] onchange_helper

Laurent-Corron:


onchange_helper/models/base.py
```diff
@@ -0,0 +1,67 @@
...
+        # we take the value of the field
+        # If self is an empty record we will have an empty value
+        if self:
+            self.ensure_one()
+            record_values = self._convert_to_write(
+                {
+                    field_name: self[field_name]
+                    for field_name, field in self._fields.items()
+                }
+            )
```
lmignon, line 48: IMO, we should adapt the code to at least don't extract computed field on existing record.

rousseldenis: @Laurent-Corron @lmignon Is this still relevant ?

rousseldenis (approved):
Code and functional review

rousseldenis:
@pedrobaeza 

This migration is immediate (see migration commit). Functionnaly tested

pedrobaeza:
Well, I'm not a very active user of this module, and I see one comment still unanswered. Better to close first that before merging.

And luckily, this can disappear on next version!

onchange_helper/models/base.py
rousseldenis: @lmignon @Laurent-Corron

lmignon: We can leave this code as is..

lmignon (approved):

rousseldenis:
@pedrobaeza Done!

PR #1708, opened 2019-10-23
IMP print failed query to help debugging

bealdav:
please @legalsylvain

### DIFF
--- a/sql_request_abstract/models/sql_request_mixin.py
+++ b/sql_request_abstract/models/sql_request_mixin.py
@@ -5,12 +5,15 @@
 
 import re
 import uuid
+import logging
 from io import BytesIO
 import base64
 from psycopg2 import ProgrammingError
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+
+logger = logging.getLogger(__name__)
 
 
 class SQLRequestMixin(models.AbstractModel):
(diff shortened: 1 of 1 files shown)


sql_request_abstract/models/sql_request_mixin.py
```diff
@@ -242,6 +242,8 @@ def _check_execution(self):
             self.env.cr.execute(query)
             res = self._hook_executed_request()
         except ProgrammingError as e:
+            # it helps a lot for debugging to have raw query string without \n
+            print(query)
```
pedrobaeza: Why not emit it as logger.warning or similar?

bealdav: logger'll display `\n` not return to the line when occurs, isn't it ?

pedrobaeza: I don't think so if you pass directly the query, or use `from pprint import pprint; pprint(query)`

foutoucour: You might want to consider `logging.exception` which is a good candidate for this case.
`logging.exception` is a log at error level that embed the previous error raised.
In the following example, I embed a `\n` to double check this case
``` python
In [8]: try:
   ...:     1/0
   ...: except:
   ...:     logger.exception("toto\ntoto")
   ...:
toto
toto
  File "<ipython-input-8-b43f6c8f4df1>", line 2, in <module>
    1/0
ZeroDivisionError: division by zero
```
test done in ipython with python 3.7
what do you think ?

bealdav: I think it's nice. Thank you

legalsylvain (approved):
LGTM for me. Thanks for the improvment.

@pedrobaeza, @foutoucour, could you update your review ? 

regards.

pedrobaeza (approved):

foutoucour (approved):

PR #1709, opened 2019-10-23
[10.0][FIX] Fix onchange_helper in case of M2M fields + add unit test.

acsonefho:
**Current issue:**
When a M2M field is updated (and this M2M is into the `_cache`) with the onchange_helper, the result `dict` contains something like:
`[(5,), (1, record_id, dict_values), ...]`.
If we use this to do a create (to fill this M2M field), this field will be empty because the `1` magic number is used to update an existing record. As we do a `5` before, this record couldn't exist.

**Fix:**
Just add the `4` (to add an existing record) before the `_cache` check, only in case of M2M and if this M2M record exists.

**Note:**
The issue doesn't happens if M2M records are not into the cache. So if you check my unit test: if I didn't do `expected_categs.mapped("name")` => no issue.

**Real example:**
To have a real example (other than the one created into the unit test): just try to create a new `sale.order.line` using `play_onchanges(...)`. You will see that `tax_id` (M2M field) is always empty. But the values returned by the `play_onchanges(...)` contains some values for this `tax_id` (but into the format `[(5,), (1, record_id, dict_values), ...]` instead of `[(5,), (4, record_id), ...]`).

### DIFF
--- a/onchange_helper/models/models.py
+++ b/onchange_helper/models/models.py
@@ -77,6 +77,12 @@ class Base(models.AbstractModel):
             result = [(5,)]
             for record in value:
                 vals = {}
+                # We have to check first if the record already exists
+                # (only in case of M2M).
+                if field.type == "many2many" and not isinstance(
+                        record.id, models.NewId):
+                    result.append((4, record.id))
+                    continue
                 for name in record._cache:
                     if name in models.LOG_ACCESS_COLUMNS:
                         continue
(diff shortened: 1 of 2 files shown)


acsonefho:
What do you think about that @lmignon ?
Thank you!

lmignon (approved):
Good catch LGTM

Cedric-Pigeon (approved):
Code review

PR #1710, opened 2019-10-29
[12.0][ADD] auto_backup_s3: database auto backup to s3

andriisem:
This module make auto backup to aws s3
will it be useful for the community? if yes then I will continue to improve it

need review


Vartkat:
Yes it will be usefull. S3 is not just AWS, it's used by several other providers.

PR #1711, opened 2019-10-30
[12.0][FIX] base_jsonify: In odoo 12, Date and Datetime fields values are d…

lmignon:
…atetime objects

Convert fields.Date and fields.Datetime values to string when serializing to json

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -3,7 +3,7 @@
 # Raphaël Reverdy <raphael.reverdy@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, models
+from odoo import api, fields, models
 from odoo.exceptions import UserError
 from odoo.tools.translate import _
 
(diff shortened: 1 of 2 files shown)


simahawk (approved):

acsonefho (approved):
Code LGTM

rousseldenis (approved):

base_jsonify/models/models.py
```diff
@@ -71,6 +71,10 @@ def jsonify(self, parser):
                     value = rec[field_name]
                     if value is False and field_type != 'boolean':
                         value = None
+                    elif field_type == "date":
+                        value = fields.Date.to_string(value)
+                    elif field_type == "datetime":
+                        value = fields.Datetime.to_string(value)
```
hparfr, line 77: I think it's better to be compliant with ISO 8601 here.
Date is ok,  but date_time defined here https://github.com/odoo/odoo/blob/12.0/odoo/tools/misc.py#L568
is with a space as separator between dates and time. Plus it's lacking timezone.

https://www.w3.org/TR/NOTE-datetime

sbidoul: @lmignon have you seen this comment? Sounds important to me too.

lmignon: @sbidoul Yes I see it, I'll provide a new PR to fix this.

PR #1712, opened 2019-10-30
[13.0][FIX] base_jsonify: In odoo 12, Date and Datetime fields values are d…

lmignon:
…atetime object

Convert fields.Date and fields.Datetime values to string when serializing to json

forward port of #1711


lmignon:
fast tracking since already approved in #1711 and travis is :green_heart:

PR #1713, opened 2019-10-30
[ADD] attachment_unindex_content: Disable the indexation of files

ebirbe:
This module is intended to disable the indexation of data on the ir.attachment model.

Attachment model has a field called 'index_content' where the content of the attachment is read and stored directly in the database. This field is useful in order to search content of a file. But most of cases it is not used, so, you can install this module in order to:

- **Avoid Duplicating Data:** Because indexation extracts text content from files and put it on the database in order it could be searched, but this implies you have the file data in your `filestore` directory, and also part (or sometimes all) of that data in your database too.
- **Improve Performance:** Since not all indexed files are plain text, they require extra process to read them.

Maybe you could try to uninstall modules like `document` in order to disable its indexation features, but you could face the uninstallation of other modules that could be useful for you (e.g,  hr_recruitment` depends on that).

But even if you don't have `document` installed, you'd still have plain text content indexation by default. 

Using this module you will not require to uninstall any module to disable the attachment content indexation, because we directly disable it at `ir.attachment` base.

/cc @moylop260


attachment_unindex_content/README.rst
```diff
@@ -0,0 +1,86 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/server-tools-12-0/server-tools-12-0-attachment_unindex_content
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/149/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module disables the indexation of attachments
```
moylop260: "attachment model has a field called 'index_content' where the content of the attachment is read and stored directly in the database. This is useful in order to search content of a file. But most of cases it is not used, so you can install this module in order to avoid duplicating data (filestore and database) and improve performance since that not all files are plain text and it requires extra process to read them"

moylop260 (approved):
👍

pedrobaeza:
Isn't the indexation done by `document` module and you only need to not install it?

ebirbe:
> Isn't the indexation done by `document` module and you only need to not install it?

Hello @pedrobaeza the indexation is being called directly from the [base module](
https://github.com/odoo/odoo/blob/9beebf2b54959727407f3959e26cd1ee874c3307/odoo/addons/base/models/ir_attachment.py#L265). It will occur even if we do not install `document`.

This is a query over a database without `document` module installed:

Área de trabajo 1_750

The `document` module only adds indexation for [additional file formats](https://github.com/odoo/odoo/blob/9beebf2b54959727407f3959e26cd1ee874c3307/addons/document/models/ir_attachment.py#L12).

Adittionally, there will be cases were you will want to have this feature of no indexation, but without uninstalling another modules that depends on `document`. That's the case of `hr_recruitment` for example.

So, this module disables indexation directly from the base, and without having to lose other functionalities.

pedrobaeza:
Thanks for the explanations. I think they are very interesting to be added to the README as well.

ebirbe:
> I think they are very interesting to be added to the README as well.

Thank you, I've updated the documentation.

moylop260:
@pedrobaeza 
Do you have any new comment?

guewen (approved):

moylop260:
Thanks for review @guewen and @pedrobaeza 

/ocabot merge

Since that it has the `approved` tag

PR #1714, opened 2019-10-31
[12.0][IMP] base_jsonify: Serialize Date and Datetime into ISO 8601 format

lmignon:
follows discussion https://github.com/OCA/server-tools/pull/1711#pullrequestreview-309157848

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -3,8 +3,6 @@
 # Raphaël Reverdy <raphael.reverdy@akretion.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-import pytz
-
 from odoo import api, fields, models
 from odoo.exceptions import UserError
 from odoo.tools.translate import _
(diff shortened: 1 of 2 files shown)


lmignon:
@hparfr @sbidoul here it's the fix following your comments on #1711 (review)

sbidoul (approved):

hparfr (approved):

hparfr:
Thanks

lmignon:
@moylop260 Thank you for the merge but I'm very annoyed  that these change are merged these  without incrementing the version......  
@sbidoul How can we create a new version?  Do we have a new version published on Pypi If I make a new PR to manually increment the version number and we use the `ocabot merge` command?

sbidoul:
@lmignon it is published, just with a .dev version.
But indeed doing ocabot merge without version bump should be exceptional.
To get a round version number on PyPI, just push a commit with the version bump to the main branch.

lmignon:
ok thank you @sbidoul

PR #1715, opened 2019-10-31
[13.0][IMP] base_jsonify: Serialize Date and Datetime into ISO 8601 format

lmignon:
follows discussion #1711 (review)

### DIFF
--- a/base_jsonify/tests/test_get_parser.py
+++ b/base_jsonify/tests/test_get_parser.py
@@ -82,6 +82,11 @@ class TestParser(TransactionCase):
                 "date": fields.Date.from_string("2019-10-31"),
             }
         )
+        # put our own create date to ease tests
+        self.env.cr.execute(
+            "update res_partner set create_date=%s where id=%s",
+            ("2019-10-31 14:39:49", partner.id),
+        )
         expected_json = {
             "lang": "en_US",
             "comment": None,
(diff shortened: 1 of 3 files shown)


base_jsonify/tests/test_get_parser.py
```diff
@@ -100,8 +100,8 @@ def test_json_export(self):
                     "email": None,
                 }
             ],
-            "create_date": fields.Datetime.to_string(partner.create_date),
-            "date": fields.Date.to_string(partner.date),
+            "create_date": partner.create_date.isoformat(),
```
sbidoul: is there timezone information in there?

lmignon: yes there is. https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

image

sbidoul: Sure, but...

```python
>>> env["res.users"].search([], limit=1).create_date.isoformat()
'2019-04-05T11:31:21.609514'
```

has no timezone information.

lmignon: indeed... , I'll fix it

lmignon: @sbidoul fixed

sbidoul: LGTM. I suppose it's safe to assume all values converted by this function are in utc? Because `replace` is not going to do timezone conversion I think.

sbidoul (approved):

hparfr (approved):
LGTM

PR #1719, opened 2019-11-05
[FIX] don't use patch_method in tests, this confuses overrides

hbrunn:
the test code for example destroys `server_environment_ir_config_parameter` because the patching restores the original function, not the overriden one

### DIFF
--- a/auth_session_timeout/tests/test_ir_config_parameter.py
+++ b/auth_session_timeout/tests/test_ir_config_parameter.py
@@ -4,13 +4,13 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.tests import common
+from odoo.tools.cache import STAT
 
 
 class TestIrConfigParameter(common.TransactionCase):
 
     def setUp(self):
         super(TestIrConfigParameter, self).setUp()
-        self.db = self.env.cr.dbname
         self.param_obj = self.env['ir.config_parameter']
         self.data_obj = self.env['ir.model.data']
         self.delay = self.env.ref(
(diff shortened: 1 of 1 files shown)


auth_session_timeout/tests/test_ir_config_parameter.py
```diff
@@ -30,47 +30,42 @@ class TestIrConfigParameterCaching(common.TransactionCase):
...
+        """First call should actually run the function."""
+        cache = self._cache('_auth_timeout_get_parameter_delay')
+        cache_misses = cache.miss
         self.param_obj._auth_timeout_get_parameter_delay()
-        self.assertTrue(self.get_param_called)
+        self.assertEqual(cache.miss, cache_misses + 1)
 
     def test_auth_timeout_get_parameter_ignored_urls_cache(self):
+        """First call should actually run the function."""
         """It should cache the parameter call."""
```
NL66278: This line is causing the travis error. You cannot have a second doc string.

NL66278:
Just wondering wether it would not be better to just backport the 11.0 version of auth_session_timeout, as suggested in the README of this module, as the current setup is apparently not working since commit da1f153 to Odoo. Well anyway, with the one line fixed that is making Travis red, this at least should make tests working again.

NL66278 (changes requested):
Looks good, but very small change needed.

hbrunn:
thanks, done. The backport should happen indeed, but is out of scope for now

NL66278:
@hbrunn Maybe we can have a look at this one... https://github.com/OCA/server-tools/pull/1540/files

NL66278:
Anyway changed method still using the same ir_config_parameter changes and tests, so the backport and this PR are both needed.

NL66278 (approved):
:+1:  LGTM. Apparant errors are due to travis failing.

gurneyalex (approved):

PR #1722, opened 2019-11-13
12.0 mig mail send bcc

legalsylvain:
Migrate a simple module from GRAP V8 incubator repository.

This module overload the send of email in odoo. For each mail sent, odoo will add, in the "Bcc" field, the email present in the "From" field.

Consequently, when the users are sending mail via Odoo to a partner (customer, supplier), they will receive a copy of the email.
Also, if the partner answer to the email, and your mail provider is handling correctly thread (as Gmail does), you'll have the whole history of the mails in your mailbox.

CC : @quentinDupont


etobella:
Wouldn't be better to add this on OCA/social?

legalsylvain:
Hi @etobella : Thanks for your comment. Well, I don't know. 
let's see what PSC members of OCA/social are thinking on that point.
CC : @OCA/social-maintainers 

otherwise, any comment on that PR ? 

thanks.

pedrobaeza:
Yes, please, this should go to OCA/social, and I think there's something similar already (used in OCA instance), and other module with `bcc` surname but other purpose.

etobella:
Well, IMO this module has more sense than the original one from OCA/social

https://github.com/OCA/social/tree/8.0/base_mail_bcc

So, if it goes to social it is :+1:  for me

legalsylvain:
Thanks for your comments.

moved into OCA/social.

Closing this PR. regards.

PR #1726, opened 2019-11-19
[12.0][ENH] excel_import_export, faster export

kittiu:
Previously export is slow on large number of rows.
This fix ensure that sheet.insert_rows is called only once for all rows,
instead of insert 1 row for every rows


kittiu:
This can be fast merge.
May I also ask, as I am maintainer of this, I can use ocabot merge xxx, right?
This is a small enh to add 1 for last digit version, is xxx = minor (or patch)?

PR #1727, opened 2019-11-21
[FIX] [11.0] base_name_search_improved: fix TypeError exception

SamuelBarron:


etobella:
Why are you making this change? Could you give some information in order to check?

gurneyalex:
fixed in 05783c64cb

PR #1728, opened 2019-11-22
[ADD] ir_sequence_standard_default: Use Standard instead No Gap sequence

ebirbe:
Set the implementation to "Standard" in all your current Sequences
(ir.sequence) and all new sequences are created as "Standard" by default
instead of "No Gap" implementation.

**What's the problem Sequences with "No Gap" Implementation?**

"No Gap" is the default value of sequences in Odoo. However, this kind of
sequences cause more locks and can turn a database slow.

Taking as example an invoice, if you assign an invoice number to one record,
but it sill not finish the process, this process must end in order to another
invoice could assign a new number and there was no gaps between the invoice
numbers. It seems to be good at first sight. But the problem start when there
is and chained process.

Imagine that there is one user that executes a process that produces 100
invoices and these at the same time produces 100 journal entries that also use
a consecutive (no gap) sequence. And also those invoices are sent to sign with
and external institution (that could take 2 seconds in giving a response
because of internet latency or server load), and maybe they made another
calculations that makes them to take 5 seconds more for each invoice, and all
this is chained to one single transaction. This means that for 8.5 minutes
anybody else could confirm invoices, neither journal entries of the involved
journals.

Now, think there is 20 users that have to execute a similar process. The
problem turns exponential. If another user comes to make an operation with the
same jornal it will thrown a concurrency failure.

You can mitigate it if you segment each transaction and don't chain them. It
means, making commit for each invoice or process. It reduces the
probability that there is a concurrency error or a lock wait. However, it still
not solve it completely.

**Why to use Sequences with "Standard" Implementation?**

If you use the standard sequence of PosgreSQL, it doesn't lock because at the
moment the request is done, the next sequence number it is changed in an
isolated transaction, and it have not to wait the other transaction to end.
However, if the transaction produces a rollback, this sequence isn't reverted,
it means, it's lost. It may be not not serious because when you cancel or
remove records that number is lost too.

**What this module does?**

To eliminate completely that concurrency/slowness problem, this module changes
all the sequences (ir.sequence) implementation from "No Gap" to "Standard" with
the awareness that it will skip numbers. In the majority of database models
and many users projects there is no problem with that jump occurs.

Selección_798

Selección_799

/cc @moylop260


moylop260 (approved):
:+1:

ir_sequence_standard_default/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, SUPERUSER_ID
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    sequences = env['ir.sequence'].sudo().search([
+        ('implementation', '=', 'no_gap'), '|', ('active', '=', False),
+        ('active', '=', True)])
```
hugho-ad: use with_context(active_test=False) instead of that domain

ir_sequence_standard_default/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+
+import logging
+from odoo import api, SUPERUSER_ID
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    sequences = env['ir.sequence'].sudo().search([
```
hugho-ad: sudo statement not needed, post init scripts are executed as superuser

ir_sequence_standard_default/README.rst
```diff
@@ -0,0 +1,140 @@
...
+    :target: https://runbot.odoo-community.org/runbot/149/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Set the implementation to "Standard" in all your current Sequences
+(ir.sequence) and all new sequences are created as "Standard" by default
+instead of "No Gap" implementation.
+
+What's the problem Sequences with "No Gap" Implementation
```
luisg123v: What's the problem with "No Gap" Sequence Implementations

ir_sequence_standard_default/README.rst
```diff
@@ -0,0 +1,140 @@
...
+=========================================================
+
+"No Gap" is the default value of sequences in Odoo. However, this kind of
+sequences cause more locks and can turn a database slow.
+
+Taking as example an invoice, if you assign an invoice number to one record,
+but it sill not finish the process, this process must end in order to another
+invoice could assign a new number and there was no gaps between the invoice
+numbers. It seems to be good at first sight. But the problem start when there
```
luisg123v: starts

ir_sequence_standard_default/README.rst
```diff
@@ -0,0 +1,140 @@
...
+is and chained process.
```
luisg123v: a chained

ir_sequence_standard_default/models/ir_sequence.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrSequence(models.Model):
+
+    _inherit = "ir.sequence"
+
+    implementation = fields.Selection(default='standard')
```
luisg123v: Did you consider using an `ir.default` linked to a config setting?

For instance:
https://github.com/odoo/odoo/blob/d34d6867f2ff9abdf303f12b3fb9b7fc20e709ab/addons/sale/models/res_config_settings.py#L36

I'm not so sure if it would be convenient/usefull but, That way, the default value could be changed to its default behavior anytime in the future.

luisg123v (changes requested):

ir_sequence_standard_default/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+    for item in sequences:
+        seq_name = "ir_sequence_%03d" % (item.id)
+        env.cr.execute("DROP SEQUENCE IF EXISTS %s" % seq_name)
+        if item.use_date_range:
+            range_ids = {}
+            for line in item.date_range_ids:
+                range_ids[line.id] = line.number_next_actual
+                seq_name = "ir_sequence_%03d_%03d" % (item.id, line.id)
+                env.cr.execute("DROP SEQUENCE IF EXISTS %s" % seq_name)
+            item.write({'implementation': 'standard'})
```
deivislaya: Can remove this call to the `write` method and remove the `continue` statement

ir_sequence_standard_default/models/ir_sequence.py
ebirbe: Thank you, I took your idea and added it to the settings as a configurable value.

Selección_782

ir_sequence_standard_default/hooks.py
ebirbe: Yes I know it could look tricky, but both lines are required, and this is why:

When you have a sequence based on Date Ranges and No Gap implementation, then you change it to Standard, all the lines at the date ranges are reset in "Next Number" to "1", in other words, the real Next Number is lost.

Selección_783

Selección_784

To avoid that we first [check](https://github.com/OCA/server-tools/blob/167102097b4cf9417f4d36dc69e82cd813ebe1d7/ir_sequence_standard_default/hooks.py#L18) if it is a sequence with Date Range values, if so, we get the current value of Next Number for each line, after that, we [change](https://github.com/OCA/server-tools/blob/167102097b4cf9417f4d36dc69e82cd813ebe1d7/ir_sequence_standard_default/hooks.py#L24) the sequence to Standard (`write`), and then, we restore the right values for Next numbers on each line. On that point we call `continue` to avoid to calling the `write` outside the `if` condition, that is only executed when the Sequence is NOT based in Date Ranges.

In short words, when the sequence is Date Ranged we need to call `write` and then restore the Next Number of the sequence lines, when it is not Date Ranged, we can directly execute the `write`.

deivislaya: LGTM  :+1:

luisg123v (approved):

PR #1733, opened 2019-11-28
[REM] module_auto_update: pyc and pyo files

remytms:
I think that these two files should not stand here. It should be generated by python when
it needs.


pedrobaeza:
They are there for serving as sample file for being ignored in tests, so they are correct.

remytms:
Ok, understood. :)

PR #1738, opened 2019-12-05
[12.0] Add module base_m2m_custom_field

grindtildeath:
This module adds a new Many2many custom field with a `create_table` attribute.

Many2manyCustom field is useful when a direct access to the relational table
is needed, for example to be editable in a dedicated tree view.

Let's consider following models:

.. code-block:: python

    class MyModelA(models.Model):

        _name = 'my.model.a'

        my_model_b_ids = fields.Many2manyCustom(
            'my.model.b',
            'my_model_a_b_rel',
            'my_model_a_id',
            'my_model_b_id',
            create_table=False,
        )

    class MyModelB(models.Model):

        _name = 'my.model.b'

        my_model_a_ids = fields.Many2manyCustom(
            'my.model.a',
            'my_model_a_b_rel',
            'my_model_b_id',
            'my_model_a_id',
            create_table=False,
        )

    class MyModelABRel(models.Model):

        _name = 'my.model.a.b.rel'

        my_model_a_id = fields.Many2one(
            'my.model.a',
            required=True,
            index=True,
        )
        my_model_b_id = fields.Many2one(
            'my.model.b',
            required=True,
            index=True,
        )

By setting `create_table=False` on the Many2manyCustom field, and using the
relational table name, as `_name` for the relational model, we're able to
define a dedicated tree view for `my.model.a.b.rel`.

.. code-block:: xml

    <record id="my_model_a_b_rel_tree_view" model="ir.ui.view">
        <field name="name">my.model.a.b.rel.tree.view</field>
        <field name="model">my.model.a.b.rel</field>
            <tree editable="top">
                <field name="my_model_a_id" />
                <field name="my_model_b_id" />
            </tree>
        </field>
    </record>


base_m2m_custom_field/readme/USAGE.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+    class MyModelABRel(models.Model):
+
+        _name = 'my.model.a.b.rel'
+
+        my_model_a_id = fields.Many2one(
+            'my.model.a',
+            required=True,
+            index=True,
```
simahawk: add a comment: indexes are mandatory

base_m2m_custom_field/fields.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields
+
+
+class Many2manyCustom(fields.Many2many):
```
simahawk, line 6: what about `Many2manyNoTableCreate` (and change module name accordingly)?

simahawk (changes requested):

base_m2m_custom_field/__init__.py
```diff
@@ -0,0 +1 @@
+from . import fields
```
Tonow-c2c:
```suggestion
from . import fields

```

Tonow-c2c (approved):

base_m2m_custom_field/fields.py
grindtildeath: IMO this module could be improved with other slots than only `create_table`. That's also why I did set it to True as default...

simahawk (approved):

PR #1747, opened 2019-12-19
[12.0][FIX] base_custom_info

AdriaGForgeFlow:
Small fix on the creation of new templates:

image

cc ~ @ForgeFlow @pedrobaeza @Tardo

### DIFF
--- a/base_custom_info/models/custom_info_property.py
+++ b/base_custom_info/models/custom_info_property.py
@@ -83,9 +83,7 @@ class CustomInfoProperty(models.Model):
         self.mapped("template_id").check_access_rule(operation)
         return super().check_access_rule(operation)
 
-    @api.constrains("default_value", "field_type")
-    def _check_default_value(self):
-        """Ensure the default value is valid."""
+    def _check_default_value_one(self):
         if self.default_value:
             try:
                 self.env["custom.info.value"]._transform_value(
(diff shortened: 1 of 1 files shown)


base_custom_info/models/custom_info_property.py
```diff
@@ -83,20 +83,22 @@ def check_access_rule(self, operation):
         self.mapped("template_id").check_access_rule(operation)
         return super().check_access_rule(operation)
 
+    @api.multi
```
pedrobaeza: No need of putting this: it's the default.

pedrobaeza (changes requested):
For not changing too much the diff, you can create a method `_check_default_value_one` with the same content, and call it from the `_check_default_value` iterating over recordset.

base_custom_info/models/custom_info_property.py
pedrobaeza: This is still here

base_custom_info/models/custom_info_property.py
```diff
@@ -98,6 +96,13 @@ def _check_default_value(self):
                     _("Default value %s cannot be converted to type %s.") %
                     (self.default_value, selection[self.field_type]))
 
+    @api.constrains("default_value", "field_type")
+    def _check_default_value(self):
+        """Ensure the default value is valid."""
+        for rec in self:
+            rec._check_default_value_one()
+
```
pedrobaeza, line 104: Extra empty line

AdriaGForgeFlow: Should be ok now :+1:

pedrobaeza (approved):
Let's fast-track this. Thanks for the patch

PR #1749, opened 2019-12-19
[IMP][8.0] base_import_odoo: show inactive records also

thomaspaulb:
Records could be imported as inactive, but it is confusing not to show them in this view


thomaspaulb:
@hbrunn

gurneyalex (approved):

hbrunn (approved):

PR #1753, opened 2020-01-08
[IMP] auth_from_http_remote_user: check session in each request

sbidoul:
Previously the consistency between the session and the remote user
was checked in the /web route. Now it is checked at each request.
This is more robust in case, for instance, the user changes login in the
SSO then comes back to Odoo via deep links.


ThomasBinsfeld (changes requested):
`from . import ir_http` in `__init__.py` is missing.

sbidoul:
@ThomasBinsfeld thanks! fixed.

ThomasBinsfeld (approved):

PR #1756, opened 2020-01-10
[12.0] Migrate and rename attachment_base_synchronize

florian-dacosta:
Hello,

Migration of attachment_base_synchronize, renamed attachment base_attachment_queue to try to represent more what the module does.
I also have removed some stuff that was never used, it could be implemented in a sub module if needed. The goal being to have the simplest module possible.

As a reminder, the module does not really do something on its own. It implements a kind of queue on top of attachment. Allowing one to do some action on this attachment.
When this kind of attachment is created, it is at pending state. and it has a "File Type".
Then a cron will process the file and do different action depending on this file type. When processed, the state's attachment is changed for done or failed, depending on what has happened.

Examples of use cases : 
With module external_file_location (renamed in version 12 by attachment_synchronize) you can create an attachment with file_type export, and also specify a storage backend (sftp, aws, ftp, ...).
Then when the cron run, the attachment will be synchronized in the chosen storage backend, for example, sent by FTP.

Another module for instance would create an attachment with file type 'import bank statement'.
Then, once the file cron runs, the attachment would create a bank statement.
You can imagine multiple use cases, the goal is to be able to process a file.

@thomaspaulb @bealdav @sebastienbeau

PR #1757, opened 2020-01-10
[12] Migrate and rename external_file_location.

florian-dacosta:
Rename to attachment_synchronize.

Depends on https://github.com/OCA/server-tools/pull/1756
Depends on https://github.com/OCA/storage/pull/35

Still need to add tests at least

The module allow export files depending on the storage backend chosen (sftp, aws...)
It also allow to import files and put a file type on the attachment, so it is then processed by the system.
(allow for instance to import bank statement file from an external system and then to process the bank statement file in order to create it in Odoo)


GSLabIt:
Hi @florian-dacosta, I'd like to contribute to this. There is something I don't understand, when and where the file-type will be executed once imported?

There are some layout issue, that can be fixed easily fixed.

florian-dacosta:
Hi @GSLabIt 
Any contribution is welcome, of course!
I am not sure I understand your question.
Once a file is imported, an object called ir.attachment.metadata is created in Odoo,  it inherits the ir.attachment. Because of the file type set on the storage backend task, the file type is also set on the attachment.
Then, there is a cron, processing all pending attachment. (cron comes from the module base_attachment_queue, which is a dependency of this one).

But anyway, in order to import a file from an external storage and assign a file type to it, you need to create another module, since this one only implement the "export" file type, used to send a file toward an external storage.

Here is an example of sub modules, that implement a file type. (not sure it works as it is because net quite finished yet) : https://github.com/akretion/server-tools/tree/12.0-mig-account_statement_move_attachment_metadata/account_statement_move_attachment_metadata
This module implement 2 file types : Account Statement Import and Account Move Import)
So, if you use this module together with attachment_synchronize, you can create a storage backend task that will import file from a external storage and assign type "Account Statement Import".
Then, each time a file is imported from external storage, the attachment created in Odoo will be processed by the  cron coming from base_attachment_queue module, and will create a bank statement.

If I resume, using this module only, you will be able to import files from an external storage, but these file won't be processed in Odoo, because they won't have any file type. You need other modules for this.

I am not sure I am clear, do not hesitates to tell me if I am not.

GSLabIt:
Hi @florian-dacosta, u got the point 😬. I got how it works now, thx. 

> Then, there is a cron, processing all pending attachment. (cron comes from the module base_attachment_queue, which is a dependency of this one).

What u think about https://github.com/OCA/queue/tree/13.0/queue_job_cron to queue the attachment jobs?

What's the best way to contribute? forking ur repo?

florian-dacosta:
@GSLabIt 
I don't see how the module queue_job_cron can help here.

But, I thought a lot about replacing the module `base_attachment_queue` by the module queue_job.
The main advantage beeing, we would not need to re-implement a model (ir.attachment.metadata) in order to process our attachments. We could instead directly create a job, with the file in attachment and let the job (fro queue_job) process everything.
So, we could drop base_attachment_queue totally with this solution.

But, it has also disadvantages. the jobs (from queue_job) are not user friendly.
The advantage of creating a new model (ir.attachment.metadata) is that we can easily add the new fields we need (storage_backend_id, file_type, etc...) and it is a classic model, so we can improve the view and make it user friendly.
Also, with queue_job, the attachments would be deleted together with the job, after one month all, and we'd lose the historic.

Anyway, I made the choice to no base the module on queue_job for now, as it would add some work to make it user friendly, but I am not totally against either.

But the module queue_job_cron is no use in our  case, it allow to run the cron as a job, but I don't see why it would be needed for attachment_synchronize.

If you want to contribute, you can fork and make Pull request on github.com/akretion/server-tools.
Then I will review it and merge it.

GSLabIt:
@florian-dacosta I see, and I agree on the choice u made. 

I was mainly thinking on processing big files, that would lead in timeouts, but it would be implemented in submodules.

I'll fork and will make pr

GSLabIt:
> Here is an example of sub modules, that implement a file type. (not sure it works as it is because net quite finished yet) : https://github.com/akretion/server-tools/tree/12.0-mig-account_statement_move_attachment_metadata/account_statement_move_attachment_metadata

Quick question:

https://github.com/akretion/server-tools/blob/d120c1dcf5ea279b36c7e6ed3a7c4748fe812365/account_statement_move_attachment_metadata/models/ir_attachment.py#L14-L16

```
    journal_id = fields.Many2one(
        'account.journal'
        )
```
Where `journal_id`is set? Shouldn't it come from `task`? Cannot figure it out, any hint?

florian-dacosta:
> Where `journal_id`is set? Shouldn't it come from `task`?

Yes it should. That is strange, I think the task files are missing.
But yes, the `journal_id` field should be added on `storage.backend.task` and in the view also.
And then, just on override on `_prepare_attachment_vals` to add the journal_id on the attachment when the file is imported.
I am not currently working on the subject and the branch are not totally complete.

sebastienbeau:
@GSLabIt I have done some change in the base module (renaming module and some clean) https://github.com/OCA/server-tools/pull/1757
I am going tp merge the dependency and then rebase this work on it

sebastienbeau:
Just need to merge https://github.com/OCA/server-tools/pull/1756 and test should be green

attachment_synchronize/models/task.py
```diff
@@ -0,0 +1,210 @@
...
+        total_import = 0
+        for file_name in filenames:
+            with api.Environment.manage():
+                with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(
+                        new_cr, self.env.uid, self.env.context
+                    )
+                    try:
+                        full_absolute_path = os.path.join(filepath, file_name)
+                        datas = backend._get_b64_data(full_absolute_path)
```
rvalyi:
```suggestion
                        data = backend._get_b64_data(full_absolute_path)
```
data is already a plural form https://www.learnenglish.de/mistakes/data.html
You should propagate the change...

attachment_synchronize/views/task_view.xml
```diff
@@ -0,0 +1,56 @@
...
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+    <record id="view_task_tree" model="ir.ui.view">
+        <field name="model">attachment.synchronize.task</field>
+        <field name="arch" type="xml">
+            <tree string="Tasks" colors="grey: not(enabled)">
```
bealdav: colors don't works here. I don't know why ?

attachment_synchronize/views/storage_backend_view.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="view_storage_backend_form" model="ir.ui.view">
```
sebastienbeau: Can we remove this and add a dedicated view for the task. So we can change the backend of a task easily?

sebastienbeau: @clementmbr

attachment_synchronize/views/task_view.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sebastienbeau: @clementmbr add a view search and a menu

bealdav (approved):

bealdav:
Finally get success: ci/rcunbot is a false positive

```

2020-08-27 08:36:17,165 171 INFO openerp_test odoo.service.server: Initiating shutdown
2020-08-27 08:36:17,165 171 INFO openerp_test odoo.service.server: Hit CTRL-C again or send a second signal to force the shutdown.
----------
Module test summary
[1;32mSuccess[0;m attachment_queue,attachment_synchronize,attachment_unindex_content,auditlog,auto_backup,autovacuum_message_attachment,base_conditional_image,base_cron_exclusion,base_custom_info,base_exception,base_fontawesome,base_jsonify,base_kanban_stage,base_kanban_stage_state,base_m2m_custom_field,base_remote,base_search_fuzzy,base_technical_user,base_view_inheritance_extension,company_country,configuration_helper,datetime_formatter,dbfilter_from_header,excel_import_export,excel_import_export_demo,fetchmail_incoming_log,fetchmail_notify_error_to_sender,html_image_url_extractor,html_text,ir_sequence_standard_default,module_auto_update,onchange_helper,scheduler_error_mailer,sentry,slow_statement_logger,sql_export,sql_export_mail,sql_request_abstract

+=======================================
|  Tests summary:
|---------------------------------------
| test_server.py              [1;32mSuccess[0;m
+=======================================
 * Stopping PostgreSQL 9.6 database server       [80G 
[74G[ OK ]

```

sebastienbeau (approved):
LGTM code review

sebastienbeau:
@gurneyalex any idea why runbot is red? even if it say it's ok?

attachment_synchronize/tests/test_import.py
```diff
@@ -0,0 +1,83 @@
...
+    @property
+    def input_files(self):
+        return self.backend._list(self.directory_input)
+
+    def _check_attachment_created(self, count=1):
+        attachment = self.env["attachment.queue"].search([("name", "=", "bar.txt")])
+        self.assertEqual(len(attachment), count)
+
+    def test_import_with_rename(self):
+        self.task.write({"after_import": "rename", "new_name": "foo.txt"})
```
mourad-ehm, line 22: Could you test  rename with jinja pattern ?

bealdav: kesako ?

mourad-ehm: I think is not necessary because def _template_render method is a basic code.

mourad-ehm (approved):
Code Review

rvalyi:
> @gurneyalex any idea why runbot is red? even if it say it's ok?

In th Runbot log there is no error indeed. All we can find are a couple of warnings:
```
2020-08-27 08:36:09,543 171 INFO openerp_test odoo.modules.loading: loading attachment_synchronize/demo/attachment_synchronize_task_demo.xml
2020-08-27 08:36:09,564 171 WARNING openerp_test odoo.models: attachment.synchronize.task.create() with unknown fields: emails
2020-08-27 08:36:09,885 171 WARNING openerp_test odoo.models: attachment.synchronize.task.create() with unknown fields: emails
```
Do you think it can come from here?
In any case it seems these 2 emails fields could be removed from the demo data, right?
https://github.com/OCA/server-tools/blob/83f2c4d434b47bbd4d550a6f300d6c0885f2c836/attachment_synchronize/demo/attachment_synchronize_task_demo.xml#L9

But the thing is there are also warnings with the server_env module:server_environment
```
2020-08-27 08:34:21,978 171 WARNING openerp_test odoo.addons.server_environment.serv_config: `running_env` not found. Using default = `test`.
2020-08-27 08:34:21,978 171 WARNING openerp_test odoo.addons.server_environment.serv_config: We strongly recommend against using the rc file but instead use an explicit config file or env variable.
```

It's worth removing the emails field, but will it be green later with this server_environment warning? I don't know and I couldn't find any older server-tools 12.0 Runbot log to know it.

cc @sebastienbeau

bealdav:
@rvalyi running key in travis shouldn't be the default one

rvalyi (approved):

PR #1762, opened 2020-01-27
13.0 mig mail server relay disallowed

JordiBForgeFlow:


mail_server_relay_disallowed/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+                'Project Expert Team, '
+                'Odoo Community Association (OCA)',
+    'contributors': [
+        'Jordi Ballester <jordi.ballester@eficent.com>',
+    ],
+    'website': 'http://project.expert',
+    'license': 'AGPL-3',
+    'category': 'Social Network',
+    'depends': ['base'],
+    'data': [],
```
MeetKD: Can we remove this empty key ?

mail_server_relay_disallowed/models/ir_mail_server.py
```diff
@@ -0,0 +1,109 @@
...
+from odoo.addons.base.models.ir_mail_server import MailDeliveryException
+from odoo.osv import osv
+from odoo.tools.translate import _
+import odoo.tools as tools
+
+_logger = logging.getLogger(__name__)
+_test_logger = logging.getLogger('odoo.tests')
+
+
+class IrMailServer(osv.Model):
```
MeetKD: models.Model ?

mail_server_relay_disallowed/models/ir_mail_server.py
```diff
@@ -0,0 +1,109 @@
...
+            msg = _(
+                "Mail delivery failed via SMTP server '%s'.\n%s: %s"
+            ) % (
+                tools.ustr(smtp_server),
+                e.__class__.__name__,
+                tools.ustr(e)
+            )
+            _logger.error(msg)
+            raise MailDeliveryException(_("Mail Delivery Failed"), msg)
+        return message_id
```
MeetKD, line 132: I think here it should adapt changes done in original v13 code.

MeetKD (changes requested):

mail_server_relay_disallowed/models/ir_mail_server.py
MiquelRForgeFlow: done

MiquelRForgeFlow (approved):

AdriaGForgeFlow (approved):
Code review :+1:

mail_server_relay_disallowed/models/ir_mail_server.py
```diff
@@ -0,0 +1,132 @@
...
+        # provided by caller.  Caller may be using Variable Envelope Return
+        # Path (VERP) to detect no-longer valid email addresses.
+        smtp_from = (
+            message["Return-Path"]
+            or self._get_default_bounce_address()
+            or message["From"]
+        )
+        assert (
+            smtp_from
+        ), "The Return-Path or From header is required for any outbound email"
```
bealdav, line 54: Thanks for your module.
It seems than assert expression shouldn't not be in production code like here
https://stackoverflow.com/a/5143044 because could be skipped according to the execution mode of your app instance. Then assert should be reserve for unit tests where there is no production mode in execution. 
But yes Odoo SA itself use this statement in its code not only for tests. As usual, OCA promotes the best practices before Odoo SA
https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement
There is a lot of documentation on this subject in python ecosystem
Not blocking for me, just for information

MiquelRForgeFlow: yeah, but here this part of code is just copied from odoo (it's an override) and we want to change the minor possible the code, in order to be able to check easy what change has been done comparing to original code.

PR #1511, opened 2019-02-14
[12.0] WIP mig letsencrypt

eLBati:

PR #1694, opened 2019-10-02
12.0 mig letsencrypt

daramousk:

PR #1763, opened 2020-02-03
[12.0] mig letsencrypt

NL66278:
This will take over from https://github.com/OCA/server-tools/pull/1511 which will not be updated.


NL66278:
Now also includes all changes in https://github.com/OCA/server-tools/pull/1694

NL66278:
Runbot is red, despite running succesfully. Only some warnings on models - not part of this PR - not having a description. Is this really the reason for failure?

database_cleanup/tests/test_database_cleanup.py
```diff
@@ -1,6 +1,7 @@
-# Copyright 2016 Therp BV <http://therp.nl>
```
emagdalenaC2i, line 1: Should this file be in other PR?

hbrunn:
runbot always turns red for warnings, you're just not used to it because I added [the possibility to ignore warnings](https://github.com/OCA/runbot-addons/blob/11.0/runbot_gitlab/models/runbot_repo.py#L53) for `runbot_gitlab` :-)

database_cleanup/tests/test_database_cleanup.py
hbrunn: there are a bunch of unrelated things that make this branch (and all other 13.0 PRs) red, so @NL66278 just fixes those here. As long as this happens in different commits, that's totally fine. When this will be merged, the whole branch will be green, and we can rebase the other PRs to also have CI succeed there and merge

NL66278: I need this to get travis green.

hbrunn (approved):

hbrunn:
I migrated #1840 to v12, so I suggest to close this one. Let's see to get #1840 and #1813 merged to that we have some consistency across versions

PR #1764, opened 2020-02-04
[13.0][ADD] base_changeset

astirpe:
This module extends the functionality of records. When a user tries to modify a record, instead of direct modifications, it creates changesets that must be validated by a manager.
Rules allow to configure which field must be validated.

This module is an evolution of module [partner_changeset](https://github.com/OCA/partner-contact/tree/10.0/partner_changeset), created by @camptocamp.
The original module `partner_changeset` works only on `res.partner` records. This one instead works with records of every model.

In this module it's also possible to see which field of the record has a change that is still pending:

image


pedrobaeza:
Have you thought about converging with `base_tier_validation` for the validation layer?

guewen:
@astirpe excellent improvement over my initial work (cool to see it's been useful), the UI is much better now :D

> Have you thought about converging with `base_tier_validation` for the validation layer?

Seems like a large work (adapting the views...) and I doubt that anyone needs a validation workflow / steps to validate a change on a single / selection of fields. But if someone needs it at some points, seems like a good idea ;)

guewen (approved):

astirpe:
@pedrobaeza I didn't think about it. I prefer to keep this module as it is for the time being.

Thank you @guewen!

astirpe:
I'm starting to have important feedback from our first customer running this module. :smiley: 

@guewen I want to share with you the main of the points I got from the feedback.
It seems that the word "Changeset", that is widely used within this module, is not well accepted by the end users, since the meaning is too technical. For the end user it sounds like meaningless.

I'm not going to change the technical name of this module but anyway I have to update the name of menuitems, actions and fields in order to make them more meaningful and user friendly.
I'm wondering if you agree and maybe you could suggest a good replacement of the word. Thanks!

guewen:
I have no problem changing the wording! No idea for now but if something pops up I'll comment ;)

andreschenkels (approved):
LGTM I also made a port to use this in all models. But with the UI change this looks pretty good. Thnx man.

PR #1688, opened 2019-10-01
[12.0][FIX] auditlog: log computed fields + create multi

sebalix:
Forward port #1686 (11.0) to 12.0.
Odoo stores values of computed fields at the end of the transaction only, as such performing a 'read()' to make a data snapshot on the record created in the current transaction doesn't return the expected result regarding these fields.
Also as a side-effect 'read()' alters the environment cache and break the values on the record inducing issues in the whole user transaction/workflow.

This fix replaces the use of 'read()' to do the data snapshot directly from the cache of the record (computed values are already there).

The second commit is a second fix (it was simpler to include it to avoid PR battle), it fixes the log creation when the `api.model_create_multi` is used. Supersedes #1627 (to fix #1619 ).
ping @mtelahun

PR #1766, opened 2020-02-10
[RFR] [FIX] 12.0 deepcopy on api.Environment class fails

PieterPaulussen:
An empty environment object will trigger an error during deepcopy.
From what I could gather, an empty Environment object is a possible value to create new records from.
Python's deepcopy fails to create a new object from this value.

Not saying my solution is the right one, but @sebalix I was interested in your feedback on this since you created the original fix (https://github.com/OCA/server-tools/pull/1688) 

The actual error is that for each key-value pair in the dictionary sent to deepcopy, python attempts to instantiate object. In the case of an api.Environment class object, a new `api.Environment` object  is created which in turn calls to `__new__()`. Unfortunately, that method requires `cr`, `uid`, and `context` as parameters and ultimately fails to create the new object.

### DIFF
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -92,12 +92,14 @@ class AuditlogCommon(object):
         groups_vals = [
             {'name': 'testgroup1'},
             {'name': 'testgroup3'},
-            {'name': 'testgroup2'},
+            {'name': 'testgroup2', 'users': self.env['res.users']},
+            {'name': 'testgroup4', 'users': self.env.user},
         ]
         groups = self.env['res.groups'].create(groups_vals)
         # Ensure that the recordset returns is in the same order
         # than list of vals
-        expected_names = ['testgroup1', 'testgroup3', 'testgroup2']
+        expected_names = [
+            'testgroup1', 'testgroup3', 'testgroup2', 'testgroup4']
         self.assertEqual(groups.mapped('name'), expected_names)
 
         logs = auditlog_log.search([
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -14,7 +14,9 @@ FIELDS_BLACKLIST = [
 EMPTY_DICT = {}
 
 
-def model__deepcopy__(self, memo={}):
+def model__deepcopy__(self, memo=None):
+    if memo is None:
+        memo = {}
     return self.browse(self.ids)
 
 
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -14,6 +14,13 @@ FIELDS_BLACKLIST = [
 EMPTY_DICT = {}
 
 
+def model__deepcopy__(self, memo={}):
+    return self.browse(self.ids)
+
+
+models.Model.__deepcopy__ = model__deepcopy__
+
+
 class DictDiffer(object):
     """Calculate the difference between two dictionaries as:
     (1) items added


PieterPaulussen:
@sebalix any comments on this issue?

StefanRijnhart:
Interesting problem. While your implementation makes sense entirely in this context, it seems to contradict the idea of a deepcopy. Then again, I cannot imagine a more appropriate implementation of deepcopy of Odoo models that will not completely derail. That may be due to my lack of imagination, so it would be interesting if you could propose something like this into odoo/master and see what they say over there. I found it helpful to have a look at the actual traceback, so I would recomment including it in your PR description over there

```
>>> deepcopy(self.env.user)
  File "<console>", line 1, in <module>
  File "/usr/lib/python3.6/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.6/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.6/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.6/copy.py", line 274, in _reconstruct
    y = func(*args)
  File "/usr/lib/python3.6/copyreg.py", line 88, in __newobj__
    return cls.__new__(cls, *args)
TypeError: __new__() missing 3 required positional arguments: 'cr', 'uid', and 'context'
```

PieterPaulussen:
PR on master has been created here: https://github.com/odoo/odoo/pull/64610

auditlog/tests/test_auditlog.py
```diff
@@ -92,12 +92,14 @@ def test_LogCreation4(self):
         groups_vals = [
             {'name': 'testgroup1'},
             {'name': 'testgroup3'},
-            {'name': 'testgroup2'},
+            {'name': 'testgroup2', 'users': self.env['res.users']},
+            {'name': 'testgroup4', 'users': self.env.user},
```
StefanRijnhart, line 96: It looks like passing recordsets to `create` is not actually supported:
```
>>> group = self.env['res.groups'].create({'name': 'bla', 'users': self.env.user})
2021-03-17 09:11:54,572 793620 WARNING 14nl odoo.models: Comparing apples and oranges: res.users(1,) == 4 (/home/odoo/buildout/14.0/parts/odoo/addons/mail/models/res_users.py:173
```
```
>>> partner = self.env.user.partner_id
>>> user = self.env['res.users'].create({'login': 'test', partner_id: partner})
  File "<console>", line 1, in <module>
NameError: name 'partner_id' is not defined
>>> user = self.env['res.users'].create({'login': 'test', 'partner_id': partner})
2021-03-17 09:10:05,222 793620 ERROR 14nl odoo.sql_db: bad query: INSERT INTO "res_users" ("id", "create_uid", "create_date", "write_uid", "write_date", "active", "company_id", "login", "notification_type", "partner_id", "signature") VALUES (nextval(%s), %s, %s, %s, %s, %s, %s, %s, %s, %s, %s) RETURNING id
ERROR: can't adapt type 'res.partner'
```
So that would make this bug invalid.

sebalix: Sorry @PieterPaulussen for being late.
Indeed, we must not use recordset in a dict parameter of `write` or `create`.

PieterPaulussen: @sebalix Thanks for the feedback. @StefanRijnhart I believe this renders this PR invalid? It must of been a bug in another module.

PieterPaulussen:
Using a recordset in a write / create dictionary is not supported by the ORM.

auditlog/tests/test_auditlog.py
StefanRijnhart: Yes, exactly!

GabbasovDinar:
@PieterPaulussen Hello,

if we want to enable audit log for quants for example we will have the same error:

```TypeError: __new__() missing 3 required positional arguments: 'cr', 'uid', and 'context'```

(in case if will don't have lot and etc we try to write empty recordset)
https://github.com/odoo/odoo/blob/b22a23435a0cc42b7084ded21dbd4ef3a4a2113e/addons/stock/models/stock_quant.py#L444

yibudak:
> @PieterPaulussen Hello,
> 
> if we want to enable audit log for quants for example we will have the same error:
> 
> `TypeError: __new__() missing 3 required positional arguments: 'cr', 'uid', and 'context'`
> 
> (in case if will don't have lot and etc we try to write empty recordset) https://github.com/odoo/odoo/blob/b22a23435a0cc42b7084ded21dbd4ef3a4a2113e/addons/stock/models/stock_quant.py#L444

here is a simple solution for stock.quant

https://github.com/yibudak/OCBAltinkaya/commit/504cecb2a520da1c62648f8c85f871bb380b079f

StefanRijnhart:
Now I'm worried. I have auditlogging on stock.quant in Odoo 12 without such a fix :scream: 

Thanks!

GabbasovDinar:
@yibudak thanks, but it was an example, only. There is no guarantee that such problems will not occur for other models, and it seems that we need a solution within the auditlog module

PR #1771, opened 2020-02-21
[IMP] Precreate display_name columns

StefanRijnhart:
When migrating a database with (only) hundred thousands of records, this script will take 5 seconds as opposed to 4.5 minutes when letting the ORM compute these values.


legalsylvain (approved):
Thanks. Code review / no test.

PR #1777, opened 2020-03-02
[13.0] MIG base_view_inheritance_extension

alexis-via:


ajaniszewska-dev:
Hello @alexis-via 
Tried to use your module by doing the following:
```
<?xml version="1.0" encoding="utf-8"?>
<odoo>
	<template id="specific_followup_report" inherit_id="account_followup.followup_filter_info_template">
		<xpath expr="//t[@t-set='contact_widget_fields']" position="attributes">
			<attribute name="t-value" operation="list_remove">'phone'</attribute>
		</xpath>
	</template>
</odoo>
```
But I get the folllowing error:
```
  File "/odoo/src/odoo/addons/base/models/ir_ui_view.py", line 713, in read_combined
    arch_tree = self.apply_inheritance_specs(arch_tree, view_arch, parent_view['id'])
  File "/odoo/external-src/server-tools/base_view_inheritance_extension/models/ir_ui_view.py", line 46, in apply_inheritance_specs
    source = handled_by(source, specs, inherit_id, pre_locate=pre_locate)
odoo.tools.convert.ParseError: "apply_inheritance_specs() got an unexpected keyword argument 'pre_locate'" while parsing /odoo/local-src/specific_report/views/report_followup.xml:2, near
<odoo>
	</odoo>
```
Do you know what's wrong?

cc @grindtildeath

alexis-via:
@ajaniszewska-dev As explained in the new README, list_remove is not provided by this module any more, because the feature is now native with a different syntax. In the README, there are pointers to the new syntax.

pedrobaeza:
@alexis-via several comments:

- For avoiding misunderstandings, you should regenerate the main README.rst, as it's still putting the 11.0 references.
- The URLs you link about native features are relative links, not permalinks, so they may change in the future (if there's code added/removed to that files). Use permalinks instead.
- AFAIK the list_remove/list_add features of this module aren't the same as the add/separator native features. `list_add="d"` acts on an attribute `x="['a', 'b', c']"`, putting `['a', 'b', c', 'd']`, while `add` will perform `['a', 'b', c'], 'd'`, but maybe I'm wrong.

alexis-via:
@pedrobaeza About your last point, you are certainly right. Let's kill this shitty PR.

PR #1779, opened 2020-03-03
[MIG][13.0] base_view_inheritance_extension: Migration to v13.0

sergio-teruel:
cc @Tecnativa TT22701


base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,131 @@
...
+            for child in spec:
+                node = etree.Element(spec.tag, **spec.attrib)
+                node.insert(0, child)
+                yield node, self._get_inheritance_handler_attributes(child)
+            return
+        yield spec, self._get_inheritance_handler(spec)
+
+    @api.model
+    def _get_inheritance_handler(self, node):
+        handler = super(IrUiView, self).apply_inheritance_specs
```
Tardo, line 67:
```suggestion
        handler = super().apply_inheritance_specs
```

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,131 @@
...
+        return handler
+
+    @api.model
+    def _get_inheritance_handler_attributes(self, node):
+        handler = super(IrUiView, self).apply_inheritance_specs
+        if hasattr(self, "inheritance_handler_attributes_%s" % node.get("operation")):
+            handler = getattr(
+                self, "inheritance_handler_attributes_%s" % node.get("operation")
+            )
+        return handler
```
Tardo, line 79:
```suggestion
        return getattr(
                self, "inheritance_handler_attributes_%s" % node.get("operation"), handler
            )
```

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,131 @@
...
+                node.get(attribute_node.get("name")) or "{}",
+                UnquoteEvalObjectContext(),
+                nocopy=True,
+            )
+            python_dict[attribute_node.get("key")] = UnquoteObject(attribute_node.text)
+            node.attrib[attribute_node.get("name")] = str(python_dict)
+        return source
+
+    @api.model
+    def inheritance_handler_attributes_list_add(self, source, specs, inherit_id):
```
Tardo, line 121: This is not the same that ```<attribute name="$attribute" add="$new_value" separator=" " />```?

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,131 @@
...
+        node = self.locate_node(source, specs)
+        for attribute_node in specs:
+            attribute_name = attribute_node.get("name")
+            old_value = node.get(attribute_name) or ""
+            new_value = old_value + "," + attribute_node.text
+            node.attrib[attribute_name] = new_value
+        return source
+
+    @api.model
+    def inheritance_handler_attributes_list_remove(self, source, specs, inherit_id):
```
Tardo, line 137: This is not the same that ```<attribute name="$attribute" remove="$new_value" separator=" " />```?

Tardo (changes requested):
Looks good! Code review...

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,131 @@
...
+        <$node position="attributes">
+            <attribute name="$attribute" operation="python_dict" key="$key">
+                $keyvalue
+            </attribute>
+        </$node>"""
+        node = self.locate_node(source, specs)
+        for attribute_node in specs:
+            python_dict = safe_eval(
+                node.get(attribute_node.get("name")) or "{}",
+                UnquoteEvalObjectContext(),
```
astirpe: I noticed a problem while porting module `account_cost_center` to V13.

Consider the following code:

```xml
    <record id="view_move_form" model="ir.ui.view">
        <field name="model">account.move</field>
        <field name="inherit_id" ref="account.view_move_form" />
            <field name="ref" position="after">
                <field name="cost_center_id" />
            </field>
            <field name="invoice_line_ids" position="attributes">
                <attribute name="context" operation="python_dict" key="cost_center_id">cost_center_id</attribute>
            </field>
        </field>
    </record>
```

The above applies to line [account/views/account_move_views.xml#L665](https://github.com/odoo/odoo/blob/13.0/addons/account/views/account_move_views.xml#L665) so that the expected result should be:

`{ ... bla bla ... , 'default_currency_id': currency_id != company_currency_id and currency_id or False, 'cost_center_id': cost_center_id}`

Instead this is what I get:

`{ ... bla bla ... , 'default_currency_id': currency_id, 'cost_center_id': cost_center_id}`


I think it depends on the `UnquoteObject` implementation.

astirpe (changes requested):

alexis-via (approved):
Tested with account_invoice_transmit_method, works well

pedrobaeza:
Oh wait, there's still comments unattended. @sergio-teruel can you check them?

astirpe:
My comment is also not attended

TDu:
Thanks for the work.
Works well for my case also testing with `account_invoice_transmit_method` would be good to get the last fixes...@sergio-teruel

sergio-teruel:
@astirpe CAn you take a look ... I create a test for this case..
Now I do not eval context....

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -0,0 +1,147 @@
...
+            </attribute>
+        </$node>"""
+        node = self.locate_node(source, specs)
+        for attribute_node in specs:
+            str_dict = node.get(attribute_node.get("name")) or "{}"
+            variables = self._list_variables(str_dict)
+            my_dict = safe_load(str_dict)
+            my_dict[attribute_node.get("key")] = attribute_node.text
+            for k, v in my_dict.items():
+                my_dict[k] = UnquoteObject(v) if v in variables else v
```
astirpe, line 116:
```suggestion
                my_dict[k] = UnquoteObject(v) if v in variables else v
            my_dict[attribute_node.get("key")] = UnquoteObject(attribute_node.text)
```

The `attribute_node.text` needs to be unquoted in order to make `account_cost_center` module work.

astirpe (changes requested):
@sergio-teruel thank you! Please have a look to my comment.

base_view_inheritance_extension/models/ir_ui_view.py
sergio-teruel: Why??... How do you use the inheritante? Can you give me it to see it?

sergio-teruel: Sorry... I see the problem

sergio-teruel:
@astirpe Changes done!!

astirpe (approved):
Thanks! It works for my needs :+1:

kittiu:
Hi there. Obviously this module can get merged without problem.
But the PR I made, https://github.com/OCA/server-tools/pull/1887 having error here,
https://travis-ci.com/github/OCA/server-tools/jobs/376854743#L1481
The problem seem to come from this module "yaml". I tried adding the lib, doing PR anew but can't solve it.
Any hint?

pedrobaeza:
This was merged because everything was green. If now it's not, then something has changed. If a Python library is missing, add it to requirements.txt file.

kittiu:
> This was merged because everything was green. If now it's not, then something has changed. If a Python library is missing, add it to requirements.txt file.

Already added, pyyaml, but still not working.
https://travis-ci.com/github/OCA/server-tools/jobs/376943711#L1411

But it is ok, will keep finding it. Please let me know if you have any other hint :+1: 
Thank you!

PR #1782, opened 2020-03-05
[13.0][MIG] html_text

newtratip:
This module provides some technical features that allow to extract text from any chunk of HTML, without HTML tags or attributes. You can chose either:
- To truncate the result by amount of words or characters.
- To append an ellipsis (or any character(s)) at the end of the result.

It can be used to easily generate excerpts.


newtratip:
@newtratip Test

PR #1785, opened 2020-03-09
v8 FIX sql_request_abstract: do not merge params when no params

bealdav:
Before PR

with this query 
`WHERE s.create_date > '2019-06-30' and t.name NOT LIKE '%notice%'`
because %n is interpreted but there is no params here

After PR
query don't fail and the code is like in v12

https://github.com/OCA/server-tools/blob/12.0/sql_request_abstract/models/sql_request_mixin.py#L151-L154


bealdav:
cc @florian-dacosta @legalsylvain 

fail is out of topic of this PR

legalsylvain:
Thanks.

PR #1786, opened 2020-03-14
[FIX] company_country: Don't fail if account is already installed

luisg123v:
When `company_country` is installed, it performs the following:
- Look for the `COUNTRY` environment variable
- If not found, look for a `l10n_*` module that is about to be installed
- If not found, raise an exception

However, If the account module is installed, it shouldn't fail even if
the environment variable is not defined, because changing the
company's country will have no effect anyway, as the account hook was
already run and an `l10n` module was already been installed.

This commit fixes the above by skipping the process if the `account`
module is installed, and print an informative message to the log.

### DIFF
--- a/company_country/models/res_config.py
+++ b/company_country/models/res_config.py
@@ -1,9 +1,12 @@
 # Copyright 2016 Vauxoo
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
 import os
 
 from odoo import _, api, models
 from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
 
 
 class CompanyCountryConfigSettings(models.TransientModel):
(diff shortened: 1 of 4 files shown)


company_country/models/res_config.py
```diff
@@ -1,21 +1,34 @@
...
 
 class CompanyCountryConfigSettings(models.TransientModel):
     _name = 'company.country.config.settings'
     _description = 'Company Country Configuration Settings'
 
     @api.model
     def load_company_country(self, country_code=None):
+        account_installed = self.env['ir.module.module'].search([
+            ('name', '=', 'account'),
+            ('state', 'in', ('installed', 'to upgrade')),
```
moylop260:
```suggestion
            ('state', '=', 'installed'),
```

luisg123v: @moylop260 
Are you sure?

Note that `to upgrade` indicates that the module is already installed, but it's about to be updated (which would be the case if updating all modules AFAIK).

luisg123v: Done @moylop260

moylop260:
Could you check the unittest, please?

luisg123v:
Done @moylop260, tests were fixed.

FYI it's still red, but it seems unrelated, I pushed my branch without any changes and got the same errors.

moylop260:
`fetchmail_notify_error_to_sender` is not related

moylop260:
I just bumped version manually and merge it
Since that the red are unrelated

moylop260:
Could you backport the module to 11.0, please?

PR #1787, opened 2020-03-15
[12.0][FIX] excel_import_export, problem on import with datetime field

kittiu:
Fix for case import if the import file has datetime field

### DIFF
--- a/excel_import_export/models/common.py
+++ b/excel_import_export/models/common.py
@@ -268,13 +268,12 @@ def _get_cell_value(cell, field_type=False):
     datemode = 0  # From book.datemode, but we fix it for simplicity
     if field_type in ['date', 'datetime']:
         ctype = xlrd.sheet.ctype_text.get(cell.ctype, 'unknown type')
-        if ctype == 'number':
+        if ctype in ('xldate', 'number'):
+            is_datetime = cell.value % 1 != 0.0
             time_tuple = xlrd.xldate_as_tuple(cell.value, datemode)
             date = dt(*time_tuple)
-            if field_type == 'date':
-                value = date.strftime("%Y-%m-%d")
-            elif field_type == 'datetime':
-                value = date.strftime("%Y-%m-%d %H:%M:%S")
+            value = (date.strftime("%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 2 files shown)


kittiu:
Note: Travis seem to fail because other modules (OCB).

JordiBForgeFlow (approved):
:+1: Works as expected.

JordiBForgeFlow:
@kittiu can you rebase? We have tried to greenify the repo here https://github.com/OCA/server-tools/pull/1789

kittiu:
> @kittiu can you rebase? We have tried to greenify the repo here #1789

Done, but seem still one issue with OCB.

MiquelRForgeFlow:
@kittiu @JordiBForgeFlow now it's green :green_apple:

excel_import_export/models/common.py
```diff
@@ -268,13 +268,12 @@ def _get_cell_value(cell, field_type=False):
     datemode = 0  # From book.datemode, but we fix it for simplicity
     if field_type in ['date', 'datetime']:
         ctype = xlrd.sheet.ctype_text.get(cell.ctype, 'unknown type')
-        if ctype == 'number':
+        if ctype in ('xldate', 'number'):
+            is_datetime = cell.value % 1 != 0.0
```
MiquelRForgeFlow, line 272: I have a question. This kind of check is safe? Cannot fail? Isn't it better to check `field_type == 'datetime'`? Just asking.

MiquelRForgeFlow (approved):
LGTM

excel_import_export/models/common.py
kittiu: I am not too certain. But as I saw in Libre office, for the field what is Date, i.e., 2010-10-10, when chagned cell type to text, number or general, it is changed to 40461. So, I just keep that on the safe side (some how).
WDYT?

kittiu: Also, comparing to 'number' used to work on my older versions (back in v8).

MiquelRForgeFlow: ok

kittiu:
@JordiBForgeFlow can you merge this? I will also fix in v13 too (now it is PR).

PR #1791, opened 2020-03-18
[IMP] company_country: Don't create DB table for model

luisg123v:
That model is intended only to set config and has no fields, so there is
no need for creating a DB table for it. This is achieved by switching
the model type from transient to abstract.


luisg123v:
@moylop260 
Could you check, please?

moylop260:
👍

PR #1796, opened 2020-03-29
[IMP] company_country: Fix and improve README and error messages

luisg123v:
Error messages and module description are improved to be more verbose,
and some typos are fixed.


luisg123v:
Merging since only changes to strings and README were performed.

CC @moylop260, I wanted to do this before backporting.

PR #1798, opened 2020-03-31
[13.0][ADD] base_time_window

grindtildeath:
This module provides base classes and models to manage time windows through
`time.window.mixin`.

Example implementation for the mixin can be found in module `test_base_time_window`.

As a time window will always be linked to a related model thourgh a M2o relation,
when defining the new model inheriting the mixin, one should pay attention to the
following points in order to have the overlapping check work properly:

- Define class property `_overlap_check_field`: This must state the M2o field to
  use for the to check of overlapping time window records linked to a specific
  record of the related model.

- Add the M2o field to the related model in the `api.constrains`:

For example:

.. code-block:: python
    class PartnerTimeWindow(models.Model):
        _name = 'partner.time.window'
        _inherit = 'time.window.mixin'
        partner_id = fields.Many2one(
            res.partner', required=True, index=True, ondelete='cascade'
        )
        _overlap_check_field = 'partner_id'
        @api.constrains('partner_id')
        def check_window_no_overlaps(self):
            return super().check_window_no_overlaps()


base_time_window/models/base_time_window.py
```diff
@@ -0,0 +1,113 @@
...
+import math
+from datetime import time
+
+from psycopg2.extensions import AsIs
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class TimeWindow(models.AbstractModel):
```
lmignon: @grindtildeath According to the naming conventions, 
filename should be: `time_window_mixin.py`
classname should be: `TimeWindowMixin`

lmignon (changes requested):
IMO, to be complete, we should add a demo module with the appropriate unittest.  I prefer to keep the tests related to the logic implemented into this new addon into the same repository and without dependencies on other addon.

grindtildeath:
@lmignon I added a test_base_time_window module in the last commits, with your test suite and an extra one related to the usage I introduced with mixin.

lmignon (approved):
Great. Thank you @grindtildeath for your refactoring and generalization of this transversal concept. (Code review)

jgrandguillaume (approved):
LGTM

base_time_window/models/time_window_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+class TimeWindowMixin(models.AbstractModel):
+
+    _name = "time.window.mixin"
+    _description = "Time Window"
+    _order = "start"
+
+    # TODO patch api.constrains with field here?
+    _overlap_check_field = False
+
+    start = fields.Float("From", required=True)
```
simahawk: being a mixin this fields will land on another model: what about prefixing them properly?
Eg: `time_window_*` or `window_*`?

base_time_window/models/time_window_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+                    _("%s must be > %s")
+                    % (
+                        self.float_to_time_repr(record.end),
+                        self.float_to_time_repr(record.start),
+                    )
+                )
+            # here we use a plain SQL query to benefit of the numrange
+            # function available in PostgresSQL
+            # (http://www.postgresql.org/docs/current/static/rangetypes.html)
+            SQL = """
```
simahawk, line 46: I'd split this to `_get_time_window_sql` which returns `(sql, sql_params)`

base_time_window/models/time_window_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+            if res:
+                other = self.browse(res[0][0])
+                raise ValidationError(
+                    _("%s overlaps %s") % (record.display_name, other.display_name)
+                )
+
+    @api.depends("start", "end", "weekday_ids")
+    def _compute_display_name(self):
+        for record in self:
+            record.display_name = _("{days}: From {start} to {end}").format(
```
simahawk, line 85: what about using `tools.misc.format_time`?

base_time_window/models/time_window_mixin.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TimeWindowMixin(models.AbstractModel):
+
+    _name = "time.window.mixin"
+    _description = "Time Window"
+    _order = "start"
+
+    # TODO patch api.constrains with field here?
+    _overlap_check_field = False
```
simahawk: `_time_window_overlap_check_field`?

test_base_time_window/models/partner_time_window.py
```diff
@@ -0,0 +1,43 @@
...
+
+    partner_id = fields.Many2one(
+        "res.partner", required=True, index=True, ondelete="cascade"
+    )
+
+    @api.constrains("partner_id")
+    def check_window_no_overlaps(self):
+        return super().check_window_no_overlaps()
+
+    @api.depends("start", "end", "weekday_ids")
```
simahawk: are the following overrides needed?

test_base_time_window/models/res_partner.py
```diff
@@ -0,0 +1,33 @@
...
+from collections import defaultdict
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    time_window_ids = fields.One2many(
```
simahawk, line 13: this field and the method IMO could stay in another mixin `time.window.consumer.mixin`

PR #1800, opened 2020-04-01
10.0 base merge

ntsirintanis:
A new module, `base_merge`, adding the possibility of merging duplicate records of a given model. For this moment, merging records of self-referring models is not possible, as this is heavily WIP.
@hbrunn


pedrobaeza:
See #1628 and my comments about it.

thomaspaulb:
@pedrobaeza I wish to finish this module. We have a similar one, and during its usage in the past years I did find some differences between Odoo versions, such as:

- v10.0 does not store any field that has `store=False` ([link](https://github.com/OCA/OCB/blob/10.0/odoo/models.py#L3666)), while other Odoo's have other logic concerning this

- v8.0 has some legacy field types compared to newer Odoo's

Do you feel these version-dependent logics should be handled as if-clauses in the openupgradelib routines, or should we still include the merge routines literally in the `base_merge` module, thus allowing versions to diverge?

pedrobaeza:
I don't really get you when you say that store=False behavior has changed over versions (it's the same AFAIK), but the general approach in openupgradelib is to put the differences by version in the lib itself. Examples:

https://github.com/OCA/openupgradelib/blob/3945526d1377533a0aa39d20be88edf2a03f3145/openupgradelib/openupgrade.py#L2429
https://github.com/OCA/openupgradelib/blob/3945526d1377533a0aa39d20be88edf2a03f3145/openupgradelib/openupgrade.py#L2296

thomaspaulb:
@pedrobaeza I might be wrong about the `store` issue, but anyway yes, I'll go by `openupgradelib` then and use the versioning where needed. Thanks!

PR #1802, opened 2020-04-01
12.0 mig keychain

cvinh:


bealdav:
FYI keychain is no more use in v12

see https://github.com/OCA/server-auth/pull/171#issuecomment-604980689

PR #1804, opened 2020-04-04
[ADD] company_country: Backport module from 12.0

luisg123v:
This module allows to set a country to the main company before the
`account` module is installed, so the hook of that module installs the
correct `l10n_***` module.

This is useful because, if the company isn't already set correctly when
the `account` module is installed, the generic accounting chart will be
installed (`l10n_generic_coa`), which may be incorrect depending on your
company's country.

This is a backport from:
https://github.com/OCA/server-tools/tree/12.0/company_country

Co-authored-by: Moisés López <moylop260@vauxoo.com>


moylop260:
@luisg123v 
Could you check the travis error, please?

luisg123v:
@moylop260
Done, error fixed.

Travis is still red but it's unrelated (only `database_cleanup`)

luisg123v:
Hi @moylop260,

The target branch is already red due to the same error. Do you think it's feasible to merge this manually?

PR #1806, opened 2020-04-06
[12.0][FIX] Auditlog: avoid access rights errors while handling creation

katyukha:
This commit fixes following case:
When record has fields that user has no access to, and user tries to
create such record, the after record created auditlog tries to read all
fields of record, but record creation fails because user cannot read one
of fields.

This fix was caught on db where full-log was enabled on Partner model,
and some user has no access to signup_token field. So he was not able to
create partners.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -225,7 +225,7 @@ class AuditlogRule(models.Model):
                 new_values.setdefault(new_record.id, {})
                 for fname, field in new_record._fields.items():
                     new_values[new_record.id][fname] = field.convert_to_read(
-                        new_record[fname], new_record)
+                        new_record.sudo()[fname], new_record.sudo())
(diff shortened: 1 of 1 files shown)


katyukha:
Hi @sebalix , it seems that CI job was stopped, so tests were not actually ran. Can you (or may be someone else) restart job and review this PR.

Thanks

sebalix:
Hi @katyukha, the build seems fine excepting the runbot. If you check the logs of it you'll see no errors related to auditlog, but there are some warnings about the `module_analysis` module. Feel free to fix in a separated commit if you are able to!
Now about `auditlog` fix, a unit reproducing the issue is welcomed, I don't think we have such tests about access rights etc. I will check later if the fix is right or not, having a test would help in this case.

thomaspaulb:
@katyukha Could you rebase to retrigger tests

PR #1813, opened 2020-04-16
[10.0] [ADD] letsencrypt: Use official acme client, support wildcard certificates

janverb:
This is a continuation of #1194.

Using `acme` instead of `acme-tiny` makes it possible to do DNS validation and request wildcard certificates. By default you have to use a shell script to set DNS records for that. Other modules can be created to support specific DNS providers, e.g. OCA/l10n-netherlands#259.

Instead of requesting a new certificate every 11 weeks, we do a daily check and renew if the certificate expires within 30 days (or if new domain names have been added). That way we retry automatically every day if it fails. `certbot` uses the same strategy.


hbrunn (approved):

letsencrypt/models/letsencrypt.py
```diff
@@ -2,167 +2,464 @@
...
+                challenge
+                for challenge in authorizations.body.challenges
+                if challenge.chall.typ == TYPE_CHALLENGE_HTTP
+            ]
+            other_challenges = [
+                challenge
+                for challenge in authorizations.body.challenges
+                if challenge.chall.typ != TYPE_CHALLENGE_HTTP
+            ]
+            for challenge in http_challenges + other_challenges:
```
hbrunn: I think we need different preferences here depending on whether we have a dns provider configured or not. I'm requesting `['*.domain', 'domain']`, and don't have an externally accessible Odoo at all. So in my case the DNS challenge is preferred.

letsencrypt/models/letsencrypt.py
```diff
@@ -2,167 +2,464 @@
...
+                    update.challenge, acme.challenges.DNSResponse()
+                )
+
+        # let them know we are done and they should check
+        backoff = int(ir_config_parameter.get_param('letsencrypt.backoff', 3))
+        deadline = datetime.now() + timedelta(minutes=backoff)
+        try:
+            order_resource = client.poll_and_finalize(authzr, deadline)
+        except acme.errors.ValidationError:
+            _logger.error("Let's Encrypt validation failed!")
```
hbrunn, line 304:
```suggestion
            _logger.error("Let's Encrypt validation failed!")
            for authz in error.failed_authzrs:
                for challenge in authz.body.challenges:
                    _logger.error(str(challenge.error))
```

letsencrypt/models/letsencrypt.py
```diff
@@ -2,167 +2,464 @@
...
+                client.answer_challenge(
+                    update.challenge, acme.challenges.DNSResponse()
+                )
+
+        # let them know we are done and they should check
+        backoff = int(ir_config_parameter.get_param('letsencrypt.backoff', 3))
+        deadline = datetime.now() + timedelta(minutes=backoff)
+        try:
+            order_resource = client.poll_and_finalize(authzr, deadline)
+        except acme.errors.ValidationError:
```
hbrunn:
```suggestion
        except acme.errors.ValidationError as error:
```

letsencrypt/models/letsencrypt.py
hbrunn: (just the validation error in the log doesn't help a lot)

hbrunn (changes requested):
when using it myself I came across a few small things

letsencrypt/models/letsencrypt.py
janverb: Hm, doing that automatically would break other use cases, like when not all of your domains go through that DNS provider. DNS challenges are also slower and less reliable.

The rigorous solution could be to add a fully-fledged `letsencrypt.domain` model for per-domain configuration records. That could also make it possible to use different DNS providers for different domains, which you currently can't do. But that would be overengineering.

A "prefer DNS challenges" checkbox seems reasonable enough. I'll add that.

I do wonder whether a self-signed certificate would make more sense if Odoo isn't externally accessible.

hbrunn: the setting is VPN + clients that aren't under our control, so self signed won't work properly there. Thanks for the flag, I indeed think a config per domain would be overkill

janverb: I've implemented this.

hbrunn (approved):

NL66278:
@janverb Can you solve the conflict on this one?

NL66278 (approved):
:+1: LGTM and great work!

janverb:
The filename for the certificate is based on the first configured altname. This is for backward compatibility with the old version. But it does mean that rearranging the altnames changes the certificate filename, which is unfortunate.

One way around it would be to make the filename configurable, but that's messy. If not for compatibility I think it would be ideal to have a fixed filename.

Maybe the migration script could replace the old certificate by a symlink pointing to the new fixed location? How does that sound?

NL66278:
@janverb Or the code could take the configured filename when filled, and else base it on the first name in altnames? Or migration could fill the configurable value with the fist name in altnames. In both cases there would be no disruption to existing installations. I would prefer the second option as in my opinion more "clean".

janverb:
Is there a good use case for a configurable filename besides compatibility? It's nice to keep the compatibility ugliness in the migration script instead of the main implementation.

The symlink solution is a little dirty, but it only affects existing installations, instead of adding a user-facing complication to all installations going forward.

NL66278:
Having the filename dependent on whatever is the first name in altnames is IMHO also something that might catch users unawares. An alternative might be to have a completely fixed position - that is fiixed relative to the data directory - like:
<datadirectory>/letsencrypt/letsencrypt.[crt|key] and the have migrationscript provide a symbolic link to the new location.

In this way nothing extra to configure for the users, and file location completely predictable....

janverb:
That was my proposal:

> Maybe the migration script could replace the old certificate by a symlink pointing to the new fixed location?

If we both arrived at the same conclusion then it's probably a good idea :slightly_smiling_face:

NL66278:
@gurneyalex This is ready to merge since a few month's. Could you have a look and maybe merge it?

NL66278:
@dreispt Could you try this one more time? I looked at the failing jobs and for some strange reason it wanted to install a non existing version of setuptools (`> 54.0`). I can not find any reason for that in the requirements.txt, so I hope this is solved by now. Otherwise I will continue investigating what the problem might be.

NL66278:
This one could be closed.

dreispt:
Superseded by #2205

PR #1814, opened 2020-04-17
[12.0][FIX] base_remote: search =ilike

etobella:
According to ietf, domain names should be case insensitive (https://tools.ietf.org/html/rfc4343)

@jarroyomorales


jarroyomorales (approved):
LGTM, can it be fast-tracked?

PR #1818, opened 2020-04-21
[6.1][FIX] Do not break on malformed messages.

NL66278:
fetchmail_attach_from_folder. backport improvements in exception handling on
10.0 to 6.1, to prevent exceptions in retireving one mail from breaking
message retrieval for all other messages.


NL66278:
With the last commit, [1343e99](https://github.com/OCA/server-tools/commit/1343e994c47d4c3368618ffee1cfa6c2ef4b2f96), performance improved a lot as well, because messages that have already been retrieved are not parsed fully, which usually will be the bulk of messages.

PR #1819, opened 2020-04-21
[12.0][MIG] dead_mans_switch_client

janverb:
There are a few general improvements to the module that could in principle be backported.

The old version generated a default configuration as demo data that wasn't really useful. That's gone now.

The old version went out of its way to work if `psutil` wasn't installed or was an old version, but `psutil` is a pinned dependency of Odoo so that's unnecessary.

@hbrunn


thomaspaulb:
Author not working with Odoo anymore

PR #1821, opened 2020-04-22
[12.0] Added purge actions for ir.actions.act_window and ir.ui.view

ypapouin:
Purge actions:
This is essentially a copy of the purge obsolete menu wizard.
I made it to be able to remove obsolete actions that blocks opening view settings in odoo 12 due to _compute_search_view and _compute_params from ir_actions.py raising an exception.

### DIFF
--- a/database_cleanup/__manifest__.py
+++ b/database_cleanup/__manifest__.py
@@ -17,6 +17,7 @@
         'views/purge_data.xml',
         "views/create_indexes.xml",
         'views/purge_properties.xml',
+        'views/purge_actions.xml',
         'views/menu.xml',
     ],
     'installable': True,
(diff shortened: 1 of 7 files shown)


database_cleanup/__manifest__.py
```diff
@@ -4,7 +4,7 @@
     'name': 'Database cleanup',
     'version': '12.0.1.0.3',
     'author': "Therp BV,Odoo Community Association (OCA)",
-    'depends': ['base'],
+    'depends': ['base', 'mail'],
```
thomaspaulb, line 7: Why the new dependency?

thomaspaulb:
@ypapouin Needs a rebase on latest

PR #1822, opened 2020-04-23
[FIX] unlink cronjobs for the model before purging it

hbrunn:
if we purge a model that has a cronjob, the server action will be [deleted](https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/ir_actions.py#L386), but as the `ir.cron` model [restricts](https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/ir_cron.py#L52) deletion of the linked server action, it breaks at this point without the patch

### DIFF
--- a/database_cleanup/models/purge_models.py
+++ b/database_cleanup/models/purge_models.py
@@ -73,6 +73,13 @@ class CleanupPurgeLineModel(models.TransientModel):
             self.env['ir.model.constraint'].search([
                 ('model', '=', line.name),
             ]).unlink()
+            cronjobs = self.env['ir.cron'].with_context(
+                active_test=False
+            ).search([
+                ('model_id.model', '=', line.name),
+            ])
+            if cronjobs:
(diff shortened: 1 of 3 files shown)


hbrunn:
PS: I took the liberty to [pin](https://github.com/OCA/server-tools/pull/1822/files#diff-b4ef698db8ca845e5845c4618278f29aR6-R7) the library [module_analysis](https://github.com/OCA/server-tools/blob/12.0/module_analysis/models/ir_module_module.py#L148) uses to a version where the API used is not deprecated. This probably should be addressed in a new PR where somebody acquainted with it like @legalsylvain can update it for the new versions.

sergiocorato (approved):
LGTM

legalsylvain (approved):
Thanks @hbrunn. 

I'll take a look on this new warnings.

CasVissers-360ERP (approved):
👍

PR #1824, opened 2020-04-24
[IMP] add helper for getting a self env with the tech user

sebastienbeau:


sebastienbeau:
@acsonefho @Cedric-Pigeon

Cedric-Pigeon (approved):
code review. Thanks

base_technical_user/models/models.py
```diff
@@ -0,0 +1,21 @@
...
+    self_sudoer = self
+    tech_user = self.env.user.company_id.user_tech_id
+    if tech_user:
+        self_sudoer = self.sudo(tech_user.id)
+    elif raise_if_missing:
+        raise UserError("The technical user is missing in the company {}".format(self.env.user.company_id.name))
+    return self_sudoer
+
+
+models.BaseModel.sudo_tech = sudo_tech
```
acsonefho: Why not doing like this:

```
class Base(models.AbstractModel):
    _inherit = 'base'

    def sudo_tech(self, raise_if_missing=False):
        self_sudoer = self
        tech_user = self.env.user.company_id.user_tech_id
        if tech_user:
            self_sudoer = self.sudo(tech_user.id)
        elif raise_if_missing:
            raise UserError("The technical user is missing in the company {}".format(self.env.user.company_id.name))
        return self_sudoer
```
It's more "Odoo"?

acsonefho (commented):
Great! Just one question

base_technical_user/models/models.py
sebastienbeau: Sorry I mix the version. I was think that this was not possible on 10 and only work since 11. But it's work since 9. I fix it

base_technical_user/models/models.py
```diff
@@ -0,0 +1,22 @@
...
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    def sudo_tech(self, raise_if_missing=False):
+        self_sudoer = self
+        tech_user = self.env.user.company_id.user_tech_id
+        if tech_user:
+            self_sudoer = self.sudo(tech_user.id)
+        elif raise_if_missing:
+            raise UserError("The technical user is missing in the company {}".format(
```
simahawk: to be translated?

base_technical_user/tests/test_sudo_tech.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo.exceptions import UserError
+from odoo import SUPERUSER_ID
+
+
+class SudoTechCase(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(SudoTechCase, cls).setUpClass()
+        cls.user_tech = cls.env["res.users"].create({"login": "tech", "name": "tech"})
```
simahawk: always use `tracking_disable` + `no_reset_password` when creating users in tests ;)

simahawk (changes requested):

base_technical_user/models/models.py
sebastienbeau: done

base_technical_user/tests/test_sudo_tech.py
sebastienbeau: done

sebastienbeau:
@simahawk fixed !

simahawk (approved):

acsonefho (approved):
Nice!

PR #1828, opened 2020-05-07
[10.0] server-tools - mass_editing many2many fields as version 13.0

SimoneVagile:
These changes modify the tree viewing of many2many fields in mass editing by multi-tag fields


mass_editing/migrations/10.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    # First remove the obsolete constraint created by the existence of the
```
legalsylvain: I think that if you change the model in the 10.0 (that is OK), it will break the migration script of the 13.0 revision. 
Maybe we should add a test in the pre-migration script of the module 13.0 testing for exemple, if mass_editing table exists. If yes, skip the migration. Don't you think ?

Maybe you should add it also here, if for example, this change is also backported in 8.0 and 9.0.

kind regards.

legalsylvain (commented):
One question. The rest LGTM. 

thanks !

mass_editing/migrations/10.0.2.0.0/pre-migration.py
SimoneVagile: @legalsylvain thanks for your answer.
That's a good advice. I'm going to work on it.

Regards

mass_editing/migrations/10.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,36 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    # Check if table has already migrated
+    try:
+        cr.execute(
```
legalsylvain: I think that this line will always fail, as pre-migration script has renamed mass_object. 
Maybe move the post-migration in the pre-migration will fix the problem. 
You should only add a script that add a table. Don't you think ?

Otherwise, the concept is correct. Thansk !

SimoneVagile: About your cue, I got idea from server-ux repository, branch 13.0. There are pre-migration.py and post-migration.py as this, and I kept this structure for readability reasons.
Guess my last modify is tricky, I want to test your suggestion.

Thank you.

legalsylvain: Indeed, the 13.0 was the only one migration script. Now if, this migration script is copied in many versions, it is not valid anymore and should be adapted.

tafaRU (changes requested):
Thanks for your contrib!
Please find my following comments:

* commit message should be more compliat with https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message
* travis and runbot builds are red: please check them

The rest LGTM.

mass_editing/tests/test_mass_editing.py
```diff
@@ -26,27 +26,14 @@ def setUp(self):
         self.partner_model = model_obj.\
             search([('model', '=', 'res.partner')])
         self.user_model = model_obj.search([('model', '=', 'res.users')])
-        self.fields_model = self.env['ir.model.fields'].\
+        self.fields_model = self.env['ir.model.fields']. \
```
tafaRU: please remove the empty space before `\`.

tafaRU (changes requested):
I think you could add yourself to the README file.
Can you also take the occasion to update it to the new [readme](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst) format? (**optional**)

tafaRU (approved):
Once runbot is green

SimoneVagile:
@legalsylvain please update your review

legalsylvain (approved):
LGTM. Thanks ! 

Note : test if ``mass_editing`` table exists should be added also in the V13 migration script.

mass_editing/wizard/mass_editing_wizard.py
```diff
@@ -41,172 +41,34 @@ def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
...
-                            'attrs': ("{'invisible':[('selection__" +
-                                      field.name + "','=','remove')]}"),
-                            'colspan': '4',
-                        })
+                etree.SubElement(
+                    xml_group,
+                    "field",
+                    {"name": "selection__" + field.name, "colspan": "2"},
+                )
+                field_vals = {"name": field.name, "nolabel": "1", "colspan": "4"}
```
rousseldenis, line 67: I will test this one.

But @SimoneVagile @legalsylvain, could we just add two hook functions to get field dict options ?

```suggestion
                field_vals = self._get_field_options(field)
```

rousseldenis: This will allow overriding (I have the case I want to add a domain here)

legalsylvain: of course ! nice addition !

rousseldenis: I would have done the PR on Simone repo, but it's not available when creating PR

rousseldenis: @SimoneVagile Could you rebase ?

rousseldenis: And can you cherry-pick the last commit from https://github.com/acsone/server-tools/commits/10.0-imp-task-13269 ?

rousseldenis:
@legalsylvain @SimoneVagile @tafaRU I did another PR with rebasing and including a commit for the hook I talked here above. Maybe we could use that one.

simahawk:
done on #1856

PR #1829, opened 2020-05-08
[13.0] base_jsonify: support callable parser

simahawk:
You can now pass a callable to manipulate output values.

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -1,6 +1,8 @@
-# © 2017 Akretion (http://www.akretion.com)
+# Copyright 2017 Akretion (http://www.akretion.com)
 # Sébastien BEAU <sebastien.beau@akretion.com>
 # Raphaël Reverdy <raphael.reverdy@akretion.com>
+# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
@@ -33,6 +35,8 @@ class Base(models.AbstractModel):
                 'number',
                 'create_date',
                 ('partner_id', ['id', 'display_name', 'ref'])
+                ('shipping_id', callable)
+                ('delivery_id', "record_method")
                 ('line_id', ['id', ('product_id', ['name']), 'price_unit'])
             ]
 
@@ -64,6 +68,7 @@ class Base(models.AbstractModel):
     def _jsonify_value(self, field_name):
         field_type = self._fields[field_name].type
         value = self[field_name]
+        # TODO: we should get default by field (eg: char field -> "")
         if value is False and field_type != "boolean":
             value = None
         elif field_type == "date":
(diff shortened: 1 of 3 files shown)


base_jsonify/models/models.py
```diff
@@ -54,31 +58,52 @@ def jsonify(self, parser):
...
-                        value = value.isoformat()
-                    res[json_key] = value
+                    res[json_key] = rec._jsonify_value(field_name)
             result.append(res)
         return result
+
+    def _jsonify_value(self, field_name):
+        field_type = self._fields[field_name].type
+        value = self[field_name]
+        # TODO: we should get default by field (eg: char field -> "")
```
janverb, line 71: Would that be a good idea? A missing value doesn't have to mean the same thing as the default value.

base_jsonify/models/models.py
```diff
@@ -54,31 +58,52 @@ def jsonify(self, parser):
...
+        elif isinstance(subparser, str):
+            # a method on the record itself
+            method = getattr(self, subparser, None)
+            if method:
+                value = method(field_name)
+            else:
+                self._jsonify_bad_parser_error(field_name)
+        else:
+            field = self._fields[field_name]
+            if not field.relational or field.type == "reference":
```
janverb: Did you mean this?
```suggestion
            if not (field.relational or field.type == "reference"):
```

base_jsonify/models/models.py
```diff
@@ -54,31 +58,52 @@ def jsonify(self, parser):
...
+        if callable(subparser):
+            # a simple function
+            value = subparser(self, field_name)
+        elif isinstance(subparser, str):
+            # a method on the record itself
+            method = getattr(self, subparser, None)
+            if method:
+                value = method(field_name)
+            else:
+                self._jsonify_bad_parser_error(field_name)
```
janverb, line 97: I'm wondering if this is the best API to provide. I would have expected just `subparser(self)`.

Why is the field name passed as an argument? Wouldn't you usually already know the field name? Or do you want to be able to reuse the same function/method for multiple fields with different names?

Do you have real-world examples?

base_jsonify/models/models.py
```diff
@@ -54,31 +58,52 @@ def jsonify(self, parser):
...
+            # Get the timestamp converted to the client's timezone.
+            # This call also add the tzinfo into the datetime
+            # object
+            value = fields.Datetime.context_timestamp(self, value)
+            value = value.isoformat()
+        return value
+
+    def _jsonify_value_subparser(self, field_name, subparser):
+        value = None
+        rec_value = self[field_name]
```
janverb: The first two cases don't use this value, so I think it's better to move it into the `else:` block below. That way you can also make "fake" computed fields that don't correspond to any field on the model. If you try that now it crashes at this line.

janverb (changes requested):
This looks like a good way to make the module more flexible. The refactoring is nice too.

I left a few notes.

base_jsonify/models/models.py
simahawk: a non valued boolean should be "False" not "None" and a non valued char should be "''" and not "None" IMO.

base_jsonify/models/models.py
simahawk: because you can pass the same subparser for different fields, this is the main reason.

base_jsonify/models/models.py
janverb: I feel like the difference between "no value" and "a value of length zero" could be significant sometimes. But I can't think of any examples and I can't find any Odoo code making the distinction so you're probably right.

base_jsonify/models/models.py
simahawk: nope. Unfortunately a "reference" field is not flagged as "relational" :sob:

janverb: But the way it's written now means that it always raises the error if the field is a reference. `not` has higher precedence, so it means this: `(not field.relational) or (field.type == "reference")`

Here's a test case:
```python
self.env.ref("base.menu_decimal_precision_form").jsonify([("action", ["name"])])
```
It currently errors. After the change it returns `[{'action': {'name': 'Decimal Accuracy'}}]` as it should.

simahawk:
@janverb thanks for your feedback, I addressed your last comments.

guewen (approved):

janverb (approved):

Issue #1122, opened 2018-01-09
auditlog -> rule ->  Log Reads then give this error


akhodifad:
![selectreallog](https://user-images.githubusercontent.com/32675315/34718940-9faa6dce-f55e-11e7-9aca-a905e68b836d.png)
![openso](https://user-images.githubusercontent.com/32675315/34718939-9f7b039a-f55e-11e7-9c57-08b986dd9235.png)

avmediadesign:
I'm facing the same error on v10 Enterprise.
```
 File "/odoo/odoo-server/odoo/models.py", line 4911, in with_context
    context = dict(args[0] if args else self._context, **kwargs)
ValueError: dictionary update sequence element #0 has length 10; 2 is required
```
It appears on any model.
The logs are properly recorded but this error message pops up.
Any idea?

uakhhhh:
I'am same error on v10 community.

Issue #946, opened 2017-08-22
Audit Log Read Problem. ValueError: dictionary update sequence element #0 has length 20; 2 is required 


anniquetapha:
module: auditlog
version: 10.0

**Steps to reproduce**
- ...When i create a read audit on a model, and i try to read open the records in that model, it gives me "Traceback (most recent call last):
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 638, in _handle_exception
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 675, in dispatch
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 331, in _call_function
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\service\model.py", line 119, in wrapper
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 324, in checked_call
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 933, in call
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 504, in response_wrap
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\addons\web\controllers\main.py", line 885, in call_kw
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\addons\web\controllers\main.py", line 877, in _call_kw
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\api.py", line 681, in call_kw
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\api.py", line 670, in call_kw_multi
File "C:\Program Files (x86)\Odoo 10.0\server\odoo\models.py", line 4890, in with_context
ValueError: dictionary update sequence element #0 has length 20; 2 is required"

**Current behavior**
The error above pops up

**Expected behavior**
It should allow me to open the record without problems

Issue #924, opened 2017-08-04
ValueError: dictionary update sequence element #0 has length 20; 2 is required


anniquetapha:
module: auditlog
version: 10.0

**Steps to reproduce**
- ...When i create a read audit on a model, and i try to read open the records in that model, it gives me "Traceback (most recent call last):
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 638, in _handle_exception
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 675, in dispatch
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 331, in _call_function
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\service\model.py", line 119, in wrapper
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 324, in checked_call
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 933, in __call__
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\http.py", line 504, in response_wrap
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\addons\web\controllers\main.py", line 885, in call_kw
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\addons\web\controllers\main.py", line 877, in _call_kw
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\api.py", line 681, in call_kw
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\api.py", line 670, in call_kw_multi
  File "C:\Program Files (x86)\Odoo 10.0\server\odoo\models.py", line 4890, in with_context
ValueError: dictionary update sequence element #0 has length 20; 2 is required"

**Current behavior**
The error above pops up
**Expected behavior**
It should allow me to open the record without problems

nikeshscs:
Hello,
module: auditlog  (https://apps.odoo.com/apps/modules/10.0/auditlog/)
version: 10.0 CE
I'm facing the same problem with auditlog module.
Below zip is video clip that shows the how that error is produce.

[auditlog-2019-06-03_17.39.16.zip](https://github.com/OCA/server-tools/files/3246933/auditlog-2019-06-03_17.39.16.zip)

Thanks.

Issue #747, opened 2017-02-27
[10.0] Audit trail logs - Errors on read


maclarensg:
When create an audit log rule with log read enable,

The following error appears

```
2017-02-27 01:01:53,853 17231 ERROR asme odoo.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 638, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 675, in dispatch
    result = self._call_function(**self.params)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 331, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/service/model.py", line 119, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 324, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 933, in __call__
    return self.method(*args, **kw)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/http.py", line 504, in response_wrap
    response = f(*args, **kw)
  File "/home/gavinyap/asme/odoo/addons/web/controllers/main.py", line 878, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/gavinyap/asme/odoo/addons/web/controllers/main.py", line 870, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/api.py", line 681, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/api.py", line 670, in call_kw_multi
    recs = self.with_context(context or {}).browse(ids)
  File "/home/gavinyap/Documents/gitlab/asme/odoo/odoo/models.py", line 4865, in with_context
    context = dict(args[0] if args else self._context, **kwargs)
ValueError: dictionary update sequence element #0 has length 9; 2 is required
```

**Step to produce**
1) Install Note, Audit trail
2) Create log rule on Note object with log read
[...]

PR #1830, opened 2020-05-15
[10.0][FIX] - Audit trail logs - Errors on read

YogeshMahera-SerpentCS:
Fix the error for users who are getting an error on read model. in view. for audit log
Getting error(`ValueError: dictionary update sequence element #0 has length) while read the log for any object, Here is the list of issues who are getting the same issue for read any object.

- #1122
- #924
- #946
- #747

Error: `ValueError: dictionary update sequence element #0 has length 10; 2 is required`

### DIFF
--- a/auditlog/__manifest__.py
+++ b/auditlog/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': "Audit Log",
-    'version': "10.0.1.0.1",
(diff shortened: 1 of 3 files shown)


TimLai125:
@YogeshMahera-SerpentCS I think the version has to be bumped?

PR #1831, opened 2020-05-15
[WIP][12.0][IMP] add by_method exception

sebastienbeau:
need to add the test, and finish it  @kevinkhao

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -47,13 +47,16 @@ class ExceptionRule(models.Model):
         "not. Use failed = True to block the exception",
     )
 
-    @api.constrains("exception_type", "domain", "code")
+    @api.constrains("exception_type", "domain", "code", "model")
     def check_exception_type_consistency(self):
         for rule in self:
             if (
                 (rule.exception_type == "by_py_code" and not rule.code)
                 or (rule.exception_type == "by_domain" and not rule.domain)
-                or (rule.exception_type == "by_method" and not rule.method)
+                or (
+                    rule.exception_type == "by_method"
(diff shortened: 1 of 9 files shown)


kevinkhao:
https://github.com/akretion/server-tools/pull/25

sebastienbeau:
@bealdav @hparfr @florian-dacosta I think you can be interested by this improvement

hparfr (changes requested):
It's a nice feature, but about the implementation, I suggest to limit methods to hard-coded known methods.

changing to `method = fields.Selection()` instead of 
`fields.Char` will do the trick.

`_check_method_valid` will be useless and it will be safer.

sebastienbeau:
@hparfr great idea, thanks, Done ok for you ?

base_exception/models/base_exception.py
```diff
@@ -53,10 +53,6 @@ def check_exception_type_consistency(self):
             if (
                 (rule.exception_type == "by_py_code" and not rule.code)
                 or (rule.exception_type == "by_domain" and not rule.domain)
-                or (
-                    rule.exception_type == "by_method"
-                    and not self._check_method_valid(rule.method, rule.model)
-                )
```
hparfr: Please add back a check on rule.method which can be undefined.
```py
 or (rule.exception_type == "by_method" and not rule.method)
```

sebastienbeau: Thanks for the feedback ! DONE

hparfr (approved):
Thanks

florian-dacosta (approved):

bealdav (approved):
LGTM

PR #1834, opened 2020-05-19
[13.0][MIG] module_analysis

xavierjimenez:
Migration to 13.0 of **module_analysis**


pedrobaeza:
Can you take the occasion to squash fixup commits together? And please check also Travis

xavierjimenez:
I've squashed the fixup commits but travis is red because of pre-commit. Apparently, black changes the format of some files so that it gives flake8 error later, how can I fix that?

misern2:
@xavierjimenez Try to fixup the precommits error before the commit. You can run precommit in your local environment in order to fix the issues

PR #1836, opened 2020-05-24
[MIG] base_fontawesome: Migration to 13.0

tonygalmiche:


legalsylvain (commented):
Hi ! Thanks for porting this module. 
Two questions :  
- why this module is not in the OCA / web repository ? 
- what about to add the major version in the name ? Base_fontawsome_5. I'm not sure fontawsome is retro compatible. I mean if we want to swtich to FA6 maybe we should have two modules (one for the 5 and one for the 6). Not sure on that point. Not a FA expert.

Otherwise LGTM : thanks ! 
Code review / no test.

Kind regards.

tonygalmiche:
Thanks @legalsylvain for your review.
As a beginner on oca I have no opinion for your 2 questions.

etobella:
@legalsylvain Hi, this module was added from the comments from https://github.com/OCA/web/pull/1042

Is not a problem moving it, but that is the main reason.

About the Compatibility, usually, FA is making it compatible, so I assume when FA6 is ready, we can make it compatible. The version is marked on the version of the module: 13.0.5.7.1

etobella (approved):
LGTM

albariera (approved):
Technical review LGTM :+1:

etobella:
@OCA/tools-maintainers Can this be merged?

PR #1840, opened 2020-05-27
[11.0] [MIG] letsencrypt: Use official acme client, support wildcard certificates

janverb:
Migration of #1813.


letsencrypt/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def migrate_cron(env):
+    # Any interval that was appropriate for the old version is inappropriate
+    # for the new one, so it's ok to clobber it.
+    # But tweaking it afterwards is fine, so noupdate="1" still makes sense.
+    jobs = (
+        env["ir.cron"]
+        .with_context(active_test=False)
+        .search([("model", "=", "letsencrypt"), ("function", "=", "cron")])
```
NL66278: This is giving errors on 11.0, as cron jobs are no longer directly using methods on models anymore, but using server actions:
```

  File "bin/upgrade_odoo", line 63, in <module>
    sys.exit(anybox.recipe.odoo.runtime.upgrade.upgrade('/home/odorboprd11/odoo/upgrade.py', 'run', '/home/odorboprd11/odoo/etc/odoo.cfg', '/home/odorboprd11/odoo'))
  File "/home/odorboprd11/odoo/eggs/anybox.recipe.odoo-1.9.3.dev0-py3.7.egg/anybox/recipe/odoo/runtime/upgrade.py", line 151, in upgrade
    statuscode = getattr(upgrade_module, upgrade_callable)(session, logger)
  File "/home/odorboprd11/odoo/upgrade.py", line 19, in run
    session.update_modules(['all'])
  File "/home/odorboprd11/odoo/eggs/anybox.recipe.odoo-1.9.3.dev0-py3.7.egg/anybox/recipe/odoo/runtime/session.py", line 424, in update_modules
    self._registry = Registry(db)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/registry.py", line 61, in __new__
    return cls.new(db_name)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/registry.py", line 85, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/loading.py", line 376, in load_modules
    force, status, report, loaded_modules, update_module, models_to_check)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/loading.py", line 274, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/loading.py", line 188, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/home/odorboprd11/odoo/parts/odoo/odoo/modules/migration.py", line 190, in migrate_module
    migrate(self.cr, installed_version)
  File "/home/odorboprd11/odoo/parts/server-tools/letsencrypt/migrations/11.0.2.0.0/post-migrate.py", line 38, in migrate
    migrate_cron(env)
  File "/home/odorboprd11/odoo/parts/server-tools/letsencrypt/migrations/11.0.2.0.0/post-migrate.py", line 25, in migrate_cron
    .search([("model", "=", "letsencrypt"), ("function", "=", "cron")])
  File "/home/odorboprd11/odoo/parts/odoo/odoo/models.py", line 1501, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/models.py", line 3838, in _search
    query = self._where_calc(args)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/models.py", line 3628, in _where_calc
    e = expression.expression(domain, self)
  File "/home/odorboprd11/odoo/parts/odoo/odoo/osv/expression.py", line 669, in __init__
    self.parse()
  File "/home/odorboprd11/odoo/parts/odoo/odoo/osv/expression.py", line 847, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'model' in leaf "<osv.ExtendedLeaf: ('model', '=', 'letsencrypt') on ir_cron (ctx: )>"
```

janverb: Ah! I should have paid more attention there. It should be fixed now.

NL66278 (approved):
:+1: LGTM

hbrunn (approved):

hbrunn:
@janverb I'm curious what's WIP here? I just yanked this into v12 and the code just works

janverb:
@hbrunn Using `web.base.url` gives trouble because it can change when you don't want it to, even to a domain that the module rejects, so I'd like to redesign that. This is my current idea:

- If `letsencrypt.altnames` is empty, use `web.base.url`.
- If `letsencrypt.altnames` has domains, ignore `web.base.url`.
- During migration, if there are altnames and `web.base.url` has a usable value (passes `._validate_domain()`), add `web.base.url` as one of the altnames.

I think that's saner behavior without causing any regressions.

(But current users of this new version would have to take care to add `web.base.url` to their altnames, if they have any.)

hbrunn:
[`web.base.url.freeze`](https://github.com/OCA/OCB/blob/13.0/odoo/addons/base/models/res_users.py#L652) FTW

It does seem a little bit weird to me to decouple the base url from the cert name as Odoo presupposes everywhere this is the address. But as it doesn't add much extra complexity I also don't oppose this.

NL66278:
@janverb If migration is needed anyway I would prefer a name like letsencrypt.domains above letsencrypt.altnames, as in many cases there will not be multiple domains, but only one, and this will be the main  or only name, not an alternative one. @hbrunn We know about the freeze parameter, but it will decrease the usability of letsencrypt if users will be forced to set this parameter. An option might be to automatically set the freeze parameter, but maybe this might cause other problems. Although I have a hard time understanding why the automatic adoption, or rather violation, of web.base.url exists in the first place.

janverb:
@NL66278 The name of the parameter is not user-facing (there's a `res.config.settings` frontend), and it's a technically correct name as even the "main" domain name gets an altname entry on the certificate, so I don't think it's worth changing.

letsencrypt/data/ir_cron.xml
```diff
@@ -2,12 +2,12 @@
 <openerp>
```
NL66278: This can be changed into odoo as well (and the noupdate made an attribute of the odoo element and the data element removed).

NL66278:
@gurneyalex Same for this one.

PR #1842, opened 2020-05-28
[13.0][ADD] Add new module: sequence_custom_data

acsonefho:
This module allow to add new variables into the sequence (ir.sequence) generation (and keep Odoo variable available).


pedrobaeza:
Please don't use plurals in module name. You can call it `sequence_custom_data` or `sequence_custom_info`.

thomaspaulb:
@acsonefho Do you have an example dependent module? We have a case whereby we want to add a random number, maybe this could become `sequence_custom_data_random`

acsonefho:
> @acsonefho Do you have an example dependent module? We have a case whereby we want to add a random number, maybe this could become `sequence_custom_data_random`

I can give you this example into a dependent module who adds new variables to generate the sequence.
```
class IrSequence(models.Model):
    _inherit = "ir.sequence"

    def _get_special_values(self, date=None, date_range=None):
        """
        Get the dict with custom values to use into prefix/suffix.
        The expected code format into the prefix/suffix is not the same than Odoo.
        You should use {my_code} who will be replaced by the value into the dict.
        :return: dict
        """
        values = super()._get_special_values(date=date, date_range=date_range)
        partner_id = self.env.context.get("partner_id", 0)
        partner_ref = self.env.context.get("purchase_partner_ref", "")
        partner = self.env["res.partner"].browse(partner_id)
        values.update(
            {
                "partner_lot_code": partner.lot_code or "",
                "purchase_reference": partner_ref or "",
            }
        )
        return values
```

And the sequence's prefix is: `{partner_lot_code}{purchase_reference}`

thomaspaulb:
@acsonefho Thank you!

thomaspaulb (approved):
Nice module and well-coded.

acsonefho:
@rousseldenis Can you check and merge please?

rousseldenis:
> @rousseldenis Can you check and merge please?

Hehe. No super powers here

rousseldenis (approved):
@acsonefho In order to avoid code to be called each time, maybe a field on ir.sequence 'Use special values' ?

acsonefho:
@OCA/server-environment-maintainers Do you have super-powers here to merge please?

PR #1844, opened 2020-06-02
[12.0][IMP] base_custom_info: Create a specific view

etobella:
This PR tries to improve the view of custom_info. It is necessary to change the view with `custom_info_ids` field and by

```
<field name="custom_info_ids"
                               colspan="4"
                               nolabel="1"
                               mode="custom_info"
                        />
```

custom_info

@Tecnativa @pedrobaeza @dalonsod

### DIFF
--- a/base_custom_info/__manifest__.py
+++ b/base_custom_info/__manifest__.py
@@ -7,11 +7,12 @@
     'name': "Base Custom Info",
     'summary': "Add custom field in models",
     'category': 'Tools',
-    'version': '12.0.1.0.2',
+    'version': '12.0.2.0.0',
     'depends': [
         'base_setup',
     ],
     'data': [
+        'views/webclient_templates.xml',
         'security/ir.model.access.csv',
         'security/res_groups_security.xml',
         'views/custom_info_category_view.xml',
@@ -23,6 +24,7 @@
         'views/res_partner_view.xml',
         'wizard/res_config_settings_view.xml',
     ],
+    'qweb': ['static/src/xml/custom_info_item.xml'],
     'demo': [
         'demo/custom.info.category.csv',
         'demo/custom.info.template.csv',
--- a/base_custom_info/models/custom_info_value.py
+++ b/base_custom_info/models/custom_info_value.py
@@ -34,7 +34,6 @@ class CustomInfoValue(models.Model):
     )
     property_id = fields.Many2one(
         comodel_name='custom.info.property', required=True, string='Property',
-        readonly=True,
     )
     property_sequence = fields.Integer(
         related="property_id.sequence", store=True, index=True, readonly=True,
@@ -50,13 +49,16 @@ class CustomInfoValue(models.Model):
     field_type = fields.Selection(
         related="property_id.field_type", readonly=True,
     )
+    widget = fields.Selection(
+        related="property_id.widget", readonly=True,
+    )
     field_name = fields.Char(
         compute="_compute_field_name",
         help="Technical name of the field where the value is stored.",
     )
     required = fields.Boolean(related="property_id.required", readonly=True)
     value = fields.Char(
-        compute="_compute_value", inverse="_inverse_value",
+        compute="_compute_value",
         search="_search_value",
         help="Value, always converted to/from the typed field.",
     )
@@ -64,6 +66,7 @@ class CustomInfoValue(models.Model):
     value_int = fields.Integer(string="Whole number value", index=True)
     value_float = fields.Float(string="Decimal number value", index=True)
     value_bool = fields.Boolean(string="Yes/No value", index=True)
+    value_date = fields.Date(string="Date value", index=True)
     value_id = fields.Many2one(
         comodel_name="custom.info.option", string="Selection value",
         ondelete="cascade", domain="[('property_ids', '=', property_id)]",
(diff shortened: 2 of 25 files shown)


jarroyomorales:
@dalonsod I saw you are migrating this module to 13.0. You may be are interested in checking this!

dalonsod:
@jarroyomorales of course I'm interested, even for https://github.com/OCA/maintenance/pull/89
Without testing it yet, looks great!

dalonsod (changes requested):
@etobella Only functional review for this very very interesting improvement :-)

I noticed that required property fields are not highlighted and, in fact, it's possible to unset them. Maybe this behavior is actually inherited and relies to an issue of this addon. Could you check?

etobella:
Sure!!! It was an error on JS. I forgot to handle the modifiers. now it should be solved.

imagen

dalonsod:
> Sure!!! It was an error on JS. I forgot to handle the modifiers. now it should be solved.
> 
> imagen

I've set a property of an existing demo template as required (the date of "gamers" template). Then, the field is now highlighted but it's still possible to leave it empty, or even unsetting it. But, tested the current addon without this specific view, I think required attribute is also ignored...

etobella:
Don't worry, I will check :wink:

etobella:
@dalonsod I have added now the checks. It should work as expected :smiley:

dalonsod:
I've done the following test:
- Set every property of a template as required.
- Go to a contact, edit and select that template.
- Save without filling any property => the proper error message (OK).
- Edit again and only fill the first property and Save => a different error is displayed and some "false" labels appeared on form:
imagen
Could you reproduce it? :(

etobella:
Thanks!! I found the issue. Also I found that sometimes a double click is needed in order to perform some changes. I will check.

etobella:
The issue was solved with the last change. I will check the navigation move

etobella:
Navigation is now improved and works with the current form

dalonsod:
> Navigation is now improved and works with the current form

The last forced push was 3 days ago, maybe you've forgotten uploading the last changes you've made :sweat_smile:

etobella:
Sure, sorry :wink:

dalonsod:
Tested on runbot, LGTM!
Reviewed code, README changes are still missing

base_custom_info/__manifest__.py
```diff
@@ -7,11 +7,12 @@
     'name': "Base Custom Info",
     'summary': "Add custom field in models",
     'category': 'Tools',
-    'version': '12.0.1.0.2',
+    'version': '12.0.1.0.3',
```
yajo: IMHO this fits better as a major release instead. Includes database migrations.

base_custom_info/static/src/js/custom_info_renderer.js
```diff
@@ -0,0 +1,157 @@
...
+    var qweb = core.qweb;
+
+    var CustomInfoRenderer = BasicRenderer.extend({
+        init: function (parent, state, params) {
+            if (params === undefined) {
+                params = {};
+            }
+            if (params.viewType === undefined ) {
+                params.viewType = 'custom_info';
+            }
```
yajo:
```suggestion
            params = _.defaults({}, params, {
                viewType: "custom_info",
            });
```

base_custom_info/static/src/js/custom_info_renderer.js
```diff
@@ -0,0 +1,157 @@
...
+            ) {
+                this.mode = 'edit';
+            }
+            this.recordWidgets = [];
+        },
+        _itemType2Object: function (type) {
+            switch (type) {
+            default:
+                return type;
+            }
```
yajo: :thinking: :question:

base_custom_info/static/src/scss/custom_info.scss
```diff
@@ -0,0 +1,3 @@
+.o_form_view .o_field_widget .o_field_widget, .o_form_view .btn .o_field_widget .custom_info_value {
+        margin-bottom: 5px;
```
yajo: Is this really used or required? Can you replace it with some class on the widget such as `mb-5`?

base_custom_info/models/custom_info_value.py
```diff
@@ -125,18 +128,6 @@ def _compute_value(self):
             else:
                 s.value = getattr(s, s.field_name, False)
 
-    @api.multi
-    def _inverse_value(self):
```
yajo, line 129: I think this one is a very valuable feature. Why do you think it's not worth it anymore? :thinking:

base_custom_info/static/src/js/custom_info_renderer.js
etobella: This function is no longer required, It was a first approach and I forgot to remove it

base_custom_info/models/custom_info_value.py
etobella: Because we are using widgets with the specified field. In order to make it work properly we cannot use a Char field on a Date widget. We could try to keep the function, but I don't know if that will work properly.

base_custom_info/static/src/scss/custom_info.scss
etobella: I was unable due to:
https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/scss/form_view.scss#L394
I think now the code is clearer.

base_custom_info/static/src/js/custom_info_renderer.js
etobella: Thanks, fixed

yajo (approved):

jarroyomorales (approved):
Tested on runbot and LGTM!!!

dalonsod:
> Reviewed code, README changes are still missing

Don't forget re-generating README files :blush:

dalonsod (approved):
👍

LoisRForgeFlow (approved):
Great UX improvement :+1:

pedrobaeza:
Good work! FYI, I have seen sparse fields, and it can do something similar, but with a serious overhead:

- Create all the possible ir.model.fields for the existing template lines
- Have an overriden fields_view_get with that fields with conditional visibility.

So it's not practical. Better this approach.

PR #1848, opened 2020-06-16
[DEL] Make obsolete base_concurrency module uninstallable

thomaspaulb:


thomaspaulb:
@NL66278 What was the rationale behind this?

NL66278:
@thomaspaulb As far as I understand there were problems in the past if a user was accessing Odoo from different tabs in concurrent updates of the last login time in the res.users model. This should now be handled in the base code by updating through a separate cursor. That is already in Odoo 8.0.

NL66278 (approved):
:+1: LGTM

pedrobaeza:
For v8, this is still needed, as it requires a change in the DB layout. For v9, Odoo already includes it in core.

thomaspaulb:
@pedrobaeza Which database change is that?

NL66278:
@thomaspaulb Do we really need to disable installation? Maybe enough just not to install the module if not needed?

thomaspaulb:
Alright.

PR #1851, opened 2020-06-18
[FIX] Fixes dbfilter_from_header if dbfilter_org returns no database

cvinh:
The original dbfilter returns [] when there is no database matching the original request
This PR fixes that case
cc @luceeasypme @kbouchek

### DIFF
--- a/dbfilter_from_header/override.py
+++ b/dbfilter_from_header/override.py
@@ -12,11 +12,13 @@ db_filter_org = http.db_filter
 
 
 def db_filter(dbs, httprequest=None):
-    dbs = db_filter_org(dbs, httprequest)
+    dbs_orig = db_filter_org(dbs, httprequest)
     httprequest = httprequest or http.request.httprequest
     db_filter_hdr = httprequest.environ.get('HTTP_X_ODOO_DBFILTER')
     if db_filter_hdr:
         dbs = [db for db in dbs if re.match(db_filter_hdr, db)]
+    else:
+        dbs = dbs_orig
     return dbs
 
 


cvinh:
I wonder why we need that fix to make it work in our case. It looks that some people has been using this module for years...
In sumup, we have one server, multidatabase, multicompany, multidomain and multiwebsite...

> 2 companies have their own website with different domain but share the same database.
> dbfilter is setup to %d$

@hbrunn could you please have a look ?
Best regards

hbrunn:
you're changing the semantics of the addon and introduce a possible security issue if the dbfilter is used this way. Your code allows to inject arbitrary dbfilters which take precedence over the filter from the config file. Depending on the setup, this can be quite a problem.

You should fix your issue by setting all filters in your proxy and close this PR.

It might however be helpful to amend the readme clarifying that the header filter works on the result of the config file filter, and doesn't replace it.

cvinh:
@hbrunn thanks for your reply...
Therefore I think we have an issue somewhere... here are steps to reproduce :

- A lot of databases on the server but only 1 database 'foo' is concerned by this case
- odoo settings : proxy=true dbfilter=%d$
- multicompany with 2 companies named 'foo' and 'bar'
- multiwebsite with 2 websites belonging to each company www.foo.com and www.bar.com
- dbfilter_from_header installed
- reverse proxy with apache on the front

Apache configuration :

www.bar.com :
> RequestHeader set X-Odoo-dbfilter foo

www.foo.com :
> Nothing special because original dbfilter works

Actual behavior :
www.foo.com goes to foo's website
www.bar.com goes to error 'database not found'

After debugging, it appears that
> db_filter_org(dbs, httprequest)

returns nothing
Here is the code :
https://github.com/OCA/server-tools/blob/cca5dffd0dec68e1aa4209a8747d4381a97efa89/dbfilter_from_header/override.py#L15

Could you give advice please ?

hbrunn:
if you just have one database, you don't need any dbfilter

cvinh:
Sorry I did not precise that I do have a lot of databases on the server but only one is used for foo and bar companies
I updated steps to reproduce above

hbrunn:
then you should use the configfile dbfilter to restrict to databases that should be accessible it the first place (something like `%d|foo` comes to mind), and pass filters in all your proxy configurations to restrict this further to the one database to be served

cvinh:
Thanks, I will try it but I'm sure you gave me the solution

PR #1855, opened 2020-06-30
[10.0][UPD] base_jsonify: export display_name by default for m2o, m2m, o2m

Cedric-Pigeon:


Cedric-Pigeon:
@lmignon

base_jsonify/models/models.py
```diff
@@ -73,6 +73,14 @@ def jsonify(self, parser):
                     value = rec[field_name]
                     if value is False and field_type != 'boolean':
                         value = None
+                    elif field_type in ('many2one', 'reference'):
```
lmignon, line 76: @Cedric-Pigeon Can you add a test to cover these UC?

PR #1861, opened 2020-07-13
[WIP][12.0][ADD] add iap_alternative_provider

sebastienbeau:
Abstract module that help to create alternative provider for IAP apps

An example of implementation is available here https://github.com/OCA/connector-telephony/pull/217 for sending sms

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ env:
   - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="database_cleanup" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="module_analysis" MAKEPOT="1"
 
-
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
--- a/iap_alternative_provider/models/iap_account.py
+++ b/iap_alternative_provider/models/iap_account.py
@@ -1,0 +1,42 @@
+# Copyright 2020 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class IapAccount(models.Model):
+    _inherit = ["iap.account", "server.env.mixin"]
+    _name = "iap.account"
+
+    name = fields.Char()
+    provider = fields.Selection([("odoo", "Odoo IAP")], required=True, default="odoo")
+
+    @property
+    def _server_env_fields(self):
+        return {
+            "provider": {},
+            "account_token": {},
+        }
+
+    def _get_service_from_provider(self):
+        """In case that the provider only propose one service you can
+        return the service_name in you module to simplify the user interface"""
+        return None
+
+    def _set_service_from_provider(self):
+        for record in self:
+            service = record._get_service_from_provider()
+            if service and record.service_name != service:
+                record.service_name = service
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        record = super().create(vals_list)
+        record._set_service_from_provider()
+        return record
+
+    def write(self, vals):
+        super().write(vals)
+        self._set_service_from_provider()
+        return True
--- a/iap_alternative_provider/__init__.py
+++ b/iap_alternative_provider/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/iap_alternative_provider/__manifest__.py
+++ b/iap_alternative_provider/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2020 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+{
+    "name": "IAP Alternative Provider",
+    "summary": "Base module for providing alternative provider for iap apps",
+    "version": "12.0.1.0.0",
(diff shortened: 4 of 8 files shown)


base_iap_alternative/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "version": "12.0.1.0.0",
+    "category": "Tools",
+    "website": "http://github.com/OCA/server-tools",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["sebastienbeau"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {"python": [], "bin": []},
+    "depends": ["iap", "server_environment"],
```
pedrobaeza: You mustn't depend on `server_environment`, as a lot of us don't use it as deployment method and this is forcing a deployment style.

pedrobaeza (changes requested):
I prefer a name without `base` in this case, like `iap_alternative_provider`.

base_iap_alternative/__manifest__.py
sebastienbeau: Thanks for you review Pedro but on this point I do not agree (let's explain my point of view)

In V12 @guewen did a huge work to make server_environment less invasive. 
Before V12 adding server_environment was forcing you to use files with password/config for the deployment. Like you I really dislike this ;) as it force to use this file and make the deployment more complex for a lot of integrator. We all have our own recipe and we need to take care of new comer.

But since the v12 (and thanks a lot @guewen for this). Now file for server_environment are optional !
If you do not have it, nothing change for you and by default the password/config are stored in the database.

The huge main point of this change is now module can directly depend of server_environment and this avoid the creation of a lot of glue module.
In my case I will need a "base glue module" then a glue module for all provider implementation (so we just duplicate the number of module by 2 ! each module will only have 5 lines of code...)

Moreover we also succeed to converge for the encryption of password in database per environment with server_environment_data_encryption https://github.com/OCA/server-env/tree/12.0/server_environment_data_encryption.
So now when a module is compatible with server_environment the enduser is able to configure the login/password for the different environment and then the database is duplicated into a different env you use the right configuration.

In my opinion storing password in clear in a database is a bad practice as it always generate issue when restoring a backup for testing (a duplicated database for testing interact with a service in production).
OCA should encourage good practice, server_environment is a really good module and we should encourage anyone to use it by default. Them as we all have our own recipe we can implement our own way (C2C and Acsone use file, on our side we use encrypted config per env), in any case if you do anything server_environment will just store it in clear like native behavior.

Interested to have your point of view (and also have other Pov like @guewen )

pedrobaeza: As said, this is your deployment way. We don't need this module for anything, and I don't store password in DBs in clear for any deployment I do, and I don't want to depend on this module for everything, but I don't really plan to use as well this iap implementation for now. It was just an observation about the increase dependency of all your work in this module that I personally don't like.

guewen: Actually, we use an environment variable, stored in a file in clear text for the non-sensitive config (SERVER_ENV_CONFIG) and in an encrypted file for sensitive config (SERVER_ENV_CONFIG_SECRET). These variables are set on the containers so a restart is enough to change an option (using files need a rebuild of image).

But I agree with @sebastienbeau even if you don't use server_environment, it won't change anything for you (if no value is set in the server env inputs (files, env variables), the fields are editable as usual.

For a module that mainly stores credentials, it makes sense to depends on it, furthermore as it doesn't force you to use its method.

iap_alternative_provider/models/iap_account.py
```diff
@@ -0,0 +1,42 @@
...
+        for record in self:
+            service = record._get_service_from_provider()
+            if service and record.service_name != service:
+                record.service_name = service
+
+    @api.model
+    def create(self, vals):
+        record = super().create(vals)
+        record._set_service_from_provider()
+        return record
```
bealdav, line 37: Not sure and not blocking, but shouldn't use odoo.api.model_create_multi decorator to make sure to have better performance in all circonstances

.travis.yml
```diff
@@ -13,15 +13,15 @@ addons:
...
   global:
-  - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
+  - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0" MQT_DEP="PIP"
   - COUNTRY="US"
 
   matrix:
   - LINT_CHECK="1"
   - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="database_cleanup"
-  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="database_cleanup"
+  - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="database_cleanup" EXCLUDE="iap_alternative_provider"
```
legalsylvain: Why excluding the module ?

legalsylvain (approved):
Lgtm. Just One question. Thanks.
(Code review / no test)
Do you have other module in the roadmap (except sms module ?)

.travis.yml
sebastienbeau: fixed I had some issue why the module code analysis, but due to the change in the master and the rebase it's not needed anymore

iap_alternative_provider/models/iap_account.py
sebastienbeau: done

bealdav: It seems syntax is 
```python
@api.model_create_multi
def create(self, vals_list):
  todos = super(IrActionsTodo, self).create(vals_list)
  ...
```
No need api.returns

sebastienbeau: done

sebastienbeau:
Hy Sylvain for now only sms module for now ;)

bealdav (approved):

PR #1862, opened 2020-07-14
[12.0][ADD] Slack addont to send messages

victoralmau:
The module contains the development to implement Odoo functionalities and integrate them with Slack


len-foss (commented):
That looks promising. 

However IMO this would be more appropriated for https://github.com/OCA/social/
Furthermore, I think it would benefit from being split; you should have a `slack` addon that purely adds the function and infra to send messages, and `slack_invoices` that adds the functionality you put on invoices. Right now it's a very particular set of features that does not feel generic enough.

You could then put some helpers in the base module to reduce the code needed to create a message. Right now, when you look at the code in invoice, it's absolutely dreadful how long it is to create such a simple message. Granted, it's mostly Slack's poor API design.
But having a `slack_button(text, url)` would greatly reduce that. Same with fields, etc.

Finally, in the base addon you could add a `slack.mixin` model. 
This way, besides having the slack helpers handily available, you could add generic methods `format_to_slack` and `post_to_slack`, with very basic default implementations that can be overriden in inheriting models.

(one last thing: Slack API is really harsh with rate-limiting, so there should be better error management to retry failed messages (I don't remember the slackclient library taking care of this, but I may be wrong). And slackclient should be in the external depends of the manifest)

victoralmau:
> That looks promising.
> 
> However IMO this would be more appropriated for https://github.com/OCA/social/
> Furthermore, I think it would benefit from being split; you should have a `slack` addon that purely adds the function and infra to send messages, and `slack_invoices` that adds the functionality you put on invoices. Right now it's a very particular set of features that does not feel generic enough.
> 
> You could then put some helpers in the base module to reduce the code needed to create a message. Right now, when you look at the code in invoice, it's absolutely dreadful how long it is to create such a simple message. Granted, it's mostly Slack's poor API design.
> But having a `slack_button(text, url)` would greatly reduce that. Same with fields, etc.
> 
> Finally, in the base addon you could add a `slack.mixin` model.
> This way, besides having the slack helpers handily available, you could add generic methods `format_to_slack` and `post_to_slack`, with very basic default implementations that can be overriden in inheriting models.
> 
> (one last thing: Slack API is really harsh with rate-limiting, so there should be better error management to retry failed messages (I don't remember the slackclient library taking care of this, but I may be wrong). And slackclient should be in the external depends of the manifest)

Thanks a lot for your comment. Now i don't use this addon because created it in my previous job in other company.
Although probably this addon is interesting to OCA (in social repository) i don't plan to work on it at the moment, therefore there are different options in this regard:
- Continue with this PR and all the improvements you have mentioned.
- Another user who finds it useful can contribute to this PR and complete it.
- Close this PR, create an issue in social repository and comment on the functionality it should have and this PR (to build on it).

PR #1863, opened 2020-07-14
[12.0][IMP] attachment_queue: small improvements

clementmbr:
- Add the State fiels in a statusbar
- Rename `date_done` and replace it in the attachment's "History" group
- Add the _Failure Emails_ field to the form view (but still invisible if empty)
- Update Readme
- Rename `state_message` into `error_message`
- Rename **Set to done** button into **Cancel** button
- Improve workflow addind a **Reset to Pending** button if operation failed or cancelled to allow running it one more time.
- Add chatter to follow the operation history

cc @sebastienbeau @florian-dacosta


attachment_queue/models/attachment_queue.py
```diff
@@ -107,5 +104,10 @@ def set_done(self):
         """
         Manually set to done
         """
-        message = "Manually set to done by %s" % self.env.user.name
-        self.write({"state_message": message, "state": "done"})
```
florian-dacosta, line 111: Any reason the `state_message` is not filled anymore ?
I think if is really important to have the information that it was manually set to done and not really processed.

florian-dacosta (commented):
One remark, LGTM otherwise

attachment_queue/models/attachment_queue.py
clementmbr: I thought it would be enough to fill the `done_uid` field with the user's name in case the attachement_queue is manually set to done (whereas it will be filled by the bot's name when done by cron).

But I can add the explicit word "manually" next to the `done_date`and `done_uid` fields when set to done manually if you think it is necessary.

My main purpose here was to avoid writing this "done manually" info in the `state_message` field as it is used for Errors messages otherwise.

florian-dacosta: @clementmbr 
Yes, it is right that it is a pity to erase the error message.
But, a user can also click on the button run, in order to process the attachment.
Of from attachment_synchronize, I think there is a button on the task to run the export of the pending attachment. So, it won't always be odoo bot that will process the attachment.

If a user process the attachment, then we have no way to know if it really processed it, or just set to done with no further action.

I prefer to erase the potential error message to indicate it has manually been set to done, so we know for sure.
I think it is important to know for sure because else, we could wonder why nothing happened even if the attachment queue is done...
What do you think?

florian-dacosta (approved):

attachment_queue/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Attachment Queue",
-    "version": "12.0.1.0.0",
+    "version": "12.0.3.0.0",
```
bealdav, line 6: There is a conflict here, please do not specify a version here

bealdav (approved):
LGTM except version

attachment_queue/__manifest__.py
clementmbr: Corrected with a rebase, hope it's allright now

dreispt:
Travis is red.

PR #1869, opened 2020-07-20
[12.0][ADD] Add new module: attachment_queue_email

clementmbr:
New module based on [attachement_queue](https://github.com/OCA/server-tools/tree/12.0/attachment_queue) allowing to create `attachment.queue` objects with the attachments from emails coming from a specific `fetchmail.server` and following some conditions (new `fetchmail.attachment.condition` objects).

cc @sebastienbeau @bealdav @florian-dacosta


rvalyi:
Hello. Not sure about the implications, but in the Odoo SA proprietary cloud, you can receive emails directly (email push) instead of fetching them from a mail server. If a model is needed, the same was done in an open source manner in the latest Rails framework with the Mailbox gem https://guides.rubyonrails.org/action_mailbox_basics.html

I understand that in the OCA the current way is via fetchmail, but may be this is a thing to keep in mind when designing modules that triggers actions when receiving emails. As far as I can see, `mail.thread` and `message_new` should work exactly the same. The only thing I can see that would be very specific to fetchmail is the `fetchmail_server_id` m2o field. I let you eval if something more generic can be done for free or if it's okay this way.

rvalyi:
So far I didn't spot any "bad design" but eventually this really worth looking at this Rails new mailbox gem. It's really been taken seriously as it is the core part of the "Hey" new Basecamp mailhosting product (where they say they want to compete with GMail being more transparent) https://github.com/rails/rails/tree/master/actionmailbox

specially, they present the gem with "The inbound emails are turned into InboundEmail records using **Active Record** (Odoo ORM) and feature **lifecycle tracking**, **storage of the original email on cloud storage via Active Storage**, and responsible data handling with on-by-default incineration. These inbound emails are routed asynchronously using Active Job to one or several dedicated mailboxes, which are capable of interacting directly with the rest of your domain model."

Not too far from mail.thread and from what we do here... 
Key parts are https://github.com/rails/rails/tree/master/actionmailbox/lib/action_mailbox
and https://github.com/rails/rails/tree/master/actionmailbox/app/models/action_mailbox
https://github.com/rails/rails/tree/master/actionmailbox/app/jobs/action_mailbox
and again detailed explanations here https://guides.rubyonrails.org/action_mailbox_basics.html
Now they don't have such a filter condition in their framework and so I guess they leave it to the callback overriders.

florian-dacosta (approved):

thomaspaulb:
@clementmbr Could you rebase so that tests are run?

PR #1872, opened 2020-07-24
[REF] company_country: Deprecate module

luisg123v:
This module was intended to avoid a wrong chart of account to be installed
by the account module's hook, but that has been fixed in Odoo v13: https://github.com/odoo/odoo/pull/52117

Because of that, a note was added on the README to mention this module is
not required anymore and it shouldn't be migrated to v14.

### DIFF
--- a/company_country/README.rst
+++ b/company_country/README.rst
@@ -25,6 +25,9 @@ Company Country
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
+**Note:** This module is not required anymore for Odoo >= 13.0, thanks to
+`this fix introduced in Odoo <https://github.com/odoo/odoo/pull/52117>`_.
+
 This module allows to set a country to the main company before the ``account``
 module is installed, so the hook of that module installs the correct
 ``l10n_***`` module.
--- a/company_country/__manifest__.py
+++ b/company_country/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Company Country",
     "summary": "Set country to main company",
-    "version": "13.0.1.0.2",
+    "version": "13.0.1.0.2",  # Note: module shouldn't be migrated, see README
     "category": "base",
     "website": "https://github.com/OCA/server-tools/tree/13.0/company_country",
     "maintainers": ["moylop260", "luisg123v"],
--- a/company_country/readme/DESCRIPTION.rst
+++ b/company_country/readme/DESCRIPTION.rst
@@ -1,3 +1,6 @@
+**Note:** This module is not required anymore for Odoo >= 13.0, thanks to
+`this fix introduced in Odoo <https://github.com/odoo/odoo/pull/52117>`_.
+
 This module allows to set a country to the main company before the ``account``
 module is installed, so the hook of that module installs the correct
 ``l10n_***`` module.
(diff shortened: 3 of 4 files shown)


legalsylvain (changes requested):
I dont think that removing the module is a possible option. It can create instablr production.
Better to update the roadmap section and write that the module is obsolete since a given revision of odoo and should not be ported to v14.
What do you think ?

luisg123v:
@legalsylvain in my understanding, when a module is not required anymore, it's removed, e.g. https://github.com/OCA/web/pull/955. Moreover, this module doesn't add any relevant data/model (actually it has effect only when being installed) so it may be safely uninstalled from production instances if required.

What do you think @moylop260?

astirpe (changes requested):
The fix https://github.com/odoo/odoo/pull/52117 was merged the 28th of may. All the Odoo installations that were not updated after that date could still need this module.
It is better to write in the README (with "highlighted, bold, capital letters, etc...") that this module is not needed anymore with a recently updated Odoo. Also it should be written that this module must be NOT ported to version 14.0 and later.

luisg123v:
@astirpe,

I'd like to add that this module makes sense only during installation. It means, it doesn't matter if Odoo installations were updated before or after the 28th of may:
- If they were installed after that date, they don't need this module
- If they were installed before that date and this module was used, then its purpose was served and it's also not required anymore
- If they were installed before that date and this module was not used, installing/using this module has no effect, even if the installation hasn't been updated, because a CoA was already installed

astirpe:
In a multicompany environment you can create new companies and install their new CoA

astirpe:
Anyway at the end the final decision should be up to you @Vauxoo authors :smiley:

luisg123v:
Yes, but:
- companies are not relevant for this module. If the `account` module is installed, [it has no effect](https://github.com/OCA/server-tools/blob/d8e1f049c2ca7092f3e9d0533b47a49695b04f3b/company_country/models/res_config.py#L21)
- This only works for the main company
- This code runs only when the module is installed
- It makes no sense for already-running instances, because it's intended to automatically set the country to the main company during installation. If you have an already running instance, you may (and should) modify the country yourself.

moylop260:
@luisg123v 

@astirpe has a good point!

Luis
What about if you install a module and after it is deleted (without uninstall it)?
Could you check it, please?

luisg123v:
@moylop260 @astirpe @legalsylvain 

I took into account your comments, and only left a note on the README and a small comment on the manifest, mentioning that this module is not required anymore and it shouldn't be ported to v14.

On the other hand, I tested the following on an instance with this module already installed:
- Removing the code
- Marking the module as uninstallable

Both with identical results,, the following error message is printed to the log:  
`Some modules have inconsistent states, some dependencies may be missing: ['company_country'] `

I also tried the module uninstalling itself, but that's not possible

I have to admit I would've preferred removing the module and then leave afected instances to complain and fix them, but I agree that approach is kind of drastic.

Since it's now a small change on the README, I'm merging.

/ocabot merge

astirpe (approved):

NL66278:
@luisg123v Maybe not remove the module then, but more or less empty it out, leaving the manifest and readme? In this way existing installations will not be affected. What do you think?

pedrobaeza:
Yes, better to deactivate the need of such environment variable.

PR #1873, opened 2020-07-25
[13.0][ADD] base_name_search_multi_lang

kittiu:
This module extends the name search feature to use other translated language.

For example, if a product name in English is "Chair", and Thai is "เก้าอี้"

Given user language preference is English, when user type name to search
on product field, "Chair", the product with this name will be found.
But if user type in Thai, "เก้าอี้", no result will be shown.

With this module installed, and model product.product is set to use "Search Translated Name".
Search by Thai name, "เก้าอี้", now find the product "Chair".

Please also note that, this search feature is available only on Many2one field.


kittiu:
Test script added, ready for review.
@ps-tubtim @Saran440 
cc @dreispt

ps-tubtim (commented):
Code LGTM. Functional Test OK.

Please check the copyright of all files.

kittiu:
> Code LGTM. Functional Test OK.
> 
> Please check the copyright of all files.

Done

Saran440 (approved):
Functional Test :+1: 

Travis is red. Could you fix it?

ps-tubtim (approved):

base_name_search_multi_lang/models/ir_model.py
```diff
@@ -0,0 +1,62 @@
...
+                    self, name=name, args=args, operator=operator, limit=limit
+                )
+                # For model with name_search_multi_lang, extend result
+                multi_lang = _get_model_name_search_multi_lang(self, self._name)
+                if multi_lang:
+                    context_lang = self._context.get("lang")
+                    installed_langs = self.env["res.lang"].get_installed()
+                    langs = [x[0] for x in installed_langs if x[0] != context_lang]
+                    for lang in langs:
+                        res = _extend_name_search_lang(
```
StefanRijnhart, line 50: Looks like this will exceed the limit easily, as well as the default model search order. Would it help to finish with a  regular id search on collected ids with the given limit, and then return name_get on the result of that?

kittiu: You mean limit of the results, i.e., 8 records?

StefanRijnhart: Yes, from the limit argument of name_search. You apply the limit in every search, but then you keep extending the result set. And extending the result set that way does not keep the ordering as defined on the model.

kittiu:
Thank you I will check. Thanks a lot for your suggestion.

base_name_search_multi_lang/models/ir_model.py
kittiu: I have made a bunch of sample data. It surprised me that, even with .extend, it does show more than 7 rows. None the less, I re-browse again this should ensure the order.

StefanRijnhart (approved):

newtratip (approved):

PR #1875, opened 2020-08-03
[10.0][FIX] Update requirements.txt

igallart:
pyotp in python2.7 is supported in versions <2.4

PR #1876, opened 2020-08-04
Pin pyotp to last version compatible with py 2.7

bealdav:
see https://github.com/pyauth/pyotp/pull/89/files

Without this fix branch fail

because of https://github.com/pyauth/pyotp/blob/master/src/pyotp/__init__.py#L14 using mypy

Also concern v10, v9

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@
   python-json-logger
   odoorpc
   raven
-  pyotp
+  pyotp<2.4


bealdav:
cc @grindtildeath
@hbrunn @StefanRijnhart @pedrobaeza

hbrunn (approved):

andhit-r (approved):

bealdav:
see https://github.com/OCA/server-tools/pull/1875

bealdav:
Now it's approved, probably it's better to merge ~ quickly because server-tools is used in many repos and it'll fail most of the new commits accross oca repos

pedrobaeza:
Well, 8.0 is not the most active OCA version, but let's go with it:

/ocabot merge nobump

PR #1877, opened 2020-08-04
[9.0][FIX] Update requirements.txt

igallart:
pyotp in python2.7 is supported in versions <2.4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ acme_tiny
 IPy
 validate_email
 pysftp
-pyotp
+pyotp<2.4
 fs==0.5.4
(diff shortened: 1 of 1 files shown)


igallart:
View https://github.com/OCA/server-tools/pull/1875

igallart:
Is there a problem with commit?

pedrobaeza:
Yes, Travis is red.

igallart:
@pedrobaeza , I think that this change in requirements is necessary because the higher versions do not work and will not start Odoo, it has been made in version 8 and 10. The fails in Travis I doubt very much had to do with this change.

PR #1878, opened 2020-08-07
[10.0][ADD] Allow login based on IP

StefanRijnhart:
This module allows you to restrict who is allowed to login to Odoo based on
the remote IP address. The restriction is applied upon login, and on RPC
calls. Note: valid web sessions are not impacted, and access to the public
website can not be restricted using this module.

When a login is rejected because of a missing IP access rule, a warning is
logged in the application log but no specific information is provided to the
user that is trying to log on. Instead, a generic access denied error is
generated (which is rendered in the web client as 'Wrong login/password'.

Installation
============

Please make the python library py2-ipaddress available in the environment in which you run Odoo before installing this module.

Configuration
=============

Go to menu *Settings -> Technical -> Security -> IP Access Rules* to create
access rules for the group whose access you want to limit. If no rules exist
for the groups of a user, it is not restricted to login from anywhere.

Each rule allows you to allow access for a specific group, or user, from IP
address or an IP network (using a netmask, e.g. 192.168.0.1/24).

Rules can be configured for a group, or for a user but not for both at the
same time. If neither group or user is configured, the rule is applied
globally.

A special case is where you allow logins from *any* private network (e.g.
networks in the 192.168.x.x or the 10.x.x.x ranges). You can create such a
rule by ticking the *private* checkbox on the rule.

If you accidentally lock yourself out you can regain access by accessing
your Odoo database through SQL and execute the following command.

```
   UPDATE ip_access_rule SET active = FALSE;
```

You will need to restart Odoo to clear the IP access cache. This will lift all IP access restrictions.

Note: if you run Odoo behind a proxy (and you should, because where do you get your
SSL encryption from?), you need to set *proxy_mode = True* in the Odoo
configuration file to ensure that the remote address is properly propaged to
Odoo and is not set to the IP address of the proxy server. A setup like this
is dependent on the proxy server setting the correct remote address in the
*X-Forwarded-For* header.


StefanRijnhart:
The change in auth_admin_passkey is needed because it creates a user with the cursor that runs the test transaction. If auth_ip_access is loaded, it browses that user id with a cursor fetched from the database pool, and there the new user does not exist. By entering test mode, the cursor that is returned from the database pool is the 'shared' test cursor (note that by default this test cursor is not the cursor used in the test class' transaction so that one has to be reassigned this 'shared' test cursor manually).

auth_ip_access/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,30 @@
+Go to menu *Settings -> Technical -> Security -> IP Access Rules* to create
+access rules for the group whose access you want to limit. If no rules exist
+for the groups of a user, it is not restricted to login from anywhere.
+
+Each rule allows you to allow access for a specific group, or user, from IP
+address or an IP network (using a netmask, e.g. 192.168.0.1/24).
```
janverb, line 6: This example doesn't work, I'm guessing it should be `192.168.1.0/24`

auth_ip_access/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Rules can be configured for a group, or for a user but not for both at the
+same time. If neither group or user is configured, the rule is applied
+globally.
+
+A special case is where you allow logins from *any* private network (e.g.
+networks in the 192.168.x.x or the 10.x.x.x ranges). You can create such a
+rule by ticking the *private* checkbox on the rule.
+
```
janverb, line 15: I can imagine wanting to deny access by default, so users can't log in at all if there's no rule for them, for extra safety. Maybe you could suggest creating a rule for address `0.0.0.0` for that use case. Or offer a pre-defined inactive rule for it, though that might be confusing since it wouldn't show up by default.

auth_ip_access/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "author": "Opener B.V., Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "preloadable": True,
+    # Note: install package as py2-ipaddress
+    "external_dependencies": {"python": ['ipaddress'], },
+    "depends": [
+        # auth_crypt overwrites methods we override
+        "auth_crypt",
```
janverb, line 20: I guess that if authentication goes through some other addon's implementation of `check_credentials` the one here may or may not take effect, depending on how the MRO ended up. That's unfortunate, but I can't think of any non-gruesome way to solve it.

Should there be a warning in the documentation that the rules work unreliably with other authentication methods? You wouldn't want people to rely on the rules taking effect, or even to rely on the rules not taking effect.

janverb (approved):
The code looks great and my test went well. I only have a few quibbles about the documentation.

PR #1874, opened 2020-07-29
[12.0][MIG] letsencrypt

hbrunn:
v12 migration of #1840

PR #1883, opened 2020-08-11
[12.0] [MIG] letsencrypt

janverb:
Migration of #1840, replaces #1874


hbrunn (approved):

janverb:
I can't quite figure out why it's [failing](https://travis-ci.com/github/OCA/server-tools/jobs/460397819). Looks like it's trying to install Python 2 versions of dependencies from `requirements.txt` and then it gives up before it gets to `dnspython`, which is necessary to use the module?

NL66278:
Lots of modules that have nothing to do with this one can't find their addon depends requirements through PIP.

And why do we need six travis jobs to all do the same tests?

NL66278 (approved):
:LGTM and working in practice.

dreispt:
Travis is red

NL66278:
@dreispt Is travis not red because of failures in other modules, related I guess to switching runbot to load stuff through pip?

letsencrypt/__manifest__.py
```diff
@@ -0,0 +1,34 @@
...
+    "demo": [
+        "demo/ir_cron.xml",
+    ],
+    "post_init_hook": "post_init_hook",
+    "installable": True,
+    "external_dependencies": {
+        "python": [
+            "acme",
+            "cryptography",
+            "dns",
```
dreispt: I believe Travis is red because of this pip dependency: [no "dns" package was found](https://app.travis-ci.com/github/OCA/server-tools/jobs/522015141#L564)

janverb: That package is imported as `dns` but installed as [`dnspython`](https://pypi.org/project/dnspython/). Is something trying to install the module names in the manifest as if they're package names?

dreispt:
Yes, Python  external dependencies must be the package name, and this is verified now.

--dr

blyxxyz:
Should I put `dnspython` instead? I thought Odoo tried to import the modules in the manifest to verify they're installed, and importing `dnspython` would fail.

(I no longer work with Odoo so I don't currently have an installation to to check with.)

dreispt:
Yes, change it please.
It used that way, but it was changed, since the package name is more 
helpful there, when ensuring dependencies are installed.

dreispt:
Build is failing: 

    odoo.exceptions.UserError: ('Unable to install module "letsencrypt" because an external dependency is not met: No module named dnspython', '')

NL66278:
@dreispt It is succesfully running in production, and 5 of the 7 travis jobs are succesfull. I do not know why there are seven travis jobs and I do not see any difference in configuration, some of this travis stuff is still a mystery to me.

thomaspaulb:
Superseded by https://github.com/OCA/server-tools/pull/2395

PR #1888, opened 2020-08-25
[11.0] base_exception improvements

gurneyalex:
Cherry pick  https://github.com/OCA/server-tools/pull/1644 fixing https://github.com/OCA/server-tools/pull/1642 

Fix exceptions which would be copied when a record is duplicated. 

Related to https://github.com/OCA/sale-workflow/pull/1218

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -142,14 +142,40 @@ class BaseException(models.AbstractModel):
     @api.multi
     def _check_exception(self):
         """
-        This method must be used in a constraint that must be created in the
-        object that inherits for base.exception.
-        for sale :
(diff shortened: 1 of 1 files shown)


base_exception/models/base_exception.py
```diff
@@ -110,7 +110,7 @@ class BaseException(models.AbstractModel):
     )
     exception_ids = fields.Many2many(
         'exception.rule',
-        string='Exceptions')
+        string='Exceptions', copy=False)
```
i-vyshnevska, line 113: question: this shouldn't have any effect m2m is not copied by default

leemannd (approved):
Functional testing only

PR #1890, opened 2020-08-27
[10.0][FIX] Authadmin passkey doesn't validate param before eval

llacroix:
Without being validated, if get_param return False, it will try to
pass False to safe_eval that will raise an exception trying to
evaluate False which isn't a string.

As a result, a user typing to password of an admin while auth passkey
params aren't configured will see an internal server error... Which
isn't as "safe" as just showing bad credentials.

### DIFF
--- a/auth_admin_passkey/models/res_users.py
+++ b/auth_admin_passkey/models/res_users.py
@@ -6,7 +6,6 @@
 import datetime
 
 from odoo import SUPERUSER_ID, _, api, exceptions, models
-from odoo.tools.safe_eval import safe_eval
 
 
 class ResUsers(models.Model):
@@ -22,17 +21,14 @@ class ResUsers(models.Model):
         admin_user = self.sudo().browse(SUPERUSER_ID)
         login_user = self.browse(user_id)
 
-        send_to_admin = icp_obj.get_param('auth_admin_passkey.send_to_admin')
-        if send_to_admin is False:
-            send_to_admin = True
-        else:
-            send_to_admin = safe_eval(send_to_admin)
+        def read_param(key):
+            record = icp_obj.search([("key", "=", key)], limit=1)
+            if not record:
+                return True
+            return record.value == "True"
 
-        send_to_user = icp_obj.get_param('auth_admin_passkey.send_to_user')
-        if send_to_user is False:
-            send_to_user = True
-        else:
-            send_to_user = safe_eval(send_to_user)
+        send_to_admin = read_param('auth_admin_passkey.send_to_admin')
+        send_to_user = read_param('auth_admin_passkey.send_to_user')
 
         mails = []
         if send_to_admin and admin_user.email:
--- a/auth_admin_passkey/models/res_users.py
+++ b/auth_admin_passkey/models/res_users.py
@@ -22,12 +22,17 @@ class ResUsers(models.Model):
         admin_user = self.sudo().browse(SUPERUSER_ID)
         login_user = self.browse(user_id)
 
-        send_to_admin = safe_eval(
-            icp_obj.get_param('auth_admin_passkey.send_to_admin')
-        )
-        send_to_user = safe_eval(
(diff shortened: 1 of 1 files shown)


auth_admin_passkey/models/res_users.py
```diff
@@ -22,12 +22,17 @@ def _send_email_passkey(self, user_id):
...
+        if send_to_admin is False:
+            send_to_admin = True
+        else:
+            send_to_admin = safe_eval(send_to_admin)
+
+        send_to_user = icp_obj.get_param('auth_admin_passkey.send_to_user')
+        if send_to_user is False:
+            send_to_user = True
+        else:
+            send_to_user = safe_eval(send_to_user)
```
janverb: I don't think this is right. We get `False` after [`base.config.settings` clears out the value](https://github.com/OCA/server-tools/blob/841afbd7ece0e4c1f3e30b6941c8da64d925ba4a/auth_admin_passkey/models/res_config.py#L26), and then we do want it to be false, not true.

This should've been changed in ad4d562eac2e2a4739c507b116a54021ee6203d7, I think. We don't have to eval anything any more. This should be enough:

```suggestion
        send_to_admin = icp_obj.get_param('auth_admin_passkey.send_to_admin')
        send_to_user = icp_obj.get_param('auth_admin_passkey.send_to_user')
```

If the settings are enabled they are `"True"`, which is truthy. If they're disabled they're `False`, which is falsey. So it works out as it should.

llacroix: Just to explain, in case the value isn't defined, it sets it to True as it's a sane default to be notified if you didn't configure anything.

While the config might set it to '', on one of our server it was indeed False because the key got removed from the ir.config_parameter or was simply absent and never created.

> This should've been changed in ad4d562, I think. We don't have to eval anything any more. This should be enough:

Yeah... thought for example in my case, the key wasn't present at all on one of our server for some reasons... And since it used to be safe_evaling, and using repr... You could see a server with `'False'` which is.. truthy... 

https://github.com/OCA/server-tools/commit/ad4d562eac2e2a4739c507b116a54021ee6203d7#diff-4b54a2cab970fc9b3616e51585b23946L49

But hey, I'm fine with getting rid of safe_eval altogether. This simplifies a lot and is less prone to errors. like the one we got a few days ago.

janverb: The thing is that if you set it to `''`, `get_param` returns `False` ([or the default](https://github.com/oca/ocb/blob/671c63424f3/odoo/addons/base/ir/ir_config_parameter.py#L62)). It doesn't distinguish between a missing parameter and an empty parameter.

If there's no parameter at all then it must have been explicitly deleted, because [it's created when you install the module](https://github.com/oca/server-tools/blob/e16b37ae/auth_admin_passkey/data/ir_config_parameter.xml#L5). I'm not sure if `True` is the right default then. Maybe it is, because then the setting doesn't change if you upgrade the module and it adds the record back. But you'd need to manually search for an `ir.config_parameter` record instead of using `get_param`.

I guess you'd want `'False'` to stay false for backward compatibility, yeah.

Maybe this?
```suggestion
        def read_param(key):
            record = icp_obj.search([("key", "=", key)], limit=1)
            if not record:
                return True
            return record.value == "True"

        send_to_admin = read_param('auth_admin_passkey.send_to_admin')
        send_to_user = read_param('auth_admin_passkey.send_to_user')
```
If there's no record or a `'True'` record, it's `True`. If there's a `''` or `'False'` record, it's `False`.

llacroix:
Sounds good to me

janverb (approved):
[`flake8` still complains about an unused import](https://travis-ci.com/github/OCA/server-tools/jobs/379489891#L426) but I've tested this and it works.

alexis-via (approved):
I was impacted by the bug too. This PR fixes the bug.

PR #1892, opened 2020-09-04
[FIX] audit_log: Permission error on auditlog.log.line missing a requ…

max3903:
…ired field (field_id)

In some wird cases, Odoo tries to create an auditlog.log record with `line_ids: [{[(6, 0, [])]}, {(0, 0, [])}]`. This pull request seems to fix the issue.


huguesdk (commented):
@max3903 could you please provide some more information on how to reproduce the problem? what are the weird cases you’re referring to, and what is the exact error message?

PR #1893, opened 2020-09-04
[12.0][ADD] base_partition: add partition method to base model

len-foss:
This module adds a "partition" method to recordsets to obtain a {value: records} structure.
Useful to for instance split work by lang or similar.
E.g.:
```
customers_by_lang = customers.partition("lang_id")
for lang, lang_customers in customers_by_lang.items():
    work.with_context(lang=lang.code).process(lang_customers)
```
or an arbitrary function:
```
customers_by_number_of_orders = customers.partition(lambda c: len(c.orders))
for number_orders in customers_by_number_of_orders:
    process(customers_by_number_of_orders[number_orders])
```

The sketchiest part of the function is that technically values can have any type; 
in particular not be hashable (nor stringifiable in the worst case).
The most common case is to use a "field(.subfield)+" ending with a nonrelational field. 
In that case the result of a "mapped" is a list, which is not hashable. 
The current implementation in that case use the `str` representation of the value as key, 
and does not handle the case where the value does not even support str (which should be vey rare...).
These two points are open to discussion.

Also maybe it could go in an existing module instead?


base_partition/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Acsone (http://www.acsone.eu)
+# Nans Lefebvre <nans.lefebvre@acsone.eu>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base Partition",
+    "summary": "Base module that provide the partition method on all models",
+    "version": "12.0.0.0.0",
```
rousseldenis: three last digits represent major/minor/patch levels. So, 12.0.1.0.0. Thanks

rousseldenis (approved):
LGTM !

pedrobaeza:
Nans, glad to see you around again. And now in the OCA side!

len-foss:
Thank you Pedro, and glad to be back (on the best side)! 
I'm still discovering the many OCA repositories so please be patient with me.

pedrobaeza:
Yeah, too much to learn also in this side. Don't hesitate to ask, and congratulations for your new job position. A good company!

rousseldenis:
@pedrobaeza Could you review it ?

pedrobaeza (commented):
Some commits should be squashed before merging, and the license change to LGPL is on the board.

rousseldenis:
> Some commits should be squashed before merging, and the license change to LGPL is on the board.

@pedrobaeza FYI, there is https://github.com/OCA/server-tools/pull/2615

thomaspaulb:
@len-foss Would you still be available to:

- Rebase
- Squash commits
- Change license to LGPL

len-foss:
@thomaspaulb I would but I have left the Acsone org so I can't push on this branch. So @rousseldenis could do it, but I can't.

rousseldenis:
@len-foss @thomaspaulb I will do it as I use it as soon as I can :smiley:

PR #1894, opened 2020-09-04
[IMP] base_jsonify: add lang, custom resolvers, alias* support

len-foss:
This PR introduces the following features:
- translate the export of fields (e.g. description_fr, description_en, description_nl)
- allow to add arguments to a list by ending the alias by * (e.g. all fields with alias customValues* will end up in customValues: [...])
- pass each field through a custom function, configurable from the interface (resolvers)
- pass the obtained json through a global resolver

A resolver is simply a record with a python code field and an eval method;
it comes in two variant, one to work at the field level and the second at the global level.
The idea is to extend the export in a way that new requirements could be met without needing to change any code.

The intended use-case is to be compatible with APIs that require all translated parameters to be exported simultaneously,
and ask for custom properties to be put in a sub-dictionary.
Since these requirements are fairly standard, we extend the module in a way that could fit most use-cases.

Note that the intermediate data structure "json parser" is insufficient for our requirements as it follows the BNF:
```
    Ps = [ps*]
    ps = Fs | (Fs, C) |(Fs, Ps)
    Fs = f | f:a
    C = method | callable
        f in field_names for the fields of the target model
        a = E* any alias, i.e. E is the alphabet 
```
We extend it to the following:
```
    Pf = {"langs": {L: P}*, "resolver"?: R}
    P = [p*]
    p = F | (F, P)
    F = {"name": f, "alias"?: a, "resolver"?: R, "callable"?: C}
    R = resolver
    L = lang
```
To put it simply, we keep the same structure be replace the simple string representation of fields by dictionaries,
and have an outer dictionary so that we can have a parser for each lang, and a global resolver.
To keep things simple we keep compatibility with that simpler API and convert the simple parser to a full parser.

### DIFF
--- a/base_jsonify/README.rst
+++ b/base_jsonify/README.rst
@@ -96,11 +96,11 @@ To use these features, a full parser follows the following structure:
             False: [
                 {'name': 'description'},
                 {'name': 'number', 'resolver': 5},
-                ({'name': 'partner_id', 'alias': 'partner'}, [{'name': 'display_name'}])
+                ({'name': 'partner_id', 'target': 'partner'}, [{'name': 'display_name'}])
             ],
             'fr_FR': [
-                {'name': 'description', 'alias': 'descriptions_fr'},
-                ({'name': 'partner_id', 'alias': 'partner'}, [{'name': 'description', 'alias': 'description_fr'}])
+                {'name': 'description', 'target': 'descriptions_fr'},
+                ({'name': 'partner_id', 'target': 'partner'}, [{'name': 'description', 'target': 'description_fr'}])
             ],
         }
     }
@@ -132,7 +132,7 @@ but other features like custom resolvers are:
         "fields": [
                 {'name': 'description'},
                 {'name': 'number', 'resolver': 5},
-                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}]),
+                ({'name': 'partner_id', 'target': 'partners'}, [{'name': 'display_name'}]),
         ],
     }
 
@@ -157,7 +157,7 @@ If the global resolver is given, then the json_dict goes through:
 Which allows to add external data from the context or transform the dictionary
 if necessary. Similarly if given for a field the resolver evaluates the result.
 
-It is possible for an alias to end with a '*':
+It is possible for a target to end with a '*':
 in that case the result is put into a list.
 
 .. code-block:: python
@@ -165,8 +165,8 @@ in that case the result is put into a list.
   parser = {
       fields: [
           {'name': 'name'},
-          {'name': 'field_1', 'alias': 'customTags*'},
-          {'name': 'field_2', 'alias': 'customTags*'},
+          {'name': 'field_1', 'target': 'customTags*'},
+          {'name': 'field_2', 'target': 'customTags*'},
       ]
   }
 
--- a/base_jsonify/README.rst
+++ b/base_jsonify/README.rst
@@ -96,13 +96,28 @@ To use these features, a full parser follows the following structure:
             False: [
                 {'name': 'description'},
                 {'name': 'number', 'resolver': 5},
-                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}])
+                ({'name': 'partner_id', 'alias': 'partner'}, [{'name': 'display_name'}])
             ],
             'fr_FR': [
                 {'name': 'description', 'alias': 'descriptions_fr'},
-                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'description', 'alias': 'description_fr'}])
+                ({'name': 'partner_id', 'alias': 'partner'}, [{'name': 'description', 'alias': 'description_fr'}])
             ],
         }
+    }
+
+
+One would get the a result having this structure (note that the translated fields are merged in the same dictionary):
+
+.. code-block:: python
+
+    exported_json == {
+        "description": "English description",
+        "description_fr": "French description, voilà",
+        "number": 42,
+        "partner": {
+            "display_name": "partner name",
+            "description_fr": "French description of that partner",
+        },
     }
 
 
@@ -114,7 +129,6 @@ but other features like custom resolvers are:
 
     parser = {
         "resolver": 3,
-        "language_agnostic": True,
         "fields": [
                 {'name': 'description'},
                 {'name': 'number', 'resolver': 5},
@@ -122,6 +136,15 @@ but other features like custom resolvers are:
         ],
     }
 
+
+By passing the `fields` key instead of `langs`, we have essentially the same behaviour as simple parsers,
+with the added benefit of being able to use resolvers.
+
+Standard use-cases of resolvers are:
+- give field-specific defaults (e.g. `""` instead of `None`)
+- cast a field type (e.g. `int()`)
+- alias a particular field for a specific export
+- ...
 
 A simple parser is simply translated into a full parser at export.
 
@@ -135,18 +158,16 @@ Which allows to add external data from the context or transform the dictionary
 if necessary. Similarly if given for a field the resolver evaluates the result.
 
 It is possible for an alias to end with a '*':
-in that case the result it put into a list.
+in that case the result is put into a list.
 
 .. code-block:: python
 
   parser = {
-      "langs": {
-          False: [
-              {'name': 'name'},
-              {'name': 'field_1', 'alias': 'customTags*'},
-              {'name': 'field_2', 'alias': 'customTags*'},
-          ]
-      }
+      fields: [
+          {'name': 'name'},
+          {'name': 'field_1', 'alias': 'customTags*'},
+          {'name': 'field_2', 'alias': 'customTags*'},
+      ]
   }
 
 
--- a/base_jsonify/README.rst
+++ b/base_jsonify/README.rst
@@ -29,7 +29,7 @@ This module adds a 'jsonify' method to every model of the ORM.
 It works on the current recordset and requires a single argument 'parser'
 that specify the field to extract.
 
-Example of parser:
+Example of a simple parser:
 
 
 .. code-block:: python
@@ -42,8 +42,8 @@ Example of parser:
         ('line_id', ['id', ('product_id', ['name']), 'price_unit'])
     ]
 
-In order to be consitent with the odoo api the jsonify method always
-return a list of object even if there is only one element in input
+In order to be consistent with the odoo api the jsonify method always
+return a list of object even if there is only one element in the recordset.
 
 By default the key into the json is the name of the field extracted
 from the model. If you need to specify an alternate name to use as key, you
@@ -79,6 +79,77 @@ you can pass a callable or the name of a method on the model:
 Also the module provide a method "get_json_parser" on the ir.exports object
 that generate a parser from an ir.exports configuration.
 
+Further features are available for advanced uses.
+It defines a simple "resolver" model that has a "python_code" field and an eval
+function so that arbitrary functions can be configured to transform fields,
+or process the resulting dictionary.
+It is also to specify a lang to extract the translation of any given field.
+
+To use these features, a full parser follows the following structure:
+
+.. code-block:: python
+
+    parser = {
+        "resolver": ir.exports.resolver(3),
+        "langs": {
+            False: [
+                {'name': 'description'},
+                {'name': 'number', 'resolver': ir.exports.resolver(5)},
+                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}])
+            ],
+            'fr_FR': [
+                {'name': 'description', 'alias': 'descriptions_fr'},
+                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'description', 'alias': 'description_fr'}])
+            ],
+        }
+    }
+
+
+A simple parser is simply translated into a full parser at export.
+
+If the global resolver is given, then the json_dict goes through:
+
+.. code-block:: python
+
+    resolver.eval(dict, record)
+
+Which allows to add external data from the context or transform the dictionary
+if necessary. Similarly if given for a field the resolver evaluates the result.
+
+It is possible for an alias to end with a '*':
+in that case the result it put into a list.
+
+.. code-block:: python
+
+  parser = {
+      "langs": {
+          False: [
+              {'name': 'name'},
+              {'name': 'field_1', 'alias': 'customTags*'},
+              {'name': 'field_2', 'alias': 'customTags*'},
+          ]
+      }
+  }
+
+
+Would result in the following json structure:
+
+.. code-block:: python
+
+    {
+        'name': 'record_name',
+        'customTags': ['field_1_value', 'field_2_value'],
+    }
+
+The intended use-case is to be compatible with APIs that require all translated
+parameters to be exported simultaneously, and ask for custom properties to be
+put in a sub-dictionary.
+Since it is often the case that some of these requirements are optional,
+new requirements could be met without needing to add field or change any code.
+
+Note that the export values with the simple parser depends on the record's lang;
+this is in contrast with full parsers which are designed to be language agnostic.
+
 **Table of contents**
 
 .. contents::
@@ -108,6 +179,7 @@ Contributors
 * BEAU Sébastien <sebastien.beau@akretion.com>
 * Raphaël Reverdy <raphael.reverdy@akretion.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
+* Nans Lefebvre <nans.lefebvre@acsone.eu>
 
 Maintainers
 ~~~~~~~~~~~
--- a/base_jsonify/README.rst
+++ b/base_jsonify/README.rst
@@ -90,11 +90,12 @@ To use these features, a full parser follows the following structure:
 .. code-block:: python
 
     parser = {
-        "resolver": ir.exports.resolver(3),
+        "resolver": 3,
+        "language_agnostic": True,
         "langs": {
             False: [
                 {'name': 'description'},
-                {'name': 'number', 'resolver': ir.exports.resolver(5)},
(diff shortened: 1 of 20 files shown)


len-foss:
@simahawk 
I started implementing this in 12.0, before seeing https://github.com/OCA/server-tools/pull/1829
As a result my own design is considerably different; in particular the main feature I wanted to keep is that everything would be available from the ir.exports object, and without requiring code change and redeploys, which is not possible with callables.

But the changes are fairly consequent, so...

len-foss:
@lmignon @sebastienbeau 
Could I have your opinion on this?

Note that in the code there was a:
`# TODO: we should get default by field (eg: char field -> "")`
With the current approach, it would be possible to add to any line a `default` parameter which would be a function taking a field name/field type as argument and returns a default value.
In the general case, if key is supposed to be of type `str`, having `{key: None}`, `{key: ""}`, and `{}` (not having key in the json) are three separate cases which may not work interchangeably or even raise an exception (some APIs require to not have the key in the json, which was the main motivation for the global resolver). 
So it's useful to be able to be able to specify the default on a field-basis, and the "full parser" API makes this easy to add and use.

base_jsonify/README.rst
```diff
@@ -79,6 +79,95 @@ you can pass a callable or the name of a method on the model:
 Also the module provide a method "get_json_parser" on the ir.exports object
 that generate a parser from an ir.exports configuration.
 
+Further features are available for advanced uses.
+It defines a simple "resolver" model that has a "python_code" field and an eval
+function so that arbitrary functions can be configured to transform fields,
+or process the resulting dictionary.
+It is also to specify a lang to extract the translation of any given field.
```
simahawk, line 86: I don't get why you need this generate translations: you should call `jsonify` with the proper lang in the ctx :thinking:

base_jsonify/README.rst
```diff
@@ -79,6 +79,95 @@ you can pass a callable or the name of a method on the model:
...
+
+.. code-block:: python
+
+    resolver.eval(dict, record)
+
+Which allows to add external data from the context or transform the dictionary
+if necessary. Similarly if given for a field the resolver evaluates the result.
+
+It is possible for an alias to end with a '*':
+in that case the result it put into a list.
```
simahawk:
```suggestion
in that case the result is put into a list.
```

base_jsonify/models/ir_export.py
```diff
@@ -4,10 +4,20 @@
 
 from collections import OrderedDict
 
-from odoo import models
+from odoo import fields, models
 
 
-def update_dict(data, fields):
+def partition(l, accessor):  # -> Dict
```
simahawk: docstring

base_jsonify/README.rst
```diff
@@ -79,6 +79,95 @@ you can pass a callable or the name of a method on the model:
...
+.. code-block:: python
+
+    parser = {
+        "resolver": 3,
+        "language_agnostic": True,
+        "langs": {
+            False: [
+                {'name': 'description'},
+                {'name': 'number', 'resolver': 5},
+                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}])
```
simahawk: I'd take the change to rename `alias` to `target`, which is more explicit IMO.

base_jsonify/models/ir_exports_line.py
```diff
@@ -13,6 +13,29 @@ class IrExportsLine(models.Model):
...
+        comodel_name="res.lang",
+        string="Language",
+        help="If set, the language in which the field is exported",
+    )
+    resolver_id = fields.Many2one(
+        comodel_name="ir.exports.resolver",
+        string="Custom resolver",
+        help="If set, will apply the resolver on the field value",
+    )
+    function = fields.Char(
```
simahawk: so this would replace the callable feature in the parser?
As far as I see below, passing the callable in the simple parse is still allowed, right?

base_jsonify/models/ir_exports_resolver.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    _name = "ir.exports.resolver"
+    _description = "Resolver"
+
+    name = fields.Char()
+    type = fields.Selection([("field", "Field"), ("global", "Global")])
+    python_code = fields.Text(
+        string="Python Code",
+        default="\n# ".join(["result = value"] + help_message),
```
simahawk: `result = value` shouldn't stay at the end?

base_jsonify/models/models.py
```diff
@@ -17,19 +17,103 @@ class Base(models.AbstractModel):
...
+                key = json_key[:-1]
+                if not root.get(key):
+                    root[key] = []
+                root[key].append(value)
+            else:
+                root[json_key] = value
+        return root
 
     def jsonify(self, parser):
         """Convert the record according to the given parser.
```
simahawk, line 94: please report in the docstring how the output looks like.

simahawk (changes requested):

base_jsonify/README.rst
lmignon: @simahawk As I understand it, to have different lang at same time into the result of jsonify

base_jsonify/README.rst
```diff
@@ -79,6 +79,95 @@ you can pass a callable or the name of a method on the model:
...
+            ],
+            'fr_FR': [
+                {'name': 'description', 'alias': 'descriptions_fr'},
+                ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'description', 'alias': 'description_fr'}])
+            ],
+        }
+    }
+
```
lmignon, line 107: I would be nice to show the result generated by this kind of parser

base_jsonify/README.rst
len-foss: We are working with an API that wants all translations at the same time; so if you want to pass the description_sale, you need also to pass description_sale_fr_BE, description_sale_nl_BE, ...

In a way that could be avoided by externalising the lang, so instead you would have one export per lang and then merge the jsons; but the json merge is also fairly complex, it's less explicit, and overall seems to be like a worse solution.

base_jsonify/models/ir_exports_line.py
len-foss: Yes the existing callable feature is still fully supported (the corresponding test has not been modified). 
This "function" field is added so that your callable feature can also be used through the interface. Although you can only pass a method name, not a lambda. 
(I called it "function" instead of callable to avoid linter warnings and writing `if callable(callable)` stuff)

base_jsonify/models/ir_exports_resolver.py
lmignon: I agress with @simahawk

base_jsonify/README.rst
len-foss: Thank you for the detailed review, I'll fix all these soon.

lmignon (changes requested):
Some little comments.... my main concern is that if becomes very difficult to know how we can use this addon into our code.

base_jsonify/README.rst
simahawk: Ok, seems legit. Is it worth to have a parser per each lang? I mean, we can rely on one parser and simply pass the list of langs code you want in the `langs` key.

len-foss: Yeah, this point was in fact driven by the implementation...
As per Laurent recommendation, records.with_context(lang=...) can be very costly. 
So putting it this way makes it very easy to only do it the minimum number of times.
To have the more natural way to just put the lang in the parser's dict, the beginning of jsonify would need to extract the translated fields to keep the same efficiency.

simahawk: Ok, but I'm saying this structure is not required IMO.

The question is: how many times you want to have a different result schema by lang? Never, likely :)

Hence, you could change this

```py
parser = {
    "resolver": 3,
    "language_agnostic": True,
    "langs": {
        False: [
            {'name': 'description'},
            {'name': 'number', 'resolver': 5},
            ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}])
        ],
        'fr_FR': [
            {'name': 'description', 'alias': 'descriptions_fr'},
            ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'description', 'alias': 'description_fr'}])
        ],
    }
}
```
to
```py
parser = {
    "resolver": 3,
    "language_agnostic": False,  # this key can be dropped or made computed on `langs`
    "parser|schema|fields":  [
            {'name': 'description'},
            {'name': 'number', 'resolver': 5},
            ({'name': 'partner_id', 'alias': 'partners'}, [{'name': 'display_name'}])
        ],
    "langs": ["fr_FR", "it_IT"],  # maybe m2m to res.lang in the resolver record
}
```
and then you generate the translated key for each field by appending - by default - the lang code to the key.
You can pass `translate_key_func` to override the behavior (eg: `lang.split("_")[0]`)

The rest of the implementation stays the same, just do not lookup for a resolver for every lang.
Am I missing anything?

len-foss: Oh, I see what you mean.
Well, typically that API we are working with is the annoying kind: any custom fields must be put into extraData* as a {name, value, type} dictionary.
But it's not that rare, I've worked with 3 platforms that allowed custom fields/columns and they all had similar (but different!) requirements.

simahawk: > 3 platforms that allowed custom fields/columns and they all had similar (but different!) requirements.

you mean per lang? Each lang w/ different keys? (eg: FR has "name" and "description" but IT has only "name"?)

len-foss: Sorry, I meant here that for English description the alias should be `Description`, but for other langs the alias should be `extraData*` (and then in a dict with `'name':  description_sale_langcode`). So:

```
{
    "Description": "English description",
    "extraData": [{"name": "description_sale_fr_BE", "value": "Description française", "type": "str"}],
}
```

(and so I meant that some API have a similar concept of extraData, but that would be slightly differently handled)
(also the field to dict transformation is handled by a generic field resolver)

With your proposed solution (which is simpler), this would require to afterwards modify the dict structure to find all translated fields and put them into the extraData* list. It's a bit less explicit and a tad tricky to do (do you go through all _fields that are translatable, or do you pass the list of fields that should go into extraData another way?).

len-foss:
Most issues should have been addressed in last update, if you can have another look at it.
I think the takeaway it that the changeset almost can be dropped with almost no change to existing codebases (except for alias -> target renaming of course, but it's in a separate commit). 
The only thing that would require changes is line https://github.com/OCA/search-engine/blob/12.0/connector_search_engine/components/mapper.py#L23, but it's because this module expose its internals, in a way.

If there are further improvements that can be done to the documentation, let me know.

lmignon (approved):
LGTM. THank you @len-foss ! (Code review only)

base_jsonify/README.rst
```diff
@@ -79,6 +79,116 @@ you can pass a callable or the name of a method on the model:
...
+If the global resolver is given, then the json_dict goes through:
+
+.. code-block:: python
+
+    resolver.eval(dict, record)
+
+Which allows to add external data from the context or transform the dictionary
+if necessary. Similarly if given for a field the resolver evaluates the result.
+
+It is possible for a target to end with a '*':
```
simahawk: can we make this more explicit?
See my example here https://github.com/OCA/website-cms/blob/11.0/cms_form/marshallers.py 
In that case if an input comes from the request and the name has a marshaller in it the value gets converted.
In your case I'd replace `*` with `:list`. This would open the door to support other marshallers in an explicit way as I did.

base_jsonify/migrations/13.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,6 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    query = """ALTER TABLE "ir_exports_line" RENAME COLUMN "alias" TO "target";"""
```
simahawk: FTR you could use `odoo.tools.sql.rename_column`

base_jsonify/models/ir_export.py
```diff
@@ -4,10 +4,26 @@
 
 from collections import OrderedDict
 
-from odoo import models
+from odoo import fields, models
 
 
-def update_dict(data, fields):
+def partition(l, accessor):
+    """Partition an iterable (e.g. a recordset) according to an accessor (e.g. a lambda)
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

base_jsonify/models/ir_export.py
```diff
@@ -23,41 +39,76 @@ def update_dict(data, fields):
...
-            parser.append(field)
+        if value[0] is True:  # is a leaf
+            parser.append(field_dict(field, value[1]))
         else:
-            parser.append((field, convert_dict(value)))
+            parser.append((field_dict(field), convert_dict(value[1])))
     return parser
 
 
+def field_dict(field, options=None):
```
simahawk, line 64: docstring

base_jsonify/models/ir_export.py
```diff
@@ -23,41 +39,76 @@ def update_dict(data, fields):
...
+    )
+
+    global_resolver_id = fields.Many2one(
+        comodel_name="ir.exports.resolver",
+        string="Custom global resolver",
+        domain="[('type', '=', 'global')]",
+        help="If set, will apply the global resolver to the result",
+    )
+
     def get_json_parser(self):
```
simahawk, line 97:
```suggestion
    @tools.ormcache("self.language_agnostic", "self.global_resolver_id.id", "self.export_fields.ids")
    def get_json_parser(self):
```
to be adapted in case there are other keys to consider

base_jsonify/models/ir_exports_line.py
simahawk: I'd rename this to `instance_method_name` to make it clear.

base_jsonify/models/ir_exports_resolver.py
```diff
@@ -0,0 +1,54 @@
...
+    ":param record: the record",
+]
+
+
+class FieldResolver(models.Model):
+    """Arbitrary function to process a field or a dict at export time.
+    """
+
+    _name = "ir.exports.resolver"
+    _description = "Resolver"
```
simahawk:
```suggestion
    _description = "Export field Resolver"
```

base_jsonify/models/ir_exports_resolver.py
```diff
@@ -0,0 +1,54 @@
...
+
+    name = fields.Char()
+    type = fields.Selection([("field", "Field"), ("global", "Global")])
+    python_code = fields.Text(
+        string="Python Code",
+        default="\n".join(["# " + h for h in help_message] + ["result = value"]),
+        help="\n".join(help_message),
+    )
+
+    def eval(self, param, records):
```
simahawk: I'd find a better name as `eval` is built-in.

base_jsonify/models/models.py
```diff
@@ -17,19 +17,103 @@ class Base(models.AbstractModel):
...
-        if isinstance(parser_field, tuple):
-            field_name, subparser = parser_field
-        json_key = field_name
-        if ":" in field_name:
-            field_name, json_key = field_name.split(":")
-        return field_name, json_key, subparser
+        return parser_field if isinstance(parser_field, tuple) else (parser_field, None)
+
+    @api.model
+    def convert_simple_to_full_parser(self, parser):
```
simahawk: this could be moved to `utils.py`

simahawk (changes requested):

base_jsonify/models/ir_export.py
len-foss: `@ormcache("self.language_agnostic", "self.global_resolver_id.id", "tuple(self.export_fields.ids)")`
 
This is the only comment that I have not implemented.
Because it depends on `self.export_fields.ids`, if you actually modify one line, the returned parser is not updated (the One2many itself hasn't changed).

You can check for yourself, as the second test in test_getting_parser breaks as soon as you put the decorator.

len-foss (commented):
Hello @simahawk ,
Sorry about the long delay, but that gave me more time to work with the module and have more use-cases. 
In hindsight, I still think it's a fairly good design, as it solves:
- default value management. There was a TODO about the default. Depending on the API for a string field you might need to give either "" or None as the field's default. That would be a bit unwieldy, but is easily solved by a field resolvers.
- It really does avoid to create new computed fields just for export purpose, so it can make you save quite some time.
- I want to have geojson support for base_geoengine, since ElasticSearch supports geo_shape objects. See https://github.com/OCA/geospatial/pull/281

base_jsonify/models/ir_export.py
simahawk: > Because it depends on self.export_fields.ids, if you actually modify one line, the returned parser is not updated (the One2many itself hasn't changed).

what about defining a custom cache key using `self.export_fields.mapped("write_date")`?

simahawk:
> It really does avoid to create new computed fields just for export purpose, so it can make you save quite some time.

this is the thing that I like the most probably :)

Amazing work, thanks @len-foss :pray: 

One last question [here](https://github.com/OCA/server-tools/pull/1894#discussion_r561022249).
Regarding merge: have you tested this in real life w/ shopinvader and/or shopfloor? If not, it would be nice to give it a try in real life before merging, if possible.

len-foss:
Thank you !
I check the ormcache a bit later.
This has been in production for a while on an instance with a shopinvader shop, although most of the features are not used for shopinvader itself (they are used for the ChannelEngine module, which I should clean up soon for OCA submission).
For the geo_point export, we added that to an elastic index that is being displayed on a map view in shopinvader with taking advantage of geo queries. It has been put in test but sadly not in production yet.

len-foss:
ormcache with write date works fine :ok_hand:

simahawk:
> ormcache with write date works fine ok_hand

cool, thanks!

simahawk (approved):

PR #1900, opened 2020-09-11
[IMP][12.0] jsonifyfull

len-foss:


thomaspaulb:
@len-foss Is this still waiting for the 13.0 version? (and what does that mean exactly)

len-foss:
@thomaspaulb At the time I expected to have some feedback before merging on the latest version. Since then I think this has been merged as the default version of the module, so everything should be good.

PR #1905, opened 2020-09-18
12.0 FIX: attachment_synchro: remove file_type from attachment.queue

bealdav:
now guessed from attachment.synchronize.task

cc @florian-dacosta @sebastienbeau @clementmbr


bealdav:
Finally @florian-dacosta convince me this file is usefull

PR #1906, opened 2020-09-24
base_jsonify: return one record on demand

simahawk:
If you use this module extensively - as I do -
you've done this tons of time. For sure :)
This little change simplifies your code base where needed.

CC @lmignon @len-foss

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -26,7 +26,7 @@ class Base(models.AbstractModel):
             field_name, json_key = field_name.split(":")
         return field_name, json_key, subparser
 
-    def jsonify(self, parser):
+    def jsonify(self, parser, one=False):
         """Convert the record according to the given parser.
 
         Example of parser:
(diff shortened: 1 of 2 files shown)


base_jsonify/models/models.py
```diff
@@ -63,6 +64,8 @@ def jsonify(self, parser):
                 else:
                     res[json_key] = rec._jsonify_value(field_name)
             result.append(res)
+        if one:
```
len-foss, line 70: Wouldn't you add a `self.ensure_one()` here? 
Otherwise a call with multiple records and a `one=True` seems very dubious?
(and I agree, I almost added a jsonify_one() method)

simahawk: good catch, done, tnx!

lmignon (approved):
@simahawk sometimes it's so obvious that one wonders why it wasn't done earlier.

simahawk:
> @simahawk sometimes it's so obvious that one wonders why it wasn't done earlier.

You don't know how many times I was tempted to do it :wink:

simahawk:
ops, closed by error :roll_eyes:

len-foss (approved):
No need to close it, Laurent's comment wasn't so harsh :->

PR #1907, opened 2020-09-25
[13.0] base_jsonify: fix rounding for float fields

i-vyshnevska:
use precision defined on fields for data preparing 
before data "13.95" returned as "13.950000000000001"

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -39,6 +39,8 @@ class Base(models.AbstractModel):
         """Override this function to support new field types."""
         if value is False and field.type != "boolean":
             value = None
+        elif field.type == "float":
+            value = float(field.convert_to_column(value, self))
         elif field.type == "date":
             value = fields.Date.to_date(value).isoformat()
         elif field.type == "datetime":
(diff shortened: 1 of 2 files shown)


i-vyshnevska:
side effect now it returned as a string

i-vyshnevska:
@simahawk can you give an opinion on this change

base_jsonify/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Base Jsonify",
     "summary": "Base module that provide the jsonify method on all models",
-    "version": "13.0.1.3.0",
+    "version": "13.0.1.4.0",
```
simahawk: rollback this pls, we manage it via ocabot. In any case this seems more a patch than an improvement :wink:

simahawk (changes requested):
can you re-word the commit msg to include module name (as per PR title)?

simahawk:
> side effect now it returned as a string

this is a problem. It will change the behavior in maaaany places.
I'd make this conditional or convert it back to number.

i-vyshnevska:
@simahawk updated

lmignon (changes requested):
Thank you @i-vyshnevska Code LGTM but can you add a tests for this change PLZ.

jcoux:
@i-vyshnevska Fix after your rebase: https://github.com/i-vyshnevska/server-tools/pull/1

ivantodorovich (approved):

ajaniszewska-dev (approved):

PR #1908, opened 2020-09-28
[IMP] base_fuzzy_search: on install automatically create required dat…

dreispt:
…abase extension


bodedra (approved):
LG :+1:

dreispt:
This feature is already there 🍅

PR #1913, opened 2020-10-14
[FIX] autovacuum_message_attachment: The field name cannot be used as it is already present in other models

etobella:
The field name used should be forbidden. With the existent code on OCA, messages are sent without attachments, as the inherits are being lost. For example, the following one:

https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_mail.py#L27

@florian-dacosta @jarroyomorales

### DIFF
--- a/autovacuum_message_attachment/models/base.py
+++ b/autovacuum_message_attachment/models/base.py
@@ -7,7 +7,7 @@ from odoo import fields, models
 class Base(models.AbstractModel):
     _inherit = "base"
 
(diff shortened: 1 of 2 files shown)


jarroyomorales (approved):
Tested locally and solves the problem. Can it be fast-tracked?

florian-dacosta (approved):

LoisRForgeFlow (approved):

PR #1915, opened 2020-10-23
[IMP][12.0] base_jsonify: optional date and datetime formatters

kevinkhao:
This PR 
- adds the possibility to add date and datetime formatters to the jsonify function
- refactors some tests


lmignon:
@kevinkhao Sorry but I don't see why we should be able to provide an other format than the ISO one. IMO that introduce lot of complexity for zero benefit.

kevinkhao:
@lmignon here is our scenario:
- need to jsonify a record to load it into a datamodel as a REST response
- record has M2Os and O2Ms. M2Os run several levels deep
- we don't want to write a recursive function to reparse the result of jsonify() and essentially write a 2nd json spec to reparse it and convert date formats

kevinkhao:
@lmignon thanks for your feedback.
In the end my use case scenario was an error on my part, so I will close this PR.

PR #1917, opened 2020-10-23
[12.0][WIP][REF] refactor code, rename enabled to active, and simplifie run code

sebastienbeau:
- [x] rename enabled to active
- [x] unify run button
- [x] improve UX

### DIFF
--- a/attachment_synchronize/models/attachment_synchronize_task.py
+++ b/attachment_synchronize/models/attachment_synchronize_task.py
@@ -89,7 +89,7 @@ class AttachmentSynchronizeTask(models.Model):
         "\nFurther operations will be realized on these Attachments Queues depending "
         "on their 'File Type' value.",
     )
-    enabled = fields.Boolean("Enabled", default=True)
+    active = fields.Boolean("Enabled", default=True, old="enabled")
     avoid_duplicated_files = fields.Boolean(
         string="Avoid importing duplicated files",
         help="If checked, a file will not be imported if an Attachment Queue with the "
@@ -139,10 +139,18 @@ class AttachmentSynchronizeTask(models.Model):
         if domain is None:
             domain = []
         domain = expression.AND(
-            [domain, [("method_type", "=", "import"), ("enabled", "=", True)]]
+            [domain, [("method_type", "=", "import")]]
         )
         for task in self.search(domain):
             task.run_import()
+
+    def run(self):
+        for record in self:
+            method = "run_{}".format(record.method_type)
+            if not hasattr(self, method):
+                raise NotImplemented
+            else:
+                getattr(record, method)()
 
     def run_import(self):
         self.ensure_one()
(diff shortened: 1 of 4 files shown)


attachment_synchronize/models/attachment_synchronize_task.py
```diff
@@ -89,7 +89,7 @@ class AttachmentSynchronizeTask(models.Model):
         "\nFurther operations will be realized on these Attachments Queues depending "
         "on their 'File Type' value.",
     )
-    enabled = fields.Boolean("Enabled", default=True)
+    active = fields.Boolean("Enabled", default=True, old="enabled")
```
GSLabIt, line 92: if task is deactivated will disappear from list, that's why used enabled instead of active. is this a wanted feature?

bealdav (commented):
Why remove enable?

sebastienbeau:
@bealdav  @GSLabIt 

The aim is to stick to odoo logic (toggle active is here natively).

Regarding the visibility it's really easy to solve it, you just need to add an "{active_test: False} in the context
Like here : https://github.com/OCA/server-tools/pull/1917/files#diff-043c362866d526d479476b871011f828b7df1faa5b2daba6b59d96088f65a925R121

So we have exactly the same UI and the same behaviour, just less code and more odoo like

GSLabIt:
> @bealdav @GSLabIt
> 
> The aim is to stick to odoo logic (toggle active is here natively).
> 
> Regarding the visibility it's really easy to solve it, you just need to add an "{active_test: False} in the context
> Like here : https://github.com/OCA/server-tools/pull/1917/files#diff-043c362866d526d479476b871011f828b7df1faa5b2daba6b59d96088f65a925R121
> 
> So we have exactly the same UI and the same behaviour, just less code and more odoo like

Yup, you're right! I've seen it after writing the comment; so i deleted it.

bealdav:
I know active test but I think it's a bad implementation because it's rely on context. Sometimes you come into a view from a secondary path without any context and then your view is broken.Real example. For me we have to go away from active test.

GSLabIt:
> I know active test but I think it's a bad implementation because it's rely on context. Sometimes you come into a view from a secondary path without any context and then your view is broken.Real example. For me we have to go away from active test.

if you call action, it will add context; if you create custom one need to remember add `active_test=False` to the context. Imho, better keep tight to odoo standards.

sebastienbeau:
I have added some action directly accessible from the tree view
And duplicate/archive button are hidden by default

image


You can still show active/duplicate button through the context. Like in my case for pattern_import_export_synchronize when I show them using a one2many

image

sebastienbeau:
@bealdav the issue with enabled is that
- you duplicate the code
- you have to add domain in several place if you when to hide the record not enable
In this case we only when to show this task in some specific place, and I prefers by default to hide disable task.

In any case I had to add the active_test only in two place (one here in the action, then one time in pattern_import_export_synchronize). I think it's cleaner, let's see what @florian-dacosta think

bealdav:
@sebastienbeau ok it you're convince of that.

florian-dacosta:
Hi @bealdav @sebastienbeau 
I also prefer to go back to active to be consistent with standard Odoo.

For me, it is not even important to pass the context active test, as I don't really see the point of seeing by default unactive records, but I don't mind if it is the case.
Anyway, even if the context is not passed for a reason or another, it is no big deal, you rarely have to deal with  inactive stuff.

florian-dacosta (approved):

bealdav (approved):
FYI, this module and its dependency needs to be hardly refactored from 13.0 because of this
https://github.com/OCA/edi/pull/243#discussion_r509216851
Don't waste too many time to fine tuning it for now. Go with that.

PR #1918, opened 2020-10-28
[IMP] override dbfilter instead of superposing it

kh-badep:
The current behavior does not correspond to the description of the module: If there is already a dbfilter defined in the config file or as a parameter, this module will add the new dbfilter to the old one hence further filtering out the databases.

This proposed changes will instead completely override the old dbfilter with the new one, giving more flexibility to its use.

### DIFF
--- a/dbfilter_from_header/override.py
+++ b/dbfilter_from_header/override.py
@@ -12,11 +12,12 @@ db_filter_org = http.db_filter
 
 
 def db_filter(dbs, httprequest=None):
-    dbs = db_filter_org(dbs, httprequest)
     httprequest = httprequest or http.request.httprequest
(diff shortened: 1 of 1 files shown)


PieterPaulussen (commented):
Semantically, this looks exactly the same.

kh-badep:
There's a difference:

- If no dbfilter is defined in the config file, then the behavior will be the same
- If there is a dbfilter in the config AND a dbfilter in the header then this new change will **disregard** the original dbfilter and replace it with the dbfilter from header, which I suppose is the intended purpose.

cvinh:
@kh-badep I did not see your PR before
I did another PR that fixes the issue here https://github.com/OCA/server-tools/pull/2353

PR #1760, opened 2020-01-20
[13.0][MIG] base_exception

jaredkipe:
Cleanup per docs including squashing "UPD" commits.
Unit tests run and pass.

PR #1025, opened 2017-10-13
11.0 mig base exception

mourad-ehm:
MIG module base_exception and add test.

PR #1920, opened 2020-10-30
[13.0][MIG] base_exception: Migration to 13.0

joao-p-marques:
Superseeds https://github.com/OCA/server-tools/pull/1760
Ran pre-commit and squashed latest commits that attended comments in previous PR.

@Tecnativa
TT21615

ping @pedrobaeza


joao-p-marques:
Also updated README

atchuthan (approved):
Code review :+1:

base_exception/tests/purchase_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class LineTest(models.Model):
+    _name = "base.exception.test.purchase.line"
+    _description = "Base Exception Test Model Line"
+
+    name = fields.Char()
+    lead_id = fields.Many2one("base.exception.test.purchase", ondelete="cascade")
+    qty = fields.Float()
+    amount = fields.Float()
```
victoralmau, line 69: It's a good idea load new model to use it later in tests?

victoralmau (approved):

base_exception/tests/purchase_test.py
joao-p-marques: Thanks for the comment @victoralmau 
Seems like a strange approach, indeed, but I have done some digging and it seems to be the best option to test this according to [this comment](https://github.com/OCA/server-tools/pull/1025#discussion_r153815296)
It already comes from previous versions...
@pedrobaeza do you think this should be refactored?

pedrobaeza: Let's keep it for now.

PR #1921, opened 2020-10-31
[12.0][FIX] excel_import_export, Change color in bold_red style

newtratip:
openpyxl after v.3.0.3 is not supported `colors.RED`, I change color to "FF0000"
https://foss.heptapod.net/openpyxl/openpyxl/-/blob/3.0.4/openpyxl/styles/colors.py#L36

cc: @kittiu

### DIFF
--- a/excel_import_export/__manifest__.py
+++ b/excel_import_export/__manifest__.py
@@ -25,6 +25,6 @@
              'views/webclient_templates.xml',
              ],
     'installable': True,
-    'development_status': 'beta',
+    'development_status': 'Beta',
     'maintainers': ['kittiu'],
 }
--- a/excel_import_export/models/styles.py
+++ b/excel_import_export/models/styles.py
@@ -7,7 +7,7 @@ import logging
 _logger = logging.getLogger(__name__)
 
 try:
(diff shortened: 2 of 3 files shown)


kittiu:
Can you add the required lib version to requirement.txt of this repo too.

newtratip:
@kittiu Do you mean `add required lib version to requirement.txt of this repo` instead of this change ?

kittiu:
Both.

newtratip:
Why we must to define openpyxl version in the requirement.txt ? I think this is fix bug for `RED` color attribute has no after v.3.0.3. After fixed it can use with all version. 

@kittiu Please suggest me if it's not correct.

kittiu:
requirement.txt is where we normally tells what lib are required. Despite
your fix, it should have had it anyway.

It think it is good practice.

kittiu:
@newtratip can you change development_status to "Beta" please.
And we will ready for merge.

newtratip:
> @newtratip can you change development_status to "Beta" please.
> And we will ready for merge.

@kittiu Edited.

newtratip:
@kittiu Done in excel_import_export_demo, Change development_status to Alpha.

PR #1925, opened 2020-11-04
[13.0][MIG] configuration_helper

phuctranfxvn:


configuration_helper/i18n/configuration_helper.pot
```diff
@@ -0,0 +1,40 @@
+# Translation of Odoo Server.
```
i-vyshnevska, line 1: you can remove this file

i-vyshnevska (approved):

simahawk (approved):

configuration_helper/i18n/configuration_helper.pot
simahawk: pot file commits can be trashed indeed

grindtildeath:
Can we maybe have a test module for that? Did anyone else test it?

I tried to test it but it seems that fields are not created properly in res.config.settings :confused:

PR #1926, opened 2020-11-10
[WIP][ADD] jsonify_stored

len-foss:
Variable/function names are not particularly good (it's a WIP after all).


jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2020 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class JsonifyStored(models.AbstractModel):
+    """Stores the data to export. When inheriting this class,
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+       _export_xmlid should be overwritten by an existing export xml_id.
+       Note that this xml_id has to exist before the model init,
+       i.e. either be in a dependency module, or created in a pre_init_hook
+       (however it's fine to modify an existing one in the module data).
+    """
+
+    _name = "jsonify.mixin"
```
simahawk:
```suggestion
    _name = "jsonify.stored.mixin"
```

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+    _description = "Stores json data"
+
+    _export_xmlid = ""
```
simahawk: I'd add a m2o which my default gets its value via `_export_xmlid`.

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+
+    data = fields.Serialized(compute="_compute_data", readonly=True, store=False)
+    data_compute = fields.Boolean(
+        compute="_compute_data_compute", default=True, readonly=True, store=True
+    )
+    data_stored = fields.Serialized(
+        compute="_compute_data_stored", readonly=True, store=True
+    )
+
```
simahawk: we should make clear the purpose of all fields and methods as this mixin can be applied to all models.
I'd prefix them all w/ `jsonify_` or `jsonify_store_`.

```suggestion
    jsonify_data = fields.Serialized(compute="_compute_jsonify_data", readonly=True, store=False)
    jsonify_data_must_compute = fields.Boolean(
        compute="jsonify_data_must_compute", default=True, readonly=True, store=True
    )
    jsonify_data_stored = fields.Serialized(
        compute="jsonify_data_stored", readonly=True, store=True
    )

```

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+    @api.model
+    def _get_export(self):
```
simahawk:
```suggestion
    def _jsonify_get_export(self):
```
aaaand repeat over the rest :)

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+        xml_id = self._export_xmlid
+        return self.env.ref(xml_id) if xml_id else self.env["ir.exports"]
+
+    @api.model
+    def get_export_depends(self):
+        export_field_names = self._get_export().export_fields.mapped("name")
+        return [fn.replace("/", ".") for fn in export_field_names]
+
+    @api.depends(lambda self: self.get_export_depends())
```
simahawk: are you sure this works at instance level? :thinking:

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+    def _compute_data_compute(self):
+        for record in self:
+            record.data_compute = True
+
+    def _compute_data_stored(self):
+        if self:
+            parser = self._get_export().get_json_parser()
+            datas = self.jsonify(parser)
+            for record, data in zip(self, datas):
+                record.data = data
```
simahawk: `data` is not countable
```suggestion
            records_data = self.jsonify(parser)
            for record, data in zip(self, records_data):
                record.data = data
```

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+       i.e. either be in a dependency module, or created in a pre_init_hook
+       (however it's fine to modify an existing one in the module data).
+    """
+
+    _name = "jsonify.mixin"
+    _description = "Stores json data"
+
+    _export_xmlid = ""
+
+    data = fields.Serialized(compute="_compute_data", readonly=True, store=False)
```
simahawk: can you clarify what's the purpose of the non stored field? I'd add an help msg

jsonify_stored/models/jsonify_mixin.py
```diff
@@ -0,0 +1,96 @@
...
+            record.data = record.data_stored
+
+    @api.model
+    def cron_recompute(self, limit=None):
+        records = self.search([("data_compute", "=", True)], limit=limit)
+        records._compute_data_stored()
+
+
+class IrExport(models.Model):
+    """Because of jsonify.mixin, some models might have field depends that depend
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

simahawk (changes requested):
tests definitely required :)

simahawk:
@len-foss would you have time to move on here?

len-foss:
Definitely, I'll get a test suite done by the end of the week.

len-foss:
@simahawk 
I think it's in a decent state now, not just a PoC anymore.

Not sure the cron should exist on the mixin - it makes sense to specialize crons for interesting inheriting models. But it also makes sense to have it as a default so that you don't need to re-implement it for basic models.
Besides I've found better to put it in a test module, this way it also provides a reference implementation for the pre-init hook.

jsonify_stored/models/ir_exports.py
```diff
@@ -0,0 +1,35 @@
...
+        return res
+
+    def _check_jsonify_stored(self):
+        # if "jsonify.stored.mixin" not in self.env:
+        #     return  # might be necessary in some weird cases?
+        xml_ids = self.get_xml_id().values()
+        jsonify_mixin = self.env["jsonify.stored.mixin"]
+        jsonified_models = [self.env[m] for m in jsonify_mixin._inherit_children]
+        models = [m for m in jsonified_models if m._export_xmlid in xml_ids]
+        for model in models:
```
simahawk, line 31: are we sure this is not over killing when you have tons of records?
Maybe a manual action would be better?

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+
+class JsonifyStored(models.AbstractModel):
+    """Stores the data to export.
+       When inheriting this class, _export_xmlid should be set with an export xml_id.
+       Note that this xml_id has to exist before the model init,
+       i.e. either be in a dependency module, or created in a pre_init_hook
+       (however it's fine to modify an existing one in the module data).
+    """
+
+    @api.depends(lambda self: self._jsonify_get_export_depends())
```
simahawk: do we really need this? You already write the flag when ir.export gets updated.

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+        if self:
+            parser = self._jsonify_get_export().get_json_parser()
+            data_list = self.jsonify(parser)
+            for record, data in zip(self, data_list):
+                record.jsonify_data_stored = data
+                record.jsonify_data_todo = False
+
+    @api.depends("jsonify_data_todo", "jsonify_data_stored")
+    def _compute_jsonify_data(self):
+        self.filtered("jsonify_data_todo")._compute_jsonify_data_stored()
```
simahawk: I don't understand why this is required. If `jsonify_data_todo` is handled correctly the field will be recomputed anyway.
Also, by having 2 fields the whole set of values will be cached twice doubling mem consumption likely.

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+        for record in self:
+            record.jsonify_data = record.jsonify_data_stored
+
+    _name = "jsonify.stored.mixin"
```
simahawk, line 21: keep these and fields definition on top. `compute` accept a string w/ the name of the compute methods.

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+        return self.env.ref(xml_id) if xml_id else self.env["ir.exports"]
+
+    @api.model
+    def _jsonify_get_export_depends(self):
+        export_field_names = self._jsonify_get_export().export_fields.mapped("name")
+        return tuple(fn.replace("/", ".") for fn in export_field_names)
+
+    @api.model
+    def cron_recompute(self, limit=None):
+        records = self.search([("jsonify_data_todo", "=", True)], limit=limit)
```
simahawk, line 134: this can be huge and make 999 compute fail if the 1st one out of 1000 fails: should chunked and use jobs.

test_jsonify_stored/hooks.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2020 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import SUPERUSER_ID, api
+
+
+def pre_init_hook(cr):
+    """The test model depends on model_export, so we need it before model init."""
+    env = api.Environment(cr, SUPERUSER_ID, dict())
+    export = env.ref("test_jsonify_stored.model_export", raise_if_not_found=False)
```
simahawk: why do you need this? The ir.export record should be created explicitly by the implementing module.

test_jsonify_stored/tests/common.py
```diff
@@ -0,0 +1,32 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.tests import SavepointCase
+
+
+class TestJsonifyMixin(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestJsonifyMixin, cls).setUpClass()
```
simahawk, line 11: do this always
```suggestion
        super(TestJsonifyMixin, cls).setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
and add `no_reset_password=True` if you create users.

simahawk (changes requested):

simahawk:
> Not sure the cron should exist on the mixin - it makes sense to specialize crons for interesting inheriting models. But it also makes sense to have it as a default so that you don't need to re-implement it for basic models.

I'd keep it on the mixin. It makes perfectly sense.

jsonify_stored/models/ir_exports.py
lmignon: Do we really have so much jsonify model instances using jsonify_mixin? IMO not really. IMO this could stay as is.

jsonify_stored/models/ir_exports.py
```diff
@@ -0,0 +1,35 @@
...
+            recompute_field = model._fields["jsonify_data_todo"]
+            recompute_field.depends = model._jsonify_get_export_depends()
+            recompute_field.setup_triggers(model)
+            model.search([]).write({"jsonify_data_todo": True})
```
lmignon:
```suggestion
            model.search([("jsonify_data_todo", "=", False)]).write({"jsonify_data_todo": True})
```

jsonify_stored/models/jsonify_stored_mixin.py
lmignon: @simahawk This is required. The depends trigger is computed from the ir.exports definition linked to the model. The flag is set into ir.exports when the definition is updated only.

jsonify_stored/models/jsonify_stored_mixin.py
lmignon: @simahawk The compute will also modify `jsonify_data_todo` from `True` to `False` Without the filtered, we'll have an infinite loop.
About the memory consumption, indeed that could become a concern... Nevertheless, by using a computed fields we keep the benefits of the batch computation mechanism from odoo (and the prefetching)

jsonify_stored/models/jsonify_stored_mixin.py
lmignon: > keep these and fields definition on top. `compute` accept a string w/ the name of the compute methods.

+1

test_jsonify_stored/tests/common.py
len-foss: Yeah I still have this on my notes from your training session. Thanks for reminding me.

test_jsonify_stored/hooks.py
len-foss: Because when the models are setup, the `@api.depends(lambda self: self._jsonify_get_export_depends())` on the implementing model would crash, as the export does not exist yet.

jsonify_stored/models/ir_exports.py
len-foss: If you modify significantly the export lines (remove one, etc), all the exports have to be recomputed, so there's no way to go around it. But it should not happen much.

simahawk: > Do we really have so much jsonify model instances using jsonify_mixin? IMO not really. IMO this could stay as is.

You might have them. Eg: in the case of search engine bindings -> possibly a ton.
Also, one of the good reasons to make this split is to make this feature reusable across models/features. 
Meaning that potentially you can use it everywhere you want to expose data w. the same structure (eg: EDI, ESB, REST)

> If you modify significantly the export lines (remove one, etc), all the exports have to be recomputed, so there's no way to go around it. But it should not happen much.

Well, you might add a new field mapping on a db w/ more than 50000 records to recompute and the update will kick your face. As mentioned above: to me it's better if this becomes an aware decision/act.

If you want to do it in this place then you should do it atomically: recompute only what has changed (eg: remove field or compute only the new field). And this reminds me that we should find a better way to check if something has changed and find a way to recompute only what has changed. This is very hard to do because you might have nested information coming from nested related records.... hence let's discuss this for the next round :wink:

test_jsonify_stored/hooks.py
simahawk: tnx for the explanation. For everyone else and your future self: mention this in a comment :wink: 

Nevertheless: can't we do this in the mixin? I mean, this is a mandatory piece. It must be there. Period.
And I can imagine a lot of headaches of ppl not adding this init hook or missing that record in many other ways.

What if `_jsonify_get_export` creates it if missing. And what if we define a computed name/xid for it?
Something like `{module}.jsonify_export_{model._table}` maybe?
It would also help in keeping ir.exports records name and xid consistent across modules.

jsonify_stored/models/ir_exports.py
simahawk: @lmignon ping :)

test_jsonify_stored/hooks.py
simahawk: @len-foss ping :)

len-foss:
@lmignon @simahawk 
So. I worked quite a bit more on it. The thing is, this PR use the compute method to update a second field (todo is actually set by either the trigger, or the recompute). It works, but it uses the ORM in a nonstandard way (and I don't see any trick to avoid that -- it could be possible to cheat using two dates, one "recompute date" and one "need update date", but then that would require a custom search method to to emulate the 'todo' state, which would be equivalent to r.recompute <= r.update. So would be a tad more complex). 
I've ported it to 13.0: https://github.com/OCA/server-tools/compare/13.0...acsone:13.0-jsonifystored-len
As shown in the third commit, it required a flush to pass the test as before, but it's the only change.

I don't know how bad of an idea it is, to be frank. So I've created an alternate version that is much simpler, at https://github.com/OCA/server-tools/compare/12.0...acsone:12.0-jsonifystoredsimple-len. 
The forward-port should be entirely trivial for this branch. The only downside is that it does not allow to batch recomputes as well. But that may be improved by improving queue_job.

Also, I don't want to miss the forest for the trees. Most of the code about the export modifications could be removed if we forbid to modify exports manually. It would mean that an export modification would require an instance restart, which is actually reasonable. In Practice, that really should not happen, especially for any big db.

simahawk:
> @lmignon @simahawk
> So. I worked quite a bit more on it. 

1st of all: thanks for your efforts! :handshake: 

> The thing is, this PR use the compute method to update a second field (todo is actually set by either the trigger, or the recompute). It works, but it uses the ORM in a nonstandard way (and I don't see any trick to avoid that -- it could be possible to cheat using two dates, one "recompute date" and one "need update date", but then that would require a custom search method to to emulate the 'todo' state, which would be equivalent to r.recompute <= r.update. So would be a tad more complex).

Another option is: build an hash using required keys and compute only when the hash changes.
In this way you won't have to recompute all data when something changes. The compute_data method will check if the hash has changed and recompute on the fly. This of course will be mitigated by a cron that will update them under the hood.

I've made something similar here to avoid computing images over and over: https://github.com/shopinvader/odoo-shopinvader/blob/13.0/shopinvader_image/models/shopinvader_image_mixin.py#L24

I think this is going to simplify a lot the machinery as you won't have to mess up w/ depends but only force hash computation when ir.exports records or resolvers change.

Also, according to https://github.com/OCA/server-tools/pull/1926#discussion_r568386034 you can simplify more and remove the commit hook as well.

> I've ported it to 13.0: [13.0...acsone:13.0-jsonifystored-len](https://github.com/OCA/server-tools/compare/13.0...acsone:13.0-jsonifystored-len)
> As shown in the third commit, it required a flush to pass the test as before, but it's the only change.

Cool, thanks!

> I don't know how bad of an idea it is, to be frank. So I've created an alternate version that is much simpler, at [12.0...acsone:12.0-jsonifystoredsimple-len](https://github.com/OCA/server-tools/compare/12.0...acsone:12.0-jsonifystoredsimple-len).

> The forward-port should be entirely trivial for this branch. The only downside is that it does not allow to batch recomputes as well. But that may be improved by improving queue_job.

what do you mean? You should create your own batches by size. 

> Also, I don't want to miss the forest for the trees. Most of the code about the export modifications could be removed if we forbid to modify exports manually. It would mean that an export modification would require an instance restart, which is actually reasonable. In Practice, that really should not happen, especially for any big db.

Hmm I'm :-1: for this: it shouldn't require a restart.

len-foss:
Is there any problem you can't solve with one more layer of indirection? I realized how to solve the compute issue when reading your comment. So the current version only uses standard ORM features, and there isn't even any trick, really.
(modifying the depends isn't really an issue, re-setuping the model is something that is already done in standard, in the mail module).

I think the alternative simple version can still be considered seriously, but there isn't any technicality favouring one over the other anymore.

jsonify_stored/models/ir_exports.py
lmignon: @len-foss Are we sure that the invalidation doesn't occur each time we update an addon defining ir.export.line into xml data files?

len-foss: We are sure that invalidation will occur n+1 time where is n is the number of lines at the export update. So that's bad.
The export should be put in noupdate, or put into a separate module that would only contain the export; otherwise, that would require really complicated mitigation strategies.

simahawk:
@len-foss @lmignon hello :)
I'm getting back to this subject. Have you done any progress here? 
I'd like to move it fwd but on v14.

len-foss:
Well. Not really, I don't have the time.
It works pretty well in production with the two timestamps system.
There's one issue with dynamic trigger: if module B > A adds some new fields to the json export, then testing A would crash at the trigger setup.
Because the new field can be on some model unrelated to the one we're working on, and that the order they are setup is 'random', there is no way around that. 
That might not be a deal breaker, but this is quite annoying (to test A, you make a fresh DB without B).
Also there was no explicit spec on how to manage the asynchronous update, while all solutions have particular trade-offs; one way around that would be to actually test implementations with some realistic enough load, but that would require some time.
All in all, it's quite finicky, has some definitive drawbacks that need to be made consciously, and has some problems. So I'm not sure it's such a good idea. 
Going forward, I'll get out of Odoo development as soon as I can, but I'll help if required before that.

jsonify_stored/models/ir_exports_line.py
```diff
@@ -0,0 +1,26 @@
...
+
+from odoo import api, models
+
+
+class IrExportLine(models.Model):
+    _inherit = "ir.exports.line"
+
+    def write(self, vals):
+        result = super(IrExportLine, self).write(vals)
+        if any(field in vals for field in ["name", "alias"]):
```
florentx, line 12: Why do you expect `"alias"` in the write fields? This field is not present in `ir.exports.line` anywhere.

test_jsonify_stored/models/test_jsonify_mixin.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestJsonifyStored(models.Model):
+    """Reference Implementation of the mixin.
+    """
+
+    _name = "test.jsonify.stored"
+    _inherit = "jsonify.stored.mixin"
+
+    _export_xmlid = "test_jsonify_stored.model_export"  # created in pre_init hook
+
```
florentx, line 15: Obsoleted?
```suggestion
```

jsonify_stored/models/ir_exports_line.py
simahawk: it has been renamed afterwards

PR #1930, opened 2020-11-17
[FIX][8.0] base_suspend_security x TransientModel

StefanRijnhart:
Grant access on transient models in the case of suspended security.
Otherwise, the ORM will try to compare the suspended security uid with
the integer create_uids from the records in the database

### DIFF
--- a/base_suspend_security/models/ir_rule.py
+++ b/base_suspend_security/models/ir_rule.py
@@ -21,6 +21,19 @@ from openerp import models, api
 from ..base_suspend_security import BaseSuspendSecurityUid, SUSPEND_METHOD
 
 
+@api.multi
+def check_access_rule(self, operation, context=None):
+    """ Grant access on transient models in the case of suspended security.
+    Otherwise, the ORM will try to compare the suspended security uid with
+    the integer create_uids from the records in the database
+    """
+    if self.is_transient() and isinstance(
+            self.env.uid, BaseSuspendSecurityUid):
+        return True
+    return self.check_access_rule_before_suspend_security(
+        operation, context=context)
+
+
 class IrRule(models.Model):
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

StefanRijnhart:
Pushing an API fix to fix this error in our own CI in which `ir.rule::_register_hook()` was called before loading `mail`'s demo data

```
  File "odoo/addons/mail/mail_followers.py", line 98, in _check_rights
    obj.check_access_rule('read')
ParseError: "check_access_rule() takes at least 5 arguments (2 given)" while parsing odoo/addons/mail/data/mail_group_data.xml:5, near
<record model="mail.group" id="group_all_employees">
            <field name="name">Whole Company</field>
            <field name="group_ids" eval="[(4, ref('base.group_user'))]"/>
            <field name="description">General announces for all employees.</field>
        </record>
```

```
--- a/base_suspend_security/models/ir_rule.py
+++ b/base_suspend_security/models/ir_rule.py
@@ -21,15 +21,17 @@ from openerp import models, api
 from ..base_suspend_security import BaseSuspendSecurityUid, SUSPEND_METHOD
 
 
-def check_access_rule(self, cr, uid, ids, operation, context=None):
+@api.multi
+def check_access_rule(self, operation, context=None):
     """ Grant access on transient models in the case of suspended security.
     Otherwise, the ORM will try to compare the suspended security uid with
     the integer create_uids from the records in the database
     """
-    if self.is_transient() and isinstance(uid, BaseSuspendSecurityUid):
+    if self.is_transient() and isinstance(
+            self.env.uid, BaseSuspendSecurityUid):
         return True
     return self.check_access_rule_before_suspend_security(
-        cr, uid, ids, operation, context=context)
+        operation, context=context)
```

PR #609, opened 2016-11-16
[MIG] 10.0 super calendar

jeanpaulrobineau:
Hello,
I have ported this super calendar for version 10.

Kind regards,
Jean-Paul Robineau

PR #1935, opened 2020-11-21
[MIG] 12.0 super_calendar

diggy128:
- Fixed some minor Python2 > 3 issues.
- Odoo keeps microseconds for write_date and create_date fields and drops them for Datetime fields. As so, in order for duration to be consistent, microseconds are dropped entire to allow using any datetime field.
- Updated file headers and README to be consistent with OCA guidelines.

I should mention migration in based on @jeanpaulrobineau and PR #609 for v10.0


thomaspaulb:
@diggy128 Seems `flake8` is complaining about some unused imports.

PR #1938, opened 2020-11-24
[14.0][MIG] base_exception: Migration to 14.0

kevinkhao:
Breaking changes: in code evaluation context, no more obj. and object. , no more time module which is disallowed.


hparfr (approved):
LGTM

kittiu:
Code reivew, LGTM.

Just my curiosity. Why is `time` disallowed?

> Breaking changes: in code evaluation context, no more obj. and object. , no more time module which is disallowed.

kittiu (approved):

base_exception/views/base_exception_view.xml
```diff
@@ -0,0 +1,107 @@
...
+                            <group>
+                                <div style="margin-top: 4px;">
+                                    <h3>Help with Python expressions</h3>
+                                    <p
+                                    >Various fields may use Python code or Python expressions. The following variables can be used:</p>
+                                    <ul>
+                                        <li><code
+                                            >self</code>: Record on which the rule is evaluated.</li>
+                                        <li><code
+                                            >time</code>: useful Python libraries</li>
```
kittiu: As you remove `time` from context eval, you might want to remove this help too.

atchuthan (approved):
Code review :+1:

pedrobaeza (dismissed):
If there are breaking changes, a migration script should convert current exceptions to new format. I don't really understand why that changes are introduced. What are the reasons for that?

base_exception/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+Terms used in old api like `pool`, `cr`, `uid` must be removed porting this module in version 12.
```
hparfr: This can be removed

kevinkhao:
Thanks for feedback.

@kittiu pls check the following:
https://github.com/odoo/odoo/blob/1557a648d96bfa725aa7bfa44773e6bb92d41112/odoo/tools/safe_eval.py#L369
This raises if we try to include time as a module.
time is already included here:
https://github.com/odoo/odoo/blob/1557a648d96bfa725aa7bfa44773e6bb92d41112/odoo/tools/safe_eval.py#L410

@pedrobaeza I saw it was noted as "deprecated" and removed it maybe a bit too quickly. I don't see a simple migration script for this (a naive replacement of obj/object with self will probably break something), and keeping obj/object seems simpler and not in contradiction with what's found in Odoo core, so it doesn't seem crazy to keep it as is

pedrobaeza:
@carlosdauden does this affect you?

kevinkhao:
sorry I should be clearer: on my last force-push I kept it as it was before, that means, we can still use obj and object

hparfr:
If they were deprecated I don't see why we should keep them.

kevinkhao:
@hparfr I think after discussion with @florian-dacosta the idea for deprecating obj/object was that self can be used instead. But to me it doesn't seem crazy to keep it either, since in Odoo we also have that in other places (unless it was changed recently)

pedrobaeza:
OK, thanks for the changes.

kittiu:
Thank you @kevinkhao and all contributors of this modules, a good one though.

Just for discussion. I am thinking that, this feature can be the extension to Automated Action in Odoo, and it can be used for all models without having to create extension on every model. Anyone has though about it?

kevinkhao:
@kittiu I think because these ideas are so abstract it is easier to talk about it with concrete examples. Anyways I have a feeling that in the end it will be the same question that comes back again: do we have business logic in the data (server actions, crons) or python code.

PR #1941, opened 2020-11-30
[14.0] ADD upgrade_analysis

legalsylvain:
Work in progress. Do not review for the time being.

### Done

* [x] rename module ``openupgrade_records`` into ``upgrade_analysis``.
* [x] recover module ``openupgrade_records`` and ``odoo_patch``, preserving author.
* [x] full refactor using new odoo convention, pre-commit, ...
* [x] in the ``upgrade.comparison.config`` get the version of the target odoo.
* [x] replace the transient model ``openupgrade.analysis.wizard`` by a non transient model ``upgrade.analysis``. it allows to make many analysis. it allows also to add in a second time, more metrics, (like number of changes, number of new models, ...)
* [x] in the ``upgrade.install.wizard``, add the possibility to choose the modules to install (all the Odoo module, all the OCA modules, custom modules, etc...)
* [x] files are now written in the path of module ``get_module_path``, if it's a "non Odoo SA module" (OCA modules, custom modules) or in the ``upgrade_path`` setting for the official modules. (this setting has to be present in the odoo.cfg file or with the option ``--upgrade-path``). (introduced here : https://github.com/odoo/odoo/pull/32650)
* [x] remove useless patches that are not needed for this module. Ref : https://github.com/OCA/OpenUpgrade/issues/2440#issuecomment-721392761. Done here https://github.com/legalsylvain/server-tools/pull/3 by @StefanRijnhart. 
* [x] Check if all patches are required in addons and for the function ``quirk_standard_calendar_attendances``. (done here https://github.com/legalsylvain/server-tools/pull/3#discussion_r533233275))

### Todo :

* [ ] rewrite readme files
* [ ] update the helper script [available here](https://github.com/OCA/OpenUpgrade/wiki/Crude-script-to-create-the-full-analysis-between-two-versions-of-Odoo) to make it working with any version
* [ ] handle the apriori duplication. (or not).

### Usage : 

- install the module ``upgrade_analysis``
- (optional) install modules via the wizard
- generate records
- Create a comparison configuration and test the connexion
- Click on "New Analysis" 
- Click on "Perform Analysis"


StefanRijnhart:
My PR here https://github.com/legalsylvain/openupgrade-framework/pull/1 improves on some of the patches and it adds a mechanism to apply a collection of patches only when we need them. Would you consider taking some of these changes? I can propose a PR onto your branch here.

legalsylvain:
Hi @StefanRijnhart. Yes, I like your ``OdooPatch`` class. I think it's a great improvment.
If you have time, yes feel free to contribute on that branch and make a PR. (If you prefer, I can give you write access on my repo, and we will squash commits at the end of the refactor.)
Let me know. I'll not work again on that topic this week.

StefanRijnhart:
You can find my PR into this branch here: https://github.com/legalsylvain/server-tools/pull/3

legalsylvain:
Hi @StefanRijnhart : 

1) it seems travis is broken because of this line : https://travis-ci.com/github/OCA/server-tools/jobs/451380734#L796
did you missed to commit the removal of an import here https://github.com/OCA/server-tools/pull/1941/files#diff-3008ab525e7f6378f5fc9b15bda124e1a301da8b39ba4f4589b92671c2192e65R2 ?

2) when I made the split of the openupgrade project into two project (``upgrade_analysis`` and ``openupgrade_framework`` I saw that a file was required for both part the ``apriori.py`` file sample [here](https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/openupgrade_records/lib/apriori.py#L1) for V13) 

As far as I can understand there dict are used : 
- for the "analysis" part, to understand which module has been renamed / merged. in the ``compare.py`` function [here](https://github.com/OCA/server-tools/pull/1941/files#diff-f73f13e2913d70d038314d08f648642baaef2309899ab66173c2af33eefaabe9R17).
- for the "upgrade" part, to make some operations at the beginning of the upgrade process. ([here](https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/base/migrations/13.0.1.3/pre-migration.py#L10), for V13)

We have basically three options : 
A) duplicate the file and the data.
B) make an optional dependency from a module to another.
As the ``openupgrade`` project allways requires this files to work properly and ``upgrade_analysis`` requires this file only to analyse odoo SA modules differencies, i propose to update the file ``apriori.py`` in the module ``upgrade_analysis`` with somethink like that : 

```
try : 
    from odoo.addons.openupgrade_framework import apriori
except:
    apriori = False

[...] 

if apriori:
    renamed_modules.update(apriori.renamed_modules)
    merged_modules.update(apriori.merged_modules)
```

It adds the constraints to have the ``openupgrade_framework`` module in the addons path when running analysis.

C)  make a standalone module ``openupgrade_module_changes`` with this file and make both module ``upgrade_analysis`` and ``openupgrade_framework`` depends on this simple module. Maybe more visible. also, when porting the module ``openupgrade_framework`` we never have to port the data from a version to another, so it makes sense to have a distinct module.

**My point of view :**

I prefer to avoid A, because it is duplicating the data.

I maybe prefer C, but B is also OK. I just think that most of the changes of this file are done when "upgrading", so anyway, it makes sense to set the data in the ``Openupgrade`` repository better than in the ``server-tools`` repository.

What do you think ?

Note : it doesn't seems to be a huge trouble for the current migration (13 -> 14) because there does not seem to be any renaming or module merging between these two versions. Ref : https://github.com/OCA/OpenUpgrade/issues/2190

StefanRijnhart:
Thanks for noting the travis breakage. Will look at that tonight. I think apriory belongs to openupgrade_scripts rather than the framework, because it is different for every edition. Maybe we should try to import apriori from `upgrade_path`? I can implement that if you like.

legalsylvain:
>  think apriory belongs to openupgrade_scripts rather than the framework, because it is different for every edition

yes you're right ! it seems the best solution. :+1: 

>  Maybe we should try to import apriori from ``upgrade_path`` ?

Not sure to understand your proposal.

- do you mean ``config.get('upgrade_path')`` ? If yes, it is now optional, due to https://github.com/legalsylvain/server-tools/pull/3/files#diff-e768c0bad8381e3d5a9cef3f3193890e5438592d5e9da9486c523bb55b19805aR37 so we can no longer rely on this value.

- if you mean ``upgrade_path`` (of model``upgrade.analysis``) it is too late, because ``apriori`` is required in the first step (generating records) so, analysis is not created and it will not allways work with the "automatic script". (https://github.com/OCA/OpenUpgrade/wiki/Crude-script-to-create-the-full-analysis-between-two-versions-of-Odoo)

For the time being, I see the following solution as the more simple 
```
try:
    from odoo.addons.openupgrade_scripts import apriori
```

We can, in that case, add a warning in the logs like ``openupgrade_scripts addons is not available on your system. Your analysis may not be complete if you try to analyse changes of Odoo SA or OCA modules``

StefanRijnhart:
Not sure if `openupgrade_scripts` will show up under odoo.addons as it is not necessarily a python module (or an Odoo addon), but if it works that would be best. Otherwise we can add the upgrade_path (or apriori path) to other wizards as is needed, using config.get('upgrade_path') as a default again.

legalsylvain:
> Not sure if openupgrade_scripts will show up under odoo.addons as it is not necessarily a python module (or an Odoo addon), but if it works that would be best

Indeed, ``openupgrade_scripts`` should be an odoo modules if we want the import of ``odoo.addons.openupgrade_scripts`` to work.
For the time being, I designed the refactor in this way : See : https://github.com/legalsylvain/openupgrade-framework/tree/14.0-upgrade-pocalypse/openupgrade_scripts

Is it an odoo module ? not really I confess, and there is no need to install it, and install it will do nothing. (as ``openupgrade_framework`` BTW that only requires to be present in ``server_wide_modules``). The original reason is the visibility of such design. If it is a module, it can be available for download on OCA appstore, and maybe on odoo appstore (if openupgrade word is not blacklisted ;-) ... ), also available on pypi without any actions. So easy to download, easy to "deploy". also it comes with a nice readme generated with fragment if we want to add extra explanation, or roadmap section). We can also use ``merge patch | minor`` command of ocabot to merge PR.

Then, the instruction for end user will be quite simple : 
For an upgrade,
1) get and install odoo in the way you prefer.
2) download both ``openupgrade_framework`` and ``openupgrade_scripts`` modules in the way you prefer. (pypi, git, download via stores)
3) add a key in your ``odoo.cfg`` file. (the ``upgrade_path``)
4) update all.
5) Pray. (Unfortunately, this step will be difficult to remove ;-))

I've already talked about the importance of simplifying the use of openupgrade here. https://github.com/OCA/OpenUpgrade/issues/2440#issuecomment-723048940

do you see any negative effects to encapsulate the migration scripts in an odoo module ?

StefanRijnhart:
OK, disguising the scripts directory as a module is fine. I'll use its location on the file system as a default for upgrade_path instead of the configuration key then.

pedrobaeza:
I'm not sure of understanding about openupgrade_scripts, but if your intention is to encapsulate all migration scripts there, that's a very bad option. We should use upgrade-paths or similar argument line option, and respect the addon hierarchy in a separate repo with that scripts. We can even add migration scripts from other OCA addons.

legalsylvain:
> I'm not sure of understanding about openupgrade_scripts, but if your intention is to encapsulate all migration scripts there, that's a very bad option. We should use upgrade-paths or similar condos line option, and respect the addon hierarchy in a separate repo with that scripts. We can even add migration scripts from other OCA addons.

``openupgrade_scripts`` is the place to store all the migration scripts for Odoo SA addons only. https://github.com/legalsylvain/openupgrade-framework/tree/14.0-upgrade-pocalypse/openupgrade_scripts/scripts

if it's not an odoo SA modules, (custom, OCA) the analysis file will be located as usual, in the ``migrations`` folder of the module and it will work correctly.

pedrobaeza:
Then I insist that shouldn't be the way to go. Scripts should be outside any Odoo module, and follow the same structure as previously:

account/migrations/14.0.1.0/*

and then use them with --migration-paths (or similar name) option when launching the upgrade. This is also the way Odoo does it.

This allows to have a more intuitive structure, we can add more Odoo major version scripts from other OCA repos, etc.

Controlling a mega-module with all the Odoo core scripts has no sense for me, it requires a big maintenance work, and even the version bumping you are commenting doesn't serve for anything. What a minor/major/patch version bumping means?

StefanRijnhart:
@pedrobaeza thank you for that pointer. I see Odoo supports a --upgrade-path parameter for this. I agree we should use this for the core modules instead of a pseudo Odoo module.

As for your suggestion to also store scripts here for OCA modules etc., I am not sure. I think keeping them within the actual module has its advantages.

StefanRijnhart:
And apologies @legalsylvain for not recognizing what the `upgrade-path` configuration key, which you already used in the code, stood for.

pedrobaeza:
I'm talking about only major Odoo versions of OCA modules. Advantages of this approach:

- Everything in the same place.
- We can avoid large CIs execution for something that is not being tested.
- We can maybe add CI to these scripts using same approach than with OpenUpgrade.
- Regular users of the OCA module doesn't need these migration scripts. Only DB migrations.
- On auto-update systems based on checksums, don't trigger an update for those added migration scripts.
- Catch attention for reviewers having them in the same repo, and not split through all the OCA repos.

For major changes on the module that require migration scripts, of course they still should be included inside OCA module itself. Example: https://github.com/OCA/credit-control/pull/86

StefanRijnhart:
For me, commiting the migration script together with the migration of the module code is 'having everything in the same place', as well as having the reviewers review both together.

The other points you mention about the infrastructure overhead of the additional changes, I don't think they are a big problem because migration scripts for modules, even between major versions, are rare, especially if the migration script is committed together with the module code which must surely be the most common case.

pedrobaeza:
@StefanRijnhart migration scripts for a change of version are not committed at the same time as the migration module. The more recent example is in all the account related stuff: only when we have cleared the structure of the `account` migration scripts and which fields we rename/fill for linking old objects with new, we have been able to create migration scripts for all the OCA modules that touch invoices.

But meanwhile the module can be migrated for being used in the new version for those not migrating DB. And not only that, you can't demand a contributor to provide migration scripts for migrating the module from an old version when contributing the module migrated to the new one.

About the CI overhead, I can assure you is not pleasant to wait 15/20 minutes for nothing. I commit directly sometimes for avoiding that thing and due to there's no real gain on running such CI.

StefanRijnhart:
Everybody knows what it's like to wait on the CI. But that's not the same thing as the load that migration script PRs add to the total CI of the OCA. I'm sure that the merge of the invoice models in the move models caused a flurry of migration scripts when the first database migrations were being developed, but in general I think it's not unreasonable to expect the migration script in the code migration PR. I have seen that many times to be the case.

pedrobaeza:
OK, we can leave such part to the contributor decision of doing one way or another, as anyway they will need to provide the general upgrade path for doing the migration, and also the addons path for the OCA modules.

legalsylvain:
Hi pedro & Stefan,

## 1) regarding pseudo module

 I think that there is some misunderstood.

> (@pedrobaeza said : ) Then I insist that shouldn't be the way to go. Scripts should be outside any Odoo module, and follow the same structure as previously: account/migrations/14.0.1.0/*

My bad ! Indeed, for the time being, analysis is generated in ``openupgrade_scripts/scripts/account/14.0.1.1/openupgrade_analysis.txt``
and the ``/migration/`` part is missing. I think that it is fixed [now](https://github.com/OCA/server-tools/pull/1941/files#diff-e768c0bad8381e3d5a9cef3f3193890e5438592d5e9da9486c523bb55b19805aR69) and that the new analysis file should be generated with the good path, I mean : ``openupgrade_scripts/scripts/account/migrations/14.0.1.1/openupgrade_analysis.txt``

> (@pedrobaeza said : ) and then use them with --migration-paths (or similar name) option when launching the upgrade. This is also the way Odoo does it.

Yes, that is exactly what I propose ! running odoo with that option. 
``--upgrade-path=[...]/openupgrade_scripts/scripts/`` should work with the current PoC.

> (@StefanRijnhart said : ) @pedrobaeza thank you for that pointer. I see Odoo supports a --upgrade-path parameter for this. I agree we should use this for the core modules instead of a pseudo Odoo module.

That is not incompatible. I propose to use a pseudo odoo module only to make the script more accessible and visible. But my objective is to stick to the standard and use the ``--upgrade-path`` option.

> (@pedrobaeza said : ) Controlling a mega-module with all the Odoo core scripts has no sense for me, it requires a big maintenance work

I don't get it. What I propose is simply to put the main ``script`` folder of the OpenUpgrade project (that contains all the analysis and the migration files) in a folder that contains a ``__manifest__.py`` file. Nothing more. Could you explain "the big maintenance" generated by this move ?

## 2) Where are the migration scripts of the OCA modules ?

> (@pedrobaeza said: ) migration scripts for a change of version are not committed at the same time as the migration module

Well, it depends. Two situations : 
- "Port without refactoring" : a module exists in a previous version (like 12.0). When a new version is available (14.0) contributors will port the module to have it available in recent version and they don't care about the migration from an old version because they just want to use it for a new customer, for the recent version. 
In a second time, people that are migrating huge instance, with a lot of OCA modules will want to have the migration scripts and so develop them. For that extra works, they often requires extra features that can be in the openupgradelib project. (like function to handle partner refactor between 6.1 and 7, or to handle binary field change from db to attachement (version 9+))
-> made in two steps, with two distinct contributors : you're right.

- "Port with refactoring" : in some cases, maintainers will take advantage of a new version to make deep refactoring. In that case, they will develop in the same time, migration script AND the port / refactor of the module. A recent personal exemple : https://github.com/OCA/web/pull/1428
Another example (futur) : I want to refactor the ``mass_editing`` [module](https://github.com/OCA/server-ux/tree/13.0/mass_editing) to inherit of core ``server.actions`` model and remove useless code. 
-> made in one step, with a single contributors : @StefanRijnhart is right, it is annoying in that case to have to make 2 PR, in two repositories, making one PR waiting for the second one.

> We can avoid large CIs execution for something that is not being tested.
> Regular users of the OCA module doesn't need these migration scripts. Only DB migrations.
> On auto-update systems based on checksums, don't trigger an update for those added migration scripts.

Well, there are not a lot of migrations folders in the 2000+ OCA modules. I think it doesn't concern more than 5% of the modules in each version. Interesting point, but not a blocking one for me.

> We can maybe add CI to these scripts using same approach than with OpenUpgrade.

That could be great ! But it requires extra work to make a specific CI.

But but there's another point you didn't address : mixted migration of instance on premise with EE modules (and contract) and OCA modules. (I may be wrong because I never used that service, so excuse me if it's not relevant)
If a user want to upgrade a instance with EE modules and OCA modules installed, _I guess_  :
- he will dump the database on https://upgrade.odoo.com/database/upload. 
- Odoo will make the upgrade of the odoo modules, letting OCA modules in a ``to upgrade`` state and provide the migrated database.
- Then user has to finish the migration, making a full ``--update all`` with all the ported modules (OCA / custom) of the targeted version. In that step, the migrations scripts will be executed. 

But until now, user doesn't have to use openupgrade project for that purpose. if the final update is done without the openupgrade project, the migration will fail silently. We can assume that EE users doens't know very well the openupgrade project.
But again, I don't know very well all the EE services, and  I'm not aware of the recent new services with extra supports, which provides support (and maybe migration) of third party modules at a very competitive price. For example :

image

StefanRijnhart:
I've submitted a PR on this PR's branch to fix the import error and refactor apriori handling. It is now imported from the --upgrade-path. It seemed safer to do this as JSON rather than a Python import.

https://github.com/legalsylvain/server-tools/pull/4

Meanwhile, the resulting analysis and an empty apriori.json are added here: https://github.com/legalsylvain/openupgrade-framework/pull/2 (oh, and I'm very sorry but I removed the manifest from openupgrade_scripts because I can see another poll coming, or more than one).

legalsylvain:
> I've submitted a PR on this PR's branch to fix the import error 

thanks ! 

> It seemed safer to do this as JSON rather than a Python import.

I have to think about it.

> (oh, and I'm very sorry but I removed the manifest from openupgrade_scripts because I can see another poll coming, or more than one)

Well, I would like to wait for valid counter-arguments from @pedrobaeza (or other people) that would indicate why a module is a bad idea. For the time being, I only see improvment explained here. https://github.com/OCA/server-tools/pull/1941#issuecomment-736542830

pedrobaeza:
I still don't see any advantage of using such complex scheme. You are saying it's better to have `openupgrade_scripts/scripts/account/migrations/14.0.1.1/openupgrade_analysis.txt` than `account/migrations/14.0.1.1/openupgrade_analysis.txt` inside OCA/OpenUpgrade?

About your arguments for being better:

- Why do you want to do `from odoo.addons import openupgrade_scripts`? It's not even possible to import anything inside `14.0.1.1` due to the dot notation, so it doesn't serve for anything.
- I repeat my same question about for what we want to bump versions. What a minor/major/patch version bumping means in this context? You should know that we can trigger the same the bot merge command without having modules. It will work the same. We are already doing it on OpenUpgrade itself.
- You have to put as --upgrade-paths `<download_folder>/openupgrade_scripts/scripts` instead of only `<download_folder>`. That's not favoring at all the simplicity.
- You are talking about publishing to pip, but that doesn't serve without the infamous `setup` directory and other required steps that will complicate the CI of the project (that's why I'm telling about maintenance). It's far easier for newbies IMO to simply download the ZIP from GitHub UI, uncompress it, and point with `--upgrade-paths` to that uncompressed folder.


About my proposal for OCA modules, you have to do 2 PRs indeed, but you don't have to wait if you don't want, unless we enable a CI for such modules, but if we enable it, people will want to test such migration scripts and will prefer to do both PRs for that reason for sure.

For EE migrations, there's also no theoretical problem: people still download whole OpenUpgrade repo and apply the same --upgrade-paths for the second round with `--update all`, and Odoo core modules won't be touched, as they already have the latest version updated in ir_module_module table and no migration scripts from OpenUpgrade will be triggered. Only OCA modules ones.

Anyway, don't be naive: nearly 0 OCA modules migration scripts will work with EE migrations. Why? Because these migration scripts expect certain data scheme that belongs to OpenUpgrade, not the way Odoo EE handles the migration. Some examples:

- Use of `openupgrade_legacy_xx_*` columns.
- In v13, we populate a column `old_invoice_id` for `account.move` and such column is used for filling fields that were in invoice and now must be in moves. I doubt Odoo uses the same column name.

Anyhow, I'm telling to do it as an option. Contributors that still want to bundle everything together, can do it as well.

StefanRijnhart:
Sylvain mentions the visibility of the migration project, having the module with the scripts on the apps store for instance. I find that slightly misleading. But the visibility of the project could be a reason to keep the migration scripts in the module directories in the case of OCA modules, so as not to completely isolate the regular contributor from these scripts (and indeed, encourage them to contribute to those as well).

yajo:
:thinking: I think what we need to answer before is: when you add `--upgrade-path`, migration scripts from the module's `migrations` folder are still loaded and processed? Because, if yes, then the whole argument is pointless, as we will be able to put the script inside OpenUpgrade or inside the module, depending on each specific situation (module needs, developer preferences...).

upgrade_analysis/compare.py
```diff
@@ -0,0 +1,484 @@
+# Copyright 2011-2015 Therp BV <https://therp.nl>
+# Copyright 2015-2016 Opener B.V. <https://opener.am>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# flake8: noqa: C901
+
+#####################################################################
+#   library providing a function to analyse two progressive database
```
yajo, line 7: I've reviewed this file, not in depth. However I see that this is not very much related to a single specific odoo version. Wouldn't `openupgradelib` be a better place to store all this code?

upgrade_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    is_oca_module = fields.Boolean(compute="_compute_is_oca_module")
+
+    def _compute_is_oca_module(self):
+        for module in self:
+            if "/OCA/" in module.website:
+                module.is_oca_module = True
+            else:
+                module.is_oca_module = False
```
yajo:
```suggestion
            module.is_oca_module = "/OCA/" in module.website
```

upgrade_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _compute_is_odoo_module(self):
+        for module in self:
+            module_path = get_module_path(module.name)
+            absolute_repo_path = os.path.split(module_path)[0]
+            x, relative_repo_path = os.path.split(absolute_repo_path)
+            if relative_repo_path == "addons":
+                module.is_odoo_module = True
+            else:
+                module.is_odoo_module = False
```
yajo:
```suggestion
            module.is_odoo_module = relative_repo_path == "addons"
```

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -0,0 +1,227 @@
...
+        comodel_name="upgrade.comparison.config",
+        readonly=True,
+        required=True,
+    )
+
+    log = fields.Text(readonly=True)
+    upgrade_path = fields.Char(
+        default=config.get("upgrade_path", False),
+        help="The base file path to save the analyse files of Odoo modules",
+    )
```
yajo: I don't understand why this is a field. Is it really possible to tell Odoo different upgrade paths per module? If not, I'd remove it and just get the info from the config.

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -0,0 +1,227 @@
...
+                os.makedirs(full_path)
+            except os.error:
+                return "ERROR: could not create migrations directory %s:\n" % (
+                    full_path
+                )
+        logfile = os.path.join(full_path, filename)
+        try:
+            f = open(logfile, "w")
+        except Exception:
+            return "ERROR: could not open file %s for writing:\n" % logfile
```
yajo, line 108: In general I always make the `odoo` user unable to write into odoo and module source files. I consider that a good security measure, because otherwise a poisoned module could infect all other modules easily.

I think you should be able to configure a specific path to output all this stuff, and that path should be inside odoo's data dir by default. If you really want a module to alter code of other modules, that should be stated explicitly IMHO.

upgrade_analysis/models/upgrade_attribute.py
```diff
@@ -0,0 +1,20 @@
...
+class UpgradeAttribute(models.Model):
+    _name = "upgrade.attribute"
+    _description = "Upgrade Attribute"
+
+    name = fields.Char(readonly=True)
+
+    value = fields.Char(readonly=True)
+
+    record_id = fields.Many2one(
+        comodel_name="upgrade.record",
```
yajo, line 17:
```suggestion
        comodel_name="upgrade.record",
        index=True,
```

https://github.com/odoo/odoo/pull/39332

upgrade_analysis/models/upgrade_comparison_config.py
```diff
@@ -0,0 +1,87 @@
...
+            "params": {
+                "type": "info",
+                "message": _(
+                    "You are correctly connected to the server {server}"
+                    " (version {version}) with the user {user_name}"
+                ).format(
+                    server=self.server,
+                    version=self.version,
+                    user_name=user_info["name"],
+                ),
```
yajo: Use % strings for translations. It is easier for translators, who are more used to it, and opens less security backdoors:

```suggestion
                "message": _(
                    "You are correctly connected to the server %(server)s"
                    " (version %(version)s) with the user %(user_name)s"
                ) % dict(
                    server=self.server,
                    version=self.version,
                    user_name=user_info["name"],
                ),
```

pedrobaeza:
AFAIK all are processed, no matter its source

upgrade_analysis/odoo_patch/odoo_patch.py
```diff
@@ -0,0 +1,59 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+class OdooPatch(object):
```
yajo, line 6: I also think this code is a good candidate to exist inside `openupgradelib` instead.

upgrade_analysis/tests/test_module.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertTrue(
+            self.upgrade_analysis.id in wizard.module_ids.ids,
+            "Select OCA module should select 'upgrade_analysis' module",
+        )
+
+        wizard.select_other_modules()
+        self.assertTrue(
+            self.product_module.id not in wizard.module_ids.ids,
+            "Select Other module should not select 'product' module",
+        )
```
yajo:
```suggestion
        self.assertFalse(
            self.product_module.id in wizard.module_ids.ids,
            "Select Other module should not select 'product' module",
        )
```

https://www.recallstack.icu/en/2020/09/01/do-not-use-negative-booleans-please/

upgrade_analysis/upgrade_log.py
```diff
@@ -0,0 +1,217 @@
...
+import logging
+
+from openupgradelib.openupgrade_tools import table_exists
+
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+
+
+def get_record_id(cr, module, model, field, mode):
```
yajo, line 14: Possibly another file that should exist in `openupgradelib`.

sebastienbeau:
Hi some feedback.

Regarding the way to publish the script of OpenUpgrade and the framework I see three possibility
- an odoo module
- a pure python package
- a simple git directory

In a pure technical point of view, I would prefers to have a dedicated command line (click-odoo can make it easier)
So basically if I want to migrate I just do "pip install openupgrade-odoo-14" and launch "openupgrade-odoo-14 run"
No configuration, no complex stuff, just 2 cmd line and migration is done.
The openupgrade-odoo-14 lib include the migration script and depend on openupgrade-framework lib.

In a pure marketing  point of view, I would prefers an odoo module.
You install it in the way you like. For pip guy you can do a "pip install odoo14-addons-openupgrade-script" update odoo config and then run an "odoo -u all"
The main advantage is the publication on the apps store of odoo and oca. And we still have a python package.

Regarding the last solution, the only advantage are you do not have to create a "__manifest__.py" (odoo module) or a setup.py (pure python package)
As we have to create this file only one time for all, I do not think it worth to use that option.

Regarding the bump of version, we just need to define a convention
major : for adding a module
minor : for a fix in a module
path : for an improvement that do not fix something (perf improvement)


As it's seem (regarding a previous discussion in openupgrade) do not want to have a dedicated command line. I am really more in favors of having an odoo module for packaging the script. And please don't do a mix where you have a module and a folder of script, go for two odoo module.

I any way, thanks for your work separating the framework and the migration outside of the current version is really a good idea.

yajo:
> In a pure technical point of view, I would prefers to have a dedicated command line (click-odoo can make it easier)

That's exactly what I also prefer, but in https://github.com/OCA/OpenUpgrade/issues/2440#issuecomment-723081749 we clearly lost :shrug:

pedrobaeza:
It's impossible to have a click-and-go solution, as it depends on a lot of factors (extra addons, database configuration, deployment option), so just forget the command-line option. About the module, as said, it adds an extra folder level without real gain. I will redirect you all the newbies issues asking about how they install the module to perform the migration :stuck_out_tongue_closed_eyes:

upgrade_analysis/models/upgrade_analysis.py
legalsylvain: Hi @Yajo : this module is not intended to be used in any production instance and should never be used. (see all the warning, when using this module).

It is just made to make analysis, one time, when a new release is out.
The process to write in the odoo folders is the same as for the previous OpenUpgrade version.

upgrade_analysis/models/upgrade_analysis.py
legalsylvain: > I don't understand why this is a field. Is it really possible to tell Odoo different upgrade paths per module

yes it is. even it is very theoritical.

upgrade_analysis/models/upgrade_comparison_config.py
legalsylvain: Thanks. Didn't know that ``format`` was opening security issues. Could you point a website with more information ? 

Anyway, commited your suggestion.

legalsylvain:
Thanks @Yajo. I commited your suggestion.

yajo:
Please amend this commit and remove the mention to `@Yajo`, because otherwise github will spam me forever:
imagen

upgrade_analysis/models/upgrade_comparison_config.py
yajo: I think this one explains it a bit: https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

upgrade_analysis/models/upgrade_analysis.py
StefanRijnhart: I introduced it before I understood --upgrade-path is a standard Odoo option now. I think we can remove this now and  then we can reinstate apriori as a python module as far as I am concerned.

upgrade_analysis/odoo_patch/odoo_patch.py
StefanRijnhart: In the current simplistic design, it is not separable from the set of patches that it serves.

StefanRijnhart:
We all have different preferences for every aspect of the framework it seems. I agree with Pedro to package the migration scripts as a subdirectory of the repo (so not as an Odoo module) and proposed it like that into @legalsylvain's branches but that was the one change that Sylvain didn't want to take because he prefers the Odoo module approach. That's fair enough and maybe we should respect Sylvain's preference because he took this initiative.

pedrobaeza:
Yeah, of course the final word should be to the contributor who is doing the hard work, so giving all the opinions and arguments, now he can decide.

StefanRijnhart:
Great! I'll propose some small changes to consolidate the 'openupgrade_scripts as an odoo module' set-up later today.

legalsylvain:
Thanks @pedrobaeza and @StefanRijnhart. 
BTW, do you know how to generate the main [analysis file summary ](https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/openupgrade/doc/source/modules120-130.rst)?
I guess that it should be generated by this module, when generating all the analysis file, but I didn't find any script to do it. Did I missed something ? 
ref question : https://github.com/OCA/OpenUpgrade/issues/2440#issuecomment-740476616

pedrobaeza:
Till now, it has been generated by hand.

StefanRijnhart:
Found this in my notes
```
import os
import sys

old = os.listdir(sys.argv[1]) + ['base']
new = os.listdir(sys.argv[2]) + ['base']
modules = []
for module in sorted(set(old + new)):
    if module not in old:
        modules.append(' |new| %s' % module)
    elif module not in new:
        modules.append(' |del| %s' % module)
    else:
        modules.append(module)

col1 = max([len(module) for module in modules])
col2 = 49

line = ("+{:-<%s}+{:-<%s}+" % (col1, col2)).format('', '')
head = ("+{:=<%s}+{:=<%s}+" % (col1, col2)).format('', '')
row = "|{: <%s}|{: <%s}|" % (col1, col2)

print(line)
print(row.format('Module', 'Status'))
print(head)
for module in modules:
    print(row.format(module, ''))
    print(line)
```

StefanRijnhart:
@legalsylvain did you try out the snippet I posted above to generate the coverage table for the new edition? Otherwise, I can start preparing the docs (it's next on the list as far as I am concerned). I wouldn't mind doing it as the docs need to be updated with the new way of working anyway.

legalsylvain:
Hi @StefanRijnhart Didn't have the time to begin this work. you can do it, if you have time.

In the meantime, I merged all your pending PR (against OpenUpgrade and server-tools), and just fixed a conflict on the file ``.pre-commit-config.yaml`` so you can base your work on a fresh branch rebased on `` legalsylvain:14.0-ADD-upgrade_analysis``.

What I had in mind : (see https://github.com/OCA/OpenUpgrade/issues/2479)

- during the last step of the analysis process (comparing ``record`` and writing file), write the file
- maybe base the generation on a jinja template to have a template easy to update (specially header part)
- try to write the file in ``os.path.join(get_module_path("openupgrade_scripts", display_warning=False), 'readme', 'ROADMAP.rst')``, if the file doesn't exist.

As said in https://github.com/OCA/OpenUpgrade/issues/2479, it's just a proposal.

thanks for all this work, it's a pleasure to co-develop this refactoring with you !

legalsylvain:
Hi I rethought about the automatic generation of files ``noupdate_changes`` that @StefanRijnhart  introduced in a recent commit. 
 I'm thinking that if a file ``noupdate_changes.xml`` is generated, we could generate and also add this file ``post-migrate.py`` in the same folder :

```
# Copyright 2020 Odoo Community Association (OCA)
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
from openupgradelib import openupgrade


@openupgrade.migrate()
def migrate(env, version):
    # Load noupdate changes
    openupgrade.load_data(env.cr, "{{module_name}}", "{{module_version}}/noupdate_changes.xml")
```

for the v13 -> v14 ``upgrade_analysis`` script generated 50 files. 

it could be worth it, and save OpenUpgrade developers time?

I can implement it, if you think it relevant.

StefanRijnhart:
I considered the same thing, as a monkey patch in the framework. But I thought it would be unwise because the changes need to be reviewed, and some changes need to be commented out. Having the developer add the loading of the xml to the post script manually is the best way to ensure that review takes place.

legalsylvain:
> Having the developer add the loading of the xml to the post script manually is the best way to ensure that review takes place.

Makes sense. :+1:

StefanRijnhart (approved):
Already used this module to generate the 14.0 analysis, so it can be moved out of WIP status I think.

upgrade_analysis/compare.py
legalsylvain: Hi. I see the ``openugpradelib`` project a s a set of functions  to make upgrade. That file is not for upgrade but for analysis. so I think it's a good place.

Anyway, such changes can be done in another PR. The objective of this one is just to migrate / rename / move the ``openupgrade_records`` module.

upgrade_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,32 @@
...
+# Copyright 2016 Opener B.V. <https://opener.am>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import os
+
+from odoo import fields, models
+from odoo.modules import get_module_path
+
+
+class UpgradeAttribute(models.Model):
```
bealdav:
```suggestion
class IrModuleModule(models.Model):
```

legalsylvain: thanks !

.pre-commit-config.yaml
```diff
@@ -43,6 +43,7 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
+        exclude: ^upgrade_analysis/odoo_patch/
```
yajo: As I maintain the template, I'm interested on knowing... what's the problem with pre-commit in these files?

upgrade_analysis/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+        "views/menu.xml",
+        "views/view_upgrade_comparison_config.xml",
+        "views/view_upgrade_analysis.xml",
+        "views/view_upgrade_record.xml",
+        "wizards/view_upgrade_generate_record_wizard.xml",
+        "wizards/view_upgrade_install_wizard.xml",
+    ],
+    "installable": True,
+    "external_dependencies": {
+        "python": ["dataclasses", "odoorpc", "openupgradelib"],
```
yajo, line 23: Will this produce any kind of failure when installing on python 3.7, where `dataclasses` is native and this backport isn't needed?

.pre-commit-config.yaml
legalsylvain: odoo_patch is a place where odoo code is patched. to simplify diff analysis. 
some times, we require overwritting odoo function.
in that case, we don't want to apply black. (and all other checks) because it is a copy of odoo code.

upgrade_analysis/__manifest__.py
legalsylvain: I don't know. I run under python 3.6. 
what are you proposing ?

StefanRijnhart: This is not a problem. If no package is found with the name of the dependency, Odoo will attempt an import which succeeds.

.pre-commit-config.yaml
StefanRijnhart: It is actually the case that so far we have been able to copy large swaths of code, so for now maybe we can do without this exclude? We can always exclude individual patches here in th4e future if needed.

upgrade_analysis/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* scripts/compare_noupdate_xml_records.py should be integrated in the analysis process (#590)
```
StefanRijnhart: Oh, this one can be removed. Ready for merge then?

legalsylvain: > Oh, this one can be removed

Done.

.pre-commit-config.yaml
legalsylvain: Hi @Yajo : I remove pre-commit exception, and so fixed some style errors.

@StefanRijnhart : I received 3 B902 errors I decided to ignore. what do you think ? it's regarding patches part.

here is the diff : 
https://github.com/OCA/server-tools/pull/1941/commits/d3d94f1d8f79a256ebbe736bde754dfecf82bea5#diff-c5d89b073f972f8dc85c7bb7e8aff14db2b2499783fcc3763fd234062de73393R1

thanks.

StefanRijnhart: Yes, very good. Thanks!

StefanRijnhart (approved):

legalsylvain:
@Yajo, @bealdav could you update your review. I think we can merge this PR, as all is green, and the generation of files are OK. (https://github.com/OCA/OpenUpgrade/tree/14.0/openupgrade_scripts/scripts)

regards.

requirements.txt
```diff
@@ -1,2 +1,5 @@
 # generated from manifests external_dependencies
+dataclasses
+odoorpc
+openupgradelib
```
yajo, line 4: Don't you need to do a git install from master for openupgradelib? AFAIK the pypi version is always bad...

yajo (approved):

requirements.txt
bealdav: I confirm it's a bear trap to install openupgrade lib from pypi

legalsylvain: Well. Not sure to understand. When executing a migration with openupgrade project, yes we have to have the up to date library, because some function are oftently added / refactored / etc...

But for that module (``upgrade_analysis``) it is using only a function ``table_exists`` [here](https://github.com/OCA/server-tools/pull/1941/files#diff-cc8b10fd780fc230005882ef34b0ad0f32d7e66fe3ee501f299f3a65e3b25005R7) and that function is quite simple & stable.
https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_tools.py#L30

But, if you prefer, I can replace ``openupgradelib`` by ``git+https://github.com/OCA/openupgradelib.git@master``

pedrobaeza: It's better to always put the latest one, as we can even patch that method or improve this module using other methods of `openupgradelib` when needed.

legalsylvain: done.

legalsylvain: Hi. I have a problem with that change. It is making fail the pre-commit process : 

```
Generate requirements.txt for an addons directory........................Failed
- hook id: setuptools-odoo-get-requirements
- files were modified by this hook
```
Could you say me what to do ? 

thanks !

yajo: ah... 🤦🏼‍♂️ 

Could you please try just removing the openupgradelib line from here? I guess that dependency should be present in CIs always... (but I'm not sure)

legalsylvain: Well, I can not remove this line, because pre-commit readd this line ;-)

``Generate requirements.txt for an addons directory``

I so commited again what pre-commit suggests. (I mean ``openupgradelib`` in ``requirements.txt``)

I propose to merge this PR as it. if someone see a better implementation, feel free to improve in another PR.

PR #1942, opened 2020-11-30
[12.0] [IMP] sentry: migrate sentry-raven to new api sentry-sdk

fernandahf:
Description of the issue/feature this PR addresses:

Current behavior before PR: events sent for sentry-raven with context, excluding loggers, exceptions and filtering sense data. In fact, Sentry warns about migrate to new api: 

image

Desired behavior after PR is merged: events sent for sentry-sdk with context, excluding loggers, exceptions and filtering sensitive data, show events in sentry sdk without warning about update api. Sentry just shows sentry-sdk version without warning migrate.

image

--
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr


sentry/__init__.py
```diff
@@ -3,80 +3,99 @@
...
         _logger.debug(
             'Odoo directory: "%s" not a valid git repository', odoo_dir)
 
 
-def initialize_raven(config, client_cls=None):
-    '''
-    Setup an instance of :class:`raven.Client`.
+def initialize_sentry(config):
+    """ Setup an instance of :class:`sentry_sdk`. """
+    """ :param config: Sentry configuration """
```
moylop260:
```suggestion
    """Setup an instance of :class:`sentry_sdk`. 
    :param config: Sentry configuration"""
```

sentry/__init__.py
```diff
@@ -3,80 +3,99 @@
...
+    http.root = SentryWsgiMiddleware(http.root)
 
+    with sentry_sdk.push_scope() as scope:
+        scope.set_extra('debug', True)
+        sentry_sdk.capture_message('Starting Odoo Server', 'info')
+
+    return client
 
-sentry_client = initialize_raven(odoo_config)
+initialize_sentry(odoo_config)
```
moylop260: In fact, to initialize these kind of feature we need to be sure that the module is loading in odoo

```suggestion
def post_load():
    _logger.info("Initializing sentry...")
    initialize_sentry(odoo_config)
```

And it should be defined in the hooks of the manifest:

```diff
diff --git a/sentry/__manifest__.py b/sentry/__manifest__.py
index 8ef85da1..732cd858 100644
--- a/sentry/__manifest__.py
+++ b/sentry/__manifest__.py
@@ -22,4 +22,5 @@
     'depends': [
         'base',
     ],
+    'post_load': 'post_load',
 }
```

It will fix the following error loading modules for the longpolling:

```txt
Traceback (most recent call last):                                                                                                                                                                                                                   [136/1808]
  File "/home/odoo/instance/odoo/odoo/modules/registry.py", line 60, in __new__
    return cls.registries[db_name]
  File "/home/odoo/instance/odoo/odoo/tools/func.py", line 69, in wrapper
    return func(self, *args, **kwargs)
  File "/home/odoo/instance/odoo/odoo/tools/lru.py", line 44, in __getitem__
    a = self.d[obj].me
KeyError: 'DATABASENAME'

```

You can use another commit for that with `git commit --author="Moisés López <moylop260@vaxuoo.com>"` :)

sentry/__manifest__.py
```diff
@@ -15,6 +15,7 @@
     'installable': True,
     'external_dependencies': {
         'python': [
+            'sentry_sdk',
```
moylop260, line 20: I think we could be added in the `author` section like "Vauxoo"

And you can add the key [`"maintainers": ["your-github-login"],`](https://github.com/OCA/maintainer-tools/blob/f1b559574648c01558b543e7550f5b9be7c92784/template/module/__manifest__.py#L13) to add original authors and we both

moylop260 (changes requested):

moylop260:
Could you check if we can remove `raven` entirely as dependency, please?

moylop260:
@barsi @naglis @Versada
We are migrating the new api of sentry
Could you check it, please?

It is still WIP but if you can get a look here we will grateful

moylop260:
For the record,
A good improvement I saw using `sentry-sdk` instead of `sentry-raven` is it is adding a tag with the hash of the trace
 - Screen Shot 2020-11-30 at 20 18 57

I had needed before to have this value and I just use the following PR to add it manually:
 - https://github.com/Vauxoo/odoo/pull/358

Now it is auto-generated.

moylop260:
For the record,
Another improvement:
 - Before of this PR  `USER: IP Address 127.0.0.1`
 - After this PR `USER: IP Address "REAL-PUBLIC-IP"`

sentry/logutils.py
```diff
@@ -1,104 +1,12 @@
...
-        session = getattr(request, 'session', {})
-    except RuntimeError:
-        ctx = {}
-    else:
-        ctx = {
-            'tags': {
-                'database': session.get('db', None),
-            },
-            'user': {
-                'login': session.get('login', None),
```
moylop260, line 51: @fernandahf 
Could you check how to add the login and uid using new package, please?

I mean,
 - Screen Shot 2020-11-30 at 21 23 16

Now, it is not added

moylop260: FYI I just found the documentation for this:
 - https://docs.sentry.io/platforms/python/guides/django/enriching-events/identify-user/

sentry/__manifest__.py
fernandahf: Done

sentry/logutils.py
fernandahf: Done

sentry/__init__.py
fernandahf: Done

sentry/__manifest__.py
fernandahf: Done

fernandahf:
> Could you check if we can remove `raven` entirely as dependency, please?

sentry/logutils.py
fernandahf: Done

sentry/__init__.py
```diff
@@ -3,80 +3,117 @@
...
+
 def get_odoo_commit(odoo_dir):
-    '''Attempts to get Odoo git commit from :param:`odoo_dir`.'''
+    """Attempts to get Odoo git commit from :param:`odoo_dir`."""
     if not odoo_dir:
         return
     try:
-        return raven.fetch_git_sha(odoo_dir)
-    except raven.exceptions.InvalidGitRepository:
+        output = subprocess.check_output(['git', '-C', odoo_dir, 'describe', '--always']).strip()
```
hparfr: I don't think it's that easy.

Please have a look in the orignal implementation of raven : https://github.com/getsentry/raven-python/blob/d7d14f61b7fb425bcb15512f659626648c494f98/raven/versioning.py#L11


By the way, I don't think it's the responsibility of this module to take assumptions over how find the commit number. I suggest to remove this feature and take the commit identifier from config

sentry/const.py
```diff
@@ -44,46 +32,9 @@ def split_multiple(string, delimiter=',', strip_chars=None):
...
-        SentryOption('list_max_length', defaults.MAX_LENGTH_LIST, None),
-        SentryOption('site', None, None),
-        SentryOption('include_versions', True, None),
-        SentryOption(
-            'ignore_exceptions', DEFAULT_IGNORED_EXCEPTIONS, split_multiple),
-        SentryOption('processors', DEFAULT_PROCESSORS, split_multiple),
-        SentryOption('environment', None, None),
-        SentryOption('release', None, None),
-    ]
+DEFAULT_ENVIRONMENT = 'production'
```
hparfr: No.
A wrongly configured instance should appear as "production".

sentry/processor.py
```diff
@@ -0,0 +1,232 @@
+""" raven.core.processors
+    ~~~~~~~~~~~~~~~~~~~~~
+:copyright: (c) 2010-2012 by the Sentry Team, see AUTHORS for more details.
+:license: BSD, see LICENSE for more details.
```
hparfr: url for AUTHORS and LICENSE  ?
Please describe why this copy paste. Are the (new) filtering options not good enough ?

sentry/__init__.py
```diff
@@ -3,80 +3,117 @@
...
+    sentry_logging = LoggingIntegration(
         level=const.LOG_LEVEL_MAP[level],
+        event_level=logging.WARNING
     )
+
+    client = sentry_sdk.init(
+        dsn= config.get('sentry_dsn'),
+        environment=config.get('sentry_environment', const.DEFAULT_ENVIRONMENT),
+        release=release,
+        send_default_pii=True,
```
hparfr: Shouldn't it be False by default ?

hparfr (changes requested):
Thanks for this great and needed work.
Didn't had time to test it yet, some remarks from the code review.

sentry/__init__.py
fernandahf: Done

sentry/__init__.py
fernandahf: I'm agree with respect to the responsability about how find the commit number. I've changed this code like raven code.

sentry/const.py
fernandahf: I've changed 'production' by 'develop'.

sentry/__init__.py
fernandahf: Done

sentry/processor.py
fernandahf: Sentry-sdk does not have a processor funcionality like raven, I've looked for a solution and have found comments about take processor class of raven and to do a custom implementation:

https://github.com/getsentry/sentry-python/issues/211#issuecomment-633101708
https://forum.sentry.io/t/how-to-customize-event-in-before-send/7457

By the way, I've changed this code.

sentry/logutils.py
```diff
@@ -1,43 +1,68 @@
...
-    from raven.handlers.logging import SentryHandler
-    from raven.processors import SanitizePasswordsProcessor
-    from raven.utils.wsgi import get_environ, get_headers
-except ImportError:
-    _logger.debug('Cannot import "raven". Please make sure it is installed.')
-    SentryHandler = object
-    SanitizePasswordsProcessor = object
+
+
+def iteritems(d, **kw):
```
moylop260: this method exists 2 times

sentry/processor.py
```diff
@@ -0,0 +1,205 @@
...
+        'access_token',
+    ])
+    VALUES_RE = re.compile(r'^(?:\d[ -]*?){13,16}$')
+
+    @property
+    def sanitize_keys(self):
+        return self.KEYS
+
+    @property
+    def FIELDS(self):
```
moylop260: I think this method is not required

sentry/__init__.py
```diff
@@ -3,80 +3,118 @@
 
 import logging
 
+import odoo.http
```
moylop260: Move all logic from `__init__.py` file to [hooks.py](https://github.com/OCA/maintainer-tools/blob/master/template/module/hooks.py)

And import hooks like:
 - https://github.com/OCA/maintainer-tools/blob/f1b559574648c01558b543e7550f5b9be7c92784/template/module/__init__.py#L7

sentry/processor.py
fernandahf: Done

sentry/__init__.py
fernandahf: Done

sentry/logutils.py
fernandahf: Done

moylop260:
@barsi @naglis @versada Could you check it, please?

hoangtrann:
Any updates on this guys?

moylop260:
@hoangtrann 

We are waiting for approves

I mean, we need almost 2 approves to install and configure this module with this new change and do feedback or approve.

Fell free to test it, please

hoangtrann:
@moylop260 let me check this with one of my client's staging instance 😉

ryantran-novobi:
I have deployed and ran this PR successfully on Odoo instance running on Odoo.sh

ping @moylop260 @hparfr @fernandahf

kirtangajjar:
I tested this PR. It works fine on my local. I'm able to see the errors in sentry, however, I'm unable to see anything in the performance screen. Is it expected?

hoangtrann:
image

Not sure if this is relavant, but how could we enhance the presentation of model object in sentry, like `sale.order(124292,)` or `sale.order(124292, 124293)`

@fernandahf @moylop260

fernandahf:
@kirtangajjar

What is the performance screen? I searched but didn't find it.

Regards.

kirtangajjar:
@fernandahf Check the below screenshot.

Screenshot from 2021-04-20 11-02-41

Here are the link to official docs: 

https://docs.sentry.io/product/performance/
https://docs.sentry.io/product/performance/getting-started/

kirtangajjar:
@fernandahf Did you get time to look at it?

hparfr (approved):
Thanks

themreza:
What's the status here?

fernandahf:
@themreza

We are waiting for approves

I mean, we need almost 1 approve to install and configure this module with this new change and do feedback or approve.

Fell free to test it, please

themreza:
I tested the code in this branch on Odoo 14 and I can confirm that issues are being correctly sent to Sentry.

image

Although the performance feature is sending data, it's really vague and rather pointless.

image

I intentionally ran some excessively slow requests to see the trace, ideally starting from the HTTP request all the way to the database query or the compute methods responsible, but no details were included:

image

Also, there is no frontend tracking via JavaScript:

image

I'm not sure if it's a bug, but when I added `sentry_traces_sample_rate` to `odoo.conf`, the parsed value was `str` rather than `float`.

image

P.S.: It wasn't working at first which threw me off for a while, but I later figured out it was because I was running Odoo via PyCharm's debug mode and it didn't send anything.

themreza:
I figured out the frontend issue and performance tracking.

With a simple template it's possible to inject Sentry's JavaScript bundle:
```
<odoo>
    <template id="assets_backend" name="sentry_javascript_assets" inherit_id="web.assets_backend">
        <xpath expr="." position="inside">
            <script
                    src="https://browser.sentry-cdn.com/6.5.1/bundle.tracing.min.js"
                    integrity="sha384-TkhX8dCdk4BiV9tdHwvmIHWhdvw+LGo0SrMWTTBZ2T/sAUnxQWPuRiXELaaiZT6A"
                    crossorigin="anonymous">
            </script>
            <script>
                Sentry.init({
                  dsn: "YOUR_DSN",

                  // This enables automatic instrumentation (highly recommended), but is not
                  // necessary for purely manual usage
                  integrations: [new Sentry.Integrations.BrowserTracing()],

                  // To set a uniform sample rate
                  tracesSampleRate: 0.2
                });
            </script>
        </xpath>
    </template>
</odoo>
```

image

I believe if we enable `propagate_traces` option, we can link the server-sided traces to frontend ones.

But still, the backend traces are not offering a detailed breakdown of what the request consisted of.

ryantran-novobi:
@moylop260 would you mind taking a look at this one? I believe that all the discussions are resolved and it's waiting for your approval so far

moylop260:
@ryantran-novobi 

Let's do it

codeagencybe:
Is this issue already resolved and ready to use now on v14? 
I just downloaded the latest version and installed and it reports to me I need to upgrade the SDK. 
But this issue seems to have resolved that issue? 
Any feedback please?

fernandahf:
Hi @codeagencybe 

It's not in forward-ported yet, but there is one PR related to v14.0

https://github.com/OCA/server-tools/pull/1976

PR #1784, opened 2020-03-06
[IMP][12.0][auto_backup] Support pysftp==0.2.9 - Auto-add host to known_hosts, if it's the first time we connect

ivantodorovich:
Adds support for `pysftp==0.2.9`

With this PR, if it's the first time we connect to that server (meaning, not in known hosts list), we add it to the list.

pinging authors:
@Yenthe666 
@archetipo 
@lasley

PR #1947, opened 2020-12-07
[IMP] auto_backup: support sftp_public_host_key to add host in known hosts.

mathben:
The host key verification protects you from man-in-the-middle attacks. Can be generated with command 'ssh-keyscan -p PORT -H HOST/IP' and the right key is immediately after the words 'ssh-rsa'.

This is another solution with manual verification of the host of #1784


mathben:
I copy the idea from project https://github.com/it-projects-llc/odoo-saas-tools/tree/12.0/saas_server_backup_ftp

PR #1948, opened 2020-12-07
[MIG] base_import_odoo: Migration to v13

hbrunn:
migration of https://github.com/OCA/server-tools/tree/10.0/base_import_odoo


SkiBY:
I'd like to up this PR.
This module works and should be accepted 
[[ADD] possibility to keep passwords](https://github.com/OCA/server-tools/pull/1948/commits/3709fd1b510e6657c829ff8e89946fdfa3c249f9) this functionality looks to paranoidal as for me in this case, but it make addon more secure for sure

StefanWiselka-Solvti:
> I'd like to up this PR. This module works and should be accepted [[ADD] possibility to keep passwords](https://github.com/OCA/server-tools/pull/1948/commits/3709fd1b510e6657c829ff8e89946fdfa3c249f9) this functionality looks to paranoidal as for me in this case, but it make addon more secure for sure

There is an issue, which probably does not happen often.
Imported data gets external identifier from that addon: 'base_import_odoo'
so, when uninstalling 'base_import_odoo' cause removing all imported data because of external identifier from that addon.
This is probably unexpected, because usually a user would like to use that tool to import data and then clean-up after, by uninstalling this addon, which will remove all imported data as well.

hbrunn:
that's something I indeed didn't consider. Should this rather be "solved" by a sentence in the readme, or should we generate an xmlid with a fake module name, similar to exports creating xmlids with module `__export__`?

SkiBY:
__export__ is better as for meBR,Sergei Ruzki09.01.2023, в 15:19, Holger Brunn ***@***.***> написал(а):﻿
that's something I indeed didn't consider. Should this rather be "solved" by a sentence in the readme, or should we generate an xmlid with a fake module name, similar to exports creating xmlids with module __export__?

hbrunn:
I made it `__base_import_odoo__`, this way it's recognizable where the id comes from, and it doesn't interfere with exports

SkiBY (approved):

thomaspaulb:
@hbrunn To fix tests, rebase when https://github.com/OCA/server-tools/pull/2518 gets merged, which should be soon

thomaspaulb (approved):

base_import_odoo/views/import_odoo_database.xml
```diff
@@ -0,0 +1,155 @@
...
+                        fields: ['res_id'],
+                        domain: [
+                            ['module', '=', 'base_import_odoo'],
+                            ['model', '=', model],
+                            ['import_database_id', '=', database_id],
+                        ],
+                    })
+                    .then(function(data)
+                    {
+                        debugger;
```
thomaspaulb: Hey, what's this `debugger` still doing here?

base_import_odoo/views/import_odoo_database.xml
```diff
@@ -0,0 +1,155 @@
...
+                        name="import_field_mappings"
+                        attrs="{'readonly': [('cronjob_running', '=', True)]}"
+                    />
+                </sheet>
+            </form>
+        </field>
+    </record>
+    <template id="view_import_odoo_database_qweb">
+        <script type="text/javascript">
+            odoo.define('base_import_odoo', function(require) {
```
thomaspaulb, line 91: I haven't tested in 13.0, but in 14.0 in debug mode, I run into [Service base_import_odoo already defined](https://github.com/OCA/OCB/blob/14.0/addons/web/static/src/js/boot.js#L137). I think if you leave it empty, then it will just assign an [incremented job number](https://github.com/OCA/OCB/blob/14.0/addons/web/static/src/js/boot.js#L112).

thomaspaulb (changes requested):

base_import_odoo/views/import_odoo_database.xml
thomaspaulb: And, this should be changed to `__base_import_odoo__`.

For the above, [this refactor](https://github.com/OCA/server-tools/pull/2519/commits/37683a58f5fb85cd3db636f3a3e8a96c12884644) seems to have done the trick.

hbrunn: thanks

thomaspaulb:
@hbrunn Could you rebase?

thomaspaulb:
@hbrunn Hmm, tests are now failing for real:

```
2024-02-03 08:08:07,494 1097 ERROR odoo odoo.addons.base_import_odoo.tests.test_base_import_odoo: ERROR: TestBaseImportOdoo.test_base_import_odoo
  File "/usr/lib/python3.6/unittest/mock.py", line 1183, in patched
    return func(*args, **keywargs)
  File "/__w/server-tools/server-tools/base_import_odoo/tests/test_base_import_odoo.py", line 62, in test_base_import_odoo
    self.env.ref("base_import_odoo.demodb")._run_import()
  File "/__w/server-tools/server-tools/base_import_odoo/models/import_odoo_database.py", line 189, in _run_import
    self._run_import_model(context)
  File "/__w/server-tools/server-tools/base_import_odoo/models/import_odoo_database.py", line 234, in _run_import_model
    record = self._create_record(context, model, data)
  File "/__w/server-tools/server-tools/base_import_odoo/models/import_odoo_database.py", line 272, in _create_record
    ).create(record)
  File "<decorator-gen-104>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 317, in _model_create_single
    return create(self, arg)
  File "/opt/odoo/addons/mail/models/mail_alias.py", line 122, in create
    return super(Alias, self).create(vals)
  File "<decorator-gen-3>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 338, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/odoo/models.py", line 3825, in create
    records = self._create(data_list)
  File "/opt/odoo/odoo/models.py", line 3931, in _create
    cr.execute(query, params)
  File "/__w/server-tools/server-tools/slow_statement_logger/__init__.py", line 38, in execute
    return super().execute(query, params, log_exceptions)
  File "/opt/odoo/odoo/sql_db.py", line 173, in wrapper
  File "/opt/odoo/odoo/sql_db.py", line 250, in execute
psycopg2.IntegrityError: duplicate key value violates unique constraint "mail_alias_alias_unique"
DETAIL:  Key (alias_name)=() already exists.
```

hbrunn:
@thomaspaulb thanks, I cherry picked 9021e8b19de0e12802f70cd4c4afe539f98f7165 from the v14 migration

thomaspaulb (approved):

PR #1952, opened 2020-12-09
[MIG] 11.0 mig base_custom_info

fshah-initos:


rousseldenis:
@fshah-initos Could you check why Travis is not happy ?

rousseldenis:
@fshah-initos I think there is a problem with configuration_helper

fshah-initos:
We close this PR as we currently focus on the newer available version.

PR #1956, opened 2020-12-10
[11.0][MIG] base locale uom default

dsolanki-initos:


fshah-initos (approved):
LGTM

dsolanki-initos:
We abort the work because of newer available Odoo versions.

PR #1957, opened 2020-12-10
11.0 mig auth ldap populate

dsolanki-initos:


dsolanki-initos:
We abort the work because of newer available Odoo versions.

PR #1963, opened 2020-12-14
[8.0][IMP] auditlog: Add indexes for autovacuum speed

thomaspaulb:
So that autovacuum of HTTP requests and sessions does not require table-scanning a huge auditlog.log file

https://stackoverflow.com/questions/48793071/does-a-postgres-foreign-key-imply-an-index


rousseldenis (approved):
LGTM. Code review

StefanRijnhart (approved):

Issue #1249, opened 2018-05-19
auditlog.rule.write_full use PostgreSQL Trigger faster


moylop260:
Description:

Currently to use `log_type='full'` is so slow the process since that read fields from database and compute the m2o fields and computed ones 2 times (before and after write).

Saw:

SQL and odoo will runs
 - `SELECT ... FROM my_table WHERE id = %s`  # To get current values of all fields
 - `INSERT INTO my_table ... VALUES (...)`
 - Odoo process fields compute and relationals
 - `SELECT field1, field2, field3... field100 FROM my_table WHERE id = %s`  # To get new values of all fields
 - Odoo process fields compute and relationals

It's a huge process.

Expected:
 - Faster process using `log_type='full'` (I know that I could use `fast` method but I need both 'fast' and 'faster')

Extra:
PostgreSQL's trigger uses `OLD.*` and `NEW.*` internal variables to get the values of the current transaction before `UPDATE` and after respectively.

Using this way SQL and odoo will runs for `my_browse.write({'field': value})` the following sentence:
 - `INSERT audit_log field_changed VALUES ('value_changed');`
 - `UPDATE my_table SET field=value WHERE id =%d`
Faster like as current `faster` type option but with the persistence of `full` type but really *faster*

There are people working on similar way:
 - 9.0< https://wiki.postgresql.org/wiki/Audit_trigger
   - Good full proof of concept used of example:
      - ![screen shot 2018-05-19 at 17 18 20](https://user-images.githubusercontent.com/6644187/40273901-2cc6cdce-5b8f-11e8-9256-5e73e4706a35.png)
 - 9.1> https://wiki.postgresql.org/wiki/Audit_trigger_91plus
   - latest: https://github.com/2ndQuadrant/audit-trigger/blob/master/audit.sql

Is somebody working with a similar trigger?
What about change the module to this way?

thomaspaulb:
@moylop260 I think this is a very good idea.

I also run into slow behaviour but was thinking other solutions:

- Dont try to read non-stored or related computed fields on "full" logging

- Only read those fields that are referenced in the logging message as o['....'], leave the rest alone.

What do you think of those options?

moylop260:
I agree

PR #1964, opened 2020-12-15
[IMP] Dont query non-stored related or computed fields on full auditlog

thomaspaulb:
Because this can slow down the system a lot and gives no important info: related fields can be logged by an audit log on the related record itself, and computed fields are always dependent on other records/fields that we can log instead.

This is a quick fix for https://github.com/OCA/server-tools/issues/1249

A rewrite to using triggers would be even better IMO, but takes much more work.

### DIFF
--- a/auditlog/README.rst
+++ b/auditlog/README.rst
@@ -62,6 +62,7 @@ Contributors
 * Sebastien Alix <sebastien.alix@osiell.com>
 * Holger Brunn <hbrunn@therp.nl>
 * Pieter Paulussen <pieter.paulussen@me.com>
+* Tom Blauwendraat <tom@sunflowerweb.nl>
 
 Images
 ------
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
LGTM. Code review

thomaspaulb:
Thanks @rousseldenis

StefanRijnhart (approved):

StefanRijnhart:
Does ocabot do autosquash or do you need to squash commits?

thomaspaulb:
I don't know, I've squashed to be sure.

StefanRijnhart (approved):

thomaspaulb:
@hbrunn Please merge.

thomaspaulb:
Hey, where's the OCA bot?

PR #1965, opened 2020-12-15
[11.0][MIG] auto_create_qweb

dsolanki-initos:


dsolanki-initos:
We close this PR as we currently focus on the newer available version.

PR #1972, opened 2020-12-21
[11.0][FIX] Dont query non-stored related or computed fields on full auditlog

### LINKED ISSUE #1249: auditlog.rule.write_full use PostgreSQL Trigger faster

thomaspaulb:
Fixes #1249, forward port of https://github.com/OCA/server-tools/pull/1964

### DIFF
--- a/auditlog/README.rst
+++ b/auditlog/README.rst
@@ -88,6 +88,7 @@ Contributors
 * Sebastien Alix <sebastien.alix@camptocamp.com>
 * Holger Brunn <hbrunn@therp.nl>
 * Holden Rehg <holdenrehg@gmail.com>
+* Tom Blauwendraat <tom@sunflowerweb.nl>
 
 Other credits
 ~~~~~~~~~~~~~
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -192,6 +192,19 @@ class AuditlogRule(models.Model):
         self.unsubscribe()
         return super(AuditlogRule, self).unlink()
 
+    @api.model
+    def get_auditlog_fields(self, model):
+        """
+        Get the list of auditlog fields for a model
+        By default it is all stored fields only, but you can
+        override this.
+        """
+        return list(
+            n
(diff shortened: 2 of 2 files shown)


StefanRijnhart (approved):

hbrunn (approved):
/ocabot merge patch

thomaspaulb:
@hbrunn Is the OCA bot broken?

hbrunn:
glitch it seems

PR #1978, opened 2021-01-03
[12.0][FIX]database_cleanup: Don't purge by default

diggy128:
Modules should not be purged when completing scan without
explicit user consent. Instead select modules to purge from results.

After applied fixed an issue similar to #1534

### DIFF
--- a/database_cleanup/models/purge_modules.py
+++ b/database_cleanup/models/purge_modules.py
@@ -89,8 +89,6 @@ class CleanupPurgeWizardModule(models.TransientModel):
                 continue
             res.append((0, 0, {'name': module.name}))
 
-        purge_lines.purge()
-
         if not res:
             raise UserError(_('No modules found to purge'))
         return res


jado95 (approved):

thomaspaulb (approved):

StefanRijnhart (approved):

thomaspaulb:
@diggy128 Interestingly, after your fix [this test](https://github.com/OCA/server-tools/blob/c9b37cb6ccef3410dabe54ee4413626de0082833/database_cleanup/tests/test_database_cleanup.py#L130) now fails, because it actually tests for a module already being purged during `default_get`.

@hbrunn Blame for that test leads to you; do you have any idea why it's there?

StefanRijnhart:
Yep, looks like purging modules immediately was a conscious change according to the explanation in this PR: https://github.com/OCA/server-tools/pull/1316

hbrunn:
thanks @StefanRijnhart, that's it. But by now we have the purge all button, and starting in version 14 they threw out this behavior anyways during migration.

thomaspaulb:
So I modified tests as per discussion above, in https://github.com/OCA/server-tools/pull/2857

Closing this.

PR #1979, opened 2021-01-04
[MIG] scheduler_error_mailer: Migration to 14.0

chafique-delli:


scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,52 @@
...
+        string="Error E-mail Template",
+        help="Select the email template that will be sent when "
+        "this scheduler fails.",
+    )
+
+    @api.model
+    def _handle_callback_exception(
+        self, cron_name, server_action_id, job_id, job_exception
+    ):
+        res = super(IrCron, self)._handle_callback_exception(
```
bealdav: Thanks for porting it.
Please update super with py3 syntax

scheduler_error_mailer/tests/test_scheduler_error_mailer.py
```diff
@@ -0,0 +1,33 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.test_scheduler_error_mailer = cls.env.ref(
+            "scheduler_error_mailer.test_scheduler_error_mailer"
+        )
+
+    def test_scheduler_failure(self):
+        try:
+            self.test_scheduler_error_mailer.method_direct_trigger()
```
hailangvn, line 17: This line triggers the job directly and bypasses `_callback` of ir_cron. That is the reason why `_handle_callback_exception` was not fired while running test thus travis build failed. Other than this test, the `ir_cron` code does work.

fredzamoabg:
quote https://github.com/OCA/server-tools/pull/1979#discussion_r633441237 and can you also squash commit history? @chafique-delli

Issue #1984, opened 2021-01-08
[12.0] auditlog: Access error when creating full log


TimLai125:
Steps to reproduce:
1. Create an auditlog rule to `sale.order.line` with "Full Log" and log all the changes.
2. Create an user which only has the "Sale: User: Own Documents Only" group.
3. Login with the user and create a sale order with some order lines.

Current behavior:
Access error when user tries to save the sale order.
![image](https://user-images.githubusercontent.com/27185427/104016208-d4419200-51f0-11eb-8871-fb0c47c3a85f.png)

atchuthan:
PR merged, we could close this issue.

PR #1985, opened 2021-01-08
[12.0][FIX] auditlog: Add sudo() to retrieve record vals

TimLai125:
This PR fixes https://github.com/OCA/server-tools/issues/1984

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -235,7 +235,7 @@ class AuditlogRule(models.Model):
             # their values exist in cache.
             new_values = {}
             fields_list = rule_model.get_auditlog_fields(self)
-            for new_record in new_records:
+            for new_record in new_records.sudo():
                 new_values.setdefault(new_record.id, {})
                 for fname, field in new_record._fields.items():
                     if fname not in fields_list:


atchuthan (approved):
:+1:

yostashiro (approved):
Functional review. Looks good. 👍

marcelsavegnago (approved):

thomaspaulb:
@TimLai125 Can you forward port to other versions also?

thomaspaulb (approved):

StefanRijnhart:
@TimLai125 thanks! A port to later versions would be appreciated indeed.

/ocabot merge patch

TimLai125:
@StefanRijnhart OK, I will port this to other versions.

pedrobaeza:
Values retrieved with sudo can vary from those returned with regular permissions.

thomaspaulb:
@pedrobaeza Should [base_suspend_security](https://github.com/OCA/server-backend/pull/31) be used instead?

pedrobaeza:
I'm not sure if that's enough. Think for example in a computed field that returns the number of records of something. Depending on the record rules, the computed value will vary according your "record visibility" from the record rules.

StefanRijnhart:
@pedrobaeza you've got a point, but I don't see an easy solution for that. I think the impact on the validity of the data is limited, and having the bug is worse, so unless you've got a pointer to a way out, I think we can keep this change and port it to later versions.

pedrobaeza:
Yes, we have come to the same conclusion on the 13.0 PR, and as this only happens on full log, it's not the usual logging option (or I think so).

thomaspaulb:
The 'computed fields' example does not hold since non-stored computed fields are [now excluded from the auditlog](https://github.com/OCA/server-tools/pull/1975) for performance reasons (same PR was merged for 12 and 13 already)

pedrobaeza:
Great then! I don't think that there are more possible differences then.

PR #1993, opened 2021-01-15
[14.0][ADD] restrict_edition_to_superuser

kevinkhao:


kevinkhao:
@bealdav could be of interest

bealdav (approved):
LGTM

restrict_edition_to_superuser/models/base.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, _, fields, models
+from odoo.exceptions import AccessError
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    restrict_edition_to_superuser = fields.Boolean()
```
legalsylvain: This is adding a field on each model of your database ?
It risk to make this mosule very slow to install, and make the database bigger !

restrict_edition_to_superuser/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Set the restriction boolean to True on your record then enjoy peace
```
legalsylvain: I dont see how ? There is no code to add this field on the form view. Right ?

legalsylvain (commented):
Some question inline.

restrict_edition_to_superuser/models/base.py
rvalyi: an alternative is to have a mixin you would inject on objects you want to have this control, not sure which is better.

Also, I did a similar POC a few years ago, but instead of being based on a cloumn flag in the record, you would look at for an eventual record ir_model_data (can be OK for some classes you write or unlik very rarely). And you could then have policies like you cannot delete a record that will be brought back with an ir_model_data on module update (like you cannot dele or change a view or a menu...). Not sure it overlapps, just mentionning the idea.

kevinkhao: thx for feedback, in the end we go with a mixin

restrict_edition_to_superuser/readme/USAGE.rst
kevinkhao: That's right, this module has its own opinion on how it should be used, but the principle is that some records, you just don't want anyone touching them

restrict_edition_to_superuser/models/restrict_edition_mixin.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def check_access_rule(self, operation):
+        if operation in ("write", "unlink"):
+            for rec in self:
+                if (
+                    rec.restrict_edition_to_superuser
+                    and not self.env.user.id == SUPERUSER_ID
+                ):
+                    raise AccessError(_("Only the administrator can edit this record"))
+        super().check_access_rule(operation)
```
sebastienbeau:
```suggestion
        return super().check_access_rule(operation)
```

restrict_edition_to_superuser/readme/USAGE.rst
sebastienbeau: @legalsylvain the main aim it to flag record created with data (in your module) that should be only edited using data.

bealdav:
After further answering I don't know which is the use case targeted but if I one day need to prevent delete records I think it could be better to add xml data to the record and prevent in the code to delete any record with xml data.
An helper in the interface could be done to add xml data by superuser. 
If yes it can works with any records and any table without any other column in each table

What do you think ?

It's not a blocking remark of me

kevinkhao:
@bealdav I think there are many ways to skin a cat; this module is only one of them, and your idea seems valid to me. Other ways to effect this will overlap with this module's goals, but this one can be used as is.

PR #1995, opened 2021-01-18
[13.0][FIX] Add sudo() to retrieve record vals

TimLai125:
Forward Port to 13.0 of https://github.com/OCA/server-tools/pull/1985

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -261,7 +261,7 @@ class AuditlogRule(models.Model):
             # their values exist in cache.
             new_values = {}
             fields_list = rule_model.get_auditlog_fields(self)
-            for new_record in new_records:
+            for new_record in new_records.sudo():
                 new_values.setdefault(new_record.id, {})
                 for fname, field in new_record._fields.items():
                     if fname not in fields_list:


len-foss (approved):

marcelsavegnago (approved):

pedrobaeza:
Values retrieved with sudo can vary from those returned with regular permissions.

len-foss:
@pedrobaeza If it was deemed necessary in 12.0 it certainly is in 13.0 (the FP was explicitly asked for)? 
Furthermore IIRC the sudo refactoring was in 12.3, so the impact should be less important than in 12.0
In the worst case, that would pollute the audit.log, which is a bad thing, but the issue would be uncovered after a debugging session. Whereas without the fix you risk to make standard flows explode because of the logging.

pedrobaeza:
See my comment for https://github.com/OCA/server-tools/pull/1985#issuecomment-762331958

I don't know if it can affect and I'm not a user of this module, but I'm just throwing out the comment for you to know about it and take it into account if applicable.

len-foss:
@pedrobaeza I see, thank you. You make a point, but since the risk/reward is asymmetrical if think it's good enough as long as nobody has time to devote to find and fix these tricky corner case(s). However it would warrant adding a todo comment to mention it.

pedrobaeza:
If the problem is with the full log, then that sudo shouldn't be put only on that mode?

len-foss:
It's in the create_full method...

pedrobaeza:
Ah, OK, sorry for the noise then.

StefanRijnhart:
~~True, a note in the USAGE would be nice.~~ Computed fields are not reflected in the auditlogs as per https://github.com/OCA/server-tools/pull/1985#issuecomment-762420934

thomaspaulb (approved):

StefanRijnhart (approved):

PR #1996, opened 2021-01-18
[IMP][8.0] auto_backup: Delete incomplete backups to avoid filling up disk

thomaspaulb:
Depends on https://github.com/OCA/server-tools/pull/1623

If a backup fails for whatever reason, delete it and dont keep the incomplete file on disk. If incomplete files are kept, the disk will start filling up after some days.

### DIFF
--- a/auto_backup/models/db_backup.py
+++ b/auto_backup/models/db_backup.py
@@ -161,7 +161,12 @@ class DbBackup(models.Model):
                     # Generate new backup
                     else:
                         with rec.custom_tempdir():
-                            db.dump_db(self.env.cr.dbname, destiny)
+                            try:
+                                db.dump_db(self.env.cr.dbname, destiny)
+                            except:  # noqa
+                                # Avoid filling up disk with failed backups
+                                os.remove(destiny)
+                                raise
                         backup = backup or destiny.name
                 successful |= rec
 
--- a/auto_backup/README.rst
+++ b/auto_backup/README.rst
@@ -89,6 +89,7 @@ Contributors
 * Yenthe Van Ginneken <yenthe.vanginneken@vanroey.be>
 * Alessio Gerace <alessio.gerace@agilebg.com>
 * Jairo Llopis <yajo.sk8@gmail.com>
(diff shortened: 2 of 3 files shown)


thomaspaulb:
@hbrunn Can you review this? I added a test case also.

auto_backup/models/db_backup.py
```diff
@@ -161,7 +161,12 @@ def action_backup(self):
                     # Generate new backup
                     else:
                         with rec.custom_tempdir():
-                            db.dump_db(self.env.cr.dbname, destiny)
+                            try:
+                                db.dump_db(self.env.cr.dbname, destiny)
+                            except:  # noqa
+                                # Avoid filling up disk with failed backups
+                                os.remove(destiny)
```
hbrunn, line 168: I'm musing if half of a dump couldn't be a live saver if that's the only backup there is. What's your take on this?

thomaspaulb: I see that side of the story, but I also see the following case:

1. User has configured 3 full backups to be held at all times
2. Healthy backups are taken Monday, Tuesday and Wednesday
3. Thursday, Friday, Saturday and Sunday the backup is interrupted, therefore incomplete
4. On Monday the issue is seen and resolved, the backup runs fully, and the older backups are deleted. Saturday, Sunday and Monday are kept, but only Monday is complete
5. Tuesday people want to restore a file from before Monday morning Lucy the secretary deleted some invoices, but they can't, because they only have corrupted backups from before Monday

But I also see your point.

A sensible approach here could maybe be that we rename incomplete backups to "_incomplete". We don't count them in the retention of backups, and we keep only the most recent one.

What do you think?

hbrunn (approved):

auto_backup/models/db_backup.py
hbrunn: hmm, that's also non-obvious behavior. Given we send a mail for failures, users should be aware that things are wrong.

You'll need to rebase now #1623 is merged.

PR #1999, opened 2021-01-25
[MIG] module_analysis: Backport to 10.0

santostelmo:
`module_analysis` migration to version 10


legalsylvain (changes requested):
Hi.

Thanks for porting this module.
You should respect history, using standard OCA procedure. It makes also review easier to do, because reviewers only have to read changes between the two versions.

Here is the doc.
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

kind regards.

santostelmo:
> Hi.
> 
> Thanks for porting this module.
> You should respect history, using standard OCA procedure. It makes also review easier to do, because reviewers only have to read changes between the two versions.
> 
> Here is the doc.
> https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
> 
> kind regards.

Hi,

In this case I don't have the history because module was given to me in a file already. I just created 10 branch and copied the module.

Best regards

legalsylvain:
> In this case I don't have the history because module was given to me in a file already. I just created 10 branch and copied the module.

OK, but you should anyway follow the procedure, and then apply your changes.

santostelmo:
Ok, but in this case I don't have the module in version 9 to keep the history.
`git format-patch --keep-subject --stdout origin/10.0..origin/9.0 -- module_analysis | git am -3 --keep`

What you mean by follow the procedure in this case?

legalsylvain:
Hi. Replace 9.0 by 8.0 should work.

santostelmo:
Hi @legalsylvain,

This is a backported module, so don't see how I can have a history in this case.

legalsylvain:
Hi, replace 9.0 by 12.0 :-)

PR #2000, opened 2021-01-25
[MIG] module_analysis: Migration to 9.0

santostelmo:
`module_analysis` version 9.0


legalsylvain (changes requested):
Could you respect history, cherry picking commits ?

Otherwise thanks for porting this module.

PR #1987, opened 2021-01-08
[MIG][14.0] attachment_queue: Migration to 14.0

kevinkhao:

PR #2001, opened 2021-01-25
[14.0][MIG] attachment_synchronize: migration to 14.0

kevinkhao:
~depends on: #1987~


bealdav:
Many of the concepts of these modules are the same that those in https://github.com/OCA/edi/tree/13.0/edi

and https://github.com/OCA/edi/tree/13.0/edi_storage

Edi seems more abstract (including web services + files) but cover similar concepts (flow direction, state of the exchange, link with storage).

It's require a deeper analyze before to have several ways to do the same thing.

If not, we risk in complex projects to have edi to achieve UBL communication and attachment_synchronize to exchange bank files. It could be really confusing for users

Thanks

attachment_queue/models/attachment_queue.py
```diff
@@ -0,0 +1,105 @@
...
+                with registry(self.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(new_cr, self.env.uid, self.env.context)
+                    attach = attachment.with_env(new_env)
+                    try:
+                        attach._run()
+                    # pylint: disable=broad-except
+                    except Exception as e:
+                        attach.env.cr.rollback()
+                        _logger.exception(str(e))
+                        attach.write({"state": "failed", "state_message": str(e)})
```
bealdav, line 83: RFC Off topic of the migration, concern improvement about retry run task
If state_message is `could not serialize access due to concurrent update` could we consider to let state to 'pending' ?
Then there'll be a retry to the next execution job without do nothing else.

I think it could worth

cc @sebastienbeau @florian-dacosta @clementmbr @bguillot @rvalyi @kevinkao

attachment_queue/models/attachment_queue.py
```diff
@@ -0,0 +1,105 @@
...
+                        emails = attach.failure_emails
+                        if emails:
+                            failure_tmpl.send_mail(attach.id)
```
bealdav, line 86: The same here. No email should be sent if state == "pending"

bealdav (commented):
Please reply to my RFC

attachment_queue/models/attachment_queue.py
florian-dacosta: Yes I guess it could be nice. But then, it would be nice to have a max retry number, it could be nice to set a delay before retry, etc... 
All of this is already manage by queue job, I wonder if it is worse it to try to re-implement all of this here.

Did you already get a lot of case of conccurent update using this module in production ?
In my cases, I don't think I do.

Working to integrate the jobs here could be better than reinvent the wheel, also because other queue job features may appears as nice in the future (the priority management for instance)...but it is not as simple as it may seem.

bealdav: Thanks a lot for your feedback. In my experience with EDI I have some examples.
Converge with jobs is an option, yes.

bealdav: Pros : all is already in queue.job

Cons : security with jobs doesn't match with end users right

attachment_queue/models/attachment_queue.py
kevinkhao: I think logically the attachment must be in either success or fail

attachment_queue/models/attachment_queue.py
kevinkhao: integrating queue.job into this would mean rethinking the module structure
It would be interesting to know what the context is to trigger "could not serialize access due to concurrent update" error

bealdav: Finally, I think that in case queue is installed, a job could be created to manage the task execution. No need to have a dependency on it. What do you think ?

kevinkhao: How about another module attachment_synchronize_async ?

florian-dacosta (approved):

kevinkhao:
I added the superuser for running the attachment queue.

The point is that there should not be a different result depending on whether the attachment.queue is run through a cron or a user clicking on "run".

attachment_queue/models/attachment_queue.py
```diff
@@ -0,0 +1,106 @@
...
+        string="Failure Emails",
+        help="Comma-separated list of email addresses to be notified in case of"
+        "failure",
+    )
+
+    def _compute_failure_emails(self):
+        for attach in self:
+            attach.failure_emails = attach._get_failure_emails()
+
+    def _get_failure_emails(self):
```
GSLabIt, line 46: Could be sent to administrator by default if not other emails set?

florian-dacosta: I think the email notification should stay optional.
For some uncritical attachment, it is possible that we don't want to be polluted by an alert.

Also, one could decide to check the failed attachments once a week for instance and not use the mail alert, etc...
I believe it is easy enough to configure this (on storage backend level) or overridable.

And in a lot of project, the mail behind the admin user is not the one which will check these cases first

GSLabIt (approved):
LGTM

sebastienbeau (approved):
LGTM tested

PR #2002, opened 2021-01-26
[14.0][FIX] only play onchange on editable field

sebastienbeau:
solve issue when have limited access right (do not try to read none accesible field)
solve perf issue by avoiding to compute computed readonly field

TODO
- [ ] write test case

### DIFF
--- a/onchange_helper/models/base.py
+++ b/onchange_helper/models/base.py
@@ -20,6 +20,17 @@ class Base(models.AbstractModel):
                 new_values[fieldname] = value
         return new_values
 
+    def _check_field_is_editable(self, field):
+        if field.readonly:
+            return False
+        elif field._related_comodel_name:
+            return self.env["ir.model.access"].check(
+                field._related_comodel_name, "write", raise_exception=False
+            )
+        elif field.groups:
+            return self.user_has_groups(field.groups)
+        return True
+
     @api.model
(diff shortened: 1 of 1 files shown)


sebastienbeau:
@PierrickBrun

legalsylvain (commented):
Hi sebastien. Thanks for this patch. I wonder if it s a good idea to remove fields if user doesnt have correct access write. I fear that it let data in inconsitent state. 
Could you provide valid use case when it is ok to do that ?
Thanks.

PierrickBrun (commented):
I agree with @legalsylvain , apart from that, that looks very good ! Is the performance boost noticeable ?

sebastienbeau:
Hi I will be back soon on this point with test and real case.
My initial issue it that I am using a user for an API with restricted access.
When I play the onchange on partner it will compute field like "rma_ids", stock information... so at this end I request access to a lot of object.
Do not merge yet, it's quiet experimental

legalsylvain:
Hi. Well, logged with your limited user, can you create an item (without ``onchange_helper`` installed) ? 
If not, you should not have the possibility to do it with the module installed.
At first glance, If an onchange requires access right, you should give the access to your user.

alexis-via (approved):

PR #2004, opened 2021-01-28
[13.0][ADD] sequence_python module

thomaspaulb:


thomaspaulb:
@gjotten @acsonefho Gently asking for a review.

sequence_python/README.rst
```diff
@@ -0,0 +1,113 @@
...
+   # To separate the Odoo-generated number with hyphens eg. 0-0-0-0-1
+   '-'.join(number_padded)
+
+   # To have an UUID as the sequence value
+   uuid.uuid4().hex
+
+   # To use an 8-digit binary number
+   {0:#010b}'.format(number)
+
+The possibilities are nearly endless.
```
legalsylvain: Is it possible to use date and date_range as regular sequence ?
Could you add a note regarding that point.

sequence_python/models/ir_sequence.py
```diff
@@ -0,0 +1,79 @@
...
+    but can also be used for any other purpose.
+    """
+
+    _inherit = "ir.sequence"
+
+    # Python code
+    use_python_code = fields.Boolean(string="Use Python", default=False)
+    python_code = fields.Text(
+        string="Python expression",
+        groups="base.group_system",
```
legalsylvain: I think this geoup is unnecessary. Users that can create sequence should use this feature

legalsylvain (commented):
Some minor remarks inline. 
Lgtm otherwise. Thanks !

sequence_python/models/ir_sequence.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _get_python_eval_context(self, number_next):
+        """
+        Get the eval context to evaluate the Python code with.
+        The format is (variable name, description, value)
+        You can inherit this in your custom module.
+        :return: tuple
+        """
+        return {
+            "number": number_next[0],
```
gjotten: Getting `'int' object is not subscriptable` here coming out of a no_gap sequence with python_code `uuid.uuid4().hex`

gjotten (changes requested):
My hunch is we're not handling no_gap implementations very well here right now

sequence_python/README.rst
gjotten: @legalsylvain @thomaspaulb are we talking about prefix and suffix here? AFAICS only the number itself is changed so on a standard 13.0 Bank Journal sequence your next number is previewed as 

> BNK1/2021/57be06c0b96045b09f066e4fbefa59ae

The 8-digit binary number-suggestion nets me `unexpected EOF while parsing (, line 1)` though.

gjotten:
@thomaspaulb looks like it's not immediately failing on no_gap in my earlier example so that's good.

sequence_python/views/ir_sequence.xml
```diff
@@ -0,0 +1,52 @@
...
+            <xpath expr="//notebook" position="inside">
+                <page name="code" string="Python Code">
+                    <group>
+                        <field name="use_python_code" />
+                        <field
+                            name="python_code"
+                            widget="ace"
+                            options="{'mode': 'python'}"
+                            placeholder="number"
+                            attrs="{'invisible': [('use_python_code', '!=', False)]}"
```
gjotten: @thomaspaulb functionally this feels the wrong way around; I have to uncheck 'use python' to be able to see my expression and its preview.

sequence_python/README.rst
thomaspaulb: The 8-bit example was wrong, I changed it and fixed two other bugs as well.

thomaspaulb: @legalsylvain The date stuff as you can put in the prefix and suffix just works:

image

image

Is that what you mean?

sequence_python/views/ir_sequence.xml
thomaspaulb: You were right, I fixed it.

sequence_python/README.rst
legalsylvain: Sorry, i did a mistake during the review !
Thanks for this module.

gjotten (approved):
Functionally, this now does all the tricks for me AFAICS.

sequence_python/README.rst
thomaspaulb: @legalsylvain No problem, thanks for the review! Could you flip it to 'approved' ?

acsonefho (approved):
Code LGTM

thomaspaulb:
Squashed. @hbrunn Can you merge this.

PR #2012, opened 2021-02-05
[13.0] base_jsonify: misc fix/imp

simahawk:
See atomic commits.

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -133,3 +133,15 @@ class Base(models.AbstractModel):
 
         results = resolver.resolve(results, self) if resolver else results
         return results[0] if one else results
+
+    def _jsonify_m2o_to_id(self, fname):
+        """Helper to get an ID only from a m2o field.
+
+        Example:
+
+            <field name="name">m2o_id</field>
+            <field name="target">m2o_id:rel_id</field>
+            <field name="instance_method_name">_jsonify_m2o_to_id</field>
+
+        """
+        return self[fname].id
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -27,7 +27,7 @@ class Base(models.AbstractModel):
 
     def _function_value(self, record, function, field_name):
         if function in dir(record):
-            method = getattr(self, function, None)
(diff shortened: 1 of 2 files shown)


base_jsonify/models/models.py
```diff
@@ -133,3 +133,15 @@ def jsonify(self, parser, one=False):
 
         results = resolver.resolve(results, self) if resolver else results
         return results[0] if one else results
+
+    def _jsonify_m2o_to_id(self, fname):
```
simahawk, line 137: @len-foss @lmignon this addresses one of the most common cases (at least in my experience): have the stupid ID only.
It would be nice to support `mapped` in the future.
One possibility would be to set `instance_method_name` as `mapped:dotted.path.to.value` and then handle this particular case in `_function_value`. For such simple cases I think defining a custom resolver seems too much.
What do you think?

len-foss: @simahawk I had the same request yesterday, so that would definitely help. The only thing is that if using it with an external API, it is also very likely to expect another identifier. E.g. you have "model.instance" with a many2one to "model.type", you probably want the type name in the json. Or in case of the lang, you'd expect the land code...

Could we use the key_marshaller? That would give relational_id:target=field.

lmignon (approved):
LGTM (code review)

base_jsonify/models/models.py
len-foss: The problem with the mapped support is 'when do you want a list'. If you put mapped:name on a many2one chain, you'd expect a single result, not a list. Which includes the particular case mapped:relational_id.id you just gave.

len-foss (approved):
If that covers most existing use-cases, my comments can be ignored :-)

len-foss (approved):

base_jsonify/models/models.py
simahawk: > Could we use the key_marshaller? That would give relational_id:target=field.

Yeah, I played a bit w/ it but it seems the wrong place to me. That should be the place for "post handling" after you've already got the value and you want to store it in a different way in the resulting dict.

PR #2014, opened 2021-02-07
[11.0][FIX] profiler module fixes

thomaspaulb:
- Exceptions raised within a profiled request were not dealt with and reraised properly at request end

- After exception was raised in the request we might not have a db cursor anymore, in that case save stats to file and log where the file is, so that you can still profile

- Postgresql profiling (for now) only used in 'full' mode


thomaspaulb:
@sbidoul I changed `dns` to `dnspython`, and it seems to install fine, yet is [not found by Odoo](https://runboat.odoo-community.org/api/v1/builds/b60fd2e87-f1c4-40e8-a182-788a8dc9b6ae/init-log). What am I doing wrong?

sbidoul:
`dns` was fine. What's needed is an external_dependencies_override in setup.py. See setuptools-odoo docs for an example. Fortunately this kind of thing has become unnecessary in Odoo 12+ (or 13 not sure).

thomaspaulb:
Are you sure that `dns` is fine?

```
(.venv) $ python --version
Python 3.5.10
(.venv) $ pip install dns
ERROR: Could not find a version that satisfies the requirement dns
ERROR: No matching distribution found for dns
(.venv) $ pip install dnspython
Collecting dnspython
  Downloading dnspython-1.16.0-py2.py3-none-any.whl (188 kB)
     |████████████████████████████████| 188 kB 5.4 MB/s 
Installing collected packages: dnspython
Successfully installed dnspython-1.16.0
```

sbidoul:
Sorry, I mean dns is fine in he manifest external_dependencies. But it must be mapped to dnspython in setup.py.

thomaspaulb:
Thanks!

setup/letsencrypt/setup.py
```diff
@@ -2,5 +2,14 @@
...
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'acme': 'acme',
+                'cryptography': 'cryptography',
+                'dns': 'dnspython==1.16.0',
+                'josepy': 'josepy',
+            }
```
sbidoul, line 12: Only dns is necessary here. Also, don't pin version unless necessary. Or use a lower bound.

sbidoul (commented):
I realize I had written but not submitted this comment.

Issue #2023, opened 2021-02-16
[10.0] auth_brute_force: incompatible with OAuth


TimLai125:
**Steps to reproduce:**
1. Set up and allow Google OAuth2
2. Reset user's password with Google login
3. Use the "Log in with Google" to login

**Current behavior:**
Even for a successful login, the "Authentication Attempts" record shows "Failed"
Users who use "Log in with Google" to login, users can even log in even the user and IP address enter "Banned" state.

**Expected behavior:**
The "Authentication Attempts" record should show "Successful", the user should not be able to log in once he/she is banned.

**Video/Screenshot link (optional):**
![image](https://user-images.githubusercontent.com/27185427/108067040-26e95600-709b-11eb-8cff-369dd8d58ad5.png)

TimLai125:
@Yajo @legalsylvain any comments on this issue?

legalsylvain:
Sorry. I dont use google oauth.

yajo:
Sorry, I also don't use this module anymore... nor Odoo v10.

PR #2024, opened 2021-02-17
[10.0][ADD] auth_brute_force_oauth

TimLai125:
This PR adds `auth_brute_force_oauth` as a bridge module between `auth_brute_force` and `auth_oauth`.
The module fixes the "Authentication Attempts" records from Oauth login, the records will be showing "Successful" with Oauth login.

Since `auth_brute_force` and `auth_oauth` are independent to each other. The execution order of their `check_credentials` could vary in different environments.

Therefore this module will conbime the `check_credentials` from both modules:
1. First the logic should check against whether the remote is banned.
2. If the user is not banned, check if the uid and password pair match any exist user's Oauth access token.
3. Lastly, the logic should check against user's uid and password.

The `_auth_oauth_signin` needs to be further adjusted as to raise `AccessDenied` during the Oauth login flow. Otherwise, there will not be any error message shown in the screen when the user is banned.

Related Issue: https://github.com/OCA/server-tools/issues/2023


auth_brute_force_oauth/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    'author': "Quartile Limited, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://github.com/OCA/server-tools',
+    'license': 'AGPL-3',
+    'depends': [
+        "auth_oauth",
+        "auth_brute_force",
+    ],
+    'data': [],
+    'installable': True,
```
legalsylvain, line 17: if it is fixing an incompatibility between both module, it should be ``auto_install: True``.

legalsylvain (commented):
partial review.

yostashiro (commented):
Would it be an option to only override `check_credentials()` and keep successful OAuth logins outside of the scope of logging, by inserting
```.py
        res = self.sudo().search([('id', '=', self.env.uid), ('oauth_access_token', '=', password)])
        if not res:
```
at the beginning?

I personally think logging of successful attempts is not important for OAuth logins (much less vulnerable to brute-force attacks than password logins) and we should still be able to capture failed attempts if I'm not mistaken.

legalsylvain (commented):
sorry I don't use oauth, so I can not review this PR.

thanks for this contribution anyway.

auth_brute_force_oauth/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Authentification - Brute-Force Filter - (Oauth Support)',
```
yostashiro:
```suggestion
    'name': 'Authentification - Brute-Force Filter (Oauth Support)',
```

auth_brute_force_oauth/models/res_users_hook_check_credentials.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+        if not trusted:
+            error = AccessDenied()
+            error.reason = "banned"
+            raise error
+        res = self.sudo().search([
+            ('id', '=', self.env.uid),
+            ('oauth_access_token', '=', password)
+        ])
+        if res:
```
yostashiro:
```suggestion
        if self.sudo().search([
            ('id', '=', self.env.uid),
            ('oauth_access_token', '=', password)
        ]):
```

yostashiro (approved):
Functional review. LGTM. 👍  Just some minor non-blocking comments.

Issue #944, opened 2017-08-21
[RFC] new api.groups decorator for function


legalsylvain:
Hi all,

I think about a new decorator like 
```
@api.groups('group_1', 'group_2', '...')
def function_name(self, args):
    ...
```
 to decorate function that are call by function in a xml file like `<button name='function_name' />`.

The decorator will check if current user is a member of one of the defined groups, if not an access error is raised. Then, in the render view, the button is hidden if the user is not a member of the defined groups automatically

That new decorator could fix security breaches that allow user to call a function of a hidden button, by xml RPC.

In a second time we could imagine other decorators @api.add_groups or @api.remove_groups to add or remove access to a function, in a custom module that inherit a model.

Thanks for your comment.

lasley:
I think this is a good idea - security by obscurity (hiding the buttons) doesn't work.

Question is where to put this. I made an [OCA Python module](https://github.com/LasLabs/python-oca) that has an `api.foreach` implementation in it, but there [wasn't much movement on the getting it into OCA front](https://odoo-community.org/groups/contributors-15/contributors-53712?mode=thread&date_begin=&date_end=)

pedrobaeza:
I still think it's a good idea to have an OCA API library. @lasley you only need to create the repo following the method put in the OCA instance task ("Create a project") and make the PR.

About this feature, :+1: for adding it, but not with the decorators `add_groups` and so on. This should be handled with the inheritance: if you add any new group in the decorator on an overwritten method, then that group is added to the list. About the removal of groups, then you should use other techniques like in core (for example, another method with less groups that calls the original method).

legalsylvain:
Hi ! Thanks for your answer. A OCA API library LGTM. Great idea ! 

> but not with the decorators add_groups and so on

You're right, it could be simplified indeed.

> About the removal of groups, then you should use other techniques like in core (for example, another method with less groups that calls the original method).

Not sure it will cover all the needs, but we'll talk about it in a specific PR against the new repo.

> @lasley you only need to create the repo following the method put in the OCA instance task ("Create a project") and make the PR.

@lasley , please ping me when the repo is created, I'll began a POC then and try to finish the work in the next Open (/ Closed) Days.

lasley:
> I still think it's a good idea to have an OCA API library. @lasley you only need to create the repo following the method put in the OCA instance task ("Create a project") and make the PR.

[...]

PR #2026, opened 2021-02-19
[12.0][add] New decorator @api.allowed_groups

legalsylvain:
Hi all,

The recent PR made by @StefanRijnhart here https://github.com/odoo/odoo/pull/66505 made me think of an old idea that we talked about here https://github.com/OCA/server-tools/issues/944 and here https://github.com/OCA/oca-decorators/issues/7.

I so made a little PoC, and it works quite well.

@OCA Members :  What do you think ? CC : @lasley, @pedrobaeza, @NL66278, @lmignon, @sbidoul 

@odoo Team : Any chance to see this feature integrated in the Odoo core directly ? I don't know who ping in the team. @Yenthe666 any idea ? 

**Note**
I finally made my PoC using a module design and not integrating the code in the current oca lib https://github.com/OCA/oca-decorators because I was more comfortable with this type of design, and that the reasons for using a library seems obsolete in recent versions of Odoo.

### Description
This module is a technical module for developpers.

It adds a new decorator, named ``@api.allowed_groups``.

- When the function is executed, it checks if the user belong to one of the given groups.

- It also adds automatically group(s) in the according form views to hide
  buttons if the user doesn't belong to one of the given groups.

#### Interests
1) it makes the application more secure, the developer only has to write the accreditation level in one place, instead of writing it twice (once in the XML view, and the other time in the python code)

In Odoo, there are many places where an action is hidden in the Form view, but the function can be called in particular by XML-RPC calls, that makes a lot a security issue, included
in recent version.

Ref : https://github.com/odoo/odoo/pull/66505

2) It makes the code more easy to maintain, because there is only one place
   where the groups are defined instead two places without this module (XML and Python code)

### Usage
Once installed, the following code:

    <button type="object" name="my_action" groups="purchase.group_purchase_manager"/>

    def my_action(self):
        if not self.env.user.has_group("purchase.group_purchase_manager"):
            raise AccessError(_("Some Error"))
        pass

can be replaced by:

    <button type="object" name="my_action"/>

    @api.allowed_groups("purchase.group_purchase_manager")
    def my_action(self):
        pass

Multi-group membership checks
-----------------------------

It is possible to list many groups. In that case, the action will be allowed
if the user belong to at least one group.

    @api.allowed_groups("purchase.group_purchase_manager", "sale.group_sale_manager")
    def my_action(self):
        pass

Conflict between view and decorators definition
-----------------------------------------------

The groups defined in the decorators take precedence over the groups that would be defined in the existing views.

Inheritance mechanisms
----------------------

it is possible to change accreditation level in custom module.

For exemple, if a module A defines a function like this:

```
    @api.allowed_groups("purchase.group_purchase_manager")
    def my_action(self):
        pass
```

In a custom module B, that depends on module A :

1) You can overwrite accreditation level by writing

```
    # Now checks if user is member of 'purchase.group_purchase_user'
    @api.allowed_groups("purchase.group_purchase_user")
    def my_action(self):
        return super().my_action()
```

2) You can remove checkes, by writing

```
    # No longer performs checks
    @api.allowed_groups()
    def my_action(self):
        return super().my_action()
```

3) Or you can only overload the function, without changing the accreditation level by writing

```
    # the user must always be a member of 'purchase.group_purchase_manager'
    def my_action(self):
        # Custom code ...
        return super().my_action()
```


Yenthe666:
Hmm it is an interesting idea. Did you check performance wise what the cost is?
If you'd like to discuss it with Odoo I bet we need @mart-e, @feyensv and possible @odony their feedback/thoughts :)

decorator_allowed_groups/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2021-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': "Decorator - Allowed Groups",
+    'summary': """
```
NL66278: The summary should be one line. Move the description to the README.

decorator_allowed_groups/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2021-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
NL66278: Put https:// on  the license lines.

lmignon:
> Note
I finally made my PoC using a module design and not integrating the code in the current oca lib https://github.com/OCA/oca-decorators because I was more comfortable with this type of design, and that the reasons for using a library seems obsolete in recent versions of Odoo.

IMO providing this functionality through a module is not appropriate for this kind of feature. It's not a per database functionality and once the addon is into the path (I think it's not required to install-it), the decorator is available for the whole runtime.

NL66278:
Seems very interesting. Would help a lot to make functions more secure in an easy way. Some tests would definitely be helpful. I do not know how to evaluate the performance impact, except by testing it in practice. But certainly for some methods, checking the groups has to be done anyway.

hparfr:
groups= in view should only be reserved to hide elements you have right to see / activate. And this PR make this possible so :+1: 

I have concerns about how it behave when :  
- somewhere in the inheritance chain the decorator is missing
- some groups are added (limit rights)
- some groups are added (increase rights)

decorator_allowed_groups/models/decorator.py
```diff
@@ -0,0 +1,53 @@
...
+def allowed_groups(*group_xml_ids):
+    """ Return a decorator that specifies group(s)
+        to which the user must belong in order to perform
+        this function.
+        - if the user does not belong to any group,
+        an AccessError will be raised if the function is called.
+        - Also in the generated views, the according button
+        will be hidden if the user doesn't belong to the group(s).
+
+            @api.allowed_groups(
```
houssine78, line 18: @api.groups should be enough and is aligned with the current declaration in the xml

legalsylvain:
Hi @lmignon.

Regarding the lib vs module design.

1) security issue :

> once the addon is into the path (I think it's not required to install it), the decorator is available for the whole runtime.

You was right, but this point is obsolete in recent Odoo version. The code of the module ``decorator_allowed_groups`` is executed only if the module ``decorator_allowed_groups`` is installed. (or to be exact, present in the ``server_wide_modules`` list).
You can test this point with this test PR https://github.com/legalsylvain/server-tools/pull/8

- as you can see, the module ``decorator_allowed_groups_for_test_only`` depends only on ``purchase`` module, and if you try to install it, you'll have an error.

```
    File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/__init__.py", line 1, in <module>
        from . import purchase_order
    File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/purchase_order.py", line 8, in <module>
        class PurchaseOrder(models.Model):
    File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/purchase_order.py", line 11, in PurchaseOrder
        @api.allowed_groups("base.group_multi_company")
    AttributeError: module 'odoo.api' has no attribute 'allowed_groups'
```

- if you manuallly add a dependency to ``decorator_allowed_groups`` module, it will works. 

You can read a more detailled explanation here : https://github.com/OCA/OpenUpgrade/issues/2440#issuecomment-722513002

2) Also I see it more difficult to maintain in a library that doesn't depends on the version of Odoo. This module overload the function ``ir.ui.view``. the ``postprocess`` function can change over the version, making the code harder to maintain with some ``if version > 12.0 and version <= 14.0`` ...

3) It has to be said that the project ``oca-decorators`` is not very active. See [8 commits](https://github.com/OCA/oca-decorators/commits/master) between October 2017 and December 2017. For now, 2 of of 4121 OCA modules are using it : 

image

decorator_allowed_groups/__manifest__.py
legalsylvain: Done.

decorator_allowed_groups/__manifest__.py
legalsylvain: Done (set in one line).

decorator_allowed_groups/models/ir_ui_view.py
```diff
@@ -0,0 +1,20 @@
...
+    _inherit = "ir.ui.view"
+
+    @api.model
+    def postprocess(self, model, node, view_id, in_tree_view, model_fields):
+        if node.tag == "button" and node.get("name", False):
+            func = getattr(self.env[model], node.get("name"), False)
+            if func:
+                group_xml_ids = getattr(func, "_allowed_groups", False)
+                if group_xml_ids:
+                    node.set("groups", ",".join(group_xml_ids))
```
StefanRijnhart, line 18: Maybe emit a warning if groups is already set and not equivalent?

legalsylvain: I thought about that point, and I'm not sure.

Consider you have an Odoo / OCA module with a ``def my_action(self)`` defined, and a button defined in a view with ``groups="module_1._group_A"``.

In a custom module, you want to change the group, so you simply write 

```
    @api.allowed_groups("module_2.group_B")
    def my_action(self):
        return super().my_action(self)
```

If you do so, it will raise a false warning, forcing developpers to overload the view in the custom module, to remove the group. 

I so just added in the ``USAGE.rst`` file this text : 

``
The groups defined in the decorators take precedence over the groups that would be defined in the existing views.
``

What do you think ?

decorator_allowed_groups/models/decorator.py
legalsylvain: Matter of taste, maybe. I followed the @lmignon's recommandation.
https://github.com/OCA/server-tools/issues/944#issuecomment-336069286

I have no opinion on the subject.

decorator_allowed_groups/models/ir_ui_view.py
StefanRijnhart: My preference would be to emit the warning. It seems to me that the value of this module is mainly for methods introduced in modules that use this decorator in the first place, rather than overriding existing methods, and in that case, aligning the views is the proper thing to do to avoid confusion. But it's just my preference.

decorator_allowed_groups/models/decorator.py
```diff
@@ -0,0 +1,53 @@
...
+            else:
+                # If it's KO, raise an error.
+                # We raise a technical message (with function name
+                # and xml_ids of the groups, because this message
+                # will be raised only in XML-RPC call.
+                raise AccessError(_(
+                    "To execute the function {}, you should be member"
+                    " of one of the following groups:\n {}".format(
+                        method,
+                        ', '.join(group_xml_ids))
```
StefanRijnhart: I believe applying `format` on user controlled input (such as a translatable string) has security implications (except that `format` is applied here to the translation source instead of the translation value but that is another thing).

StefanRijnhart:
Do you know if decorators like this have to be repeated when their methods are overridden?

decorator_allowed_groups/models/decorator.py
NL66278: In my opinion allowed_groups is much clearer and makes the code self documenting.

decorator_allowed_groups/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2021-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': "Decorator - Allowed Groups",
```
NL66278: You mix single and double quotes in this file. I would recommend to black the whole module, easing migration to subsequent versions where this has to be done anyway.

decorator_allowed_groups/models/ir_ui_view.py
NL66278: What about combining the groups in the button and in the decorator, so you can only limit the authorities by either method, but never grant authorities?

StefanRijnhart: I would say that is not the way I would expect and want such a mechanism to work. I guess it depends on whether you take a security perspective or a usability perspective.

decorator_allowed_groups/models/decorator.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def decorator(method):
+        def secure_method(*args, **kwargs):
+            _self = args[0]
+            # Check if current user is member of correct groups
+            if any([
+                _self.env.user.has_group(group_xml_id)
+                for group_xml_id in group_xml_ids
+            ]):
```
sbidoul:
```suggestion
            if any(
                _self.env.user.has_group(group_xml_id)
                for group_xml_id in group_xml_ids
            ):
```

Otherwise the full list is built before any starts looking into it.

sbidoul (commented):
nice!

decorator_allowed_groups/models/ir_ui_view.py
NL66278: It is what I would expect. But I think I expressed myself poorly. Combining should be interpreted as: when groups are defined on the button and in the decorator, only the groups that are valid in both should be shown. When this leaves no valid groups, at least a warning should be shown, or maybe even an exception thrown.

decorator_allowed_groups/models/decorator.py
legalsylvain: You're right ! In fact, I replace by build in function that does exactly what I want. ``user_has_groups``.
(https://github.com/odoo/odoo/blob/12.0/odoo/models.py#L1204)

decorator_allowed_groups/__manifest__.py
legalsylvain: Thanks ! Done.

decorator_allowed_groups/models/decorator.py
legalsylvain: > applied here to the translation source instead of the translation value but that is another thing

Thanks ! Fixed.

> applying format on user controlled input (such as a translatable string) has security implications

I didn't know. I replaced by ``%s``. Is it better ? ;-)

decorator_allowed_groups/models/ir_ui_view.py
legalsylvain: > Maybe emit a warning if groups is already set and not equivalent?

You're right. :+1: 
I added this point to the roadmap. Better to add a warning. We can remove it in a second time, if we consider later, that it is not relevant in some cases.

> when groups are defined on the button and in the decorator, only the groups that are valid in both should be shown

do you mean : 
- if ``groups="base.group_A,base.group_B"`` 
- and ``@api.allowed_groups("base.group_B", "base.group_C")``
-> the result should ``groups="base.group_B"`` ?

legalsylvain:
Regarding inheritance, 

@hpar said : 

> I have concerns about how it behave when :
> * somewhere in the inheritance chain the decorator is missing
> * some groups are added (limit rights)
> * some groups are added (increase rights)

and @StefanRijnhart said : 

> Do you know if decorators like this have to be repeated when their methods are overridden?


Well, it was not correctly handled, but I refactored a little bit the code to make inheritance working. here is the text added in the USAGE.rst file.


Inheritance mechanisms
----------------------

it is possible to change accreditation level in custom module.

For exemple, if a module A defines a function like this:

    @api.allowed_groups("purchase.group_purchase_manager")
    def my_action(self):
        pass

In a custom module B, that depends on module A :

1) You can overwrite accreditation level by writing

```
    # Now checks if user is member of 'purchase.group_purchase_user'
    @api.allowed_groups("purchase.group_purchase_user")
    def my_action(self):
        return super().my_action()
```

2) You can remove checkes, by writing

```
    # No longer performs checks
    @api.allowed_groups()
    def my_action(self):
        return super().my_action()
```

3) Or you can only overload the function, without changing the accreditation level by writing

```
    # the user must always be a member of 'purchase.group_purchase_manager'
    def my_action(self):
        # Custom code ...
        return super().my_action()
```

I wrote some dummy code, available here. you can play with that code https://github.com/grap/odoo-addons-extra/pull/1, if you want to test.

decorator_allowed_groups/models/ir_ui_view.py
StefanRijnhart: > only the groups that are valid in both should be shown

Ah indeed, I misunderstood. This would improve the usability as well as the security. :+1:

decorator_allowed_groups/models/decorator.py
StefanRijnhart: Yes, see https://github.com/OCA/pylint-odoo/issues/302

StefanRijnhart:
Thank you for your clarification regarding overrides.

> Or you can only overload the function, without changing the accreditation level by writing

Just to be clear, is it the case that the check is executed when `super()` is called?

legalsylvain:
> Thank you for your clarification regarding overrides.

In fact, writing test, It seems that it doesnt work exactly as I have described it. But that is the objective ! 

> Just to be clear, is it the case that the check is executed when super() is called?

I'll test it, but I guess, yes.

lmignon:
> ```
>     File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/__init__.py", line 1, in <module>
>         from . import purchase_order
>     File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/purchase_order.py", line 8, in <module>
>         class PurchaseOrder(models.Model):
>     File "/src_code/server-tools/decorator_allowed_groups_for_test_only/models/purchase_order.py", line 11, in PurchaseOrder
>         @api.allowed_groups("base.group_multi_company")
>     AttributeError: module 'odoo.api' has no attribute 'allowed_groups'
> ```
> 

For sure the decorator will not be available in the `odoo.api` package since your [patch](https://github.com/OCA/server-tools/pull/2026/files#diff-0afba1657b7c53cdbaff376c71688997ff7e19a85d1bd29452217e0e05f8e1d8R55)  is no executed while the addon is not imported by the addons loader. BUT the best practice is not to import the new decorator from `odoo.api` but from the its real source package `from odoo.addons.decorator_allowed_groups.models.decorator import allowed_groups`. In such a case, I'm pretty sure that you don't need to install the addon to use it once it's available into the addons_path.  If you add an addon to `server_wide_modules` it's available for all the databases even these where it's not into the dependency graph. 

My comment is not a blocking comment, it's just my thought about the right way to provide such a functionality.

decorator_allowed_groups/models/decorator.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (C) 2021-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api
+from odoo.exceptions import AccessError
+
+
+def allowed_groups(*group_xml_ids):
```
lmignon, line 9: The decorator is not a model. IMO it should be moved into `decorator_allowed_groups/api.py`

decorator_allowed_groups/readme/USAGE.rst
```diff
@@ -0,0 +1,86 @@
...
+
+can be replaced by:
+
+.. code-block:: xml
+
+    <button type="object" name="my_action"/>
+
+.. code-block:: python
+
+    @api.allowed_groups("purchase.group_purchase_manager")
```
lmignon, line 22: @legalsylvain IMO it's not a good practice to import the decorator from `odoo.api` this will only works if you add the addon into the `server_wide_modules` list and will require to modify the configuration file each time this addon is used to avoid strange error at server start. (Could become a pain in OCA repos). ` from odoo.addons.decorator.allowed_groups.api import allowed_groups` will work in every case without requiring to add the addon into the `server_wide_modules` list

lmignon (changes requested):

thomaspaulb:
@legalsylvain Will you still be working on this?

legalsylvain:
> @legalsylvain Will you still be working on this?

Hi @thomaspaulb. Thanks for your interest.

My PoV :
 
I still think that : 
- this feature could be implemented. (Less code / avoid security by obscurity)
- put the code in the external library [oca-decorators](https://github.com/OCA/oca-decorators/) is not a good idea. (In fact there is not a single commit on that library since 7 years). The code should be put in an odoo module.

But, now, I think that creating a dedicated OCA module is not a good idea. This feature should be put in the odoo ``base`` module to be used widely.
In fact, "good decorators" have been widely used because there are in the core. (like the "new" ``@api.model_create_multi`` that improves performance)

At the last Open Days 2023, I spoke briefly about this subject with @rco-odoo. Perhaps the idea will finally be taken up by the Odoo Core team...

PR #2027, opened 2021-02-19
[13.0][MIG] html_text: Migration to 13.0

victoralmau:
Migration to 13.0

@Tecnativa TT27586


CarlosRoca13 (approved):
:+1: LGTM

CarlosRoca13:
But please squash the administrative commits

html_text/README.rst
```diff
@@ -0,0 +1,82 @@
...
+=====
+
+This module just adds a technical utility, but nothing for the end user.
+
+If you are a developer and need this utility for your module, see these
+examples and read the docs inside the code.
+
+Python example::
+
+    @api.multi
```
joao-p-marques: Maybe remove this from the docs, as it is no longer used.

joao-p-marques (approved):

html_text/README.rst
joao-p-marques: @victoralmau you need to remove it from the `USAGE.rst` file, otherwise it will just be overwritten by the bot

victoralmau: In this addon there is no readme folder, you should create it (with the corresponding files) so that the bot correctly regenerates this file when it is merged?.

joao-p-marques: Oh ok. In that case, you should probably follow the 2nd point in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration to create the correct folder

pedrobaeza (approved):

PR #2029, opened 2021-02-24
[13.0][MIG] datetime_formatter: Migration to 13.0

victoralmau:
Migration to 13.0

Please @ernestotejeda   and @joao-p-marques can you review it?

@Tecnativa TT27104


datetime_formatter/models/res_lang.py
```diff
@@ -0,0 +1,98 @@
...
+MODE_DATETIME = "MODE_DATETIME"
+MODE_DATE = "MODE_DATE"
+MODE_TIME = "MODE_TIME"
+
+
+class ResLang(models.Model):
+    _inherit = "res.lang"
+
+    @api.model
+    @api.returns("self")
```
joao-p-marques: I think this is not necessary. The migration steps specify that this decorator can be removed, and besides it returns "self", so I think it can be removed:
```suggestion
```

ernestotejeda (approved):

joao-p-marques (approved):

PR #2044, opened 2021-03-17
[14.0][IMP] auditlog: prevent cascading delete of logs when models or fields are unlinked

StefanRijnhart:
When a field or a model is unlinked, keep the related audit logs. Denormalize the field and model info on the logs and log lines so that the information is still available after the deletion of the related data model.

Also, to improve the performance of the deletion of fields and models, add indexes on the log's `model_id` and log line's `field_id`.

Port of https://github.com/OCA/server-tools/pull/1262.

### DIFF
--- a/auditlog/migrations/14.0.1.1.0/pre-migration.py
+++ b/auditlog/migrations/14.0.1.1.0/pre-migration.py
@@ -1,0 +1,73 @@
+# © 2018 Pieter Paulussen <pieter_paulussen@me.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    logger = logging.getLogger(__name__)
+    logger.info(
+        "Creating columns: auditlog_log (model_name, model_model) "
+        "and auditlog_log_line (field_name, field_description)."
+    )
+    cr.execute(
+        """
+    ALTER TABLE auditlog_log
+    ADD COLUMN IF NOT EXISTS model_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS model_model VARCHAR;
+    ALTER TABLE auditlog_log_line
+    ADD COLUMN IF NOT EXISTS field_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS field_description VARCHAR;
+    ALTER TABLE auditlog_rule
+    ADD COLUMN IF NOT EXISTS model_name VARCHAR,
+    ADD COLUMN IF NOT EXISTS model_model VARCHAR;
+    """
+    )
+    logger.info(
+        "Creating indexes on auditlog_log column 'model_id' and "
+        "auditlog_log_line column 'field_id'."
+    )
+    cr.execute(
+        """
+        CREATE INDEX IF NOT EXISTS
+        auditlog_log_model_id_index ON auditlog_log (model_id);
+        CREATE INDEX IF NOT EXISTS
+        auditlog_log_line_field_id_index ON auditlog_log_line (field_id);
+    """
+    )
+    logger.info(
+        "Preemptive fill auditlog_log columns: 'model_name' and " "'model_model'."
+    )
+    cr.execute(
+        """
+    UPDATE auditlog_log al
+    SET model_name = im.name, model_model = im.model
+    FROM ir_model im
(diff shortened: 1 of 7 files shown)


auditlog/models/log.py
```diff
@@ -9,7 +10,11 @@ class AuditlogLog(models.Model):
     _order = "create_date desc"
 
     name = fields.Char("Resource Name", size=64)
-    model_id = fields.Many2one("ir.model", string="Model")
+    model_id = fields.Many2one(
+        "ir.model", string="Model", index=True, ondelete="set null"
+    )
+    model_name = fields.Char(string=u"Model Name", readonly=True)
+    model_model = fields.Char(string=u"Technical Model Name", readonly=True)
```
PieterPaulussen: @StefanRijnhart the `unicode` is probably not required here?

auditlog/migrations/14.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,73 @@
...
+    if not version:
+        return
+    logger = logging.getLogger(__name__)
+    logger.info(
+        "Creating columns: auditlog_log (model_name, model_model) "
+        "and auditlog_log_line (field_name, field_description)."
+    )
+    cr.execute(
+        """
+    ALTER TABLE auditlog_log
```
rconjour, line 16: The alignment of those lines seems odd

rconjour (changes requested):
Looking good, only the syntax offset in the migration seems odd.

PieterPaulussen (changes requested):

auditlog/models/log.py
StefanRijnhart: Ah yes. As a matter of fact it wasn't necessary in Python 2.7, but they are particularly offensive in Python 3.x. Removed now.

auditlog/migrations/14.0.1.1.0/pre-migration.py
StefanRijnhart: Same as https://github.com/OCA/server-tools/blob/8.0/auditlog/migrations/8.0.1.4.0/pre-migration.py. I don't think there is a leading convention on the formatting of multiline SQL statements so I would prefer to leave it like this.

StefanRijnhart: The only change was applied to the position of the quotes, by python-black in the prehook, and we can't derive from that.

StefanRijnhart:
Oh, just pushed a version that set the new fields to readonly in the view.

rconjour (approved):

auditlog/models/log.py
```diff
@@ -9,7 +10,11 @@ class AuditlogLog(models.Model):
     _order = "create_date desc"
 
     name = fields.Char("Resource Name", size=64)
-    model_id = fields.Many2one("ir.model", string="Model")
+    model_id = fields.Many2one(
+        "ir.model", string="Model", index=True, ondelete="set null"
+    )
+    model_name = fields.Char(readonly=True)
+    model_model = fields.Char(string="Technical Model Name", readonly=True)
```
PieterPaulussen, line 17: Stupid question probably, but why aren't these stored related fields?

auditlog/models/log.py
```diff
@@ -35,5 +60,30 @@ class AuditlogLogLine(models.Model):
     new_value = fields.Text()
     old_value_text = fields.Text("Old value Text")
     new_value_text = fields.Text("New value Text")
-    field_name = fields.Char("Technical name", related="field_id.name")
-    field_description = fields.Char("Description", related="field_id.field_description")
+    field_name = fields.Char("Technical name", readonly=True)
+    field_description = fields.Char("Description", readonly=True)
```
PieterPaulussen, line 64: Same question as above?

auditlog/models/log.py
StefanRijnhart: Pieter, you *wrote* this code ;-) The reason they are not stored related fields is because those would get wiped when the field/model is deleted, whereas the exact point of these fields is to keep the info when the field/model is deleted.

PieterPaulussen (approved):

auditlog/models/log.py
StefanRijnhart: This OK for you @PieterPaulussen?

PR #1720, opened 2019-11-06
[10.0][FIX] issue 1227. auth_admin_passkey. Error on editing general settings.

NL66278:
This is an alternative for the stale PR: https://github.com/OCA/server-tools/pull/1228

PR #2047, opened 2021-03-18
[10.0][FIX] auth_admin_passkey: issue with 'bool' object has no attribute 'strip' after login

thomaspaulb:
@NL66278 After your [last update](https://github.com/OCA/server-tools/pull/1720/files), things went haywire in my system.

What happened was:

- The config parameters in my system previously set, I think, to a string of "None", "False" or "0", were now interpreted as a truthy value, so the system started to send out emails on login
- I went into the settings to see what was happening and the setting was now on. I assumed the customer had changed it or something, so I switched it off again.
- Now after logging in with the passkey, I received 'bool' has no attribute 'strip' error.

I think this should fix it, I will test it also.

### DIFF
--- a/auth_admin_passkey/__manifest__.py
+++ b/auth_admin_passkey/__manifest__.py
@@ -4,26 +4,13 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 {
-    'name': 'Authentification - Admin Passkey',
-    'version': '10.0.1.0.2',
-    'category': 'base',
-    'author': "GRAP,Odoo Community Association (OCA)",
-    'website': 'http://www.grap.coop',
-    'license': 'AGPL-3',
-    'depends': [
-        'mail',
-    ],
-    'data': [
-        'data/ir_config_parameter.xml',
-        'views/res_config_view.xml',
-    ],
-    'demo': [],
-    'js': [],
-    'css': [],
-    'qweb': [],
-    'images': [],
-    'post_load': '',
-    'application': False,
-    'installable': True,
-    'auto_install': False,
+    "name": "Authentification - Admin Passkey",
+    "version": "10.0.1.0.2",
+    "category": "base",
+    "author": "GRAP,Odoo Community Association (OCA)",
+    "website": "http://www.grap.coop",
+    "license": "AGPL-3",
+    "depends": ["mail"],
+    "data": ["data/ir_config_parameter.xml", "views/res_config_view.xml"],
+    "installable": True,
 }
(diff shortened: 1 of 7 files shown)


auth_admin_passkey/models/res_config.py
```diff
@@ -14,8 +15,8 @@ class BaseConfigSettings(models.TransientModel):
     def get_default_auth_admin_passkey_send_to_admin(self, fields):
         return {
             'auth_admin_passkey_send_to_admin':
-            self.env["ir.config_parameter"].get_param(
-                "auth_admin_passkey.send_to_admin")
+            safe_eval(self.env["ir.config_parameter"].get_param(
+                "auth_admin_passkey.send_to_admin") or 'False')
```
NL66278: Let us start to use black code conventions for new code. Will ease migration to 13.0 and beyond. So quoting strings with double quotes.

NL66278:
@thomaspaulb  Sorry to hear about your troubles.

Maybe add some unit tests for the situations I did not recognize, to make sure they will no longer occur?

auth_admin_passkey/models/res_config.py
thomaspaulb: In order to make reviewing easier I've run 'black' before the two other commits, which are the functional fixes.

thomaspaulb:
@NL66278 I've created a unit test and made the commit messages more descriptive, actually there were two separate issues at hand.

auth_admin_passkey/__manifest__.py
```diff
@@ -4,26 +4,21 @@
...
-    'auto_install': False,
+    "name": "Authentification - Admin Passkey",
+    "version": "10.0.1.0.2",
+    "category": "base",
+    "author": "GRAP,Odoo Community Association (OCA)",
+    "website": "http://www.grap.coop",
+    "license": "AGPL-3",
+    "depends": ["mail"],
+    "data": ["data/ir_config_parameter.xml", "views/res_config_view.xml"],
+    "demo": [],
```
NL66278: Empty keys should be deleted. Actually you can remove application and auto_install keys as well.

NL66278:
@thomaspaulb Can you delete the unneeded keys from the manifest - now that all those lines change because of the blacking anyway - and I think we are good to go. Great work!

auth_admin_passkey/__manifest__.py
thomaspaulb: I've removed those, and ran 'isort' as well over the first commit.

NL66278 (approved):
:+1: LGTM

ntsirintanis (approved):

thomaspaulb:
An alternative fix was already merged here: https://github.com/OCA/server-tools/pull/1890/files

PR #2048, opened 2021-03-19
[IMP] base_custom_info enable to easily import custom info values

Rad0van:
I use base_custom_info module a bit differently than its original plan. I combine properties from multiple templates to form product custom information like technical specs, params, ...  The values come from externals sources - I import them mostly. However as it stands now I'd need to write specific field depending on type (i.e. value_str, value_int, ...).
This simple change should allow for just writing into value field using already present _inverse_value function. I haven't found any side effects so far.

### DIFF
--- a/base_custom_info/models/custom_info_value.py
+++ b/base_custom_info/models/custom_info_value.py
@@ -61,6 +61,7 @@ class CustomInfoValue(models.Model):
     required = fields.Boolean(related="property_id.required")
     value = fields.Char(
         compute="_compute_value",
+        inverse="_inverse_value",
         search="_search_value",
         help="Value, always converted to/from the typed field.",
     )


Rad0van:
@pedrobaeza, @dalonsod, @Tardo could you please have a look? Simple change that looks quite "natural" to me.

Also I have a question: I am importing the values from CSV/XLS. This eases importing of values for properties. However I have to address the owner of the property (owner_id) by its database id. Thus I have to take an extra step of exporting the parent records I am interested in together with their ids and match them when importing property values. I am sure there can be a clever hack done that would allow me to address the parent records by their external ids. Could you please give any pointers where to look? I guess it would probably has to be some computed field with inverse function that would enable import process to match through that...

dalonsod:
@Rad0van only code review, LGTM. Maybe commit message could be changed to be prefixed by "[IMP] base_custom_info: " or similar

Rad0van:
> @Rad0van only code review, LGTM. Maybe commit message could be changed to be prefixed by "[IMP] base_custom_info: " or similar

@dalonsod done. If I may ask: does anything come to mind how to fool import process to allow import using external ids (of owning objects - for example product.template records) instead of database ids?

dalonsod (approved):

dalonsod:
> If I may ask: does anything come to mind how to fool import process to allow import using external ids (of owning objects - for example product.template records) instead of database ids?

@Rad0van maybe I can't help you, as I'm not an expert of Odoo data import, sorry. I only know the ways described here:  https://www.odoo.com/documentation/user/13.0/general/base_import/import_faq.html#what-s-the-difference-between-database-id-and-external-id

base_custom_info/models/custom_info_value.py
```diff
@@ -225,12 +226,6 @@ def _onchange_property_set_default_value(self):
             if not record.field_type and record.property_id.field_type:
                 record.field_type = record.property_id.field_type
 
-    @api.onchange("value")
-    def _onchange_value(self):
```
pedrobaeza: This is still needed anyway. Please don't remove it.

Rad0van: Sure. But would you care to explain? I'd think it would get triggered with inverse function in place.

pedrobaeza: Just read the docstring for the explanation.

Rad0van: @pedrobaeza done.

PR #2050, opened 2021-03-22
[14.0][MIG] sql_export

florian-dacosta:
Standard migration


sql_export/README.rst
```diff
@@ -0,0 +1,120 @@
...
+  - create
+  - update
+
+See sql_request_abstract module to fix this issue.
+
+* checking SQL request by execution and rollback is disabled in this module
+  since variables features has been introduced. This can be fixed by
+  overloading _prepare_request_check_execution() function.
+
+* Move modules sql_request_abstract and sql_export to oca/reporting-engine
```
petrus-v, line 72: I'm wondering when is the best time to attempt this kind of works ?

I don't know if discussion already take place somewhere ?

 `sql_request_abstract` is already merged in branch 14.0 so I don't know what's the best to do. At least tell when it's the best time to do it and makes some consistent with `sql_request_abstract` README  (also not sure what's the OCA best practice for that kind of cross ref do it in different PR or in the same atomic commit ?)

petrus-v (changes requested):
Quick review and test few query on run bot (I'havn't plays with access write) LGTM once future is clarify

Other few remarks that I won't block that could be done after migration:
- [ ] it would be nice display a message in the execute query wizard if there are not parameters or directly execute and download the file
- [ ] I don't really understand states on sql.export model as long you can still edit the sql query after validation maybe move that field to readonly ?
- [ ] it would be nice to add in the query field help how to use parameters

florian-dacosta:
Hello @petrus-v 
The improvement proposal you made seem good to me. 
As you said it could be done after or during the migration, I have no time right now but will keep it in mind of if someone want to contribute this.
About the repo, I still believe it should move to report engine, as it really is a report, but I also think it should stays with sql_request_abstract. (so maybe do it in a next version? Not sure though...

petrus-v:
> Hello @petrus-v
> The improvement proposal you made seem good to me.
> As you said it could be done after or during the migration, I have no time right now but will keep it in mind of if someone want to contribute this.
> About the repo, I still believe it should move to report engine, as it really is a report, but I also think it should stays with sql_request_abstract. (so maybe do it in a next version? Not sure though...

At least (I believe) a choice should be token and explicit in the readme to avoid loosing time at the next migration.

I suppose one of the @OCA/tools-maintainers team have the authority to make the decision !

> I don't really understand states on sql.export model as long you can still edit the sql query after validation maybe move that field to readonly

And I think it would be nice to clarify that point to be consistency, probably move the sql query to readonly once it has been validated.

florian-dacosta:
@petrus-v 
I've added a Roadmap on `sql_request_abstract` so that the repo change is not forgotten on next version.

>And I think it would be nice to clarify that point to be consistency, probably move the sql query to readonly once it has been validated.
You are right, I've change the view and put some fields in readonly mode once validated.

I've also made some usability change about parameters and added a description on the field.
There are still a lot of room for improvement  thought.

sql_export/models/sql_export.py
```diff
@@ -27,7 +27,22 @@ class SqlExport(models.Model):
         "sql_id",
         "field_id",
         "Parameters",
-        domain=[("model", "=", "sql.file.wizard")],
+        domain=[("model", "=", "sql.file.wizard"), ("state", "=", "manual")],
+        help="Before adding parameters, make sure you have created one that fill your "
+        "need in the dedicated menu with the right type dans label. \n"
```
petrus-v: in label ?

sql_export/wizard/wizard_file.py
```diff
@@ -70,12 +70,10 @@ def export_sql(self):
                 else:
                     variable_dict[field.name] = self[field.name]
         if "%(company_id)s" in sql_export.query:
-            company_id = self.env.context.get(
-                "force_company", self.env.user.company_id.id
-            )
+            company_id = self.env.company.id
```
petrus-v, line 73: I suppose that the force_user/ force_company was to be able to create dedicated views according action menu context ie: one dedicated view per company (instead of the current company) can be confused at some point, not sure what's the best todo !

petrus-v (approved):
A note and type otherwise you get my :+1: 

thanks for migration

sql_export/wizard/wizard_file.py
florian-dacosta: No, these are kind of special parameters, you can use in the sql query %(company_id)s so it will be replaced by the company of the user.
Same for the user_id

It may be usefull to have a dynamic query, that depends on the user and/or of the company of the user, that execute it.
So, the same query won't give the same csv file depending on which user execute it.

The force_company context is deprecated now in favor of the current company on the environment

But indeed, the sql export menu does not manage multi company for now, we may add a company_id field on it and add some ir.rules, in order to display only the queries made for the allowed companies of the user, it would be a nice improvement for intercompany projects

sql_export/models/sql_export.py
florian-dacosta: Miss typing "and"
Thanks!

bealdav (approved):
LGTM

PR #2052, opened 2021-03-23
[14.0][IMP] upgrade_analysis: make us of odoorpc/openupgradelib master

MiquelRForgeFlow:
It is always a good practice to use the last version (`master`) of OCA libraries, which are contin**u**ously maintained and thus not depend of pypi outdated versions.

Includes minor fixes.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,0 +1,2 @@
+odoorpc @ git+https://github.com/OCA/odoorpc.git@master
+openupgradelib @ git+https://github.com/OCA/openupgradelib.git@master
--- a/upgrade_analysis/blacklist.py
+++ b/upgrade_analysis/blacklist.py
@@ -2,7 +2,7 @@ BLACKLIST_MODULES = []
 
 # the hw_* modules are not affected by a migration as they don't
 # contain any ORM functionality, but they do start up threads that
-# delay the process and spit out annoying log messages continously.
+# delay the process and spit out annoying log messages continuously.
 
 # We also don't want to analyze tests modules
 BLACKLIST_MODULES_STARTS_WITH = ["hw_", "test_"]
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -14,14 +14,14 @@ import copy
 try:
     from odoo.addons.openupgrade_scripts import apriori
 except ImportError:
-    from dataclasses import dataclass, field
+    from dataclasses import dataclass, field as dc_field
 
     @dataclass
     class NullApriori:
-        renamed_modules: dict = field(default_factory=dict)
-        merged_modules: dict = field(default_factory=dict)
-        renamed_models: dict = field(default_factory=dict)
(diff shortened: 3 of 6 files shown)


MiquelRForgeFlow:
@sbidoul Does a way exist to avoid the call to `pkg_resources.pyparsing` or any workaround? It's to avoid the error:
```yml
pkg_resources.extern.packaging.requirements.InvalidRequirement: Parse error at "'+https:/'": Expected stringEnd
```

I call you because travisCI is using your `acsone/setuptools-odoo` branch.

upgrade_analysis/__manifest__.py
```diff
@@ -20,7 +20,11 @@
     ],
     "installable": True,
     "external_dependencies": {
-        "python": ["dataclasses", "odoorpc", "openupgradelib"],
+        "python": [
+            "dataclasses",
+            "git+https://github.com/OCA/odoorpc.git@master",
```
pedrobaeza: Why forcing this here?

legalsylvain (approved):
Thanks !

upgrade_analysis/__manifest__.py
legalsylvain: Otherwise, it will be erased by new repo tools, when merging.

pedrobaeza: Yeah, but I think this is the cause of the red CI.

sbidoul: Put the url link in test-requirements.txt. but it would be better to publish odoorpc and open upgrade lib to pypi. It's easy to automate.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # generated from manifests external_dependencies
 dataclasses
-odoorpc
-openupgradelib
+git+https://github.com/OCA/odoorpc.git@master
+git+https://github.com/OCA/openupgradelib.git@master
```
sbidoul: Assuming we want to use the master branch in tests, these should go in `test-requirements.txt` with this form:

```suggestion
odoorpc @ git+https://github.com/OCA/odoorpc.git@master
openupgradelib @ git+https://github.com/OCA/openupgradelib.git@master
```

`test-requirements.txt` are installed by MQT in PIP and OCA modes and will be installed by GitHub actions, and are very useful to cleanly separate test dependencies from runtime dependencies.

upgrade_analysis/__manifest__.py
sbidoul: To say a bit more now that I have a keyboard: we should never use URLs in python external dependencies because it would force all users of the module to deploy with the exact same URL. And people deploying from git will probably want to use their own branch or pin a specific commit. This is basically for the same reason that we must never pin a project version in external dependencies.

MiquelRForgeFlow: thanks

MiquelRForgeFlow:
PR is ready :)

pedrobaeza (approved):

PR #2054, opened 2021-03-24
[13.0][IMP] base_exception add blocking exceptions

TDu:
By default exceptions can be ignored by the click of a button.
Users begin human they will just click that button what ever
the internal rules.

So this PR adds the option to set specific exceptions as blocking.

When exceptions are detected if one of them is blocking, the user
will not be able to ignore them.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -42,6 +42,9 @@ class ExceptionRule(models.Model):
         help="Python code executed to check if the exception apply or "
         "not. Use failed = True to block the exception",
     )
+    is_blocking = fields.Boolean(
+        string="Is blocking", help="When checked the exception can not be ignored",
+    )
 
     @api.constrains("exception_type", "domain", "code")
     def check_exception_type_consistency(self):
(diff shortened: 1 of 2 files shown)


base_exception/models/base_exception.py
```diff
@@ -224,8 +234,17 @@ def _compute_exceptions_summary(self):
...
-                        "<li>%s: <i>%s</i></li>"
-                        % tuple(map(html.escape, (e.name, e.description)))
+                        "<li>%s: <i>%s</i> <b>%s<b></li>"
+                        % tuple(
+                            map(
+                                html.escape,
+                                (
+                                    e.name,
+                                    e.description,
+                                    "(Blocking exception)" if e.is_blocking else "",
```
simahawk: I guess that this should be translatable, no?

TDu: Yes, thanks, I fixed that

TDu: Yes fixed.

simahawk (approved):
LG but if you have time pls add a test for this change

jbaudoux (approved):

sebalix (approved):

PR #2055, opened 2021-03-25
[14.0][OU-MIG] upgrade_analysis: add migration scripts

MiquelRForgeFlow:
Just in case.

The module rename is added in https://github.com/OCA/OpenUpgrade/pull/2629.

### DIFF
--- a/upgrade_analysis/migrations/14.0.1.0.0/pre-migrate.py
+++ b/upgrade_analysis/migrations/14.0.1.0.0/pre-migrate.py
@@ -1,0 +1,27 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+_model_renames = [
+    ("openupgrade.analysis.wizard", "upgrade.analysis"),
+    ("openupgrade.attribute", "upgrade.attribute"),
+    ("openupgrade.comparison.config", "upgrade.comparison.config"),
+    ("openupgrade.record", "upgrade.record"),
+    ("openupgrade.generate.records.wizard", "upgrade.generate.record.wizard"),
+    ("openupgrade.install.all.wizard", "upgrade.install.wizard"),
+]
+
+_table_renames = [
+    ("openupgrade_analysis_wizard", "upgrade_analysis"),
+    ("openupgrade_attribute", "upgrade_attribute"),
+    ("openupgrade_comparison_config", "upgrade_comparison_config"),
+    ("openupgrade_record", "upgrade_record"),
+    ("openupgrade_generate_records_wizard", "upgrade_generate_record_wizard"),
+    ("openupgrade_install_all_wizard", "upgrade_install_wizard"),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_models(env.cr, _model_renames)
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
@pedrobaeza

pedrobaeza:
I don't think this is going to happen, as this tool can be uninstalled and re-install. Moreover, I doubt the records created from one version serves for another, so I would instead write in the README about not migrating having this module installed.

MiquelRForgeFlow:
I know this might don't happen, but just for the fun of continuation, as a keepsake of the old openugrade_records. I think it doesn't add any harm in adding that :)

I could also add a text in the readme, no problem.

pedrobaeza:
But the main question remains: if you generate the records for the analysis in N - 1, are that records removed in version N?

MiquelRForgeFlow:
No, why should them be removed? The person who happens (if happens) migrating a db with the old module installed will decide. Here it's just the table renames, to make the migration work. I don't mind here if there are generated records or not.

pedrobaeza:
But that's why I told you about never do that, as the module will do wrong things if kept migrating from one version to another.

pedrobaeza:
@StefanRijnhart do you follow my rationale?

MiquelRForgeFlow:
> write in the README about not migrating having this module installed.

I added a text in the readme.

StefanRijnhart:
The upgrade records are supposed to be volatile. Up to 13.0, they were truncated before each new analysis: https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/openupgrade_records/models/generate_records_wizard.py#L38-L44
But this is missing in this module: https://github.com/OCA/server-tools/blob/14.0/upgrade_analysis/wizards/upgrade_generate_record_wizard.py#L20-L27. @legalsylvain what was the reason to leave it out?

Having that said, there is no harm in having this migration here as far as I'm concerned.

pedrobaeza:
OK, let's merge it and wait for @legalsylvain to say if the clean operation should be restored.

legalsylvain:
> @legalsylvain what was the reason to leave it out?

Hi all. The offending commit seems this one : https://github.com/OCA/server-tools/commit/f8f40d36174c9bc04ec6d4db93c7d856f7f7c1fe#diff-cea8bc48b215b1ea3ddbc78a8fdd4eb94fd4e9c1670cfa07199e820dfb21cb05L36

I must confess I don't remember why I removed that lines, and the commit message doesn't help me...

So yes, if we want to call many times the ``generate`` function, data should be dropped, indeed. What I just don't understand in the previous implementation is the test, to check if tables exists. if the module ``upgrade_anlysis`` is installed, the two tables should exist. Don't you tink ? 

So i'm in favour to readd simply

    self.env.cr.execute("TRUNCATE upgrade_attribute, upgrade_record;")

Anyway, thanks @MiquelRForgeFlow to play with the new implementation of ``upgrade_analysis`` module and OpenUpgrade and thanks for your fixes !

pedrobaeza:
OK for me

MiquelRForgeFlow:
Can I add this in the other PR o make a new one?

pedrobaeza:
I think you can add it as separate commit.

MiquelRForgeFlow:
Ok, done in https://github.com/OCA/server-tools/pull/2057.

StefanRijnhart:
Thanks @legalsylvain and thanks @MiquelRForgeFlow for the follow-up.

PR #2057, opened 2021-03-27
[14.0][FIX] upgrade_analysis: fix methods related to noupdate data

MiquelRForgeFlow:
Last time I did the analysis I detected that the noupdate data code was broken and thus, not updated. The break of that is silenced, so I didn't see it. But with more time, I have been fixing that to make it works. It will also be needed https://github.com/OCA/OpenUpgrade/pull/2632 compatibility to make it work. The result is https://github.com/OCA/OpenUpgrade/pull/2631.

Errors fixed:
- ~~Method `list_modules` was returning a list of lists, which was wrong.~~
- Method `get_xml_records` was reading the files to get parsed later with `_parse_files`. That breaks when the xml file uses a different enconding than `utf-8`, like `data/account_chart_template_data.xml` in `l10n_ma`. In fact, the reading step proves to be unnecessary, as it can be parsed directly, that's what it does now (and what it was already doing in the script of v13).
- Also merged_modules of apriori was not being considered.

And then, I update my company name :P

~~Note: I also added the analysis.txt to justify the migration scripts haha~~

### DIFF
--- a/upgrade_analysis/models/upgrade_record.py
+++ b/upgrade_analysis/models/upgrade_record.py
@@ -1,6 +1,6 @@
 # Copyright 2011-2015 Therp BV <https://therp.nl>
 # Copyright 2016-2020 Opener B.V. <https://opener.am>
-# Copyright 2019 Eficent <https://eficent.com>
+# Copyright 2019 ForgeFlow <https://forgeflow.com>
 # Copyright 2020 GRAP <https://grap.coop>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
@@ -159,7 +159,7 @@ class UpgradeRecord(models.Model):
         # The order of the keys are important.
         # Load files in the same order as in
         # module/loading.py:load_module_graph
-        files = []
+        paths = []
         for key in ["init_xml", "update_xml", "data"]:
             if not manifest.get(key):
                 continue
@@ -167,6 +167,5 @@ class UpgradeRecord(models.Model):
                 if not xml_file.lower().endswith(".xml"):
                     continue
                 parts = xml_file.split("/")
-                with open(os.path.join(addon_dir, *parts), "r") as xml_handle:
-                    files.append(xml_handle.read())
-        return files
+                paths.append(os.path.join(addon_dir, *parts))
+        return paths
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

pedrobaeza (approved):

upgrade_analysis/models/upgrade_record.py
```diff
@@ -140,7 +140,7 @@ def list_modules(self):
             """SELECT DISTINCT(module) FROM upgrade_record
             ORDER BY module"""
         )
-        return [module for module, in self.env.cr.fetchall()]
+        return [module[0] for module in self.env.cr.fetchall()]
```
StefanRijnhart: Why did you change this? I think the original is a very elegant way to parse tuples of size one.

upgrade_analysis/migrations/14.0.1.0.0/upgrade_analysis.txt
```diff
@@ -0,0 +1,72 @@
+---Models in module 'upgrade_analysis'---
```
StefanRijnhart: Do we want to include the analysis of non-core modules? I would prefer not to.

upgrade_analysis/models/upgrade_record.py
MiquelRForgeFlow: No, without that `[0]` then you return something like [("module1",), ("module2",), ...] which later breaks. But if you return instead ["module1", "module2", ...] then later it doesn't break. My changes have been tested.

upgrade_analysis/migrations/14.0.1.0.0/upgrade_analysis.txt
MiquelRForgeFlow: Ok, then I will remove then.

upgrade_analysis/models/upgrade_record.py
StefanRijnhart: The comma after `module` unpacks the tuple fine.
```
>>> [module for module, in  [("module1",), ("module2",),]]
['module1', 'module2']
```

MiquelRForgeFlow: Oh, shit, then it was my bad. Yeah, it's elegant, but where does it fall the readability? hehe

StefanRijnhart (approved):

PR #2058, opened 2021-03-30
 [FIX+IMP] module_analysis : fix analysis during update and make update faster

legalsylvain:
**Rational**

for the time being (before this patch), the update of the analysis is done manually (by Update module List) AND when an update of modules is done (update all, or partial update).
The second analysis is not OK : 
- It is incorrect, because it is partial : module loaded before ``module_analysis`` module are not correctly analysed, due to Odoo design. So for exemple, ``base`` module is never analysed that way.
- It makes the update process slower.

**Description of the PR**
- Remove analysis when updating modules, 

- Add instead a cron task that is executed nightly to update analysis automatically and fully.

thanks for your review

### DIFF
--- a/module_analysis/__manifest__.py
+++ b/module_analysis/__manifest__.py
@@ -10,6 +10,7 @@
     'author': 'GRAP, Odoo Community Association (OCA)',
     'website': "https://github.com/OCA/server-tools",
     'version': '12.0.1.0.4',
+    'maintainers': ['legalsylvain'],
     'license': 'AGPL-3',
     'depends': [
         'base',
--- a/module_analysis/__manifest__.py
+++ b/module_analysis/__manifest__.py
@@ -22,6 +22,7 @@
         'views/view_ir_module_type.xml',
         'views/view_ir_module_type_rule.xml',
         'views/view_ir_module_module.xml',
+        'data/ir_cron.xml',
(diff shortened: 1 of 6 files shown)


PierrickBrun (approved):

quentinDupont (approved):

module_analysis/__manifest__.py
```diff
@@ -8,11 +8,12 @@
     " to know which installed modules comes from Odoo Core, OCA, or are"
     " custom modules",
     'author': 'GRAP, Odoo Community Association (OCA)',
+    'maintainers': ['legalsylvain'],
     'website': "https://github.com/OCA/server-tools/",
     'version': '12.0.1.0.2',
     'license': 'AGPL-3',
     'depends': [
-        'base',
+        'base_automation',
```
StefanRijnhart: Why is the dependency on `base_automation` added? Is what you need from ir.cron provided by `base`?

legalsylvain: 1 year later ... 
You're right ! Sorry for the delay. I tested with ``base`` dependency and it works. 
maybe a residual from another implementation I tried...

reverted.

thomaspaulb:
Well apparently I hijacked this without knowing, but let's finish it in https://github.com/OCA/server-tools/pull/2858

PR #2059, opened 2021-04-01
[14.0][ADD] sync_auditlog

Chanakya-OSI:


sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+            ("Pulled", "Pulled"),
+            ("Processed", "Processed"),
+            ("Error", "Error"),
+            ("Failed", "Failed"),
+            ("Cancelled", "Cancelled"),
+        ],
+        string="State",
+        default="logged",
+    )
+    uuid = fields.Char("UUID")
```
dreispt: Would it be better to have a default method to generate the UUID?

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+    timestamp = fields.Datetime("Timestamp")
+    resource_ids = fields.Char("Resource IDs")
+    raw_args_kwargs = fields.Char("Raw Args and kwargs")
+    prepared_args_kwargs = fields.Char("Prepared Args and kwargs")
```
dreispt, line 75: I was originally thinking on having args as kwargs as two separate fields.
No change needed but lets keep this in mind, in case it is helpful when implementing the "Event Apply" logic.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+    context = fields.Char("Context")
+    external_id = fields.Char("External ID")
+    parent_log = fields.Char("Parent log")
+    result = fields.Char("Result")
```
dreispt:
```suggestion
    result = fields.Text("Result")
```

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+    remote = fields.Char("Remote")
+    message = fields.Char("Message")
```
dreispt: What is this filed? I don't see it in the design.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                "Record will be skipped to avoid duplication"
+            ),
+        )
+    ]
+
+    def _cron_prepare_auditlog_events(self):
+        logs = self.search([("state", "=", "captured")])
+        for log in logs:
+            # ToDo - Prepare data Logic
+            print("============ToDo Prepare Data Logic============")
```
dreispt: This method is responsible to populate `prepared_args_kwargs ` and  `external_id` fields

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def _push_event_data(self, addr, uid, password, dbname):
+
+        try:
```
dreispt, line 175: Try-catch should span one or a few tightly related lines of code.
So, one try-catch for the connection, another for the execute call.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+            model = xmlrpclib.ServerProxy("%s/xmlrpc/object" % (addr)).execute(
```
dreispt: I'm quite sure we can skip this call. AFAICS it is checking the log model exists in the server. This can be also checked when try-ing the execute call.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                    "resource_ids": self.resource_ids,
+                    "raw_args_kwargs": self.raw_args_kwargs,
+                    "prepared_args_kwargs": self.prepared_args_kwargs,
+                    "context": self.context,
+                    "external_id": self.external_id,
+                    "parent_log": self.parent_log,
+                    "result": self.result,
+                    "remote": self.remote,
+                }
+                log_id = xmlrpclib.ServerProxy("%s/xmlrpc/object" % (addr)).execute(
```
dreispt: Add a try-except

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                    "model_id": model_id,
+                    "method": self.method,
+                    "uuid": self.uuid,
+                    "res_id": self.res_id,
+                    "user_id": uid,
+                    "log_type": self.log_type,
+                    "state": "Pulled",
+                    "timestamp": self.timestamp,
+                    "resource_ids": self.resource_ids,
+                    "raw_args_kwargs": self.raw_args_kwargs,
```
dreispt:
```suggestion
```

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+            ("Processed", "Processed"),
+            ("Error", "Error"),
+            ("Failed", "Failed"),
+            ("Cancelled", "Cancelled"),
+        ],
+        string="State",
+        default="logged",
+    )
+    uuid = fields.Char("UUID")
+    timestamp = fields.Datetime("Timestamp")
```
dreispt: Add a default to now

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+            if model:
+                model_id = model[0]
+                log_dict = {
+                    "name": self.name,
+                    "model_id": model_id,
+                    "method": self.method,
+                    "uuid": self.uuid,
+                    "res_id": self.res_id,
+                    "user_id": uid,
```
dreispt: The local user might not exist on the remote instance, so this can error. 
```suggestion
```

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                    "log_type": self.log_type,
+                    "state": "Pulled",
+                    "timestamp": self.timestamp,
+                    "resource_ids": self.resource_ids,
+                    "raw_args_kwargs": self.raw_args_kwargs,
+                    "prepared_args_kwargs": self.prepared_args_kwargs,
+                    "context": self.context,
+                    "external_id": self.external_id,
+                    "parent_log": self.parent_log,
```
dreispt: Not sure this ID will exist in the remote...

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                    "result": self.result,
```
dreispt:
```suggestion
```

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                    "remote": self.remote,
+                }
+                log_id = xmlrpclib.ServerProxy("%s/xmlrpc/object" % (addr)).execute(
+                    dbname, uid, password, "auditlog.log", "create", log_dict
+                )
+                if log_id:
+                    for event_line in self.line_ids:
```
dreispt: The log lines are not needed.
If something like this is needed, it should be sent in a single payload, in the previous execute, as a write on a to-many field.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,237 @@
...
+                                uid,
+                                password,
+                                "auditlog.log.line",
+                                "create",
+                                log_line_dict,
+                            )
+                    self.state = "Pushed"
+        except Exception as e:
+            self.state = "Cancelled"
+            self.message = e
```
dreispt: Missing a commit per record

sync_auditlog/models/rule.py
```diff
@@ -0,0 +1,455 @@
+# Copyright (C) 2021 Open Source Integrators
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models, modules
+import uuid
+
+from datetime import datetime
+
+FIELDS_BLACKLIST = [
```
dreispt: You can import this from auditlog, instead of repeating it

sync_auditlog/models/rule.py
```diff
@@ -0,0 +1,455 @@
...
+
+            rule_model.sudo().with_context(
+                black_list_fields=black_list_fields, white_list_fields=white_list_fields
+            ).create_logs(
+                self.env.uid,
+                self._name,
+                new_records.ids,
+                "create",
+                None,
+                new_values,
```
dreispt, line 179: The new values/old values are not needed for the sync log records.
```suggestion
```

sync_auditlog/models/rule.py
```diff
@@ -0,0 +1,455 @@
...
+                additional_log_values,
+            )
+            return new_records
+
+        return create_full if self.log_type == "full" else create_fast
```
dreispt: create_full/create_fast are not needed for the sync log records.
We roll out our won log logic that does not look into new/old values, only to args/kwargs.

sync_auditlog/models/rule.py
```diff
@@ -0,0 +1,455 @@
...
+
+    def _make_read(self):
```
dreispt: Remove - we don't log reads, only regular auditlog does that.

sync_auditlog/models/log.py
omahadik: I have added this call as we require the Model ID to create logs in the remote system. As the model database ID can be different in the remote system, I am fetching the model_id by using the name of the model and then using it in the dictionary to create the log entry.

dreispt: Got it. Then I suggest adding a `model` char field where you can write the model identifier ("res.partner").
Doing so would trigger logic populatig the model_id on that system.
That way we can avoid the Model DB ID issue and simplify the upload.

sync_auditlog/models/log.py
```diff
@@ -0,0 +1,313 @@
...
+import ast
+
+from odoo import fields, models
+from datetime import datetime
+try:
+    from xmlrpc import client as xmlrpclib
+except ImportError:
+    import xmlrpclib
+
+def _generate_xml_id(self, skip=False):
```
dreispt: @vimalpatelserpentcs 
This does the same as `__ensure_xml_id()`, so I see no point in having it, and can remote.

dreispt: Here is a try
```suggestion
def _get_external_id(rec):
    self and self.ensure_one()
    res = list(rec.__ensure_xml_id())
    return res[0][1]

def _generate_xml_id(self, skip=False):
```

rvalyi:
guys, I'm commenting just in case, another/complementary architectural choice could be to expose the auditlog events using https://github.com/OCA/rest-framework, just as explained in this popular post:
https://blog.syncinc.so/events-not-webhooks
and have it consumed by clients possibly different than only Odoo instances (as with this module). **I think that would be a killer module**. I'm not asking to change anything in this sync_auditlog module, but simply bringing awareness about the potential of exposiing this via a REST API...

cc @dreispt @lmignon @kittiu @pedrobaeza @guewen @sebastienbeau @hparfr

pedrobaeza:
Are you thinking on monitoring purposes?

rvalyi:
> Are you thinking on monitoring purposes?

possibly.
But I mas more thinking about the generic need to connect Odoo to any third party tools in a small company. Odoo has some kind of built-in low level JSON-RPC API and XMLRPC API and it's easy to create REST ones (and we might end up having a generic one). But we typically miss that /events end point for third party tools trying to figure out what happended on the Odoo side since the last time they interracted with Odoo so that can know if they should pull some record again from Odoo because it was updated (possibly when they were offline) or do something with Odoo in general...

pedrobaeza:
Well, I think it's a bit off-topic of this module pull request, but my general trend is to avoid intermediate layers.

rvalyi:
a bit off

> Well, I think it's a bit off-topic of this module pull request, but my general trend is to avoid intermediate layers.

In part off topic but may be not as much as you may think. Indeed such a REST exposition of the auditlog logs would probably not be worth depending on sync_auditlog (or would it?). Nonetheless, some fields and concepts such as the uuid and timestamp log fields would typically be shared by such a REST API module. So considering syn_auditlog is still in draft, may be this gives ideas how to architecture things a bit differently.

In fact, Akretion has a v14 customer with such a need that I would have addressed using webhooks until the hackernews post convinced me an /events end point would just be simpler and more powerfull to address future needs. So I'll see if this customer is ready to fund such a POC, over the next weeks, but it is quite likely...

In fact people could also think about a convergence between the OCA/connector event system approach and this new auditlog approach (I think both approach make sense, but may be the two approaches could be more consistent to work well together).

sync_auditlog/views/remote_server_view.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="model">auditlog.remote.server</field>
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <group>
+                        <field name="name"/>
+                        <field name="url" />
+                        <field name="dbname" />
+                        <field name="user" />
+                        <field name="password" password="True"/>
```
hparfr, line 14: Keep in mind: It's not secure to put the password in the views.

PR #2061, opened 2021-04-02
 [IMP] sql_export : add preview button

legalsylvain:
Add a preview button.

image

When clicking, you'll have a quick preview of your request.

image

CC : original commiters : @florian-dacosta , @bealdav, @bguillot 
CC : @quentinDupont

### DIFF
--- a/sql_export/views/sql_export_view.xml
+++ b/sql_export/views/sql_export_view.xml
@@ -8,6 +8,7 @@
                 <header>
                     <button name="button_validate_sql_expression" type="object" states="draft"
                             string="Validate SQL Expression" class="oe_highlight"/>
+                    <button name="button_preview_sql_expression" type="object" states="draft" string="Preview Results" attrs="{'invisible': [('field_ids', '!=', False)]}"/>
                     <button name="button_set_draft" type="object" states="sql_valid"
                             string="Set to Draft" groups="sql_request_abstract.group_sql_request_manager"/>
                     <button name="export_sql_query" string="Execute Query" states="sql_valid" type="object" class="oe_highlight"
--- a/sql_request_abstract/models/sql_request_mixin.py
+++ b/sql_request_abstract/models/sql_request_mixin.py
@@ -270,3 +270,9 @@ class SQLRequestMixin(models.AbstractModel):
         """
         self.ensure_one()
         return False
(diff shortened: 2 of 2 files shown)


len-foss (approved):
Cool feature.
Name seems a bit confusing, maybe something like 'test query' or 'preview results'?

legalsylvain:
Hi @len-foss. Thanks for your review. I updated the code, to replace by "Preview Results". 

Note : could you also review trivial related PR ? https://github.com/OCA/reporting-engine/pull/496

thanks.

florian-dacosta (commented):
I am not so sure about this one.
I don't really see the point of preview a part of the query when it is as fast as generate the full file, but why not if you have a use case, but, I think it won't work if query uses variable. (I guess it would fail for now)
You have to manage this case. (simplest would be to just not to allow preview if query uses variables, hiding the button?) Or allowing user to set variables also in this case...

legalsylvain:
Hi @florian-dacosta, I updated the code, hidding the button in case of variables. 
could you update your review ?

florian-dacosta (approved):

legalsylvain:
hi maintainers. Could you merge this one ? thanks !

pedrobaeza (approved):

legalsylvain:
I don't understand why oca-git-bot doesn't merge. This message (https://travis-ci.com/github/OCA/server-tools/jobs/512030152#L1336)
```
 ! [rejected]        HEAD -> 12.0-ocabot-merge-pr-2061-by-pedrobaeza-bump-minor (fetch first)
error: failed to push some refs to 'https://github.com/OCA/server-tools'
hint: Updates were rejected because the remote contains work that you do

hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

I rebased the branch. @pedrobaeza could you try again to merge. (or do you think I have something to fix ?)

pedrobaeza:
There has been an outage of GitHub, Pypi, etc due to a failure in Fastly, and it was due to that. Restarting again:

/ocabot merge minor

PR #2063, opened 2021-04-06
[IMP] base_jsonify: Allow to export many2one and referene field as si…

len-foss:
…mple field

Before this change, the result of the export of a relational field was always a dict or a list of dict.

ex:

<record id="ir_exp_shopinvader_variant_lang" model="ir.exports.line">
  <field name="name">publication_language_id/name</field>
  <field name="alias">publication_language_id/name:lang</field>
  <field name="export_id" ref="shopinvader.ir_exp_shopinvader_variant" />
</record>

will output:
{
    ...
    "publication_language_id" : {
        "lang": "French"
    }
    ...
}

After this change it's now possible to define simple exporter for relational fields where the value into the result will be the display_name or a list of display_name of the related records.

ex:

<record id="ir_exp_shopinvader_variant_lang" model="ir.exports.line">
  <field name="name">publication_language_id</field>
  <field name="alias">publication_language_id:lang</field>
  <field name="export_id" ref="shopinvader.ir_exp_shopinvader_variant" />
</record>

will output:
{
    "lang": "French"
}

(Cherry-pick of 652c0167a after this commit was dropped by #1894's refactoring)
(+ added a test so that it cannot be forgotten again)

### DIFF
--- a/base_jsonify/models/models.py
+++ b/base_jsonify/models/models.py
@@ -48,6 +48,10 @@ class Base(models.AbstractModel):
             # This call also add the tzinfo into the datetime object
             value = fields.Datetime.context_timestamp(self, value)
             value = value.isoformat()
+        elif field.type in ("many2one", "reference"):
+            value = value.display_name if value else None
+        elif field.type in ("one2many", "many2many"):
+            value = [v.display_name for v in value]
         return value
 
     @api.model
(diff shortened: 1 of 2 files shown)


hailangvn (approved):
LGTM

lmignon (approved):

PR #2066, opened 2021-04-09
[13.0][FIX] base_exception: import correctly osv.expression

MiquelRForgeFlow:
When doing `osv.expression`, python assumes that the `__init__.py` file in the `odoo/osv` folder contains a `from . import expression` call, but this is not the case. Thus, the correct way is importing the `expression` instead of the `osv` (also because `import osv` leads to confusion, there is an `osv.py` in the `odoo/osv` folder).

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -7,8 +7,9 @@
 import html
 import time
 
-from odoo import _, api, fields, models, osv
+from odoo import _, api, fields, models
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):

JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
what a :bug:

PR #2071, opened 2021-04-21
[14.0][FIX] upgrade_analysis: merged modules affecting noupdate files

MiquelRForgeFlow:
Due to https://github.com/OCA/server-tools/pull/2057, noupdate data now considers merged modules. But the fix was incomplete, as the diff files between merged modules and local_module were rewritting already generated diff between original remote_module and local_module.

### DIFF
--- a/upgrade_analysis/models/upgrade_record.py
+++ b/upgrade_analysis/models/upgrade_record.py
@@ -5,12 +5,15 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import ast
+import logging
 import os
 
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.modules.module import MANIFEST_NAMES, get_module_path
 from odoo.tools.translate import _
+
+_logger = logging.getLogger(__name__)
 
 
 class UpgradeRecord(models.Model):
@@ -159,7 +162,7 @@ class UpgradeRecord(models.Model):
         # The order of the keys are important.
         # Load files in the same order as in
         # module/loading.py:load_module_graph
-        paths = []
+        files = []
         for key in ["init_xml", "update_xml", "data"]:
             if not manifest.get(key):
                 continue
@@ -167,5 +170,13 @@ class UpgradeRecord(models.Model):
                 if not xml_file.lower().endswith(".xml"):
                     continue
                 parts = xml_file.split("/")
-                paths.append(os.path.join(addon_dir, *parts))
-        return paths
+                try:
+                    with open(os.path.join(addon_dir, *parts), "r") as xml_handle:
+                        files.append(xml_handle.read())
+                except UnicodeDecodeError:
+                    _logger.warning(
+                        "Encoding error: Unable to read %s",
+                        os.path.join(addon_dir, *parts),
+                    )
+                    continue
+        return files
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Will you also revert the functionality of `get_xml_records` to return xml data instead of paths? Otherwise we will have a problem with OpenUpgrade 15.

MiquelRForgeFlow:
> Will you also revert the functionality of get_xml_records to return xml data instead of paths?

I could do that if there was a problem. As the issue of the noupdate data don't affect if we do the 'paths' approach or 'files' approach, for now I decided to maintain the 'paths' approach (in order to include the files that have a different encoding).

> Otherwise we will have a problem with OpenUpgrade 15

Which problem?

StefanRijnhart:
To perform the noupdate data comparison in upgrade_analysis in Odoo 15, it needs this module in version 14 to provide the XML data (the remote paths to the XML data are useless). It's the same discussion as https://github.com/OCA/OpenUpgrade/pull/2481#discussion_r616535367

pedrobaeza:
I also think that providing whole XML data instead of paths is better for isolating each environment (previous version and this one). Think for example with Doodba, that each Odoo is isolated with no access to their local file paths.

MiquelRForgeFlow:
Yeah, but the upgrade_analysis (of v15) will be created doing a forward port of upgrade_analysis (of v14), which will have the same code (despite number version), and thus it will also have the same methods and approach.

StefanRijnhart:
Oh now I see why you think it is not a problem, you are running the two versions side by side on the same machine on the same user account! Not everybody does that.

MiquelRForgeFlow:
> Think for example with Doodba

Okey, that would be a reason, but I don't imagine anybody executing this analysis with doodba haha

MiquelRForgeFlow:
> you are running the two versions side by side on the same machine on the same user account!

Of course. That's full control!

StefanRijnhart:
Please revert to the version that is generic and not depending on your local setup.

pedrobaeza:
@MiquelRForgeFlow what I don't imagine is doing this manually in your local machine without Doodba! You have to comply with both versions requirements, do the setup, etc. It's counter-productive.

As the data transmission way serves for both setups, please go with that standard.

MiquelRForgeFlow:
Ok, reverted everywhere, all it's ready to merge.

MiquelRForgeFlow:
@StefanRijnhart could you check? sure @pedrobaeza is waiting you hehe

upgrade_analysis/models/upgrade_record.py
```diff
@@ -159,13 +159,17 @@ def get_xml_records(self, module):
...
             for xml_file in manifest[key]:
                 if not xml_file.lower().endswith(".xml"):
                     continue
                 parts = xml_file.split("/")
-                paths.append(os.path.join(addon_dir, *parts))
-        return paths
+                try:
+                    with open(os.path.join(addon_dir, *parts), "r") as xml_handle:
+                        files.append(xml_handle.read())
+                except UnicodeDecodeError:
```
StefanRijnhart, line 176: I'm thinking a warning would be good at this point

StefanRijnhart (approved):
~~Just a small non blocking comment.~~  Thanks for the follow-up!

upgrade_analysis/models/upgrade_record.py
MiquelRForgeFlow: Done! :)

PR #2072, opened 2021-04-23
[MIG][12.0] backport base report auto create qweb

NL66278:
Rather a back-port from 14.0 then a migration, to minimize the work to do.


NL66278:
@lfreeke Could you do a functional review of this? @oihane Would you do a technical review?

thomaspaulb:
@NL66278 

- pre-commit issue
- failing test, because `unidecode` cannot be found. I think this might be a case of having to use `external_dependencies_override` such as [here](https://github.com/OCA/server-tools/blob/12.0/setup/sentry/setup.py#L8) or [here](https://github.com/OCA/server-tools/blob/12.0/setup/letsencrypt/setup.py#L10), to specify that the package name is `Unidecode` and not `unidecode`.

NL66278:
@thomaspaulb I think the needs fixing label can be removed now.

thomaspaulb (approved):

PR #2074, opened 2021-04-28
[10.0][FIX] auto_backup: open backup file locally produce empty buffer for backup actions after the first

SimoneVagile:


auto_backup/models/db_backup.py
```diff
@@ -163,14 +163,15 @@ def action_backup(self):
         # Ensure a local backup exists if we are going to write it remotely
         sftp = self.filtered(lambda r: r.method == "sftp")
         if sftp:
-            if backup:
```
SimoRubi: It just seemed reasonable to do the backup once, cache it and send it in all configured locations; why are you now creating the backup again and again for each `sftp` configuration?

auto_backup/models/db_backup.py
```diff
@@ -180,8 +181,8 @@ def action_backup(self):
 
                             # Copy cached backup to remote server
                             with remote.open(
-                                    os.path.join(rec.folder, filename),
-                                    "wb") as destiny:
+                                os.path.join(rec.folder, filename), "wb"
```
SimoRubi: Looks like this isn't changing anything, can you remove it from the commit?

SimoRubi (changes requested):
Please also check why bots checks are different from `10.0` (https://github.com/OCA/server-tools/commit/4c5ad84b600a2120381c54bd3e35f35f15b9b6a5):

- Runbot is 🔴 in `10.0` too, can you fix it here or in a dedicated PR?
- Travis was 🟢 in `10.0` so it should be 🟢 here too.

auto_backup/models/db_backup.py
SimoneVagile: Removed

auto_backup/models/db_backup.py
SimoneVagile: I backported from [https://github.com/OCA/server-tools/commit/52b070e839fd83b60bd1ff4eef797d29bdcb02e7#diff-75853cf520b8f3ccf275c9a73a465d1fddc919bb908fa19627211491deb69f37](https://github.com/OCA/server-tools/commit/52b070e839fd83b60bd1ff4eef797d29bdcb02e7#diff-75853cf520b8f3ccf275c9a73a465d1fddc919bb908fa19627211491deb69f37), I changed from this version, I suppose may be changed for further versions. Correct?

SimoRubi: The changes you linked (https://github.com/OCA/server-tools/commit/52b070e839fd83b60bd1ff4eef797d29bdcb02e7) had a valid reason to re-create the dump at each iteration: the backup format might change between `sftp` configurations, here we have nothing like that and it would be better to keep the old logic _dump once, send everywhere_.

Even with latest changes, it would be better to send the pre-existing backup instead of doing it again.

You can choose if backporting the linked changes entirely with backup formats (that should also fix your issue) or just fix the issue you found

SimoneVagile: I restored the condition, replacing the _with_ statement into the loop, to allow file closing each step.

.travis.yml
```diff
@@ -81,7 +81,7 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
   # Install libraries that require specific development headers, but not during lint test
-  - if ! [ "$LINT_CHECK" = "1" ]; then pip install "pymssql<3.0" MySQL-python pyodbc; fi
+  - if ! [ "$LINT_CHECK" = "1" ]; then pip install pymssql==2.1.5 MySQL-python pyodbc; fi
```
SimoRubi: Better not to use fixed versions because it is a big constraint on other packages
```suggestion
  - if ! [ "$LINT_CHECK" = "1" ]; then pip install "pymssql<=2.1.5" MySQL-python pyodbc; fi
```

auto_backup/models/db_backup.py
```diff
@@ -168,22 +168,22 @@ def action_backup(self):
...
+                    with rec.sftp_connection() as remote:
+                        # Directory must exist
+                        try:
+                            remote.makedirs(rec.folder)
+                        except pysftp.ConnectionException:
+                            pass
 
-                            # Copy cached backup to remote server
+                        # Copy cached backup to remote server
+                        with cached:
```
SimoRubi: I have tried the following that should be close enough to what happens in the module:
```
>>> f = open('draft')
>>> f
<open file 'draft', mode 'r' at 0x7fda9e548db0>
>>> with f:
...     s = f.read()
... 
>>> s
'hi\n'
>>> f
<closed file 'draft', mode 'r' at 0x7fda9e548db0>
>>> with f:
...     s = f.read()
... 
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file
```
And the file is closed the 2nd time I tried to read it, so how is this working?

auto_backup/tests/test_db_backup.py
```diff
@@ -187,6 +175,30 @@ def test_action_backup_all_return(self):
...
+            'sftp_port': '222',
+            'sftp_user': 'tuser2',
+            'sftp_password': 'password',
+            'folder': '/folder/',
+        }
+        rec2 = self.Model.create(new_vals)
+        with mock.patch("%s.search" % class_name):
+            res = self.Model.action_backup_all()
+            self.assertEqual(
+                self.Model.search('id', '=', rec1.id).action_backup(), res
```
SimoRubi: Both `action_backup_all` and `action_backup` always return `None`, so what are you testing here?

Moreover, this test passes even without your edit to the module; if you add a fix it is best to add a test that fails without your fix and passes when your fix is included.

SimoRubi (changes requested):

auto_backup/models/db_backup.py
SimoneVagile: This is what happens when you go crazy using CTRL+c CTRL+v :smile:

SimoRubi (approved):
👍🏻 thanks

SimoneVagile:
@SimoRubi  I run into this error:
```
  File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/base/ir/ir_cron.py", line 125, in _callback
    getattr(model, method_name)(*args)
  File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/auto_backup/models/db_backup.py", line 195, in action_backup_all
    return self.search([]).action_backup()
  File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/auto_backup/models/db_backup.py", line 168, in action_backup
    cached = open(backup)
IOError: [Errno 2] No such file or directory: '<fdopen>'
```

This is caused by file closure after `with open` statement. The file created is temporary, so once it's closed, system deletes it.
I revert code and fixed this with last modifies, functional test is ok.
Can you check?

SimoRubi (changes requested):
> @SimoRubi I run into this error:
> 
> ```
> Traceback (most recent call last):
>   File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/base/ir/ir_cron.py", line 125, in _callback
>     getattr(model, method_name)(*args)
>   File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/auto_backup/models/db_backup.py", line 195, in action_backup_all
>     return self.search([]).action_backup()
>   File "/srv/odoo-10-customer-prod/customer/odoo/lib/python2.7/site-packages/odoo/addons/auto_backup/models/db_backup.py", line 168, in action_backup
>     cached = open(backup)
> IOError: [Errno 2] No such file or directory: '<fdopen>'
> ```
> 
> This is caused by file closure after `with open` statement. The file created is temporary, so once it's closed, system deletes it.
> I revert code and fixed this with last modifies, functional test is ok.
> Can you check?

Travis is failing the test you added

auto_backup/tests/test_db_backup.py
```diff
@@ -187,6 +175,25 @@ def test_action_backup_all_return(self):
...
+            'sftp_port': '222',
+            'sftp_user': 'tuser2',
+            'sftp_password': 'password',
+            'folder': '/folder/',
+        }
+        self.Model.create(new_vals)
+        with self.mock_assets() as assets:
+            with mock.patch('%s.%s.sftp_connection' % (model, "DbBackup")):
+                rec_id.action_backup_all()
+                self.assertEqual(assets['shutil'].copyfileobj.call_count, 2)
```
SimoRubi, line 195: > this test passes even without your edit to the module; if you add a fix it is best to add a test that fails without your fix and passes when your fix is included.

SimoRubi (changes requested):

PR #2080, opened 2021-05-08
[14.0][ADD] base_sequence_option

kittiu:
This module allow user to add optional sequences to some document model.
On which sequence is used, is based on domain matching with document values
(and original sequence will be bypassed).

For example, it is now possible to,

* Avoid using Odoo automatic sequence on invoice and vendor bill, and back to old style ir.sequence.
* Customer payment and vendor payment to run on different sequence.
* Assign different sales order sequence based on customer region.

This is a base module and does nothing by itself. Following are modules
that will allow managing sequence options for each type of documents, I.e.,

* Purchase Order: purchase_sequence_option, https://github.com/OCA/purchase-workflow/pull/1173
* Invoice / Bill / Refund / Payment: account_sequence_option, https://github.com/OCA/account-invoicing/pull/932
* Others: create a new module with few lines of code

**Note:** Partly, this module is intended to fall back to the old style invoice sequence. Would love to hear some comments.

image

image

TODO:

- [x] Test Script


remi-filament (approved):
I have tested this module with account one and it works fine, thank you @kittiu for your work !

base_sequence_option/views/sequence_option_view.xml
```diff
@@ -0,0 +1,123 @@
...
+                        <field name="prefix" />
+                        <field name="suffix" />
+                        <field name="implementation" readonly="1" />
+                    </group>
+                    <group>
+                    </group>
+                </group>
+            </form>
+        </field>
+    </record> -->
```
ps-tubtim: Are you still using this code?

kittiu: OK, I removed it.

ps-tubtim (approved):

alexis-via:
A really great module with a big potential, in particular with account_sequence_option that will allow us to revert to the pre-v14 behavior where the number of invoices/moves where generated from a sequence.
But, after reading the code and made some tests in multi-company, it seems to me that this module doesn't handle multi-company. Obviously, for invoices and other documents, you need a sequence per company. But base.sequence.option and sequence.option are not company dependent. The object sequence.option has a M2O link to ir.sequence and this M2O field is not company dependant. I really don't understand how this module handle multi-company... @kittiu could you tell us how multi-company is supposed to work in this module ?

Another not-important remark : the 2 objects are named "sequence.option" and "base.sequence.option". I would suggest "ir.sequence.option" and "ir.sequence.option.line" (advantage: when you look at a list of tables, the 2 tables are together). But it's your module, so you decide how you name the tables :)

kittiu:
> A really great module with a big potential, in particular with account_sequence_option that will allow us to revert to the pre-v14 behavior where the number of invoices/moves where generated from a sequence.
> But, after reading the code and made some tests in multi-company, it seems to me that this module doesn't handle multi-company. Obviously, for invoices and other documents, you need a sequence per company. But base.sequence.option and sequence.option are not company dependent. The object sequence.option has a M2O link to ir.sequence and this M2O field is not company dependant. I really don't understand how this module handle multi-company... @kittiu could you tell us how multi-company is supposed to work in this module ?
> 
> Another not-important remark : the 2 objects are named "sequence.option" and "base.sequence.option". I would suggest "ir.sequence.option" and "ir.sequence.option.line" (advantage: when you look at a list of tables, the 2 tables are together). But it's your module, so you decide how you name the tables :)
 

@alexis-via thanks for your comment really. You are right, I didn't think about multi-company yet, I will do so, also other of your comment. If you have any good sample on multi-company code, please feel free to share. Thanks!

kittiu:
@alexis-via your comment, fixed.
For multi-company, now each company can have its own set of sequence option.

alexis-via:
Thanks for taking into account my remarks!
I see that you've set company_id as a required field. On ir.sequence, the company_id field is not required; cf https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_sequence.py#L148
If left empty, the sequence is shared between several companies, if set, if is restricted to a single company. Did you do it on purpose because you think sequence options should never be shared between companies ? I haven't thought about the consequences of sharing sequence options between companies, it's just a question.

In the sale or purchase module, the native code generate the sequence by an inherit of the create() before calling super(), cf:
https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py#L501
But, for base_sequence_option to work, you need to have a full recordset. So, if you develop a module sale_sequence_option, do you plan to inherit create(), call super() to create the record set, and then generate the sequence for each recordset and do a write() to update the 'name' field ?

alexis-via:
Sorry, I didn't see your PR for purchase.order; I'll study it and it will certainly answer my second question.

kittiu:
@alexis-via about company_id field, it is on purpose. I think user should make the sequence option for each company. Then, on option line, if user can still share the same sequence_id (that has no company_id). I think this make more organized, and can lead to less error, WDYT?

matteoopenf:
> @alexis-via about company_id field, it is on purpose. I think user should make the sequence option for each company. Then, on option line, if user can still share the same sequence_id (that has no company_id). I think this make more organized, and can lead to less error, WDYT?

I think is now ready to merge

kmatthes:
Hello @alexis-via Are you satisfied with this PR now? :slightly_smiling_face:  Can it be merged?

(@kittiu) We just discovered this PR and the PRs that build on it and would love to see the PRs completed soon. I'll try to organize a review for OCA/account-invoicing#932 asap. (Would be easier with this PR already merged. :sweat_smile: ) Please don't hesitate to ask for additional reviews by mentioning my fellow colleague @alihashaam , who just started to manages a special team that probably still has free capacity for community reviews. :smile:

alexis-via:
@kmatthes I am now focused on my PR account_move_name_sequence https://github.com/OCA/account-financial-tools/pull/1228
so I don't plan to investigate this PR any further.

kmatthes:
> @kmatthes I am now focused on my PR account_move_name_sequence [OCA/account-financial-tools#1228](https://github.com/OCA/account-financial-tools/pull/1228) so I don't plan to investigate this PR any further.

@alexis-via ah, I see. Thx for clarification. Tell us (me +  @alihashaam ) if you need an additional review or any other support. :slightly_smiling_face:

kittiu:
@kmatthes this PR is ready and we also use them in projects. But may be it is up to the maintainer to merge. :)

HaraldPanten (approved):
Tested with https://github.com/OCA/account-invoicing/pull/932 and it seems to be working fine.

HaraldPanten:
@OCA/tools-maintainers I think that this one could be merged in order to continue with other extensions. THX

PR #2082, opened 2021-05-13
[FIX] sentry: Warning when using classes from collections.abc

luisg123v:
The following warning is fixed:

    DeprecationWarning: Using or importing the ABCs from 'collections' instead of
    from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop
    working


luisg123v:
@moylop260 & @fernandahf

Could you review, please?

moylop260 (approved):
👍

PR #2085, opened 2021-05-18
[14.0][MIG] module_analysis

kevinkhao:
trivial migration
manual tests on menus


legalsylvain:
Hi @kevinkhao. Thanks for porting this module.
Could you takes this bugfix also https://github.com/OCA/server-tools/pull/2058 ?

thanks.

module_analysis/models/ir_module_module.py
```diff
@@ -0,0 +1,178 @@
...
+        }, [...]
+        extension: extension of the file, with the '.'
+        data_to_analyse : possible value : code, documentation, empty, string
+        field_name: Odoo field name to store the analysis
+        """
+        return {
+            ".py": {"code": "python_code_qty"},
+            ".xml": {"code": "xml_code_qty"},
+            ".js": {"code": "js_code_qty"},
+            ".css": {"code": "css_code_qty"},
```
MeetKD, line 59: Doest it cover scss too?

kevinkhao: it seems there is about a 50/50 split between CSS and SCSS files in the OCA, so it seems reasonable to add

kevinkhao:
> Hi @kevinkhao. Thanks for porting this module.
> Could you takes this bugfix also #2058 ?
> 
> thanks.

done here https://github.com/OCA/server-tools/pull/2085/commits/fece1ffdf916f84c9449679135cc07abc36b846b

module_analysis/models/ir_module_module.py
kevinkhao: done here https://github.com/OCA/server-tools/pull/2085/commits/ecbfdcbcab9a2182bed5d57c52330bae32e71d85

legalsylvain:
Hi @kevinkhao. Could you rebase ? some commits doesn't belong to the migration work.

Also squash migration 14.0 commits could make the review easier.

thanks !

sebastienbeau:
@kevinkhao  can you rebase ? thanks

kevinkhao:
@legalsylvain sorry this slipped by, done now
@sebastienbeau done
however there is one issue you can see in Travis. module_analysis should be installed last so that the tests can run properly. I am not sure, but I think it would be sufficient to just test one or several base modules and assume the rest are analysed. What do you think ?

legalsylvain:
Hi kevin.
Could you cherry pick this one https://github.com/OCA/server-tools/pull/2250

Regarding tests, no idea. Post install doesnt solve the issue ?

module_analysis/__manifest__.py
```diff
@@ -9,9 +9,9 @@
     " custom modules",
     "author": "GRAP, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-tools",
-    "version": "13.0.1.0.0",
+    "version": "14.0.1.0.0",
     "license": "AGPL-3",
-    "depends": ["base"],
+    "depends": ["base", "base_automation"],
```
legalsylvain: ``base`` is so unnessary.

legalsylvain (approved):
Diff review. 

thanks for porting this module !

PierrickBrun:
@legalsylvain I rebased and removed the `base` dependency FYI

legalsylvain:
here is an error on the CI, due, I think to the magic ``sever_environment`` module. Not sure to understand, maybe change the test (as proposed by @kevinkhao could fix the problem.

regards.

PierrickBrun:
> here is an error on the CI, due, I think to the magic `sever_environment` module. Not sure to understand, maybe change the test (as proposed by @kevinkhao could fix the problem.
> 
> regards.

I added a call to the cron on the beginning of the test, thinking it would resolve the issue (I think it is still a good idea to keep it).

The solution proposed by Kevin Khao is OK for me too but I have no idea how to choose which module gets to be tested, maybe `base_automation` as it is a dependency ?

rvalyi:
@PierrickBrun @legalsylvain @sebastienbeau I rebased and amended the latest commit to test only on base and base_automation modules as suggested by @PierrickBrun. It seems we have a different error than the last time this time (an error with Sentry). However it might be related to module_analysis because I see no such Sentry error with the regular 14.0 branch. Also, it seems this module makes the whole CI terribly slow: like 28 minutes instead of 5 minutes max usually. At the end I think something should be done to avoid running the analysis in the CI environment (test for env variable?) and avoid killing the CI just for this module...  cc @renatonlima

module_analysis/post_init_hook.py
```diff
@@ -2,13 +2,23 @@
...
+                [
+                    "|",
+                    ("state", "=", "installed"),
+                    ("name", "in", ("base", "base_automation", "module_analysis")),
+                ]
+            )
+        else:
+            installed_modules = env["ir.module.module"].search(
+                ["|", ("state", "=", "installed"), ("name", "=", "module_analysis")]
+            )
```
legalsylvain: I don't get that change. 
AFAIU, if it we are in a CI context, the analysis is done on more modules. Is it desired ?

rvalyi: @legalsylvain not my change!  (and yes I don't get it either)
https://github.com/OCA/server-tools/pull/2085/commits/58c2304a349c3899a41d351b5cea8b5440db5054#diff-91be04b85dd85fbd3a212c85ecb10524d6239d5a204189493aa46df065b0562fL12

Spider_Man_meme

legalsylvain: Ahaha ! 

No. 
- I'm talking about that commit https://github.com/OCA/server-tools/pull/2085/commits/204afc6a87a67cf66f3b015749f192fd91a3b7a5 and it's your.
- regarding the "["|", ("state", "=", "installed"), ("name", "=", "module_analysis")]" I don't remember precisely, but I guess that when installing module_analysis module, in the post_init_hook step, the module is not in the installed state. (I think the module is set as installed after the post_init_hook).

legalsylvain: Instead of adding condition in that function, I think we could just remove the call of analyse_installed_modules in the test.
I mean, it's useless to add write a test, if it's to call a function with if==CI inside. Don't you think ?

rvalyi: @legalsylvain in fact the analysis was triggered both by the test and by the post_init hook (and that is why my previous change failed to skip the analysis in the CI context). So I kept an if os.environ.get("CI") test in the _analyse_code method in my last commit. As you can see it succeed in keeping the analysis only for 3 modules max on the CI (and it's probably better to still limit to these 3 modules in the test if your run the test without setting your local CI environment variable.

rvalyi:
@PierrickBrun @legalsylvain @sebastienbeau @renatonlima at the end I got the tests working in the decent manner as you can see in the log.

Still there is a cryptic error with the Sentry module at the end of the run and this is not the module_analysis fault:


Without understanding the root of this Sentry error, I tried to pin versions that were differing from the latest successful CI run on the 14.0 branch with that kind of commit (removed later) https://github.com/OCA/server-tools/pull/2085/commits/88af573617ee9e07a0bbd7769862c50781d4beae but it didn't fix the problem, so I'm giving up on this Sentry error at this point. cc to the Sentry maintainer and latest committers @barsi, @naglis, @versada, @moylop260, @fernandahf @paradoxxxzero @njeudy any idea?

thomaspaulb:
Superseded by https://github.com/OCA/server-tools/pull/2656

PR #2094, opened 2021-05-28
[14.0][FIX] base_report_auto_create_qweb, conflict test report name

kittiu:
If module `product` is installed, the test will not pass, because keyword "report_test" in https://github.com/odoo/odoo/blob/14.0/addons/product/views/product_templates.xml#L11

Which result in test error,

```
  File "/opt/odoo/auto/addons/base_report_auto_create_qweb/tests/test_base_report_auto_create_qweb.py", line 42, in test_creation_html
    self.assertEqual(view_num, 1, "Only one view must be created.")
AssertionError: 2 != 1 : Only one view must be created.
```

Just ensure that, the report name in test is unique.


kittiu:
cc @pedrobaeza

pedrobaeza (approved):
Seems correct

newtratip (approved):

ps-tubtim (approved):

PR #2098, opened 2021-06-01
[IMP] database_cleanup : delete also obsolete database views

legalsylvain:
**Context**

- database migrated with openupgrade from 8.0 to 12.0.
when trying to cleanup the database, some tables are not dropable, for exemple ``stock_quant_move_rel`` with the following errors : 

```cannot drop table stock_quant_move_rel because other objects depend on it
DETAIL:  view stock_history depends on table stock_quant_move_rel
```

this PR improve the concept purge tables to also purges obsoletes views.

**sample**

image

**Note**

Materialized views are still not handled.

### DIFF
--- a/database_cleanup/models/purge_tables.py
+++ b/database_cleanup/models/purge_tables.py
@@ -4,6 +4,13 @@
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError
 from ..identifier_adapter import IdentifierAdapter
+from psycopg2.extensions import AsIs
+
+
+_TABLE_TYPE_SELECTION = [
+    ('base', 'SQL Table'),
+    ('view', 'SQL View'),
+]
 
 
 class CleanupPurgeLineTable(models.TransientModel):
@@ -13,6 +20,10 @@ class CleanupPurgeLineTable(models.TransientModel):
 
     wizard_id = fields.Many2one(
         'cleanup.purge.wizard.table', 'Purge Wizard', readonly=True)
+
+    table_type = fields.Selection(
+        string='Table Type', selection=_TABLE_TYPE_SELECTION
+    )
 
     @api.multi
     def purge(self):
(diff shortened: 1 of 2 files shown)


legalsylvain:
CC : original authors : @StefanRijnhart, @pedrobaeza, @hbrunn

rvalyi (approved):

hbrunn (approved):

StefanRijnhart (approved):

database_cleanup/models/purge_tables.py
```diff
@@ -6,6 +6,12 @@
 from ..identifier_adapter import IdentifierAdapter
 
 
+_TABLE_TYPE_SELECTION = [
+    ('BASE TABLE', 'Base Table'),
+    ('VIEW', 'View'),
```
pedrobaeza: Please normalize DB selection value to lower and without spaces, and also be more specific in the label for avoiding misunderstandings with `ir.ui.view` records:
```suggestion
    ('base', 'SQL Table'),
    ('view', 'SQL View'),
```
I know you are using this value for using such `SQL code` in the query, but you can perfectly put there both values without using this list.

database_cleanup/models/purge_tables.py
```diff
@@ -65,8 +75,12 @@ def purge(self):
 
             self.logger.info(
                 'Dropping table %s', line.name)
```
pedrobaeza: Change this logger accordingly the type.

pedrobaeza (changes requested):

GSLabIt:
@legalsylvain @pedrobaeza is this ready to merge?

StefanRijnhart:
I see @pedrobaeza 's comment was followed up, so I consider this ready for merge indeed.

/ocabot merge minor

legalsylvain:
Thanks for merging.

PR #2108, opened 2021-06-18
[14.0][ADD] auditlog_report

kittiu:
cc @sebalix @hbrunn @StefanRijnhart 

We need the view auditlog.report that join log and lines so we can see detail in one table.  I.e., from a PO, to see log of both PO and PO Line.

image

Note: In fact, this is just the database view that is added. Should this be included in auditlog module?


sebalix:
IMO it makes sense to have such handy views included in the base module.

kittiu:
> IMO it makes sense to have such handy views included in the base module.

Thanks I will.

kittiu:
Move to https://github.com/OCA/server-tools/pull/2109

PR #1659, opened 2019-09-02
[12.0] Add module sql_export_excel

florian-dacosta:
This new module extend the sql_export module and add the possibility to extract the data in an excel file.
It also make possible to insert the result of the sql query directly in a excel template file. Which can be usefull when the result of the query is used in other sheets with a lot of calculation for example.

I did a refactore in sql_export to allow to have different format (csv and excel).
And also a very small refactore on sql_request_abstract to allow to add header or not in the excel file.
@legalsylvain @bealdav @aheficent

PR #2115, opened 2021-06-23
[14.0][mig]sql_export_excel

hkapatel-initos:
Dependent: sql_request_abstract 
[Fix]: Add header feature in latest v14. #2144


hkapatel-initos:
@florian-dacosta  It would be great if we could get the reviews here.

hkapatel-initos:
@atchuthan it would be great if we could get the reviews here.

hkapatel-initos:
@dsolanki can you please review this pr?

fshah-initos (approved):
LGTM

sql_export_excel/models/sql_export.py
```diff
@@ -69,20 +71,16 @@ def check_column_position(self):
...
         if self.file_format == "excel":
             return "xlsx"
         else:
             return super()._get_file_extension()
 
-    @api.multi
     def excel_get_data_from_query(self, variable_dict):
         self.ensure_one()
-        res = self._execute_sql_request(
-            params=variable_dict, mode="fetchall", header=self.header
```
florian-dacosta: It is not ok to remove the header parameter here.
If you does it, the header field become useless, but more importante, the whole feature allowing to insert data in an existing xls file (with a lot of formulas for instance) does not work anymore.

This header feature need a modification in `sql_request_abstract` module, that were added here : https://github.com/OCA/server-tools/pull/1659/files

I guess, these modifications are not in version 14, because the port (to v13) was initiated before the PR in v12 was merged.
I believe, it should be added back in this PR to avoid loosing the feature.

hkapatel-initos: Hello, @florian-dacosta 
As you know the header parameter is removed from the dependent module, 
can I change it in the dependent module, or can you please suggest to me another good way to add a header feature?

florian-dacosta: You should add it in the dependent module.
It does not really have been removed from dependent module, it is kind on an omission during the migration, so it should be added back. You can do this in this PR IMHO

hkapatel-initos: suggestions are implemented can you please re-review it

sql_export_excel/models/sql_export.py
```diff
@@ -69,15 +72,87 @@ def check_column_position(self):
...
     def _get_file_extension(self):
         self.ensure_one()
         if self.file_format == "excel":
             return "xlsx"
         else:
             return super()._get_file_extension()
 
-    @api.multi
+    # API Section
+    def _execute_sql_request(
```
florian-dacosta: You can't just copy past the whole method.
The header modification should be done in base module ( `sql_request_abstract`) also because it could be re-used by another module.
Maybe I was not specific enough, I think you can do the header modification in this PR, but it should be done in `sql_request_abstract` directly, the same way it was done in v12, in file  sql_request_abstract/models/sql_request_mixin.py (https://github.com/OCA/server-tools/pull/1659/files).

sql_export_excel/tests/test_sql_query_excel.py
```diff
@@ -41,12 +41,7 @@ def test_excel_file_generation(self):
                 "sql_export_id": query.id,
             }
         )
-        workbook = self.get_workbook_from_query(wizard)
-        ws = workbook.active
-        # Check values, header should be here by default
-        self.assertEqual(ws.cell(row=1, column=1).value, "firstcol")
-        self.assertEqual(ws.cell(row=2, column=1).value, "testcol1")
```
florian-dacosta: I don't see any reason to remove these lines from the test.
If it fails, it means there is a problem with the module

sql_export_excel/models/sql_export.py
hkapatel-initos: The suggestion is implemented can you please re-review it

florian-dacosta:
Thanks a lot, seems good.
Let's wait the other PR to be merged to see if the tests passed but it should be good!

florian-dacosta:
@hkapatel-initos The PR https://github.com/OCA/server-tools/pull/2144 has been merged, could you rebase to make the tests pass?

hkapatel-initos:
> @hkapatel-initos The PR #2144 has been merged, could you rebase to make the tests pass?

Done

florian-dacosta (approved):

hkapatel-initos:
Hello maintainers,
Can you please try to merge it?

atchuthan (approved):

CasVissers-360ERP (approved):

PR #2125, opened 2021-06-29
[14.0][FIX+IMP] base_technical_user: Multi company context

rousseldenis:


rousseldenis:
@simahawk @sebastienbeau

rousseldenis:
The default behaviour is problematic for two reasons:

- First, the company should be taken through `self.env.company` as user.company_id is now the default company for the user (and not the current one).
- Secondly, the actual code does not allow to do `recordset.with_company(B).sudo_tech()....` as allowed_company_ids in context can be filled in with current company (and not the one of the tech user - which is unique)

carlosdauden (approved):

base_technical_user/tests/test_sudo_tech.py
```diff
@@ -12,12 +12,23 @@ class SudoTechCase(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super(SudoTechCase, cls).setUpClass()
+        cls.company_2 = cls.env["res.company"].create(
```
simahawk, line 15:
```suggestion
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True, no_reset_password=True))
        cls.company_2 = cls.env["res.company"].create(
```
then you can remove the other `with_context` calls

base_technical_user/models/models.py
```diff
@@ -12,9 +12,21 @@ class Base(models.AbstractModel):
 
     def sudo_tech(self, raise_if_missing=False):
         self_sudoer = self
-        tech_user = self.env.user.company_id.user_tech_id
+        tech_user = self.env.company.user_tech_id
         if tech_user:
             self_sudoer = self.with_user(tech_user.id)
+            # We restrict the allowed companies to the one of the tech user
+            allowed_company_ids = self.env.context.get("allowed_company_ids")
+            # TODO: Is any(...) part necessary as we can consider company should always be
```
simahawk, line 20: if the check is not necessary then even the check on `allowed_company_ids` seems useless because it comes from the the same context of `self` :thinking:

rousseldenis: In allowed_company_ids, you can have several companies depending on e.g. the user interface behaviour.

The problem is this check: https://github.com/odoo/odoo/blob/14.0/odoo/api.py#L585

As user that launches action can have access to company A and B. But, tech user has only access to company A.
So, allowed_company_ids comes with [A, B] and raises.

Maybe, we can limit the checks indeed and only force the value of allowed_company_ids. What do you think ?

simahawk: I see, thanks for the detail. Yes, I think it makes sense to simply clean them up and keep only the ones available.

acsonefho (approved):
That make sense!

acsonefho:
@OCA/server-environment-maintainers Do you have super-power here to merge it? Please :pray:

PR #2137, opened 2021-07-11
[12.0][IMP] bus_alt_connection

nilshamerlinck:
Allow to rely on a different database then default postgres for longpolling notifications


thomaspaulb:
@nilshamerlinck Since this is a quite technical module of which you know the most, don't you want to become `maintainer` of it? Then I'd more readily merge in your improvements.

PR #2142, opened 2021-07-15
[12.0][MIG][IMP] base_import_security_group

kirca:
Migration and improvements:

+ Add the possibility to restrict import for specific model (ir.model).
+ Change the implementation to server-side by using the import
attribute on the view definition (backport from v13).


base_import_security_group/models/models.py
```diff
@@ -0,0 +1,72 @@
...
+    _inherit = 'ir.model'
+
+    import_group_id = fields.Many2one(
+        comodel_name='res.groups')
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
+
+    def allow_import(self):
```
thomaspaulb: Not blocking but in general functions that dont have to be called directly through RPC, should be prefixed with `_`

thomaspaulb (approved):

thomaspaulb:
@kirca Nice work! Could you do a rebase, so that tests are run?

kirca:
@thomaspaulb thanks for resurrecting this PR :). The rebase is done and the tests are passing

thomaspaulb:
Nice! Then we need just another review, perhaps a functional review from you @Saran440 ?

PR #2143, opened 2021-07-16
[14.0][MIG] base_custom_info

kittiu:
Standard Migration, ready for review.


base_custom_info/tests/test_partner.py
```diff
@@ -117,7 +117,7 @@ def test_owner_id(self):
         """Check the computed owner id for a value."""
         self.set_custom_info_for_agrolait()
         self.assertEqual(
-            self.agrolait.mapped("custom_info_ids.owner_id"), self.agrolait
+            set(self.agrolait.mapped("custom_info_ids.owner_id")), set(self.agrolait)
```
kittiu, line 120: Other adjustment is sound to me (as Odoo logs more complains)

This one here, is it a bit strange to me. If anyone knows why, please let me know.
Normally, `mapped()` function should already return just 1 object !!!
```
print(self.agrolait.mapped("custom_info_ids.owner_id"))
[res.partner(10,), res.partner(10,), res.partner(10,), res.partner(10,), res.partner(10,)]
```

So, I assume this as new thing on v14 when running test. So I tested with normal data (in the test script),

So, I test with other model,
```
sale = self.env.ref("sale.sale_order_7")

print(sale.order_line)
sale.order.line(14, 15, 16, 17)

print(sale.mapped("order_line.product_id")
product.product(31,)
```

cc @pedrobaeza @Yajo

pedrobaeza: Remove the `mapped` to see if it returns just one.

kittiu:
```
  File "/opt/odoo/auto/addons/base_custom_info/tests/test_partner.py", line 119, in test_owner_id
    print("-------------->", self.agrolait.custom_info_ids.owner_id)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 963, in __get__
    record.ensure_one()
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4987, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: custom.info.value(506, 507, 508, 509, 510)
```

pedrobaeza: The problem should be due to the field being `Reference` instead of `Many2one`

kittiu: > Remove the `mapped` to see if it returns just one.

Thanks @pedrobaeza for a super quick response once again :).

kittiu: OK, then, this fix should be fine. Also just for the record, then this is the new behavior of `Reference` field in v14, as I also test in v13 without issue.

fshah-initos (approved):
Code review and functional tests. LGTM

dsolanki-initos:
it would be nice if we get reviews here

dsolanki-initos:
it would be nice if we get reviews here

dsolanki-initos:
it would be nice if someone add reviews here.

dsolanki-initos:
it would be nice if someone add reviews here.

dsolanki-initos:
it would be nice if someone add reviews here.

dsolanki-initos:
it would be nice if someone add reviews here.

dsolanki-initos:
it would be nice if someone add reviews here.

dalonsod:
Tested on runbot 👍

dalonsod (approved):

PR #2144, opened 2021-07-19
[Fix]: Add header feature in latest v14.

hkapatel-initos:
Add header feature in latest v14

### DIFF
--- a/sql_request_abstract/__manifest__.py
+++ b/sql_request_abstract/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "SQL Request Abstract",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
     "author": "GRAP,Akretion,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-tools",
     "license": "AGPL-3",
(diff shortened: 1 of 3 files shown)


legalsylvain (approved):
Lgtm. Where is it used ?

hkapatel-initos:
> Lgtm. Where is it used ?

This feature is used in sql export_excel #2115 module the header parameter was removed but it is added back in this PR to avoid losing the feature.

florian-dacosta (approved):
@legalsylvain 
It was initially added here in v12 : https://github.com/OCA/server-tools/pull/1659
But, the before it was merged, the v13 PR of sql_request_abstract was created and this small addition was lost.

The goal is for sql_export_excel to be able to add or not the header in the generated xlsx file.
(It can be usefull not not to add it because the sql_export_excel file allow to inject the result of the query inside an existing xlsx file/template)

legalsylvain:
Hi @florian-dacosta Thanks for the explanation.

/ocabot merge patch

legalsylvain:
> Sorry @legalsylvain you are not allowed to merge.

that's not very kind of you, dear bot.

Could you approve https://github.com/OCA/server-tools/pull/2146 ?

dsolanki-initos (approved):
LGTM

sql_request_abstract/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "SQL Request Abstract",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
```
legalsylvain, line 7: for your information, that change is not necessary, as  the  oca bot bump the version.

PR #2146, opened 2021-07-20
14.0: sql_request_abstract : add legalsylvain as maintainers

legalsylvain:
trivial PR. adding me as maintainers. (I'm the original co-author of the module)

CC : @florian-dacosta

### DIFF
--- a/sql_request_abstract/__manifest__.py
+++ b/sql_request_abstract/__manifest__.py
@@ -6,6 +6,7 @@
     "name": "SQL Request Abstract",
     "version": "14.0.1.0.0",
     "author": "GRAP,Akretion,Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
     "website": "https://github.com/OCA/server-tools",
     "license": "AGPL-3",
     "category": "Tools",


florian-dacosta (approved):

sebastienbeau (approved):

legalsylvain:
@sebastienbeau : this merge request failed due to ``letsencrypt`` error due to connexion error.
maybe relaunching the order could fix the problem...

https://app.travis-ci.com/github/OCA/server-tools/jobs/525671402#L2552

sebastienbeau:
I have relaunched the job

sebastienbeau:
note let'sencrypt module should use a mock to avoid such issue

legalsylvain:
> note let'sencrypt module should use a mock to avoid such issue

@NL66278, what do you think about that point ?

NL66278:
@legalsylvain I did not see that letsencrypt was the culprit here. In general I am in favour of mocking external servers. I realize that you then do not test the stability of those servers, but IMHO that is not the purpose of unit tests. Anyway letsencrypt does use mock, but I have not studied the code enough to check whether no testing with external services is done at all.

PR #2152, opened 2021-08-01
[14.0][FIX] base_exception, error if rule's description is null

kittiu:
If I create an exception rule and leave description null, when use it, I got the error on click close.

cc @MiquelRForgeFlow @rvalyi 

image


MiquelRForgeFlow (approved):
LGTM

Saran440 (approved):
Functional Test :+1:

PR #2164, opened 2021-08-28
[12.0][ADD] attachment_delete_restrict

yostashiro:
Odoo out-of-the-box allows all the internal users to delete any attachments, which is problematic when attachments play some important roles in the workflow/recordkeeping.

This module provides the ability to restrict the deletion of the attachments to certain users per related model.


legalsylvain:
Hi @yostashiro. Thanks a lot for sharing this module. I have a similar need and was thinking about such module. 
I like the restriction by model, because indeed it depends on items and user can maybe delete purchase attachment, but should not delete invoicing attachment.

However, I think that the restriction by ``res.users`` field is not the best approach. For an instance with a few users, it's ok, but this design does'nt scale for instances with a lot of users. A design per groups could be better in my opinion. If I want to allow my accountant to delete attachment for invoices, it's a mess if I have to select all the users, and then add new users each time a new accountant is in my company. What do you think ? I think replacing ``delete_attachment_user_ids`` by ``delete_attachment_group_ids`` could be more user friendly.

kind regards.

yostashiro:
@legalsylvain Thanks for your comments! I actually compared between the options of `delete_attachment_user_ids` and `delete_attachment_group_ids` before developing the module, and as you know I decided to go for the former because I couldn't imagine there would be real need to grant the permissions per group (but I might be wrong with this).

IMO, if the delete is restricted for a model, that's in the context of risk management which usually requires stricter and more specific control over who should have the delete permission than how the Odoo standard groups are designed. For example, there should maybe just one or at most two people from the accounting department who should have the delete permission of the attachments, and they may not even be billing managers. In that case, controlling the delete permission by groups may make the configuration quite cumbersome (you'd need to create quite a few 'xxx Attachment Delete Permission' groups and assign users to each of them).

Would be nice to hear more opinions on this. I will consider changing the design if control by groups seems more reasonable.

legalsylvain:
Hi. Thanks for your deep answer. Well i think that groups can cover 90% of the use case and in 10% doens't work. That case, create some custom groups is a solution. (Maybe not the  best)

I still think that groups should Be implemented.
Why not implementing both ? Then "can delete = current user In user_ids OR current user is member of any group_ids"
What do you think ?

legalsylvain:
Hi. Thanks for your deep answer. Well i think that groups can cover 90% of the use case and in 10% doens't work. That case, create some custom groups is a solution. (Maybe not the  best)

I still think that groups should Be implemented.
Why not implementing both ? Then "can delete = current user In user_ids OR current user is member of any group_ids"
What do you think ?

yostashiro:
> Why not implementing both ?

@legalsylvain Yes, I listened to you. 😉

kakurai8 (approved):
As for the functional review, it is fine.

andhit-r (approved):
Functional review

PR #2165, opened 2021-08-30
[14.0][MIG] base_model_restrict_update

kittiu:


kittiu:
@yostashiro I think the module works for me. But I am coming from the opposite direction, so there are some usability issue for me.

Current behavior:
* Define some model to restrict, then, some user to by pass that restriction.

What we need, is simply to define some user to be restricted to all models. So, by current module, we need to,
* Define all model to be restricted, then, set all users to by pass the restriction first
* Then set some user to NOT by pass that restriction (to be auditing user).

Which mean, we will to do a lot of configuration in order to apply for one user.

So, I am thinking about following enhancement,

---

**Option 1:**

In general configuration, under model restriction, option to choose either mode,
1. Model restriction (restrict some model, set some user to by edit) -- (current behavior)
2. User restriction (no model restriction,  set some user to restrict)

(1) is exactly the same as now.
(2) if set, system will 
  * In Model window, system will set True to all model. And will invisible the field `Update Restrict Model`
  * In User window, system will set `Unrestrict` to True to all user and visible it. It will visible the opposite field `Restrict User`

**Option 2:**

This option, there are no alternate behavior. But we can add helper action to ease thing up.

* In Model tree view, add the action "Restrict Model(s)" and "Unrestrict Model(s)"
* In User tree view, add the action "Restrict User(s)" and "Unrestrict User(s)"

---

Personally, for my use case, I would love option 1. But I also know that it will add new behavior to the module.

WDYT?

Note: Can I add your name as the maintainers of this module?

yostashiro:
> **Option 1:**
> 
> In general configuration, under model restriction, option to choose either mode,
> 
> 1. Model restriction (restrict some model, set some user to by edit) -- (current behavior)
> 2. User restriction (no model restriction,  set some user to restrict)
> 
> (1) is exactly the same as now.
> (2) if set, system will
> 
> * In Model window, system will set True to all model. And will invisible the field `Update Restrict Model`
> * In User window, system will set `Unrestrict` to True to all user and visible it. It will visible the opposite field `Restrict User`

Thanks @kittiu !  I think this option would be nice.  The points 1 and 2 are to handle two different use cases and I believe the enhancement for the point 2 is a nice addition.

> Note: Can I add your name as the maintainers of this module?

I don't feel perfectly qualified to be a maintainer if it starts to involve more of low-level stuff, but I don't mind being one if that helps.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,35 @@
...
+    _inherit = "ir.model.access"
+
+    @api.model
+    @tools.ormcache_context(
+        "self._uid", "model", "mode", "raise_exception", keys=("lang",)
+    )
+    def check(self, model, mode="read", raise_exception=True):
+        res = super(IrModelAccess, self).check(model, mode, raise_exception)
+        if self._uid == 1:
+            return True
```
yostashiro:
```suggestion
        if self.env.su:
            return True
        res = super().check(model, mode, raise_exception)
```
This may be slightly better.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,35 @@
...
+        self._cr.execute(
+            "SELECT restrict_update FROM ir_model WHERE model = %s", (model,)
+        )
+        query_res = self._cr.dictfetchall()[0]
```
yostashiro:
```suggestion
        query_res = self._cr.dictfetchone()
```

yostashiro (commented):
Just some minor suggestions.

kittiu:
@yostashiro I have pending test. Can you see if this is ok.

Selection_098

Selection_097

yostashiro:
> In general configuration, under model restriction, option to choose either mode,

@kittiu Sorry I wasn't paying enough attention to this sentence.  I thought these two options could co-exist and no need to force the user to select either one.  Do you see some problems with having them both enabled?

kittiu:
> > In general configuration, under model restriction, option to choose either mode,
> 
> @kittiu Sorry I wasn't paying enough attention to this sentence. I thought these two options could co-exist and no need to force the user to select either one. Do you see some problems with having them both enabled?

Yes, because, when choose the option 2 User Restriction, system will set restriction to all models and allow edit to all user at first place. And then, allow admin to select just some user to take out that permission.

https://github.com/OCA/server-tools/pull/2165/files#diff-3c79753bce60a251eb71c2e5e4aff80ef28a32299d8d551bcfd6306c65641ebeR51-R56

This is, of course, to preserve the way your original code works.

So, I see the two option are coming from different perspective. I can't see how it can coexists. May be you have better approach, you can describe if you see otherwise.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,35 @@
...
+            query_res["restrict_update"]
+            and mode != "read"
+            and not self.env.user.unrestrict_model_update
+        ):
+            if raise_exception:
+                raise AccessError(
+                    _("You are only allowed to read this record. (%s - %s)")
+                    % (model, mode)
+                )
+            return False
```
yostashiro, line 34:
```suggestion
        if mode == "read":
            return res
        self._cr.execute(
            "SELECT is_readonly_user FROM res_users WHERE id = %s", (self.env.uid,)
        )
        user_query_res = self._cr.dictfetchone()
        if user_query_res["is_readonly_user"]:
            if raise_exception:
                self._raise_access_error()
            return False
        self._cr.execute(
            "SELECT restrict_update FROM ir_model WHERE model = %s", (model,)
        )
        model_query_res = self._cr.dictfetchone()
        if (
            model_query_res["restrict_update"]
            and not self.env.user.unrestrict_model_update
        ):
            if raise_exception:
                self._raise_access_error()
            return False
```

@kittiu What I had in mind is something like this, given that `is_readonly_user` is newly added to res.users.  Does this make sense?

kittiu: IC what you mean, is this correct?

* If user is marked as "readonly", regardless of any model restriction or user permit, just don't allow update/create/delete.

In this case, there is nothing to do with model restriction, right? (what I propose was to still use the model restriction ways, but yes, it could be too restricted).

I would think, we should have another module that do only this, i..e, `base_user_readonly`. This will be simple module, which of course can work along side with `base_module_restrict_update`.

What do you think?

yostashiro: You understand me correctly, Kitti.

> I would think, we should have another module that do only this, i..e, `base_user_readonly`. This will be simple module, which of course can work along side with `base_module_restrict_update`.
> 
> What do you think?

While I think it makes sense to create a separate module `base_user_readonly`, I can also think of a couple of benefits to keep both functions in one module:

1. It makes it easier to avoid test conflicts.
2. Restrict/unrestrict toggle button can be hidden when user is a readonly user.

That said I do not have strong preferences.

kittiu: OK Yoshi. You are the original author, I would go by your will :).
For me, I am ok both way.
I will create another PR, and let you know when it is ready.

kittiu:
Moved to https://github.com/OCA/server-tools/pull/2169

PR #2167, opened 2021-09-03
[MIG][12.0] backport base_changeset

hbrunn:


hbrunn:
he @OCA-git-bot, this addon [has a maintainer](https://github.com/OCA/server-tools/pull/2167/commits/fb135a3236bf0a442762e88678fa3048df204415#diff-d342be9a2d196bbde132681a41efce8147ffdd1c9d4894b08b26f4a5186e7331R10) - are you reading the files wrongly?

hbrunn:
@astirpe @dsolanki-initos I think commits b6998d0 and ea1c0ce are relevant for later versions too, you should be able to just cherry-pick them

astirpe:
Thank you @hbrunn !

astirpe (approved):

StefanRijnhart:
Looks like this module's tests fail
```
2021-10-28 14:27:54,492 7531 ERROR openerp_test odoo.sql_db: bad query: INSERT INTO "mail_followers" ("id", "partner_id", "res_id", "res_model") VALUES (nextval('mail_followers_id_seq'), 2, 82, 'res.partner') RETURNING id

ERROR: duplicate key value violates unique constraint "mail_followers_mail_followers_res_partner_res_model_id_uniq"

DETAIL:  Key (res_model, res_id, partner_id)=(res.partner, 82, 2) already exists.

2021-10-28 14:27:54,504 7531 ERROR openerp_test odoo.addons.base_changeset.tests.test_changeset_flow: ERROR
```

hbrunn:
now without breaking when mail is installed

base_changeset/models/record_changeset.py
```diff
@@ -0,0 +1,169 @@
...
+        readonly=True,
+    )
+    change_ids = fields.One2many(
+        comodel_name="record.changeset.change",
+        inverse_name="changeset_id",
+        string="Changes",
+        readonly=True,
+    )
+    date = fields.Datetime(
+        string="Modified on", default=fields.Datetime.now(), index=True, readonly=True
```
StefanRijnhart: Could it be that the callable was intended?
```suggestion
        string="Modified on", default=fields.Datetime.now, index=True, readonly=True
```

astirpe:
@hbrunn maybe it is worth to backport https://github.com/OCA/server-tools/pull/2184 within this PR as well, what do you think?

hbrunn:
I totally think so, but I'm not sure if this should be in this PR or in a subsequent one

base_changeset/models/record_changeset.py
hbrunn: oh yes, this seems wrong, thanks! @astirpe are we missing something fancy?

hbrunn:
@StefanRijnhart fixed your last point + squashed commits

StefanRijnhart (approved):
Thanks!

PR #2169, opened 2021-09-05
[14.0][MIG] base_model_restrict_update

kittiu:
* Migration to v14
* Add option to set Readonly User to all models

Selection_106


kittiu:
@Yoshiro009 I also do a bit of reorganize code, please feel free to comment.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -17,19 +17,37 @@ def check(self, model, mode="read", raise_exception=True):
...
+            "SELECT restrict_update FROM ir_model WHERE model = %s", (model,)
+        )
+        query_res = self.env.cr.dictfetchone()
+        if query_res["restrict_update"] and not self.env.user.unrestrict_model_update:
+            return True
+        return False
+
+    @api.model
+    def _readonly_exclude_models(self):
+        """ Models updtate/create by system, and should be excluded from checking """
```
eLBati, line 52: I'm not sure, why we need this?

eLBati:
@kittiu thanks!
Backporting `Add option 'Readonly User' to all models` to 12.0 with https://github.com/OCA/server-tools/pull/2188

base_model_restrict_update/models/ir_model_access.py
kittiu: Because res.user itself will be updated when switch back and forth, does it make sense?

eLBati: :thinking:  could you describe a use case where this is needed?

kittiu:
> @kittiu thanks! Backporting `Add option 'Readonly User' to all models` to 12.0 with #2188

You can try to remove those code. And switch user as read-only back and forth. I remember it does update res.users table and got some error.

base_model_restrict_update/models/ir_model_access.py
eLBati: I tried removing it https://github.com/OCA/server-tools/pull/2188/commits/607eee8f12764bc1e594a01ab92cf624e6f34860 and did not get errors:

read

kittiu: Oh ok. May be it was what I did the first iteration. I will review and if ok fix that asap. Sorry for any misleading.

kittiu: Done remove the unused method. thanks for your review!

dsolanki-initos (approved):
Code review and Functional tests LGTM

dsolanki-initos:
@eLBati  can you please re-check it?

eLBati:
I did not test on v14 but we deployed in production the [12.0 version](https://github.com/OCA/server-tools/pull/2188). I don't know if this counts as an approval

dsolanki-initos:
Hello, @TimLai125 it would be great if you add a review here.

dsolanki-initos:
Hello, @TimLai125 it would be great if you add a review here.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,42 @@
...
+    @api.model
+    @tools.ormcache_context(
+        "self._uid", "model", "mode", "raise_exception", keys=("lang",)
+    )
+    def check(self, model, mode="read", raise_exception=True):
+        if self.env.su:
+            return True
+        res = super().check(model, mode, raise_exception)
+        if mode != "read" and raise_exception:
+            if self._test_readonly(model) or self._test_restrict_update(model):
```
yostashiro, line 26: @kittiu Could we exclude some models from this checking (e.g. a read-only user may want to create a new custom filter)? ref. https://gist.github.com/moylop260/43748e848dfc9916fd2f0fd51a69a12a#file-read_only_user-py-L11-L18

Thank you very much for the nice addition.

kittiu: Hi Yoshi, let's discuss together what should be excluded. I am actually ok with your proposal.
(this means bring back https://github.com/OCA/server-tools/compare/44005c10f2854150dd80755ed49852d8ff91bafb..130ca667f8f9c5c4b855e06d35fcf1ac07a99024)

At first I think readyonly is readonly. But it also make sense for, i.e., readonly auditor to have own filter.

@dsolanki-initos  @eLBati what is your opinion on this? If we can't tell what else, we can start with only ir.filters and add more excluded model later if needed.

yostashiro: One way to go, I think, is to get a list of models through raw SQL based on create access = true and group is not set, which should be how @moylop260 came up with the list in the gist.

There are 161 such models in 14.0 EE as I look in runbot.
image

eLBati: @yostashiro @kittiu thanks, it is clear the purpose of `_readonly_exclude_models` now.
I restored it and refactored to dinamically retrieve model names, see https://github.com/OCA/server-tools/pull/2188/commits/c69eba0fd0f4497c694fd7dbe16f56fff795acf7 (in PR for v12 https://github.com/OCA/server-tools/pull/2188 )

kittiu: @eLBati I saw your excluded model.
```
        return self.sudo().search([
            ("group_id", "=", False),
            "|", ("perm_write", "=", True),
            "|", ("perm_create", "=", True), ("perm_unlink", "=", True)
        ]).mapped("model_id.model")
```
Not sure what is the rationale. If you find all ACL group_id = False with updatable flag, that will open a lot of model for update, isn't it? While, I think we have this module to restricted those models.
Please feel free to discuss, I never put this to real use and may not know exact problem.

eLBati: @kittiu I retrieve all the access rules not associated to a group where create, write or unlink is allowed.
That is, models where everyone should be able to create, write or delete.

On a v12, this is what it returns:
```
>>> env["ir.model.access"].search([
...     ("group_id", "=", False),
...     "|", ("perm_write", "=", True),
...     "|", ("perm_create", "=", True), ("perm_unlink", "=", True)
... ]).mapped("model_id.model")
['ir.ui.view.custom', 'res.users.log', 'report.layout', 'web_editor.converter.test', 'web_editor.converter.test.sub', 'decimal.precision.test', 'ir.attachment.add_url', 'mail.thread', 'publisher_warranty.contract']
```

kittiu: @eLBati I think the first intention is to not allow any edition to any model (as much as it could), i..e, most of the returns you show shouldn't be editable as well.
So I think to be compromised for everyone requirement, I think we start with a minimum list and make the list extensible on custom module.

I would start with,
- ir.filters (auditor should be able to create his/her own filter)

WDYT?

eLBati: I made https://github.com/OCA/server-tools/commit/c69eba0fd0f4497c694fd7dbe16f56fff795acf7 because I got into errors while a read-only user logs in and gets error because not able to create/write `res.users.log`.
So, the minimum set for me, AFAIK now, is `['res.users.log']`.
Thus, I made the `search` implementation instead of a fixed list, to let the system tell us which models to allow.

The other way I see is to restore your list
```
        [
            "res.users",
            "res.users.log",
            "mail.channel",
            "mail.channel.partner",
        ]
```

Other models like `ir.filters` should be considered in a second step I think; considering however that, with the `search` implementation, access could be granted modifying ACL

yostashiro: IMO @eLBati 's approach is more generic and therefore is better. To me, what's globally updatable in the first place in the ACL should remain updatable even for read-only users. I think you can override `_readonly_exclude_models()` if there is a situation that goes against this idea.

@kittiu Can you think of a case where you may have some problem with above approach?

kittiu: Alright, I understand the point of @eLBati now. It is in a way configurable also. So I restored back using search()
Thanks both.

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,61 @@
...
+        return (
+            self.sudo()
+            .search(
+                [
+                    ("group_id", "=", False),
+                    "|",
+                    ("perm_write", "=", True),
+                    "|",
+                    ("perm_create", "=", True),
+                    ("perm_unlink", "=", True),
```
yostashiro, line 62: @kittiu Can you please see my comment here https://github.com/OCA/server-tools/pull/2188/files#r749825989?

yostashiro (approved):
Code review. LGTM. 👍

dsolanki-initos:
@dreispt Can we merge this?

Saran440:
@kittiu Functional Test
It's error with module hr_expense

Step to error

1. Install module `hr_expense`
2. Create new user and change permission Employee to null
Selection_008
3. Create employee from this user
4. Login with new user
5. Go to menu Expenses > My Expenses > My Reports > Create
6. It will error wrong menu
Selection_009

------------------------------------------------------------------------------------

Test before install module `base_model_restrict_update`. I can create new expense


After install module `base_model_restrict_update`

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,61 @@
...
+    _inherit = "ir.model.access"
+
+    @api.model
+    @tools.ormcache_context(
+        "self._uid", "model", "mode", "raise_exception", keys=("lang",)
+    )
+    def check(self, model, mode="read", raise_exception=True):
+        if self.env.su:
+            return True
+        res = super().check(model, mode, raise_exception)
```
Saran440, line 24:
```suggestion
        self.flush()
        res = super().check(model, mode, raise_exception)
```

Saran440 (commented):
I try to add self.flush() and it' work.
WDYT?

kittiu:
@Saran440 I can't get this error. Are you sure you click "Update" module properly.

base_model_restrict_update/models/ir_model_access.py
kittiu: @Saran440 
I tested by removing rights on Emplyee, Contact and Expense. Now, I can replicate the problem.

I try with .fulsh() and still have the same error.

kittiu:
I think I find the root cause. The error is from this line,
```
sheet.address_id = sheet.employee_id.sudo().address_home_id
```
and the error,
```
ValueError: Invalid field 'address_home_id' on model 'hr.employee.public'
```
As we remove permissions, the only object that this user have access to is only `hr.employee.public` and not `hr.employee` (which need to be at least officer).
As such,  it doesn't have `address_home_id` field.

I think your error is not from this module, but the standard Odoo itself.

kittiu:
image

Problem is solved if this access is added.

kittiu:
OK, I found the problem and fixed now.

https://github.com/OCA/server-tools/compare/7808bcf19cec072eaea625135bb9f10eea81c68e..9fd5e8ccd3aa8771d26743e52af4355e68ff95bc

Thanks! @Saran440 for this important catch!

eLBati:
@Saran440 @kittiu thanks, also updated https://github.com/OCA/server-tools/pull/2188

eLBati:
Oops we don't have `self.env.su` in v12

Saran440 (approved):
Functional Test :+1:

PR #2170, opened 2021-09-08
[WIP][13.0] Add method_control_execution_order

mmequignon:
This module adds a configurable way to manage the order in which method overloads are executed.

### DIFF
--- a/method_control_execution_order/README.rst
+++ b/method_control_execution_order/README.rst
@@ -1,0 +1,1 @@
+wait for the bot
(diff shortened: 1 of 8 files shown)


legalsylvain (commented):
hi @mmequignon. Looks interesting ! 
I'm not sure to understand concretly how it can works (despite the documentation). 
- Could you point a real example that uses that feature ? 
- also adding test with demo python code could help.

kind regards.

sebalix:
We fixed this by merging some addons into one to get the expected execution order: https://github.com/OCA/sale-workflow/pull/1736
I'm closing this PR as we won't work anymore on it.

PR #2171, opened 2021-09-10
[8.0] pin unidecode==1.2.0

andhit-r:
Pin unidecode on [lates version (1.2.0) ](https://pypi.org/project/Unidecode/1.2.0/)that support python 2.7

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
   pstats_print2list
   python-ldap
-  unidecode
+  unidecode==1.2.0
   validate_email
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):

dreispt:
Is the pinning because later versions are known to be incompatible?

StefanRijnhart:
@dreispt later versions don't advertise compatibility with Python 2.7 anymore, indeed.

dreispt (approved):

PR #2173, opened 2021-09-14
[12.0][FIX] Scheduler_error_mailer: fixed error printing in email

### LINKED ISSUE #1927: [12.0] scheduler_error_mailer: Email has never been succeeded to retrieve error message
Hi all,
As mentioned in the title, I am feeling kinda strange that nobody has ever had the same issue.
In the error email, the message is rendered by calling attribute `value` of object `job_exception`:
`${ctx.get('job_exception') and ctx.get('job_exception').value or 'Failed to get the error message from the context.'}`

Meanwhile in the Odoo core function `_handle_callback_exception`. The `job_exception` is supposed to be an **Exception** object. And it does not have `value` attribute, means it has never been working the entire time.
image

My current solution is just simply cast the `job_exception` to string (`str(job_exception)`). If anybody has different idea, feel free to give your opinion.

HekkiMelody:
Currently, the email template for the scheduler_error_mailer can never print the exception. With this fix the exception is printed correctly in the email.

Fixes #1927

### DIFF
--- a/scheduler_error_mailer/data/ir_cron_email_tpl.xml
+++ b/scheduler_error_mailer/data/ir_cron_email_tpl.xml
@@ -18,7 +18,7 @@
 <p>Odoo tried to run the scheduler <em>${object.name or ''}</em> in the database <em>${ctx.get('dbname')}</em> but it failed. Here is the error message :</p>
 
 <strong>
-${ctx.get('job_exception') and ctx.get('job_exception').value or 'Failed to get the error message from the context.'}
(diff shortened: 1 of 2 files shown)


gurneyalex (approved):

StefanRijnhart (commented):
The email template is noupdate. Can you add a migration script to replace the text of existing installs?

StefanRijnhart:
Oh, the migration script is added in https://github.com/OCA/server-tools/pull/2262 now. Closing this then. Thank you @aleuffre for your contribution!

PR #2177, opened 2021-09-22
[14.0] Updated from template

hbrunn:
this updates the repo to the latest template, mainly auto closing stale PRs


.travis.yml
```diff
@@ -19,23 +19,32 @@ stages:
...
-      name: "pre-commit"
-      before_install:
-      install: pip install pre-commit
-      script: pre-commit run --all --show-diff-on-failure --verbose --color always
-      after_success:
+    # Test separately: database_cleanup
     - stage: test
       env:
-        - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
+        - TESTS=1 ODOO_REPO="odoo/odoo" MAKEPOT=1 INCLUDE="database_cleanup"
```
pedrobaeza: This shouldn't be needed, and more as the module hasn't been migrated yet.

hbrunn: I just copied the existing config, but indeed this should go then. Just for clarity, if we add/remove a rebel module, this should happen via a new `copier update` call, no manual fiddling with the files, right?

hbrunn:
...and what was the solution to runbot pointing to the wrong logfile links?

.travis.yml
pedrobaeza: I would say yes, but @Yajo can confirm

pedrobaeza:
> ...and what was the solution to runbot pointing to the wrong logfile links?

@gurneyalex can you answer this?

.travis.yml
hbrunn: just noticed we exclude things in [v14](https://github.com/OCA/server-tools/blob/14.0/.travis.yml#L34) too, so I added that as rebel modules. It's not exactly the same because the former version didn't test those modules at all.

yajo: > Just for clarity, if we add/remove a rebel module, this should happen via a new `copier update` call, no manual fiddling with the files, right?

That'd be the best, yes.

yajo (approved):

pedrobaeza (approved):
Let's go for it!

PR #2178, opened 2021-09-22
[14.0][MIG] mail_server_relay_disallowed

JoanSForgeFlow:
Standard migration to 14.0.

cc @AdriaGForgeFlow @MateuGForgeFlow


pedrobaeza:
Having the new system parameter for not doing spoofing, this module is not needed IMO.

AaronHForgeFlow:
What's that new system parameter? Thanks in advance :innocent:

ronozoro:
> What's that new system parameter? Thanks in advance 😇

If you want all your outgoing messages to use the same “From” address, use the key `mail.force.smtp.from` and set that address as value (such as `outgoing@mycompany.example.com`).

mail_server_relay_disallowed/models/ir_mail_server.py
```diff
@@ -0,0 +1,132 @@
...
+
+            # exact name and address
+            (oldname, oldemail) = parseaddr(message["From"])
+            # use original name with new address
+            newfrom = formataddr((oldname, smtp.user))
+            # need to use replace_header instead '=' to prevent
+            # double field
+            message.replace_header("From", newfrom)
+            smtp.sendmail(smtp.user, smtp_to_list, message.as_string())
+            # END OF CODE ADDED
```
thomaspaulb, line 120: I didn't know about the new Odoo configuration parameter, but for this purpose I always use [mail_outbound_static](https://github.com/OCA/social/tree/14.0/mail_outbound_static)... in comparison to that, why would this module be better?

PR #2183, opened 2021-10-01
[13.0][FIX] base_changeset: fix validations not displayed in form view

JordiMForgeFlow:
When fields view get method for form view is executed, self does not contain any record. This is causing the button to navigate to related changesets to never display because the length of self is 0, and the computed value user_can_see_changeset can't get properly evaluated.

I have tested both in local and runbot.
To fix it I am proposing:

- Use method directly instead of computed field to check user rights to the changeset, as the compute method does not evaluate correctly in empty record.

- Do not display button box if no changesets are pending.

### DIFF
--- a/base_changeset/models/base.py
+++ b/base_changeset/models/base.py
@@ -26,7 +26,6 @@ class Base(models.AbstractModel):
     count_pending_changeset_changes = fields.Integer(
         compute="_compute_count_pending_changesets"
     )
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow:
Hi @astirpe,

Could you review this issue I have found? If it was working before my guess is that perhaps something changed in Odoo core, but not sure.

JordiMForgeFlow:
Hi @astirpe,

After testing locally and in runbot I can see this problem affects the button for related changesets that should appear in the button box, which does not show.

PR #2184, opened 2021-10-04
[13.0][IMP] base_changeset: show change pop up in list view

AdriaGForgeFlow:
This PR intends to improve the module to show the pop-up in the list view allowing a user to accept / reject changes from the list view.

Default List View:
image

X2Many field in a  list view:
image

cc ~ @LoisRForgeFlow @JordiMForgeFlow

### DIFF
--- a/base_changeset/README.rst
+++ b/base_changeset/README.rst
@@ -173,6 +173,7 @@ Known issues / Roadmap
 
 * Only a subset of the type of fields is actually supported
 * Multicompany not fully supported
+* When accepting a change in a X2Many field (list view) the page will be reloaded to display the new value.
 
 Bug Tracker
 ===========
@@ -201,6 +202,7 @@ Contributors
 * Yannick Vaucher <yannick.vaucher@camptocamp.com>
 * Dennis Sluijk <d.sluijk@onestein.nl>
 * Andrea Stirpe <a.stirpe@onestein.nl>
+* Adria Gil Sorribes <adria.gil@forgeflow.com>
 
 Maintainers
 ~~~~~~~~~~~
--- a/base_changeset/models/record_changeset_change.py
+++ b/base_changeset/models/record_changeset_change.py
@@ -361,8 +361,9 @@ class RecordChangesetChange(models.Model):
         return change, pop_value
 
     @api.model
-    def get_fields_changeset_changes(self, model, res_id):
+    def get_fields_changeset_changes(self, model, res_id=False):
         fields = [
+            "record_id",
             "new_value_display",
             "origin_value_display",
             "field_name",
@@ -371,9 +372,10 @@ class RecordChangesetChange(models.Model):
         states = self.get_pending_changes_states()
         domain = [
             ("changeset_id.model", "=", model),
-            ("changeset_id.res_id", "=", res_id),
             ("state", "in", states),
         ]
+        if res_id:
+            domain.append(("changeset_id.res_id", "=", res_id))
         return self.search_read(domain, fields)
 
     def _compute_user_can_validate_changeset(self):
--- a/base_changeset/static/src/js/list_backend.js
+++ b/base_changeset/static/src/js/list_backend.js
@@ -1,0 +1,159 @@
+odoo.define("base_changeset.list_backend", function(require) {
+    "use strict";
+
+    var ListRenderer = require("web.ListRenderer");
+    var ListController = require("web.ListController");
+    var core = require("web.core");
+    var qweb = core.qweb;
+
+    ListController.include({
+        start: function() {
+            return this._super
+                .apply(this, arguments)
+                .then(this._updateChangeset.bind(this));
+        },
+
+        update: function() {
+            var self = this;
+            var res = this._super.apply(this, arguments);
+            res.then(function() {
+                self._updateChangeset();
+            });
+            return res;
+        },
+
+        _updateChangeset: function() {
+            var self = this;
+            var state = this.model.get(this.handle);
+            this.model.getChangeset(state.model, false).then(function(changeset) {
+                self.renderer.renderChangesetPopovers(changeset);
+            });
+        },
+
+        applyChange: function(id) {
+            this.model.applyChange(id).then(this.reload.bind(this));
+        },
+
+        rejectChange: function(id) {
+            this.model.rejectChange(id).then(this.reload.bind(this));
+        },
+    });
+
+    ListRenderer.include({
+        /* eslint-disable no-unused-vars */
+        init: function(parent, state, params) {
+            this._super.apply(this, arguments);
+            this.lastChangeset = false;
+        },
+
+        _onToggleOptionalColumnDropdown: function(ev) {
+            this._super.apply(this, arguments);
+            // HACK to trigger the renderChangesetPopovers when user has added / removed an optional column
+            if (!$(ev.target).hasClass("o_optional_columns_dropdown_toggle")) {
+                this.renderChangesetPopovers(this.lastChangeset);
+            }
+        },
+
+        renderChangesetPopovers: function(changeset) {
+            var self = this;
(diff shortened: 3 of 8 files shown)


base_changeset/static/src/js/relational_fields.js
```diff
@@ -0,0 +1,61 @@
...
+        },
+
+        applyChange: function(id) {
+            return this._rpc({
+                model: "record.changeset.change",
+                method: "apply",
+                args: [[id]],
+                context: _.extend({}, this.context, {set_change_by_ui: true}),
+            }).then(function() {
+                window.location.reload();
```
astirpe: did you try if `this.reload()` works?

astirpe (changes requested):
Thank you for this great contribution!

I just tried in Runbot to open a list view of partners and I got this error:

```
Traceback:
Error: self.renderer.renderChangesetPopovers is not a function
_updateChangeset/<@http://3505617-2184-5a37f7.runbot2.odoo-community.org/base_changeset/static/src/js/relational_fields.js:34:35
```

astirpe:
@tarteo @hbrunn you may be interested in this PR!

AdriaGForgeFlow:
> Thank you for this great contribution!
> 
> I just tried in Runbot to open a list view of partners and I got this error:
> 
> ```
> Traceback:
> Error: self.renderer.renderChangesetPopovers is not a function
> _updateChangeset/<@http://3505617-2184-5a37f7.runbot2.odoo-community.org/base_changeset/static/src/js/relational_fields.js:34:35
> ```

Thanks! It should be fixed now.

base_changeset/static/src/js/relational_fields.js
AdriaGForgeFlow: Yes, I tried and it does not work for me. Could not find another way of reloading the value.

JordiMForgeFlow (approved):
Functional and code review. Really nice feature! 🚀

base_changeset/static/src/js/relational_fields.js
astirpe: @AdriaGForgeFlow can you try with this?


```javascript
        applyChange: function(id) {
            var self = this;
            return this._rpc({
                model: "record.changeset.change",
                method: "apply",
                args: [[id]],
                context: _.extend({}, this.context, {set_change_by_ui: true}),
            }).then(function() {
                self.trigger_up('reload');
            });
        },
```

AdriaGForgeFlow: Genius! :clap: I added the change, thank you!

astirpe (approved):
Looks good to me! I'm looking forward for the review by @tarteo and @hbrunn :slightly_smiling_face:

base_changeset/README.rst
```diff
@@ -173,6 +173,7 @@ Known issues / Roadmap
 
 * Only a subset of the type of fields is actually supported
 * Multicompany not fully supported
+* When accepting a change in a X2Many field (list view) the page will be reloaded to display the new value.
```
astirpe, line 176: This comment is obsolete now

base_changeset/static/src/js/list_backend.js
```diff
@@ -0,0 +1,159 @@
...
+        },
+
+        rejectChange: function(id) {
+            this.model.rejectChange(id).then(this.reload.bind(this));
+        },
+    });
+
+    ListRenderer.include({
+        /* eslint-disable no-unused-vars */
+        init: function(parent, state, params) {
```
tarteo, line 44: You can remove the eslint exception and remove the parameters and it will work fine.

base_changeset/static/src/js/list_backend.js
```diff
@@ -0,0 +1,159 @@
...
+                    var resId = changes[0].record_id.split(",")[1];
+                    var dataId = self.state.data.filter(function(element) {
+                        return element.res_id == resId;
+                    });
+                    if (dataId.length > 0) {
+                        var bodyRow = self.$el.find("[data-id='" + dataId[0].id + "']");
+                        var $tr = self.$el.find("thead").find("tr");
+                        var i = 0;
+                        var cellPos = 0;
+                        _.each($tr.children(), function(td) {
```
tarteo, line 89: You can change this to ` _.each($tr.children(), function(td, i) {` so you don't need to create the variable on line 87 (`var i = 0;`).

You can also use `_.findIndex` https://underscorejs.org/#findIndex

tarteo (approved):
LGTM (I only reviewed the code) only some none blocking issues.

hbrunn (approved):

hbrunn:
@AdriaGForgeFlow do you plan to follow up on @tarteo's comments? I plan to backport this after merging

JordiMForgeFlow:
Hi @hbrunn

We tested this improvement in a database with multiple records and noticed a high performance issue due to it. We used the changeset changes on purchase order line fields like the price unit and product qty, and when there were several changeset changes in the system it was taking up to 15 seconds to open one purchase order.

We did not have time to keep investigating but seems that when you open one purchase order from the tree view, the new change tries to fetch the changeset changes for all the purchase orders that you had loaded in the tree view.

hbrunn:
so I'll sit back until you've fixed that in v13, and then backport that :-) Maybe better set the PR to draft meanwhile?

LoisRForgeFlow (changes requested):
performance issue pending to attend.

LoisRForgeFlow:
> so I'll sit back until you've fixed that in v13, and then backport that :-) Maybe better set the PR to draft meanwhile?

Hi @holdenrehg to be honest I don't think we are going to fix it soon, we have worked around the problem and we don't have funding for more changes/improvements. So in the short term, I think this will sit here waiting for a superhero :)

Anyway for me the real problem of this module that really needs some solution is https://github.com/OCA/server-tools/issues/2187.

holdenrehg:
> > so I'll sit back until you've fixed that in v13, and then backport that :-) Maybe better set the PR to draft meanwhile?
> 
> Hi @holdenrehg to be honest I don't think we are going to fix it soon, we have worked around the problem and we don't have funding for more changes/improvements. So in the short term, I think this will sit here waiting for a superhero :)
> 
> Anyway for me the real problem of this module that really needs some solution is #2187.

Hi @LoisRForgeFlow, I think I was maybe accidentally tagged on this one? :)

LoisRForgeFlow:
> > > so I'll sit back until you've fixed that in v13, and then backport that :-) Maybe better set the PR to draft meanwhile?
> > 
> > 
> > Hi @holdenrehg to be honest I don't think we are going to fix it soon, we have worked around the problem and we don't have funding for more changes/improvements. So in the short term, I think this will sit here waiting for a superhero :)
> > Anyway for me the real problem of this module that really needs some solution is #2187.
> 
> Hi @LoisRForgeFlow, I think I was maybe accidentally tagged on this one? :)

Hi, it seems so, my bad :sweat_smile: 

@hbrunn The message was for you :)

base_changeset/models/record_changeset_change.py
```diff
@@ -371,9 +372,10 @@ def get_fields_changeset_changes(self, model, res_id):
         states = self.get_pending_changes_states()
         domain = [
             ("changeset_id.model", "=", model),
-            ("changeset_id.res_id", "=", res_id),
             ("state", "in", states),
         ]
+        if res_id:
+            domain.append(("changeset_id.res_id", "=", res_id))
```
victoralmau, line 378: I think you have to add a domain "impossible" when you don't specify `res_id`, otherwise you will get all records (something wrong).  Could this be the cause of https://github.com/OCA/server-tools/pull/2184#issuecomment-963081075 ?

victoralmau:
Backport to 14.0 https://github.com/OCA/server-tools/pull/2759 and extra changes added to improve performance.

PR #2185, opened 2021-10-04
[13.0] [IMP] database_cleanup: purge orphaned SQL views

Linekio:
Improve the "Purge obsolete tables" feature to purge SQL views.
SQL views are not automatically dropped after uninstalling an addons, or upgrading odoo.
For example, the view mail_statistics_report is not dropped after migration to v13: https://github.com/odoo/odoo/blob/12.0/addons/mass_mailing/models/mass_mailing_report.py#L32

### DIFF
--- a/database_cleanup/models/purge_tables.py
+++ b/database_cleanup/models/purge_tables.py
@@ -5,6 +5,13 @@ from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
 from ..identifier_adapter import IdentifierAdapter
+from psycopg2.extensions import AsIs
+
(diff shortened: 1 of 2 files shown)


legalsylvain (commented):
Hi. Did you took a look on that PR ? Seems similar : https://github.com/OCA/server-tools/pull/2098

Linekio:
No, I searched for it but not found.
I guess it should be merged in v13.

legalsylvain:
Could you cherry pick it ?

Linekio:
Done

PR #2188, opened 2021-10-08
[12.0] base_model_restrict_update Add option 'Readonly User' to all models

eLBati:


mrcast (commented):
@eLBati a situation where readonly doesn't work
STEP
- create a user readonly


- login with this user
- create a sale order:
+ observed: the user can create a draft sale order
+ expected: the user can not create a draft sale order

TKS

mrcast (approved):
Functional test ok!

GSLabIt (approved):
LGTM

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -17,19 +17,26 @@ def check(self, model, mode="read", raise_exception=True):
...
-        )
-        query_res = self._cr.dictfetchall()[0]
-        if (
-            query_res["restrict_update"]
-            and mode != "read"
-            and not self.env.user.unrestrict_model_update
-        ):
-            if raise_exception:
+        if mode != "read" and raise_exception:
+            if self._test_readonly(model) or self._test_restrict_update(model):
```
yostashiro, line 25: Could we exclude some models from this checking? ref. https://gist.github.com/moylop260/43748e848dfc9916fd2f0fd51a69a12a#file-read_only_user-py-L11-L18

Thank you very much for the nice addition.

eLBati: https://github.com/OCA/server-tools/pull/2169#discussion_r745432119

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -17,19 +17,36 @@ def check(self, model, mode="read", raise_exception=True):
...
+            return True
+        return False
+
+    @api.model
+    def _readonly_exclude_models(self):
+        """ Models updtate/create by system, and should be excluded from checking """
+        return self.sudo().search([
+            ("group_id", "=", False),
+            "|", ("perm_write", "=", True),
+            "|", ("perm_create", "=", True), ("perm_unlink", "=", True)
```
yostashiro, line 55:
```suggestion
            ("group_id", "=", False), ("perm_create", "=", True)
```

To slightly lighten the load (although the difference is probably negligible).

I suppose just having `("perm_create", "=", True)` is sufficient.  I haven't seen a case where `perm_write` or `perm_unlink` is true and `perm_create` is false.

kittiu: I could in theory if, i..e, edit the ACL after records already created in order to prevent some master from further creation. May not be common, but nonetheless.

yostashiro: @kittiu Thanks for the comment. I'm OK with either way.

yostashiro (approved):
Code review. LGTM. 👍

PR #2189, opened 2021-10-11
[14.0][IMP] base_jsonify: Allow to export many2one and referene field as simple field

Cedric-Pigeon:
foward port #2063


rousseldenis (approved):

lmignon (approved):

lmignon:
Tired..... now it's not possible to merge a PR since some addons are calling external resources into their unittest without mocking the calls. And since these services are not reliable, (limit in number of calls / days / sec, ...) we are stuck.   This repo is too big and addons calling external resources, if not adapted to mock the call, should be isolated into a single addon repo...

PR #2190, opened 2021-10-12
[14.0][FIX] Other modules upgrade write file full path

Yasaie:
Parentheses placement was wrong in other modules full_path that lead to create migrations on root of modules


legalsylvain (approved):
Thanks ! 
Fast tracking patch

/ocabot merge patch

pedrobaeza:
Curious that this wasn't detected before. The bot doesn't react to review comment, so:

/ocabot merge patch

Issue #2187, opened 2021-10-07
[13.0] base_changeset: mass deletion of field records on update


JordiMForgeFlow:
To reproduce in a blank database:

- Install some module (for example sale_management)
- Install base_changeset module
- Update base_changeset module

When the update process of the module is triggered, a massive deletion of field records is done for all models in the system (see picture below)

![image](https://user-images.githubusercontent.com/71635103/136412681-066dd647-2417-4ff2-a7fa-a31c5eea730b.png)

I have tried to debug when this is happening and seems to be executed on the method _process_end from the base module (https://github.com/odoo/odoo/blob/9d72e8cf5aff5a3449175a8f048894f66c09e5fd/odoo/addons/base/models/ir_model.py#L2005). I am not sure on why it happens but I have not seen any other examples in other modules where fields are added into the base  model.

I have not tried in other versions.

cc @LoisRForgeFlow @pedrobaeza @astirpe @guewen

PR #2194, opened 2021-10-18
[13.0][REF] base_changeset: remove base model inheritance

JordiMForgeFlow:
Using the base model implies adding the changeset fields and methods for all models in the system. This is causing issues when the module is updated, as recorded in #2187.

Usually the practice is not to perform this kind of inheritance, but to create a new Abstract Model to be used to add such functionalities.

This MR intends to perform this modifications so that the models that should be provided with the changeset fields and methods can simply inherit from this new Abstract Model. To do so, new modules can be developed on top of this base module (for example a module purchase_changeset or sale_changeset).

**This MR is currently a proposal opened to discussion. The tests have not yet been adapted.**


JordiMForgeFlow:
cc @LoisRForgeFlow @astirpe @guewen @yvaucher

astirpe:
So this way the module become useless by itself, because it will require extra development for each model we want to use this functionality with. This was unwanted in my initial idea of this module.

JordiMForgeFlow:
It's true, but inheriting directly from base and adding all the fields for all models seems to cause issues, specially when dealing with DB having lots of models. In other similar cases I have seen that the typical approach is to have a base module adding the core fields and methods in a new abstract model, and then modules are developed on top for the different areas where it could be needed.

I have tried this new approach together with a new purchase_changeset module for purchase related models and worked fine without the issue detected when inheriting from base. I wanted to share it to have opinions on it.

pedrobaeza:
One intermediate option can be to attach this feature to `mail.thread`, that is inherited by most of the models, but not having the issue with the `base` model.

JordiMForgeFlow:
> One intermediate option can be to attach this feature to `mail.thread`, that is inherited by most of the models, but not having the issue with the `base` model.

I guess this would cover a wide quantity of models, but still models like `purchase.order.line` or `sale.order.line` miss that inheritance. At the end, then, we would still need specific modules to leverage the feature in these models.

astirpe:
Attaching this feature to `mail.thread` would mitigate the requiring of extra development for each model, however I'm afraid there still will be a lot of models excluded, like for example lines (eg.: invoice lines, order lines, ...), configuration records, etc... I'm not against the principle of this PR, I'm just noticing that this PR introduces a regression for the way this module was meant. Is it possible for this module to skip [_process_end()](https://github.com/odoo/odoo/blob/9d72e8cf5aff5a3449175a8f048894f66c09e5fd/odoo/addons/base/models/ir_model.py#L2005) somehow?

JordiMForgeFlow:
I am not sure on changing such core method behavior...

PR #2197, opened 2021-10-27
[14.0][IMP] base_exception: Allow to skip exception checks through context

ivantodorovich:
In some cases, you may want to skip exceptions from being checked, or even raised. (ie: you know exceptions are going to be checked anyways later on, so you avoid checking multiple times)


jcoux (approved):

kittiu:
Good addition, thanks!

kittiu (approved):

PR #2199, opened 2021-11-01
[15.0][MIG] base_exception

TheCloneMaster:
Migration of base_exception to 15.0


TheCloneMaster:
Will apply new PR as this doesn't include previous history...

PR #2201, opened 2021-11-05
[IMP] base_changeset: overhaul security

hbrunn:
the issue here is that as it is currently, every user can read all changesets, and this means you can't have confidential data in changesets.

The proposal is to make changes only accessible for validators (and the user who effected the change), as they are the only ones having any business with those changes.

It's the responsibility of the integrator then to have only people as validators who can be trusted with all data.

### DIFF
--- a/base_changeset/models/changeset_field_rule.py
+++ b/base_changeset/models/changeset_field_rule.py
@@ -46,7 +46,7 @@ class ChangesetFieldRule(models.Model):
     )
     validator_group_ids = fields.Many2many(
         'res.groups', 'changeset_field_rule_validator_group_rel',
-        string='Validators', default=lambda self: self.env.ref(
+        string='Validator Groups', default=lambda self: self.env.ref(
             'base_changeset.group_changeset_user',
             raise_if_not_found=False,
         ) or self.env['res.groups'],
--- a/base_changeset/models/changeset_field_rule.py
+++ b/base_changeset/models/changeset_field_rule.py
@@ -44,6 +44,13 @@ class ChangesetFieldRule(models.Model):
         help="Use this rule only on records where this is true. "
         "Available variables: object, user",
     )
+    validator_group_ids = fields.Many2many(
+        'res.groups', 'changeset_field_rule_validator_group_rel',
+        string='Validators', default=lambda self: self.env.ref(
+            'base_changeset.group_changeset_user',
+            raise_if_not_found=False,
+        ) or self.env['res.groups'],
+    )
 
     def init(self):
         """Ensure there is at most one rule with source_model_id NULL.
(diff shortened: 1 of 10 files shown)


base_changeset/security/groups.xml
```diff
@@ -15,12 +15,11 @@
         </record>
     </data>
     <data noupdate="1">
-        <record id="group_changeset_manager" model="res.groups">
-            <field name="users" eval="[(4, ref('base.user_root'))]" />
-        </record>
         <record id="group_changeset_user" model="res.groups">
-            <field name="implied_ids" eval="[(4, ref('group_changeset_manager'))]" />
```
astirpe, line 22: I would keep `<field name="users" eval="[(4, ref('base.user_root'))]" />` for `group_changeset_user` as well

base_changeset/models/record_changeset_change.py
```diff
@@ -383,9 +383,12 @@ def get_fields_changeset_changes(self, model, res_id):
 
     def _compute_user_can_validate_changeset(self):
         is_superuser = self.env.user.id == SUPERUSER_ID
-        has_group = self.user_has_groups("base_changeset.group_changeset_user")
+        user_groups = self.env.user.groups_id
         for rec in self:
-            can_validate = rec._is_change_pending() and (is_superuser or has_group)
+            can_validate = rec._is_change_pending() and (
+                is_superuser or
+                rec.rule_id.validator_group_ids & user_groups
```
astirpe: In case field Validators (validator_group_ids) is empty, shouldn't `group_changeset_user` be enabled anyways? Or for example in case `group_changeset_user` is accidentally removed from validator_group_ids.

base_changeset/models/changeset_field_rule.py
```diff
@@ -44,6 +44,13 @@ class ChangesetFieldRule(models.Model):
         help="Use this rule only on records where this is true. "
         "Available variables: object, user",
     )
+    validator_group_ids = fields.Many2many(
+        'res.groups', 'changeset_field_rule_validator_group_rel',
+        string='Validators', default=lambda self: self.env.ref(
```
astirpe: I would name this field as "Validator Groups"

astirpe (changes requested):

base_changeset/security/groups.xml
hbrunn: with the now correct implied ides and admin added to the manager group, admin also ends up in the user group, so that's fine I think? I do notice however that we have an empty xml node then, so I'm going to remove that

astirpe: you are right, thanks!

hbrunn: thanks, followed up on all your input by now

StefanRijnhart (approved):

astirpe (approved):

PR #2203, opened 2021-11-09
Update dotfiles, rebel fetchmail_notify_error_to_sender

### LINKED ISSUE #2191: [14.0] MAKEPOT loop
Since 2 jobs are declared into travis with MAKPOT=1, The result of generated pot diverge than the one from oca_transbot. Indeed, `base_changeset`, `base_autovacuum_message_attachment` and `base_name_search_improved`  introduces new fields on "base". As consequence, once installed, all the generated pot file will contains these fields. Pot are generated a first time for a restricted list of addons excluding these base addons, a second time for all the others addons and a third time by oca_transbot on a db will all the addons installed. As result, files generated by transbot are always different than the files generated by the first travis job on restricted list of addons.

sbidoul:
fetchmail_notify_error_to_sender must be in the rebel
modules group, because it depends on fetchmail_notify_error_to_sender_test
which is a rebel module. Otherwise fetchmail_notify_error_to_sender
is in both jobs which creates issues for makepot.

Attempt to fix #2191

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.3.4
+_commit: v1.3.5
 _src_path: gh:oca/oca-addons-repo-template
 ci: Travis
 dependency_installation_mode: PIP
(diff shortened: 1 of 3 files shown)


lmignon:
Thank you @sbidoul I didn't know this 'rebel group'.

PR #2210, opened 2021-11-17
[IMP] auditlog: Improve performance of creating logs for recordsets

emilkholod:
Description of the issue/feature this PR addresses:
This PR contains performance improvement on creating logs for recordsets. 

Current behaviour before PR:
There was one-by-one creating log and log line records. It was very slow when we want to create logs for recordsets. 

Desired behaviour after PR is merged:
I group recordset by batches, so each log record will write in two stages (SQL) requests: 
- create log
- audit log lines for this log
Also there was added optimization for reading `current_http_request` and `current_http_session` - because before this it was calling inside the cycle and it read data from database - it was also slowing down performance. In this PR they were moved outside the cycle.

### DIFF
--- a/auditlog/__manifest__.py
+++ b/auditlog/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Audit Log",
-    "version": "14.0.1.2.0",
+    "version": "14.0.1.3.0",
     "author": "ABF OSIELL,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "website": "https://github.com/OCA/server-tools",
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -473,44 +473,69 @@ class AuditlogRule(models.Model):
         if new_values is None:
             new_values = EMPTY_DICT
         log_model = self.env["auditlog.log"]
-        http_request_model = self.env["auditlog.http.request"]
-        http_session_model = self.env["auditlog.http.session"]
-        for res_id in res_ids:
-            model_model = self.env[res_model]
-            name = model_model.browse(res_id).name_get()
-            model_id = self.pool._auditlog_model_cache[res_model]
-            auditlog_rule = self.env["auditlog.rule"].search(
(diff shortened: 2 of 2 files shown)


auditlog/models/rule.py
```diff
@@ -473,44 +473,69 @@ def create_logs(
...
+        model_id = self.pool._auditlog_model_cache[res_model]
+        auditlog_rule = self.env["auditlog.rule"].search(
+            [("model_id", "=", model_id)],
+            limit=1,
+        )
+        log_rules_values_to_create = []
+        for batched_res_ids in self._cr.split_for_in_conditions(res_ids):
+            names = self.env[res_model].browse(batched_res_ids).name_get()
+            for name, res_id in zip(names, batched_res_ids):
+                res_name = name and name[1]
```
StefanRijnhart, line 488: For this to work with the `zip` and all, it would have to be the case that name_get returns an empty tuple or something instead of the regular (id, name) format. Does that really happen?

auditlog/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Audit Log",
-    "version": "14.0.1.2.0",
+    "version": "14.0.1.3.0",
```
StefanRijnhart, line 6: Because no versioned migration script is present in this PR, you can leave the version number increment to the OCA bot.

auditlog/models/rule.py
```diff
@@ -473,44 +473,69 @@ def create_logs(
...
+                    "res_id": res_id,
+                    "method": method,
+                    "user_id": uid,
+                    "http_request_id": current_http_request,
+                    "http_session_id": current_http_session,
+                }
+                vals.update(additional_log_values or {})
+                log_rules_values_to_create.append(vals)
+
+            log_rules = log_model.create(log_rules_values_to_create)
```
StefanRijnhart, line 501: `log_rules` is a little confusing here and in the method signature below. Maybe `logs` would be clearer?

PR #2215, opened 2021-11-24
[FIX] base_exception: use odoo-test-helper

yajo:
The library is made for what this module was doing, but it cleans it better.

Other tests were failing in an integration environment and I suspect this is the culprit.

@Tecnativa


joao-p-marques (approved):
Code review. LGTM :+1:

carlosdauden (approved):

pedrobaeza (approved):

PR #2216, opened 2021-11-24
[14.0][IMP] base_view_inheritance_extension: real python_dict from ast tree

ivantodorovich:
Use `ast` parsing for `python_dict`, which enables its use in other cases than simply just `context`, for example `attrs`. It treats the value as a real python expression and not just a string

The added unit test fails without this change.

[btw `astor` is already an optional dependency on odoo core](https://github.com/odoo/odoo/blob/b3302398bc79f107588ff755666bf74a2ef4e11a/openerp/addons/base/ir/ir_qweb/qweb.py#L116-L120)

### DIFF
--- a/base_view_inheritance_extension/__manifest__.py
+++ b/base_view_inheritance_extension/__manifest__.py
@@ -10,6 +10,6 @@
     "summary": "Adds more operators for view inheritance",
     "website": "https://github.com/OCA/server-tools",
     "depends": ["base"],
-    "external_dependencies": {"python": ["pyyaml"]},
+    "external_dependencies": {"python": ["astor"]},
     "demo": ["demo/ir_ui_view.xml"],
 }
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -1,38 +1,14 @@
 # Copyright 2016 Therp BV <https://therp.nl>
 # Copyright 2018 Tecnativa - Sergio Teruel
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com).
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+import ast
+
+import astor
 from lxml import etree
-from yaml import safe_load
 
-from odoo import api, models, tools
-
-
-class UnquoteObject(str):
-    def __getattr__(self, name):
-        return UnquoteObject("{}.{}".format(self, name))
-
-    def __repr__(self):
-        return self
-
-    def __call__(self, *args, **kwargs):
-        return UnquoteObject(
-            "%s(%s)"
-            % (
-                self,
-                ",".join(
-                    [
-                        UnquoteObject(a if not isinstance(a, str) else "'%s'" % a)
-                        for a in args
-                    ]
-                    + ["{}={}".format(UnquoteObject(k), v) for (k, v) in kwargs.items()]
-                ),
-            )
-        )
-
-
-class UnquoteEvalObjectContext(tools.misc.UnquoteEvalContext):
-    def __missing__(self, key):
-        return UnquoteObject(key)
+from odoo import api, models
 
 
 class IrUiView(models.Model):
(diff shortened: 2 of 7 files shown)


SilvioC2C (approved):
LGTM

ivantodorovich:
@pedrobaeza could this interest you?

pedrobaeza:
Well, just in the shake of the ecosystem, but I'm afraid I have to reject it if you are changing dependencies in a stable version. Why do you need `astor` instead of plain `ast`? Odoo doesn't need extra things for doing its evaluations. I don't get also what is the use case you are trying to cover.

ivantodorovich:
Hello @pedrobaeza !

The usecase I'm trying to cover is basically the one in the new unit test:

Given this view:

```xml
<form>
    <field name="ref" attrs="{'invisible': [('state', '=', 'draft')]}" />
</form>
```

Try to add a `readonly` attr like so, without caring nor changing the existing `invisible` attr:

```xml
<field name="ref" position="attributes">
    <attribute name="attrs" operation="python_dict" key="required">
        [('state', '!=', 'draft')]
    </attribute>
</field>
```

Without this PR you get an error related to the current hacky string treatment

----

> Why do you need `astor` instead of plain `ast`

Plain `ast` is perfect to parse source code into a syntax tree, but not the other way around.
(at least not until python `>=3.9` with the addition of [ast.unparse](https://docs.python.org/3/library/ast.html#ast.unparse))

`astor` provides the `to_source` method required to transform the syntax tree back to a string.

----

> [...] I have to reject it if you are changing dependencies in a stable version

What do you mean *stable* version?
I agree it would require a `major` module version bump, but that's what versions are for ^^

pedrobaeza:
> What do you mean stable version?

It means that people updating the module inside same Odoo version will find that it fails to load, and their installation breaks. This kind of things are not usually allowed even in OCA. Major version bumps are for migration scripts, not for this kind of changes.

ivantodorovich:
Well.. updating the module source is something technical ppl do, and a  `Unable to upgrade module "base_view_inheritance_extension" because an external dependency is not met: astor` error is easily understood and solved IMO..

Keep in mind current implementation only works on simple flat dicts (just simple `context` cases) 😅

pedrobaeza:
I let @hbrunn, who was the original author of such code, to decide, as all his code has been totally replaced.

hbrunn (approved):
actually, [my code](https://github.com/OCA/server-tools/commit/204c9a7554dcbcffbc691615b9b29f596090d6a2#diff-459a940a21549a47db709daa41758f6efea750902795c286577d2ca9dccfe78cL91-L97) has been butchered to death during the v13 migration, so I'm very happy with this

hbrunn:
PS: It's not in the tests, but this supports modifying a dict like `{'test': some_variable}` to `{'test': some_variable, 'test2': some_variable2}`, right?

hbrunn:
PPS: reason for changing this back then seems to have been expressions, but that should be covered by this approach too, right? Could you add some expression in your tests?

ivantodorovich:
> PS: It's not in the tests, but this supports modifying a dict like {'test': some_variable} to {'test': some_variable, 'test2': some_variable2}, right?

> PPS: reason for changing this back then seems to have been expressions, but that should be covered by this approach too, right? Could you add some expression in your tests?

That's correct! This was actually already tested, but I just added a few more tests anyway :)
Please check if all the cases you have in mind are covered.

The way it works, though, is by using `ast` to parse python source code, and apply the transformations directly to the ast tree. The code needs to be syntactically correct, but it's not evaluated -- which is what we need here.

ivantodorovich:
Btw since we're here, I'm failing to understand the reason for [`list_add` and `list_remove`](https://github.com/OCA/server-tools/tree/14.0/base_view_inheritance_extension#id1)

Isn't that already covered in core with

```xml
<attribute name="csv" separator="," add="something" remove="oldstuff" />
```

hbrunn:
it is, but nobody removed this when migrating to versions where core had this already (but don't remove it now because v14 modules might rely on this way of list manipulations)

hbrunn:
maybe add a note about this to the [roadmap](https://github.com/OCA/server-tools/tree/14.0/base_view_inheritance_extension#known-issues--roadmap)? Then you could also remove the line about features of moving stuff, that's implemented in the core version

ivantodorovich:
> maybe add a note about this to the [roadmap](https://github.com/OCA/server-tools/tree/14.0/base_view_inheritance_extension#known-issues--roadmap)? Then you could also remove the line about features of moving stuff, that's implemented in the core version

done 👍🏻

PR #2220, opened 2021-12-04
[14.0] base_jsonify: fail gracefully

simahawk:
It could happen that a module adds an ir.export.line for a new field
but when tests run the field is not loaded
if this module is not a dependency of the module running tests.

Furthermore, in general, it seems too strict to break the whole computation
if there's a field that is missing.
It seems more reasonable to log an error when this happen
so that technical action can be taken to fix it.


sebalix (approved):
Seems legit

TDu (approved):
LG, would be good to have a not in the readme about that feature

simahawk:
@sebastienbeau any feedback?

ivantodorovich (approved):
LG ;)

PR #1997, opened 2021-01-22
[14.0][MIG] sentry

moylop260:
Continuation of https://github.com/OCA/server-tools/pull/1929

Thanks to @traviswaelbro

PR #2223, opened 2021-12-04
[14.0] [IMP] sentry: Switch from raven to sentry_sdk

dnplkndll:


dnplkndll:
to implement the odoo_dir sha fetch and pass as release without a tag need to pick

https://forum.sentry.io/t/fetch-git-sha-equivalent-in-the-unified-python-sdk/5521
```python
raven.fetch_git_sha(odoo_dir)
```

dnplkndll:
> to implement the odoo_dir sha fetch and pass as release without a tag need to pick
> 
> https://forum.sentry.io/t/fetch-git-sha-equivalent-in-the-unified-python-sdk/5521
> 
> ```python
> raven.fetch_git_sha(odoo_dir)
> ```
oh nevermind https://github.com/OCA/server-tools/pull/2223/files#diff-297136b7cbfb1c11044f0db7d8fa3da6b2f9a50b6dcb046f94f12068b7ae7468R76 

so I think this is covering fine after debug. 

I actually don't pass in any .git refs in my docker build if anyone else can test that. I am not sure how you get the .git in. maybe on a local non docker setup?

legalsylvain:
12.0 version still uses new sentry_sdk lib. 
https://github.com/OCA/server-tools/blob/12.0/sentry/__manifest__.py#L20
The work has been 
That in Nov 2020.  https://github.com/OCA/server-tools/pull/1942
Could you better cherry peak that commit, and then apply extra changes, if required.

And the migration on January 2021 (from 13.0) to 14.0 doesn't take into account that change. https://github.com/OCA/server-tools/pull/1997  cc : @moylop260

dnplkndll:
I see CI fail, redis package needs added to support that or removed. I was going to remove as I do not find the metadata useful. But in general not sure how may installs would have redis sessions and otherwise would not be a useful requirement to always have. 
```
redis
GET 'session:9ffae8e961b33e78db9ee7fdb540fe2ba25fea80'
redis.key: session:9ffae8e961b33e78db9ee7fdb540fe2ba25fea80
redis.command: GET
info
15:19:27
```
image

stanevicius1:
Can you make pull request with fixes?

thomaspaulb:
Closing, since already handled by https://github.com/OCA/server-tools/pull/2254

PR #2232, opened 2021-12-13
[15.0][MIG] `base_view_inheritance_extension`

ivantodorovich:
* Removed `list_add` and `list_remove`, they've been deprecated and implemented in Odoo core since several versions ago.
* Removed `move` from docs, as it has also already been implemented in core several versions ago.
* Replaced `python_dict` by `update`, that performs an operation similar to :meth:`dict.update`, but on the `ast.Dict`.

The most important change is the replacement of `operation="python_dict"` by `operation="update"`.
Rather than having an operation per key, we can now perform a `dict.update`-like operation with multiple keys at once:

```xml
<field position="attributes">
    <attribute name="context" operation="update">
        {
            "key": "value",
        }
    </attribute>
</field>
```


ivantodorovich:
@hbrunn what do you think of this change? 👀

damdam-s (approved):
LG thanks

hbrunn:
doesn't this make it difficult to unset keys, and to modify existing modifications again with inheritance?

ivantodorovich:
> doesn't this make it difficult to unset keys [...]

Real unsetting is indeed not possible, closest thing is setting to `None`. But I don't see how this is different than what we had before 🤔 

If we want real unsetting, an option would be to do it when keys are assigned `None`; or adding an `unset="key1,key2,key3"` attribute perhaps.

----

> and to modify existing modifications again with inheritance?

I think it's the opposite, this makes it easier to read and to inherit code, IMO

See for example: https://github.com/OCA/account-invoicing/blob/338d3d85e0bebc7f95925db5caf3e37dc6f67853/account_invoice_transmit_method/views/partner.xml#L36-L45

vs: https://github.com/OCA/account-invoicing/blob/c630385b95de77309861ef289915cecf8aeb0623/account_invoice_transmit_method/views/res_partner.xml#L30-L35

alexis-via (approved):
operation=update works well

hbrunn (approved):

hbrunn:
restarted runbot however to see why it complains

ivantodorovich:
> restarted runbot however to see why it complains

I believe runbot doesn't work on `15.0` branches --> this is why runboat was created

hbrunn:
> I believe runbot doesn't work on `15.0` branches --> this is why runboat was created

please elaborate?

ivantodorovich:
> > I believe runbot doesn't work on `15.0` branches --> this is why runboat was created
> 
> please elaborate?

I understand there were some issues trying to have runbot working on 15.0 branches, and this is why @sbidoul created runboat. I think the reasons were mentioned during the OCA talks, but I can't find them right now. I believe it's related to travis2docker and the fact that we are not using travis for testing anymore

hbrunn:
aah, so actually the runbot hook shouldn't fire for 15.0 branches

ivantodorovich:
> aah, so actually the runbot hook shouldn't fire for 15.0 branches

I think so, yes ^^

sbidoul:
> aah, so actually the runbot hook shouldn't fire for 15.0 branches

Yes. Do you know how one does that ?

hbrunn:
yes, good question, now you ask I don't think we can do this from the github side, but we need to have runbot ignore some branches. A quick grep through the sources revealed [`forbidden_regex`](https://github.com/odoo/runbot/blob/13.0/runbot/models/repo.py#L245) (don't use runbot myself here any more), so just adding `^15.0-` in the oca runbot configurations should suffice probably?

sbidoul:
@hbrunn forbidden_regex is on runbot 13, and I think OCA is on runbot 11.

Anyway, runboat is now active on all branches from 10 to 15. So if things continue to go well, I'm inclined to decommission the OCA runbot sooner than later.

hbrunn:
but we need some intermediary. currently red = no merges from ocabot, or can/should we force everything in lower branches?

sbidoul:
> but we need some intermediary. currently red = no merges from ocabot, or can/should we force everything in lower branches?

@hbrunn in fact, "ocabot merge" ignores runbot, runboat and coverage

We need to see if it works well for older branches but I see nothing that would prevent it in principle.
Do not hesitate to ping me with runboat build issues that are not obvious.

ivantodorovich:
Merge failed due to a weird pre-commit error:
```
Error: reserveCache failed: Cache already exists. Scope: refs/heads/15.0-ocabot-merge-pr-2232-by-dreispt-bump-nobump, Key: pre-commit-2-5291cb15e157ded572a7fee6447ac205de50e27c4937998a3ac705325b1e7cba-ef4292b95f935ef17263c92de7e92b5ea01a560626d2f5812d60fa9821978fdd, Version: 28cdb9f5496f334116f23e86f0063f5d3a9348c2e22425a33171e071aadada7e
 ```

Could you please retry? 🤔

ivantodorovich:
Could we merge this? 🙏🏻

PR #2237, opened 2021-12-16
[FIX][12.0] base_changeset: fix write return type

dkubiak789:


dkubiak789:
It returned a recordset instead of boolean.

base_changeset/models/base.py
```diff
@@ -114,7 +114,7 @@ def write(self, values):
         for record in self:
             local_values = self.env["record.changeset"].add_changeset(record, values)
             super(Base, record).write(local_values)
-        return self
+        return bool(self)
```
StefanRijnhart: Good catch! I would simply return True as does the upstream write method: https://github.com/odoo/odoo/blob/12.0/odoo/models.py#L3403

StefanRijnhart: Also, please add the module name in the commit message and the PR title `[FIX] base_changeset: fix write return type`.

StefanRijnhart (approved):

StefanRijnhart:
Merging, as it was included in later versions.

/ocabot merge patch

StefanRijnhart:
Amazing, thanks @thomaspaulb

PR #2238, opened 2021-12-16
[14.0][IMP] `base_jsonify`: Add `_jsonify_x2m_to_ids` and `_jsonify_format_duration` helpers

ivantodorovich:
IMO these basic helpers (and possible others, as long as they're not opinionated) belong here, as you would need them quite often

### DIFF
--- a/base_jsonify/tests/test_helpers.py
+++ b/base_jsonify/tests/test_helpers.py
@@ -9,6 +9,7 @@ class TestJsonifyHelpers(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
         cls.partner = cls.env["res.partner"].create(
             {
                 "name": "My Partner",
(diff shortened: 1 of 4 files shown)


base_jsonify/tests/test_helpers.py
```diff
@@ -0,0 +1,44 @@
...
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+
+
+class TestJsonifyHelpers(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 11:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (commented):
LG, we should probably think about another way of defining/using these converters instead of polluting the base model.

damdam-s (approved):

SilvioC2C (approved):

ivantodorovich:
@sbidoul runboat is failing here for no apparent reason: `{"detail":"No log found."}` 😅

PR #2240, opened 2021-12-17
[IMP] sql_export_mail : add mute mechanism

legalsylvain:
1) Add "mute mechanism" to implement "Mail alert" on sql_export_mail.

Add an option ``option_mail_mute`` (disabled by default). 
if checked, when sending a mail, the send of mail will be disabled, until users enable it.

this is usefull with the "mail_condition" option set to "not_empty", to check regurlarly if all is OK.

2) minor fix in the data. removing ``admin@example.com``

3) propose @florian-dacosta and me as maintainers.

kind regards.


houssine78 (approved):

remytms (approved):
LGTM

thomaspaulb (approved):

PR #2244, opened 2021-12-23
[14.0][FIX] base_changeset: assign values for all computed fields

StefanRijnhart:
Prevents error on record.changeset.change's read() method
```
    changeset.change_ids.read()
  File "/home/odoo/odoo/models.py", line 3022, in read
    return self._read_format(fnames=fields, load=load)
  File "/home/odoo/odoo/models.py", line 3042, in _read_format
    vals[name] = convert(record[name], record, use_name_get)
  File "/home/odoo/odoo/models.py", line 5667, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/odoo/odoo/fields.py", line 1028, in __get__
    raise ValueError("Compute method failed to assign %s.%s" % (record, self.name))
ValueError: Compute method failed to assign record.changeset.change(1403,).origin_value_date
```

### DIFF
--- a/base_changeset/models/record_changeset_change.py
+++ b/base_changeset/models/record_changeset_change.py
@@ -155,14 +155,24 @@ class RecordChangesetChange(models.Model):
     @api.depends("changeset_id.res_id", "changeset_id.model")
     def _compute_origin_values(self):
         states = self.get_pending_changes_states()
+        field_names = [
+            field_name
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
makes sense. thanks !

thomaspaulb (approved):

PR #2246, opened 2021-12-23
[14.0][IMP] base_changeset: security overhaul and other ports from 12.0

StefanRijnhart:
Port of improvements made during the 12.0 backport of this module by @hbrunn: https://github.com/OCA/server-tools/pull/2167
plus the security improvements from https://github.com/OCA/server-tools/pull/2201 also by @hbrunn

### DIFF
--- a/base_changeset/models/base.py
+++ b/base_changeset/models/base.py
@@ -87,7 +87,8 @@ class Base(models.AbstractModel):
                     {
                         field_name: value
                         for field_name, value in vals.items()
-                        if self._fields[field_name].required
+                        if field_name in self._fields and
+                        self._fields[field_name].required
                         and isinstance(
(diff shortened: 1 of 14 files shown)


StefanRijnhart:
@hbrunn I found that passing [a new() for `record`](https://github.com/OCA/server-tools/blob/12.0/base_changeset/models/base.py#L82-L99) does not work in 14.0 as it inherits the cache of its origin so that the field values are not recognized as modified. I replaced it by passing an explicit `create` flag across the mechanism.

hbrunn (approved):
I think this whole part should be refactored, but neither the backport nor this PR are the good places to do that, so I think the flag is perfectly fine

MiquelRForgeFlow (approved):

PR #2256, opened 2022-01-13
[MIG] html_text: Migration to 15.0

Mantux11:
Standard migration to version `15.0`.


gaikaz:
Sad... Had to try 😛

StefanRijnhart:
I'm sorry @gaikaz! Maybe we can open up this function for non-maintainers.

/ocabot migration html_text

StefanRijnhart (approved):

pedrobaeza:
It was done on purpose. Check comments about this on https://github.com/OCA/oca-github-bot/pull/97#discussion_r363111092

gaikaz:
It's ok. Maybe some day... 😄

pedrobaeza:
You can apply for being a PSC if your efforts concentrate in one specific area: http://odoo-community.org/page/psc

StefanRijnhart:
@pedrobaeza thanks for the pointer. I understand the considerations, but it makes me feel sad, as I was hoping that the OCA bot would allow the OCA to grant more powers to incidental contributors in a flexible and controlled way.

pedrobaeza:
If you think this can be used without a abuse on some way, please comment. Ideas welcome.

StefanRijnhart:
@pedrobaeza no new ideas, I just think opening up the feature to all contributors might be worth a little damage but I'm well aware that it's people like you who have to clean it up afterwards so I'll leave it at this.

pedrobaeza:
We can detect if CIs are green (with exceptions), no other PR is already proposed for the module, and then let any contributor to mark it, but it's a very hard work to do on the bot side, and the performance may be drained with this, and this is also not good enough, as a previous migration may not be still marked by anyone. I think we should focus more on recruiting more PSCs to perform this easy task than to sophisticate the bot in this way.

gaikaz (approved):

gaikaz:
I personally would gladly help tagging PRs that I come across with the `/ocabot migration ...` as I know how to check for duplicates in all Github Issues/PRs. Tho, I don't see myself as PSC for a single repo. Maybe it is a little egoistic to think this, but @pedrobaeza, You must have been seeing me around for some time now. I do get to see you a lot, can only imagine the amount of effort you put here 😄
If there must be more than trust involved, maybe simply asking a contributor to not only be some-what active but also have a Member status?
Or, here from Via laurea we push PRs to OCA only from OCA Forks on our Github org. (e.g. this PR) Maybe allow tagging those to specific people form the company. Could that be a way to check?

Side-question: How does OCA feel about people from the same company approving their own PRs on OCA?
I always feel kind of "cheaty" when I do that, tho I actually do check it properly.

pedrobaeza:
Right now, there's no mechanism on ocabot to add some trusted logins/orgs: all works through GitHub permissions. That can be a way to enlarge the maintainers of this part, but not sure the technical difficulties for implementing it.

About the reviews from the same organization, it's not something incorrect, while they are real, not only approvals for circumvent the review guidelines. We do continuously in my organization for being self-dependent in all of our contributions.  I think it's very easy to detect fake reviews most of the times. There's still the requirement of one of the reviews being from a trusted reviewer/PSC, or at least the merge being triggered by a PSC/maintainer, so that keep the situation controlled.

gaikaz:
@OCA/tools-maintainers
Could we please get a review/merge here?
We have multiple modules that depend on this one.

gaikaz:
Hmm... Looks like it failed by some Github cache thing.
Let's try merging one more time?

StefanRijnhart:
@gaikaz indeed!

/ocabot merge nobump

PR #2258, opened 2022-01-20
branch 12.0, fix 2 bugs in database_cleanup

maisim:
https://github.com/OCA/server-tools/commit/b7da38debb8936bf4ef67945522741c2d62b7dae 
Bug : database_cleanup does not ignore inherited field so it throw an exception when it tried to delete one
Fix: ignore inherited columns (they  are removed when the field they are linked to is deleted)

https://github.com/OCA/server-tools/commit/63b4aef077113993d2475988782c7e0515e3464d
Bug: When I ran the tests to validate the previous fix I got an exception
FIx: fix typo in logger.info() call

### DIFF
--- a/module_auto_update/i18n/tr.po
+++ b/module_auto_update/i18n/tr.po
@@ -9,20 +9,21 @@ msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-07-21 02:43+0000\n"
-"PO-Revision-Date: 2017-07-21 02:43+0000\n"
-"Last-Translator: OCA Transbot <transbot@odoo-community.org>, 2017\n"
+"PO-Revision-Date: 2022-01-06 16:39+0000\n"
+"Last-Translator: Yavuz Avci <yavuzavci@gmail.com>\n"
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi. Thanks for your conyribution. Could you take a look on the red CI. (Travis / runboat).
Also add tests could be great if possible.

Kind regards.

maisim:
> Also add tests could be great if possible.

Ok, I think I will refactor the tests file first. 
Any way to tell travis to use black instead of flake for a specific file?
If so, is OCA ok to use black instead of flake when changes are done one old versions?

Regards,

legalsylvain:
>  Any way to tell travis to use black instead of flake for a specific file?

Not sure on that point. As far as i understand. For branch since 13.0, black is used in the OCA. before, not.

StefanRijnhart:
This would be cleaner without the code style changes indeed.

maisim:
Hi, I will split this PR, the first new one : https://github.com/OCA/server-tools/pull/2295

maisim:
The build fail with 
`  File "/mnt/data/odoo-addons-dir/setup/company_country/odoo/addons/company_country/models/res_config.py", line 39, in load_company_country
    'COUNTRY environment variable with country code is not '
odoo.exceptions.ValidationError: ('COUNTRY environment variable with country code is not set and no localization module is marked to be installed.', None)`

Any idea?

PR #2261, opened 2022-01-24
[14.0][IMP] base_changeset: optionally prevent self validation

Shamsidinkhon:


StefanRijnhart (approved):
Thanks!

StefanRijnhart:
@hbrunn FYI, we have this as a customization on our own project I believe.

hbrunn (approved):
yes, but slightly differently, and I think what this PR implements is how it should be

StefanRijnhart:
Thanks for fixing the iteration variable there!

PR #2269, opened 2022-02-01
[IMP] rely on standard for first onchange call

PierrickBrun:
Make use of this refactor from Odoo: https://github.com/odoo/odoo/commit/488e334fc8fd6601e304635fa577c98b9bfd18e6

It is more in line with the standard while also improving perfs

### DIFF
--- a/onchange_helper/models/base.py
+++ b/onchange_helper/models/base.py
@@ -38,6 +38,12 @@ class Base(models.AbstractModel):
         # If self is a record (play onchange on existing record)
         # we take the value of the field
         # If self is an empty record we will have an empty value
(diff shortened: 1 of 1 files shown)


onchange_helper/models/base.py
```diff
@@ -38,6 +38,12 @@ def play_onchanges(self, values, onchange_fields):
         # If self is a record (play onchange on existing record)
         # we take the value of the field
         # If self is an empty record we will have an empty value
+        if not self:
```
rousseldenis, line 41: @PierrickBrun Line 55 we should remove else.

PR #2270, opened 2022-02-03
[10.0][FIX] base_suspend_security:  Allows to use suspended uid into search domain

lmignon:


base_suspend_security/base_suspend_security.py
```diff
@@ -30,3 +30,16 @@ def __eq__(self, other):
...
     def __iter__(self):
         yield super(BaseSuspendSecurityUid, self).__int__()
+
+    def __len__(self):
+        return 1
+
+    def __getitem__(self, key):
+        int_value = super(BaseSuspendSecurityUid, self).__int__()
+        if isinstance(key, slice):
+            if key.start > 0:
```
thomaspaulb, line 40: Shouldn't this also trigger if `key.start` is negative?

thomaspaulb (commented):
Could you add some more context to this PR? An example of something that didn't work before, and which works now?

thomaspaulb:
Tests also seem to fail: `AccessError not raised`

PR #2271, opened 2022-02-03
[FP][FIX][ENG][14.0] base_name_search_improved: forward port of patches and fix to interactive search

Rad0van:
This has 2 parts:

1. Forward port of patches from 13.0 to improve GUI, tests, ...
2. Fix to interactive search where code taken from 13.0 leads to exceptions in `_extend_name_results` as `results` contains list of ids directly (instead of expected records). Upon fixing that extending results with `lazy_name_get` results into exception in ORM like `psycopg2.ProgrammingError: can't adapt type 'lazy'` So I am trying to fix both of these.
3. Translate Spanish texts to English

Feedback is welcome.

### DIFF
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -25,7 +25,11 @@ def _get_rec_names(self):
 
 @tools.ormcache(skiparg=0)
 def _get_use_smart_name_search(self):
-    return self.env['ir.model'].search([('model', '=', str(self._name))]).use_smart_name_search
+    return (
+        self.env["ir.model"]
+        .search([("model", "=", str(self._name))])
+        .use_smart_name_search
+    )
 
 
 @tools.ormcache(skiparg=0)
@@ -55,17 +59,28 @@ def _extend_name_results(self, domain, results, limit, name_get_uid):
     result_count = len(results)
     if result_count < limit:
         domain += [("id", "not in", [x[0] for x in results])]
-        rec_ids = self._search(domain, limit=limit - result_count, access_rights_uid=name_get_uid)
-        results.extend(models.lazy_name_get(self.browse(rec_ids).with_user(name_get_uid)))
+        rec_ids = self._search(
+            domain, limit=limit - result_count, access_rights_uid=name_get_uid
+        )
+        results.extend(
+            models.lazy_name_get(self.browse(rec_ids).with_user(name_get_uid))
+        )
     return results
 
 
 def patch_name_search():
     @api.model
-    def _name_search(self, name='', args=None, operator='ilike', limit=100, name_get_uid=None):
+    def _name_search(
+        self, name="", args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
         # Perform standard name search
         res = _name_search.origin(
-            self, name=name, args=args, operator=operator, limit=limit, name_get_uid=name_get_uid
+            self,
+            name=name,
+            args=args,
+            operator=operator,
+            limit=limit,
+            name_get_uid=name_get_uid,
         )
         if name and _get_use_smart_name_search(self.sudo()) and operator in ALLOWED_OPS:
             limit = limit or 0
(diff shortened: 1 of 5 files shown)


base_name_search_improved/models/ir_model.py
```diff
@@ -175,11 +193,41 @@ def _search_smart_search(self, operator, value):
 class IrModel(models.Model):
     _inherit = "ir.model"
 
-    add_smart_search = fields.Boolean(
-        help="Add Smart Search on search views",
+    add_smart_search = fields.Boolean(help="Add Smart Search on search views",)
```
legalsylvain: there is something wrong in that change.
it is making breaking Test.
https://github.com/OCA/server-tools/runs/5059800229?check_suite_focus=true#step:4:121

legalsylvain (changes requested):
one minor remarks. 
Otherwise LGTM.

Could you also translate spanish text ? (comments, error, etc...)

Thanks !

legalsylvain (approved):
Partial review. Thanks.

Rad0van:
@jjscarafia @kittiu would you care to take a look? Thanx.

kittiu:
But out of curiosity, do you know what is "Smart Search Domain" used for? Because I did add arbitrary domain in res.partner to test, i..e, [('name', '=', 'xxx')]

It seem to affect only on "Smart Search" (Top Search Bar)

For name search field, it passing through this line, https://github.com/Rad0van/server-tools/blob/base_name_search_improved-fp_patches/base_name_search_improved/models/ir_model.py#L87
`args = args or [] + _get_name_search_domain(self.sudo())`
the arg become `['|', ['company_id', '=', False], ['company_id', '=', 1]]` again, hence no effect from the input domain.

If this is expected behavior, and "Smart Search Domain" is only for "Smart Search", should the field be invisible if it is not enbled?

**Note:** my opinion is, both options (Search Domain and Search Fields) should be equally work for both Smart Search and Name Search, in this case this is a bug?

Rad0van:
> But out of curiosity, do you know what is "Smart Search Domain" used for? Because I did add arbitrary domain in res.partner to test, i..e, [('name', '=', 'xxx')]
> 
> It seem to affect only on "Smart Search" (Top Search Bar)
> 
> For name search field, it passing through this line, https://github.com/Rad0van/server-tools/blob/base_name_search_improved-fp_patches/base_name_search_improved/models/ir_model.py#L87 `args = args or [] + _get_name_search_domain(self.sudo())` the arg become `['|', ['company_id', '=', False], ['company_id', '=', 1]]` again, hence no effect from the input domain.
> 
> If this is expected behavior, and "Smart Search Domain" is only for "Smart Search", should the field be invisible if it is not enbled?
> 
> **Note:** my opinion is, both options (Search Domain and Search Fields) should be equally work for both Smart Search and Name Search, in this case this is a bug?

Sure I know what it's supposed to be for but I must admit I didn't test it with it. Will have a look at it. However it seems the same line exists in 13.0: https://github.com/OCA/server-tools/blob/0564c5e426908553a025ea29d2d3e62a0ebffe0c/base_name_search_improved/models/ir_model.py#L89
So probably it's needed to have a look there as well because I agree the Search Domain should be applied on both options.

Rad0van:
> But out of curiosity, do you know what is "Smart Search Domain" used for? Because I did add arbitrary domain in res.partner to test, i..e, [('name', '=', 'xxx')]
> 
> It seem to affect only on "Smart Search" (Top Search Bar)
> 
> For name search field, it passing through this line, https://github.com/Rad0van/server-tools/blob/base_name_search_improved-fp_patches/base_name_search_improved/models/ir_model.py#L87 `args = args or [] + _get_name_search_domain(self.sudo())` the arg become `['|', ['company_id', '=', False], ['company_id', '=', 1]]` again, hence no effect from the input domain.
> 
> If this is expected behavior, and "Smart Search Domain" is only for "Smart Search", should the field be invisible if it is not enbled?
> 
> **Note:** my opinion is, both options (Search Domain and Search Fields) should be equally work for both Smart Search and Name Search, in this case this is a bug?

I know what it's supposed to be for. The code in question is port from 13.0 so I didn't touch it. Will test and if @jjscarafia doesn't provide fix I'll try to do so...

jjscarafia:
Hi! thakns @Rad0van 
I'm in holidays and returning on one week.
We're not using v14 so please fill free to merge/continue.
On my return, if you didn't before, I will take a look to the bug + this PR

Thanks!

jjscarafia (approved):
LGTM

jjscarafia:
@Rad0van @kittiu 
I've work on the fix/improvement of "name_search_domain".
Check here https://github.com/OCA/server-tools/pull/2297

Thanks!

MiguelMachadoM:
i try this version 14.0.1.1.0 and not work same.
Before is want find "jonh Smith" typing "jo smi" get results.
Now need use % for get any result, ex. "jo%smi"
The question is correct? or  have any other problem.

PR #2274, opened 2022-02-07
[MIG] datetime formatter: Migration to 15.0

tbanevicius:


tbanevicius:
Can someone review this merge request? 
It seems CI/runbot fails without any eligible reason.

tbanevicius:
@dreispt @guewen @pedrobaeza Is there any chance someone could review this MR?

pedrobaeza:
Have you signed CLA?

tbanevicius:
@pedrobaeza I did signed CLA. Sent signed **ICLA** document to **cla@odoo-community.org** on February 28.

pedrobaeza:
Now you can review other PRs and ask in exchange that they review yours.

aisopuro:
The actual diff of the migration is here: https://github.com/OCA/server-tools/compare/723ac37584fa0b136edc546c119b83d7c95c0160...tbanevicius:15.0-mig-datetime-formatter

datetime_formatter/tests/test_best_matcher.py
```diff
@@ -50,7 +50,9 @@ def test_user(self):
                 self.rl.with_context(lang=False).best_match().code, lang.code
             )
             # Lang not found in context
-            self.assertEqual(self.rl.with_context(dict()).best_match().code, lang.code)
+            self.assertEqual(
+                self.rl.with_context(**dict()).best_match().code, lang.code
```
aisopuro: `pylint` seems to be mistaken here, this does not seem like a safe change.

The old call in version 14 used an empty dict, which means `with_context` empties the context, removing any `lang` key if it was present.

However this new version unpacks an empty dictionary, which will not override anything. In particular in this test, if `lang` is in the context, this will _not_ clear it, thus making it uncertain whether this test is actually testing the correct functionality.
To placate `pylint`, consider changing it to perhaps use `{}` instead of `dict()`. If that doesn't work, you may be able to [explicitly ignore it](https://stackoverflow.com/a/29562085).

aisopuro (changes requested):
One small linting conflict.

datetime_formatter/tests/test_best_matcher.py
tbanevicius: {} did the trick. 
Now pylint has some issues with readme file, though README.rst, readme folder with CONTRIBUTORS.rst, DESCRIPTION.rst, USAGE.rst is present in the module..

tbanevicius:
Hi @Yajo @victoralmau  any idea why pylint catches error on README.rst? Even though it exists there..

yajo:
No idea, it's weird indeed 😕 

Any idea @sbidoul @moylop260?

moylop260:
> No idea, it's weird indeed 😕
> 
> Any idea @sbidoul @moylop260?

It looks like `sample_module` was identified as a real module in pylint + pre-commit

Since that pre-commit run for each individual path instead of run over the root path only.

I mean,

old: `pylint module/`
new (pre-commit): `pylint module_auto_update/tests/sample_module/models/stuff.py; pylint module_auto_update/tests/sample_module/models/otherfiles`

sbidoul:
@moylop260 your explanation makes sense. Not sure why the error did not trigger in 14.0 though.
Anyway, the easy fix should be to add a README.rst in that `sample_module`.

tbanevicius:
@moylop260  @Yajo  @sbidoul  thanks! Seems now all checks have passed.

tbanevicius:
@Yajo could you review this MR?

yajo (approved):

PR #2268, opened 2022-01-31
[MIG] auditlog: Migration to 15.0

ghost:

PR #2280, opened 2022-02-11
[15.0][MIG] auditlog

HviorForgeFlow:
Standard migration

CC @ForgeFlow


pedrobaeza:
What about #2268 ?

HviorForgeFlow:
Didn't find it before. Thanks for pointing it

PR #1899, opened 2020-09-10
[12.0][FIX] excel_import_export: external ids naming.

LoisRForgeFlow:
Because of the naming given to the external ID's, Odoo interprets all this data as regular module data and when you update the module and those external ID's are not present in the module files, it tries to delete them.

Fixes https://github.com/OCA/server-tools/issues/1898.

cc @kittiu 

@ForgeFlow

PR #2283, opened 2022-02-12
[14.0][FIX] excel_import_export: external ids naming.

kittiu:
FW Port from https://github.com/OCA/server-tools/pull/1899

Because of the naming given to the external ID's, Odoo
interprets all this data as regular module data and when
you update the module and those external ID's are not present
in the module files, it tries to delete them.

### DIFF
--- a/excel_import_export/models/xlsx_import.py
+++ b/excel_import_export/models/xlsx_import.py
@@ -50,7 +50,7 @@ class XLSXImport(models.AbstractModel):
             ModelData.create(
                 {
                     "name": "{}_{}".format(record._table, record.id),
-                    "module": "excel_import_export",
+                    "module": "__excel_import_export__",
                     "model": record._name,
                     "res_id": record.id,
                 }


luc-demeyer (changes requested):
This fix prevents the deletion of the records since __excel_import_export__ is a non-existing module name.

But I think we should be consistent and hence change also the following line accordingly:

https://github.com/OCA/server-tools/blob/9378a9f649449d8c42af4371de24f53ca09cf417/excel_import_export/models/xlsx_import.py#L209

or "{}.{}".format("xls", uuid.uuid4()) : xls -> __excel_import_export

Otherwise we have two different  module names in the xml_id for records updated or created by the excel_import_export module.

Mantux11:
In commit there is `[12.0]`, but this PR would merge to 14.0. Change that:)

kittiu:
I thought for FW port it is correct like this, just pull the commit here. No?

Mantux11:
Oh well. I don't catch that. Maybe that's correct

Mantux11 (approved):

norlinhenrik (approved):
The suggestion by @luc-demeyer is included in #2329

norlinhenrik:
This PR is approved by 2 reviewers. Can we merge it?

norlinhenrik:
According to this PR and the discussion, `xlsx_import.py` has `__excel_import_export__` **twice** in both version [15.0](https://github.com/OCA/server-tools/blob/15.0/excel_import_export/models/xlsx_import.py) and [16.0](https://github.com/OCA/server-tools/blob/16.0/excel_import_export/models/xlsx_import.py).

This PR has `__excel_import_export__` only **once**. With no activity in a year, I think it is time to close it.

PR #2287, opened 2022-02-14
[12.0] IMP base_model_restrict_update: readonly users can create wizards

eLBati:
This is useful for reporting wizards typically

### DIFF
--- a/base_model_restrict_update/models/ir_model_access.py
+++ b/base_model_restrict_update/models/ir_model_access.py
@@ -49,13 +49,11 @@ class IrModelAccess(models.Model):
 
     @api.model
     def _readonly_exclude_models(self):
-        transient_models = self.env["ir.model"].sudo().search(
-            [("transient", "=", True)]).mapped("model")
         skipped_models = self.env["ir.model"].sudo().search([
-            ("skip_check_for_readonly_users", "=", True)
+            "|", ("transient", "=", True), ("skip_check_for_readonly_users", "=", True)
         ]).mapped("model")
         # Models updtate/create by system, and should be excluded from checking
-        return skipped_models + transient_models + self.sudo().search([
+        return skipped_models + self.sudo().search([
             ("group_id", "=", False),
             "|", ("perm_write", "=", True),
             "|", ("perm_create", "=", True), ("perm_unlink", "=", True)
--- a/base_model_restrict_update/models/ir_model_access.py
+++ b/base_model_restrict_update/models/ir_model_access.py
@@ -48,8 +48,10 @@ class IrModelAccess(models.Model):
 
     @api.model
     def _readonly_exclude_models(self):
-        """ Models updtate/create by system, and should be excluded from checking """
-        return self.sudo().search([
+        transient_models = self.env["ir.model"].sudo().search(
+            [("transient", "=", True)]).mapped("model")
+        # Models updtate/create by system, and should be excluded from checking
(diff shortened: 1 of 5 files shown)


yostashiro (approved):
Code review + functional test. LGTM 👍

eLBati:
Hi @yostashiro , please note I just updated this PR with https://github.com/OCA/server-tools/pull/2287/commits/263144692a357360503bf3c82deb100478828b11

yostashiro:
> Hi @yostashiro , please note I just updated this PR with [2631446](https://github.com/OCA/server-tools/commit/263144692a357360503bf3c82deb100478828b11)

@eLBati If I'm not mistaken, I think this change opens more possibility to bypass the ACL check in an unexpected way for non-readonly users.

To make the new settings applicable only to the readonly users, how about updating this part
https://github.com/OCA/server-tools/blob/311142a8aab02bb9f6a2fee9fa0d01d539bacaee/base_model_restrict_update/models/ir_model_access.py#L32-L37
to something like this:
```py
    @api.model
    def _test_readonly(self, model):
        if not self.env.user.is_readonly_user:
            return True
        if model not in exclude_models self._readonly_exclude_models():
            return True
        return False
```

I also think the new changes deserve updates on tests and readme.

eLBati:
```
        if not self.env.user.is_readonly_user:
            return True
```
:thinking: why should `_test_readonly` return True if user is not readonly?

> I also think the new changes deserve updates on tests and readme

Right, I proceed

eLBati:
> Right, I proceed

@yostashiro done

yostashiro:
> 🤔 why should `_test_readonly` return True if user is not readonly?

@eLBati You are right. I made a mistake.

I think, however, it's worth adjusting the method to
```py
    @api.model
    def _test_readonly(self, model):
        if not self.env.user.is_readonly_user:
            return False
        if model in exclude_models self._readonly_exclude_models():
            return False
        return True
```
to skip unnecessary searches for non-readonly users now that you are adding more searches in _readonly_exclude_models().

eLBati:
@yostashiro ok, changed, thanks

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -48,8 +49,13 @@ def _test_restrict_update(self, model):
...
     def _readonly_exclude_models(self):
-        """ Models updtate/create by system, and should be excluded from checking """
-        return self.sudo().search([
+        transient_models = self.env["ir.model"].sudo().search(
+            [("transient", "=", True)]).mapped("model")
+        skipped_models = self.env["ir.model"].sudo().search([
+            ("skip_check_for_readonly_users", "=", True)
+        ]).mapped("model")
+        # Models updtate/create by system, and should be excluded from checking
+        return skipped_models + transient_models + self.sudo().search([
```
yostashiro:
```suggestion
        skipped_models = self.env["ir.model"].sudo().search(
            ["|", ("transient", "=", True), ("skip_check_for_readonly_users", "=", True)]
        ).mapped("model")
        # Models updtate/create by system, and should be excluded from checking
        return skipped_models + self.sudo().search([
```
Let's combine the searches into one. ;)

eLBati: committed, thanks

yostashiro (approved):
LGTM 👍

newtratip (approved):
LGTM :+1:

PR #2289, opened 2022-02-16
[14.0] Add module cron_daylight_saving_time_resistant

florian-dacosta:
The module allows to configure crons so they always run at the same hour, no matter the timezone and the change of hours.
For instance, if you are in Brussels and you have a cron that run every days at 11H (local time), when swtiching from summer time to winter time, the  cron will be planned for the next day at 10H (local time)
At the contrary when switching from winter time to summer time, the cron that was configured to run at 11H will be planned for 12H.
If it is important that the cron always run at the same hour no matter the change of time, then you can check the box "daylight_saving_time_resistant" on the cron.


manengel:
The cron job required for this does take the user's time zone into account, does it?

thomaspaulb:
It does look like it [takes the user into account when creating an env to take the timezone from](https://github.com/OCA/server-tools/pull/2289/files#diff-03fdd80d5f6bf892d9402e38f658370d4a407f36355c4988602f7c09b891eaedR55)?

I like this a little bit better than [this one](https://github.com/OCA/server-tools/pull/2348/files), because it doesn't write to other cron jobs, it only writes to the next execution time of the current cron job, which seems less invasive to me and less prone to create conflict issues with other cron jobs.

thomaspaulb (approved):
Code review

thomaspaulb:
@NL66278 Maybe another review?

NL66278:
Normally I would say that code should document itself, but I have a hard time understanding how this works. So some documentation of what happens seems to be in order. Then I might be able to do a proper review.

thomaspaulb:
@hparfr @clementmbr Since you reviewed the 16.0 version, maybe you can check if this one can be merged too?

gurneyalex (approved):

PR #2290, opened 2022-02-16
[14.0] ADD sql_export_external_database

florian-dacosta:
This module extends sql_export to be able to make the query on an external database (Common case would be a replication of the production database).
The goal beeing to not overload the production database server in case many heavy queries are done frequently.


sql_export_external_database/models/sql_export.py
```diff
@@ -0,0 +1,29 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, fields, models
+
+from ..sql_db import get_external_cursor
+
+
+class SqlExport(models.Model):
+    _inherit = "sql.export"
```
legalsylvain, line 9: why not inherit from ``sql.request.mixin``. This feature could be available for all modules that depends on this module.

legalsylvain (changes requested):
Hi. Nice addition.
I'll see better this boolean (and your commit) directly in the module ``sql_request_abstract`` instead of having a new module.
Otherwise, LGTM.
thanks !

sql_export_external_database/models/sql_export.py
florian-dacosta: @legalsylvain Thanks for the review.

You are right, I could put this directly in `sql.request.mixin`.
But I wonder if it is good. This feature should be used exclusively for reading queries (SELECT/COPY) it can't work with the modes `view` and `materialized_view` since we need the views created in the production database.
If I add this in sql.request.mixin, I guess I should add a check to raise if we got the `use_external_database` for a `view` or `materialized_view` mode, what do you think ?

Also I am not sure it should go in `sql_request_abstract` directly. I is a feature for some specific project with heavy use/big database IMO and is useless for at least 80% of the users (if not more...).
So  it seems best to me to leave this in a separate module

Also we'll need 2 modules instead of one, indeed, we still have to have a module `sql_export_*` to add the new field `use_external_database` in the view.

Should I do something like : 
sql_request_abstract_external_database (adding a raise if we use an external database with a other mode than (fetchone/fetchall/stdout)
and sql_export_external_database with only the field in the view (could be auto install)
Or is it too much?

legalsylvain: Hi. 
Well, indeed, the ``use_external_database`` will not work directly with ``bi_sql_editor``, but it could ! with some works. (and it could be great, to avoid to increase production database size, with a lot of materialized views). But it requires patch to make the request on the correct database.

My proposal is : 
- move your patch against ``sql_request_abstract`` module
- add the field ``use_external_database`` in ``sql_export`` view. (and not in ``bi_sql_editor`` view.
- (no new modules) keep it simple.

> it is a feature for some specific project with heavy...

Indeed, but it's just a boolean default False. it doesnt hurt. Just an option. and it makes the feature visible. (if you add a new module, most of the people will not discover it). There a lot of other options that are not used a lot. No need to have extra modules (like having restriction by user or group). --> it reduces maintenance when migrating modules.

what do you think ?

florian-dacosta: Ok, let's do this.
I'll also add a constraint on the `use_external_database` to raise if someone try to check it when there are no config for external database, because this feature needs a configuration on server side (global odoo config file)

PR #2291, opened 2022-02-16
[14.0][IMP] sql_export: Make possible to make sql queries on an external database

florian-dacosta:
Replace https://github.com/OCA/server-tools/pull/2290

Make possible to make queries on an external database (Common case would be a replication of the production database).
The goal beeing to not overload the production database server in case many heavy queries are done frequently.

### DIFF
--- a/sql_request_abstract/models/sql_request_mixin.py
+++ b/sql_request_abstract/models/sql_request_mixin.py
@@ -12,8 +12,10 @@ from io import BytesIO
 from psycopg2 import ProgrammingError
 from psycopg2.sql import SQL
 
-from odoo import _, api, fields, models
-from odoo.exceptions import UserError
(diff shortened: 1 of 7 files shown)


sql_request_abstract/models/sql_request_mixin.py
```diff
@@ -182,42 +204,56 @@ def _execute_sql_request(
         else:
             raise UserError(_("Unimplemented mode : '%s'" % mode))
 
+        # helper usefull for sql_export_external_database module where
```
legalsylvain: this comment is outdated.

legalsylvain (approved):
Thansk ! 
A minor remark. Otherwise, LGTM.

(code review. no test.)

ntsirintanis (approved):

PR #2293, opened 2022-02-18
[14.0] Add rpc_helper

simahawk:
Decorate an Odoo model class like this::

    from odoo.addons.rpc_helper.decorator import disable_rpc

    @disable_rpc()
    class AverageModel(models.Model):
        _inherit = "avg.model"

This will disable ALL calls.

To selectively disable only some methods::

    @disable_rpc("create", "write", "any_method")
    class AverageModel(models.Model):
        _inherit = "avg.model"

Old RFC on Odoo core https://github.com/odoo/odoo/issues/62299


legalsylvain (approved):
nice !

rpc_helper/tests/test_decorator.py
```diff
@@ -0,0 +1,26 @@
...
+
+from ..decorator import disable_rpc
+
+
+@disable_rpc()
+class All:
+    pass
+
+
+@disable_rpc("create", "write")
```
SilvioC2C, line 21: Since we have `target._disable_rpc = config if len(config) > 1 else ("all",)` in `decorator.py`, what happens if we want to disable exactly 1 method (eg: `@disable_rpc("create",)`)?

simahawk: good catch, added a test for it and fixed. Grazie caro :)

SilvioC2C (approved):
LGTM :muscle:

PR #2294, opened 2022-02-21
[14.0] rpc_helper: add config via UI

simahawk:
Enable debug mode and go to "Technical -> Database Structure -> Models".

Open the model that you like to configure and go to the tab "RPC config".

There you see a text field which supports JSON configuration.

The configuration is the same you can pass via decorator.
The only difference is that you have to wrap values in a dictionary
like `{"disable": [...values...]}`.

To disable all calls::

    {
        "disable": ["all"],
    }

To disable only some methods::

    {
        "disable": ["create", "write", "another_method"],
    }

NOTE: on the resulting JSON will be automatically formatted on save for better readability.

### DIFF
--- a/rpc_helper/__init__.py
+++ b/rpc_helper/__init__.py
@@ -1,1 +1,2 @@
+from . import models
 from .hooks import post_load_hook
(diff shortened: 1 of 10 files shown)


SilvioC2C (approved):
LGTM :+1:

ajaniszewska-dev (approved):

mmequignon (approved):

PR #2297, opened 2022-02-22
[FIX] search_improved: use search domain on name_search

jjscarafia:
Previous to this commit "name_search_domain" was only used on search view but not in name_search.
We also change and use expression.AND instead of directly append the domains


gurneyalex:
@jjscarafia can you check the CI status?

PR #2298, opened 2022-02-22
14.0 1831 port rebased and cleanup

sebastienbeau:
Port PR #1831
- squash for simplier history
- add test improvement + finish to clean test using odoo-test-helper


base_exception/models/base_exception.py
```diff
@@ -35,7 +39,7 @@ class ExceptionRule(models.Model):
         "           are evaluated with several records",
```
hparfr: There is no help for by_method

sebastienbeau: @hparfr done

hparfr (approved):

base_exception/tests/test_base_exception.py
```diff
@@ -2,115 +2,107 @@
...
-        potest1.action_ignore_exceptions()
-        self.assertTrue(potest1.ignore_exception)
-        self.assertFalse(potest1.exceptions_summary)
-        potest1.button_confirm()
-        self.assertTrue(potest1.state == "purchase")
+        self.po.action_ignore_exceptions()
+        self.assertTrue(self.po.ignore_exception)
+        self.assertFalse(self.po.exceptions_summary)
+        self.po.button_confirm()
+        self.assertTrue(self.po.state == "purchase")
```
bealdav: not blocking, just for information but `assert self.po.state == "purchase"` is a more pythonic syntax

bealdav (approved):

base_exception/tests/test_base_exception.py
sebastienbeau: I didn't see it, thx. I replace it with : self.assertEqual(self.po.state, "purchase")

hparfr: It's not the same !

sebastienbeau: @hparfr you talk about 

 ```assert self.po.state == "purchase" ``` (this is indeed not valid)

or

 ```self.assertEqual(self.po.state, "purchase") ```

bealdav: Why not valid?
Assert is s built-in dedicated for tests. In production running configuration, assert outside of tests, is not executed

hparfr: David says you can use the `assert` keyword instead of `self.assertSomething()` method.
It's one of the key arguments of pytest: 

> Due to pytest’s detailed assertion introspection, only plain assert statements are used
> [...]    Detailed info on failing assert statements (no need to remember self.assert* names)
https://docs.pytest.org/en/7.0.x/index.html

hparfr: And yes, `assertEquals`, in this case is more expressive than `assertTrue`, but it was not the point here :dart: .

bealdav: https://realpython.com/python-assert-statement/#testing-your-code-with-assertions

florian-dacosta (approved):

PR #2301, opened 2022-02-24
[13.0] [ADD] base_import_xlsx: New module.

NL66278:
This module is needed for installaions with python library xlrd > 2.0,
because this version dropped support for xlsx files.


pedrobaeza:
Why not pinning library version in requirements?

NL66278:
@pedrobaeza According to the developers of xlrd they dropped the support fo xlsx because they could not or would not maintain it, and according to them they could not guarantee its security. Of course pinning might solve it for the moment, but we encountered a situation on odoo.sh where the library was updated, as Odoo only sets a minimum version.

pedrobaeza:
OK then.

NL66278:
runbout fails with this unrelated message. Suggestions on how to solve this are welcome:
  File "/mnt/data/odoo-addons-dir/setup/company_country/odoo/addons/company_country/models/res_config.py", line 39, in load_company_country
    "COUNTRY environment variable with country code is not "
odoo.tools.convert.ParseError: "COUNTRY environment variable with country code is not set and no localization module is marked to be installed.
None" while parsing /mnt/data/odoo-addons-dir/setup/company_country/odoo/addons/company_country/data/res_config.xml:2, near
<odoo noupdate="1">
    <function name="load_company_country" model="company.country.config.settings"/>
</odoo>

sbidoul:
Hi @NL66278,

The `company_country` module requires a COUNTRY environment variable to be set. If not it apparently crashes.

This environment variable is [set in the travis config](https://github.com/OCA/server-tools/blob/0564c5e426908553a025ea29d2d3e62a0ebffe0c/.travis.yml#L39).

Runboat does not parse the travis config, as runbot did, so it does not have this environment variable set.

That said:
- The readme of company_country says this module is not necessary for Odoo >= 13, so why do we keep it here ?
- Creating a module that places new constraints on Odoo environmental conditions to just "not crash" is questionnable, IMO. It should have a default value if the COUNTRY env var is not set
- The runboat failure is not blocking (it reports status here so we have an easy link to open the instance), but will become blocking when we switch from Travis to GitHub actions, since GitHub actions tests use the same mechanism as runboat for installing - so this has to be fixed sooner or later

NL66278:
@sbidoul Thank you for your explanations!

If company_country is causing this problem and is no longer needed, I will gladly remove it. Note that I did not touch this module until now.

Just to understand things, how is runboat triggered/configured to run now?

NL66278:
See here: https://github.com/OCA/server-tools/pull/2302

sbidoul:
> how is runboat triggered/configured to run now?

Runboat is launching a build for every push to PRs and branches 10 to 15. There is no specific configuration per repo. Not sure if that answer your question ?

NL66278:
@sbidoul I is clear now, thank you! Do you have merge rights in this repo? Please have a look at the company_contry removal PR, if you have time. I could rebase this one if that one is merged to get a green runboat...

sbidoul:
I have merge right everywhere ;) But it's probably best to ask authors of that module if they agree to remove it.

base_import_xlsx/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Base import Excel",
+    "version": "13.0.0.0.0",
```
thomaspaulb: It's common to start at `1.0.0` not at `0.0.0` (probably a debate that goes back to the start of computing, but lets not go into that here)

base_import_xlsx/models/base_import.py
```diff
@@ -0,0 +1,85 @@
...
+        sheet = book.worksheets[0]
+        for row in sheet.rows:
+            values = []
+            for cell in row:
+                if isinstance(cell, openpyxl.cell.read_only.EmptyCell):
+                    values.append("")
+                    continue
+                # Would like to skipp NULL values, but for some incomprehensible
+                # reason TYPE_NULL and TYPE_NUMERIC have the same value in openpyxl.
+                if cell.value is None:
```
thomaspaulb, line 52: I saw that the types are both `n` indeed, strange... but I think this is the right solution: checking for `is None`.

thomaspaulb (changes requested):
Solid work! Code review only. Tests appear to be running well.

thomaspaulb:
@NL66278 Perhaps change the version number and rebase, to see if CI is now running well?

NL66278:
@thomaspaulb I changed the version and everything is green now.

thomaspaulb (approved):

thomaspaulb:
@fkantelberg Maybe we could ask for your support here in return :-)

base_import_xlsx/models/base_import.py
```diff
@@ -0,0 +1,85 @@
...
+                    values.append("")
+                    continue
+                if cell.data_type == "d" and isinstance(cell.value, datetime.date):
```
fkantelberg, line 55: Is it intended that `timedelta` and `time` aren't handled? If I see correctly they will just appended in L82 and not even converted to string. I know it might be a weird case for export->import but if a customer is in between I could imagine it happening with the automatic typing of the calc tools.

base_import_xlsx/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+It is expected that in future versions of Odoo support for xlsx files will be fully
+restored. However, as Odoo will not ontroduce new libraries in stable versions,
```
fkantelberg: `introduce`

base_import_xlsx/models/base_import.py
NL66278: timedelta and time are not in the Odoo datatypes, so I do not quite see how this should be supported. I will convert unknown datatypes to string not in L82.

base_import_xlsx/readme/ROADMAP.rst
NL66278: Corrected

base_import_xlsx/models/base_import.py
fkantelberg: Conversion to str would be handling it correctly I guess.

NL66278:
@fkantelberg Thanks for your input! Can you approve the PR now?

fkantelberg:
The error regarding the unhandled types if data_type is d isn't yet handled. See the following example. No one would understand what this message means. I think that should be handled with a better error or just convert to str.

test
[test.xlsx](https://github.com/OCA/server-tools/files/14176288/test.xlsx)

PR #2303, opened 2022-02-25
[14.0][ADD] base_generate_code module

Kev-Roche:
This module with a mixin model,  can be plugged to other models, in order to generate random codes  (e.g. for customers' purposes : coupon code, Gift card code, etc.).

It's inspired by <https://github.com/akretion/sale-promotion/tree/14.0-sale-coupon-custom-code-format/sale_coupon_custom_code_format> from @paradoxxxzero

Two questions : 
1) It is the good place for this module? 
2) The test include a fake model, maybe in a second time, split the tests in a second module for test only ?

cc @hparfr, @PierrickBrun

### DIFF
--- a/base_generate_code/models/code_format_mixin.py
+++ b/base_generate_code/models/code_format_mixin.py
@@ -1,0 +1,89 @@
+# Copyright (C) 2021 Akretion (<http://www.akretion.com>).
+# @author Florian Mounier <florian.mounier@akretion.com>
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from random import choice
+from string import ascii_lowercase, ascii_uppercase, digits
+
+from odoo import _, fields, models
+from odoo.exceptions import ValidationError
+
+
+class CodeFormatMixin(models.AbstractModel):
+    _name = "code.format.mixin"
+    _description = "Customizable code format Mixin"
+
+    #  1/ _code_mask must be added in the related model
+    # e.g. in "gift.card":
+    #     _code_mask = {
+    #       "mask": "code_mask",
+    #       "template": "gift_card_tmpl_id"
+    #      }
+
+    # 2/ In addition, update a method of the target model to
+    # compute the code. For example the create method:
+    #     @api.model
+    #     def create(self, vals):
+    #         res = super().create(vals)
+    #         vals["code"] = res._generate_code()
+
+    _code_mask = {}
+
+    _default_mask = "XXXXXX-00"
+    _forbidden_characters = "iI1oO0"
+    _dedup_max_retries = 20
+
+    def _generate_code(self):
+        choices = self._get_choice_collections()
+        for rec in self:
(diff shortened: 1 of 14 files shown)


pedrobaeza:
Why not removing the word `customer` from the module name? Or is it linked to customers somehow?

Kev-Roche:
> Why not removing the word `customer` from the module name? Or is it linked to customers somehow?

Indeed, the module, in some ways, can be usefull for partners in general or users.

base_generate_code/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module can be plugged to other models, in order to generate random codes (e.g. for customer's purposes : coupon code, Gift card code, etc.).
```
hparfr: The goal is to have human readable code: not too long, random enough, without some ambiguous char like ( "O" "0", "I", "L"), not memorable, without meaning (no words), acceptable as a barcode (like code128).

base_generate_code/models/code_format_mixin.py
```diff
@@ -0,0 +1,89 @@
...
+            while len(self.search([("code", "=", code)])):
+                code = rec._generate_code_from_mask(choices, mask)
+                retries += 1
+                if retries > self._dedup_max_retries:
+                    raise ValidationError(
+                        _("Unable to generate a non existing random code.")
+                    )
+            return code
+
+    code = fields.Char(compute=_generate_code, readonly=True, store=True)
```
hparfr: fields are in the top  usually

hparfr (changes requested):
You may update the readme to more reflects the purpose of the module.

base_generate_code/readme/DESCRIPTION.rst
Kev-Roche: good idea, it's done.

Kev-Roche:
@pedrobaeza , @PierrickBrun what do you think of the failed runboat/build here?

pedrobaeza:
https://github.com/sbidoul/runboat/issues/48

Kev-Roche:
> [sbidoul/runboat#48](https://github.com/sbidoul/runboat/issues/48)

Thank you!

hparfr (approved):

rousseldenis (approved):
LGTM

sebastienbeau:
desolé mon petit bot pour l'erreur de frappe : /ocabot merge patch

base_generate_code/models/code_format_mixin.py
```diff
@@ -0,0 +1,89 @@
...
+                char
+                for char in ascii_uppercase
+                if char not in self._forbidden_characters
+            ],
+            "x": [
+                char
+                for char in ascii_lowercase
+                if char not in self._forbidden_characters
+            ],
+            "0": digits,
```
petrus-v, line 88: I was just given an eye on this module, I suspect that `0` can be part of the final result as there is no filtering chars here like for ascii_uppercase and ascii_lowercase. I'm wondering why ?

PR #2305, opened 2022-03-01
[14.0] base_view_inheritance_extension: wrap import

simahawk:


mmequignon (approved):

thomaspaulb (approved):

pedrobaeza:
This technique is not needed in v13+. Why do you need it?

simahawk:
@pedrobaeza well, since the instance does not break when you start it, if the module is installed you won't notice you miss a lib till the piece of code that requires it runs and breaks.
As overriding ir_ui_view is quite sensitive, we should probably let it fail gracefully and skip the this machinery if astor is not there. This way, rendering views won't be deadly broken.

pedrobaeza:
OK then:

/ocabot merge patch

PR #2306, opened 2022-03-06
[BCP][13.0] base_view_inheritance_extension python_dict

thomaspaulb:
Backport from 14.0 of:

- [x] https://github.com/OCA/server-tools/pull/2216
- [x] https://github.com/OCA/server-tools/pull/2305

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 80
+max-line-length = 88
 max-complexity = 16
 # B = bugbear
 # B9 = bugbear opinionated (incl line length)
--- a/base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
+++ b/base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
@@ -1,11 +1,19 @@
 # Copyright 2016 Therp BV <http://therp.nl>
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
 from lxml import etree
 
-from odoo.tests.common import TransactionCase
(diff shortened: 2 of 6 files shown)


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 80
+max-line-length = 88
```
ntsirintanis, line 2: Isn't there a command that ignores this line check? So that we won't have to edit .flake8 itslef

ntsirintanis (commented):
We would also need a README here I guess, otherwise LGTM

.flake8
thomaspaulb: In the latest OCA repository template, they [ignore regular line length warnings](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.flake8#L8) and instead replace by [bugbear line warnings of max-line-length 88](https://github.com/OCA/oca-addons-repo-template/blob/master/src/.flake8#L2). Bugbear looks at the (rolling?) average, so that if one line goes over 88 to like 92 but the rest of the lines are OK, it will tolerate that one line.

thomaspaulb: So this is just an application of the latest template really.

thomaspaulb:
@ntsirintanis About the README, the module already has one - this is just a backport of some refactored functionality.

ntsirintanis (approved):

base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
```diff
@@ -1,11 +1,19 @@
 # Copyright 2016 Therp BV <http://therp.nl>
```
NL66278: Make this https (and maybe extend to 2022).

NL66278 (approved):
:+1: LGTM

AaronHForgeFlow (approved):
LGTM

thomaspaulb:
@hbrunn Could you merge?

PR #2307, opened 2022-03-08
[14.0][IMP] onchange_helper: Call onchange only if fields are concerned by

rousseldenis:
As the onchange method can be a little bulky, we don't call it if fields passed as
arguments are not in _onchange_methods


rousseldenis:
@guewen @lmignon @sbejaoui @len-foss

lmignon (approved):
Makes sense. LGTM (code review)

guewen (approved):

atchuthan (approved):

PR #2308, opened 2022-03-08
[15.0][MIG] [INVALID] base search fuzzy

Shide:
Migration of base search fuzzy

MT-358 @moduon


Shide:
@pedrobaeza please check out this pr in order to migrate other modules that depends on this module

pedrobaeza:
It seems the commit history is not correct. Can you please check?

Shide:
I've opened a new PR with the commit history https://github.com/OCA/server-tools/pull/2310

PR #2312, opened 2022-03-11
[14.0][IMP] video_link add the possibility to not specify any provider and to manually set url

sebastienbeau:
In case that you import data from a external provider that give you links for video, we need to have a simple posibilty to set the url and thumbnail url without configuring a provider


sebastienbeau:
@simahawk @shopinvader

PR #1683, opened 2019-09-30
[9.0] [ADD] Base User Role History

baimont:

PR #2314, opened 2022-03-16
[MIG] [10.0]  base_user_role_history from 9.0

FrancoMaxime:
Migration of baser_user_role_history from 9.0
(related to https://github.com/OCA/server-tools/pull/1683 )


FrancoMaxime:
@sbidoul Bonjour Stéphane, pourrais tu regarder au runbot?

Merci

apineux (approved):

sbidoul:
The error is 
```
ERROR: Could not find a version that satisfies the requirement dns (from -r test-requirements.txt (line 3)) (from versions: none)
ERROR: No matching distribution found for dns (from -r test-requirements.txt (line 3))
```

That is because the `letsencrypt` addon on the 10.0 branch has `dns` in its external dependencies.
And `dns` is not a package that is available on PyPI, or rather, it is on PyPI but has no file to download.

I don't know which PyPI package is supposed to provide this `dns` external dependency.
@NL66278 @blyxxyz do you know ?

NL66278:
@sbidoul The package is called dnspython

sbidoul:
Thanks Ronald! A fix for that issue and a couple of other is in https://github.com/OCA/server-tools/pull/2315.

base_user_role_history/models/base_user_role_line_history.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
-# Copyright 2019 ACSONE SA/NV
+# Copyright 2022 ACSONE SA/NV
```
sbidoul: Please keep 2019 (or put 2019-2022, but that is not necessary)

sbidoul (commented):
A minor remark about the copyright that which does not need changing. Otherwise LGTM

sbidoul:
The runboat issue does not prevent merging.

base_user_role_history/models/base_user_role_line_history.py
FrancoMaxime: Changes are made

baimont (approved):

PR #2318, opened 2022-03-22
[14.0] auditlog: Add option to Eliminate user and fields

atchuthan:
Added option to eliminate specific Users and fields to exclude from the auditlog.

cc: @SodexisTeam

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -142,6 +142,19 @@ class AuditlogRule(models.Model):
         "Capture Record",
         help="Select this if you want to keep track of Unlink Record",
     )
+    users_to_exclude_ids = fields.Many2many(
+        "res.users",
+        string="Users to Exclude",
+        context={"active_test": False},
+        states={"subscribed": [("readonly", True)]},
+    )
+
+    fields_to_exclude_ids = fields.Many2many(
+        "ir.model.fields",
+        domain="[('model_id', '=', model_id)]",
+        string="Fields to Exclude",
+        states={"subscribed": [("readonly", True)]},
+    )
 
     _sql_constraints = [
         (
@@ -265,6 +278,7 @@ class AuditlogRule(models.Model):
         """Instanciate a create method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude = self.mapped("users_to_exclude_ids")
 
         @api.model_create_multi
         @api.returns("self", lambda value: value.id)
@@ -286,6 +300,8 @@ class AuditlogRule(models.Model):
                     new_values[new_record.id][fname] = field.convert_to_read(
                         new_record[fname], new_record
                     )
+            if self.env.user in users_to_exclude:
+                return new_records
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -307,6 +323,8 @@ class AuditlogRule(models.Model):
             new_values = {}
             for vals, new_record in zip(vals_list2, new_records):
                 new_values.setdefault(new_record.id, vals)
+            if self.env.user in users_to_exclude:
+                return new_records
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -324,6 +342,7 @@ class AuditlogRule(models.Model):
         """Instanciate a read method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude = self.mapped("users_to_exclude_ids")
 
         def read(self, fields=None, load="_classic_read", **kwargs):
             result = read.origin(self, fields, load, **kwargs)
@@ -343,6 +362,8 @@ class AuditlogRule(models.Model):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude:
+                return result
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -360,6 +381,7 @@ class AuditlogRule(models.Model):
         """Instanciate a write method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude = self.mapped("users_to_exclude_ids")
 
         def write_full(self, vals, **kwargs):
             self = self.with_context(auditlog_disabled=True)
@@ -378,6 +400,8 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            if self.env.user in users_to_exclude:
+                return result
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -400,6 +424,8 @@ class AuditlogRule(models.Model):
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
+            if self.env.user in users_to_exclude:
+                return result
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -417,6 +443,7 @@ class AuditlogRule(models.Model):
         """Instanciate an unlink method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude = self.mapped("users_to_exclude_ids")
 
         def unlink_full(self, **kwargs):
             self = self.with_context(auditlog_disabled=True)
@@ -428,6 +455,8 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            if self.env.user in users_to_exclude:
+                return unlink_full.origin(self, **kwargs)
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -442,6 +471,8 @@ class AuditlogRule(models.Model):
         def unlink_fast(self, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude:
+                return unlink_fast.origin(self, **kwargs)
             rule_model.sudo().create_logs(
                 self.env.uid,
                 self._name,
@@ -475,17 +506,16 @@ class AuditlogRule(models.Model):
         log_model = self.env["auditlog.log"]
         http_request_model = self.env["auditlog.http.request"]
         http_session_model = self.env["auditlog.http.session"]
+        model_model = self.env[res_model]
+        model_id = self.pool._auditlog_model_cache[res_model]
+        auditlog_rule = self.env["auditlog.rule"].search([("model_id", "=", model_id)])
+        fields_to_exclude = auditlog_rule.fields_to_exclude_ids.mapped("name")
         for res_id in res_ids:
-            model_model = self.env[res_model]
             name = model_model.browse(res_id).name_get()
-            model_id = self.pool._auditlog_model_cache[res_model]
-            auditlog_rule = self.env["auditlog.rule"].search(
-                [("model_id", "=", model_id)]
-            )
             res_name = name and name[0] and name[0][1]
             vals = {
                 "name": res_name,
-                "model_id": self.pool._auditlog_model_cache[res_model],
+                "model_id": model_id,
                 "res_id": res_id,
                 "method": method,
                 "user_id": uid,
@@ -498,18 +528,26 @@ class AuditlogRule(models.Model):
                 new_values.get(res_id, EMPTY_DICT), old_values.get(res_id, EMPTY_DICT)
             )
             if method == "create":
-                self._create_log_line_on_create(log, diff.added(), new_values)
+                self._create_log_line_on_create(
+                    log, diff.added(), new_values, fields_to_exclude
+                )
             elif method == "read":
                 self._create_log_line_on_read(
-                    log, list(old_values.get(res_id, EMPTY_DICT).keys()), old_values
+                    log,
+                    list(old_values.get(res_id, EMPTY_DICT).keys()),
+                    old_values,
+                    fields_to_exclude,
                 )
             elif method == "write":
                 self._create_log_line_on_write(
-                    log, diff.changed(), old_values, new_values
+                    log, diff.changed(), old_values, new_values, fields_to_exclude
                 )
             elif method == "unlink" and auditlog_rule.capture_record:
                 self._create_log_line_on_read(
-                    log, list(old_values.get(res_id, EMPTY_DICT).keys()), old_values
+                    log,
+                    list(old_values.get(res_id, EMPTY_DICT).keys()),
+                    old_values,
+                    fields_to_exclude,
                 )
 
     def _get_field(self, model, field_name):
@@ -534,11 +572,14 @@ class AuditlogRule(models.Model):
                 cache[model.model][field_name] = field_data
         return cache[model.model][field_name]
 
-    def _create_log_line_on_read(self, log, fields_list, read_values):
+    def _create_log_line_on_read(
+        self, log, fields_list, read_values, fields_to_exclude
+    ):
         """Log field filled on a 'read' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        fields_to_exclude = fields_to_exclude + FIELDS_BLACKLIST
         for field_name in fields_list:
-            if field_name in FIELDS_BLACKLIST:
+            if field_name in fields_to_exclude:
                 continue
             field = self._get_field(log.model_id, field_name)
             # not all fields have an ir.models.field entry (ie. related fields)
@@ -565,11 +606,14 @@ class AuditlogRule(models.Model):
             vals["old_value_text"] = old_value_text
         return vals
 
-    def _create_log_line_on_write(self, log, fields_list, old_values, new_values):
+    def _create_log_line_on_write(
+        self, log, fields_list, old_values, new_values, fields_to_exclude
+    ):
         """Log field updated on a 'write' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        fields_to_exclude = fields_to_exclude + FIELDS_BLACKLIST
         for field_name in fields_list:
-            if field_name in FIELDS_BLACKLIST:
+            if field_name in fields_to_exclude:
                 continue
             field = self._get_field(log.model_id, field_name)
             # not all fields have an ir.models.field entry (ie. related fields)
@@ -614,11 +658,14 @@ class AuditlogRule(models.Model):
             vals["new_value_text"] = new_value_text
         return vals
 
-    def _create_log_line_on_create(self, log, fields_list, new_values):
+    def _create_log_line_on_create(
+        self, log, fields_list, new_values, fields_to_exclude
+    ):
         """Log field filled on a 'create' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        fields_to_exclude = fields_to_exclude + FIELDS_BLACKLIST
         for field_name in fields_list:
-            if field_name in FIELDS_BLACKLIST:
+            if field_name in fields_to_exclude:
                 continue
             field = self._get_field(log.model_id, field_name)
             # not all fields have an ir.models.field entry (ie. related fields)
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -3,7 +3,7 @@
 # © 2021 Stefan Rijnhart <stefan@opener.amsterdam>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo.modules.migration import load_script
-from odoo.tests.common import SavepointCase, TransactionCase
+from odoo.tests.common import Form, SavepointCase, TransactionCase
 
 from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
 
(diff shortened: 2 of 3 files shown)


amkarthik:
Looks good to me

amkarthik (approved):

xavier-dass (approved):
Changes looks good.

atchuthan:
@OCA/oca-contributors please review

matteoopenf (approved):

yweng8111 (commented):
Nice Feature!!! Can you please add some unit test to cover your changes? Please use sudo() instead of SUPERUSER_ID

atchuthan:
> Nice Feature!!! Can you please add some unit test to cover your changes? 
Added unit test

> Please use sudo() instead of SUPERUSER_ID
Odoo also uses SUPERUSER_ID in the base code, and it was preferred in one of the mailing list discussions.

auditlog/models/rule.py
```diff
@@ -286,6 +295,11 @@ def create_full(self, vals_list, **kwargs):
                     new_values[new_record.id][fname] = field.convert_to_read(
                         new_record[fname], new_record
                     )
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Why using `with_user` instead of doing a `sudo().search(...)` ?
I'm asking because it's not used in any write or position where the user is required. If it's performed to gain rights I'll favor `sudo`

auditlog/models/rule.py
```diff
@@ -307,6 +321,11 @@ def create_fast(self, vals_list, **kwargs):
             new_values = {}
             for vals, new_record in zip(vals_list2, new_records):
                 new_values.setdefault(new_record.id, vals)
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

auditlog/models/rule.py
```diff
@@ -343,6 +362,18 @@ def read(self, fields=None, load="_classic_read", **kwargs):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

auditlog/models/rule.py
```diff
@@ -378,6 +409,18 @@ def write_full(self, vals, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

auditlog/models/rule.py
```diff
@@ -400,6 +443,18 @@ def write_fast(self, vals, **kwargs):
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

auditlog/models/rule.py
```diff
@@ -428,6 +483,11 @@ def unlink_full(self, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

auditlog/models/rule.py
```diff
@@ -442,6 +502,11 @@ def unlink_full(self, **kwargs):
         def unlink_fast(self, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            rule = rule_model.with_user(SUPERUSER_ID).search(
```
leemannd: Same question as above

atchuthan:
@yweng8111 @leemannd Changed to sudo() instead of with_user

leemannd (approved):

auditlog/models/rule.py
```diff
@@ -343,6 +358,16 @@ def read(self, fields=None, load="_classic_read", **kwargs):
...
+            rule = rule_model.sudo().search([("model_id.model", "=", self._name)])
+            if self.env.user in rule.mapped("users_to_exclude"):
+                return result
+            fields_to_exclude_match = [
+                d in rule.fields_to_exclude.mapped("name") for d in fields
+            ]
+            if (len(fields) == 1 and any(fields_to_exclude_match)) or (
+                all(fields_to_exclude_match)
+            ):
+                return result
```
coleste, line 366: What makes 1 special? `any([x]) == all([x])` anyway

auditlog/models/rule.py
```diff
@@ -142,6 +142,15 @@ class AuditlogRule(models.Model):
         "Capture Record",
         help="Select this if you want to keep track of Unlink Record",
     )
+    users_to_exclude = fields.Many2many(
```
simahawk:
```suggestion
    users_to_exclude_ids = fields.Many2many(
```

auditlog/models/rule.py
```diff
@@ -142,6 +142,15 @@ class AuditlogRule(models.Model):
...
+        "res.users", string="Users to Exclude", context={"active_test": False}
+    )
+
+    fields_to_exclude = fields.Many2many(
```
simahawk:
```suggestion
    fields_to_exclude_ids = fields.Many2many(
```

auditlog/models/rule.py
simahawk: wondering: wouldn't be better to use groups instead of users?

auditlog/models/rule.py
```diff
@@ -343,6 +358,16 @@ def read(self, fields=None, load="_classic_read", **kwargs):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            rule = rule_model.sudo().search([("model_id.model", "=", self._name)])
```
simahawk: could we benefit from mem cache here instead of querying everytime for these rules and fields?

auditlog/models/rule.py
```diff
@@ -343,6 +358,16 @@ def read(self, fields=None, load="_classic_read", **kwargs):
...
+            if self.env.user in rule.mapped("users_to_exclude"):
+                return result
+            fields_to_exclude_match = [
+                d in rule.fields_to_exclude.mapped("name") for d in fields
```
simahawk: instead of looping on all of them I would do a simple loop and break on the first that match

auditlog/models/rule.py
```diff
@@ -378,6 +403,16 @@ def write_full(self, vals, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            rule = rule_model.sudo().search([("model_id.model", "=", self._name)])
```
simahawk: same as above

auditlog/models/rule.py
simahawk: also this block could be moved to its on method

auditlog/models/rule.py
```diff
@@ -400,6 +435,16 @@ def write_fast(self, vals, **kwargs):
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
+            rule = rule_model.sudo().search([("model_id.model", "=", self._name)])
```
simahawk: same + reinforce the fact that w/ a specific method you don't repeat this code everywhere

auditlog/models/rule.py
```diff
@@ -537,8 +588,11 @@ def _get_field(self, model, field_name):
     def _create_log_line_on_read(self, log, fields_list, read_values):
         """Log field filled on a 'read' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        rule = self.search([("model_id", "=", log.model_id.id)])
+        fields_to_excluded = rule.fields_to_exclude.mapped("name")
```
simahawk:
```suggestion
        fields_to_exclude = rule.fields_to_exclude.mapped("name")
```

auditlog/tests/test_auditlog.py
```diff
@@ -368,3 +368,173 @@ def setUp(self):
     def tearDown(self):
         self.groups_rule.unlink()
         super(TestAuditlogFullCaptureRecord, self).tearDown()
+
+
+class AuditLogRuleTestCommon(TransactionCase):
```
simahawk: please use SavepointCase and move setUp code to setUpClass + set tracking_disable on the whole env

auditlog/tests/test_auditlog.py
```diff
@@ -368,3 +368,173 @@ def setUp(self):
...
+                "log_read": True,
+                "log_create": True,
+                "log_write": True,
+                "log_unlink": True,
+                "log_type": "full",
+                "capture_record": True,
+            }
+        )
+
+    def test_01_AuditlogFull(self):
```
simahawk: would be nice to split this tests in smaller tests

simahawk (changes requested):

simahawk:
@sebalix would you mind to have a look here?

bealdav:
I think it's a fantastic feature. Thanks a lot
On my case it could be better to define which fields have to be audited, but I suppose it could be managed by an additionnal m2m in other PR

clementmbr:
Nice feature!
But +1 for @simahawk requests

dreispt:
I’m curious: what is the use case to exclude some users from audit logs? Sounds shady…

odoo-sh:
@dreispt excluding Odoobot.  It will be the main user filling up the log.

auditlog/models/rule.py
atchuthan: @simahawk if we change this to group, it will change the whole concept.  So, we might as well do it in future PR.

auditlog/tests/test_auditlog.py
atchuthan: added smaller test splits

auditlog/tests/test_auditlog.py
atchuthan: Corrections done in test file

auditlog/models/rule.py
atchuthan: Done

auditlog/models/rule.py
atchuthan: Done with break in loop

auditlog/models/rule.py
atchuthan: Done

auditlog/models/rule.py
atchuthan: Corrections done for users_to_exclude_ids

auditlog/models/rule.py
atchuthan: @coleste we optimized the code and this line section is removed now

auditlog/models/rule.py
```diff
@@ -265,6 +278,7 @@ def _make_create(self):
         """Instanciate a create method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude_ids = self.mapped("users_to_exclude_ids")
```
simahawk:
```suggestion
        users_to_exclude = self.users_to_exclude_ids
```

auditlog/models/rule.py
```diff
@@ -324,6 +342,8 @@ def _make_read(self):
         """Instanciate a read method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude_ids = self.mapped("users_to_exclude_ids")
+        fields_to_exclude_ids = self.fields_to_exclude_ids.mapped("name")
```
simahawk:
```suggestion
        users_to_exclude = self.users_to_exclude_ids
        field_names_to_exclude = self.fields_to_exclude_ids.mapped("name")
```

auditlog/models/rule.py
```diff
@@ -343,6 +363,12 @@ def read(self, fields=None, load="_classic_read", **kwargs):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude_ids:
+                return result
+            for d in fields:
+                if d not in fields_to_exclude_ids:
+                    break
+                return result
```
simahawk: hmmm this will return out of read on the very first iteration. What are you trying to achieve?

auditlog/models/rule.py
```diff
@@ -360,6 +386,8 @@ def _make_write(self):
         """Instanciate a write method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude_ids = self.mapped("users_to_exclude_ids")
```
simahawk: same comments as above

auditlog/models/rule.py
```diff
@@ -343,6 +363,12 @@ def read(self, fields=None, load="_classic_read", **kwargs):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude_ids:
+                return result
+            for d in fields:
+                if d not in fields_to_exclude_ids:
```
sebalix:
```suggestion
                if d not in fields_to_exclude:
```

But one question, the field description tells "Field to exclude", as a user I expect to not have these fields logged, but we could still have a log for the other fields linked to the request, but here you avoid the log creation as soon as one of the field to exclude is present, is it really what we want?
I would instead filter out the fields to exclude in the line above:
```python
read_values = {d["id"]: d for d in result2}
```
What do you think?

auditlog/models/rule.py
```diff
@@ -378,6 +406,12 @@ def write_full(self, vals, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            if self.env.user in users_to_exclude_ids:
```
simahawk: as per my former review: why don't you move these checks to a reusable method?

auditlog/models/rule.py
sebalix:
```suggestion
        users_to_exclude = self.mapped("users_to_exclude_ids")
```
We use the suffix `_ids` for 2many field declaration, but for internal variables better to not use it or we could think it's a list of IDs.

auditlog/models/rule.py
```diff
@@ -534,11 +586,14 @@ def _get_field(self, model, field_name):
                 cache[model.model][field_name] = field_data
         return cache[model.model][field_name]
 
-    def _create_log_line_on_read(self, log, fields_list, read_values):
+    def _create_log_line_on_read(
+        self, log, fields_list, read_values, fields_to_exclude
+    ):
         """Log field filled on a 'read' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        fields_to_exclude.extend(FIELDS_BLACKLIST)
```
simahawk:
```suggestion
        _fields_to_exclude = fields_to_exclude + FIELDS_BLACKLIST
```
Redefine the var inside the method, otherwise you are modifying the argument you received and it might make debugging tricky.

auditlog/models/rule.py
```diff
@@ -565,11 +620,14 @@ def _prepare_log_line_vals_on_read(self, log, field, read_values):
             vals["old_value_text"] = old_value_text
         return vals
 
-    def _create_log_line_on_write(self, log, fields_list, old_values, new_values):
+    def _create_log_line_on_write(
+        self, log, fields_list, old_values, new_values, fields_to_exclude
+    ):
         """Log field updated on a 'write' operation."""
         log_line_model = self.env["auditlog.log.line"]
+        fields_to_exclude.extend(FIELDS_BLACKLIST)
```
simahawk: same

auditlog/models/rule.py
```diff
@@ -286,6 +300,8 @@ def create_full(self, vals_list, **kwargs):
                     new_values[new_record.id][fname] = field.convert_to_read(
                         new_record[fname], new_record
                     )
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/tests/test_auditlog.py
```diff
@@ -368,3 +368,197 @@ def setUp(self):
...
+            .with_context(tracking_disable=True)
+            .with_user(cls.user.id)
+            .create(
+                {
+                    "name": "testpartner2",
+                }
+            )
+        )
+
+    def test_01_AuditlogFull_field_exclude_create_log(self):
```
simahawk, line 466: `AuditLogRuleTestCommon` being a "common" class, should not have any test. 
Also, why not moving it to its own file?

simahawk (changes requested):

auditlog/models/rule.py
```diff
@@ -307,6 +323,8 @@ def create_fast(self, vals_list, **kwargs):
             new_values = {}
             for vals, new_record in zip(vals_list2, new_records):
                 new_values.setdefault(new_record.id, vals)
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/models/rule.py
sebalix:
```suggestion
        users_to_exclude = self.mapped("users_to_exclude_ids")
        fields_to_exclude = self.fields_to_exclude_ids.mapped("name")
```

auditlog/models/rule.py
```diff
@@ -343,6 +363,12 @@ def read(self, fields=None, load="_classic_read", **kwargs):
                 return result
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/models/rule.py
```diff
@@ -360,6 +386,8 @@ def _make_write(self):
         """Instanciate a write method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude_ids = self.mapped("users_to_exclude_ids")
+        fields_to_exclude_ids = self.fields_to_exclude_ids.mapped("name")
```
sebalix:
```suggestion
        users_to_exclude = self.mapped("users_to_exclude_ids")
        fields_to_exclude = self.fields_to_exclude_ids.mapped("name")
```

auditlog/models/rule.py
```diff
@@ -378,6 +406,12 @@ def write_full(self, vals, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            if self.env.user in users_to_exclude_ids:
+                return result
+            for d in vals.keys():
+                if d not in fields_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
                return result
            for d in vals.keys():
                if d not in fields_to_exclude:
```
Same remark regarding the fields to exclude: why not filtering out these fields from the log?

auditlog/models/rule.py
```diff
@@ -400,6 +434,12 @@ def write_fast(self, vals, **kwargs):
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/models/rule.py
```diff
@@ -400,6 +434,12 @@ def write_fast(self, vals, **kwargs):
...
+                return result
+            for d in vals.keys():
+                if d not in fields_to_exclude_ids:
```
sebalix:
```suggestion
                if d not in fields_to_exclude:
```

auditlog/models/rule.py
```diff
@@ -417,6 +457,7 @@ def _make_unlink(self):
         """Instanciate an unlink method that log its calls."""
         self.ensure_one()
         log_type = self.log_type
+        users_to_exclude_ids = self.mapped("users_to_exclude_ids")
```
sebalix:
```suggestion
        users_to_exclude = self.mapped("users_to_exclude_ids")
```

auditlog/models/rule.py
```diff
@@ -428,6 +469,8 @@ def unlink_full(self, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/models/rule.py
```diff
@@ -442,6 +485,8 @@ def unlink_full(self, **kwargs):
         def unlink_fast(self, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude_ids:
```
sebalix:
```suggestion
            if self.env.user in users_to_exclude:
```

auditlog/tests/test_auditlog.py
atchuthan: We changed the class name to AuditLogRuleTestForUserFields

atchuthan:
@simahawk @sebalix 
We made corrections once again.  Please check it.

bealdav:
OK, now all is green @simahawk thanks @atchuthan

sebalix (approved):

sebalix:
thanks @atchuthan , LGTM

PR #2292, opened 2022-02-18
[WIP][ADD] jsonify_stored

mmequignon:
Based on https://github.com/OCA/server-tools/pull/1926

Only basic features ATM, I'll improve it later

PR #2324, opened 2022-04-15
[14.0] Add jsonifier_stored

simahawk:
Replaces https://github.com/OCA/server-tools/pull/2292

TODO:

- [x] rename according to https://github.com/OCA/server-tools/pull/2327

### DIFF
--- a/jsonifier_stored/README.rst
+++ b/jsonifier_stored/README.rst
@@ -1,0 +1,1 @@
+wait 4 the bot
--- a/jsonifier_stored/__init__.py
+++ b/jsonifier_stored/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/jsonifier_stored/__manifest__.py
+++ b/jsonifier_stored/__manifest__.py
@@ -1,18 +1,18 @@
-# Copyright 2022 Camptocamp SA
+# Copyright 2022 Camptocamp SA (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# @author Matthieu Méquignon <matthieu.mequignon@camptocamp.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 {
-    "name": "Jsonify Stored",
-    "summary": "Store Jsonified data based on an ir.export",
+    "name": "JSONify Stored",
+    "summary": "Base module to store JSON data based on any model",
     "version": "14.0.1.0.0",
     "category": "Uncategorized",
     "website": "https://github.com/OCA/server-tools",
-    "author": ("ACSONE SA/NV, " "Camptocamp SA, " "Odoo Community Association (OCA)"),
-    "maintainers": ["mmequignon"],
+    "author": ("Camptocamp, " "Odoo Community Association (OCA)"),
+    "maintainers": ["simahawk", "mmequignon"],
     "license": "AGPL-3",
     "installable": True,
-    "auto_install": False,
     "depends": ["base_jsonify", "base_sparse_field", "queue_job"],
     "data": ["data/ir_cron.xml", "data/queue_job.xml"],
-    "demo": ["demo/tests.xml"],
 }
--- a/jsonifier_stored/__manifest__.py
+++ b/jsonifier_stored/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Jsonify Stored",
+    "summary": "Store Jsonified data based on an ir.export",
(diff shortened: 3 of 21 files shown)


jsonify_stored/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Camptocamp SA (http://www.camptocamp.com).
+# @author Simone Orsi <simone.orsi@camptocamp.com>
+# @author Matthieu Méquignon <matthieu.mequignon@camptocamp.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "JSONify Stored",
```
ivantodorovich: As it depends on `base_jsonify`, maybe this module could be named `base_jsonify_stored`

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,179 @@
...
+    ):
+        """Generate jobs to compute JSON data for given model.
+
+        If the curent model has a `lang_id` field, one job-ification per lang will be triggered.
+        """
+        _logger.info("cron_update_json_data_for: %s", model_name)
+        model = self.env[model_name]
+        domain = domain or []
+        if "lang_id" in model._fields and not langs:
+            langs = model._json_data_compute_get_all_langs()
```
ivantodorovich: Seems a bit unnecessary having to maintain `_json_data_compute_get_all_langs` with a raw sql that only works if the field is stored, and having to worry about cases with inherits.

Why not simply get the list of active res.lang records, instead?
Most databases have very few languages installed, so the performance imp you get with the distinct select sql query is not much.

jsonify_stored/models/jsonify_stored_mixin.py
```diff
@@ -0,0 +1,179 @@
...
+        _grouper_and_sorter = self._jsonify_data_grouper_and_sorter
+        return groupby(self.sorted(_grouper_and_sorter), key=_grouper_and_sorter)
+
+    def _get_default_lang(self):
+        return get_lang(self.env)
+
+    @staticmethod
+    def _jsonify_data_grouper_and_sorter(rec):
+        lang_code = None
+        if "lang_id" in rec._fields:
```
ivantodorovich: Maybe get the `"lang_id"` field name from a property.
I understand it's a standard but maybe some model uses a different field for the lang field

ivantodorovich (approved):
non blocking comments

jsonify_stored/__manifest__.py
simahawk: actually I'd like to change the name of base_jsonify to `jsonify` or `jsonifier` because: 
1. there's a conflict on odoo apps store
2. there's nothing "basic" here: it's a tool that you can use as it is on every model w/ out having to code anything

jsonify_stored/models/jsonify_stored_mixin.py
simahawk: I don't get your point. What shall I do w/ the list of active languages? Here I want to load only the records for a given language.

jsonify_stored/models/jsonify_stored_mixin.py
simahawk: I don't think we need this. We are defining a convention here which most of times will be satisfied as this is the most common name for relations to languages. If you want to change it, you override this. To ease it I could put the lang in its own class prop or method

jsonify_stored/models/jsonify_stored_mixin.py
ivantodorovich: I mean, `_json_data_compute_get_all_langs` could be removed and simply do this:

```suggestion
            langs = self.env["res.lang"].search([]).mapped("code")
```

In most databases you'll find not more than 3~4 languages installed, so no need to get the list of languages from the records table IMO.

jsonify_stored/models/jsonify_stored_mixin.py
ivantodorovich: Noted 👍🏻 Just to keep in mind though, there's also the "standard" `lang` Selection field (for example in `res.partner`, and some other models)

jsonify_stored/models/jsonify_stored_mixin.py
simahawk: well, yeah but not all languages might be used for a given model, that's why I'm looking only for real records' values.

jsonify_stored/models/jsonify_stored_mixin.py
simahawk: I let the implementer handle this on his/her model if wants to work w/ it :stuck_out_tongue_winking_eye:

simahawk:
@ivantodorovich module renamed, lang lookup refactored.

ivantodorovich (approved):

Issue #2134, opened 2021-07-07
[14.0] apps.odoo.com status


simahawk:
```
Module base_jsonify already exists for another serie. Please select a free module name or use the right user account.
Module onchange_helper already exists for another serie. Please select a free module name or use the right user account.
```
Indeed base_jsonify seems published only for v8 https://apps.odoo.com/apps/modules/8.0/base_jsonify same for https://apps.odoo.com/apps/modules/8.0/onchange_helper/
but seems not to be in control of OCA and that's why we cannot publish it for other versions.

@sebastienbeau @bguillot @hparfr  can you check?

hparfr:
@bealdav can you please check ?

simahawk:
Current status

Module onchange_helper already exists for another serie. Please select a free module name or use the right user account.
Module auto_backup for series 14.0 belongs to another repository. Please select a free module name or user the right user account.
Module base_jsonify already exists for another serie. Please select a free module name or use the right user account.

@hparfr @bealdav @sebastienbeau we should do something, especially because these modules might be used as base modules for other modules and this state blocks their publishing on the apps store for all them (eg: shopfloor*).

hparfr:
ping @rvalyi

simahawk:
@sebastienbeau @rvalyi piiiiiing :)

The module is already published here https://apps.odoo.com/apps/modules/8.0/base_jsonify/ and maybe you registered it manually?

In any case, I'm considering to move `base_jsonify` code to `jsonify_base` and make `base_jsonify` depend on this to preserve v13 and v14. Then in v15 we should kill `base_jsonify`.

sebastienbeau:
@rvalyi can you remove it ?

rvalyi:
> @rvalyi can you remove it ?

I answered your email, it seems I have no control over the module in the appstore sadly.

sebastienbeau:
I found the right person, it was linked to david account.
The apps have been removed

sebastienbeau:
can you try to publish the oca version ?

simahawk:
still getting the same error "Module base_jsonify already exists for another serie. Please select a free module name or use the right user account." :sob:

bealdav:
It requires 24h to be taken in account I suppose

simahawk:
nada, the error is still there :sob: 

I'll try to get in touch w/ someone at Odoo SA.

simahawk:
[...]

PR #2327, opened 2022-04-22
[14.0] Replace base_jsonify w/ jsonifier

simahawk:
Continues #2326 to address #2134.

Rationale:

1. the name reflects better what it does
   as this is not _just_ a base module

2. solves publication issue on odoo apps store
   due to an old module registered w/ the same name
   that even if un-published blocks publishing the module
   for all versions.
   This in turn, blocks the publication of ALL modules
   that depend on base_jsonify.

This change makes `base_jsonify` dependent on `jsonifier`: no harm on existing modules AFAIK.

**TODO**

- [x] test integration in existing projects


simahawk:
I've tested this on 2 existing projects: seems all good.
No change is required on depending modules, so if you update only server-tools on your instances nothing will be broken (if you install `jsonifier`).

However, I'll start replacing `base_jsonify` dependencies on `shopfloor*` and `connector_search_engine` because it's a pity that those modules are not on the apps store.

simahawk:
@sebastienbeau @lmignon @hparfr @ivantodorovich could any of  you review this?

mmequignon (approved):

lmignon (approved):

simahawk:
image

@sbidoul does it sound familiar?

sbidoul:
@pedrobaeza mentioned the same this morning. No idea what's happening...

simahawk:
@pedrobaeza @sbidoul according to https://github.com/actions/checkout/issues/363 and https://github.com/actions/checkout/issues/335 upgrading GIT might solve the problem. 
I could test it here.

simahawk:
bah, probably is not "the" solution... I'll check again later

simahawk:
I see PRs on other repo succeeding... let's try again!

/ocabot merge nobump

pedrobaeza:
The problem has escalated everywhere: https://github.com/OCA/l10n-spain/runs/6135379246?check_suite_focus=true

I don't know which part do the POT update, but may it be that an update on GH machinery puts the target directory in another environment variable or similar?

pedrobaeza:
Now it should be working:

/ocabot merge nobump

PR #2328, opened 2022-04-23
[14.0][ADD] base_disable_field_translate

kittiu:
When it is sure user don't want to use data translation. Translation of data can cause problem than being useful.
This technical module simply remove translation from all fields' data (same as set translate=False)

Selection_555

After install and restart server

image


kittiu:
Moved to https://github.com/OCA/server-ux/pull/491

PR #2336, opened 2022-05-10
[12.0][FIX] company_country: handle nonexisting env variable gracefully

hbrunn:
@moylop260 @luisg123v I'd like to make this module play nice with runboat (we can't inject env vars there, and probably shouldn't want that), does this totally destroy the point of this module for you or is this change fine?

### DIFF
--- a/company_country/models/res_config.py
+++ b/company_country/models/res_config.py
@@ -35,10 +35,12 @@ class CompanyCountryConfigSettings(models.AbstractModel):
                 ('state', '=', 'to install'),
                 ('name', '=like', 'l10n_%')], limit=1)
             if not l10n_to_install:
-                raise ValidationError(_(
+                _logger.error(_(
                     'COUNTRY environment variable with country code is not '
                     'set and no localization module is marked to be '
-                    'installed.'))
+                    'installed.'
+                ))
+                return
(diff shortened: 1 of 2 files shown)


luisg123v (changes requested):
I agree with the changes, I'd just adjust commit message. I would:
- Add ":" after module name
- Add a small description on why is this change necessary

hbrunn:
thanks, done

moylop260:
👍 
LGTM

I think the user even will not see the logs error and the instance will be displayed with l10n wrong module but if there is not another solution so go ahead

The good news is new versions has already fixed this matter

and maybe the new instances for old versions are less

so 👍 

(I even have the doubt about queue_* modules since that they will need environment variables to be tested in runboat)

luisg123v:
@hbrunn typo in description: neccessary -> necessary

hbrunn:
thanks, I keep doing this wrong since I learned English I think

@moylop260 I think the queue modules default do something safe for similar reasons, stuff like https://github.com/OCA/queue/blob/14.0/queue_job/jobrunner/runner.py#L170

luisg123v (approved):
/ocabot merge minor

hbrunn:
@luisg123v I repeated your command as you don't seem to have writing permissions in this repo

PR #2339, opened 2022-05-11
[14.0][ADD] attachment_restrict_deletion module

Kev-Roche:
This module restricts the deletion of attachments to their creator, as well as to administrative groups.

### DIFF
--- a/attachment_restrict_deletion/__manifest__.py
+++ b/attachment_restrict_deletion/__manifest__.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Invoice Attachment Owner Rule",
+    "summary": "SUMMARY",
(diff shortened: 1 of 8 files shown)


yostashiro:
Maybe the objectives are a bit different, but in case it helps: https://github.com/OCA/server-tools/pull/2164

andhit-r (approved):
#2164  provide better control. This PR is easier to use. While it is possible for both module to exist, it's better to combine both. 

What i have in mind:

* If not **restrict_delete_attachment**  then odoo original rules will be applied
* if **restrict_delete_attachment** and not **delete_attachment_group_ids** or **delete_attachment_user_ids** then attachment can only be deleted by it's creator or admin
* if **restrict_delete_attachment** and not **delete_attachment_group_ids** or **delete_attachment_user_ids** then #2164 rules will be applied

But like i said, i am not against both modules exist. So :heavy_check_mark: on functional review

Kev-Roche:
cc @andhit-r @yostashiro 
I'm agree with you, both modules are so close, it will be better to combine them. 
As a proposition to combine both models, please check my proposition here : https://github.com/OCA/server-tools/pull/2341
If it's ok with you, I ll close that one and continue on the new PR.

PR #2341, opened 2022-05-12
[14.0][ADD] module : attachment_delete_restrict

Kev-Roche:
According to the discussion : https://github.com/OCA/server-tools/pull/2339 , the PR here combine two PRs :
1) Migration of the PR https://github.com/OCA/server-tools/pull/2164
2) New 'strict' feature from PR https://github.com/OCA/server-tools/pull/2339

The purpose of the module is to restrict acces to attachement deletion at different degrees. 
For consistency between versions, I add a post_init hook to keep activated the parameters of groups and users. 

cc @yostashiro I also add you as maintainer with me, if you don't mind. 
cc @andhit-r, @legalsylvain


yostashiro:
@Kev-Roche Thanks for the new PR! IMO, however, strict/custom/none options should not be a global configuration, because that would diminish the point of combining the modules together.

It would make more sense to me if 'custom' settings of the model are first checked, and then fall back on the global setting - 'restrict' (which you currently call 'strict') or 'none' - in case there is no custom settings for the model.

The reason I'd rename 'strict' to 'restrict' is that 'custom' restriction can be stricter in a sense - 'custom' does not necessarily allow the creator of the attachments to delete their own ones.

andhit-r:
@Kev-Roche i agree with @yostashiro. strict/custom/none options should not be a global configuration.

Kev-Roche:
I understand, but I need to set a 'restrict' option on all models. It will be very long to set it on each model. 
One solution will be : 
- on global settings : remove restrict_delete_attachment field and add a default value (restrict or none) for restrict_delete_attachment in ir_model when no delete_attachment_group_ids nor delete_attachment_user_ids. 
- on res_model : restrict_delete_attachment field as compute and readonly=False, taking default value from global settings. 

What do you think ?

andhit-r:
@Kev-Roche what if:

1. Do not relate *restrict_delete_attachment* on ir.model
2. Add *use_global* selection **only on** ir.model. Make **use_global** as default value.

attachment_delete_restrict/models/ir_model.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+from .res_config import RESTRICT_DELETE_ATTACH
+
+
+class IrAttachment(models.Model):
```
andhit-r:
```suggestion
class IrModel(models.Model):
```

andhit-r (changes requested):

Kev-Roche:
Hi @andhit-r , @yostashiro, 
I rework the module in order to suit both objectives:
- restrict_delete_attachment is not more related to the global settings.
- global (all models) and model restrictions are independent
- I change "strict" into "restrict" as suggested
- I add "restrict + custom" option.

attachment_delete_restrict/models/ir_model.py
```diff
@@ -0,0 +1,58 @@
...
+
+    delete_attachment_user_ids = fields.Many2many(
+        "res.users",
+        string="Attachment Deletion Users",
+        relation="delete_attachment_user_rel",
+        help="The users selected here can delete the attachments related to this "
+        "model.",
+    )
+
+    global_delete_attachment_group_ids = fields.Many2many(
```
sebastienbeau: do we need this ?

attachment_delete_restrict/models/ir_model.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+RESTRICT_DELETE_ATTACH = [
+    ("restrict", "Restrict: Only creator and admin can delete them"),
```
sebastienbeau: I will add here a ("default", "use global configuration")

attachment_delete_restrict/models/ir_attachment.py
```diff
@@ -0,0 +1,122 @@
...
+                "Only the owner and administrators can delete it."
+            )
+        )
+
+    def unlink(self):
+        for rec in self:
+            if not rec.res_model:
+                continue
+            model = self.env["ir.model"].search([("model", "=", rec.res_model)])
+            model_restrict = model.restrict_delete_attachment
```
sebastienbeau: in case that we have a "default" on model it will simplify the code.
We use the model restriction or the global one.

sebastienbeau (changes requested):

attachment_delete_restrict/hooks.py
```diff
@@ -5,19 +5,27 @@
...
-    # delete_attachment_user_ids, change the default level of
-    # attachment deletion at 'custom' to keep it.
-    with api.Environment.manage():
-        env = api.Environment(cr, SUPERUSER_ID, {})
-        previous_group_or_user = False
-        for model in env["ir.model"].search([]):
-            if model.delete_attachment_group_ids or model.delete_attachment_user_ids:
-                model.is_restrict_delete_attachment = True
-                previous_group_or_user = True
+def pre_init_hook(cr):
```
sebastienbeau: use migration script like : https://github.com/OCA/search-engine/tree/14.0/connector_search_engine/migrations/14.0.1.1.0

attachment_delete_restrict/models/ir_model.py
Kev-Roche: no

attachment_delete_restrict/models/ir_model.py
Kev-Roche: Good idea, done

sebastienbeau (dismissed):
LGTM (code review + refactor) tested

sebastienbeau:
@yostashiro @andhit-r are you ok with the change proposed ?

yostashiro:
Hi @sebastienbeau @Kev-Roche , thanks for updating the code. My apologies for not checking the updates sooner.

The new design looks nice.

If anything, I'd prefer the style of first letting users select whether restriction should be applied or not (i.e. turn `global_restrict_delete_attachment` into a boolean field), and then let them configure following fields to define who are exempt from the restriction (and of course do the same for the model level):

- Owner (a boolean field)
- Groups (admin can be covered here)
- Users

I think this would make the design a bit more granular and make the code easier to extend.

It's not a strong preference and the current design looks to me good enough.  Thank you.

attachment_delete_restrict/models/ir_model.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2021 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+RESTRICT_DELETE_ATTACH = [
+    ("default", "Use global configuration"),
+    ("owner", "Owner: Only creator and admin can delete them"),
+    ("custom", "Custom: For each model, selected groups and users can " "delete them"),
```
yostashiro:
```suggestion
    ("custom", "Custom: For each model, selected groups and users can delete them"),
```

attachment_delete_restrict/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,25 @@
+Select level:
```
yostashiro, line 1: This content needs to be updated.

andhit-r (approved):
@Kev-Roche @sebastienbeau my apologize for the delayed responses. I already test latest revision. Looks good. Lets merge it.

attachment_delete_restrict/models/ir_model.py
Kev-Roche: done

attachment_delete_restrict/readme/CONFIGURE.rst
Kev-Roche: I updated the readme file.

Kev-Roche:
> Hi @sebastienbeau @Kev-Roche , thanks for updating the code. My apologies for not checking the updates sooner.
> 
> The new design looks nice.
> 
> If anything, I'd prefer the style of first letting users select whether restriction should be applied or not (i.e. turn `global_restrict_delete_attachment` into a boolean field), and then let them configure following fields to define who are exempt from the restriction (and of course do the same for the model level):
> 
>     * Owner (a boolean field)
> 
>     * Groups (admin can be covered here)
> 
>     * Users
> 
> 
> I think this would make the design a bit more granular and make the code easier to extend.
> 
> It's not a strong preference and the current design looks to me good enough. Thank you.

Hi @yostashiro, 
We thought about your proposition but, from our point of view,  it seems more complex and brings more line of codes. If you don 't mind, we prefer to keep it like that.

attachment_delete_restrict/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,26 @@
...
+=============
+#. Go to *Settings > General Settings> Permission section*.
+#. Choose the level of *Restrict Delete Attachment* for all models by default or for models with "default" restriction level.
+
+Five levels:
+=============
+* Default : Use global configuration
+* Owner : Owner and admins only
+* Custom : Certain groups or users per related model.
+* Custom + Strict : Owner, admins and Certain groups or users per related model.
```
yostashiro:
```suggestion
* Owner + Custom : Owner, admins and Certain groups or users per related model.
```

attachment_delete_restrict/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,26 @@
...
+* None : all users can delete them
+
+Only Custom and Custom + Strict need specific configuration on models.
```
yostashiro:
```suggestion
Only Custom and Owner + Custom need specific configuration on models.
```

yostashiro (approved):
Code review and no test. Just a couple of points to adjust in readme. LGTM 👍

attachment_delete_restrict/readme/CONFIGURE.rst
Kev-Roche: > Code review and no test. Just a couple of points to adjust in readme. LGTM +1

Both adjustements are done, thank you

PR #2342, opened 2022-05-13
[14.0][FIX] auditlog: view log multi record

Saran440:
BUG:
1. create audit rule on some module (sale.order) -> subscribe
2. create sale order 2 document
3. select SO 2 document -> Action -> View logs
4. it show 1 document only

This PR support with multi record when view logs on tree view

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -645,16 +645,26 @@ class AuditlogRule(models.Model):
             vals["new_value_text"] = new_value_text
         return vals
 
+    def _except_active_ids(self):
+        """Extend for more field exceptions."""
+        return ["res.users"]
+
     def subscribe(self):
         """Subscribe Rule for auditing changes on model and apply shortcut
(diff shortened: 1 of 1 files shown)


Saran440:
It error some view (ex. res.users) because no have active_ids
I think it should be fix in my project.

auditlog/models/rule.py
```diff
@@ -645,16 +645,26 @@ def _prepare_log_line_vals_on_create(self, log, field, new_values):
             vals["new_value_text"] = new_value_text
         return vals
 
+    def _except_active_ids(self):
+        """Extend for more field exceptions."""
```
StefanRijnhart, line 649: Can you document here what would be a good reason to refuse to show the logs for all active ids?

StefanRijnhart:
Looks like this would not fix the issue for existing rules. Would you consider rewriting the domains of existing actions in a migration script?

PR #2348, opened 2022-05-27
[WIP][14.0][ADD] Introduce new module 'cron_dst' for handling timezones and Daylight Saving Changes for cronjobs.

manengel:
Hello dear contributors,
we would like to introduce the module cron_dst to the OCA.

It has been created with the purpose to add time zones as a factor to the time of execution of cronjobs and, for regions
affected by that, also ease the handling of Daylight Saving Changes and ensuring that manual adjustments
are no longer required.

We hope that the tool can be useful for many other individuals.

Possibly, it does need improvements. It has some tests.


florian-dacosta:
Hello @manengel 
I think this module does the same more or less : https://github.com/OCA/server-tools/pull/2289
What do you think ?

manengel:
Hello @florian-dacosta,
yes, in general you are right about them being quite similar (at least in purpose). We might need to compare the both to each other more closely and then decide which PR to keep and maintain.

thomaspaulb:
Since it looks like the other one was already [merged in 16.0](https://github.com/OCA/server-tools/pull/2578), am closing this

PR #2349, opened 2022-05-28
[REF] excel_import_export: remove xlrd & xlwt from external dependencies

luisg123v:
Because those dependencies are already part of Odoo's base dependencies.

### DIFF
--- a/excel_import_export/__manifest__.py
+++ b/excel_import_export/__manifest__.py
@@ -10,7 +10,7 @@
     "website": "https://github.com/OCA/server-tools",
     "category": "Tools",
     "depends": ["mail"],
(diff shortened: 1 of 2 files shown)


luisg123v:
Hi @kittiu, @moylop260,

Could you review, please?

CC @hugho-ad

rvalyi (approved):

kittiu (approved):

moylop260 (approved):
Odoo 14.0 has a pinned version where it is compatible

but this unpinned version is doing odoo fails trying to import MS files

👍

moylop260:
Reference

https://github.com/odoo/odoo/blob/970e60d1c1426fed6e7280d3a71a7275a3dd188b/requirements.txt#L52-L53

PR #2352, opened 2022-06-02
[WIP][14.0][ADD] Introduce new modules import_processor and import_processor_async

manengel:
Hello dear contributors,
we developed two new modules which we think might be useful for others as well. Normally, you should open one PR per module but these modules are very closely related to each other.

import_processor does in general provide a framework which allows to create configurable data imports.
import_processor_async allows to enque an import for asynchronous execution.

We would be happy to receive your feedback on them.

What is yet missing?
* Codecov
* Unit tests.


Rad0van:
Hi!

At first glance this seems interesting to ma as I am building sort of mini ETL to import data into Odoo. However it would be very helpful if you provided some demo data to show the intended use or at least provide better description in README or Help. As it stands now it would take some time by trial & error to find out how to use it.

Just my $0.02.

kmatthes:
@Rad0van Thx for your feedback. We'll proof if/how we can provide some demo data.

thomaspaulb:
Nice work.

I don't know if this should be thrown on the `server-tools` pile, but it also looks weird on `queue` (not all is async), `connector` (it doesnt use the framework) or `edi` (it's not documents being imported). Perhaps there should be a new `etl` or `connectivity` repo?

I would really love if there was something like this built on top of the `connector` framework, which would make its components be dynamically creatable. For example such a module, call it `connector_dynamic`, would define `DynamicMapper(Mapper)` as a class, which would then be a fully configurable mapper with the Python pre- and post- conversions that you are adding here. There would also be `DynamicFileImporter(Importer)` with which you would be able to specify a filename for import, and specify code or templates with which to handle a file. Also you could have `DynamicRESTfulAPIimporter`, or something like that. So that a consultant can build his own importing tool.

fkantelberg:
@thomaspaulb The placement of the module is quite awkward. Nothing fits to 100% because it's quite generic and basic.

The same technic with safe_eval should be possible with the connector framework or building on top of the rest-framework but there is a downside. The interpretation of the safe_eval takes a good overhead making it less efficient to run on small and frequent jobs compared to the same code in a module. This has something to do with the prevention of arbritaty code execution and running in a safe environment. Nevertheless it's especially useful if the API/structure of the connected system changes.

thomaspaulb:
Isn't that something that could be solved by [pre-parsing](https://docs.python.org/2/library/ast.html#ast.parse) the Python expression and storing it in Odoo's ORM cache for fast retrieval by any job that would need to access it?

fkantelberg:
@thomaspaulb 
Yes indeed and it's maybe a good opportunity to look into it when I look on my side projects. I guess the biggest improvements would be possible going with `compile(...)` and `eval(...)` using a cache inside of the ORM. I'll play around and see if I can offer a nice patch to the safe_eval function of Odoo. `safe_eval` is a must have nontheless because we want the checks atleast once for the code otherwise we would offer too much potential for ACE.

ortlam:
CI looks good now. Can someone review this?

thomaspaulb:
Could you help potential functional reviewers, as well as future module users, by adding a `readme/USAGE.rst` and if necessary an `readme/INSTALL.rst` file?

ortlam:
Yes, thank you. We're also in the process of adding unit tests.

PR #2353, opened 2022-06-04
[14.0][FIX] dbs value has to be kept in case of special dbfilter config

cvinh:
I had to commit this fix to make it work on our architecture
Our dbfilter is `dbfilter = %d$`
Our proxy is Apache 2.4.6

### DIFF
--- a/dbfilter_from_header/override.py
+++ b/dbfilter_from_header/override.py
@@ -13,11 +13,13 @@ db_filter_org = http.db_filter
 
 
 def db_filter(dbs, httprequest=None):
-    dbs = db_filter_org(dbs, httprequest)
+    dbs_orig = db_filter_org(dbs, httprequest)
     httprequest = httprequest or http.request.httprequest
     db_filter_hdr = httprequest.environ.get("HTTP_X_ODOO_DBFILTER")
     if db_filter_hdr:
         dbs = [db for db in dbs if re.match(db_filter_hdr, db)]
+    else:
+        dbs = dbs_orig
     return dbs
 
 


dbfilter_from_header/override.py
```diff
@@ -13,11 +13,13 @@
...
 
 def db_filter(dbs, httprequest=None):
-    dbs = db_filter_org(dbs, httprequest)
+    dbs_orig = db_filter_org(dbs, httprequest)
     httprequest = httprequest or http.request.httprequest
     db_filter_hdr = httprequest.environ.get("HTTP_X_ODOO_DBFILTER")
     if db_filter_hdr:
         dbs = [db for db in dbs if re.match(db_filter_hdr, db)]
+    else:
+        dbs = dbs_orig
```
thomaspaulb, line 22: Could you explain what this actually changes, because I don't see it.

```
dbs = something
if condition:
    dbs = something_other
return dbs
```

Should be functionally equivalent to:

```
dbs_orig = something
if condition:
    dbs = something_other
else:
    dbs = dbs_orig
return dbs
```

cvinh: thanks for reviewing
In the previous code, as dbs was already in the arguments, it was overwritten by 
`dbs = db_filter_org(dbs, httprequest)`

cvinh:
ping @kh-badep

dbfilter_from_header/override.py
thomaspaulb: Ahhh i see it now. Yes

thomaspaulb (approved):

thomaspaulb:
@hbrunn Could you perhaps glance over quickly and approve also? It's a good fix IMO

hbrunn (approved):
if using the default filter mechanism, why have this module installed in the first place? But I also don't see how this hurts

PR #2354, opened 2022-06-06
[13.0][FIX] base_view_inheritance_extension: fix a bug that was preventing the first key of a dictionary to be replaced by a new value (because a 0 index is False)

joao-jeronimo:
I tried to always show the account.move type on top of the form view, using the following patch:
<pre>
    &lt;xpath expr="//span[hasclass('o_form_label')]/field[@name='type']" position="attributes"&gt;
        &lt;attribute name="attrs" operation="python_dict" key="invisible"&gt;[('type', '=', False)]&lt;/attribute&gt;
    &lt;/xpath&gt;
</pre>
However, the result was:
<pre>
    &lt;field name="type" attrs="{'invisible': ['|', ('type', '=', 'entry'), ('state', '=', 'draft')], 'invisible': [('type', '=', False)]}" readonly="1" nolabel="1"/&gt;
</pre>
Whereas it should be:
<pre>
    &lt;field name="type" attrs="{'invisible': [('type', '=', False)]}" readonly="1" nolabel="1"/&gt;
</pre>

This pull request fixes this. The problem was that the python code was confusing a zero with a none object.

### DIFF
--- a/base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
+++ b/base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
@@ -235,6 +235,38 @@ class TestBaseViewInheritanceExtension(SavepointCase):
             "'required': [('state', '!=', 'draft')]}",
         )
 
+    def test_python_dict_inheritance_attrs_update_dictionary_0th_key(self):
+        """Test that we can replace the first key of the dictionary (there was a bug that preventied this)"""
+        inherit_id = self.env.ref("base.view_partner_form").id
+        source = etree.fromstring(
+            """
+            <form>
+                <field
+                    name="ref"
+                    attrs="{
+                        'required': [('state', '=', False)],
+                    }"
+                />
+            </form>
+            """
+        )
+        specs = etree.fromstring(
+            """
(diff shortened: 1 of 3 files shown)


base_view_inheritance_extension/tests/test_base_view_inheritance_extension.py
```diff
@@ -235,6 +235,38 @@ def test_python_dict_inheritance_attrs_update(self):
             "'required': [('state', '!=', 'draft')]}",
         )
 
+    def test_python_dict_inheritance_attrs_update_dictionary_0th_key(self):
+        """Test that we can replace the first key of the dictionary (there was a bug that preventied this)"""
```
thomaspaulb: "prevented"

thomaspaulb (approved):
Great find and nice that you made a test

thomaspaulb:
@joao-jeronimo Could you sign the CLA? https://odoo-community.org/about/cla

MiquelRForgeFlow (approved):

thomaspaulb:
Looks like a similar thing was done in https://github.com/OCA/server-tools/pull/2557

PR #2356, opened 2022-06-08
[WIP][14.0][ADD] Introduce new module "schedule"

manengel:
Hello dear contributors,
we would like to possibly bring this new module "schedule" into the OCA and receive your feedback for this.

It might be useful as a technical module in order to develop other modules for automation. It allows for a user-friendly approach to plan events on a weekly, monthly, yearly regularity with custom intervals and has, as a dependency for future modules, potential to ease the development of modules with timed activities in case and situations cronjobs might either not be sufficient or hard to maintain.


legalsylvain (commented):
Thanks for sharing. Not sure to understand the design. Could you mention a réal application of this module ?

fkantelberg:
@legalsylvain 
One of our customers wanted a possibility to run specific things (for us it was the procurement) based on some specific times based on partners. E.g. that the procurement for a supplier X only runs on Mondays and Tuesdays but for another one it can run every day. This is why we needed a structure to define specific intervals in a more user-friendly way. Out of the box this module here does nothing but can be used as a base like described above.

thomaspaulb:
Two comments here:

1. Isn't this module more useful as eg. `base_cron_scheduler` which will inherit `ir.cron` and enhance it with more advanced scheduling features? That way it can directly be used by people, and your customer will also be helped, since pretty much anything can be run from an `ir.cron` record, also the procurements.

2. Did you take a look at [field_rrule](https://github.com/OCA/server-tools/tree/8.0/field_rrule)? It leverages rrules to define repetitivity, maybe it's nice to use as a base, or to enhance it. There's also a widget that comes with it.

fkantelberg:
@thomaspaulb Thank you for your feedback.

1. Indeed that can work as an additional module. For cron that would mean that you can run specific cron jobs in specific days which whould be quite nice and I also see 1/2 cases to use it. The thing with our customer was that there was the need to have procurement special for suppliers which is why a hook for ir.cron wasn't enough and we went for the more generic model here.

2. Indeed rrule might be a good way to handle it as a field. I wasn't aware of the rrule field module. This might help with storing the data and making it more powerful if it's managable in v14.

thomaspaulb:
@fkantelberg Could you explain how that 'procurement for suppliers' flow is implemented and why that didn't work as a cron?

fkantelberg:
@thomaspaulb 
Just an example: Procurement for the supplier A should run only on monday while the procurement of supplier B should run only once every 4 weeks. This means that every monday you see new POs for supplier A if products can be ordered from there but the POs for supplier B will only appear in a 4 week rhythm.

With the normal procurement job and especially the entrypoint you can't say Odoo to only manage specific suppliers. Odoo only checks against the stock.rule, evaluates what's missing and looks up the vendor capable of delivering. Afterwards a PO will be created/updated. If you have a vendor which only delivers every 4 weeks you have an open PO for about that time. This is why the customer wished that they can configure when POs for specific vendors are created. The entrypoint of the cron is bad and you can't create cron jobs for different suppliers without hooking into them anyways.

thomaspaulb:
Right, I get why in that case you want to put the schedule on another model than `ir.cron`. In case of a lot of suppliers, you don't want each of them having their own custom cron job. Thanks for explaining.

PR #2359, opened 2022-06-11
[WIP][12.0][ADD] Add base_external modules, use Odoo to administer data in external API

YannickB:
Hello everyone,

I'm willing to move under the OCA umbrella some modules I am working on : the ability for Odoo to connect to external, CRUD-like, API so we can use Odoo to administer the data there.

The concept here is similar to https://marmelab.com/react-admin/, a React framework you can use to connect to any API so you have an admin panel ready to use. It's essentially use by some startups who develop their own backend and need to quickly have an admin panel.

To be honest, React-Admin is really lacking on a lot of features, and the Odoo UI is far better. That's why I believe there is a great opportunity here for Odoo to take a part of this Admin UI for backend market, and theses modules are the first step for this. 
This is also an essential part of my personal strategy to move Odoo out of it's monolithic architecture and embrace the microservices world (connecting to SaaS services, other custom microservices etc...) while keeping an unique admin panel for employees.

You shall be able to mix regular Odoo models and external models in the same instance
I already made PoC for my customers using Odoo Cloud instances, Pipedrive API, Shopify API, and some others SaaS providers.

- base_external is the base module, doing the heavy lifting of overriding ORM functions
- base_external_odoo is a provider using Odoo XML-RPC. Use this module as base to create your own provider depending of the API you want to connect
- base_external_example is an example of code, creating external.sale.order and external.sale.order.line models which are a mirror and their sale.order and sale.order.line counterpart.

Please let me know if you believe theses modules shall go in another repository. This PR is for 12.0 because that's the version I am currently working on, but I intend to make the 16.0 version as soon as it's released.

Thank you,
Yannick


thomaspaulb:
@YannickB Some comments:
- It seems like this has some overlap with `base_external_system` and/or `base_external_dbsource`, could you comment on this?
- These modules reside in [server-backend](https://github.com/OCA/server-backend/tree/12.0), shouldn't this belong there then also?
- Looking back on this after now almost 2 years, are you still pursuing this idea of administering a React backend from Odoo? What are your experiences with it?

YannickB:
Hello @thomaspaulb 

1) 

I may be wrong, but from what I understand `base_external_system` and `base_external_dbsource` create a new model to store credentials of remote systems, and provide functions on this model to interact with this remote system in your custom python code. They do not really modify the Odoo framework itself.

I don't have the same goal. My goal is to 'deceive' the Odoo framework, by allowing to create a model which inherit from base_external.model. This inheritance will override some important standard methods, like _search and _read_from_database, so they are not using PostgreSQL anymore but instead find the data directly in the remote system, for each requests.

Also, I really disagree with storing remote credentials in database, for security reason it shall be in environment variables.

2) 

I suppose I wasn't sure about this one, server-tools and server-backend are quite close and I believe you know better than me here. I'd like to at least bring the critical question on the table : I use inheritance, like for instance mail.thread, to extend models capabilities. According to you, if mail.thread was from an OCA module and not from Odoo native, should it be in server-tools or server-backend ? I'd make my recommendation depending of the answer to this question.

3) 

Indeed, two years is a long time. I'm not sure it shall be merged since nobody seems to be really interested at this point.
What I can say, yes I continue to move to this direction. In general :
I expect all my future customers using Odoo to use this module, because I encourage them to also use external SaaS services (which has a lot of added value in their specialty) and use Odoo to have models specific to the company and provide a centralised interface for the employees.
I may also decide to completely move out of Odoo for new customers in the future. About this point I'm kinda at the crossroad currently, I'll see what happen.

That being said, right now it is in use in one customer, essentially between their on premise Odoo and an Expense SaaS service. I created a new expense model in Odoo, which display in the Odoo interface the expenses in the SaaS service. This module is currently used in production in v16, and I expect to extend this approach to other SaaS services used by this big customer.

So here we are. I think that without further interest from members of the OCA, we shall probably close this PR. Theses modules represent my vision of entreprise architecture, in this world of microservices and SaaS platforms, but maybe it's premature to involve the OCA until I can do some proper communication about this vision.

thomaspaulb:
@YannickB Thanks for the explanation! I think you're right in your assessments, if this makes its way into OCA it will need to be in another form and maybe on server-backend repo 16.0

Perhaps at that point you could start by proposing them in smaller pieces, for example an addition to base_external_system where passwords are stored outside of the database, makes sense. (By the way I think the documentation of that module already describes how to do that, using some kind of vault module, but I digress)

Meanwhile, OCA days always seem far away but in reality they arent - perhaps you can come and do a talk about this vision. It would be interesting.

PR #2364, opened 2022-06-22
[ADD] base_time_dependent_parameter

nimarosa:
Hello, i bring here a new module which in collaboration with @appstogrow we think it might be useful for the server-tools repository.

Basically the module allows the user to define parameters with a time_dependent value. That is, each parameter has "revisions" and a start_date, so you can fetch this parameters from any module and get the actual value for the today's date or you can pass a "date" param, to fetch the value from a given date.

Please let me know if i need to change anything to comply to the OCA repo structure. It's my first time making a PR for a totally new module so i don't know quite well how to make the readme.rst scripts and other guidelines work.

Hope we can get it merged. Regards.


nimarosa:
Please @appstogrow and @pedrobaeza if you could check this module to get it merged so we can finish the PR in payroll #31 with the implementation of this.

pedrobaeza:
Pre-commit seems to fail. I also don't understand the use case of this to be a separate module and increase the overhead of this.

nimarosa:
Hello @pedrobaeza , if you see the PR https://github.com/OCA/payroll/pull/31 in payroll you will see that @appstogrow suggested that I make a new module with the functionality because he thinks it might be useful for another modules to, so we agreed to make a generic model and then using that in payroll. 

For me we can merge the PR in payroll but that was the reason of this PR. 

If you think that it don't have to be a different module I ask you to please continue in merging the payroll PR since is a crucial functionality for my localization.

pedrobaeza:
On both places, being as a separate module or integrated, I'm not able to understand why a new model is needed instead of being extra fields on the salary rule defining the time dependent thing.

nimarosa:
@pedrobaeza Because this are parameters that should be used within a salary rule code (submitted by the payroll user) like the worked_hours and inputs are right now. 

The use case for this in payroll is that during the salary rule computation you can use this parameters in the SR formula and the value returned will be set based on the date of the payslip. So if you are doing payslips for last month for example you will get x value for that date and if you use it in todays Month you will get y value according the revisions date and values that you set. 

The use case in payroll is that in some countries, you have calculation values that changes depending of the fiscal period you are making a payslip for. So the PR allows the user to set this parameters with a starting date and then it doesn't need to change the value manually according of the date of the payslip, it always will return the value for the right month according to the date. 

In my localization for example we use that to supply different minimum salary values that the labor union set for the year. So we use that parameters in salary rules to ship the localization with this values in data files by default so the user don't have to do it manually and be constantly checking labor union changes. 

I hope I made it clear to you, if not please ask i Can try to clarify it more. Then @appstogrow has a use case in the account.tax module for this functionality and maybe he can explain his implementation too. That's why we prepared this module but the main and production use for me will be in payroll.

pedrobaeza:
OK, I think I'm understanding, but shouldn't that be formula embedded in the salary rule computation instead of having a new model and to assign codes that you later have to define in such formula? I think you should do something similar to Excel/Calc, where you have functions that do several computations, and this kind is one of them.

nimarosa:
@pedrobaeza That's the way it work in the Payroll PR if you want to check. In there I did changes to payslip calculation that allows to use a formula object calling payslip.rule_parameter("CODE"). The same way that you consume worked days (worked_days.CODE.number_of_hours). 

In this PR in particular I only ship the basic module and then if the user want to integrate it to another module he will have to code the implementation according to the need of the parameters. The plan is to consume this module and adapt the payroll implementation to like you said, have this parameters accessed in the formula like any other payroll parameter. 

Also I'm thinking that if we drop the "base" module and do the payroll implementation like I've done initially we can also refactor it to use the new payslip refactoring that it's being discussed in another payroll PR.

pedrobaeza:
That's not exactly. You still need the new model `hr.rule.parameter` and `hr.rule.parameter.value` and use `CODE` to "invoke it". I'm talking about embedding the needed formula in the payroll formula field. Example: `SUM(x)`.

nimarosa:
@pedrobaeza i think I don't quite understand so. If we have no model how can we set the parameters and it's dates that the function will use to decide which value to return? 

Maybe I don't understand you approach or maybe I don't explained my use case in detail. 

Can you elaborate? I don't see a way for this to be a embedded formula, because it use parameters that change over time and the payroll manager should be able to construct salary rules with them.

nimarosa:
I think I understand that you are talking about the other modes in the salary rule that could be ranges, fixed amounts, percentages and so? 

If that's the case I think it will not work because this values often are used in complex formula. Like the example I described I use this to calculate the minimum salary settled by the labor union for that month and that way I can compare this with the actual payslip salary, and if the param is greater than the actual salary I have to supplement the difference in the salary rule.

pedrobaeza:
Let's do one thing: put a real example of the data you configure for such models and the salary rule, and I will tell you my counter-example with my approach for the same use case.

nimarosa:
@pedrobaeza okay great. I have two examples, one basic and one more complex that require more code un salary rule. I will start with the basic one to see if it's more clear. 

**Example 1: Rates**
In some countries the rates for the tax that you deduct from employee payslip changes on time to time. So supone you have a 10% tax from January to June and a 20% tax from July to December. You set the models like this: 

Salary rule parameter: 
- Code -> TAX_RATE 
Salary rule values (revisions):
- From date: 01/01/2022 - Value: 10.0 
- From date: 01/07/2022 - Value: 20.0 

So we set two revisions on the salary rule parameter TAX_RATE. 

Then in the salary rule we can use this code: 

result_rate = payslip.rule_parameter("TAX_RATE") 
result = categories.GROSS 

Using that, the user can set the salary rule only one time, and the rate amount will change according to the payslip date to the correct value (10.0 or 20.0) and then calculate that percentage for the GROSS category of the payslip. 

So without this module the user don't have a easy way to set a global parameter that changes depending on time, and has to: manually change the salary rule every month the rate changes or set the value in a fixed param (like configuration ones) and manually change it when it's time. This could lead to errors and forgets.

I hope I made it clear. If not I have another example that it's more complex but I think this shows the functionality well.

nimarosa:
@pedrobaeza Did you have oportunity to see this?

ghost:
I have similar use cases as @nimarosa has described above, and I think this is a good base module.

__manifest__.py should have maintainers in one line instead of two lines.

Regarding the failing pre-commit:
1) ./README.rst will be created from readme folder right?
2) Is "LGPL-3" not a valid value for license?
3) The third error is related to excel_import_export which is in this repo but is not a dependency of this module.

nimarosa:
@appstogrow Hello Henrik, thanks for replying the PR. I hope we can get this merged here or at least in the payroll module so i can continue with new ones. 

Yes i saw the errors. 

1. For the README i thought it will be generated on merge, but i was thinking in creating a empty README.rst that maybe then will be replaced? 
2. It's weird it's not a valid licence because I have seen it in another modules. Maybe not in this repo. If someone can point a valid one i can change it. 
3. Don't know why we are receiving that error here, it's not related to the module at all. 

Also about the manifest i see the maintainers in one line. Did you saw it differently? 

@pedrobaeza Can you point us in a direction about it? Did you read the example I made? 

Wait for your comments and thanks in advance.

pedrobaeza:
The error on pre-commit is:

`base_time_dependent_parameter/__manifest__.py:7:162: B950 line too long (161 > 88 characters)`

Just fix that.

I have checked the use case and now I understand the usefulness. It's OK to continue with the module. I would remove the model field though, as it's not adding anything useful. Another problem is to add this dependency in the payroll module in the middle of the lifecycle of the version. I still would prefer to include this feature in the module itself.

nimarosa:
Hello @pedrobaeza if you like we can finish this module here for anyone who wants to use it like @appstogrow and then finish and merge the payroll commit with this functionality.

Also, the model field is there because this module is intended to be integrated in another modules so this field allows to create views filtered by that field in the other modules and give the user the ability to only see parameters of the module he is using. 

Will do this and check the PR in payroll repo and let you know. The payroll PR is ready but I will check it don't affect the recent changes we made there. 

Thanks. Will be back with news.

ghost:
I agree we can remove the `module` field.

With this module installed, a hr.salary.rule may have python code like
`value = payslip.get_time_dependent_parameter('my_parameter_name', payslip.date_to)`
also when the payroll module does NOT depend on base_time_dependent_value.

Please change the name of this PR: dependent instead of dependant.

nimarosa:
The changes are ready in this module. @appstogrow Note that if we use this module in payroll it must depend on it, because the usage I do i need to add views for the user to edit the time_parameters into the GUI. I mean, a payroll manager should have easy access to this. That's why this module to really be useful in payroll has to depend on it. 

That's why @pedrobaeza point us to merge this here, for usage in another modules and also merge the payroll salary_rule_parameters PR to not have to add the dependency in the middle of the lifecycle of the version.

ghost (approved):

nimarosa:
@pedrobaeza This can be also merged or is something missing?

pedrobaeza:
Can you check pre commit that is failing?

nimarosa:
@pedrobaeza This are the errors that i see in pre-commit: 

************* Module base_time_dependent_parameter
base_time_dependent_parameter/__init__.py:1: [C7902(missing-readme), ] Missing ./README.rst file. Template here: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

**- Should i add a empty README.rst file and that will be re-generated on merge?**

************* Module module_prototyper.models.licenses
module_prototyper/models/licenses.py:64: [W0622(redefined-builtin), get_license_text] Redefining built-in 'license'

**Not sure about licenses needed. Can you suggest a licence that the script will accept? I always use that license but maybe in this repo is not allowed?**

************* Module excel_import_export.models.xlsx_import
excel_import_export/models/xlsx_import.py:10: [W7936(missing-manifest-dependency), ] Missing external dependency "xlrd" from manifest. More info: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#external-dependencies

**This error and module is not related in any way with 'excel_import_export' module. What should i do?**

Please point me in one direction about this errors because i'm not sure what to do to fix them.

pedrobaeza:
You should look to the red line:

https://github.com/OCA/server-tools/runs/7187294541?check_suite_focus=true#step:7:101

The other are just warnings.

ghost:
@nimarosa Will you fix this?
base_time_dependent_parameter/__manifest__.py:7:108: B950 line too long (107 > 88 characters)

nimarosa:
@appstogrow @pedrobaeza Okay i fixed line length. Will see if it pass now when the tests finish. 

@appstogrow Sorry the delay, i was in a busy week.

PR #2367, opened 2022-07-04
[REF] base_import_module_group: Remove not working auto-install

SirTakobi:


thomaspaulb (approved):

base_import_module_group/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "auto-install": True,
+    "auto_install": True,
```
thomaspaulb: Actually, no. Why should this module be `auto_install` anyway? Let's just remove the whole key instead.

thomaspaulb (changes requested):

base_import_module_group/__manifest__.py
SirTakobi: Seems reasonable: no-one will be missing a feature that doesn't work

thomaspaulb (approved):

SirTakobi:
@OCA/tools-maintainers can this be merged? Thanks!

dreispt (approved):

PR #2373, opened 2022-07-15
[15.0][FIX] base_exception: add _popup_exceptions deleted in migration

MiquelRForgeFlow:
This method is used by other modules that depend on 'base_exception'.

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -265,6 +265,27 @@ class BaseExceptionModel(models.AbstractModel):
             else:
                 rec.exceptions_summary = False
 
+    def _popup_exceptions(self):
+        """This method is used to show the popup action view.
+        Used in several dependent modules."""
+        record = self._get_popup_action()
+        action = record.sudo().read()[0]
+        action = {
+            field: value
+            for field, value in action.items()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I already commented it in the migration, but this method is only used in `purchase_exception` AFAIK, so it should be located there instead.

MiquelRForgeFlow:
`account_move_exception` also uses it.

base_exception/models/base_exception.py
```diff
@@ -265,6 +265,26 @@ def _compute_exceptions_summary(self):
             else:
                 rec.exceptions_summary = False
 
+    def _popup_exceptions(self):
+        """Used in several dependent modules"""
```
pedrobaeza: Explain also what this method does

pedrobaeza (approved):

base_exception/models/base_exception.py
MiquelRForgeFlow: Well, I added a bit of explanation 😅

The `account_move_exception` uses this method this way:
Selection_485

nicomacr (approved):
Also used in `sale_exception` on https://github.com/OCA/sale-workflow

PR #2375, opened 2022-07-18
[14][FIX] Keep computed editable field value when playing onchanges

florian-dacosta:
When playing the `play_onchange` on a dictionary, `onchange_helper` remove the value of the stored computed fields with no inverse method, but it should also check if the field is readonly. If the field is not readonly, it does not need an inverse method and it should be kept.
This is a issue that will become more frequent as onchanges are replaced by computed not readonly fields.

Example with `account_sale_payment` : 
```
vals = {"payment_mode_id": 1}
vals = self.env["sale.order"].play_onchanges(vals, ["payment_mode_id"])
```

Then the key `payment_mode_id` is not present in the dict `vals`

### DIFF
--- a/onchange_helper/models/base.py
+++ b/onchange_helper/models/base.py
@@ -64,6 +64,10 @@ class Base(models.AbstractModel):
         return {
             f: v
             for f, v in all_values.items()
-            if not (self._fields[f].compute and not self._fields[f].inverse)
+            if not (
+                self._fields[f].compute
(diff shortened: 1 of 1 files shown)


florian-dacosta:
@sebastienbeau @rousseldenis @lmignon

lmignon (approved):

rousseldenis (approved):
Indeed

PR #2376, opened 2022-07-18
[MIG] sentry: Migrated to 15.0

airlessproject:


aisopuro:
It looks like code coverage is failing due to some lines in the configuration reading not being covered by tests.

I don't know the policy regarding these kinds of things: are some of these lines OK to skip? I wouldn't know how to set up tests that would cover them.

@extrememicro you've contributed here recently: do you know where the author might find guidelines regarding how to do coverage for lines that are in `__init__.py` files?

airlessproject:
@traviswaelbro please check previous comment by @aisopuro, maybe you can help as well, considering you migrated the module to 14.0 and made other improvements?

traviswaelbro:
I've not worked with code coverage at all, but you may be able to mock out import errors. You can also mock out return exceptions to trigger the `except` clause.

I'm not sure how you can tweak the Odoo configuration for some of the other complaints from coverage.

aisopuro:
Introducing mocking just to avoid coverage errors seems a little overkill IMO. In my estimation this is confuration-reading code that is difficult to test with unit tests like this. I would be fine with putting in `# pragma: no cover` comments in order to get this PR moving.

traviswaelbro:
I'd agree with that

aisopuro:
@traviswaelbro I made a new PR with fixes to tests that resolved the coverage failures: #2386. I've kept all of @airlessproject commits so they're credited properly.

thomaspaulb:
Replaced by #2386

PR #2254, opened 2022-01-04
[14.0][MIG] sentry: migrate sentry-raven to new api sentry-sdk

paradoxxxzero:
This is a migration done by cherry-picking https://github.com/OCA/server-tools/pull/1942 on top of current 14.0.
This superseeds #2223

PR #2386, opened 2022-08-10
Migrate Sentry to 15.0

aisopuro:
This adds fixes to the migration started in #2376. Most notably fixes the tests to actually run.


thomaspaulb (approved):

RaveendarU:
@aisopuro aisopuro Any specific reason for using old 'raven' package for Sentry(version:15) instead of 'sentry-sdk' package which we used in previous versions.

aisopuro:
~~Unfortunately @RaveendarU just lack of time. I don't have the bandwidth to refactor/rebuild the module to use the new SDK, and we're succesfully running this code as is.~~

~~I agree it would be better to move to the new SDK since Sentry themselves recommend it, but unfortunately I won't have time to implement that in the foreseeable future. If someone knows that it is super-easy to do and has a fast test iteration setup ready, I'm more than willing to let this PR be stolen from me~~ :smile: 

OK, now I actually read your comment and see that you are referring to the fact that the 14 version already uses the new SDK. :facepalm: 

No, there's no reason for that other than oversight: we either did not notice that code when we started migrating this, or it wasn't available. We should most likely simply re-do the migration basing it off the 14 version. Though again, I don't have a lot of bandwidth. I'll see what I can come up with.

thomaspaulb:
I'm thinking it could be as simple as rebasing this PR's branch on the latest 14.0, to which the `sentry_sdk` refactor was merged a couple of days ago: https://github.com/OCA/server-tools/pull/2254

aisopuro:
Alright @thomaspaulb I've opened a third attempt at migration: #2428 . See the notes on that PR for explanation. I'll close this one as unnecessary.

thomaspaulb:
Thanks for the work! You could have force-pushed it over here as well, but it doesnt matter, I'll review the other one.

PR #2392, opened 2022-08-29
sql_export: improve menu position

sebastienbeau:
Change the parent to avoid to have in first position the SQL Export.
SQL export is less important then native dashboard so it should be after

Before when opening dashboard menu

image

After When opening dashboard menu

image


sebastienbeau:
@Kev-Roche @florian-dacosta

rvalyi (approved):

Kev-Roche (approved):
LGTM

legalsylvain (approved):

PR #2396, opened 2022-09-01
[FIX][11.0] Make branch green

thomaspaulb:
All new 11.0 PR's currently turn red because of these two problems, [example here](https://github.com/OCA/server-tools/pull/2014):

**Problem 1** `letsencrypt` depends on `dns` Python package but the real name is `dnspython`, which confuses Travis and Runboat. See discussion in https://github.com/OCA/server-tools/pull/2014.

**Problem 2** `COUNTRY` variable does not exist, and `company_country` module does not handle that gracefully. Backport of https://github.com/OCA/server-tools/pull/2336

### DIFF
--- a/company_country/models/res_config.py
+++ b/company_country/models/res_config.py
@@ -35,10 +35,12 @@ class CompanyCountryConfigSettings(models.AbstractModel):
                 ('state', '=', 'to install'),
                 ('name', '=like', 'l10n_%')], limit=1)
             if not l10n_to_install:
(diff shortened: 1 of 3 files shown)


thomaspaulb:
@luisg123v @hbrunn @sbidoul Gently asking for a review for this maintenance PR

luisg123v (approved):

rvalyi (approved):

PR #2398, opened 2022-09-08
[12.0][IMP]sentry: Allows to set sentry configuration as environment variables.

jesusVMayor:
- Allow to setup sentry options as environment variables, allowing to have various environments with the same configuration file.
- Define the event_level with the configuration parameter instead of fix it on warning.


thomaspaulb:
@jesusVMayor Since something else was merged into `sentry` meanwhile, this needs a rebase.

PR #2408, opened 2022-09-28
[FIX] attachment_unindex_content: unsupported param on _index method

luisg123v:
Odoo introduced a new param `checksum` to the method `_index` on the attachment model [1]. Since that param is not supported on this model, the following error is now being raised:

    TypeError: _index() got an unexpected keyword argument 'checksum'

[1] https://github.com/odoo/odoo/commit/3d59cc84


moylop260 (approved):
👍

PR #2413, opened 2022-10-04
[MIG] base_m2m_custom_field: Migration to 16.0

rousseldenis:


rvalyi (approved):
no test but LGTM. FYI here is the original Many2Many#update_db method that is bypassed when 
`create_table=False` https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L4637 
(it's exactly the same code as in v14 or in v15)

lmignon (approved):

PR #2416, opened 2022-10-05
[15.0][FIX] upgrade_analysis: misleading representation of read/write computed fields

StefanRijnhart:
Odoo 14 introduced the widescale usage of computed fields with readonly=False. In that case, the compute method functions as a _record specfic_ default method that can also be used to compute a value some time *after* the initial creation of the record.

In the OpenUpgrade analysis files, these fields would be misrepresented as computed fields rather than fields with a default function. This change fixes that. See https://github.com/OCA/OpenUpgrade/pull/3587 for the result.

image

### DIFF
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -134,8 +134,6 @@ def report_generic(new, old, attrs, reprs):
         if attr == "required":
             if old[attr] != new["required"] and new["required"]:
                 text = "now required"
-                if new["req_default"]:
-                    text += ", req_default: %s" % new["req_default"]
                 fieldprint(old, new, "", text, reprs)
         elif attr == "stored":
             if old[attr] != new[attr]:
@@ -284,14 +282,18 @@ def compare_sets(old_records, new_records):
         ],
     )
 
-    printkeys = [
+    # Info that is displayed for deleted fields
+    printkeys_old = [
         "relation",
         "required",
         "selection_keys",
-        "req_default",
         "_inherits",
         "mode",
         "attachment",
+    ]
+    # Info that is displayed for new fields
+    printkeys_new = printkeys_old + [
+        "hasdefault",
     ]
     for column in old_records:
         if column["field"] == "_order":
@@ -304,7 +306,7 @@ def compare_sets(old_records, new_records):
         extra_message = ", ".join(
             [
                 k + ": " + str(column[k]) if k != str(column[k]) else k
-                for k in printkeys
+                for k in printkeys_old
                 if column[k]
             ]
         )
@@ -312,11 +314,6 @@ def compare_sets(old_records, new_records):
             extra_message = " " + extra_message
         fieldprint(column, "", "", "DEL" + extra_message, reprs)
 
-    printkeys.extend(
-        [
-            "hasdefault",
-        ]
-    )
     for column in new_records:
         if column["field"] == "_order":
             continue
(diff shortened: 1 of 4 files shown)


pedrobaeza:
For new fields,I think it's interesting to know that they are compute / stored (like a default). For existing fields, not sure if being put if any change of a field may trigger a recomputation of this field.

StefanRijnhart:
@pedrobaeza that implies you'd want to see the `depends` in the OpenUpgrade analysis. Are there examples where that is actually relevant when developing migration scripts?

pedrobaeza:
Not really. It's like the default. When a new field is present, and it has a default, we put in the analysis file that it has default. You go to the definition to see which default is and if it serves. The same can apply to the computed writable: it can put: has compute, and if you want to see which value it will take (or want to disable as very heavy), you see the code and decide, but not having this information will make an immediate "Nothing to do", where it's not.

StefanRijnhart:
@pedrobaeza not sure I understand, but let's recap.

Before this PR, it is reported for a ~~new,~~ required field if it has a `default` attribute and whether that is a literal or a function. This is key `req_default` in the analysis.

It is also reported if a field that did not have a default has a default in the new version or the other way around (as per https://github.com/OCA/OpenUpgrade/commit/4e60a82b6030c5b214f44547bc76f7e6c04f7bef, discussion in https://github.com/OCA/OpenUpgrade/issues/1204 ). This is key `default` in the analysis.

What this PR changes, is that for a ~~new,~~ required field that has a compute method, it will show the name of that compute method in the analysis.

I think the discussion in #1204 was not very clear. ~~I don't think it's very interesting if anything changes wrt. the default or compute method of an existing read/write field.~~

Instead, I do think it is interesting to see for ~~new,~~ non-required field the same information about any kind of default. So my proposal would be extend this PR to
* reverse https://github.com/OCA/OpenUpgrade/commit/4e60a82b6030c5b214f44547bc76f7e6c04f7bef (which would rename `req_default` back to `default`)
* show the default info also for ~~new~~ fields that are not required. That was my original intention with https://github.com/OCA/OpenUpgrade/issues/1204.

Is this in line with what you had in mind?

-- edit -- actually, the default for required fields was already reported for existing and new fields, so we'll keep it that way.
-- edit 2 -- no it wasn't and we'll keep it that way.

StefanRijnhart:
@pedrobaeza my refactoring is done. The 'default' key is now merged into the 'hasdefault' key and it can have two values: `default` or` compute`. That is to say that for new fields, it is shown if there is a default method or a compute method. Note that static values (e.g. `default=True`) are wrapped in a lambda function by the Odoo framework and cannot be distinguished from default methods anymore.

The information is not shown for deleted fields anymore as I think this is useless noise, and as for changes with existing fields in the previous version, this was not shown and it is still not shown as I still think this information is only useful for new fields.

I've already included this change in the initialization of 16.0 (https://github.com/OCA/OpenUpgrade/pull/3583), but you can inspect the change in the analysis more easily in the 15.0 analysis update here: https://github.com/OCA/OpenUpgrade/pull/3587.

Let me know if we are on the same line.

MiquelRForgeFlow (approved):
Is it retro-compatible with previous versions, right?

pedrobaeza:
Checked the result on OU:

/ocabot merge major

pedrobaeza:
It seems there's a problem with `letsencrypt` module due to a new LE policy:

```
 acme.messages.Error: urn:ietf:params:acme:error:rejectedIdentifier :: The server will not issue certificates for the identifier :: Error creating new order :: Cannot issue for "*.example.com": The ACME server refuses to issue a certificate for this domain name, because it is forbidden by policy
```

@NL66278 you are the migrator to this version. Can you please check? Maybe we should disable tests for now?

NL66278:
@pedrobaeza I did a mostly technical migration, not being really an expert on the intricacies of letsencrypt. As far as I can see it is the acme client complaining about a generic domain that the certificates in the request are requested for. I have some hesitancy just disabling the tests... Although that would maybe be the most easy option.

pedrobaeza:
Let's try again to see if it's a transient question:

/ocabot merge major

MiquelRForgeFlow:
Maybe the acme library version has to be restricted 🤔

StefanRijnhart:
@MiquelRForgeFlow it is not entirely backwards compatible, in the sense that we need one more backport: https://github.com/OCA/server-tools/pull/2431

StefanRijnhart:
This fixes the letsencrypt tests, for now: https://github.com/OCA/server-tools/pull/2432

PR #2422, opened 2022-10-11
[FIX] base_time_dependent_parameter

norlinhenrik:
I have fixed **datetime** and **ast.literal_eval** and now the module should work properly.

### DIFF
--- a/base_time_dependent_parameter/__manifest__.py
+++ b/base_time_dependent_parameter/__manifest__.py
@@ -5,9 +5,9 @@
     "name": "Time Dependent Parameter",
     "version": "14.0.1.0.0",
     "summary": """
-    Time dependent parameters
-    Adds the feature to define parameters
-    with time based versions.
+        Time dependent parameters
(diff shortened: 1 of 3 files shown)


nimarosa (approved):

nimarosa:
@norlinhenrik Henrik, can you make another push to retrigger the tests? They seems successful in first tests but then fails in the second test. The failing is related to letsencrypt module, so i think i might be a temporary error. If you make a force-push we can have the tests running again since i don't have access to this repo actions to retrigger them manually.

norlinhenrik:
As you say, "The failing is related to **letsencrypt** module". There are also an error related to **attachment_synchronize**. I think the errors will not disappear until these modules are fixed.

There are no errors related to **base_time_dependent_parameter**. Anyway I have force-push to run the tests again.

nimarosa:
@norlinhenrik now tests are okay. I think the bot losses connection and that's why failed. 

We can merge now.

/ocabot merge patch

PR #2423, opened 2022-10-13
[ADD] company_today

carmenbianca:
Purpose of the module is described well in DESCRIPTION.rst. It originally lives in https://github.com/coopiteasy/addons. Example usage can be found [here](https://github.com/coopiteasy/cie-timesheet/blob/16f5aabc9f198bece6d943060b2ac5f862d2d39c/hr_timesheet_overtime/models/hr_timesheet_sheet.py#L138).

In preparing this PR, I ran into one issue: The value of `fields.Date.today()`/`date.today()` (effectively equivalent) is determined by the local time on the server. For a server where the local time is configured as UTC and a client in Brussels (CEST, UTC +2:00), this means that the value of `.today()` won't change until 02:00 local time for the client. This gets worse the further the client is from Greenwich. This problem is bigger than this module, but this module is especially affected by it. I don't know what the conventions are to address this problem.


carmenbianca:
Ping @thomaspaulb

company_today/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,24 @@
+Store today's date on the company model. A cronjob makes sure this field stays
+up-to-date.
+
+The use-case for this module is as follows. Imagine you have two regular fields
+and a computed field::
+
+    amount = fields.Monetary()
+    payment_date = fields.Date()
+    cost_per_day = fields.Monetary(compute="_compute_cost_per_day")
```
thomaspaulb, line 9: A question I had here was: in your real-world examples where you use this module, are these non-stored fields like in this example, or stored fields? Because I can imagine the use case for stored fields, but for non-stored the value is generally recomputed on the fly so it would hold to just use `fields.Date.today()`.

company_today/data/ir_cron.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field name="model_id" ref="model_res_company" />
+        <field name="code">model.cron_update_today</field>
+        <field name="active" eval="True" />
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field
+            name="nextcall"
+            eval="(datetime.now() + timedelta(days=1))
+            .replace(hour=0, minute=0, second=0)"
+        />
```
thomaspaulb, line 19: So about the timezone issue, if you look at [how Odoo handles this field](https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/ir_cron.py#L132) you'll see it interpretes the stored `nextcall` value as UTC, then converts it to the timezone of the user that is running the cron, which in this case would be Administrator. So in your calculation above, you're now placing the midnight datetime in UTC, but you should actually be placing the midnight datetime in the timezone of the `user_id` defined on the cron record.

This now limits the usefulness of the `company_today` field, since, assuming your Odoo installation spans multiple timezones, it won't be "today" for everyone - this is specifically a field that is updated at midnight every day in the timezone of the Administrator user, and shifts the value by 1. That's probably good enough for most of your installations where you use this, but it doesn't make it universally applicable - so at least that has to mentioned in the documentation/disclaimer of the module. I usually place it in "ROADMAP" - something like "Works when you have a single timezone instance, but needs to be adapted for use in mul ttiple timezones"

Thinking about an implementation that would work in all timezones, I could imagine the `today` field moving to `res.users` instead, and having a cron that runs at least every 30 minutes, and then updates the value for each of the users. But that would of course make the module harder to use, because you'll have to find an `api.depends` route to the current user.

thomaspaulb:
So I guess for the case where you have stored computed fields that are dependent on the current date, in a single-timezone instance, this module can work well.

My own strategy for dealing with date-dependent computed fields is that I always make them non-stored, and compute them on the fly, and then you can also just use the timezone of the user that is doing the action - `self.env.user`.

In case there is a necessity for them to be stored because the recalculation is really slow, then I try to make a stored computed field "up until the point" that the date is needed. Eg in your example I would have the total cost as a stored computed field, and then have a non-stored computed field that does the date-dependent part of the calculation based on that.

In case there is a necessity to have a stored field because it needs to be sortable in the view, I would resort to an SQL view.

company_today/readme/DESCRIPTION.rst
carmenbianca: Ah, yes, this concerns stored fields. I believe the use-case is to be able to sort on those computed-and-stored fields.

carmenbianca:
> In case there is a necessity to have a stored field because it needs to be sortable in the view, I would resort to an SQL view.

I didn't know this was possible! I think that this would negate the use case for this module. Is there a piece of documentation where I can learn more?

thomaspaulb:
> > In case there is a necessity to have a stored field because it needs to be sortable in the view, I would resort to an SQL view.
> 
> I didn't know this was possible! I think that this would negate the use case for this module. Is there a piece of documentation where I can learn more?

I don't know of any, but you can find examples of this pattern all over the odoo code. Just a random example is this stock forecasting report:

https://github.com/OCA/OCB/blob/11.0/addons/stock/report/report_stock_forecast.py

Anywhere in the odoo source code where you see a model with `_auto = False` and then an init function with `CREATE VIEW` in it, follows this concept.

Note that you'll still need to account for timezones, eg you can calculate your values in postgres using UTC, but if you need an output in the user's timezone that's still sortable, this may be tricky.

thomaspaulb:
@carmenbianca Do you still see a use case for this module, or not anymore?

PR #2425, opened 2022-10-13
[16.0][MIG] dbfilter_from_header

carmenbianca:

PR #2429, opened 2022-10-17
[16.0][MIG] dbfilter_from_header

nilshamerlinck:


legalsylvain:
Hi.

It seems to be a duplicate of https://github.com/OCA/server-tools/pull/2425

CC : @carmenbianca 

Edit : sorry, did'nt saw your comment in the other PR.

dbfilter_from_header/override.py
```diff
@@ -6,24 +6,99 @@
...
 
+# we don't rely on original http.db_list
+# because it calls db_filter(dbs, host)
+# while we also need to pass httprequest
+def db_list(force=False, host=None, httprequest=None):
+    try:
+        dbs = odoo.service.db.list_dbs(force)
+    except psycopg2.OperationalError:
+        return []
+    return db_filter(dbs, host, httprequest)
```
carmenbianca: I propose to use keyword arguments here, so

```suggestion
    return db_filter(dbs, host=host, httprequest=httprequest)
```

dbfilter_from_header/override.py
```diff
@@ -6,24 +6,99 @@
...
+            session.context["lang"] = self.default_lang()
+
+        dbname = None
+        host = self.httprequest.environ["HTTP_HOST"]
+        if session.db and db_filter(
+            [session.db], host=host, httprequest=self.httprequest
+        ):
+            dbname = session.db
+        else:
+            all_dbs = db_list(force=True, host=host, httprequest=self.httprequest)
```
carmenbianca: At the very least, mark this line as diverging from upstream, with the rest being identical. Also at the very least, mark the date and commit from which the upstream code was taken.

carmenbianca (commented):
I made some comments.

I don't think this solution is good, though, unless there's really no other way. It completely overrides a very core private method, which is potentially really bad for security and reliability reasons.

Is there a way that we can petition upstream to adjust their code a little, or to provide us a hook that we need?

Possible hooks:

1. Upstream can pass httprequest as parameter without using it.
2. Upstream can create a method `Request.[_]list_all_databases` and call that from `_get_session_and_dbname`. We can then override _that_ method instead.

nilshamerlinck:
Hi @carmenbianca I agree with your concerns. When I worked on this initially, I needed the module asap so I adopted a pragmatic approach. I've just opened https://github.com/odoo/odoo/pull/104573 with the smallest possible change in numbers of lines to increase its chances of getting merged. If it does, we'll be in a position to get rid of the dirty overrides.

nilshamerlinck:
Hello @carmenbianca odoo PR was merged, so I will close this one. Your PR is ok now (once squashed) :)

carmenbianca:
@nilshamerlinck That's awesome news! Thank you for your work.

I will have time tomorrow to adjust my PR.

PR #2431, opened 2022-10-19
[14.0][FIX] upgrade_analysis: misleading representation of read/write computed fields

StefanRijnhart:
Backport of https://github.com/OCA/server-tools/pull/2416

Odoo 14 introduced the widescale usage of computed fields with readonly=False. In that case, the compute method functions as a default that can also be used to compute a value some time *after* the initial creation of the record.

In the OpenUpgrade analysis files, these fields would be misrepresented as computed fields rather than fields with a default function. This change fixes that.

### DIFF
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -134,8 +134,6 @@ def report_generic(new, old, attrs, reprs):
         if attr == "required":
             if old[attr] != new["required"] and new["required"]:
                 text = "now required"
-                if new["req_default"]:
-                    text += ", req_default: %s" % new["req_default"]
                 fieldprint(old, new, "", text, reprs)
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow:
Sad it's not retro-compatible haha

MiquelRForgeFlow:
Oh, this version also has the same problem with `letsencrypt` as https://github.com/OCA/server-tools/pull/2416.

StefanRijnhart:
@MiquelRForgeFlow yes, we should move this code to an oca-upgrade-analysis-lib some time.

MiquelRForgeFlow (approved):

StefanRijnhart:
CI is fixed here: https://github.com/OCA/server-tools/pull/2433

PR #2434, opened 2022-10-20
[14.0][ADD] tracking_manager

Kev-Roche:
This module allows to track all fields on every model that has a chatter, including one2many and many2many ones. This excludes the computed, readonly, related fields by default.
In addition, line changes of a one2many field are also tracked (e.g. product_uom_qty of an order_line in a sale order).


legalsylvain (commented):
Partial review. Nice job ! Looks interesting.

nimarosa:
Looks like a useful module. Can you check precommit so we can move forward?

Kev-Roche:
@nimarosa done

nimarosa (approved):

sebastienbeau (dismissed):
Some feedback after testing

First I like the UI ! 
To improve it, maybe we should add a mass action to track everything (when you are in the list of  tracking.model.field)

Regarding the case of inherit module (ex product.product)
We should customize on the product.product only the field that really belong to the product.product (product.template should be configured on product.template)

Regarding the datamodel, I think we can make some simplication.
Indead of creating a new model (tracking.model.field) I would simplify and put everything on ir.model.field
I will just add a computed boolean field "trackable_field" that will be tick for all field trackable (the fields where you create a tracking.model.field).
Then I will keep the same UI (the smartbutton, and view) to show all "trackable field".


Using the same model will simplify the code and avoid the case where you install a new module that create a new field and this field do not exist in the  tracking.model.field.



Thanks for your work

sebastienbeau:
Hi, I have done a lot of change, now the code should be cleaner and the support of o2m improve
@legalsylvain @Kev-Roche @nimarosa can you check my change ?
Thanks

tracking_manager/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+- In setting > models: select a model
+- Check "Apply custom tracking on fields", all the non readonly, related, computed fields will be tracked.
+- Button or smart button "Tracked Fields" allow to activate / deactivate custom_tracking on each field
+- By default, all sub fields of a one2many field are tracked if they changed. To manage the sub fields, you need to activate "Apply custom tracking on fields" on the one2many model. Button "One2many related models" is a shortcut for these models.
```
yostashiro: This content looks obsolete after the recent changes, perhaps.

It's probably more appropriate to have this content in CONFIGURE.rst.

yostashiro (commented):
Nice module!

tracking_manager/readme/USAGE.rst
sebastienbeau: @Kev-Roche can you help me to update the doc and add some screenshot ?

nimarosa (approved):
Like i said, very nice module. Congrats. 
Code seems good for a initial version, tested locally and seems to work (i tested two fields in two different models). 

Please check why tests are red, but it's approved for me.

sebastienbeau:
Test are green, it's was not related to the PR but due the an availability issue on a external source (some repo for debian package)

nimarosa:
Great. For me this is ready to merge.

tracking_manager/readme/USAGE.rst
Kev-Roche: I did it

Kev-Roche:
Thank you @sebastienbeau ! Simplier is better, very nice improvement of my initial code. 
I improve the readme and add some filters on fields view (tracking on and off)

cc @yostashiro , @legalsylvain

tracking_manager/readme/USAGE.rst
```diff
@@ -0,0 +1,12 @@
+- In setting > models: select a model
+- Check "Active" under Custom Tracking.
+- Check "Automatic configuration" in order to track new fields
+- Choose a domain or not
+- Click on "Update"
+- You can check tracked fields by clicking the smart button
```
yostashiro:
```suggestion
- You have two options - 1) manually configure tracked fields one by one, or 2) determine tracked fields based on a specific domain.
- For 1) manually configure tracked fields one by one
  - Click on Tracked Fields smart button, and select/unselect Custom Tracking.
- For 2) determine tracked fields based on a specific domain
  - Select "Automatic configuration", and then set the domain accordingly.
  - Click "Update" for the domain to take effect.
```

This is how I understood the design to be (correct me if I'm mistaken).

Kev-Roche: My bad, I was too focus on automatic configuration, you're right.

yostashiro:
@Kev-Roche Thanks for updating the readme.  Tests are failing now somehow...

hparfr:
Tests failing because of attachment_delete_restrict
> 2023-01-28 18:04:39,184 868 INFO odoo odoo.addons.attachment_delete_restrict.tests.test_attachment_delete_restrict: > ====================================================================== 
2023-01-28 18:04:39,184 868 ERROR odoo odoo.addons.attachment_delete_restrict.tests.test_attachment_delete_restrict: ERROR: 
>odoo.exceptions.AccessError: Sorry, you are not allowed to access this document.

tracking_manager/models/models.py
```diff
@@ -0,0 +1,151 @@
...
+from odoo import api, models, tools
+
+from ..tools import format_m2m
+
+# To avoid conflict with other module and avoid too long function name
+# specific tracking_manager method are prefixed with _tm
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
```
petrus-v, line 17: I'm wondering why we could not move all of this in `mail.thread` AbstractModel ? Using this code what's happen on models that do not inherit  from `mail.thread` ?

petrus-v (approved):
Looks nice addition, code reviews only!

tracking_manager/models/models.py
sebastienbeau: In fact we need to inherit all models.
For example you want to track the change on "sale.order.line" and post this information on the related "sale.order"

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,205 @@
...
+    _inherit = "ir.model"
+
+    active_custom_tracking = fields.Boolean(default=False)
+    tracked_field_count = fields.Integer(compute="_compute_tracked_field_count")
+    automatic_custom_tracking = fields.Boolean(
+        compute="_compute_automatic_custom_tracking",
+        readonly=False,
+        store=True,
+        default=False,
+        help=("If tick new field will be automatically tracked " "if the domain match"),
```
damdam-s:
```suggestion
        help=("If tick new field will be automatically tracked if the domain match"),
```

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,205 @@
...
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    active_custom_tracking = fields.Boolean(default=False)
+    tracked_field_count = fields.Integer(compute="_compute_tracked_field_count")
+    automatic_custom_tracking = fields.Boolean(
+        compute="_compute_automatic_custom_tracking",
+        readonly=False,
+        store=True,
+        default=False,
```
damdam-s:
```suggestion
```
no need of default `False` value on Boolean

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,205 @@
...
+
+from ast import literal_eval
+
+from odoo import api, fields, models, tools
+
+
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    active_custom_tracking = fields.Boolean(default=False)
```
damdam-s:
```suggestion
    active_custom_tracking = fields.Boolean()
```
no need of default `False` value on Boolean

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,205 @@
...
+    native_tracking = fields.Boolean(
+        compute="_compute_native_tracking",
+        store=True,
+    )
+    trackable = fields.Boolean(
+        compute="_compute_trackable",
+        store=True,
+    )
+
+    @api.depends("native_tracking")
```
damdam-s, line 165:
```suggestion
    @api.depends("native_tracking", "model_id", "model_id.automatic_custom_tracking", "model_id.automatic_custom_tracking_domain")
```
to be sure the field is properly recomputed

damdam-s:
@ivantodorovich just for you to know this module exists 😄

tracking_manager/models/ir_model.py
sebastienbeau: Hi dam, it's wanted, let's explain

The field automatic_custom_tracking is a computed readonly=False field.

So the idea is to only update field in some case keep existing value in most of the case.


Case 1: You install a new module that add some extra field to an existing model => if the field is track natively or match the domain, it's should be tracked.

Case 2: you activate or desactivate the option automatic_custom_tracking on the model => we do nothing as user can tick and untick and we do not want to reset all existing configuration.

Case 3: you modify the default domain (used as rule when a module install new field) => we do nothing as user can update the domain for new installed field but we do not want do change all existing configuration.

If a user want to reapply the domain on all fields then he can click on the button "update_custom_tracking" that we reset the value of all field https://github.com/OCA/server-tools/pull/2434/files/233e08cef4e959f7754f93e5ce0b3c6a3c84c732#diff-d2127356ce28776a51b642c4bf0e6c7762702ac27abc335c3b9eb9c204cb0d32R129

Thanks for your feedback

tracking_manager/models/ir_model.py
Kev-Roche: Thanks for your review @damdam-s , I applied your suggestions, except the last one.

PR #2439, opened 2022-10-27
[14.0][IMP] base_time_parameter: LGPL-3; rename model to base.time.parameter.version

ghost:
LGPL-3 licence since it is a dependency for payroll.

Each parameter version has a start date and a value.

### DIFF
--- a/base_time_parameter/__manifest__.py
+++ b/base_time_parameter/__manifest__.py
@@ -10,7 +10,7 @@
         with time based versions.
     """,
     "license": "AGPL-3",
-    "author": "Nimarosa, Odoo Community Association (OCA)",
+    "author": "Nimarosa, appstogrow, Odoo Community Association (OCA)",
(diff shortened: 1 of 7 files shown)


legalsylvain (changes requested):
- could you split commits (. Licence change + model rename) ?
- could you provide migration scripts ?

Thanks.

nimarosa:
@appstogrow Hi Henrik, i think like @legalsylvain said, we have to split the commits. I don't think migration scripts are necessary since the module it's too new and i don't think anyone is using it right now. 

I think we can merge with splitted commits. Can you do it?

norlinhenrik:
I have written [migration script for payroll](https://github.com/OCA/payroll/pull/87) which will depend on base_time_parameter.

nimarosa:
Okay so we can merge this so we can continue and test the PR in payroll. 

/ocabot merge patch

PR #2441, opened 2022-10-29
[16.0] session_db

sbidoul:
Migrate `session_db` from https://github.com/odoo/odoo-extra/tree/master/session_db to 16.0.

I'm going to play with this and see if it fares well enough, so I keep the PR draft for now.

@nseinlet are you ok with us maintaining your session_db module in OCA? As there was no license in odoo-extra, I chose LGPL. Let me know if there is any concern with that.


session_db/session.py
```diff
@@ -0,0 +1,126 @@
...
+        self._cr.execute("DELETE FROM http_sessions WHERE sid=%s", [session.sid])
+
+    @with_cursor
+    def get(self, sid):
+        self._cr.execute(
+            "UPDATE http_sessions "
+            "SET write_date = now() at time zone 'UTC' "
+            "WHERE sid=%s",
+            [sid],
+        )
```
sbidoul: Updating the write_date when reading is likely not necessary. If I read the code correctly the FilesystemSessionStore does not do that.

session_db/session.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # This method is not part of the Session interface but is called nevertheless,
+    # so let's get it from FilesystemSessionStore.
+    rotate = http.FilesystemSessionStore.rotate
+
+    @with_cursor
+    def vacuum(self):
+        self._cr.execute(
+            "DELETE FROM http_sessions "
+            "WHERE now() at time zone 'UTC' - write_date > '7 days'"
```
sbidoul: use SESSION_LIFETIME constant (which is now 90 days, btw)

session_db/session.py
nseinlet: The point of doing this is to have the session not expire after 8 days.

nseinlet:
Yes, nice to see it useful

session_db/session.py
sbidoul: Yes, but the FilesystemSessionStore does not do that right? Any specific reason to do it differently here?

adrienpeiffer (approved):

lmignon (approved):
LGTM (Code + functional)

AnizR (approved):
Code LGTM, do you plan to make PGSessionStore thread safe?

sbidoul:
> Code LGTM, do you plan to make PGSessionStore thread safe?

No, this is really meant to be used with multiple workers.

I'm going ahead and merging this. I'll remain available in case any issue show up.

/ocabot merge patch

PR #2444, opened 2022-10-31
[FIX] auditlog: autovacuum performance

gurneyalex:
Add database indices on the foreign key fields of the auditlog models. Without these indices, the performance of the autovacuum cron are terrible because the "ON DELETE SET NULL" trigger has to make a full table scan on the auditlog_log and auditlog_log_line tables when auditlog_http_session and auditlog_http_request rows are deleted.


nimarosa (approved):

rvalyi (approved):
LGTM

StefanRijnhart (approved):

PR #2448, opened 2022-11-02
[15.0][IMP] auditlog: dedicated security groups for model access

BertVGroenendael:
We would like to see a change in the user rights of auditlog. In this to have an easier way to link permissions to other users and not just the default.


StefanRijnhart:
Hi @BertVGroenendael, thanks for the PR! Please run the OCA pre-commit, and mention the module in the commit name using the conventional '[IMP]' prefix and a clear description. I would use
```
[IMP] auditlog: dedicated security groups for model access
```

StefanRijnhart:
Please also add a line or two to https://github.com/OCA/server-tools/blob/15.0/auditlog/readme/USAGE.rst about assigning the Auditlog User group to users that need to be able to access the auditlogs through the `Action -> View logs` actions on the records of each logged model.

auditlog/security/res_groups.xml
```diff
@@ -0,0 +1,25 @@
+<odoo>
+    <record model="ir.module.category" id="security_auditlog_groups">
+        <field name="name">Auditlog Rights</field>
+    </record>
+
+    <record model="res.groups" id="group_auditlog_user">
+        <field name="name">Auditlog User</field>
+        <field name="category_id" ref="auditlog.security_auditlog_groups" />
+        <field name="implied_ids" eval="[(4, ref('base.group_user'))]" />
```
StefanRijnhart: I would leave this implied group out.

auditlog/readme/USAGE.rst
```diff
@@ -19,3 +19,8 @@ To activate it and/or change the delay, go to the
 `Auto-vacuum audit logs` entry:
 
 .. image:: ../static/description/autovacuum.png
+
+There are two possible groups configured to which one may belong. The first
+is group_auditlog_user. This group has read-only rights to auditlog. The second
+is the group_auditlog_manager. This has the same rights as the group_auditlog_user,
+but extended with Create, Update, Delete to CRUD.
```
StefanRijnhart:
```suggestion
There are two possible groups configured to which one may belong. The first
is the Auditlog User group. This group has read-only access to the auditlogs of
individual records through the `View Logs` action. The second group is the
Auditlog Manager group. This group additionally has the right to configure the
auditlog configuration rules.
```

sebalix (approved):
Nice addition, need the last comment of @StefanRijnhart to be addressed and we are good, pre-approving.

StefanRijnhart (approved):

etobella (approved):
Makes sense to me

StefanRijnhart:
Oh no, it's flake8-gitlapocalypse! Fixed here: https://github.com/OCA/server-tools/pull/2467

pedrobaeza:
Yes, it's an annoyance, but there's no other solution than updating each repo...

StefanRijnhart:
Thanks for merging the fixing PRs @pedrobaeza!

/ocabot merge minor

PR #2450, opened 2022-11-02
[16.0][MIG] dbfilter_from_header

ryanc-me:
The base `httprequest` isn't passed to `db_filter()` anymore.

Also, it looks like `http.request` sometimes is not bound - specifically in [this call](https://github.com/odoo/odoo/blob/16.0/odoo/http.py#L1144), so we may return multiple DBs in some cases. Unfortunately it looks non-trivial to cover properly.

I think probably this will be fine. We've tested with several client sites.

It would be great if someone can test with their own setup also!


ryanc-me:
Closing, I see there's another PR already.

PR #2457, opened 2022-11-08
[14.0][FIX] base_view_inheritance_extension: must strip before parsing

MiquelRForgeFlow:
Selection_650


rousseldenis (approved):

LoisRForgeFlow (approved):
LGTM

For instance the problem was happening in this standard odoo view:

image

between `}` and `</attribute>` there is a space

PR #2461, opened 2022-11-10
[14.0][IMP] base_custom_info: new field: code

norlinhenrik:
I would like to add the field "code", because I would like to use it for payroll calculations.


nimarosa (approved):

norlinhenrik:
`record.get_custom_info_value(properties)` is not so useful when only the code is known. Now this also works:
`record.get_custom_info_value("my_code")`

norlinhenrik:
When the type is Selection, by default _compute_value() will return the option's display_name. Now it may return the code:
`custom_info_value_record.with_context(custom_info_option_field="code").value`

norlinhenrik:
I have just improved the code and code descriptions.
In payslip rules I want to look for custom info by code, e.g.
`value = employee.address_home_id.get_custom_info_value(code="house_tax_value").value`

@pedrobaeza @Rad0van What do you think?

Rad0van:
@norlinhenrik I actually have a small module in place where I add code to custom.info.property but you use it in an extended way. I like it.

pedrobaeza (approved):

norlinhenrik:
Inside the `custom.info` model, I made `get_custom_info()` to replace `get_custom_info_value()` (the original method is still there, with a deprecation notice and returning `get_custom_info()`).

Then a typical use case would be `get_custom_info(properties or code).value`, and I would prefer this.

What do you think?

pedrobaeza:
Seems OK as you are respecting previous API. Some tests are failing.

nimarosa:
Agree with the new changes.

nimarosa (approved):

norlinhenrik:
Can we please set this PR in **draft** state?

I will change so that `get_custom_info()` will return not the value record, but the actual value, similar to `get_time_parameter()`.

norlinhenrik:
`get_custom_info()` is ready now, with a test. I think the PR is ready for migration.

nimarosa:
@appstogrow @norlinhenrik Tests shows this error: 
`TypeError: get_custom_info() got an unexpected keyword argument 'properties'`

The rest of the changes look good.

norlinhenrik:
Now I have fixed the test, rebased and squashed commits.

PR #2462, opened 2022-11-14
[16.0][MIG] base_fontawesome

legalsylvain:
Port ``base_fontawesome`` from 15.0 to 16.0
- remove useless file
- adapt asset replacement to new location

- DONT change the fontawesome. (still the 5.15.4, not the 6.2.0)
(I don't need it, but feel free to make a PR against mine, or a second PR once this one is merged).


legalsylvain:
/ocabot migration base_fontawesome

CC : @simahawk, @etobella , @olgamarcocb

base_fontawesome/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Base Fontawesome",
     "summary": """Up to date Fontawesome resources.""",
-    "version": "15.0.5.15.4",
+    "version": "16.0.1.0.1",
```
etobella: We usually tried to keep the version of font awesome on the version of the module, but it is ok for me to change it

etobella (approved):

base_fontawesome/__manifest__.py
simahawk: I'd stick to that rule

simahawk (approved):

nimarosa (approved):

StefanRijnhart:
Setting to 'Needs fixing' for the version number alignment.

base_fontawesome/__manifest__.py
legalsylvain: Fixed. I didn't know that rule. Thanks for your review.

(could be merged I guess now).

etobella (approved):

PR #2463, opened 2022-11-16
[MIG] base_exception: Migration to 16.0

matiasperalta1:


base_exception/views/base_exception_view.xml
```diff
@@ -39,6 +39,7 @@
                             <field name="description" />
                         </group>
                     </group>
+                    <field name="exception_type" invisible="1" />
```
sebalix, line 42: Why it is needed? This field is already displayed in the group below.

matiasperalta1: Before, when Odoo rendered a field, used to declare the invisible=1. As of this version, it no longer takes this into account for those users who are not part of this group. Later we see that the field is used in a domain and odoo understands that that field was never brought into view. The field is displayed into the group but for users who are not part of that group the field does not exist and that is where the error occurs. In order to this not to happen, the field without group must be called with invisible=1

aliciagaarzo (approved):

sebalix (approved):

sebalix:
@matiasperalta1 can you add `odoo_test_helper` dependency in the `test-requirements.txt` file to get the CI green?
https://github.com/OCA/server-tools/blob/14.0/test-requirements.txt

matiasperalta1:
@sebalix Done

sebastienbeau (approved):
LGTM code review

/ocabot merge

PR #2464, opened 2022-11-17
[14.0][IMP] base_time_parameter: more data types

norlinhenrik:
@nimarosa This is a quick draft. Will finish on Sunday.

field value_text -> value (will require migration script)

new function _validate() will check that the value can convert to the desired data type.

Another function/method can return the desired data type (not implemented yet).

### DIFF
--- a/base_time_parameter/migrations/14.0.3.0.0/pre-migration.py
+++ b/base_time_parameter/migrations/14.0.3.0.0/pre-migration.py
@@ -1,0 +1,28 @@
+from openupgradelib import openupgrade
+
+
+def rename_fields(env):
+    openupgrade.rename_fields(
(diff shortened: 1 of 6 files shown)


nimarosa:
@norlinhenrik i like the approach. Great. I will merge the other one in payroll and then when this is ready I change the function. 

Thanks Henrik.

norlinhenrik:
base_time_parameter is ready now.

nimarosa (approved):
Looks very good. Thanks @norlinhenrik I think this approach will suit us more and leave us with a lot of possibilities.

nimarosa:
Merging as major to trigger migration scripts. 

/ocabot merge major

PR #2470, opened 2022-11-21
[14.0] [IMP] base_time_parameter: Add get param 

nimarosa:
New data_type param that allows to get the date of the version. This is useful for cases when you need to know the start_date of the version, maybe for an average function or some logic with dates. 

Minor improvements in value types, boolean value now return False and not and error, simplify logic.

### DIFF
--- a/base_time_parameter/models/base.py
+++ b/base_time_parameter/models/base.py
@@ -10,9 +10,7 @@ from odoo import models
 class Base(models.AbstractModel):
     _inherit = "base"
 
-    def get_time_parameter(
-        self, code, date=None, raise_if_not_found=True, get="value"
-    ):
+    def get_time_parameter(self, code, date=None, raise_if_not_found=True, get="value"):
         if type(date) is str:
             # Get the date/datetime from a field on the record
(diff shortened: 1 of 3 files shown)


nimarosa:
@norlinhenrik Henrik, can you fast review this? Since i need this modifications for a use case i have in production. If you have any suggestions i will apply them fast.

norlinhenrik:
Interesting improvement. Yes it is good to be able to get the start date. I suggest the name `get` instead of `data` / `data_type`. pre-commit will accept `get` as a variable, so I think it is not a reserved keyword in python.

norlinhenrik:
@nimarosa 
Reviewing the code, I see that it is not consistent to say `_get_value(get=something_else)` Could you please rename
_get_value_from_model_code_date() to `_get_from_model_code_date()`
_get_value() to `_get()`
Thank you!

nimarosa:
@norlinhenrik Did the changes, please check it so we can move forward.

norlinhenrik (approved):
Code review, looks good!

PR #2162, opened 2021-08-19
[14.0][IMP] auditlog, add auditlog.log.line view

kittiu:
Add auditlog.log.line menu.

PR #2471, opened 2022-11-22
[OU] auditlog: Improve time of upgrade

etobella:
When upgrading a DB with 11.000.000  auditlog.log.line, it took more than 2 hours (Probably more than 3, I did 6 of the 10 fields in 1.5 hours :sob: ) to update this module (more than 10 minutes for each field on #2162). This change should improve the time (it changed to 9 minutes), as we are doing all the changes at the same time :smile:


etobella:
@kittiu

kittiu (approved):
Great improvement.

pedrobaeza (approved):

PR #2459, opened 2022-11-09
[13.0] update dotfiles and use GitHub actions

sbidoul:

PR #2474, opened 2022-11-25
[13.0] Create module base_domain_inverse_function

grindtildeath:
This module provides functions to decompose normalized domains 
into domains operands, as these functions are the inverse of
`AND` and `OR` functions available in `odoo.osv.expression`.

### DIFF
--- a/base_domain_inverse_function/inverse_expression.py
+++ b/base_domain_inverse_function/inverse_expression.py
@@ -1,0 +1,92 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo.osv.expression import (
+    AND,
+    AND_OPERATOR,
+    DOMAIN_OPERATORS,
+    NOT_OPERATOR,
+    OR,
+    OR_OPERATOR,
+)
+
+
+def inverse_combine(domain, operator):
+    """Decompose normalized domain into operands according to operator
+
+    The result can be then altered easily to inject domain operands before
+    rebuilding a new domain using the corresponding function from osv.expression.
+
+    :param domain: A normalized domain
+    :param operator: The "main" domain operator of the domain being either '&' or '|'
+                     (must be the first operator in a normalized domain)
+
+    :return list: A list of domains
+    """
+    if operator not in DOMAIN_OPERATORS:
+        raise Exception("Unsupported operator parameter: %s" % operator)
+    operator_func = {AND_OPERATOR: AND, OR_OPERATOR: OR}
+    other_operator = OR_OPERATOR if operator == AND_OPERATOR else AND_OPERATOR
+    result = []
+    operator_elements_stack = []
+    other_elements_stack = []
+    elements_stack = []
+
+    last_element = False
+
+    # 1. Loop over the domain in reverse order
+    for element in reversed(domain):
+        if element == NOT_OPERATOR:
+            raise Exception(
+                "Inversing domains including NOT operator ('!') is not supported"
+            )
+        if element in DOMAIN_OPERATORS:
+            # 3. When we reach an operator:
+            # - pop the last item from the element stack to the corresponding operator stack
+            # - if such stack contains only one element, the actual operator applies to the two
+            #   last items in the elements stack, so pop the penultimate item as well
+            if element != operator:
+                if len(elements_stack) > 0:
+                    other_elements_stack.append([elements_stack.pop()])
+                    if (
+                        len(other_elements_stack) == 1
+                        and last_element not in DOMAIN_OPERATORS
+                    ):
+                        other_elements_stack.append([elements_stack.pop()])
+            else:
+                if len(elements_stack) > 0:
+                    operator_elements_stack.append([elements_stack.pop()])
+                    if (
+                        len(operator_elements_stack) == 1
+                        and last_element not in DOMAIN_OPERATORS
+                    ):
+                        operator_elements_stack.append([elements_stack.pop()])
+            last_element = element
+        else:
+            # 4. If actual element is a tuple, but last element was an operator, empty the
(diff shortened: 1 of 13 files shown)


i-vyshnevska:
to fix the build you need to change from gitlab to github
https://github.com/OCA/server-tools/blob/ef4e864ee4b6dcd4e30ece4c26c279b68e4d563b/.pre-commit-config.yaml#L66

sebalix (approved):

grindtildeath:
@i-vyshnevska that probably won't be enough, we need to move to GH actions but there are some issues in this repo, cf https://github.com/OCA/server-tools/pull/2459

base_domain_inverse_function/inverse_expression.py
```diff
@@ -0,0 +1,90 @@
...
+    rebuilding a new domain using the corresponding function from osv.expression.
+
+    :param domain: A normalized domain
+    :param operator: The "main" domain operator of the domain being either '&' or '|'
+                     (must be the first operator in a normalized domain)
+
+    :return list: A list of domains
+    """
+    if operator not in DOMAIN_OPERATORS:
+        raise Exception("Unsupported operator parameter: %s" % operator)
```
mmequignon, line 26: should we make those exceptions translatable?

base_domain_inverse_function/tests/test_inverse_function.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_inverse_complex_or_and_and(self):
+        result = [
+            ["&", "&", ("a", "=", 1), ("b", "=", 2), ("c", "=", 3)],
+            ["&", "&", ("a", "=", 1), ("b", "=", 2), ("c", "=", 3)],
+        ]
+        self.assertEqual(
+            inverse_OR(self.complex_domain_or_and_and),
+            result
+        )
```
mmequignon: shouldn't you add a test that ensures that an OR domain isn't split by `inverse_AND`, and an AND domain isn't split by `inverse_OR`
```
self.assertEqual(inverse_OR(self.complex_domain_and_or_or),  self.complex_domain_and_or_or)
self.assertEqual(inverse_AND(self.complex_domain_or_and_and), self.complex_domain_or_and_and)
```

base_domain_inverse_function/inverse_expression.py
```diff
@@ -0,0 +1,90 @@
...
+    elements_stack = []
+
+    last_element = False
+
+    # 1. Loop over the domain in reverse order
+    for element in reversed(domain):
+        if element == NOT_OPERATOR:
+            raise Exception("Inversing domains including NOT operator ('!') is not supported")
+        if element in DOMAIN_OPERATORS:
+            # 3. When we reach an operator:
```
mmequignon, line 43: there's no 2.

base_domain_inverse_function/inverse_expression.py
```diff
@@ -0,0 +1,90 @@
...
+    AND,
+    AND_OPERATOR,
+    DOMAIN_OPERATORS,
+    NOT_OPERATOR,
+    OR,
+    OR_OPERATOR,
+)
+
+
+def inverse_combine(domain, operator):
```
mmequignon, line 13: I find it very difficult to understand.
Domains use infix notation `a + b -> + b a`.
I remember in compilation courses at school, we created a lexer for that notation, and I would have gone this way.
The grammar here is quite easy: `|, &, ! and tuples`, I think it would be easier to review and to maintain.
However, it might take some time to implement.
Ping me if you have questions.

base_domain_inverse_function/inverse_expression.py
grindtildeath: As it's a technical module, IMO it doesn't make sense to translate it for the end user.

base_domain_inverse_function/inverse_expression.py
grindtildeath: It's there: https://github.com/OCA/server-tools/pull/2474/files#diff-09b04ca66e438a334ce118b2870d239e3d8612f6043059bfbdfd6f084ff291bdR73

base_domain_inverse_function/tests/test_inverse_function.py
grindtildeath: There's no sense to use the `inverse_OR` function on a domain that doesn't begin with an OR operator, as it means such domain wasn't created with `OR` function. I could still add a test and raise an error in this case.

base_domain_inverse_function/inverse_expression.py
grindtildeath: Feel free to propose another implementation if you can do better :stuck_out_tongue_closed_eyes:

base_domain_inverse_function/tests/test_partner_domains.py
```diff
@@ -0,0 +1,80 @@
...
+        decomposed_and_domains = inverse_AND(composed_domain)
+        decomposed_or_domains_1 = inverse_OR(decomposed_and_domains[0])
+        decomposed_or_domains_2 = inverse_OR(decomposed_and_domains[1])
+        # Ensure there is at least 1 result
+        self.assertTrue(self.partner_model.search(composed_domain))
+        # Ensure result is same after inverse
+        self.assertEqual(
+            self.partner_model.search(composed_domain),
+            self.partner_model.search(
+                OR([AND(decomposed_or_domains_1), AND(decomposed_or_domains_2)])
```
sonhd91:
```suggestion
                AND([OR(decomposed_or_domains_1), OR(decomposed_or_domains_2)])
```
Due to purpose of this tests, It's should be `AND([OR(....), OR(....)])`, right ?

grindtildeath: Good catch :+1:

gurneyalex (approved):

PR #2479, opened 2022-11-30
[16.0][MIG] onchange_helper: Migration to 16.0

kevinkhao:


bealdav (approved):

chandni299 (approved):

angelmoya (approved):

sebastienbeau:
Thanks for the work but before merging it, where do you still need this module in V16?
Indeed odoo have dropped a lot of onchange (no more on sale, account move) and replace them by compute readonly=False so in several case we can drop the dependency on this module to simplify the code and improve the perf

rousseldenis (approved):
Code review.

@kevinkhao Maybe we could add something in README that onchanges tend to disappear due to the introduction of 'computed stored readonly False fields' in most cases

StephaneMangin (approved):

simahawk:
> Thanks for the work but before merging it, where do you still need this module in V16? Indeed odoo have dropped a lot of onchange (no more on sale, account move) and replace them by compute readonly=False so in several case we can drop the dependency on this module to simplify the code and improve the perf

I agree but I doubt we'll have them disappear everywhere. Of course we should put a lot of effort in reviews for v16 to ask for refactoring of business logic out of onchanges.... but we won't succeed everywhere/every time I fear :wink: 

I'd keep the module and document this in the roadmap.
We could even an annoying post_load hook to say that the module is deprecated and the you should refactor the modules that use it (or their dependencies) to not use onchanges for business logic.

florian-dacosta:
> We could even an annoying post_load hook to say that the module is deprecated and the you should refactor the modules that use it (or their dependencies) to not use onchanges for business logic.

I would not go this far, there are still a lot of onchange in Odoo core modules that we just can't refactore ourselves.

rousseldenis:
> > We could even an annoying post_load hook to say that the module is deprecated and the you should refactor the modules that use it (or their dependencies) to not use onchanges for business logic.
> 
> I would not go this far, there are still a lot of onchange in Odoo core modules that we just can't refactore ourselves.

Agree. Something that should more go into odoo-module-migrator IMHO

kevinkhao:
> Thanks for the work but before merging it, where do you still need this module in V16? Indeed odoo have dropped a lot of onchange (no more on sale, account move) and replace them by compute readonly=False so in several case we can drop the dependency on this module to simplify the code and improve the perf

@sebastienbeau sorry for late reply, I migrated it because it was a dependency on PIM

Updated readme with ROADMAP with your comments

PR #2485, opened 2022-12-02
[14.0][FIX] auditlog: consistency with Many2one fields

SilvioC2C:
Odoo allows creating or updating records with values like ``{"many2one_field_id": empty.recordset()}``, but this crashes when using ``deepcopy``, which fails in recreating the ``api.Environment`` object attached to the recordset.
This commit should fix the issue by updating the values before passing them to ``deepcopy``.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -318,6 +318,7 @@ class AuditlogRule(models.Model):
         def create_fast(self, vals_list, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            vals_list = rule_model._update_vals_list(vals_list)
             vals_list2 = copy.deepcopy(vals_list)
             new_records = create_fast.origin(self, vals_list, **kwargs)
             new_values = {}
@@ -723,3 +724,15 @@ class AuditlogRule(models.Model):
                 act_window.unlink()
         self.write({"state": "draft"})
         return True
+
+    @api.model
+    def _update_vals_list(self, vals_list):
+        # Odoo supports empty recordset assignment (while it doesn't handle
+        # non-empty recordset ¯\_(ツ)_/¯ ), it could be an Odoo issue, but in
+        # the meanwhile we have to handle this case to avoid errors when using
+        # ``deepcopy`` to log data.
+        for vals in vals_list:
+            for fieldname, fieldvalue in vals.items():
+                if isinstance(fieldvalue, models.BaseModel) and not fieldvalue:
+                    vals[fieldname] = False
+        return vals_list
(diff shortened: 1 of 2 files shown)


SilvioC2C:
@sebalix I took the opportunity to refactor the tests a bit, for example by removing the `write` and `unlink` tests from `test_LogCreation()` and moving them to their own test methods.

auditlog/models/rule.py
```diff
@@ -318,6 +318,7 @@ def create_full(self, vals_list, **kwargs):
         def create_fast(self, vals_list, **kwargs):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
+            vals_list = self.env["auditlog.rule"]._update_vals_list(vals_list)
```
sebalix:
```suggestion
            vals_list = rule_model._update_vals_list(vals_list)
```

auditlog/models/rule.py
```diff
@@ -723,3 +724,12 @@ def unsubscribe(self):
                 act_window.unlink()
         self.write({"state": "draft"})
         return True
+
+    @api.model
+    def _update_vals_list(self, vals_list):
+        # Odoo supports empty recordset assignment: handle this case
```
sebalix:
```suggestion
        # Odoo supports empty recordset assignment (while it doesn't handle non-empty
        # recordset ¯\_(ツ)_/¯ ), it could be an Odoo issue, but meanwhile we have to handle
        # this case to avoid errors when using `deepcopy` below.
```

auditlog/tests/test_auditlog.py
```diff
@@ -194,6 +172,103 @@ def test_LogCreation6(self):
         if self.groups_rule.capture_record:
             self.assertTrue(len(log_record.line_ids) > 0)
 
+    def test_LogCreation7(self):
+        """Seventh test, create several records at once (with create multi
+        feature starting from Odoo 12) and different values on Many2one fields.
+        Check that creation goes as planned (no error coming from ``deepcopy``)
+        """
```
sebalix:
```suggestion
        """Seventh test, create several records at once with different M2O values.

        Check that creation goes as planned (no error coming from ``deepcopy``)
        """
```

auditlog/tests/test_auditlog.py
```diff
@@ -194,6 +172,103 @@ def test_LogCreation6(self):
...
+            [
+                ("model_id", "=", self.groups_model_id),
+                ("method", "=", "create"),
+                ("res_id", "in", groups.ids),
+            ]
+        )
+        self.assertEqual(len(logs), len(groups))
+
+    def test_LogUpdate(self):
+        """Tests write results, using multiple values for ``Many2one`` field"""
```
sebalix: To align with the other docstring
```suggestion
        """Tests write results with different M2O values."""
```

sebalix (commented):
Small comments, LG.
Run pre-commit too to fix the lints.

sebalix (approved):

jcoux (approved):

sebalix:
Note for later: I don't think Odoo should support that kind of data in `create` and `write` calls, but it is used as a side-effect when creating orderpoints (empty recordset are seen like false values), so... better to handle this case in auditlog as there is little chance that Odoo fixes it.

PR #2486, opened 2022-12-02
[REF] date_range: Generate new stable version

SirTakobi:
I'm having trouble with Github actions for `10.0`: in https://github.com/OCA/l10n-italy/actions/runs/3603460390/jobs/6071687725#step:4:50 it is choosing a very old release (**10.0.1.0.2**) of module **date_range**.

I have found out that in https://wheelhouse.odoo-community.org/oca-simple/odoo10-addon-date-range/ (as well as https://pypi.org/project/odoo10-addon-date-range/) **10.0.1.0.2** is the most recent stable version, so maybe that is the reason.
Hoping this PR creates a new stable version.


users_ldap_populate/__manifest__.py
```diff
@@ -5,6 +5,8 @@
     "name": "LDAP Populate",
     "version": "10.0.1.0.3",
     "author": "Therp BV, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-tools"
+               "/tree/10.0/users_ldap_populate",
```
sbidoul, line 9: These are website urls are unusual. Did you make them manually?

SirTakobi: Thanks for having a look!
Yes I did them _manually_ (with a script) because I like the website of a module to point to the module's information instead of the repository's information

sbidoul: Yeah but you are actually exploiting a bug in the linter so it may break in the future. If you want that it's better to work to have the feature in the official repo template.

SirTakobi: > Yeah but you are actually exploiting a bug in the linter so it may break in the future. If you want that it's better to work to have the feature in the official repo template.

:sweat_smile: I thought it was allowed on purpose, since this kind of URL is accepted by the guidelines:
> The website key must be [https://github.com/OCA/<repo](https://github.com/OCA/%3Crepo)>, so as to provide the most relevant link to discover more information about the addon. That link shows the repository README. Alternatively [https://github.com/OCA/<repo>/tree/<branch>/<addon](https://github.com/OCA/%3Crepo%3E/tree/%3Cbranch%3E/%3Caddon)> may be used, to provide a direct link to the addon README, which includes proper credits (authors, contributors and theirs companies), and links to the relevant information on the OCA website.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules)

PR #2489, opened 2022-12-05
[IMP] 14.0 base_generate_code allow to manually set code

PierrickBrun:
In my case this is needed to import data from another system.

I think this module should allow to manually set a code, it should be removed in the implementing modules if needed


Kev-Roche (approved):
LGTM

hparfr (approved):

PR #2295, opened 2022-02-21
[12.0] database_cleanup: refactor tests file

maisim:
* database_cleanup: [IMP] refactoring tests file
Split the big test method in tiny ones to ease read of results and to make adding tests easier 

* database_cleanup: [FIX] missing cleanup in tests
Remove tests data in db to avoid fail on next run on the same db

PR #2491, opened 2022-12-06
[12.0][ADD] database_cleanup: Purge fields (ir.model.fields) added

ypapouin:
Usefull to remove orphaned fields after long path database migration

### DIFF
--- a/database_cleanup/models/purge_fields.py
+++ b/database_cleanup/models/purge_fields.py
@@ -1,0 +1,122 @@
+# Copyright 2014-2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# pylint: disable=consider-merging-classes-inherited
+from odoo import _, api, models, fields
+from odoo.exceptions import UserError
+from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
+from ..identifier_adapter import IdentifierAdapter
+
+
+class CleanupPurgeLineField(models.TransientModel):
+    _inherit = 'cleanup.purge.line'
+    _name = 'cleanup.purge.line.field'
+    _description = 'Purge fields'
+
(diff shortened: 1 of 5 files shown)


thomaspaulb:
@ypapouin Could you rebase?

alexis-via (approved):
Successfully tested

alexis-via:
This feature is absolutely needed in database_cleanup ! It solves some blocking problems in migrations.

database_cleanup/models/purge_fields.py
```diff
@@ -0,0 +1,123 @@
...
+# pylint: disable=consider-merging-classes-inherited
+from odoo import _, api, models, fields
+from odoo.exceptions import UserError
+from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
+from ..identifier_adapter import IdentifierAdapter
+
+
+class CleanupPurgeLineField(models.TransientModel):
+    _inherit = 'cleanup.purge.line'
+    _name = 'cleanup.purge.line.field'
```
thomaspaulb, line 12: How about adding a test [here](https://github.com/OCA/server-tools/blob/74bee5ff8678ef01db56e6d93aa5734f82f8d9b1/database_cleanup/tests/test_database_cleanup.py#L71) that, after creation of the dummy model, creates a field on it as well, then purges the field?

A test would also make it easier to review when porting this feature to versions 13.0 and higher.

thomaspaulb: Or if you like, throw in a review [here](https://github.com/OCA/server-tools/pull/2295) so that I can merge it, and the tests will be that much nicer.

astirpe:
Forward-ported to V15 in https://github.com/OCA/server-tools/pull/2905 and added some tests

thomaspaulb:
Superseded by https://github.com/OCA/server-tools/pull/2906 with backported tests.

PR #2492, opened 2022-12-06
[14.0] autovacuum_message_attachment: Support inherited models

grindtildeath:
Selection of attachments through inheriting model can be useful if
we want to purge attachments that are extended through another
model.

For example, shipping.label (from base_delivery_carrier_label)
inherits from ir.attachment but the attachment itself is linked to
the stock.picking model.
By using inheriting_model defined to shipping.label and setting
model_ids to stock.picking, we can safely select the attachments
linked to stock.pickings that are shipping labels and avoid deleting
any picking list or delivery note reports.


grindtildeath:
ping @florian-dacosta @etobella @hkapatel-initos

guewen (approved):

StephaneMangin (approved):

florian-dacosta (approved):

leemannd (approved):

gurneyalex (approved):

PR #2493, opened 2022-12-06
[14.0] base_conditional_image: Migrate to new image fields

grindtildeath:
partial cherry-pick of https://github.com/OCA/server-tools/commit/8076a0168b1362f2a27cb3b947dd8dbe4df3df12


leemannd (approved):

StephaneMangin (approved):

base_conditional_image/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "category": "Misc",
     "depends": [
-        "base",
+        "mail",
```
sebalix, line 12: This change deserves a comment if we go this way, no?

simahawk: Yes, at least... However, I'd prefer to something like this instead https://github.com/OCA/web-api/pull/3/files#diff-5d31acaa664c7c8d9dd2b526e6f5acf160e25fd3cd837db900de3e6657bb5b68

PR #2497, opened 2022-12-07
[14.0][IMP] base_time_parameter: value_reference: standard selection

norlinhenrik:
Inherit like this:
```
value_reference = fields.Reference(
    selection_add=[("account.account", "Account")],
)
```

### DIFF
--- a/base_time_parameter/models/__init__.py
+++ b/base_time_parameter/models/__init__.py
@@ -1,6 +1,6 @@
 # Author Copyright (C) 2022 Nimarosa (Nicolas Rodriguez) (<nicolasrsande@gmail.com>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from . import base
 from . import base_time_parameter
 from . import base_time_parameter_version
-from . import base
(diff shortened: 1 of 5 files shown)


nimarosa (approved):

nimarosa:
Very great change. Now it's easy to use value_reference.

Another improvement I have in mind is a view tweak maybe we can include it here: 

- Hide value reference field in views table when value_reference it's not selected

Do you think it's possible?

norlinhenrik:
> Hide value reference field in views table when value_reference it's not selected

I think this would require javascript customization since the optional value is "hide" or "show".
If it would be "1" or "0", probably we could hide value_reference like this: attrs="{'hide':[('type', '!=', 'reference')]}"

So I set optional="hide" and added to the Reference a short description of how to show the column.

nimarosa:
@norlinhenrik I think you can use the "column_invisible", something like this: 

attrs="{'column_invisible':[('parent.type', '!=', 'reference')]}"

"parent" is necessary to access the version parent class (base Time parameter) because the type field is in main class not in versions one. 

This should work, I used in some modules. We can also do the other part, and hide the value field if reference is selected. That way we only show only one field (value or value_reference) depending on type.

norlinhenrik:
Thank you, @nimarosa !!! I think you can merge it.

nimarosa (approved):

Issue #2502, opened 2022-12-16
letsencrypt module is not compatible with acme>=2.0.0


petrus-v:
## Module
letsencrypt

## Describe the bug

`letsencrypt` module rely on [acme](https://pypi.org/project/acme/#history) library.

This module is not compatible with the latest acme version >= 2.0.0 

## To Reproduce
**Affected versions**: at lease v15.0

1. update acme lib
2. launch letsencrypt module unittest

**Expected behavior**
test pass

**Additional context**
This has been discover while merging https://github.com/OCA/server-tools/pull/2340

I've suggested a PR #2503  for the time being to freeze acme version prior to v2.0.0. I suppose maintainers should take care of that and migrate the odoo module to be compatible with the latest acme package.

petrus-v:
Root cause may be related to a flaky test as suggested by https://github.com/OCA/server-tools/pull/2340#issuecomment-1354949191

PR #2503, opened 2022-12-16
[FIX] letsencrypt: freeze acme lib before version 2.0.0

petrus-v:
I like to suggest this PR to avoid slowdown merge process on other PR but should be improved by #2502 (which I've not plan to deep into right now).


pedrobaeza (approved):
/ocabot merge patch

PR #2507, opened 2022-12-21
[15.0][FIX] fetchmail_incoming_log: delete test_mail dependency

MiquelRForgeFlow:
If it's merged, I will make backports.


StefanRijnhart:
Seems like a good idea! I'd appreciate a note in fetchmail_incoming_log's README to point out that the tests are in this separate module.

StefanRijnhart (approved):
Thanks!

PR #2509, opened 2022-12-21
[IMP] auditlog: make the line views using a non auto model

etobella:
The change applied on #2162 multiplies de size of the table just to make some views. with the proposed change, we have the same functionality without changing the table. WDYT @kittiu ?


auditlog/models/log.py
```diff
@@ -106,3 +91,75 @@ def write(self, vals):
                 {"field_name": field.name, "field_description": field.field_description}
             )
         return super().write(vals)
+
+
+class AuditlogLogLineView(models.Model):
```
legalsylvain: Could you move the new class in a distinct file ?

auditlog/models/log.py
```diff
@@ -106,3 +91,75 @@ def write(self, vals):
...
+    _name = "auditlog.log.line.view"
```
legalsylvain: If you both use name and inherits, it could avoir to redefine all the fields, that all comes from log and log line models, dont you think ?

legalsylvain (commented):
Thanks for your contribution. 

Some remarks inline.

General question : did you faced performance issue on huge databases ?

etobella:
@legalsylvain Changes applied. The thing is that the size of this table is being too large by adding some fields that are not really necessary there. For example, before the change, I had a table with about 3Gb. With the change it is 13Gb :sob:

etobella:
Also, it took too much time to apply the changes, but it was improved with the migration script. With the change it is no longer needed

pedrobaeza (approved):

etobella:
@kittiu as original developer of the change, May I ask for your review? :pray:

sanchonuria (approved):

kittiu:
Thanks @etobella for the improvement. I agree with it.

kittiu (approved):

PR #2515, opened 2023-01-03
[ADD] new module base_selection

hailangvn:
Native Odoo field Selection does not allow removing items. This module provides a base model and an abstract model to support CRUD when working with selection items.

It is not useful by itself. You can see an example of implementation in the 'report_product_label' module. (stock-logistics-workflow repository).

### Result
image
image
image


base_selection/models/base_selection.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2022 Camptocamp SA, Trobz
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+
+from odoo import fields, models
+
+
+class Selection(models.Model):
```
legalsylvain:
```suggestion
class BaseSelection(models.Model):
```

base_selection/models/base_selection.py
```diff
@@ -0,0 +1,42 @@
...
+    _name = "base.selection"
+    _description = "Store values in db for more flexible selections"
```
legalsylvain: Thar doesnt describe the field imo.

base_selection/models/base_selection.py
```diff
@@ -0,0 +1,42 @@
...
+    _order = "field_id, sequence, id"
+    field_id = fields.Many2one("ir.model.fields")
+    code = fields.Char(translate=False, index=True, required=True)
```
legalsylvain:
```suggestion
    code = fields.Char(index=True, required=True)
```

legalsylvain (commented):
Partial review. Interesting module.

legalsylvain:
Général question : did you tried to implement some new field argument "selection_remove" (like selection_add) ?

hailangvn:
> Général question : did you tried to implement some new field argument "selection_remove" (like selection_add) ?

Thanks for your suggestions. They are applied. About the general question, looking at [the sample implementation](https://github.com/OCA/stock-logistics-workflow/pull/1177/files#diff-2317bc6a44c12bf42bfd6f1d300175e8245cc7bd96dd2059d6b9502a4a1a8e3fR15), I believe the update or delete will be taken care by `ondelete` of [Many2one field](https://www.odoo.com/documentation/15.0/developer/reference/backend/orm.html#odoo.fields.Many2one). After adding new option to the list, it appears on UI without any coding.
image
image

Similarly, remove an option (i.e. `4 x 7`) from the list, it disappears as well. Archiving and unarchiving options make them updated in UI.
image
image

PR #2518, opened 2023-01-13
[13.0][FIX] pre-commit; letsencrypt tests; module_analysis tests; nsca_client tests

StefanRijnhart:
* Reinitialization of pre-commit using copier, from https://github.com/OCA/server-tools/pull/2483 by @adrienpeiffer 
* Incremental update of pre-commit configuration
* Backport of https://github.com/OCA/server-tools/pull/2433 to fix `letsencrypt` tests
* Fix `module_analysis` tests by not expecting an analysis for addons installed as Python packages
* Fix `nsca_client` tests by installing Ubuntu package `nsca-client`

PR #2520, opened 2023-01-15
[13.0][FIX] nsca client: external dependencies

StefanRijnhart:
Follow-up from https://github.com/OCA/server-tools/pull/2518, using the proper fix from https://github.com/OCA/server-tools/pull/2459 by @sbidoul


sbidoul:
Thanks Stefan!

/ocabot merge patch

PR #2521, opened 2023-01-15
[14.0][IMP] base_time_parameter: type reference -> type record

ghost:
When the value should be a record:

* Select `Type` = Record
* Select `Record Model`
  - If empty, for each value select a model and a record (`value_reference`)
  - If Account, for each value select an account (`value_account_id`, see the [account_time_parameter](https://github.com/appstogrow/apps/tree/14.0/account_time_parameter) module)

### DIFF
--- a/base_time_parameter/__manifest__.py
+++ b/base_time_parameter/__manifest__.py
@@ -21,6 +21,7 @@
         "security/security.xml",
         "security/ir.model.access.csv",
         "views/base_time_parameter_views.xml",
+        "views/base_time_parameter_version_views.xml",
     ],
     "development_status": "Beta",
     "installable": True,
--- a/base_time_parameter/migrations/14.0.4.0.0/pre-migration.py
+++ b/base_time_parameter/migrations/14.0.4.0.0/pre-migration.py
@@ -1,0 +1,13 @@
+from openupgradelib import openupgrade
+
+
+def type_reference2record(cr):
+    openupgrade.logged_query(
+        cr, "UPDATE base_time_parameter SET type = 'record' WHERE type = 'reference';"
+    )
+
+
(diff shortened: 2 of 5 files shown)


ghost:
I wonder if we should change the model name to e.g. `res.time.parameter` or just `time.parameter`. It is a bit strange to work with `account_time_parameter` and similar modules and do all changes in base.time.parameter model.

nimarosa (approved):
Changes look okay to me.

@appstogrow i suggest we don't change the name of the model again, since I had a lot of headaches in production databases when we changed from rule_parameters to this new module. 

What is the thing you perceive odd when working in account_time_parameter?

norlinhenrik:
@nimarosa Are you using type reference? Then potentially this update could cause you problems, just like when we changed the model name. I think the solution to both scenarios is to **update base_time_parameter in the database the first time** the database is exposed to the code update.

I just checked the non-transient models in my production database.
61 model names start with ir. (internal resource)
22 model names start with res. (resource)
3 model names start with base. (base.time.parameter, base.time.parameter.value & another model which I have made.)
Since starting with base. is not common, _I suggest changing the model name when migrating to 15.0._
I would prefer "time.parameter" or "res.time.parameter".

I came back to this PR now because of a small bug that I introduced in the beginning and now I fixed it. When a parameter value exists, always return it.

norlinhenrik:
@nimarosa Would you like to merge this? (reference -> record, and always return value)

The failing tests are not related to base_time_parameter.

nimarosa:
Lets rebase to see if tests pass now.

/ocabot rebase

nimarosa:
@appstogrow i think we have to wait till tests fixed. Will see tomorrow if I can make a PR but not sure which module is failing.

PR #2522, opened 2023-01-15
[12.0][FIX] excel_import_excel: fix adding new row in excel for pre-defined row and add extend attribute

AungKoKoLin1997:
Before this commit, if we export excel with multiple rows (eg.sale order line) that has pre-defined row with attributes in excel file(eg. line rule and font),  the first output row will only affect the attributes and the attributes of the rest is none because the system insert new rows after added the first row.
After this commit, the above problem will be solved and inserting new row will work when we enable extend attribute (reference from  v13.0).

Please see the reference.
This is the default excel file for exporting.
[sale_order.xlsx](https://github.com/OCA/server-tools/files/10420366/sale_order.xlsx)

This is the export excel before this commit.
[SO007(8).xlsx](https://github.com/OCA/server-tools/files/10420368/SO007.8.xlsx)

This is the export excel after commit.
[SO007(20).xlsx](https://github.com/OCA/server-tools/files/10420371/SO007.20.xlsx)

@qrtl

### DIFF
--- a/excel_import_export/models/xlsx_export.py
+++ b/excel_import_export/models/xlsx_export.py
@@ -50,6 +50,7 @@ class XLSXExport(models.AbstractModel):
         """
         line_field, max_row = co.get_line_max(line_field)
         line_field = line_field.replace('_CONT_', '')  # Remove _CONT_ if any
+        line_field = line_field.replace('_EXTEND_', '')  # Remove _EXTEND_ if any
         lines = record[line_field]
         if max_row > 0 and len(lines) > max_row:
             raise Exception(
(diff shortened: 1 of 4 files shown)


AungKoKoLin1997:
@kittiu Can you please review this PR?

kanda999:
Function review: it works as expected.
What is this Travis error?

AungKoKoLin1997:
> What is this Travis error?

@kanda999 I don't know what is the reason why travis failed? I think we need to change CI from travis to github in 12.0 branch.

yostashiro:
> I think we need to change CI from travis to github in 12.0 branch.

@AungKoKoLin1997 Please work on this yourself if you can.  Expect to have several hiccups in the process.  See how it went for 13.0 here: https://github.com/OCA/server-tools/pull/2518

thomaspaulb:
@AungKoKoLin1997 Maybe rebase this as well

AungKoKoLin1997:
@thomaspaulb Done!

kanda999 (approved):

PR #2533, opened 2023-01-30
[16.0][MIG] base_sparse_field_list_support

glitchov:
Migrated unchanged


base_sparse_field_list_support/models/fields.py
```diff
@@ -0,0 +1,17 @@
+import json
+
+from odoo import fields
```
lmignon:
```suggestion
from odoo.addons.base_sparse_field.models import fields
```

lmignon (approved):
LGTM, A little change to be more precise about the origin of the field definition...

lmignon (approved):

CasVissers-360ERP (approved):

sbejaoui (approved):

CasVissers-360ERP (approved):

PR #2536, opened 2023-02-02
[IMP] Subscribe rule on creation.

amh-mw:
When creating an `auditlog.rule` from XML, I found myself unable to subscribe the rule as part of its creation.


amh-mw:
Marking as draft as I'm still testing this locally and haven't added any unit test coverage.

PR #2540, opened 2023-02-04
[FIX] Double requirement given

cubells:
If I execute in this repo

`pip install -r requirements.txt` 

I've got this error:

<pre><span style="color:#C01C28">ERROR: Double requirement given: cryptography==2.6.1 (from -r requirements.txt (line 5)) (already in cryptography (from -r requirements.txt (line 4)), name=&apos;cryptography&apos;)</span></pre>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 # generated from manifests external_dependencies
 acme<2.0.0
 astor
-cryptography
 cryptography==2.6.1
 dataclasses
 dnspython


pedrobaeza (approved):

pedrobaeza:
But I think this could be because a manifest contains the external dependency without version and another with the version. All manifests should be assigned.

cubells:
ok @pedrobaeza ,so I close this and I'll do the work manually.

legalsylvain:
No I think something should be fixed. But as Said by predo, manifest should be fixed. Not requirements file that is autogenerated.

PR #2545, opened 2023-02-09
[15.0][ADD] base_sequence

Saran440:
This module refactor the `ir.sequence` allowing more legends to be added with minimal code duplication and
adding `preview` fields on both `ir.sequence` and `ir.sequence.date_range` models.

- Add `qoy` for sequence with quarter of the year
- Add `range_period_month` for sequence with period
Example
date range sequence 2022-10-01 to 2023-09-30
Current month is 2023-02-01 sequence with range_period_month should be 05

This module move from https://github.com/OCA/l10n-thailand to here and I think this module generic for other country use it.


Saran440:
This case for special Thailand.
We move it back to https://github.com/OCA/l10n-thailand/pull/369

PR #2550, opened 2023-02-10
[16.0] session_db: gevent and thread support

sbidoul:
Some concurrency issue were discovered in the gevent process under load.
So while I was at it I added thread mode support.


adrienpeiffer (approved):
Code review and (deep) functional testing

PR #2557, opened 2023-02-20
[13.0][IMP] base_view_inheritance_extension: Add new options domain_add and text_add

carlosdauden:
@Tecnativa 
ping @sergio-teruel @chienandalu


sergio-teruel (approved):

pedrobaeza:
Please add tests for the new features.

stefan-tecnativa:
TT43770

This commit adds test coverage to the new options domain_add and text_add. It also corrects a deficiency of a method and adapts and modify how the tests are done for adjusting it to interface elements behaviour. Note that the operations of the tests are not following a functional logic, just are done for testing the methods. A summary of what is changed:

- In `inheritance_handler_attributes_list_add` method, check if the list was empty before adding the value. Previously if the list was empty, the new_value was added with a ',' in front.

- Inside `inheritance_handler_attributes_python_dict` when evaluating `if key_idx` expresion, if key_idx equals 0 it is not evaluated as if key_idx exists and then instead of updating the attribute of that position, it is added. By returning -1 is the attribute is not present and changing the if condition we ensure that the value is update correctly when the attribute is in first position.

- Mainly the tests were performing the methods by code, but performing them by adding attributes to the views makes more sense as we try to replicate how odoo works by interface, and respecting the process that shows you the results. The inherited view is disable by default for not interacting with the user that installs the module, it is only enable in tests.

@carlosdauden please review!

pedrobaeza (approved):
/ocabot merge major

PR #2558, opened 2023-02-21
[14.0][ADD] force_https_redirection: perform http -> https redirections

petrus-v:


petrus-v:
I expect test to fails, I'm not sure how set server_wide_modules that is required while launching test

PR #2564, opened 2023-02-24
[14.0][FIX] excel_import_export, make new xml_id not deletable on update process

ChrisOForgeFlow:
@ForgeFlow

cc. @mariadforgeflow

### DIFF
--- a/excel_import_export/models/xlsx_import.py
+++ b/excel_import_export/models/xlsx_import.py
@@ -50,7 +50,7 @@ class XLSXImport(models.AbstractModel):
             ModelData.create(
                 {
                     "name": "{}_{}".format(record._table, record.id),
-                    "module": "excel_import_export",
+                    "module": "__excel_import_export__",
                     "model": record._name,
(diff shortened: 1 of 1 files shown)


excel_import_export/models/xlsx_import.py
```diff
@@ -50,7 +50,7 @@ def get_external_id(self, record):
             ModelData.create(
                 {
                     "name": "{}_{}".format(record._table, record.id),
-                    "module": "excel_import_export",
+                    "module": "__export__",
```
yostashiro: `__excel_import_export__` might be better, borrowing the idea from https://github.com/OCA/server-tools/pull/1948#issuecomment-1376286339?

ChrisOForgeFlow: You are right, is better set more specific module name, thank you

yostashiro (approved):

mariadforgeflow (approved):

PR #2572, opened 2023-03-06
[ADD] field name

gfcapalbo:
Adds a parameter to jsonifier with_fieldnames.
THis will allow jsonifier to return a different type of result: instead of : 
`       {'product_qty' : 18}`
we will have:
`    {   product_qty : { 'value':18, 'fieldname': 'Product Quantity'} }`

               
Very useful to have , with one RPC call , the translated value of the field we are displaying, 100% the same as odoo backend.

### DIFF
--- a/jsonifier/models/models.py
+++ b/jsonifier/models/models.py
@@ -160,16 +160,13 @@ class Base(models.AbstractModel):
              'field_name:json_key'
         ]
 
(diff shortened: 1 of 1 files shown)


thomaspaulb (changes requested):
Please resolve conflicts :-)

gfcapalbo:
Closing , replaced by https://github.com/OCA/server-tools/pull/2668

PR #2574, opened 2023-03-07
[16.0][MIG] fetchmail_notify_error_to_sender: Migration to 16.0

dsolanki-initos:


fetchmail_notify_error_to_sender/models/mail_thread.py
```diff
@@ -21,7 +21,7 @@ def message_route(
                 thread_id=thread_id,
                 custom_values=custom_values,
             )
-        except ValueError as ve:
+        except Exception as ve:
```
MiquelRForgeFlow: why? https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_thread.py#L1118

dsolanki-initos: Before I had tested it, It wasn't going inside the ValueError. To Catch all the exceptions I added the Exception. I've changed it to ValueError. Please review it. 
Thank you.

CasVissers-360ERP (approved):
Functional review

MiquelRForgeFlow (approved):

wpichler:
I've tested it - and found a bug. 
https://github.com/initOS/server-tools/blob/ac196d1eb051a2dadf45fb97bc2310b7ce71e04f/fetchmail_notify_error_to_sender/models/mail_thread.py#L35

You do catch the ValueError Exception - then you use the template.send_mail function. This functions is declared with force_send=False - so it will queue the mail for sending - and will not send immediatly. After this you raise the original exception - this will rollback the transaction - and the mail will never go out.

So - add force_send=True or do not raise the exception here.

I would prefer to not raise the exception - because the exception got handled - so why raise it here ?

ventor-dev:
Do you need any help in migrating it?

NICO-SOLUTIONS:
As there is another PR... #2681
What is the status here?

yboujraf:
Dear @MiquelRForgeFlow , @CasVissers-360ERP 

Any approval to get the migration for 16.0 ?

Best Regards

LoisRForgeFlow:
Fixes pending here seem that have been attended in https://github.com/OCA/server-tools/pull/3330. Closing in favor of it.

PR #2580, opened 2023-03-10
[IMP] sentry: allows to use server_environment config result to setup…

petrus-v:
… sentry


petrus-v:
I'm closing this issue because of https://github.com/OCA/server-env/pull/159 seems a better approch

PR #2585, opened 2023-03-17
tracking_manager: fix access right issue if sudo have been used

sebastienbeau:
If a record have been modify using a sudo.
We may not have the access right to check the value to track so we should use sudo

@kev-roche


Kev-Roche (approved):
Lgtm

sebastienbeau:
@bealdav

gbrito (approved):

PR #2586, opened 2023-03-17
[FIX] sentry: change with_locals to include_local_variables variable

deivislaya:
Currently, version 1.17.0 of sentry_sdk is causing the following error:

SentryOption("with_locals", DEFAULT_OPTIONS["with_locals"], None), KeyError: 'with_locals'.

Where the with_locals key is not found in the dictionary, generating an error, stopping the installation of the sentry module.

In version 1.17.0 rename 'with_locals'  to 'include_local_variables' https://github.com/getsentry/sentry-python/commit/79e33169aa629ec67cf9636b8440f64bf0a6d566

This commit adjust the  get_sentry_options() method in https://github.com/Vauxoo/server-tools/blob/14.0/sentry/const.py file, set the new variable.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,5 +10,5 @@ odoorpc
 openpyxl
 openupgradelib
 pysftp
-sentry_sdk
+sentry_sdk>=1.17.0
 unidecode
--- a/sentry/__manifest__.py
+++ b/sentry/__manifest__.py
@@ -17,7 +17,7 @@
     "installable": True,
     "external_dependencies": {
         "python": [
-            "sentry_sdk",
+            "sentry_sdk>=1.17.0",
         ]
     },
     "depends": [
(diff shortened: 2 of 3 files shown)


luisg123v (changes requested):
Please include a commit prior to yours where you apply pre-commit, so this doesn't fail.

You should name as follows:

`[IMP] sentry: pre-commit stuff`

luisg123v (approved):
LGTM :+1: 

@moylop260 or @fernandahf,

Could you merge, please?

CC @randall-vx

fernandahf (approved):

luisg123v:
@fernandahf hold

luisg123v:
Nevermind, I was thinking we could support both versions with something like:

`DEFAULT_OPTIONS.get("include_local_variables", DEFAULT_OPTIONS["with_locals"])`

But that's not possible because first param for `SentryOption` also changed, so I think better supporting only the latest one.

moylop260:
if we are not supporting the last version of sentry, so, you need to use in the requirements.txt the minimal version to work

I mean, something like

pkg >= VERSION

luisg123v (approved):

luisg123v:
@deivislaya could you forward-port this change to 15.0 and 16.0, please?

deivislaya:
@luisg123v 

forward-port 15.0 https://github.com/OCA/server-tools/pull/2587
forward-port 16.0 https://github.com/OCA/server-tools/pull/2588

Regards

PR #2587, opened 2023-03-18
[FIX] sentry: change with_locals to include_local_variables variable

deivislaya:
Currently, version 1.17.0 of sentry_sdk is causing the following error:

SentryOption("with_locals", DEFAULT_OPTIONS["with_locals"], None), KeyError: 'with_locals'.

Where the with_locals key is not found in the dictionary, generating an error, stopping the installation of the sentry module.

In version 1.17.0 rename 'with_locals'  to 'include_local_variables' https://github.com/getsentry/sentry-python/commit/79e33169aa629ec67cf9636b8440f64bf0a6d566

This commit adjust the  get_sentry_options() method in https://github.com/Vauxoo/server-tools/blob/15.0/sentry/const.py file, set the new variable.

related #2586

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,4 +13,4 @@ openpyxl
 openupgradelib
 pygount
 pysftp
-sentry_sdk
+sentry_sdk>=1.17.0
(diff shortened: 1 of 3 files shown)


moylop260 (approved):
Based on approved PR

https://github.com/OCA/server-tools/pull/2586

PR #2335, opened 2022-05-10
[12.0][FIX] auditlog: Allow passing a chunk size for autovacuum

hbrunn:
this allows administrators to fine tune the amount of records to delete in one run of the autovacuum.

I also think we can get away with not recomputing after deletion, as we delete all relevant records eventually.

I'll be happy to cherry pick this to 13, 14, 15 when merged

PR #2591, opened 2023-03-22
[13.0][FIX] auditlog: Allow passing a chunk size for autovacuum

BT-tbaechle:
(cherry picked from commit d4f8eb345acfb60d7ff28180614340944c03225e)
The commit was already cherry-picked into all branches v14+, only v13 is missing.

Original PR in v12 #2335 

This allows administrators to fine tune the amount of records to delete in one run of the autovacuum.

### DIFF
--- a/auditlog/models/autovacuum.py
+++ b/auditlog/models/autovacuum.py
@@ -27,7 +27,9 @@ class AuditlogAutovacuum(models.TransientModel):
         for data_model in data_models:
             records = self.env[data_model].search(
                 [("create_date", "<=", fields.Datetime.to_string(deadline))],
-                limit=chunk_size, order="create_date asc")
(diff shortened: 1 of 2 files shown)


BT-dmoreno (approved):
Looks good to me :+1: , just wondering about the use of `norecompute` since I see it's marked as deprecated in the method definition.

BT-tbaechle:
> Looks good to me +1 , just wondering about the use of `norecompute` since I see it's marked as deprecated in the method definition.

This commit is already migrated to all versions 14+ and all still include the `norecompute` so assumed it would be fine to keep it.

BT-mgomez (approved):
LGTM

BT-tbaechle:
Hi @hbrunn do you think this can be merged?

PR #2593, opened 2023-03-23
[13.0] session_db: reconnect if needed

sbidoul:
If the connection to the database fails when
retrying a session operation, we end up
with no cursore, which makes subsequent
session operations fail.

We fix this by ensuring we have cursor
before any operations.

### DIFF
--- a/session_db/pg_session_store.py
+++ b/session_db/pg_session_store.py
@@ -47,6 +47,7 @@ def with_cursor(func):
         while True:
             tries += 1
             try:
+                self._ensure_connection()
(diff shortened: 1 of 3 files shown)


adrienpeiffer (approved):
Thanks ! It works like a charm !

sbidoul:
@adrienpeiffer I added tests, which led me to a little additional tweak. Can you try this version?

adrienpeiffer:
still working fine :+1:

sbidoul:
Good. So merging and I'll forward port.

/ocabot merge patch

PR #2595, opened 2023-03-25
[15.0][FIX] database_cleanup: 'Environment' object has no attribute 'invalidate_all'

hhgabelgaard:
Replaced by invalidate_cache on the model, which end up calling env.cache.invalidate()


pedrobaeza:
Please change the commit message to `[FIX] database_cleanup: 'Environment' object has no attribute 'invalidate_all'` (with the tag without space and the module name after).

fuentes73 (approved):

fuentes73:
@pedrobaeza can be merged?

PR #2598, opened 2023-03-28
[ADD] base_sequence_default: add default values with sequences

yajo:
This module extends the defaults functionality to allow filling *Character*
type fields on creation with a predefined sequence.

@moduon MT-1506


fcvalgar (approved):
LGTM, Thank you @yajo

Shide (approved):
LGTM code

PR #2601, opened 2023-04-04
[16.0][MIG] base_multi_image: Migration to 16.0

clb-openfire:


clb-openfire:
I changed my mind, too many things are wrong, I think the module needs a rework:
- restrict the folders from which files can be read for security reasons, through the odoo config file
- fix hooks
- remove "database" type of storage, which was actually storing in filestore since v13 anyway (I don't think that users can or should decide if files are stored in DB or in filestore)
- remove "url" and "file" types of storage. image.mixin class stores all versions of images for performance reasons. I believe it would be even slower to get them from a file/remote url before resizing them. We should as well store everything in filestore. And by the way, restrict the size of those images, like it is done in base_import module, _import_image_by_url()
- simplify the filestore type to make it work like the current database type.
- adapt module to image.mixin (remove image_main, image_main_medium, etc. from owner)

I've done all that work in a fork, I don't know the correct way to share though

PR #2602, opened 2023-04-07
[16.0][MIG] scheduler_error_mailer

hbrunn:


CasVissers-360ERP (approved):
Functional review

kikopeiro (approved):
:+1: LGTM (without code review)

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @api.model
+    def _handle_callback_exception(
+        self, cron_name, server_action_id, job_id, job_exception
+    ):
+        res = super()._handle_callback_exception(
+            cron_name, server_action_id, job_id, job_exception
+        )
+        my_cron = self.browse(job_id)
+
```
jcdrubay, line 36: I believe that people are installing this module if they want an email to be sent when there is an error in the cron. 

But from my understanding when reading the code, the module doesn't do anything when installed.

The user needs to manually set a mail template in the ir.cron records. And then will need to remember doing so for every new Scheduled Actions.

Should we change the logic a bit?

### Suggestion

With an ir.config.parameter , set which mail template should be used by default. Something like this:

key: scheduler_error_mailer_default_mail_template
value: scheduler_error_mailer.scheduler_error_mailer

In this file, adjust like this:

```python

cron_error_mail_template = my_cron.email_template_id 

if not cron_error_mail_template:
    default_mail_template = tools.config.get("scheduler_error_mailer_default_mail_template")
    if default_mail_template:
        cron_error_mail_template = self.env.ref("default_mail_template")
)

if cron_error_mail_template:
    ...  # the rest would be the same, only replacing my_cron.email_template_id by cron_error_mail_template
```

### One step further

We could add a field on the ir.cron: `disable_scheduler_error_mailer` if some cron should be ignored.

jcdrubay: ### One more step further

Put the default email template in General Config which is a bit more user friendly.

CasVissers-360ERP: I agree with this approach.

jcdrubay: @hbrunn Let me know your opinion and if you would like me to submit a PR to your branch or to branch 15.0 of this repo (which could later on be forward ported). Thanks :)

hbrunn: note this is a draft, so I don't consider this ready for primetime, ie tests should be improved.

Why so complicated? I planned to make the default for the field the template that comes with the module, and have a post-install script that sets the default template on all existing crons. With this, you can configure each cron however you like, but by default you'll get a mail

jcdrubay: Hi @hbrunn I think that your suggestion is also good (or better) and achieves the same main objective :) Thanks.

jcdrubay (approved):
Code review looks good to me.

Thanks @hbrunn :)

liebana:
Functional review LGTM. Can you merge @moylop260 , please?

moylop260:
Could you rebase it, please?

PR #2605, opened 2023-04-14
[16.0][MIG] configuration_helper: Migration to 16.0

lehoangan:


configuration_helper/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2014 David BEAL Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Configuration Helper",
+    "version": "16.0.1.0.0",
```
simahawk, line 5: there's no summary here, is it correct?

simahawk (commented):
Curious: how many modules rely on this? Is it worth to keep maintaining it?

configuration_helper/__manifest__.py
lehoangan: I added summary

nilshamerlinck:
Hello @simahawk

> Curious: how many modules rely on this? Is it worth to keep maintaining it?

`delivery_carrier_label_postlogistics` up to 12.0 (now replaced by `delivery_postlogistics`), `l10n_ch_delivery_carrier_label_quickpac` up to 14.0, none anymore in 16.0.

Probably not worth indeed.

PR #2607, opened 2023-04-17
[ADD][14.0] slow_wsgi_logger: log slow wsgi request

petrus-v:
This module lets you log slow WSGI requests. It is useful to analysis
performance issue given more context regarding the request and
only log if processed time is higher to the defined limit.


damdam-s (approved):
🚀 🚀 LGTM

Koudja (approved):

thomaspaulb:
@petrus-v Tests are still failing

slow_wsgi_logger/tests/test_slow_wsgi_logger.py
```diff
@@ -0,0 +1,190 @@
...
+        wait = self.min_duration + 0.1
+        with self.assertLogs("odoo.addons.slow_wsgi_logger", level=self.level) as cm:
+            self._json_call(
+                self.db_name,
+                self.admin_uid,
+                "admin",
+                "fake.test",
+                "name_search",
+                wait,
+            )
```
thomaspaulb, line 164: Test is [failing](https://github.com/OCA/server-tools/actions/runs/8802245846/job/24157487806#step:8:1891). Not quite sure what to make of it, to my best interpretation skills the `fake.test` model is not properly loaded, hence test fails and Odoo attempts to restore the savepoint but can't find it and the cursor is reset to an unusable state. The next query that has to do with registry restore, then also fails to execute.

PR #2608, opened 2023-04-18
[16][MIG ] module_analysis

florian-dacosta:


legalsylvain (commented):
Why https://github.com/OCA/server-tools/pull/2608/commits/8e5457e805d46c7e65fd9eb2a57189cb504660a6
and https://github.com/OCA/server-tools/pull/2608/commits/796e9cab879c8cf739db9ea660c28fb8b4b7d412 ? 
otherwise LGTM. Thanks for porting this module.

florian-dacosta:
> Why [8e5457e](https://github.com/OCA/server-tools/commit/8e5457e805d46c7e65fd9eb2a57189cb504660a6)

In general I prefer all cron to be unactive at installation, mainly this kind of cron that you have to do some configuration (better change the time to run it at night at least, etc). It does not change a lot to unactivate it because you have to configure it anyway.
Having active cron of the kind is annoying for test database, CI, preprod, etc...



> and [796e9ca](https://github.com/OCA/server-tools/commit/796e9cab879c8cf739db9ea660c28fb8b4b7d412) ?

I went a bit too fast with the cherry-pick, I guess the second line could stay the same.
For the first line, I guess we need to launch a `cron_analyse_code` because it is not done during the installation anymore (removed post_init_hook)

PR #2609, opened 2023-04-18
[16][MIG ] module_analysis

florian-dacosta:
Replace https://github.com/OCA/server-tools/pull/2608

Sorry I guess I did a wrong manipulation, I could not reopen the other after a force push.

@legalsylvain


florian-dacosta:
@legalsylvain 
>>Why [8e5457e](https://github.com/OCA/server-tools/commit/8e5457e805d46c7e65fd9eb2a57189cb504660a6)

>In general I prefer all cron to be unactive at installation, mainly this kind of cron that you have to do some configuration (better change the time to run it at night at least, etc). It does not change a lot to unactivate it because you have to configure it anyway.
Having active cron of the kind is annoying for test database, CI, preprod, etc...

You'll tell me if it is acceptable for you, else I'll remove the commit so we can go forward

legalsylvain:
I have no clear point of view. 
In one hand, you removed the analysis in the post_install part, and you set active to false for the cron. So when we install the module, now, nothing "work", (and will work) and the configuration is now mandatory.
In the other hand, your argument are valid. ;-)

So, it's as you wish on that topic ! 
But if we go in that direction, you should rewrite some readme files, because some part are obsolete / wrong now.

florian-dacosta:
Thanks @legalsylvain I always forget to check the readme files...
I've updated it the readme file and I also added the possibility to add a custom domain in the cron, because we have a use case where we'd like the analysis to run on uninstalled modules. (I will allow to know the "cost" of your module before installing it...)

module_analysis/readme/CONFIGURE.rst
```diff
@@ -16,9 +16,9 @@ developped for exemple,
 
 to update the data manually, you have to :
 
-* Go to 'Apps' / 'Update Apps List'
+* Go to 'Settings' / 'Technical' / 'Scheduled Actions'
 
-* Check the box 'Analyse Installed modules'
+* Run manually the action : 'Update Module Analysis'
 
     .. image:: ../static/description/base_module_update.png
```
legalsylvain: I guess that line has to be removed. (and the according png file).

legalsylvain:
> (I will allow to know the "cost" of your module before installing it...)

good idea !

module_analysis/readme/CONFIGURE.rst
florian-dacosta: I've added the line => To run an update manually, you can run the scheduled action manually.
The image is indeed to remove

PierrickBrun (approved):
looks good, I tested it in a customer project and the analysis worked well

florian-dacosta:
@legalsylvain 
Would you mind updating your review ?

legalsylvain (approved):
Thanks ! 

/ocabot merge nobump

bguillot:
Thanks !

PR #2612, opened 2023-04-20
[14.0][FIX] *: remove `lxml` from python depends

luisg123v:
Such dependency is already included in Odoo requirements using a pinned
version [1]. Adding here could cause to upgrade the library to an
incompatible version.

[1] https://github.com/odoo/odoo/blob/54e58b3e47ee/requirements.txt#L23

### DIFF
--- a/html_image_url_extractor/__manifest__.py
+++ b/html_image_url_extractor/__manifest__.py
@@ -12,6 +12,5 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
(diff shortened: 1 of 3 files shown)


luisg123v:
Hi @moylop260,

Could you review, please?

CC @randall-vx @desdelinux

moylop260 (approved):
👍

luisg123v:
Hi @pedrobaeza, what do you think about this one?

pedrobaeza (approved):
/ocabot merge patch

PR #2615, opened 2023-04-24
[16.0][ADD] base_partition

glitchov:
Migrtion from https://github.com/OCA/server-tools/pull/1893.

This module adds a `partition(self, accessor)` method to every model.
It accepts for accessor any parameter that would be accepted by `mapped`,
i.e. a string `"field(.subfield)*"` or a function `(lambda x: not x.b)`.
It returns a dictionary with keys that are equal to `set(record.mapped(accessor))`,
and with values that are recordsets.


rvalyi:
are you sure AGPL is the best fit for such a module? It seems it's a small utility module that could have a broad usage and LGPL might fit better. If such module is AGPL, people not playing the AGPL game could easily just copy the feature instead of adopting the module and help to debug and maintain it... But of course this decision is up to you the authors.

base_partition/models/models.py
```diff
@@ -0,0 +1,73 @@
+# © 2020 Acsone (http://www.acsone.eu)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+LIKE_COMPARATORS = (
```
rousseldenis: I think you can remove this

rousseldenis (approved):
Code review. Seems great. Minor comment

Agree with @rvalyi. Better to move to LGPL

rousseldenis:
@glitchov

rousseldenis:
This is a better implementation of https://github.com/odoo/odoo/blob/15.0/odoo/tools/misc.py#L260

lmignon (approved):
LGTM (Code review)

A very useful addon heavily used in Odoo 10....

len-foss (approved):

PR #2618, opened 2023-04-26
[16.0][MIG] base_multi_image

yibudak:


clb-openfire:
Hi,
To me this module has several problems that should be fixed (mentioned in #2601 ):
- restrict the folders from which files can be read for security reasons, through the odoo config file.
- fix hooks
- remove "database" type of storage, which was actually storing in filestore since v13 anyway (I don't think that users can or should decide if files are stored in DB or in filestore)
- remove "url" and "file" types of storage. image.mixin class stores all versions of images for performance reasons. I believe it would be even slower to get them from a file/remote url before resizing them. We should as well store everything in filestore. And by the way, restrict the size of those images, like it is done in base_import module, _import_image_by_url()
- simplify the filestore type to make it work like the current database type.
- adapt module to image.mixin (remove image_main, image_main_medium, etc. from owner)

I've done all that work already in the following fork if anyone interested: https://github.com/odof/server-tools/commits/16.0-base_multi_image

thomaspaulb:
Superseded by https://github.com/OCA/server-tools/pull/3010 as well.

@clb-openfire We're going to try integrating your code, it looks like a lot of improvements.

Maybe you can give it a review once it's done?

Next time you can also just propose it yourself in a new PR.

clb-openfire:
@thomaspaulb thanks, review done.
I wasn't confident enough to post a PR because a migration + rework seemed a lot to review, I'll be sure to do it next time.

PR #2620, opened 2023-05-11
[15.0][MIG] base_model_restrict_update

Saran440:
No change code


StefanRijnhart:
Thanks! Can you include https://github.com/OCA/server-tools/pull/2626?

Saran440:
@StefanRijnhart Thanks for review. This PR add commit https://github.com/OCA/server-tools/pull/2626
and 1 commit for fix error

Peek 2023-06-15 15-35

base_model_restrict_update/models/res_users.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+    @api.constrains("is_readonly_user")
+    def _check_is_readonly_user(self):
+        for user in self:
+            if self.env.ref("base.group_system") in user.groups_id:
+                raise UserError(_("You cannot set admin user as a readonly user."))
+
+    def toggle_unrestrict_model_update(self):
+        pass
```
StefanRijnhart: If it is necessary to add this code override to make the GUI widget work (and I played with it a bit, and I don't see another way myself), maybe Odoo is telling us we should not be doing this 'boolean field in a stat button' thing. Would you consider putting the binary toggles somewhere else on the user form?

yostashiro: I also checked the behavior and found that if you replace the buttons with the following divs then the presentation and behavior become OK without the dummy toggle methods:

```.xml
                <div
                    name="toggle_unrestrict_model_update"
                    class="btn-group btn oe_stat_button"
                    attrs="{'invisible': [('is_readonly_user', '=', True)]}"
                    style="align-items: center;"
                >
                    <span style="padding: 0 10px;">Unrestrict Update</span>
                    <field name="unrestrict_model_update" widget="boolean_toggle" />
                </div>
                <div
                    name="toggle_is_readonly_user"
                    class="btn-group btn oe_stat_button"
                    style="align-items: center;"
                >
                    <span style="padding: 0 10px;">Read-only</span>
                    <field name="is_readonly_user" widget="boolean_toggle" />
                </div>
```

But I agree with Stefan that we may put the toggles elsewhere if it becomes a maintenance burden to put them in the button box.

StefanRijnhart: Oh, I can live with this `div` solution!

Saran440: @StefanRijnhart @yostashiro Thanks for solution :) I fixed it.

base_model_restrict_update/views/res_users_views.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="view_users_form" model="ir.ui.view">
+        <field name="name">res.users.form</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form" />
+        <field name="arch" type="xml">
+            <div class="oe_button_box" name="button_box" position="inside">
+                <div
```
yostashiro, line 10: A little comment would be nice.

```suggestion
                <!-- "button" tag cannot be used without object/action assignment as of 15.0, therfore we have chosen to use "div" here. -->
                <div
```

yostashiro (approved):
Functional test.  Thanks for the update!

base_model_restrict_update/models/res_users.py
StefanRijnhart: Thanks! I think this all counts as migration stuff so can you squash the two last commits into the migration commits?

base_model_restrict_update/views/res_users_views.xml
StefanRijnhart: Not sure. This is the story of the road we traveled, but maybe we can let the code speak for itself here :grinning:?

base_model_restrict_update/models/res_users.py
Saran440: I think migration commit should original and rebase https://github.com/OCA/server-tools/pull/2626 and [Fix error](https://github.com/OCA/server-tools/pull/2620#discussion_r1232949173) into 1.

StefanRijnhart (approved):
Thanks for the update!

PR #2624, opened 2023-05-17
[FIX] upgrade_analysis : avoid KeyError "module"

metaminux:
Check "module" is in context when patching init_models


StefanRijnhart (approved):

legalsylvain (approved):
/ocabot merge patch

PR #2628, opened 2023-05-22
[16.0] [FIX] Multi-Company issue with M2M field

rven:
In a Multi-Company environment with the auditlog rule on product.template enabled, the taxes_id field gets always overwritten with the taxes chosen in the current company. The cause of this is that the old_values are read with sudo, which has access to not only the current company. When doing the actual write of the field, this cached value is taken into account, which removes them from the M2M relation. Before doing the actual write, the cache of the recordset(s) need to be cleared again.

The following method causes the problem in Odoo:
```python
    def write_real(self, records_commands_list, create=False):
        # records_commands_list = [(records, commands), ...]
        if not records_commands_list:
            return

        model = records_commands_list[0][0].browse()
        comodel = model.env[self.comodel_name].with_context(**self.context)
        cr = model.env.cr

        # determine old and new relation {x: ys}
        set = OrderedSet
        ids = set(rid for recs, cs in records_commands_list for rid in recs.ids)
        records = model.browse(ids)

        if self.store:
            # Using `record[self.name]` generates 2 SQL queries when the value
            # is not in cache: one that actually checks access rules for
            # records, and the other one fetching the actual data. We use
            # `self.read` instead to shortcut the first query.
            missing_ids = list(records.env.cache.get_missing_ids(records, self))
            if missing_ids:
                self.read(records.browse(missing_ids))

        # determine new relation {x: ys}
        old_relation = {record.id: set(record[self.name]._ids) for record in records}
        new_relation = {x: set(ys) for x, ys in old_relation.items()}
```

The **old_relation** field will be populated with all the taxes, not taken into account the current company.
So later on it will remove them.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -386,7 +386,7 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
-            # self.invalidate_recordset()
+            self.invalidate_recordset()
             result = write_full.origin(self, vals, **kwargs)
             new_values = {
                 d["id"]: d
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -386,6 +386,7 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            # self.invalidate_recordset()
             result = write_full.origin(self, vals, **kwargs)
             new_values = {
                 d["id"]: d
--- a/auditlog/tests/__init__.py
+++ b/auditlog/tests/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from . import test_auditlog
 from . import test_autovacuum
+from . import test_multi_company
--- a/auditlog/tests/test_multi_company.py
+++ b/auditlog/tests/test_multi_company.py
@@ -1,0 +1,194 @@
+import json
+
+from odoo.tests.common import Form, TransactionCase, new_test_user
+
+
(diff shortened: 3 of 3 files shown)


rven:
@StefanRijnhart What do you think about this solution?

You could also say that this is an Odoo bug, that Odoo does not take into account the company which has been provided in the context...

StefanRijnhart:
It does not sit entirely well with me. I'm wondering if it might hurt performance and I'm not entirely sure it would not introduce any side issues. I would be tempted to patch `self.env.cache` for my own odoo.api.Cache() instance for the duration of the `read`. But this is more a gut feeling, to be honest.

Apart from that, it would be nice if you could reproduce the error. Would adding a test with a rule on `res.partner.category`'s `partner_ids` work for that?

rven:
@StefanRijnhart
I don't see directly how to fix this issue in another way then I proposed.

I tried to add a test case on the `category_id` field on `res.partner`, but here the scenario is not going wrong.
Only in the test case `test_20_02_multi_company_restricted` when using `with_user` is acting weird.
With the `self.partner.invalidate_recordset()` enabled in the test cases everything is working fine.

Because of that I taught it had to do with a customization in our project, so I created a test scenario using the `taxes_id` field on `product.template`, and yes the problem came up here.
The next thing I did was creating a module based on only standard Odoo modules, and there the problem also happens.

Can you maybe have a look at this problem?

Here the module with all the test cases:
[auditlog_bug.zip](https://github.com/OCA/server-tools/files/11567009/auditlog_bug.zip)

StefanRijnhart:
@rven Of course it was only a matter of time before I would be hit by this myself. Thanks to your effort, I was able to reproduce the issue in #2814 and apply your fix to solve the issue.

PR #2630, opened 2023-05-23
[16.0][MIG] base_name_search_multi_lang: Migration to 16.0

therbin200:
Modules works on v16 as it was on v14. Just changed the version number.


kanda999:
@therbin200 
Can you update the README.rst for V16?

base_name_search_multi_lang/models/ir_model.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2020 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+from odoo import api, fields, models, tools
+
+
+@tools.ormcache("model")
+def _get_model_name_search_multi_lang(self, model):
+    multi_lang = self.env["ir.model"].search_read(
```
AungKoKoLin1997, line 8: The error will be raised when we try to search from normal users. Since odoo15, the access right was changed.
```suggestion
    multi_lang = self.env["ir.model"].sudo().search_read(
```

PR #2562, opened 2023-02-22
[16.0][MIG] attachment_synchronize: Migration to 16.0

kevinkhao:
depends on https://github.com/OCA/server-tools/pull/2560

PR #2632, opened 2023-05-29
[16.0][MIG] attachment_synchronize

florian-dacosta:
Depends on https://github.com/OCA/storage/pull/252

Adapt the module to the new fs_storage module which will replace storage_backend very soon.
Aims to replace https://github.com/OCA/server-tools/pull/2562 when fs_storage will be merged.

@bealdav @kevinkhao


attachment_synchronize/models/attachment_synchronize_task.py
```diff
@@ -0,0 +1,241 @@
...
+    avoid_duplicated_files = fields.Boolean(
+        string="Avoid importing duplicated files",
+        help="If checked, a file will not be imported if an Attachment Queue with the "
+        "same name already exists.",
+    )
+    failure_emails = fields.Char(
+        help="Used to fill the 'Failure Emails' field in the 'Attachments Queues' "
+        "related to this task.\nAn alert will be sent to these emails if any operation "
+        "on these Attachment Queue's file type fails.",
+    )
```
bealdav, line 97: Hi @florian-dacosta thanks for this nice integration.
Side consideration. Odoo introduced activities in 11.0 as internal communication system.
Clearly, users in odoo do not use it as much as they could do.
Flows in this module are fully managed by Odoo objects but mail is a communication mode to exchange informations between two  systems. Here we have a unique system, `Odoo`
Currently our odoo users in our projects complains they receive too many mails in a day from misc sources for so many usage.
I don't talk about environmental considerations, but we could have this argument too
I think we should promote definitively this kind of communication in a such version.
Replace here `failure_emails` here by `failure_activity_user_ids` could definitively imply users to use activities instead of emails.
It's an opportunity to improve communication, make it smarter, interactive, only few bytes consumer

#makestheplanetgreenagain

My 2 cts

cc @kevinkhao

florian-dacosta: Hi @bealdav 
Thanks for the remark. It is true we may want to implement activities.
This module is not responsible for this part though, as it just extends it, but the base implementation of these failure email is in the base module (`attachment_queue`) which is merged.

I think it would be nice to have an implementation with the activities indeed, and we could have another one for those still prefering email.
We'd just have to decide if we'd implement both in base module, or only activities and email as a separated module, or both in separated modules...

Overall I agree,  but I think it should be put on the roadmap. (I may work on it but I don't want it to block the PR)

florian-dacosta:
@bealdav @kevinkhao @bguillot 
Ready for review now that https://github.com/OCA/storage/tree/16.0/fs_storage has been merged.

florian-dacosta:
Hi @bealdav @kevinkhao @bguillot 
I would still apreciate if you could do a review on this one.
It is used in production since some weeks now with no issue.

kevinkhao (approved):
code review LGTM

attachment_synchronize/tests/common.py
```diff
@@ -2,34 +2,36 @@
...
+        path = fs.sep.join([self.directory_input, "bar.txt"])
+        with fs.open(path, "wb") as f:
+            f.write(self.filedata)
 
     def setUp(self):
         super().setUp()
-        self.env.cr.commit = mock.Mock()
-        self.registry.enter_test_mode(self.env.cr)
+        #        self.env.cr.commit = mock.Mock()
+        #        self.registry.enter_test_mode(self.env.cr)
```
clementmbr, line 32: better to delete instead of comment ;) @florian-dacosta

clementmbr (changes requested):

clementmbr (approved):
LGTM!

PR #2490, opened 2022-12-05
[14.0] Sentry APM integration

sebastienbeau:
Working progress

The aim of this PR is to integrate correctly Sentry APM feature.

- [x] generate transaction with human readable name for http
- [x] generate transaction with human readable name for cron process
- [x] track queue job correctly
- [x] integrate span for database
- [x] integrate span for external request
- [x] improve logger integration (add event_log_level and breadcrum_log_level)
- [ ] write test

Note: I have work on a previous integration with elastic APM : https://github.com/akretion/odoo-elasticapm

PR #2638, opened 2023-06-03
[FIX] sentry: downgrade sentry-sdk to compatible version

deivislaya:
Odoo requires urllib3 between 1.21.1 and 1.25.
https://github.com/odoo/odoo/blob/6ef080a4818dd0bb85aa4ef257900522d9e9fbd6/requirements.txt#L51
https://github.com/psf/requests/blob/5a1e738ea9c399c3f59977f2f98b083986d6037a/setup.py#L47

sentry-sdk > 1.9.0 required urllib3 >= 1.26.11
https://github.com/getsentry/sentry-python/blob/4f1f782fbedc9adcf1dfcd2092bb328443f09e8c/setup.py#L43

Currently, urllib3 >= 1.26.11 is causing the following error in
response.py:

``` Traceback (most recent call last):
"/home/odoo/.local/lib/python3.8/site-packages/urllib3/response.py",
line 705, in _error_catcher
    yield
  File
"/home/odoo/.local/lib/python3.8/site-packages/urllib3/response.py",
line 830, in _raw_read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
urllib3.exceptions.IncompleteRead: IncompleteRead(1501 bytes read, -827
more expected)
```

On the other hand, sentry 1.9.0 supports urllib3 >= 1.10.0, satisfying
odoo requirements.

This partially reverts
https://github.com/OCA/server-tools/commit/d7ae024951d30b9409104d0bf294f8e1cf05064f.
That was initially introduced to support newer versions of `sentry_sdk`, but won't be required anymore
due to this downgrade.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,5 +9,5 @@ odoorpc
 openpyxl
 openupgradelib
 pysftp
-sentry_sdk>=1.17.0
+sentry_sdk<=1.9.0
 unidecode
--- a/sentry/__manifest__.py
+++ b/sentry/__manifest__.py
@@ -17,7 +17,7 @@
     "installable": True,
     "external_dependencies": {
         "python": [
-            "sentry_sdk>=1.17.0",
+            "sentry_sdk<=1.9.0",
         ]
     },
     "depends": [
--- a/sentry/const.py
+++ b/sentry/const.py
@@ -77,9 +77,7 @@ def get_sentry_options():
         SentryOption("dsn", "", str.strip),
         SentryOption("transport", DEFAULT_OPTIONS["transport"], select_transport),
         SentryOption("logging_level", DEFAULT_LOG_LEVEL, get_sentry_logging),
-        SentryOption(
-            "include_local_variables", DEFAULT_OPTIONS["include_local_variables"], None
-        ),
+        SentryOption("with_locals", DEFAULT_OPTIONS["with_locals"], None),
         SentryOption(
             "max_breadcrumbs", DEFAULT_OPTIONS["max_breadcrumbs"], to_int_if_defined
         ),


deivislaya:
@luisg123v 
Please review this MR

Regards!

luisg123v (changes requested):
Hi @deivislaya,

Code LGTM, but description needs to be updated:
- Point to odoo/odoo, not the Vauxoo's fork
- Provide permalinks, i.e. links should point to a hash instead a commit, otherwise it could get outdated at any time
- Links should always go on their own lines, for readability
- Where you say "MR related...":
  * That's a PR, not a MR :smile: 
  * You should say instead something like:
    
    This partially reverts `<hash>`. That was initially introduced to support newer versions of `sentry_sdk`, but won't be required anymore due to this downgrade.
    
    The hash is preferred (IMHO) instead of the PR, because it can be checked easily from git

CC @randall-vx

deivislaya:
@luisg123v 

Please review again this MR

Regards!!

cc @randall-vx

luisg123v (approved):
LGTM :+1: 

@moylop260, @fernandahf,

Could you review, please?

CC @randall-vx

fernandahf (approved):

moylop260:
@deivislaya 

Did you check the CI error?
 - `AttributeError: 'InMemoryTransport' object has no attribute 'hub_cls'`

cc @luisg123v @fernandahf

deivislaya:
@moylop260 
 
Yes, I did
In this PR 
https://github.com/OCA/server-tools/pull/2645  
previously joined, a few hours ago. The CI Error was already present
https://github.com/OCA/server-tools/actions/runs/5179816806/jobs/9333206935#step:8:1519

Regards!

cc @luisg123v @fernandahf

StefanRijnhart:
@deivislaya do you have a complete traceback of the exception? We have been running Odoo 14 with Sentry-sdk 1.17.0 and urllib3 1.26.x for some time now, without seeing this particular error. I'd like to have a way to reproduce the error to see if our setup has this issue.

StefanRijnhart:
Ah, I see that in our setup, we are running with requests 2.28.1 precisely to match sentry-sdk 1.17.0's urllib3 version dependency. So this change is valid if you stick to Odoo's requirements, but Odoo is running fine with an upgraded version of requests as well (also, we are running Odoo 14 on Python 3.9 so it all makes sense now).

hparfr:
This PR breaks our setup too. sentry 1.17 is working fine for us (urlib 1.26.15  requests 2.28.2)

ioans73:
@deivislaya It also gives us an error with version 1.9.0 `KeyError: 'include_local_variables'`, having this PR https://github.com/OCA/server-tools/pull/2490/files#diff-1869ad05ba60dc51f66ca0d19e0fdef1f27fdcded364b7774015ac525af3fe31R93

What would be better, update the version or remove the previous line?

cc @sebastienbeau @moylop260

moylop260:
@deivislaya 

Could you check the @ioans73 comment, please?

@ioans73 

Could you share us the output of packages installed in your environment, please?

`python3 -m pip list --format=freeze`

PR #2642, opened 2023-06-05
[16.0][MIG] base_conditional_image: migration module

polqubiq:
Migrating the base_conditional_image module from v15.0 to v16.0


chusamo (approved):
LGTM, just keep de commit patch of the module from 15.0 version

PR #2645, opened 2023-06-05
[14.0][FIX] database_cleanup: preserve spaces in quoted identifiers

StefanRijnhart:
Meant to drop a column with a trailing space in its name that was created by accident, but database_cleanup dropped the column by the same name without the trailing space.


pedrobaeza (approved):

legalsylvain (approved):
Thanks !

StefanRijnhart:
Tests are fixed here: https://github.com/OCA/server-tools/pull/2647

PR #2652, opened 2023-06-06
[15.0][FIX] *: remove lxml, cryptography from python depends

EmilioSerna:
Such dependency is already included in Odoo requirements using a pinned
version [1][2]. Adding here could cause to upgrade the library to an
incompatible version.

[1] https://github.com/odoo/odoo/blob/710a2b2a7af68e8f2f249ef9fc3146f44d3266a5/requirements.txt#L18
[2] https://github.com/odoo/odoo/blob/710a2b2a7af68e8f2f249ef9fc3146f44d3266a5/requirements.txt#L3

### DIFF
--- a/html_image_url_extractor/__manifest__.py
+++ b/html_image_url_extractor/__manifest__.py
@@ -12,6 +12,5 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "external_dependencies": {"python": ["lxml"]},
     "depends": ["base"],
 }
--- a/html_text/__manifest__.py
+++ b/html_text/__manifest__.py
@@ -14,6 +14,5 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "external_dependencies": {"python": ["lxml"]},
     "depends": ["base"],
 }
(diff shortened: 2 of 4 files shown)


hugho-ad (approved):

legalsylvain (approved):

EmilioSerna:
Hi @luisg123v 

Could you please review?

cc @hugho-ad

luisg123v (changes requested):
Remove also `cryptography`. It was not removed from v14 because it was not in Odoo's main requirements (see https://github.com/OCA/reporting-engine/pull/757#issuecomment-1516833160). But, starting from v15, it is.

EmilioSerna:
> Remove also `cryptography`. It was not removed from v14 because it was not in Odoo's main requirements (see [OCA/reporting-engine#757 (comment)](https://github.com/OCA/reporting-engine/pull/757#issuecomment-1516833160)). But, starting from v15, it is.

@luisg123v 
Changes applied

luisg123v (changes requested):
Combine the two last commits (letsencrypt and the one that removes cryptography from requirements) because they both belong to the same change. Take into account the `requirements.txt` file is autogenerated from module's manifest

EmilioSerna:
@luisg123v 

Done

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review, please?

yaazkal:
I know this is old now, but I wonder why you are keeping `cryptography==2.6.1` anyway in the `requirements.txt` file and in the `auto_backup` module.

PR #2654, opened 2023-06-07
IMP attach_queue: abstract notification on failure

bealdav:
cc @florian-dacosta @kevinkhao @sebastienbeau 

This is the first step to use activity to notify users instead of mails in this module

### DIFF
--- a/attachment_queue/models/attachment_queue.py
+++ b/attachment_queue/models/attachment_queue.py
@@ -73,6 +73,9 @@ class AttachmentQueue(models.Model):
 
     def _get_failure_emails(self):
         # to be overriden in submodules implementing the file_type
+        _logger.info(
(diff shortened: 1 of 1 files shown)


attachment_queue/models/attachment_queue.py
```diff
@@ -73,6 +73,9 @@ def _compute_failure_emails(self):
 
     def _get_failure_emails(self):
         # to be overriden in submodules implementing the file_type
+        _logger.info(
```
kevinkhao, line 76: suggestion for deprecation

https://github.com/odoo/odoo/blob/83a7daa07fedcaeddb983a94a86e4c7a187915d9/odoo/tools/misc.py#L86

LGTM

kevinkhao (approved):

PR #1982, opened 2021-01-07
[14.0][MIG] module_analysis: Migration to 14.0

santostelmo:

PR #2656, opened 2023-06-07
[14.0] [MIG] `module_analysis`

ivantodorovich:
There are two other PRs migrating this module to `14.0`, but they're old and CI is red:
- https://github.com/OCA/server-tools/pull/1982 @santostelmo 
- https://github.com/OCA/server-tools/pull/2085 @kevinkhao 

This is a new attempt. It migrates from `13.0`, but it also cherry-picks some commits from `15.0`.

Additionally, it includes 2 more commits of my own:

1) **To refactor tests**, it solves [the red CI issue the other 2 PRs had](https://github.com/OCA/server-tools/pull/2085#issuecomment-1407217371)

> Before this commit, the tests consisted in:
> 
> 1) Analyze all installed modules.
> 2) Check that we have at least 1 code line for each installed addon.
> 3) Check that not installed addons have 0 python code lines in the analysis.
> 
> This test is pretty naive, as it's not making any assertions on the resulted
> analysis. It's also highly inefficient, as it's analyzing all installed addons
> and takes ~10 minutes in our CI.
> 
> Moreover, at the time of this writing it was not possible to have our CI pass
> with the exiting tests due to some incompatibility with the `sentry` addon.
> 
> The new approach, on the other hand, uses a mocked function in order to analyze
> a dummy module instead, located in this module's `tests/module` directory. As
> we have full control over this fake module source code, we can make assertions
> on the analysis results.

2) **To fix a bug where `scss` files are being ignored**

These two commits are done separately so that we can easily forward-port them to `15.0` and `16.0`

ping @legalsylvain


module_analysis/post_init_hook.py
```diff
@@ -0,0 +1,16 @@
...
+
+def analyse_installed_modules(cr, registry):
+    if tools.config["test_enable"]:
+        return
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        installed_modules = env["ir.module.module"].search(
+            ["|", ("state", "=", "installed"), ("name", "=", "module_analysis")]
+        )
+        installed_modules.button_analyse_code()
```
legalsylvain, line 16: why call this function in each load ?

module_analysis/tests/module/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.parnter"
```
legalsylvain:
```suggestion
    _inherit = "res.partner"
```

legalsylvain (commented):
nice improvment ! scss and test.

a question regarding post_init_hook.

otherwise, LGTM ! 
thanks !

module_analysis/post_init_hook.py
ivantodorovich: It's called once, right after the module is installed.
It's true that, with the cron you implemented, it may not be needed -- but I guess it's so that users don't have to wait for the cron to run the first time

legalsylvain: You're right. I confused it with post_load. sorry for the noice.

legalsylvain (approved):
thanks ! 

CC : @PierrickBrun, you could be interested by this one.

PierrickBrun (approved):
Very nice, thanks !

hparfr (approved):
Ready for a merge.

PR #2659, opened 2023-06-15
[14.0] sentry: add optin to not send  startup event to Sentry

petrus-v:
On odoo saas platform event has cost. So ignoring odoo starting as signifiant effect on platform likes odoo.sh where cron workers start before each calls or where worker restarts.


hparfr (approved):

ioans73 (approved):
Nice feature 👍🏻

ioans73:
@moylop260 can you please review and merge the PR?

PR #2664, opened 2023-06-29
 14.0 mig auth generate password

gfcapalbo:


auth_generate_password/tests/tests.py
```diff
@@ -0,0 +1,25 @@
+from openerp.tests.common import TransactionCase
```
ddejong-therp: Could you change `openerp` to `odoo`?

gfcapalbo: done

gfcapalbo:
@ddejong-therp fixed the old lib name to odoo,  increased coverage and linted. 

there is one failure on tests:
(odoo.addons.base_model_restrict_update.tests.test_base_model_restrict_update.TestBaseModelRestrictUpdate)

this module wants to create a user without an email.  Now this should be possible, but the module as it was written 
raises an error in this case to warn the client.

The res.user email field is NOT a required field, so raising an error if there is no email makes no sense in odoo 14.0

I will functionally modify this module like this:

                   - no error raised if user is created without email. 
                   -  if user without email is then modified to have one, email is sent.
                   -  if user WITH email  has email is modified with different email, NOTHING is sent (password is sent once).

Fixing for odoo 14 compatribility.

gfcapalbo:
@ddejong-therp  all tests passing now.

will not raise error anymore when creating a user without an email. like v7.0 did..

I believe it might not had been a good idea in 7.0 either. 
https://github.com/OCA/OCB/blob/7.0/openerp/addons/base/res/res_partner.py#L260


Now the module behaves as so:

           -  if user is created with email , all normal, we generate password and send it.
           -  if  user is created without email, on first write with "email" password is set and sent. It is never sent again on subsequent changes. existing users with emails will not be affected.


ready for new review.

auth_generate_password/README.rst
```diff
@@ -0,0 +1,109 @@
...
+~~~~~~~
+
+* Groupement Regional Alimentaire de Proximite
+* Therp B.V.
+
+Contributors
+~~~~~~~~~~~~
+
+* Sylvain LE GAL (https://twitter.com/legalsylvain)
+* Giovanni Francesco Capalbo , Therp B.V. <giovanni@therp.nl>
```
NL66278, line 92: No dots in BV, just Therp BV

auth_generate_password/data/email_template.xml
```diff
@@ -0,0 +1,27 @@
...
+            name="body_html"
+        ><![CDATA[
+Your Odoo credentials have been set:
+<ul>
+<li>Login : ${object.login}</li>
+<li>Password : ${ctx.get('generated_password')}</li>
+</ul>
+Please,
+<ul>
+<li>remember this password and delete this email;</li>
```
NL66278, line 21: We certainly should not encourage multiple people to use the same credentials. Actually sending out passwords by email is a doubtful practice from a security perspective.

PR #2660, opened 2023-06-23
[14.0][ADD] base_future_response

sbidoul:
This is a backport of Odoo 16 `request.future_response` mechanism, to allow
setting cookies in the response before the response object is constructed.

PR #2665, opened 2023-06-29
[IMP] base_future_response: add test

sbidoul:


sbidoul:
A quick followup to #2660 with a test.

PR #2668, opened 2023-06-30
[ADD] jsonifier: with_fieldname option

gfcapalbo:
Jsonifier adds an extra key called fieldname_$field with field name  on the same level of the $field key.
Does not disrupt, function feature or specialized resolvers.


jsonifier/models/models.py
```diff
@@ -194,6 +201,62 @@ def jsonify(self, parser, one=False):
              'field_name:json_key'
         ]
 
+        if with_fieldname=True,the name of the field will be added
+        to the value. the name of the field will be translated.
```
thomaspaulb: (...) translated to the language currently selected in context['lang']

jsonifier/models/models.py
```diff
@@ -194,6 +201,62 @@ def jsonify(self, parser, one=False):
...
+
+                result['name']['value']  # get value of field.name
+                result['name']['fieldname']   # get the translated fieldname of name
+                # get value of field.order_line.product_uom.
+                result['order_line']['value']['product_uom']['value']
+                # fieldname of  field.order_line.product_uom
+                result['order_line']['value']['product_uom']['fieldname']
+
+        I need to use structure {'value':x, 'fieldname': y} because the json name
+        (order_line) may be duplicated in the various levels of relational fields.
```
thomaspaulb: We went for the nested dictionary approach, but maybe it would have been better to go for sister keys at the same level named `_fieldname_$field` instead; so that the with_fieldname context key won't break code that isn't expecting the context key to be there.

jsonifier/models/models.py
```diff
@@ -194,6 +201,62 @@ def jsonify(self, parser, one=False):
...
+
+        returns:
+
+        [{'name': 'SO/2022/006',
+                   'create_date': '2022-10-14T15:55:39.823019+02:00',
+                   'order_line': {
+                         'id': 4,
+                         'product_uom': 'Stuk(s)
+                         'is_expense': False}
+                    }]
```
thomaspaulb: I don't feel that this is the place for an elaborate documentation of the usage of this module with examples, that's for `readme/USAGE.txt`. It also needs an extra line that explains the use cases in which this is necessary, and that's when you are using `jsonifier` in the context of an API, and you want to build a visual UI around the nested dataset that you are fetching - then you need the field names.

thomaspaulb (changes requested):

gfcapalbo:
Failing for reduced coverage, will add tests-

jsonifier/models/models.py
gfcapalbo: Moved my extra  explanations in readme.

gfcapalbo:
@thomaspaulb I had started this on the very, very old version that was available on clients server.
The jsonifier has been completely refactored (and inproved).  What we wanted could had been obtained with the many many more options available.  I rewrote the MR to give customer an easy way to obtain the results we need.

Re-review needed.

jsonifier/README.rst
```diff
@@ -192,8 +192,45 @@ new requirements could be met without needing to add field or change any code.
...
+    parser=['name', 'create_date', ('order_line', ['id' , 'product_uom', 'is_expense'])]
+    >>> a.jsonify(parser=parser, default_function=False)
+    [{'name': 'SO3996', 'create_date': '2015-06-02T12:18:26.279909+00:00', 'order_line': [{'id': 16649, 'product_uom': 'stuks', 'is_expense': False}, {'id': 16651, 'product_uom': 'stuks', 'is_expense': False}, {'id': 16650, 'product_uom': 'stuks', 'is_expense': False}]}]
+    >>> a.jsonify(parser=parser, default_function=True)
+    [{'name': {'fieldname': 'Order Reference', 'value': 'SO3996'}, 'create_date': {'fieldname': 'Creation Date', 'value': datetime.datetime(2015, 6, 2, 12, 18, 26, 279909)}, 'order_line': {'fieldname': 'Order Lines', 'value': sale.order.line(16649, 16651, 16650)}}]
+
+
+
+
+NOTE this module was named `base_jsonify` till version 14.0.1.5.0.
```
thomaspaulb: It seems more logical if this notice resides just before the new doc section you're introducing. Besides - `README.rst` is a generated file - it's [DESCRIPTION.rst](https://github.com/OCA/server-tools/blob/14.0/jsonifier/readme/DESCRIPTION.rst) you should be modifying, or rather `USAGE.rst`, which does not exist yet.

gfcapalbo:
WIP. small fixes and tests needed.

jsonifier/README.rst
```diff
@@ -192,8 +192,45 @@ new requirements could be met without needing to add field or change any code.
...
+
+.. code-block:: python
+   
+    # example 1 
+    parser = [('name')]
+    a.jsonify(parser=parser)
+    [{'name': 'SO3996'}]
+    >>> a.jsonify(parser=parser, default_function=False)
+    [{'name': 'SO3996'}]
+    >>> a.jsonify(parser=parser, default_function=True)
```
thomaspaulb: TBH I find the parameter name `default_function` confusing. I can understand where you get it from, but to the end user it suggests that some kind of default would be applied, whereas that's not what it does, what it does is to add field names to the output. So `with_field_names=True` would be more intuitive.

jsonifier/models/models.py
```diff
@@ -83,6 +83,11 @@ def _jsonify_record(self, parser, rec, root):
             except SwallableException:
                 continue
             json_key = field_dict.get("target", field_dict["name"])
+            if rec.env.context.get("default_function") and not field_dict.get(
+                "function"
+            ):
+                # after all processing we post-process the parser to impose a function.
+                field_dict["function"] = "default_function"
```
thomaspaulb: If you do it like this, then, if the `field_dict` already has a function, this won't work. That would be a limitation. You could overcome that by moving this to the end of the function, where `value` is already guaranteed to be defined and where you can get `field` by doing `field = rec._fields[field_dict["name"]]`. Then, you can do:

```
if self.env.context.get("default_function"):
    value = {'value': self.default_function(field)}
```

jsonifier/models/models.py
```diff
@@ -256,3 +266,7 @@ def _jsonify_format_duration(self, fname):
 
         """
         return format_duration(self[fname])
+
+    def default_function(self, field):
```
thomaspaulb: After the refactor suggested above, this should probably also be prefixed with `_`, and probably also with `_jsonify` in order not to pollute the namespace.

jsonifier/README.rst
```diff
@@ -192,8 +192,45 @@ new requirements could be met without needing to add field or change any code.
...
+    [{'name': {'fieldname': 'Order Reference', 'value': 'SO3996'}}]
+
+
+    # example 2
+    parser=['name', 'create_date', ('order_line', ['id' , 'product_uom', 'is_expense'])]
+    >>> a.jsonify(parser=parser, default_function=False)
+    [{'name': 'SO3996', 'create_date': '2015-06-02T12:18:26.279909+00:00', 'order_line': [{'id': 16649, 'product_uom': 'stuks', 'is_expense': False}, {'id': 16651, 'product_uom': 'stuks', 'is_expense': False}, {'id': 16650, 'product_uom': 'stuks', 'is_expense': False}]}]
+    >>> a.jsonify(parser=parser, default_function=True)
+    [{'name': {'fieldname': 'Order Reference', 'value': 'SO3996'}, 'create_date': {'fieldname': 'Creation Date', 'value': datetime.datetime(2015, 6, 2, 12, 18, 26, 279909)}, 'order_line': {'fieldname': 'Order Lines', 'value': sale.order.line(16649, 16651, 16650)}}]
```
thomaspaulb: It would be nice to add a test to test one of these usecases.

thomaspaulb (changes requested):

jsonifier/models/models.py
gfcapalbo: Now changed it.  I leave function alone.  
Simplified and working well as "Sister Key", as you suggested.

Still in WIP.

jsonifier/models/models.py
gfcapalbo: Very nice, but it was wrong of me to piggyback on the function feature, wich sould remain untouched.
Plus it did not allow me to add a "sister" key, just change the value.  Alnmost done with a simple  version.
Works. Still in WIP

jsonifier/README.rst
gfcapalbo: Moved to Usage.rst

jsonifier/README.rst
gfcapalbo: added

jsonifier/models/models.py
gfcapalbo: initial solution was best. now a sister key is there no more function -meddling.

jsonifier/models/models.py
gfcapalbo: function solution was trashed. tested and works with a more simple solution.

jsonifier/README.rst
gfcapalbo: I put it as "default function" because that solution could had been used for anything and kept a generic name.
The default was adding the fieldnames. but it could had been something else.
that solution was thrashed and now is with_fieldnames as it should be.

jsonifier/README.rst
gfcapalbo: Now in USAGE.rst

jsonifier/models/models.py
gfcapalbo: $field_fieldname is now  a "sister key".  but now i see that fieldname_$field is better. Changing.

gfcapalbo: done.

gfcapalbo:
@thomaspaulb  complete with passing tests. Removing WIP.
the new structure will not make the old code fail, because it does not change the value fetched, but adds on same level a new key.  I had an idea to override function feature of jsonifier, but that is not needed anymore.
.
(see test data and usage.rst)

As a possible jsonifier addition the feature "default_function" could be implemented in a separate MR, unrelated to this

All explanations and examples moved to usage.rst

jsonifier/readme/USAGE.rst
```diff
@@ -0,0 +1,30 @@
+with_fieldname parameter
+==========================
+
+The with_fieldname option of jjsonify() method, hen true,  will inject on
```
thomaspaulb:
```suggestion
The with_fieldname option of jsonify() method, when true, will inject on
```

jsonifier/readme/USAGE.rst
```diff
@@ -0,0 +1,30 @@
...
+the same level of the data fieldname_$field keys that will
+give us the field name , in current user language of the field itself.
```
thomaspaulb:
```suggestion
contain the field name, in the language of the current user.
```

jsonifier/readme/USAGE.rst
```diff
@@ -0,0 +1,30 @@
...
+This will add keys for translated front-end labels, without disrupting the access
+to existing code.
```
thomaspaulb: I would remove this sentence, it's self explanatory.

jsonifier/models/models.py
```diff
@@ -105,6 +105,13 @@ def _jsonify_record(self, parser, rec, root):
                     value, json_key = self._jsonify_record_handle_resolver(
                         rec, field, resolver, json_key
                     )
+            # whatever json value we have found in subparser or not ass a sister key
+            # on the same level {jsoni_key}_fieldname
+            if rec.env.context.get("with_fieldname"):
+                json_key_fieldname = "_".join(["fieldname", json_key])
```
thomaspaulb: Would it make sense to prefix the whole key with `_` as well? In the odd case that some Odoo implementation has both `foo` and `fieldname_foo` as fields in a model

thomaspaulb (changes requested):

thomaspaulb:
I like the refactor to using a sister key, it's less intrusive.

thomaspaulb (approved):

jsonifier/models/models.py
```diff
@@ -105,6 +105,13 @@ def _jsonify_record(self, parser, rec, root):
                     value, json_key = self._jsonify_record_handle_resolver(
                         rec, field, resolver, json_key
                     )
+            # whatever json value we have found in subparser or not ass a sister key
+            # on the same level {jsoni_key}_fieldname
```
RuudFreshfilter: It looks like this comment is outdated, might want to review this.

jsonifier/models/models.py
```diff
@@ -105,6 +105,13 @@ def _jsonify_record(self, parser, rec, root):
...
+            if rec.env.context.get("with_fieldname"):
+                json_key_fieldname = "_".join(["_fieldname", json_key])
```
RuudFreshfilter: To me this is less readable than something like
`"_fieldname_" + json_key` or `f"_fieldname_{json_key}"`

jsonifier/models/models.py
gfcapalbo: fixed

jsonifier/models/models.py
gfcapalbo: I've went  with a more readable version

RuudFreshfilter (approved):

simahawk:
I'm forward porting from 14 to 16... it's a bit painful as commit were not squashed here. Next time, please, do it if the feature is just one at the end :wink: 
Danke! :pray:

PR #2670, opened 2023-07-04
[16.0] base_exception: Improves performances

lmignon:
To follow the changes you can read the commits one by one.


lmignon:
ping @rousseldenis

rousseldenis (approved):
Code review.

Indeed, before :

image


After :

image



Even if few tests, quite 25% gain...

base_exception/models/exception_rule.py
```diff
@@ -70,3 +70,59 @@ def _get_domain(self):
...
         return safe_eval(self.domain)
+
+    def _get_rules_info_for_domain(self, domain):
+        """returns the rules that match the domain
+
+        This method will call _get_cached_rules_for_domain to get the rules
+        that match the domain. It will then transform the result into a list
+        of objects that have the same attributes as the exception.rule model.
+        """
+        cache_info = self._get_cached_rules_for_domain(tuple(domain)).copy()
```
hparfr: why copy() ?

base_exception/models/exception_rule.py
```diff
@@ -70,3 +70,59 @@ def _get_domain(self):
...
+        return {
+            "id": self.id,
+            "name": self.name,
+            "description": self.description,
+            "sequence": self.sequence,
+            "model": self.model,
+            "exception_type": self.exception_type,
+            "domain": self._get_domain()
+            if self.exception_type == "by_domain"
+            else None,
```
hparfr, line 118: is None a valid domain ?

hparfr (approved):
thanks for addressing this subject.

Code read; just two small remarks

base_exception/models/exception_rule.py
lmignon: should be removed... a trace from a previous implementation

base_exception/models/exception_rule.py
lmignon: No but the domain is only used  if `exception_type == "by_domain"`

hparfr (approved):
LGTM

hparfr:
can we merge this one ?

lmignon:
> can we merge this one ?

@hparfr yes plz

hparfr:
I'm not PSC here :superhero:

PR #2676, opened 2023-07-14
15.0 mig base name search improved

thinkwelltwd:


base_name_search_improved/models/ir_model.py
```diff
@@ -233,13 +235,18 @@ def update_search_wo_restart(self):
...
-            name_search_domain = False
             try:
                 name_search_domain = literal_eval(rec.name_search_domain)
-            except Exception as error:
+            except (
+                ValueError,
+                TypeError,
+                SyntaxError,
+                MemoryError,
+                RecursionError,
```
thomaspaulb, line 245: Why this particular list and not others?

PR #2547, opened 2023-02-09
[15.0][MIG] database_cleanup: Migration to 15.0

yankinmax:

PR #2219, opened 2021-12-03
[15.0][MIG] database_cleanup

damdam-s:
based on #1990

PR #2682, opened 2023-07-20
[15.0][FIX]database_cleanup: endless loop

Camille0907:
We have some cases where database_cleanup loop endlessly because of referential integrity. 

In my case, the issue was caused by tables having foreign keys on tables which are not in the list of tables to delete. Updating the code to drop all constraints, even on tables which are not meant to be dropped, seems to solve the issue 

___
[OUTDATED]
_Here for example, DROP TABLE is failing because some tables have foreign keys on a studio table, and the error message suggest using DROP TABLE ... CASCADE, which seems indeed to fix the problem._
```
        Start purging tables attempt n° 158
        Try to purge table: transport_vehicle
2023-07-20 07:50:32,502 347 ERROR test_error odoo.sql_db: bad query: DROP TABLE "transport_vehicle"
ERROR: cannot drop table transport_vehicle because other objects depend on it
DETAIL:  constraint x_journaux_x_studio_field_PE7Np_fkey on table x_journaux depends on table transport_vehicle
 
        Try to purge table: transport_vehicle_type
2023-07-20 07:50:32,513 347 ERROR test_error odoo.sql_db: bad query: DROP TABLE "transport_vehicle_type"
ERROR: cannot drop table transport_vehicle_type because other objects depend on it
DETAIL:  constraint transport_vehicle_vehicle_type_id_fkey on table transport_vehicle depends on table transport_vehicle_type
constraint x_journaux_x_studio_field_mHvR3_fkey on table x_journaux depends on table transport_vehicle_type
```

_First introduced by https://github.com/damdam-s/server-tools/pull/1 in https://github.com/OCA/server-tools/pull/2219_

_But lost when this PR has been superseded by https://github.com/OCA/server-tools/pull/2547_

### DIFF
--- a/database_cleanup/models/purge_tables.py
+++ b/database_cleanup/models/purge_tables.py
@@ -63,13 +63,19 @@ class CleanupPurgeLineTable(models.TransientModel):
                         constraint[0],
                         constraint[3],
                     )
-                    self.env.cr.execute(
-                        "ALTER TABLE %s DROP CONSTRAINT %s",
-                        (
-                            IdentifierAdapter(constraint[3]),
-                            IdentifierAdapter(constraint[0]),
-                        ),
+                else:
+                    self.logger.info(
+                        "Dropping constraint %s on table %s (not to be dropped)",
+                        constraint[0],
+                        constraint[3],
                     )
+                self.env.cr.execute(
+                    "ALTER TABLE %s DROP CONSTRAINT %s",
+                    (
+                        IdentifierAdapter(constraint[3]),
+                        IdentifierAdapter(constraint[0]),
+                    ),
+                )
 
             self.logger.info("Dropping table %s", line.name)
             self.env.cr.execute("DROP TABLE %s", (IdentifierAdapter(line.name),))


database_cleanup/models/purge_tables.py
```diff
@@ -72,7 +72,9 @@ def purge(self):
                     )
 
             self.logger.info("Dropping table %s", line.name)
-            self.env.cr.execute("DROP TABLE %s", (IdentifierAdapter(line.name),))
+            self.env.cr.execute(
+                "DROP TABLE %s CASCADE", (IdentifierAdapter(line.name),)
```
StefanRijnhart: This looks dangerous and I would never knowingly run this on a production database. Do you think it would work to drop *all* constraints for all tables in the `line` loop in a first pass and then drop the tables in a second pass?

Camille0907: You're right :) I tried your proposition and it works, at least in my case. Updating the PR to this less dangerous solution ;) Thank you !

StefanRijnhart: What used to be the case is that constraints between tables that were going to be dropped were removed explicitely. That can be considered safe, because the related tables were going to be dropped anyway. What happens now in the suggested change is that constraints from a table that is not going to be dropped to a table that is going to be dropped are being dropped explicitely.
I would not consider this safe, because the existence of such a constraint is an indication that it is not safe to drop the table in the first place. It also breaks dropping tables that still (only) depend on each other.

thomaspaulb: So, what would the safe solution be? Provide a bugfix just for the endless loop, but fail to drop the table?

StefanRijnhart: Considering that
* The current code in this PR is unsafe,
* My original suggestion is very much what is already implemented in the released version of the module,
* The issue occurs when dealing with a manually created `x_` model according to the traceback,

I would make an exception only when dealing with a manually created model (c.q. table) that starts with `x_`

PR #2473, opened 2022-11-24
[16.0][MIG] database_cleanup: Migration to 16.0

JoelZilli:

PR #2683, opened 2023-07-25
16.0][MIG] database_cleanup: Migration to 16.0

miikanissi:
Original PR author seems to be unavailable at https://github.com/OCA/server-tools/pull/2473

This PR continues the progress made in that and includes a cherry pick to fix tests


pedrobaeza:
Thanks for the contribution.

You should also preserve the previous migrator commit and put on top of it your changes.

PR #2688, opened 2023-07-27
[16.0][FIX] sentry: Prevent capturing ignored exceptions

MrGigSolutions:
With this change, any exceptions that are in the DEFAULT_IGNORE_LIST will not be captured by Sentry.

### DIFF
--- a/sentry/tests/test_client.py
+++ b/sentry/tests/test_client.py
@@ -17,6 +17,8 @@ from ..hooks import initialize_sentry
 GIT_SHA = "d670460b4b4aece5915caf5c68d12f560a9fe3e4"
 RELEASE = "test@1.2.3"
 
+_logger = logging.getLogger(__name__)
+
 
 def remove_handler_ignore(handler_name):
     """Removes handlers of handlers ignored list."""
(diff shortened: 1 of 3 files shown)


sentry/tests/test_client.py
```diff
@@ -118,6 +120,35 @@ def test_ignore_exceptions(self):
...
+        ignore_exceptions is not captured (there is no exc_info in the
+        ValidationError exception)"""
+        client = initialize_sentry(config)._client
+        client.transport = InMemoryTransport({"dsn": self.dsn})
+        level, msg = logging.WARNING, "Test exception"
+        try:
+            raise exceptions.ValidationError(msg)
+        except exceptions.ValidationError as exception:
+            # Odoo handles UserErrors by logging the exception
+            _logger.warning(exception)
```
StefanRijnhart, line 148: Is this not equivalent to
```suggestion
        _logger.warning(exceptions.ValidationError(msg))
```

StefanRijnhart: Is this not equivalent to
```suggestion
        _logger.warning(exceptions.ValidationError(msg))
```

MrGigSolutions: Very good point. I have adjusted the code accordingly, and improved that test to be more understandable too.

StefanRijnhart (approved):
Thanks!

moylop260 (approved):

PR #2689, opened 2023-07-28
[15.0][MIG] base_translation_field_label

JasminSForgeFlow:
Standard Migration

@ForgeFlow


GuillemCForgeFlow (changes requested):
I see that tests are failing because it doesn't find the language used on the test. Perhaps we need to unarchive it first on the test so it can be found.

JasminSForgeFlow:
> I see that tests are failing because it doesn't find the language used on the test. Perhaps we need to unarchive it first on the test so it can be found.

Its done,
Thanks

GuillemCForgeFlow (approved):
LGTM! 👍🏽

JordiMForgeFlow (approved):
Code + functional review 👍🏼

LoisRForgeFlow (approved):
LGTM

PR #2690, opened 2023-07-29
[ADD][14.0] new module: base_order_by_related

thomaspaulb:
This module allows a search to order by a non-stored related field.

### DIFF
--- a/base_order_by_related/__manifest__.py
+++ b/base_order_by_related/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright (C) 2023 Therp (<http://www.therp.nl>).
+# @author Tom Blauwendraat <tblauwendraat@therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Base order by related",
+    "summary": "Order by non-stored related fields",
+    "version": "14.0.1.0.0",
+    "category": "Tools",
+    "website": "https://github.com/OCA/server-tools",
+    "author": "Therp BV, Odoo Community Association (OCA)",
+    "maintainers": ["thomaspaulb"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["base"],
+    "data": [],
+}
--- a/base_order_by_related/readme/USAGE.rst
+++ b/base_order_by_related/readme/USAGE.rst
@@ -1,0 +1,9 @@
+In any model where you have a non-stored related field, you can make it sortable::
+
+    class ResPartner(models.Model):
+        _inherit = "res.partner"
+
+        country_name = fields.Char(related="country_id.name")
+
+        # make some non-stored related fields sortable
+        _order_by_related = ["country_name"]
--- a/base_order_by_related/tests/fake_models.py
+++ b/base_order_by_related/tests/fake_models.py
@@ -1,0 +1,20 @@
+# Copyright (C) 2023 Therp (<http://www.therp.nl>).
+# @author Tom Blauwendraat <tblauwendraat@therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class FakeCustomer(models.Model):
+    _name = "fake.customer"
+    _description = "fake model for customers"
+
+    _order_by_related = ["country_name", "country_code"]
+
+    name = fields.Char(required=True)
+    sortkey1 = fields.Integer()
+    sortkey2 = fields.Integer()
+    sortkey3 = fields.Integer()
+    country_id = fields.Many2one("res.country")
+    country_name = fields.Char(related="country_id.name")
+    country_code = fields.Char(related="country_id.code")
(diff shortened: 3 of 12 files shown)


base_order_by_related/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
gjotten: @thomaspaulb needs different attribution methinks.

base_order_by_related/tests/fake_models.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (C) 2022 Akretion (<http://www.akretion.com>).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
gjotten: @thomaspaulb here as well.

base_order_by_related/tests/test_order_by_related_mixin.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (C) 2022 Akretion (<http://www.akretion.com>).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
gjotten: @thomaspaulb here as well.

base_order_by_related/models/order_by_related_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @api.model
+    def _generate_order_by_inner(
+        self, alias, order_spec, query, reverse_direction=False, seen=None
+    ):
+        order_parts = order_spec.split(",")
+        order_fields = [p.strip().split(" ")[0].strip() for p in order_parts]
+        order_by_related = self._order_by_related or {}
+        if not any(f in order_by_related for f in order_fields):
+            return super(OrderByRelatedMixin, self)._generate_order_by_inner(
```
hbrunn:
```suggestion
            return super()._generate_order_by_inner(
```

base_order_by_related/models/order_by_related_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+class OrderByRelatedMixin(models.AbstractModel):
+    _name = "order.by.related.mixin"
+    _description = "Mixin to allow ordering by non-stored related fields"
+
+    @api.model
+    def _related_join_calc(self, alias, fname, query, real_fname):
+        model, field = self, self._fields[fname]
+        while field.related:
+            # retrieve the parent model where field is inherited from
+            related_model = self.env[field.related_field.model_name]
```
hbrunn:
```suggestion
            if field.related_field.model_name not in self.env:
                continue
            related_model = self.env[field.related_field.model_name]
```

for this kind of low-level stuff I'd suggest to be cautious about models/fields not existing due to a half loaded registry, also above with the field

base_order_by_related/readme/USAGE.rst
```diff
@@ -0,0 +1,46 @@
+In our case we use this for the case of ordering common Odoo models by a non-stored computed field.
+
+This is normally not possible, but in this case it can be hacked in by creating a view such as::
```
hbrunn: I found your example actually a bit too detailed. Don't we have some non-stored related field in the base module that you can use to make your point, same for tests?

base_order_by_related/models/order_by_related_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            return '"%s"."%s"' % (alias, related_fname2)
+
+    @api.model
+    def _generate_order_by_inner(
+        self, alias, order_spec, query, reverse_direction=False, seen=None
+    ):
+        order_parts = order_spec.split(",")
+        order_fields = [p.strip().split(" ")[0].strip() for p in order_parts]
+        order_by_related = self._order_by_related or {}
```
hbrunn:
```suggestion
        order_by_related = getattr(self, '_order_by_related',  {})
```

hbrunn (approved):
code looks good to me, but personally I think I'd implement this as overrides in the `base` model, and calculate `_order_by_related` in `base#init` - that should make it more convenient for devs to use, don't you think?

thomaspaulb:
> code looks good to me, but personally I think I'd implement this as overrides in the `base` model, and calculate `_order_by_related` in `base#init` - that should make it more convenient for devs to use, don't you think?

It would; I tried in two ways:

- Calculating `_order_by_related` in `base#init` for all non-stored related fields: this enables search and modifies the ORM call for a lot of problematic fields as well, so I think this should be an opt-in thing only
- Opt-in via a kwarg in `Field#init`: looked easy to do, but it will require another hook on the Field object which will complicate the module

So, I stick with `_order_by_related`, which has now simplified to being just a list.

hbrunn (approved):

gjotten (approved):

PR #2692, opened 2023-08-01
[IMP] auditlog/models/rule.py: Add condition to skip logging if no fields changed in write method

chupaSV:
Add a condition to check if any fields have changed before creating a log entry for the "write" method. This improves performance and reduces unnecessary log entries.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -516,10 +516,16 @@ class AuditlogRule(models.Model):
                 "http_session_id": http_session_model.current_http_session(),
             }
             vals.update(additional_log_values or {})
-            log = log_model.create(vals)
+
             diff = DictDiffer(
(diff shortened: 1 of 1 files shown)


ChristophAbenthungCibex:
Great functionality. 
However, when the only changed field is a field of the list fields_to_exclude a log entry is still created but without any entry at the updated fields table. I see no problem to prevent log creation if those fields are changed.
I changed the line changed_fields = diff.changed() - set(FIELDS_BLACKLIST) to changed_fields = diff.changed() - set(FIELDS_BLACKLIST) - set(fields_to_exclude) and it works fine.

Steps to reproduce:
1. Go to "Rules"
2. Add the active field of the model to "fields_to_exclude_ids"
3. "Subscribe" the rule
4. Archive a record of the model of the rule
5. A log without any entry at updated fields is created

PR #2696, opened 2023-08-14
[15.0][ENH] auditlog: add message tracking back to origin document

Saran440:
Concept: Use in case log message change in line (Example: account move)

Step:
1. Create auditlog rule > check `Message Track` and select field reference (field reference to header)
Selection_016
2. Test edit value in line. it will log note
Selection_017

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -147,6 +147,17 @@ class AuditlogRule(models.Model):
         string="Fields to Exclude",
         states={"subscribed": [("readonly", True)]},
     )
+    message_track = fields.Boolean(
+        help="Select this if you want to keep track message note of this rule",
+        states={"subscribed": [("readonly", True)]},
+    )
+    field_reference_id = fields.Many2one(
+        comodel_name="ir.model.fields",
+        domain="[('model_id', '=', model_id)]",
+        string="Reference Track",
+        states={"subscribed": [("readonly", True)]},
+        help="Select field reference for track message",
+    )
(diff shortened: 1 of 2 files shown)


auditlog/models/rule.py
```diff
@@ -518,10 +529,14 @@ def create_logs(
             diff = DictDiffer(
                 new_values.get(res_id, EMPTY_DICT), old_values.get(res_id, EMPTY_DICT)
             )
+            origin_model = self.env[log.model_model].browse(log.res_id)
```
thomaspaulb, line 532: - Maybe instead of `log.res_id` just use `res_id`
- Instead of `self.env[log.model_model]` you can just use `model_model` like defined above
- If you move this code up you can change `name = model_model.browse(res_id).name_get()` to `name = origin_model.name_get()` as well

auditlog/models/rule.py
```diff
@@ -533,13 +548,31 @@ def create_logs(
...
+                    msg += "</ul>"
             elif method == "unlink" and auditlog_rule.capture_record:
                 self._create_log_line_on_read(
                     log,
                     list(old_values.get(res_id, EMPTY_DICT).keys()),
                     old_values,
                     fields_to_exclude,
                 )
+                if auditlog_rule.message_track:
+                    msg = "<ul><li>%s deleted</li></ul>" % (log.name)
```
thomaspaulb, line 570: In this way the word `deleted` is not translatable

PR #2698, opened 2023-08-16
[14.0][MIG] profiler

LoisRForgeFlow:
Standard migration to v14.

@ForgeFlow


thomaspaulb (approved):
LGTM, functional test

profiler/static/src/js/tour.js
```diff
@@ -0,0 +1,74 @@
...
+
+    var _t = core._t;
+
+    tour.register(
+        "profiler_tour",
+        {
+            url: "/web",
+        },
+        [
+            tour.STEPS.SHOW_APPS_MENU_ITEM,
```
thomaspaulb, line 15: Getting `Error while loading profiler.tour: TypeError: Cannot read properties of undefined (reading 'SHOW_APPS_MENU_ITEM')`

May have to be rewritten as [example](https://github.com/OCA/OCB/blob/14.0/addons/sale/static/src/js/tours/sale.js#L13)

thomaspaulb (changes requested):

PR #2684, opened 2023-07-25
[16.0][MIG] database_cleanup: Migration to 16.0

miikanissi:
Original PR author seems to be unavailable at https://github.com/OCA/server-tools/pull/2473

This PR continues the progress made in that and includes a cherry pick to fix tests

PR #2699, opened 2023-08-25
[16.0][MIG] database_cleanup

letzdoo-js:


dariodelzozzo:
There are already two open migration PRs:
https://github.com/OCA/server-tools/pull/2684
https://github.com/OCA/server-tools/pull/2473

Maybe is better to review them instead of opening a new one

PR #2700, opened 2023-08-25
[FIX] sentry: respect sentry_logging_level

aisopuro:
Before this fix, the Sentry module sent events for WARNING- level logs, even if `sentry_logging_level` was registered as "error" or higher.

The fix itself is minor: setup of the integration mistakenly set the hardcoded WARNING level to the event handler and the sentry_logging_level to the breadcrumb handler, when they should have been the other way around.

The largest part of the diff is a reworking of the tests in order to properly replicate the issue:

* The test previously emitted a fake log event directly using the integration's handler's emit-method, which skipped the part of the logic that actually filters based on logging level. This has been changed to use a bespoke `NoopHandler` and dedicated `Logger`, so that the tests can emit "actual logs" and test Sentry as accurately as possible.
* The tests were not configured to use a non-default logging level, thus making it so that none of them caught the fact we were basically hard-coding the setting to WARNING-level. The tests now set the logging level to ERROR in order to make sure the configuration parameter works when it is non-default.
* Changes to configuration (especially ignored loggers) were leaking from one test into others. The tests were directly mutating the `odoo.tools.config.options` mapping, without resetting it afterward, leaving the changes in place for subsequent tests. Introduced a helper method `patch_config` that can be used to patch the config object so that the patch is undone at the end of the test.

### DIFF
--- a/sentry/const.py
+++ b/sentry/const.py
@@ -69,7 +69,11 @@ def get_sentry_logging(level=DEFAULT_LOG_LEVEL):
     if level not in LOG_LEVEL_MAP:
         level = DEFAULT_LOG_LEVEL
 
-    return LoggingIntegration(level=LOG_LEVEL_MAP[level], event_level=logging.WARNING)
+    return LoggingIntegration(
+        # Gather warnings into breadcrumbs regardless of actual logging level
+        level=logging.WARNING,
+        event_level=LOG_LEVEL_MAP[level],
+    )
 
 
 def get_sentry_options():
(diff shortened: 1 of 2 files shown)


moylop260 (approved):
👍 

LGTM

Thank you!

aisopuro:
I amended my original commit with a slight fix: `test_exclude_logger` was removing the incorrect logger name from the set, leaving the excluded logger to still be excluded after the test completed.

I also added a new commit to increase test coverage to make the CI pass.

PR #2702, opened 2023-08-30
[13.0][IMP]base_exception: skip check if context is passed

GuillemCForgeFlow:
It can be useful for the cases where we want to skip the check validation if we are writing via code

cc @ForgeFlow


GuillemCForgeFlow:
closing as no reviews have been done

PR #2704, opened 2023-09-06
[14.0][IMP] base_custom_info: only save if data has been modified

etobella:
@dalonsod this should fix #2330

PR #2705, opened 2023-09-06
[IMP] base_custom_info: only save if data has been modified

etobella:
Port of #2704


pedrobaeza (approved):
Fast-tracking this:

/ocabot merge patch

PR #2707, opened 2023-09-10
base_contextvars: pytest workaround

sebastienbeau:
@sbidoul @lmignon When running test with pytest (on shopinvader_api_address for example). I had some issue with the contextvars env (the env is not found so everything crash).
It's seem that pytest do not support it correctly base_contextvars.
After some investigation I found this workaround.

Thanks for your feedback


sbidoul:
Hm, setting `_odoo_environments_ctx.set(envs)` is normally done by the monkey patched `Environment.manage`.
So it's likely that base_context var was imported too late.
Do you know if pytest-odoo supports server-wide-modules?

PR #2657, opened 2023-06-09
[16.0][FIX] Traceback FIX

Hardik-OSI:
Error on Auditlog create_fast deepcopy - TypeError: __new__() missing 3 required positional arguments: 'cr', 'uid', and 'context'
FIX.

PR #2710, opened 2023-09-11
[16.0] [FIX] auditlog consistency with Many2one fields

PieterPaulussen:
Odoo allows creating or updating records with values like ``{"many2one_field_id": empty.recordset()}``, but this crashes when using ``deepcopy``, which fails in recreating the ``api.Environment`` object attached to the recordset.
This commit should fix the issue by updating the values before passing them to ``deepcopy``.

This is a forward port of https://github.com/OCA/server-tools/pull/2485
And superseeds: https://github.com/OCA/server-tools/pull/2657


antonioburic (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2715, opened 2023-09-22
[16.0][MIG] base_multi_image

taiebnourddine12:


florian-dacosta:
Hello @taiebnourddine12,

You need  to preserve the commit history (see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0)
Also, this module does not belong to this repository, but to https://github.com/OCA/product-attribute
And there is already a PR on this module : https://github.com/OCA/product-attribute/pull/1329

It would probably be better to review this existing PR and/or contribute to it if necessary.

PR #2716, opened 2023-09-22
 [16.0][MIG] base_multi_image

taiebnourddine12:


pedrobaeza:
Thanks for the contribution.

florian-dacosta:
Also, there is an existing PR here : https://github.com/OCA/server-tools/pull/2618

PR #2733, opened 2023-10-13
[12.0][FIX] sentry: Patch odoo.service.server wsgi app

Rickpla:
Because post_load is called after odoo.service.server start
It has already registered the unpatched   odoo.service.wsgi_server.application
So patch it here too.
This enables wsgi performance reporting with sentry_traces_sample_rate

Cherry picking the fix by [paradoxxxzero](https://github.com/OCA/server-tools/commits?author=paradoxxxzero) from 14.0 to 12.0

### DIFF
--- a/sentry/hooks.py
+++ b/sentry/hooks.py
@@ -5,6 +5,7 @@ import logging
 
 import odoo.http
 from odoo.service import wsgi_server
+from odoo.service.server import server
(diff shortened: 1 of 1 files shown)


moylop260:
👍

moylop260 (approved):
👍

thomaspaulb (approved):

PR #2734, opened 2023-10-13
[14.0][IMP] model_read_only

ilyasProgrammer:
Now you can create multiple rules. Set domain and group for each rule.

going to be major (breaking)

not sure how to test migration script here.


model_read_only/models/models.py
```diff
@@ -1,14 +1,29 @@
...
 class IrModel(models.Model):
     _inherit = "ir.model"
 
-    readonly_group_ids = fields.Many2many(
-        "res.groups", relation="ir_model_group_readonly_rel"
-    )
+    readonly_restriction_ids = fields.One2many("model.readonly.restriction", "model_id")
+    # readonly_group_ids = fields.Many2many(
+    #     "res.groups", relation="ir_model_group_readonly_rel"
+    # )
```
HekkiMelody: **chore**: remove commented code

model_read_only/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+mrr1,mrr1,model_model_readonly_restriction,base.group_system,1,1,1,1
+mrr2,mrr2,model_model_readonly_restriction,base.group_user,1,0,0,0
```
HekkiMelody: **suggestion(non-blocking)**: I wonder if the user should even have read access. An alternative could be to remove this permission, and access restrictions using `sudo()` when checking access rights.

model_read_only/models/models.py
```diff
@@ -1,14 +1,29 @@
...
+    _name = "model.readonly.restriction"
+
+    model_id = fields.Many2one("ir.model")
+    model_name = fields.Char(related="model_id.model")
+    restriction_domain = fields.Char("Domain")
+    group_ids = fields.Many2many(
+        "res.groups",
+        relation="res_group_model_readonly_restriction_rel",
+        string="Groups",
+    )
```
HekkiMelody: **suggestion**: If I understand correctly:

- if `restriction_domain` is empty, the restriction is always active
- if `group_ids` is empty, the restriction applies to everyone

could you add this bit of documentation to the `help` of the fields as well as the README?

HekkiMelody (changes requested):
Code review: LGTM, only a couple of minor comments.

Please squash commits and add a message according to the guidelines. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

In general, I think this module doesn't quite fit with Odoo's way of managing permissions, since it's meant to restrict operations based on the user HAVING a group, whereas normally Odoo sets a restriction for everyone and then lifts it if you have a specific group. But this was the case before this PR as well.

ilyasProgrammer:
@aleuffre Thanks for the review. Updated.

HekkiMelody (approved):
Code review, LGTM

francesco-ooops (approved):
Functional ok!

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+            SELECT ir_model_id, res_groups_id
+            FROM ir_model_group_readonly_rel_tmp;
+        """,
+    )
+    data = env.cr.fetchall()
+    grouped = {}
```
geomer198: Could you use `defaultdict` function from [collections](https://docs.python.org/3/library/collections.html#collections.defaultdict)
`Defaultdict` allow you to quickly group a sequence of key-value pairs into a dictionary of lists.
```py
grouped = defaultdict(list)
```

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+    for item in data:
```
geomer198: Using unpacking allows you to decompose the sequence into individual values. You can decompress a list of values pairs in a data variable in the way:
```py
for key, value in data:
```

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+        key = item[0]
+        value = item[1]
+        if key in grouped:
+            grouped[key].append(value)
```
geomer198: Then you can do:
```py
grouped[key].append(value)
```

Using standard  language methods can improve your code and make it more readable.
In this case we can get the following code:
```py
grouped = defaultdict(list)
for key, value in data: grouped[key].append(value)
```

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+        else:
+            grouped[key] = [value]
+    for k, v in grouped.items():
```
geomer198: Please don't use short name for variables, this one not readable. Better, if you use understandable names for variables. For example:
```py
for key, values_list  in grouped.items():
    ...
```

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+        env["model.readonly.restriction"].create(
```
geomer198: Instead of calling create in a cycle you should use @api.create_multi and call create once for a list of values. This will result in a single db operation istead of multiple ones.
For example:
```py
vals_list = [{"model_id": model_id, "group_ids": group_ids} for model_id, group_ids in grouped]
env["model.readonly.restriction"].create(vals_list)
```

geomer198 (changes requested):
Please fix comments

model_read_only/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+            FROM ir_model_group_readonly_rel_tmp;
+        """,
+    )
+    data = env.cr.fetchall()
+    grouped = defaultdict(list)
+    for key, value in data:
+        grouped[key].append(value)
+    vals_list = [
+        {"model_id": model_id, "group_ids": group_ids}
+        for model_id, group_ids in grouped
```
geomer198: Grouped is a dict variable that should use `.item()` for iterating dict items. I got around this point in the last comment, sorry for that. Than your code must be looks like this:
```py
        for model_id, group_ids in grouped.items()
```

geomer198 (changes requested):
@ilyasProgrammer  
I found a one moment. Please fix this bug.

francesco-ooops:
@geomer198 good to go?

geomer198:
> @geomer198 good to go?

Yes)

geomer198 (approved):
LGTM!

PR #2739, opened 2023-10-22
[16.0][FIX] upgrade_analysis: blacklist + no reinstall + UI tweaks

pedrobaeza:
Several patches for problems found when performing a OU analysis:

- **[FIX] Add to blacklist deprecated modules**

  Odoo has deprecated such modules with a pre-hook raising an error, so they are not installable anymore.

  This way, we avoid them to be included in the Install Modules Wizard.

- **[IMP] Don't reinstall installed modules**

  If a module is already installed, there's no need of reinstalling it.

- **[FIX] Proper UI in forms**

  The colspan property of some UI elements were not correctly adjusted to the v16 sytem.

  This commits fixes it.

@Tecnativa

### DIFF
--- a/upgrade_analysis/blacklist.py
+++ b/upgrade_analysis/blacklist.py
@@ -1,4 +1,9 @@
-BLACKLIST_MODULES = []
+BLACKLIST_MODULES = [
+    "payment_alipay",
+    "payment_ogone",
+    "payment_payulatam",
+    "payment_payumoney",
+]
 
 # the hw_* modules are not affected by a migration as they don't
 # contain any ORM functionality, but they do start up threads that
--- a/upgrade_analysis/tests/test_module.py
+++ b/upgrade_analysis/tests/test_module.py
@@ -6,7 +6,7 @@ class TestUpgradeAnalysis(common.TransactionCase):
     def setUp(self):
         super().setUp()
         self.IrModuleModule = self.env["ir.module.module"]
-        self.product_module = self.IrModuleModule.search([("name", "=", "product")])
+        self.website_module = self.IrModuleModule.search([("name", "=", "website")])
         self.sale_module = self.IrModuleModule.search([("name", "=", "sale")])
         self.upgrade_analysis = self.IrModuleModule.search(
             [("name", "=", "upgrade_analysis")]
@@ -18,19 +18,19 @@ class TestUpgradeAnalysis(common.TransactionCase):
 
         wizard.select_odoo_modules()
         self.assertTrue(
-            self.product_module.id in wizard.module_ids.ids,
+            self.website_module.id in wizard.module_ids.ids,
             "Select Odoo module should select 'product' module",
         )
-
-        wizard.select_oca_modules()
-        self.assertTrue(
-            self.upgrade_analysis.id in wizard.module_ids.ids,
-            "Select OCA module should select 'upgrade_analysis' module",
-        )
+        # New patch avoids to reinstall already installed modules, so this will fail
+        # wizard.select_oca_modules()
+        # self.assertTrue(
+        #     self.upgrade_analysis.id in wizard.module_ids.ids,
+        #     "Select OCA module should select 'upgrade_analysis' module",
+        # )
 
         wizard.select_other_modules()
         self.assertFalse(
-            self.product_module.id in wizard.module_ids.ids,
+            self.website_module.id in wizard.module_ids.ids,
             "Select Other module should not select 'product' module",
         )
 
--- a/upgrade_analysis/tests/test_module.py
+++ b/upgrade_analysis/tests/test_module.py
@@ -21,19 +21,11 @@ class TestUpgradeAnalysis(common.TransactionCase):
             self.website_module.id in wizard.module_ids.ids,
             "Select Odoo module should select 'product' module",
         )
-        # New patch avoids to reinstall already installed modules, so this will fail
-        # wizard.select_oca_modules()
-        # self.assertTrue(
-        #     self.upgrade_analysis.id in wizard.module_ids.ids,
-        #     "Select OCA module should select 'upgrade_analysis' module",
-        # )
-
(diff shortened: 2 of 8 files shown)


legalsylvain:
hi @pedrobaeza. Thanks for working on this module.

> [FIX] Add to blacklist deprecated modules

:+1: 

> [FIX] Proper UI in forms

:+1: 

> [IMP] Don't reinstall installed modules. If a module is already installed, there's no need of reinstalling it.

I don't use this module very often, so I don't have a clear point of view. But I remember that the analysis was performed when modules already installed were reinstalled. (but I don't know why). 

I'm not available for the next two weeks. I can look at this topic during OCA days in a little more detail.

@StefanRijnhart perhaps you have an opinion on the subject?

pedrobaeza:
That part is not altered. The reinstallation takes place the same:

https://github.com/OCA/server-tools/blob/3a5305184a8eb87ce60b7acaa0b0465b1e5f8afc/upgrade_analysis/wizards/upgrade_generate_record_wizard.py#L41

This is only for avoiding to reinstall on the first wizard.

StefanRijnhart:
From https://github.com/OCA/server-tools/blob/15.0/upgrade_analysis/upgrade_log.py#L192-L197 I gather that not reinstalling all modules at this point might give different results if the code was updated in the mean time and new data records were added. This is a common situation in the timeframe surrounding a new Odoo release, so it's not completely safe. Maybe this can be made into an option defaulting to 'reinstall already installed modules'?

pedrobaeza:
Are you sure that statement is still applying? I did the analysis for 16.0, and didn't notice any problem. Anyway, it was just an improvement, so it's not the main fix at all, and I can remove it.

upgrade_analysis/blacklist.py
```diff
@@ -1,4 +1,9 @@
-BLACKLIST_MODULES = []
+BLACKLIST_MODULES = [
+    "payment_alipay",
```
legalsylvain, line 2: Hi. Naive question. How do you know this module is deprecated. it is installable in V16. the only way to find the information is to read the description of the manifest. 

https://github.com/odoo/odoo/blob/16.0/addons/payment_alipay/__manifest__.py#L8

upgrade_analysis/wizards/upgrade_install_wizard.py
```diff
@@ -34,7 +34,7 @@ class UpgradeInstallWizard(models.TransientModel):
     def _module_ids_domain(self, extra_domain=None):
         domain = [
             "&",
-            ("state", "not in", ["uninstallable", "unknown"]),
+            ("state", "not in", ["installed", "uninstallable", "unknown"]),
```
legalsylvain, line 38: I'm not sure to understand. how do you run the analysis of the ``base`` module, that is allways installed.

legalsylvain (commented):
>  [FIX] Add to blacklist deprecated modules

LGTM, but I wonder this list should be better in the apriori.py file of the openupgrade project. (to have all the merged / renamed / deprecated changes in the same place).
   
> [IMP] Don't reinstall installed modules

Not sure to understand. See inline.

> [FIX] Proper UI in forms

:+1:

StefanRijnhart:
> Are you sure that statement is still applying? I did the analysis for 16.0, and didn't notice any problem. Anyway, it was just an improvement, so it's not the main fix at all, and I can remove it.

This is a valid question. If it is safe to perform the analysis without a reinstall of installed modules (even if new xml records were added to these installed modules in the mean time), it should also be safe to skip this step entirely in the analysis process. We can then generate an analysis by only installing each module once on a fresh database.

This is also easily testable: does a one step analysis and a two step analysis both yield equivalent results?

hbrunn (approved):
I've scripted this for myself as creating all databases from scratch for all analysis runs, so I can't really say something about the reinstall question.

upgrade_analysis/blacklist.py
pedrobaeza: Yes, @legalsylvain, I found it by the hard way, when errors arose, as @hbrunn found out in #3086

upgrade_analysis/wizards/upgrade_install_wizard.py
pedrobaeza: This is for reinstalling things, so no need to reinstall what is already installed.

pedrobaeza:
@StefanRijnhart 

> This is also easily testable: does a one step analysis and a two step analysis both yield equivalent results?

Yes, I ran it and got the same results, that were reflected in https://github.com/OCA/OpenUpgrade/pull/4180.

upgrade_analysis/tests/test_module.py
```diff
@@ -18,19 +18,19 @@ def test_upgrade_install_wizard(self):
...
+            self.website_module.id in wizard.module_ids.ids,
             "Select Odoo module should select 'product' module",
         )
-
-        wizard.select_oca_modules()
-        self.assertTrue(
-            self.upgrade_analysis.id in wizard.module_ids.ids,
-            "Select OCA module should select 'upgrade_analysis' module",
-        )
+        # New patch avoids to reinstall already installed modules, so this will fail
```
StefanRijnhart: OK, you can remove this commented code given that reinstalling seems save.

pedrobaeza: Done @StefanRijnhart

StefanRijnhart (approved):

StefanRijnhart:
Thanks! The commit to not reinstall modules would have to be ported to17.0

pedrobaeza:
This is #3066 hitting

I'm going to deactivate such harmful test.

pedrobaeza:
Let's encrypt is saturated:

```
There were too many requests of a given type :: Service busy; retry later.
```

maybe we need to disable those tests as well:

/ocabot merge minor

pedrobaeza:
Test disabled.

/ocabot merge minor

Issue #2694, opened 2023-08-10
[16.0] Can't use base_fontawesome, not showing icons


iggups:
## Module
base_fontawesome

## Describe the bug
Tried to use the new icons in an XML view but I don't see any icon at all. Maybe I am using it wrong? I followed the documentation.

## To Reproduce
**Affected versions**: 16.0.5.15.4

1. Install module
2. Add `<button brand_icon="fa-amazon-pay" string=" fa-amazon-pay"/>` to a form view
3. Only string is shown, no icon

**Expected behavior**
Icon is shown

danielduqma:
Found similar bug, an easy way to reproduce is installing just `base_fontawesome` and `website` on a clean installation, for example on [runboat](http://oca-server-tools-16-0-3a5305184a8e.runboat.odoo-community.org/). Tested on linked runboat and right after installation icons of social networks on footer are broken:

  ![image](https://github.com/OCA/server-tools/assets/6009746/7f7f1f4e-2a95-4c4e-b7df-ca90a4c22495)

Looks like not all assets are loaded:

  ![image](https://github.com/OCA/server-tools/assets/6009746/3d61f6e9-a496-4acf-9108-65867b1800c9)

Could anyone take a look of this please @OCA/tools-maintainers?

Best regards.

thomaspaulb:
It seems to be something amiss in [web.assets_frontend](https://github.com/OCA/server-tools/blob/3a5305184a8eb87ce60b7acaa0b0465b1e5f8afc/base_fontawesome/__manifest__.py#L25).

When inspecting `/web/assets/debug/1/web.assets_frontend.css` on Runboat, `fontawesome.css` is included, but `all.css` isn't, and I think that's the one needed for the correct interpretation of the `fa-*` classes. In the backend, it does work.

danielduqma:
Solved in https://github.com/OCA/server-tools/pull/2740, could someone take a look please? Thanks!

iggups:
Hi folks,
I am still not seeing the icons in the odoo backend. I am using them as described in the OP.

I updated the module to 16.0.5.16.0 and wanted to use the new icons in a form view button like:

```
<button class="oe_stat_button" type="object" name="action_foobar" brand_icon="fa-amazon-pay">
   <field name="foobar_count" widget="statinfo" string="Foo Bar"/>
</button>
```

thomaspaulb:
@iggups I think it is now fixed

iggups:
[...]

PR #2740, opened 2023-10-25
[FIX] base_fontawesome: fixed replace files to show icons in website

Pablocce:
**Description:**
This pull request addresses an issue where icons were not displaying in the website module when the base_fontawesome module was installed. To reproduce the issue, follow these steps:

- Set up an Odoo environment running version 16.
- Install the base_fontawesome module.
- Install the website module.
- Access the website.
- Observe that the icons are not being displayed.

**Solution:**
To resolve this problem, we have made the following modifications:

- Updated the replace statement in the manifest file to load the custom CSS with the new version.
- Loaded the icon font of brands.
- Added the fa class to the corresponding font-family.

This should ensure that icons are displayed correctly in the Odoo website module when the base_fontawesome module is installed.

Original issue --> https://github.com/OCA/server-tools/issues/2694

### DIFF
--- a/base_fontawesome/__manifest__.py
+++ b/base_fontawesome/__manifest__.py
@@ -28,6 +28,10 @@
                 "web/static/src/libs/fontawesome/css/font-awesome.css",
                 "base_fontawesome/static/src/css/fontawesome.css",
             ),
+            "base_fontawesome/static/lib/fontawesome-5.15.4/css/all.css",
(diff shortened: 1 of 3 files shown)


thomaspaulb (approved):

xAdrianCif (approved):
Technical and functional review from differents componets, thanks for the work ❤️

PR #2742, opened 2023-10-27
[16.0][IMP] excel_import_export: possibility to apply unidecode

RodrigoBM:
Add the possibility to apply unidecode to the data to be set in the exporter.
It is possible that the file you export has only the characters ascii.

e.g.
unidecode.unidecode("ESPAÑA")
'ESPANA'
unidecode.unidecode("forçut")
'forcut'

It is possible that the system where you import the file only supports ascii characters.


xAdrianCif (approved):
Tech review ok! thanks!

liebana (approved):
Functional review OK

pedrobaeza (changes requested):
Adding extra dependencies in a merged module is not advisable, and in this case being an extra feature, it should be done in a new module `excel_import_export_unidecode` or similar.

RodrigoBM:
Hi @pedrobaeza,

I understand what you are saying, but I don't see a way to do this in another module, the methods are not very inheritable.

Best regards

pedrobaeza:
You can provide hooks in this base module for easing that.

RodrigoBM:
> You can provide hooks in this base module for easing that.

Done @pedrobaeza

requirements.txt
```diff
@@ -7,3 +7,4 @@ openpyxl
 openupgradelib
 pygount
 sentry_sdk<=1.9.0
+unidecode
```
pedrobaeza, line 10: Put this in the other commit.

pedrobaeza (commented):
The rest seems OK.

requirements.txt
RodrigoBM: Done, I forgot this in the other commit.

pedrobaeza (approved):
/ocabot merge patch

PR #2747, opened 2023-10-31
[10.0][ADD] base_external_system_[http|oauth]

NL66278:


thomaspaulb:
@NL66278 CI is broken and there is talk to phase out python2 odoo versions anyway. I'd say keep this as a private branch and re-propose to OCA upon migration

PR #2754, opened 2023-11-12
[17.0][MIG] session_db

sbidoul:


session_db/pg_session_store.py
```diff
@@ -0,0 +1,169 @@
...
+            except Exception:  # pylint: disable=except-pass
+                pass
+            self._cr = None
+
+    @with_lock
+    @with_cursor
+    def _setup_db(self):
+        self._cr.execute(
+            """
+                CREATE TABLE IF NOT EXISTS http_sessions (
```
ivantodorovich, line 100: This could be a good fit for an unlogged table.

session_db/pg_session_store.py
```diff
@@ -0,0 +1,169 @@
...
+                    sid varchar PRIMARY KEY,
+                    write_date timestamp without time zone NOT NULL,
+                    payload text NOT NULL
```
ivantodorovich, line 103: We could use a jsonb column here nowadays

ivantodorovich (approved):
LGTM 💯

non-blocking suggestions

session_db/pg_session_store.py
```diff
@@ -0,0 +1,169 @@
...
+    import gevent.lock
+
+    lock = gevent.lock.RLock()
+elif odoo.tools.config["workers"] == 0:
+    import threading
+
+    lock = threading.RLock()
+
+
+def with_lock(func):
```
ivantodorovich, line 29: This could possibly be replaced with `odoo.tools.func.locked`

session_db/readme/ROADMAP.md
```diff
@@ -0,0 +1,2 @@
+This module does not work with multi-threaded workers, so it requires
+workers \> 0.
```
ivantodorovich: Isn't this implemented now? Since https://github.com/OCA/server-tools/commit/8023bff1b654ab0d27a6aebcdee4f7edb9dd3d0b

ivantodorovich (approved):

session_db/pg_session_store.py
sbidoul: Hm, I don't think so. I want the sessions to be replicated to be available in case of failover.

session_db/pg_session_store.py
sbidoul: What benefit do you expect of a Jsonb column in this case?

session_db/readme/ROADMAP.md
sbidoul: I removed ROADMAP.md

session_db/pg_session_store.py
ivantodorovich: Given payload is a JSON, it simply seems right to use the column type made for this

sbidoul: I don't know. Intuitively I'd say a jsonb column would be less performant as postgres would need to parse it somehow. And we don't need to query the content of the sessions so that does not seem useful to me at first sight.

session_db/pg_session_store.py
sbidoul: Interesting, does not seem to be a 1:1 replacement at first sight but could be interesting to investigate in another PR.

adrienpeiffer (approved):

adrienpeiffer:
Let's merge it !

PR #2755, opened 2023-11-14
[17.0][MIG] auditlog: Migration to 17.0

rven:


rven:
A "big" remark here is that the methods `_patch_method` and `_revert_method` are not available anymore in Odoo.
I re-implemented them in the `base.py` file.
If somebody knows a better way, let me know!

pedrobaeza:
There should be then an alternative implementation. Have you dug into the commit removing it to check?

rven:
> There should be then an alternative implementation. Have you dug into the commit removing it to check?

Yes, they removed it, because it was only used in Odoo test classes.
Here they are using the patch method now from the test.common package, which is not the way to go I guess ;-)

pedrobaeza:
OK, I let others used to this module to check and give their opinion.

nilshamerlinck:
Hello @rven what about using the same approach as their fix on `base_automation` in [PR](https://github.com/odoo/odoo/pull/110370):

image

To avoid re-introducing these risky methods ;-)

rven:
@nilshamerlinck Thanks for the tip!

ahaly-compass:
Whats the status here?

esakarya:
Not working properly, when u try open Logs, Menü happen following error. also delete never logged,

 File "/opt/odoo/odoo-enterprise/odoo/models.py", line 2717, in read_group
    rows = self._read_group(domain, annoted_groupby.values(), annoted_aggregates.values(), offset=offset, limit=limit, order=orderby)
  File "/opt/odoo/odoo-enterprise/odoo/models.py", line 1906, in _read_group
    self.env.cr.execute(SQL("\n").join(query_parts))
  File "/opt/odoo/odoo-enterprise/odoo/sql_db.py", line 332, in execute
psycopg2.errors.SyntaxError: syntax error at or near "SELECT"
LINE 2: FROM SELECT 
             ^

rven:
@esakarya The error should be fixed now.

Are you sure the unlink of a record is not logged?
image

StefanRijnhart (approved):
Looking good, and seems to work fine on runboat.

auditlog/models/autovacuum.py
```diff
@@ -0,0 +1,37 @@
...
+        deadline = datetime.now() - timedelta(days=days)
+        data_models = ("auditlog.log", "auditlog.http.request", "auditlog.http.session")
+        for data_model in data_models:
+            records = self.env[data_model].search(
+                [("create_date", "<=", fields.Datetime.to_string(deadline))],
+                limit=chunk_size,
+                order="create_date asc",
+            )
+            nb_records = len(records)
+            with self.env.norecompute():
```
StefanRijnhart:
```
2024-02-02 12:22:58,062 3016808 WARNING 17auditlog py.warnings: /usr/lib/python3.10/contextlib.py:135: DeprecationWarning: `norecompute` is useless. Deprecated since 17.0.
  File "/home/odoo/server-tools/auditlog/tests/test_autovacuum.py", line 40, in test_autovacuum
    autovacuum_model.autovacuum(days=0)
  File "/home/odoo/server-tools/auditlog/models/autovacuum.py", line 34, in autovacuum
    with self.env.norecompute():
  File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
```

StefanRijnhart (commented):
Oh, tests emit a deprecation warning

StefanRijnhart:
If you want you can pick https://github.com/OCA/server-tools/pull/2814 here. It's a clean pick onto your branch and the issue still applies.

auditlog/models/autovacuum.py
rven: Removed!

StefanRijnhart (approved):
Thanks for the updates!

xFingerlinKx:
when will the version for 17 Odoo be available?

StefanRijnhart:
@xFingerlinKx as soon as more people have done a code review or a functional test. People like you!

StefanRijnhart:
@rven please check this fix of a regression introduced by the cache invalidation improvement: https://github.com/OCA/server-tools/pull/2834

rven:
@StefanRijnhart I added the specific commit into this PR, I had to change the test a little bit because of a ensure_one error.

StefanRijnhart:
@rven Great, thanks!

heccjj:
Is this a bug?
I override the method do_print_picking() of stock.picking model, and there is code like this:
self.write({'printed': True})
Then add and subscribe stock.picking auditlog rule.
When I print a stock picking, I will get an empty log, without log_line!

StefanRijnhart:
@heccjj Your change does not seem substantial, you just leave out actually rendering the report. Therefore, I tested this on this PR's runbot, and I can't reproduce:

image

So it does not seem like a bug in the module. Perhaps there is some other customization in your setup that interferes here.

heccjj:
@StefanRijnhart  Thanks for your reply. I try it again. If I print the picking slip first time, I get correct log, but if I print this picking slip again(field printed is already "True"), get an empty log.

StefanRijnhart:
@heccjj That makes sense, because the written values are only logged if the values differ from the current field values. The log is still present to represent the write itself (this is a design choice), but it is empty because there were no actual changes. So, not a bug.

DucTruongKomit (approved):
Look good

PR #2759, opened 2023-11-16
[14.0][IMP] base_changeset: Show change pop up in list view

victoralmau:
Superseed from 13.0: https://github.com/OCA/server-tools/pull/2184

Extra changes:
- Change the `get_fields_changeset_changes()` method to indicate the ids to be queried.
- Remove show changesets in tree views.
- Change so that in the form view only the data of the corresponding lines are requested.
- Refactor code (reduce code).

form

@Tecnativa TT45801


pedrobaeza:
With this, I'm concerned about the performance. Not sure if it's a good idea to put it on the tree. OK to put it in the lines o2m, but not on the tree...

victoralmau:
I am also not sure that it does not have performance problems displaying the information in the tree view (maybe someone can test it in a database with many changesets), now you get the data only from the records that are displayed (before it was done from all and probably that was the problem).

Showing the data in the tree view is something that was already in the v13 PR and I have kept, although I personally would not add the information in tree view.

victoralmau:
Removed display of changes in tree view.

PR #2760, opened 2023-11-17
[14.0][WIP] base_changeset: Allow one2many fields from rules

victoralmau:
Allow one2many fields from rules

Changes done:
- [x] Allow one2many fields from rules.
- [x] Show the virtual one2many line in the form view. 
ejemplo-1
ejemplo-2

- [x] Indicate in the one2many line that this record is to be deleted.
ejemplo-eliminar

Example use case: `order_line` from `purchase.order`

@Tecnativa TT45801


victoralmau:
@pedrobaeza what do you think about this approach?

Pending changes:
- Show the virtual one2many line in the form view. Where to indicate that the line is approved or cancelled? Always show it when access to the record or when clicking on some button?
- Indicate in the one2many line that this record is to be deleted. Where to indicate that the line is approved or cancelled?

PR #2762, opened 2023-11-18
[16.0][FIX] jsonifier: Call callable for unknown field

lmignon:
If a field name into the parser definition doesn't exist into the model but is resolved by a callable, call the method. Prior to this change, it was o more possible to define computed json value for custome keys


thomaspaulb:
@lmignon Could you rebase to retrigger tests?

lmignon:
@thomaspaulb rebase done

thomaspaulb (approved):

PR #720, opened 2017-01-28
Isolate database_cleanup tests

dreispt:
Fixes #689

PR #2763, opened 2023-11-24
[17.0][MIG] module_auto_update

fkantelberg:
I had to adjust some tests because:

- Odoo core removed the methods `_patch_method` and `_revert_method`
- There was no `i18n/en.po` in general


thomaspaulb:
@fkantelberg Could you retrigger the runboat with a rebase?

thomaspaulb:
@fkantelberg Looks like tests are failing in conjunction with `database_cleanup_test`

fkantelberg:
@thomaspaulb True and I wonder why. I suspect that the tests of `database_cleanup` aren't cleaning up the database afterwards. I do believe that the `module_auto_update` works just fine. Maybe we have to think about a `tearDown` function in `database_cleanup` to remove the test data. Why do you think?

thomaspaulb:
@fkantelberg After some investigation, found that [database_cleanup tests are doing a commit](https://github.com/OCA/server-tools/blob/ad3febf4a202255bef0612605c7c1f7bc241b791/database_cleanup/tests/common.py#L16)

And I think in previous versions this was [noted](https://github.com/OCA/server-tools/pull/2547) also and solved by running [tests in isolation](https://github.com/OCA/server-tools/pull/720) or [setting it as a rebel module](https://github.com/yankinmax/server-tools/pull/1/files). Maybe we can try the same here on 17.0?

fkantelberg:
Sure lets try the rebel one :)

thomaspaulb:
Cool, will you cherry pick it as part of this PR?

fkantelberg:
Seems to work. Not sure if I would rename the tests now to make it more descriptive.

thomaspaulb (approved):

thomaspaulb:
Renaming which tests in which way?

fkantelberg:
The tests are currently named:
- `tests / test with Odoo`
- `tests /test with OCB`
- `tests / test with Odoo`
- `tests /test with OCB`

The cherry pick duplicated the existing tests BUT the first to only test `database_cleanup` and the other 2 all other non-rebel modules. Not really descriptive but I could live with it.

thomaspaulb:
Ahhhh yes. Well if they can be renamed, I would name the second one "Rebel tests with Odoo", it even sounds cool. But I wouldn't really know where to put it. If you do, and it's easy, then why not.

fkantelberg:
It's defined in `.github/worksflows/test.yml`. I renamed the one which is only testing `database_cleanup` `Rebel test ...` because the cherry picked defined that modul as rebel :)

Thanks for the support.

thomaspaulb:
@fkantelberg I don't see the addon hash tests running here in 17.0 either, so maybe this needs a cherry pick of https://github.com/OCA/server-tools/pull/2847

fkantelberg:
I cherry picked the interesting parts of it and I see the tests now in the logs. Seems like Odoo runs tests differently than I do locally.

thomaspaulb:
What exactly did you cherry-pick though? I see the changes in the test file, but also [changes](https://github.com/OCA/server-tools/pull/2763/commits/29ac79dfac7e57b8a6a7bdb1fd2d0c800191fe2e) in Contributors and other readme files

fkantelberg:
cherry picked your linked commit. pre-commit did the rest but I squashed the commits.

dreispt (approved):

PR #2768, opened 2023-11-28
[16.0] auditlog: Log exports

tarteo:
image

### DIFF
--- a/auditlog/models/log.py
+++ b/auditlog/models/log.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+from odoo.tools.safe_eval import safe_eval
 
 
 class AuditlogLog(models.Model):
@@ -16,6 +17,7 @@ class AuditlogLog(models.Model):
     model_name = fields.Char(readonly=True)
     model_model = fields.Char(string="Technical Model Name", readonly=True)
     res_id = fields.Integer("Resource ID")
+    res_ids = fields.Char("Resource IDs")
     user_id = fields.Many2one("res.users", string="User")
     method = fields.Char(size=64)
     line_ids = fields.One2many("auditlog.log.line", "log_id", string="Fields updated")
@@ -48,6 +50,15 @@ class AuditlogLog(models.Model):
             model = self.env["ir.model"].sudo().browse(vals["model_id"])
             vals.update({"model_name": model.name, "model_model": model.model})
         return super().write(vals)
+
+    def show_res_ids(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "tree,form",
+            "res_model": self.model_id.model,
+            "domain": [("id", "in", safe_eval(self.res_ids))],
+        }
 
 
 class AuditlogLogLine(models.Model):
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -106,6 +106,15 @@ class AuditlogRule(models.Model):
         ),
         states={"subscribed": [("readonly", True)]},
     )
+    log_export_data = fields.Boolean(
+        "Log Exports",
+        default=True,
+        help=(
+            "Select this if you want to keep track of exports "
+            "of the model of this rule"
+        ),
+        states={"subscribed": [("readonly", True)]},
+    )
     log_type = fields.Selection(
         [("full", "Full log"), ("fast", "Fast log")],
         string="Type",
@@ -206,6 +215,12 @@ class AuditlogRule(models.Model):
                 model_model._patch_method("unlink", rule._make_unlink())
                 setattr(type(model_model), check_attr, True)
                 updated = True
+            #   -> export_data
+            check_attr = "auditlog_ruled_export_data"
+            if rule.log_export_data and not hasattr(model_model, check_attr):
+                model_model._patch_method("export_data", rule._make_export_data())
+                setattr(type(model_model), check_attr, True)
+                updated = True
         return updated
 
     def _revert_methods(self):
@@ -213,7 +228,7 @@ class AuditlogRule(models.Model):
         updated = False
         for rule in self:
             model_model = self.env[rule.model_id.model or rule.model_model]
-            for method in ["create", "read", "write", "unlink"]:
+            for method in ["create", "read", "write", "unlink", "export_data"]:
                 if getattr(rule, "log_%s" % method) and hasattr(
                     getattr(model_model, method), "origin"
                 ):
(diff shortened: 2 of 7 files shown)


thomaspaulb:
@tarteo Could you rebase?

tarteo:
> @tarteo Could you rebase?

Done!

auditlog/views/auditlog_view.xml
```diff
@@ -150,8 +162,21 @@
                                 attrs="{'invisible': [('model_id', '!=', False)]}"
                                 readonly="1"
                             />
-                            <field name="res_id" readonly="1" />
-                            <field name="name" readonly="1" />
+                            <field
+                                name="res_id"
+                                readonly="1"
+                                attrs="{'invisible': [('res_id', '=', 0)]}"
```
thomaspaulb, line 168: What about False or None?

auditlog/views/auditlog_view.xml
```diff
@@ -150,8 +162,21 @@
...
+                            />
+                            <field
+                                name="res_ids"
+                                readonly="1"
+                                attrs="{'invisible': [('res_ids', '=', False)]}"
+                            />
+                            <field
+                                name="name"
+                                readonly="1"
+                                attrs="{'invisible': [('name', '=', False)]}"
```
thomaspaulb: Why invisible? If name is empty, people at least can see that it's empty.

auditlog/models/rule.py
```diff
@@ -267,6 +282,31 @@ def get_auditlog_fields(self, model):
             if (not f.compute and not f.related) or f.store
         )
 
+    def _make_export_data(self):
+        """Instanciate a create method that log its calls."""
```
thomaspaulb: "...an export method that logs..."

auditlog/models/rule.py
```diff
@@ -507,20 +547,25 @@ def create_logs(
...
-                "res_id": res_id,
-                "method": method,
-                "user_id": uid,
-                "http_request_id": http_request_model.current_http_request(),
-                "http_session_id": http_session_model.current_http_session(),
-            }
-            vals.update(additional_log_values or {})
-            log = log_model.create(vals)
+
+            log_vals = {**vals, **{"name": res_name, "res_id": res_id}}
```
thomaspaulb: Reads a bit nicer as `{**vals, "name": res_name, "res_id": res_id}`

auditlog/models/rule.py
```diff
@@ -507,20 +547,25 @@ def create_logs(
...
+        vals = {
+            "model_id": model_id,
+            "method": method,
+            "user_id": uid,
+            "http_request_id": http_request_model.current_http_request(),
+            "http_session_id": http_session_model.current_http_session(),
+        }
+        vals.update(additional_log_values or {})
+        if method == "export_data":
+            vals.update({"res_ids": str(res_ids)})
```
thomaspaulb: Conceptual issue: if the ids that were exported are logged, but a record is deleted later on, you don't know what was exported. It would be nice (perhaps later, eg in roadmap) to be able to log also "something about the record" eg the display_name or such.

As a workaround, people could just add an unlink auditlog as well.

auditlog/models/log.py
```diff
@@ -49,6 +51,15 @@ def write(self, vals):
...
         return super().write(vals)
 
+    def show_res_ids(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "tree,form",
+            "res_model": self.model_id.model,
+            "domain": [("id", "in", safe_eval(self.res_ids))],
+        }
```
thomaspaulb, line 61: Perhaps to increase coverage, this function can just be called once from a test.

auditlog/models/log.py
```diff
@@ -49,6 +51,15 @@ def write(self, vals):
             vals.update({"model_name": model.name, "model_model": model.model})
         return super().write(vals)
 
+    def show_res_ids(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "view_mode": "tree,form",
+            "res_model": self.model_id.model,
+            "domain": [("id", "in", safe_eval(self.res_ids))],
```
thomaspaulb, line 60: `safe_eval` seems a bit overkill here, maybe just trim brackets and split on commas?

auditlog/models/rule.py
```diff
@@ -267,6 +282,31 @@ def get_auditlog_fields(self, model):
...
+        self.ensure_one()
+        log_type = self.log_type
+        users_to_exclude = self.mapped("users_to_exclude_ids")
+
+        def export_data(self, fields_to_export):
+            res = export_data.origin(self, fields_to_export)
+            self = self.with_context(auditlog_disabled=True)
+            rule_model = self.env["auditlog.rule"]
+            if self.env.user in users_to_exclude:
+                return res
```
thomaspaulb, line 296: Maybe can add a `sudo(demo_user)` export in the test as well, and beforehand exclude the demo user, then see if it refrains from logging.

thomaspaulb (changes requested):

auditlog/views/auditlog_view.xml
AungKoKoLin1997: I don't think we can use `False` because res_id is `integer` field and default is `0`.

AungKoKoLin1997 (approved):
Functional Review: LGTM

auditlog/models/rule.py
yostashiro:
```suggestion
            vals.update({"name": res_model, "res_ids": str(res_ids)})
```
Just to assign something as a name, or the log record shows as 'New' when opened in a form. A trivial point.
image

yostashiro (commented):
Nice improvement.

auditlog/models/log.py
tarteo: I think this is the clearest way of doing it.

auditlog/models/log.py
tarteo: I'm normally not in favor of creating trivial tests just to increase coverage but I can do this.

auditlog/models/rule.py
tarteo: Good idea 👍

auditlog/views/auditlog_view.xml
tarteo: True! For cases where the resource name is False 👍

auditlog/models/rule.py
tarteo: Oopsy 😅

auditlog/models/rule.py
tarteo: I'm not sure I understand what you mean, is this a blocker?

tarteo:
Wow, the time flies by. I come here again requesting your review 😬 @yostashiro @thomaspaulb 
Error seems to be unrelated to this: https://github.com/OCA/server-tools/actions/runs/16772374931/job/47490378259?pr=2768#step:8:4924

auditlog/models/rule.py
thomaspaulb: Not a blocker, just a suggestion on how to create a unit test that tests this behavior, in order to increase coverage.

tarteo: Ah okay! I've applied your suggestions, can you take a look again and see if you're content with the code 🙏

StefanRijnhart:
@thomaspaulb Is this one alright for you now?

@tarteo Needs another rebase because of conflicts.

auditlog/models/rule.py
thomaspaulb: @tarteo I don't see my suggestion applied in your unit test, but as I said it's not a blocker, it was just an idea how to increase coverage.

thomaspaulb:
I'm good, so let's rebase and merge.

tarteo:
@StefanRijnhart Done!

PR #2774, opened 2023-12-04
[14.0] compute_field_after_install: Add Module for computed field in background at installation

sebastienbeau:
Add module that differs the compute field in background (cron) when installing a new module.

When having huge database with million of records, installing a new module that add  computed stored field can take a long time. To avoid a waiting too long the update the idea is to compute the value after the installation.

@legalsylvain @bealdav this may interest you ;)


base_compute_field_after_install/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+Contributors
+------------
+
+* Sébastien BEAU <sebastien.beau@akretion.com>
```
legalsylvain:
```suggestion
* Sébastien BEAU <sebastien.beau@akretion.com>
```

header is added automatically.

legalsylvain (commented):
Looks Great ! cool R&D.
A design question : what about to use queue_job. That way, we can reduce the time to compute. Something like : 
- on install, create a new job per field to compute.
- in the job, compute XXX items. if there are remaining items, create a new job.

So the computation will be done as soon as possible. Also, computing 1000 field x can be very more longer that computing 1000 fields y.

Not a blocking point, though.

base_compute_field_after_install/readme/CONTRIBUTORS.rst
sebastienbeau: done

bealdav:
Queue job could be an option, yes.

In this module or in another additional one I don't know.

About module name. I remenber than Holger Brunn said "I don't know what base_ means in a module name"

I think he is right, in this case particurlarly.

bealdav:
compute field after install

is the right word combo : it names what it do, no more, no less


Thanks for this contribution

sebastienbeau:
I hesitate to depend on queue_job.
But finally I want to keep it as simple as possible (avoid extra dependency)
And keep a simple view to track that is the progress for the re-computation.
If you want to recompute it asap you can run it every minute and then inactivate it after the installation of a new module (or always keep it active because if they is nothing to recompute the cron will just run in a second).

For the name of the module what to you prefers ? "compute_field_after_install" is ok for me

rvalyi:
I'm afraid there might be situations where a module is updated in such a way that running the database for some minutes before the new fields are computed could lead to doing business with corrupted data or corrupting the database... 

Imagine in new version invoice total is now a computed stored field. Eventually you would ask the customer to pay 0 (empty value) while computation is delayed or even write this corrupted value somewhere else in the database...

This is probably a rare situation, but if you will be running years with such a module it could happen eventually and the risk is you might not get a clue why about the kind of heisenbugs it may lead to. What do you think?

Eventually this is mitigated with good logging and ways to force some computation immediately, like very big warning in the log and command line option to force immediate computes in some modules.

There are also Docker based "zero downtime deploys" technologies that could eventually be less dangerous that such approach:
https://dokku.com/docs/deployment/zero-downtime-deploys/
In this case the old code will be running on the new database schema before computes occurs. IMHO this is safer while still not 100% safe.

compute_field_after_install/models/recompute_field.py
```diff
@@ -0,0 +1,99 @@
...
+            else:
+                limit = task.step
+            while True:
+                _logger.info(
+                    "Recompute field %s for model %s in background. Offset %s",
+                    task.field,
+                    task.model,
+                    offset,
+                )
+                records = model.search(domain, limit=limit, offset=offset, order="id")
```
legalsylvain, line 58:
```suggestion
                records = model.search(domain, limit=limit, offset=offset, order="id desc")
```

I think we should compute first the last created items. Due to @rvalyi remarks, if we try to mark as paid invoices, for sure we will work on recent invoices, not invoices created in 2009. (well, it can occures, but less frequently).

This proposal should come with other changes, regarding last_id, etc..

legalsylvain (commented):
Please add a section in ROAMAP.rst. something like : 

know limits : 
Never use this module in an OpenUpgrade context.

rvalyi:
I think this module will make Odoo blazing fast on Odoo.sh! https://twitter.com/PCatinean/status/1731614421753958526?t=C4IRO676ixBHJt-N4Yr2fQ&s=19

etobella:
Just one question, wouldn't have more sense to use pre_init_hook and post_init_hooks in a proper way in order to avoid this extra engineering?

I mean, the same happens on OpenUpgrade. We create the field on pre-migration and update it in post-migration. Or it might even happen in a migration in a usual Database:

https://github.com/OCA/queue/tree/15.0/queue_job/migrations/15.0.1.1.0

Even so, we can find some examples inside OCA for this case using pre_init_hook:

https://github.com/OCA/sale-workflow/blob/4a72d881252062c8576e3a97cd2c8015d986d446/sale_manual_delivery/hook.py#L4

https://github.com/OCA/brand/blob/719ea9089a2607d5965687d474508efc85214144/product_brand_stock/__init__.py#L9

pedrobaeza:
I totally agree with @etobella, you should rationalize individually on each module:

- Pre-create the column in pre.
- If the computed doesn't make sense for old data, just do that.
- If you need the computed value, put an SQL in pre/post init hooks for filling the value.

With this module, all kind of side-effects or unexpected things will appear, so it's opening a gate to the hell.

legalsylvain:
> so it's opening a gate to the hell.

;-)

@sebastienbeau : maybe you could describe the use case you faced, that is not solved by "classic" way  ?
thanks !

compute_field_after_install/__manifest__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2016 Akretion (http://www.akretion.com)
+# Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Base compute field after install",
+    "summary": "Computed field are computed after the install process",
+    "version": "14.0.1.0.0",
+    "category": "e-commerce",
```
dreispt, line 9: Not really...

compute_field_after_install/models/recompute_field.py
dreispt: I agree, I prefer newest IDs to be recomputed first, rather than the oldest IDs.

compute_field_after_install/models/recompute_field.py
```diff
@@ -0,0 +1,99 @@
...
+    state = fields.Selection(
+        [
+            ("todo", "Todo"),
+            ("done", "Done"),
+        ]
+    )
+
+    @api.model
+    def _run_all(self):
+        return self.search([("state", "=", "todo")]).run()
```
thomaspaulb, line 36: If this compute took too long on module installation, it will also take too long in a cron job, unless it's run in batches. Can you support a batch_size=SOME_DEFAULT parameter? People can then adjust this in the scheduled action, before they enable it.

PR #2780, opened 2023-12-09
16.0 mig base_custom_info

rquevedo:
We made some significant advances but couldn’t continue, we left this PR here in case anybody would like to take it from here.


thomaspaulb:
Thanks! Since you're not going to finish it yourself, I'll close, and people will be able to find it in the closed PR's.

PR #2525, opened 2023-01-16
[12.0][UPD] Update from copier. Update pre-commit.

AungKoKoLin1997:
- Update dotfiles using copier and pre-commit
- Backport of https://github.com/OCA/server-tools/pull/2433 to fix letsencrypt tests
- Fix module_analysis tests by cherry-pick commit from  [13.0][FIX] pre-commit; letsencrypt tests; module_analysis tests; nsca_client tests #2518 

@qrtl

PR #2782, opened 2023-12-09
[12.0][REF] dotfiles: copier copy

dreispt:


thomaspaulb:
@dreispt Shouldn't we better finish https://github.com/OCA/server-tools/pull/2525, or a reincarnation of it?

dreispt:
Indeed.

PR #2788, opened 2023-12-19
[15.0][MIG] tracking_manager

CRogos:

PR #2789, opened 2023-12-19
[16.0][MIG] tracking_manager

CRogos:


CRogos:
@sebastienbeau @Kev-Roche
I've migrated tracking_manager to v15 which is working fine. https://github.com/OCA/server-tools/pull/2788
But this PR with the migration to v16 is not working. 

Maybe you have an hint what might be the problem. I am not that deep into this module yet.
Also a review of the v15 migration would be nice.

Kev-Roche:
Hi @CRogos, I don't have errors when I launch the tests locally, do you ?

CRogos:
I've fixed one issue yesterday, but indeed I do not get any error when running the tests locally. But when I run the tests locally with a debugger attached, the tests fail as well.
Maybe an issue because of a parallel execution? I am not that deep into odoo tests yet. 
I haven't tested the modules after the latest changes. Maybe it is already working and only the tests make trouble.

CRogos:
@Kev-Roche  I found the issue. When the module "account" is installed, additional messages are logged. I've separated the tests, so that tracking_manager is executed on its own.

In v15 the log message looked like this: 
image

In v16 two messages are added out of the messages when "account" is installed:
image

Kev-Roche:
@CRogos Nice ! 
I try the module fonctionnally, from my point of view, it was only a test issue.
Thank you for the migration work.

Kev-Roche (approved):

lucuswolfius (approved):
Code looks good, testing worked fine.

CRogos:
@pedrobaeza could you merge this and also v15 #2788 ?

pedrobaeza:
I don't have experience with this module and a PSC should validate it.

/ocabot migration tracking_manager

AnizR (approved):
The code looks good to me and I tried it.
It seems to work properly!

sebastienbeau:
LGTM (code review) thanks for the migration
/ocabot merge patch

PR #2795, opened 2023-12-29
[17.0][MIG]  base_view_inheritance_extension

alexis-via:
No code change.
This module is above my technical skills, so review carefully and make sure we didn't need to change anything.


alexis-via:
Can a maintainer have a look at it ? We have PRs that depend on this PR.

StefanRijnhart (approved):
Also above my technical skill level, but I checked the upstream ir.ui.view API that this module overrides and it does not have changed between 16 and 17.

thomaspaulb (approved):

PR #2802, opened 2024-01-09
[15.0][ADD] monitoring: Generic monitoring module with API

fkantelberg:
This module allows monitoring systems (like icinga, prometheus ...) to look inside of Odoo to watch for problem within instances and inform the proper people to handle them. For example your monitoring system can monitor if queue.jobs or the email transmissions are failing or you can receive an email instead.


em230418:
There should be historical data. For example to check when script started failing and then when continued to work fine.

em230418:
> There should be historical data. For example to check when script started failing and then when continued to work fine.

Nevermind. Did not understand purpose of this module correctly.

gurneyalex (approved):

PR #2807, opened 2024-01-16
[17.0][MIG] base_multi_image: Migration to 17.0

omalbastin:


omalbastin:
How do we add the odoo_test_helper package?

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

PR #2812, opened 2024-01-18
[16.0] [FIX] database_cleanup: override create in batch

ivantodorovich:
This suppresses the following WARNING:

```
The model odoo.addons.database_cleanup.models.purge_wizard is not overriding the
create method in batch
```

Note: There's no need to forward port this change because it has already been taken care of in the `17.0` migration https://github.com/OCA/server-tools/commit/fbb5dc1fe7167d1904fd79d67e51ab74583b040f

### DIFF
--- a/database_cleanup/models/purge_wizard.py
+++ b/database_cleanup/models/purge_wizard.py
@@ -25,7 +25,7 @@ class CleanupPurgeLine(models.AbstractModel):
     def purge(self):
         raise NotImplementedError
 
-    @api.model
+    @api.model_create_multi
     def create(self, values):
         # make sure the user trying this is actually supposed to do it
         if self.env.ref("base.group_erp_manager") not in self.env.user.groups_id:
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Don't understand why CI is red...

ivantodorovich:
Me neither 🥲 I'll check tomorrow

database_cleanup/models/purge_wizard.py
```diff
@@ -25,7 +25,7 @@ class CleanupPurgeLine(models.AbstractModel):
     def purge(self):
         raise NotImplementedError
 
-    @api.model
+    @api.model_create_multi
     def create(self, values):
```
legalsylvain, line 29:
```suggestion
    def create(self, vals_list):
```

nitpicking : by convention, AFAIK, the argument is vals_list when we use the decorator model_create_multi.

(same remarks below).
otherwise, LGTM.

legalsylvain (approved):

database_cleanup/models/purge_wizard.py
ivantodorovich: Yeah, I totally agree with you

The reason I didn't change it is because in 17.0 it's done exactly like this, without renaming the arguments. And so I figured this way there's absolutely nothing to forward-port.

But I'm not sure my reasoning is correct. I can change it if you think it's better

legalsylvain: If it's not upstream, it doesnt matter for me. Improvment should ne done on latest révision.

SirAionTech (approved):
Thanks!

PR #2578, opened 2023-03-07
[16.0][MIG] cron_daylight_saving_time_resistant

florian-dacosta:
Based on https://github.com/OCA/server-tools/pull/2289

The module allows to configure crons so they always run at the same hour, no matter the timezone and the change of hours.
For instance, if you are in Brussels and you have a cron that run every days at 11H (local time), when swtiching from summer time to winter time, the cron will be planned for the next day at 10H (local time)
At the contrary when switching from winter time to summer time, the cron that was configured to run at 11H will be planned for 12H.
If it is important that the cron always run at the same hour no matter the change of time, then you can check the box "daylight_saving_time_resistant" on the cron.

PR #2813, opened 2024-01-19
[16.0][IMP] tracking_manager: allow tracking readonly fields

CRogos:
@AnizR, @Kev-Roche, @sebastienbeau 

What is the reason for the limitation that read_only fields cannot be tracked? 
When there have a look at the Odoo native tracking, there are also read_only fields with tracking enabled:
image

In our case we what to track changes on an enterprise table account.asset, where most of the fields are read_only and can only be changed by a separate wizard. But through this wizard, user changes are possible and therefore tracking might be necessary.

Are there any doubt to remove the read_only filter?

### DIFF
--- a/tracking_manager/models/ir_model.py
+++ b/tracking_manager/models/ir_model.py
@@ -100,21 +100,27 @@ class IrModel(models.Model):
     def _default_automatic_custom_tracking_domain_rules(self):
         return {
             "product.product": [
+                ("readonly", "=", False),
                 "|",
                 ("ttype", "!=", "one2many"),
                 ("name", "in", ["barcode_ids"]),
             ],
             "sale.order": [
+                ("readonly", "=", False),
                 "|",
                 ("ttype", "!=", "one2many"),
                 ("name", "in", ["order_line"]),
             ],
             "account.move": [
+                ("readonly", "=", False),
                 "|",
                 ("ttype", "!=", "one2many"),
                 ("name", "in", ["invoice_line_ids"]),
             ],
-            "default_automatic_rule": [("ttype", "!=", "one2many")],
+            "default_automatic_rule": [
+                ("ttype", "!=", "one2many"),
+                ("readonly", "=", False),
+            ],
         }
 
     @api.depends("automatic_custom_tracking")
@@ -176,7 +182,7 @@ class IrModelFields(models.Model):
         for record in self:
             record.native_tracking = bool(record.tracking)
 
-    @api.depends("readonly", "related", "store")
+    @api.depends("related", "store")
     def _compute_trackable(self):
         blacklists = [
             "activity_ids",
@@ -185,13 +191,9 @@ class IrModelFields(models.Model):
             "message_main_attachment",
             "message_main_attachement_id",
         ]
+
         for rec in self:
-            rec.trackable = (
-                rec.name not in blacklists
-                and rec.store
-                and not rec.readonly
-                and not rec.related
-            )
+            rec.trackable = rec.name not in blacklists and rec.store and not rec.related
 
     def write(self, vals):
         custom_tracking = None
--- a/tracking_manager/migrations/16.0.1.0.3/post-migration.py
+++ b/tracking_manager/migrations/16.0.1.0.3/post-migration.py
@@ -1,0 +1,21 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_model_fields SET trackable = false;
+        UPDATE ir_model_fields SET trackable = true
+            WHERE name NOT IN (
+                'activity_ids',
+                'message_ids',
+                'message_last_post',
+                'message_main_attachment',
+                'message_main_attachement_id'
+            )
+            AND store AND related IS NULL;
+        """,
+    )


legalsylvain (approved):

tracking_manager/models/ir_model.py
```diff
@@ -159,7 +158,6 @@ class IrModelFields(models.Model):
     )
     trackable = fields.Boolean(
         compute="_compute_trackable",
-        store=True,
```
AnizR: Why did you remove store=True?

I think that you may need to add something in the depends of your compute if the field isn't (re-)computed correctly.

CRogos: The trackable attribute is already computed, but with this PR we change the implementation how it is computed. 
Changing the implementation does not change the stored value automatically.
I agree another options would be an migration update script to recalculate all values. 

In think this is the only place where trackable is used. I am not sure if this syntax requires a stored field. But this could also be replaced by an lambda. Beside not changing it, I do not see the benefit of storing this simple value. 

https://github.com/OCA/server-tools/blob/485de9f8f8bba0df1fb905a47bc99e5cccb434a3/tracking_manager/models/ir_model.py#L128-L135

AnizR: I think that a migration script might be needed like you suggested but I am not 100% sure about that..

CRogos: Maybe there is a 3rd opinion over the weekend. I am personally not so a big fan of store=True on computed values. But indeed, changing it has also its risks.

Kev-Roche: I ll prefer to keep it stored because in some cases, we want to save a custom choice of fields tracked, and this independently to its kind (read only, related, ...).

Kev-Roche (changes requested):
I m not sure about this. 
In a way, the behavior of readonly field is known and not changed by users. In order to avoid some noise with too many tracked fields by default, I will prefer to not include the readonly fields as tracked. 
IMHO, maybe just set as tracked the relevant fields to follow, as you mention in your specific case, will be enough.

tracking_manager/models/ir_model.py
CRogos: I think "trackable" is only used as a filter when showing all trackable fields in the list. If this value is False, the field is not in the list and you cannot activating the track. There is also no possibility to change this value in the UI.
image

I do not see why this field needs to be persistent, but I agree changing it could also have unwanted side effects. But I did not have time yet to write the necessary upgrade script, to recalculate all values with the update. Without an update of the stored values, removing the readonly filter has no effect.

CRogos:
> I m not sure about this. In a way, the behavior of readonly field is known and not changed by users. In order to avoid some noise with too many tracked fields by default, I will prefer to not include the readonly fields as tracked. IMHO, maybe just set as tracked the relevant fields to follow, as you mention in your specific case, will be enough.

I only give the user the option to track fields, and do not activate the tracking. (trackable vs. tracking_custom field)
Do you still have concerns?

Kev-Roche:
> > I m not sure about this. In a way, the behavior of readonly field is known and not changed by users. In order to avoid some noise with too many tracked fields by default, I will prefer to not include the readonly fields as tracked. IMHO, maybe just set as tracked the relevant fields to follow, as you mention in your specific case, will be enough.
> 
> I only give the user the option to track fields, and do not activate the tracking. (trackable vs. tracking_custom field) Do you still have concerns?

My Bad, you are right, sorry I confound the two !

Kev-Roche:
> > I m not sure about this. In a way, the behavior of readonly field is known and not changed by users. In order to avoid some noise with too many tracked fields by default, I will prefer to not include the readonly fields as tracked. IMHO, maybe just set as tracked the relevant fields to follow, as you mention in your specific case, will be enough.
> 
> I only give the user the option to track fields, and do not activate the tracking. (trackable vs. tracking_custom field) Do you still have concerns?

But if you set tracking on a new model or just update it with update_custom_tracking button, it will add all the readonly fields as tracked and we don't want too many automatic and unwanted tracked fields. Maybe just `trackable` as readonly=False and store will be enough to deal with specific cases, or making optional the readonly fields as trackable ?

CRogos:
I've tested it on res.partner and there are 74 fields tracked instead of 59. Yes these are 15 fields more.
How about initializing the default domain this way to prevent readonly fields from automatic configuration?
image

CRogos:
I've now re-added the store flag and added an upgrade script to recompute the existing value.

tracking_manager/models/ir_model.py
```diff
@@ -73,8 +73,7 @@ def _get_model_tracked_by_o2m(self):
                 tracked_fields = custom_tracked_fields[model.model]
             else:
                 tracked_fields = model.field_id.filtered(
-                    lambda s: not s.readonly
-                    and not s.related
+                    lambda s: not s.related
```
Kev-Roche: It will activate the tracking on all readonly fields which are related to the main model by a o2m field. 
No need code here, for exemple a sale order with lines (o2M field). You have to activate the custom tracking on the sale.order.line model and select the pertinent fields to track.

tracking_manager/models/ir_model.py
```diff
@@ -100,21 +99,27 @@ def _compute_automatic_custom_tracking(self):
...
             "account.move": [
+                ("readonly", "=", False),
                 "|",
                 ("ttype", "!=", "one2many"),
                 ("name", "in", ["invoice_line_ids"]),
             ],
-            "default_automatic_rule": [("ttype", "!=", "one2many")],
+            "default_automatic_rule": [
+                ("ttype", "!=", "one2many"),
+                ("readonly", "=", False),
```
Kev-Roche, line 122:
```suggestion
                ("readonly", "=", True),
```
or delete it, in order to prevent readonly fields from automatic tracking.

Kev-Roche (changes requested):
as you said "to prevent readonly fields from automatic configuration", my main point is to keep the configuration and reading of tracking fields simple.
From my opinion, the tracking of readonly fields is marginal and can not be the main rule of this module. 

cc @sebastienbeau

tracking_manager/models/ir_model.py
CRogos: Are you sure? This should prevent adding readonly fields by default.
image
image

tracking_manager/models/ir_model.py
CRogos: I'll revert this change. I am not that deep into this part.

tracking_manager/models/ir_model.py
Kev-Roche: My bad, you are right !

CRogos:
@Kev-Roche I think everything is done now. Could you update your review to approved?

CRogos:
I think everything is final now. Could you review again?

CRogos:
@Kev-Roche @AnizR @legalsylvain could update your reviews?

CRogos:
@Kev-Roche 
> From my opinion, the tracking of readonly fields is marginal and can not be the main rule of this module.

Unfortunately the account_asset module of odoo enterprise is one, were the fields are readonly and the user changes the fields via a wizard. So technically they are readonly, but actually they are not and they are very important to track.

If you really see a risk, I could also add a global setting that needs to be enabled first before allowing to track readonly attributes. But "Keep the configuration simple" could also mean not to add this global setting ;-) . 

@pedrobaeza do you have an opinion on this?

Kev-Roche:
> @Kev-Roche
> 
> > From my opinion, the tracking of readonly fields is marginal and can not be the main rule of this module.
> 
> Unfortunately the account_asset module of odoo enterprise is one, were the fields are readonly and the user changes the fields via a wizard. So technically they are readonly, but actually they are not and they are very important to track.
> 
> If you really see a risk, I could also add a global setting that needs to be enabled first before allowing to track readonly attributes. But "Keep the configuration simple" could also mean not to add this global setting ;-) .
> 
> @pedrobaeza do you have an opinion on this?

I m good with the current way

Kev-Roche (approved):

pedrobaeza:
I think it's usual that some fields are readonly, but unlocked on certain conditions (like state and having `force_save="1"` parameter) or modified by code, so they should be tracked as well.

CRogos:
@sebastienbeau could you merge?

CRogos:
@Kev-Roche do you have an idea what's the issue with the merge? Error does not look related to this change? Shall I rebase or could you give it another try?

Kev-Roche:
@CRogos I have no Idea, don't see any related reason... can you rebase please, then I ll try again.

CRogos:
This themes to be the problem: https://github.com/OCA/server-tools/pull/2578

CRogos:
@Kev-Roche the problem with the repo has been solved. Could you retry the merge?

PR #2814, opened 2024-01-19
[160][FIX] auditlog: prevent removal of x2many values from inaccessible companies

StefanRijnhart:
Having a product.template with taxes from multiple companies assigned, let a user without all of these companies enabled set the taxes of that product.

Before this change, the taxes related to inaccessible companies will be inadvertently removed.

Based on https://github.com/OCA/server-tools/pull/2628 by @rven

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -387,6 +387,9 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            # Prevent the cache of modified fields from being poisoned by
+            # x2many items inaccessible to the current user.
+            self.invalidate_recordset(vals.keys())
             result = write_full.origin(self, vals, **kwargs)
             new_values = {
                 d["id"]: d
--- a/auditlog/tests/__init__.py
+++ b/auditlog/tests/__init__.py
@@ -1,3 +1,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from . import test_auditlog
 from . import test_autovacuum
+from . import test_multi_company
--- a/auditlog/tests/test_multi_company.py
+++ b/auditlog/tests/test_multi_company.py
@@ -1,0 +1,115 @@
+from unittest.mock import patch
+
+from odoo.fields import Command
+from odoo.tests.common import TransactionCase
+
+from odoo.addons.base.models.res_users import Groups
+
+
+class TestMultiCompany(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Disarm any existing auditing rules.
+        cls.env["auditlog.rule"].search([]).unlink()
+        cls.env["auditlog.log"].search([]).unlink()
+        # Set up a group with two share users from different companies
+        cls.company1 = cls.env["res.company"].create({"name": "c1"})
+        cls.company2 = cls.env["res.company"].create({"name": "c2"})
+        cls.group = cls.env["res.groups"].create({"name": "g1", "share": True})
+        cls.user1 = cls.env["res.users"].create(
+            {
+                "name": "u1",
+                "login": "u1",
+                "groups_id": [Command.set(cls.group.ids)],
+                "company_ids": [Command.set(cls.company1.ids)],
+                "company_id": cls.company1.id,
+            }
+        )
+        cls.user2 = cls.env["res.users"].create(
+            {
+                "name": "u2",
+                "login": "u2",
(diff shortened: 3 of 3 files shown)


hbrunn (approved):

auditlog/models/rule.py
```diff
@@ -387,6 +387,9 @@ def write_full(self, vals, **kwargs):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            # Prevent the cache of modified fields from being poisoned by
+            # x2many items inaccessible to the current user.
+            self.invalidate_recordset(vals.keys())
```
SirAionTech, line 392:
```suggestion
            self.invalidate_recordset(fnames=fields_list)
```
All the fields read with `sudo` have a cache value that is not what the current user would read, not only the ones in `vals`.
Shouldn't we reset the cache of all those fields instead of only the ones that are being written?

SirAionTech (approved):
Thanks!
I reviewed the code (big thanks for the test) and verified locally that the issue is fixed.

auditlog/models/rule.py
StefanRijnhart: @SirAionTech it's a good question. However, invalidating more fields might have an additional performance impact, and it's not clear to me under what circumstances these alternate cached values would cause an issue outside of the issue fixed here. Do you have a concrete scenario?

SirAionTech: > invalidating more fields might have an additional performance impact

I know, I have looked for an alternative to invalidating the cache (maybe a per-user cache?) and have found nothing useful.

> Do you have a concrete scenario?

A concrete scenario would be something like:
https://github.com/odoo/odoo/blob/1b1ee1b04948708fbcd9ccf78432183d2458b209/addons/account/models/account_journal.py#L560
an override of `write` wants to read the value of a field that is not in `vals`, after calling `super`.
It would probably read what `sudo` read a few lines above this one, instead of what the current user would read; if that's the case, we probably have to reset the cache everytime `sudo` reads anything.
What do you think?

Just to be clear: I still think the fix is fine (the review is an approval), I am just trying to understand a bit more of how the ORM works.

StefanRijnhart: Yes, I appreciate the approval!

As for your example, I don't see how the values read with sudo() by the auditlog override influences the fields in that section. Am I missing something?

I'm thinking that the only fields that will be different are these multicompany x2many fields, and if an override does something with those it implies a multicompany approach using sudo() in the first place.

SirAionTech: > As for your example, I don't see how the values read with sudo() by the auditlog override influences the fields in that section. Am I missing something?

It's more likely that I am missing something, this is how I think it works:

1. `sudo` reads the fields and puts the read value in the cache
2. every following access to the field sees the cached value

IF (notice the big IF) that's how it works, then the scenario I have linked has an issue.

StefanRijnhart: I agree with your analysis, but I think that in the linked code the consulted data like the currency_id of a journal will be the same when read with or without sudo so I'm wondering if it makes for a good example.

SirAionTech: > I agree with your analysis, but I think that in the linked code the consulted data like the currency_id of a journal will be the same when read with or without sudo so I'm wondering if it makes for a good example.

I didn't find an example but it probably would happen with `company_dependent` fields or fields that are accessible only by specific `groups`.

StefanRijnhart: Yes, those are interesting cases. Company dependent fields depend on the company which is kept when switching to sudo regardless of the default company of the superuser so it should not be an issue.

Fields protected by a groups attribute can indeed be cached when switching to sudo (in auditlog or anywhere else for that matter) and are from then on accessible by the non-sudo user.
```
>>> action = env["ir.actions.server"].search([], order="id asc", limit=1).with_user(env.ref("base.user_demo"))
>>> action.code
  File "/home/odoo/odoo/models.py", line 2944, in check_field_access_rights
    raise AccessError(_(
odoo.exceptions.AccessError: The requested operation can not be completed due to security restrictions.
>>> action.sudo().code
"\nconfig = model.next() or {}\nif config.get('type') not in ('ir.actions.act_window_close',):\n    action = config\n"
>>> action.code
"\nconfig = model.next() or {}\nif config.get('type') not in ('ir.actions.act_window_close',):\n    action = config\n"
```

If auditlog would use the sudo read for `read` operation rules this might be possible to exploit, but auditlog only uses the sudo read for `write` operations, and at the moment I find it difficult to imagine how this could be used to effectively leak information.

PR #2819, opened 2024-01-26
[16.0][IMP] base_time_window: Add helper function to get next Date matching weekday

grindtildeath:


base_time_window/models/time_weekday.py
```diff
@@ -60,3 +60,20 @@ def unlink(self):
...
+        """
+        self.ensure_one()
+        if not date_from:
+            date_from = fields.Date.today()
+        next_date = date_from
+        if next_date.weekday() == int(self.name) and not include_date_from:
+            next_date = fields.Date.add(next_date, days=1)
+        while next_date.weekday() != int(self.name):
+            next_date = fields.Date.add(next_date, days=1)
+        return next_date
```
StephaneMangin:
```suggestion
        next_date = date_from
        if next_date.weekday() == int(self.name) and not include_date_from:
            next_date = fields.Date.add(next_date, days=1)
        return fields.Date.add(next_date, days=7)
```

Or even better readability using your logic
```suggestion
        same_day_as_self = next_date.weekday() == int(self.name)
        next_weekday_delta = 0
        # Select next day as we do not want the current one
        if same_day_as_self and not include_date_from:
            next_weekday_delta += 1
        return fields.Date.add(date_from, days=next_weekday_delta)
```

Or by changing logic and using timedelta which are very readable too IMHO
```suggestion
        exclude_condition = (
            date_from.weekday() == int(self.name)
            and not include_date_from
        )
        next_weekday_delta = 1 if exclude_condition else 0
        return date_from + timedelta(days=next_weekday_delta)
```

StephaneMangin (commented):
The next week day is always 1 days later, na ? If you don't include the current day, it's 7 days more ?

base_time_window/models/time_weekday.py
StephaneMangin: I think the last one improves readability by a factor of + infinity ^^ And if it is not your use case, be aware to be readable because i didn't get it.

base_time_window/tests/test_weekday.py
```diff
@@ -0,0 +1,61 @@
...
+        # 2024-01-01 is Monday, next Saturday is 2024-01-06
+        self.assertEqual(
+            self.saturday._get_next_weekday_date(Date.to_date("2024-01-01")),
+            Date.to_date("2024-01-06"),
+        )
+        # 2024-01-01 is Monday, next Sunday is 2024-01-07
+        self.assertEqual(
+            self.sunday._get_next_weekday_date(Date.to_date("2024-01-01")),
+            Date.to_date("2024-01-07"),
+        )
```
StephaneMangin, line 61:
```suggestion
    def test_next_weekday_date_same_day_included(self):
        """include_date_from to True should return the same day"""
        # 2024-01-01 is Monday, next Monday (including date_from) is 2024-01-01
        self.assertEqual(
            self.monday._get_next_weekday_date(Date.to_date("2024-01-01")),
            Date.to_date("2024-01-01"),
        )
        
    def test_next_weekday_different_day_included(self):
        # 2024-01-01 is Monday, next Tuesday is 2024-01-02
        self.assertEqual(
            self.tuesday._get_next_weekday_date(Date.to_date("2024-01-01")),
            Date.to_date("2024-01-02"),
        )
        
    def test_next_weekday_different_day_not_included(self):
        """include_date_from to False should not alter next week day if days are different"""
        # 2024-01-01 is Monday, next Tuesday (excluding date_from) is 2024-01-02
        self.assertEqual(
            self.tuesday._get_next_weekday_date(Date.to_date("2024-01-01")),
            Date.to_date("2024-01-02"),
        )
        
    def test_next_weekday_same_day_not_included(self):
        # 2024-01-01 is Monday, next Monday (excluding date_from) is 2024-01-08
        self.assertEqual(
            self.monday._get_next_weekday_date(
                Date.to_date("2024-01-01"), include_date_from=False
            ),
            Date.to_date("2024-01-08"),
        )
```

StephaneMangin (commented):
One option is changed, one different test should be added. And testing all days is not relevant i presume

base_time_window/tests/test_weekday.py
StephaneMangin: On the proposed test_next_weekday_same_day_not_included i definitly don't get it with your PR information. Very sorry

Understood use case regarding the tests:
 - get the next day of the date provided by default.
 - if included and identical get the current day (?!)
 - if not included and identical get the next same week day one week later

base_time_window/models/time_weekday.py
gurneyalex:
```
next_weekday_delta = int(self.name) - date_from.weekday()
if next_weekday_delta < 0:
    next_weekday_delta += 7
elif next_weekday_delta == 0 and not_include_date_from:
    next_weekday_delta += 7
return datefrom + timedelta(days=next_weekday_delta)
```

At least this is what I would write from what I understand of the docstring.

grindtildeath:
@StephaneMangin @gurneyalex Thanks for your comments.

The idea is only to get the next date matching the weekday. I structured the tests to make sure it works with whichever weekday is used and if the target is before or after the date from during a single week. Hence why it looks a bit repetitive, but IMO it's fine as long as we cover everything and we don't rely on the date when the test is executed.

The function goal is to return the "next date" for a selected weekday as for example:
- Next Monday after today
- Next Monday after YYYY-MM-DD

The include parameters allows to return the same date passed as parameter or look for the "next one", such as:
- Today is Monday and I want the function to return Today when looking for a Monday
- Today is Monday and I want the function to return Monday next week when looking for a Monday

Now if the "from date" is a Thursday, and I want the next Monday, the function will need to add 4 days to the date from to find the next Monday, so it seems to me the propositions from @StephaneMangin won't work as it's always adding 0, 1 or 7 days.
But it looks to me like @gurneyalex 's proposal would work, so I'll ditch the while loop, as I guess it must be much more performant.

base_time_window/models/time_weekday.py
StephaneMangin: @gurneyalex Nice to directly compute the delta

grindtildeath:
@thomaspaulb what's the reason for adding the "Needs fixing" here?

@gurneyalex @StephaneMangin Can you please update your review?

thomaspaulb:
@grindtildeath It was based on your comment of "I'll ditch the while loop, as I guess it must be much more performant." but seeing no commits after that, so I assumed that's a WIP.

grindtildeath:
@thomaspaulb Thanks for checking. Actually there's something weird because I did amend my commit after the reviews (I'm not sure if I did before or after my last comment) but the commit still appears before all the conversation, which I agree looks like if I didn't take care of these, what I did.

base_time_window/models/time_weekday.py
```diff
@@ -60,3 +60,20 @@ def unlink(self):
...
+
+        :param date_from: Date object from which we start searching for next weekday
+        :param include_date_from: Allows to return date from if it's same weekday
+        :return Date object matching the weekday
+        """
+        self.ensure_one()
+        if not date_from:
+            date_from = fields.Date.today()
+        next_date = date_from
+        if next_date.weekday() == int(self.name) and not include_date_from:
```
jbaudoux: Can we just do
```suggestion
        if not include_date_from:
```

grindtildeath: Must be fixed in fd25b083bbeb86b47a4e771ca3652e4c4d441276 , IDK why but it seems I forgot to push it.

thomaspaulb:
pre-commit

gurneyalex (approved):

cyrilmanuel:
@grindtildeath  can you update the pre-commit please ?

PR #2820, opened 2024-01-30
[14.0][FIX] base_name_search_improved: avoid error search type query

Saran440:
Step to test:
1. install module `account` and `base_name_search_improved`
2. Go to Settings > Smart Searches > active model `Analytic Account` and add somefield to search
Selection_002
3. Create new invoice/bill and search value in analytic account. it will error.
Selection_003
Selection_004

it error because some module like `account.analytic.account` will return result to query not list.

### DIFF
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -89,6 +89,11 @@ def patch_name_search():
             # Support a list of fields to search on
             all_names = _get_rec_names(self.sudo())
             base_domain = args or []
+
+            # Check type res, avoid error some model.
+            if not isinstance(res, list):
+                res = []
+
             # Try regular search on each additional search field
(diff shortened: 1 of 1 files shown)


StefanRijnhart (changes requested):
Looks like in Odoo 14, `_name_search` is already refactored to return a Query so you need this fix from the 16.0 migration: https://github.com/adhoc-dev/server-tools/commit/1d55e6280aab73fb84236480ddd304e29ea4721f#diff-1e21cf02cf8be2b178bb9fb61f5d1b9690cef4f747ff05121854535604cdb64f

pedrobaeza:
Please review #2940 that should solve all the problems.

StefanRijnhart:
Closing in favour of https://github.com/OCA/server-tools/pull/2940. Thanks for noting the issue and proposing your fix though!

PR #2824, opened 2024-02-02
[16.0][IMP] base_view_inheritance_extension: attrs_domain_add

BT-cserra:
New operation 'attrs_domain_add', used to add a domain to the attribute 'attrs'.

### DIFF
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -190,6 +190,42 @@ class IrUiView(models.Model):
         return source
 
     @api.model
+    def inheritance_handler_attributes_attrs_domain_add(self, source, specs):
+        """Implement attrs_domain_add
+
+        <attribute name="$attribute" operation="attrs_domain_add"
+                   key="$attrs_key" join_operator="OR">
+            $domain_to_add_to_attrs_key
+        </attribute>
+        """
+        node = self.locate_node(source, specs)
+        for attribute_node in specs:
+            attribute_name = attribute_node.get("name")
+            key = attribute_node.get("key")
+            join_operator = attribute_node.get("join_operator") or "AND"
+            old_value = node.get(attribute_name) or ""
+            if old_value:
+                old_value = ast.literal_eval(
+                    self.var2str_domain_text(old_value.strip())
+                )
+                old_domain_attrs = old_value.get(key)
+                new_domain = ast.literal_eval(
+                    self.var2str_domain_text(attribute_node.text.strip())
+                )
+                if join_operator == "OR":
+                    new_value = expression.OR([old_domain_attrs, new_domain])
+                else:
+                    new_value = expression.AND([old_domain_attrs, new_domain])
+                old_value[key] = new_value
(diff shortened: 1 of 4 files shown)


BT-cserra:
Hello @sergio-teruel , @carlosdauden , @ivantodorovich , I added a new option to extend the domains _inside_ `attrs`. Otherwise, that domain has to be substituted as a whole. Could you please have a look and tell me what do you think?

BT-rmartin (changes requested):
Really interesting addition Carlos!
You don't need to bump the version. It's done automatically when merging. 
Would it make sense to add also the operator NOT?

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -189,6 +189,42 @@ def inheritance_handler_attributes_domain_add(self, source, specs):
...
+            key = attribute_node.get("key")
+            join_operator = attribute_node.get("join_operator") or "AND"
+            old_value = node.get(attribute_name) or ""
+            if old_value:
+                old_value = ast.literal_eval(
+                    self.var2str_domain_text(old_value.strip())
+                )
+                old_domain_attrs = old_value.get(key)
+                new_domain = ast.literal_eval(
+                    self.var2str_domain_text(attribute_node.text.strip())
```
BT-rmartin, line 213: I would first normalize the domain just in case before appending new ones with the method normalize_domain of expression

BT-etejeda: I think that odoo does a normalization when in the call to OR or AND https://github.com/odoo/odoo/blob/16.0/odoo/osv/expression.py#L264

BT-etejeda (approved):
LGTM

BT-etejeda:
> Really interesting addition Carlos! You don't need to bump the version. It's done automatically when merging. Would it make sense to add also the operator NOT?

Maybe that would be another different functionality (instead of attrs_domain_add, maybe attrs_domain_negation), we would no longer be combining a new domain with an existing one, but to negate the existing domain and perhaps with another logic.

ping @BT-cserra

BT-cserra:
> Really interesting addition Carlos! You don't need to bump the version. It's done automatically when merging. Would it make sense to add also the operator NOT?

The other operators only provide the OR and the AND, so I followed the trend here. I think adding the NOT would complicate things because the NOT is not a combination of one domain with the existing one, but a negation of the existing one. I prefer to leave it out.

Thanks for the reminder about the version, I'll do a patch on it.

base_view_inheritance_extension/models/ir_ui_view.py
BT-cserra: Yes, there seems to be a discrepancy here between the documentation in the doc-string and the implementation. @BT-etejeda since you found it maybe you can do a PR to Odoo to correct it.

BT-cserra:
Note: just tested it in a real project and it works.

BT-rmartin (approved):
LGTM!

carlosdauden (approved):
Code LGTM

PR #2828, opened 2024-02-05
[16.0][FIX] dbs value has to be kept in case of special proxy config

cvinh:


cvinh:
@thomaspaulb @hbrunn same fix as in v14
https://github.com/OCA/server-tools/pull/2353
Thanks

thomaspaulb (approved):

hbrunn (approved):

PR #2833, opened 2024-02-08
[MIG][17.0] base_name_search_improved

PieterJanCassiman:


base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,264 @@
...
+    smart_search = fields.Char(
+        compute="_compute_smart_search",
+        search="_search_smart_search",
+    )
+
+    def _compute_smart_search(self):
+        self.smart_search = False
+
+    @api.model
+    def _search_smart_search(self, operator, value):
```
thomaspaulb, line 151: While it looks like [smart search is included in tests](https://github.com/ACCOMODATA-BVBA/server-tools/blob/f5992d2f224cc0f87b54986c09e6670b3b781825/base_name_search_improved/tests/test_name_search.py#L15), this function does not seem to receive test coverage. Any idea why?

StefanRijnhart:
Please apply the technical method described in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#how-to to preserve the full commit history.

Bart-dh:
Hi! I was wondering - is there still ongoing work on this? Does it work functionally?

PR #2834, opened 2024-02-08
[16.0][FIX] audit_log: fix regression (Add/Remove User Group)

BT-anieto:
When adding/removing a group in a user (and tracking a user or a partner), odoo sends a 'reified' val 'in_group_{group_id}'. We must convert this value into the real groups_id one.

This bug is now appearing after the merge of #2814 because of https://github.com/OCA/server-tools/blob/7d79324f01920496ad34af33b819597a63b9b55b/auditlog/models/rule.py#L395

Steps to reproduce:
1. Set a rule for re.users models
2. Go to a user and select/unselect one of the groups he belongs to (only visible in debug mode)
3. Save

image

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -387,6 +387,9 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
+            # invalidate_recordset method must be called with existing fields
+            if self._name == "res.users":
+                vals = self._remove_reified_groups(vals)
             # Prevent the cache of modified fields from being poisoned by
             # x2many items inaccessible to the current user.
             self.invalidate_recordset(vals.keys())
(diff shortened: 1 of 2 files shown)


BT-anieto:
@StefanRijnhart Could you please take a look?

StefanRijnhart (commented):
Thanks! This should be safe as it's what core Odoo does first thing in res.users' own `write` method: https://github.com/odoo/odoo/blob/47efddb/odoo/addons/base/models/res_users.py#L1611-L1612

To make sure this does not drop out in a future version, would it be possible to add a test?

BT-anieto:
@StefanRijnhart I'll do it asap (Next week)

BT-anieto:
@StefanRijnhart Done! :)

StefanRijnhart (approved):
Those are beautiful tests, thank you!

BT-anieto:
@StefanRijnhart Do you know if it needs another review in order to be merged?

StefanRijnhart:
@BT-anieto Yes, it does.

flotho (approved):
3 lines of code
80lines of test !
thanks for this
LGTM

PR #2836, opened 2024-02-08
[15.0][FIX] Missing fontawesome icons in some browsers

ozoromo:
`fa-style-family` was not being set correctly on all browsers causing FontAwesome icons to be broken in some (most notably chrome).

### DIFF
--- a/base_fontawesome/static/src/css/fontawesome.css
+++ b/base_fontawesome/static/src/css/fontawesome.css
@@ -13,6 +13,11 @@
     font-display: block;
 }
 
+:root,
+:host {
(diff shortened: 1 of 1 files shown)


anmarmo1 (commented):
I don't know if it's for the same reason, but the main menu icon still needs fixing
image

The other icons already look correct

image

ozoromo:
Think it might actually be displaying correctly:
https://fontawesome.com/icons/th-large?f=classic&s=solid

zamberjo (approved):

anmarmo1 (approved):
Ok

PR #2596, opened 2023-03-27
[14.0][PORT] session_db 2593 from 13.0

sbidoul:
Port #2593 from 13.0 to 14.0.

PR #2840, opened 2024-02-13
[15.0] [MIG] session db: Migration to 15.0

IT-Ideas:


thomaspaulb:
@IT-Ideas 

- The version update should be part of the migration commit, so, needs to be separate from the pre-commit fixes
- You probably also want to include the fix [here](https://github.com/OCA/server-tools/pull/2596)
- And maybe you can leave your review there as well in order to get it merged quicker

IT-Ideas (commented):
No real code change with 14.0 version but:
* disabling two pylint `except-pass`
* modifying tests as assertRaises uses a savepoint in 15.0, which was failing because of the patching made upfront


Tested functionally

IT-Ideas:
@thomaspaulb , here you are!

thomaspaulb (approved):

thomaspaulb:
@IT-Ideas thanks! If you can, your review still on https://github.com/OCA/server-tools/pull/2596 would be greatly appreciated

IT-Ideas:
@lmignon

lmignon (approved):

PR #1657, opened 2019-08-28
[12.0] IMP dbfilter_from_header allowing to use hostname (%h) or the first subdomain (%d)…

eLBati:
… through which the system is being accessed, same feature  as odoo itself.

This can be useful when you have a virtualhost with "server_name *.domain.com"

See https://www.odoo.com/documentation/12.0/setup/deploy.html#dbfilter

PR #2841, opened 2024-02-13
[FIX] dbfilter_from_header issue when header does not exist

thomaspaulb:


GSLabIt (approved):
it should work

cvinh:
Hello
Do you think this commit has to be forwarded to last version ?

thomaspaulb:
I don't think so, unless someone would want to forward-port https://github.com/OCA/server-tools/pull/1657/files also, because the introduction of that code made it necessary for the result to always be a string. In current 18.0 it's fine if it's `None`.

PR #2843, opened 2024-02-14
[12.0][UPD] module_auto_update: enable test that got disabled

thomaspaulb:
Found this out while fixing tests for 11.0, noticing the hash had changed but for 12.0 the hash was still the same. Turns out test didn't run at all.


thomaspaulb:
Seeing the test running successfully in the logs now, so fast-tracking for merge.

/ocabot merge patch

PR #2846, opened 2024-02-14
[15.0][UPD] module_auto_update: enable test that got disabled

thomaspaulb:
Same as https://github.com/OCA/server-tools/pull/2843

### DIFF
--- a/module_auto_update/tests/test_addon_hash.py
+++ b/module_auto_update/tests/test_addon_hash.py
@@ -2,13 +2,14 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 import os
-import unittest
+
+from odoo.tests import TransactionCase
 
 from .. import addon_hash
 from ..models.module import DEFAULT_EXCLUDE_PATTERNS
 
 
-class TestAddonHash(unittest.TestCase):
+class TestAddonHash(TransactionCase):
     def setUp(self):
         super(TestAddonHash, self).setUp()
         self.sample_dir = os.path.join(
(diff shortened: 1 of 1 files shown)


thomaspaulb:
@sbidoul This runboat fail persists after retriggering, and I don't see any recent commits on 15.0 that would have broken this since a few days ago when I merged a 15.0 PR.

Arguably [pyopenssl](https://stackoverflow.com/questions/73830524/attributeerror-module-lib-has-no-attribute-x509-v-flag-cb-issuer-check) needs an update, but how could this have broken? And if it is, would this happen in other repos as well?

sbidoul:
@thomaspaulb that's always this same story of compatibility of the latest `cryptography` with the old `pyopenssl` version that is installed by default.

The oca-ci image preinstall odoo's [requirements.txt](https://github.com/odoo/odoo/blob/15.0/requirements.txt).
This one has `cryptography==2.6.1` and `pyopenssl==19.0.0`, both pretty old.

Now, very likely, some external dependency of this repo require a more recent version of cryptography (for instance there is `acme` which requires `cryptography>=3.2.1`). So we see in the log that pip installs (correctly) cryptography 42. And that version is incompatible with pyopenssl 19.

The best solution I can think of now is to find the addon here that pulls `acme` and add, say, `pyopenssl>19` to its external dependencies, with a comment explaining it is to force pulling a more recent pyopenssl than the default one from Odoo for compatibility with cryptography that is an indirect dependency (yeah, it's complicated). This should force pip to find a correct solution.

thomaspaulb:
Thanks, @sbidoul. I see in 14.0 they have solved by pinning cryptography to `<23.2.0`, and here in 15.0 someone else pinned `acme<2.0.0`, and I guess your trick would work as well. It seems to build now.

PR #2848, opened 2024-02-15
[16.0][MIG] base_report_auto_create_qweb: Migration to 16.0

unaiberis:


pedrobaeza:
Thanks for the contribution.

unaiberis:
Thanks for the explanation.

I will change the source branch to make the commit history more visible and explainatory.

Issue #2851, opened 2024-02-20
[16.0] database_cleanup - Purge columns purges totp_secret, resulting in server crash


MrGigSolutions:
## Module
database_cleanup

## Describe the bug
With Odoo 16.0, running the "Purge columns" database cleanup, a field called `totp_secret` shows up for purging when the `auth_totp` module is installed. This field seems to be defined as a computed non-stored field (https://github.com/odoo/odoo/blob/93aa48c972889f941b97a60738d33f571204bb2c/addons/auth_totp/models/res_users.py#L24), but also has a query referring to it. (https://github.com/odoo/odoo/blob/93aa48c972889f941b97a60738d33f571204bb2c/addons/auth_totp/models/res_users.py#L160)

## To Reproduce
**Affected versions**: 16.0

1. Ensure `auth_totp` is installed
2. Install `database_cleanup`
3. In Settings, open "Purge columns"
4. Purge the `totp_secret` column
5. Refresh the page, or logout and try to log in

**Expected behavior**
Expect `totp_secret` to not show up.

**Additional context**
The `totp_secret` field would have to be defined as `stored=True`, but it is not defined as such. Instead, it's added in `res.users` init method: https://github.com/odoo/odoo/blob/93aa48c972889f941b97a60738d33f571204bb2c/addons/auth_totp/models/res_users.py#L32

A solution could be to add the "totp_secret" field to the blacklisted fields of res_users.

MrGigSolutions:
Resolved by https://github.com/OCA/server-tools/pull/2852.

PR #2852, opened 2024-02-20
[FIX][16.0] database_cleanup - Prevent deletion of columns added in the init method of the model.

MrGigSolutions:
See issue https://github.com/OCA/server-tools/issues/2851.

The same issue was found in the `res_partners` table, with the `signup_token` field.


StefanRijnhart (approved):
Nice, thanks! Next time you don't have to create a separate issue if you are going to propose a fix. You can put all info on the pull request instead.

legalsylvain:
/ocabot merge patch

Thanks !

PR #2853, opened 2024-02-20
[16.0][ADD] base_force_record_noupdate

imlopes:


base_force_record_noupdate/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,62 @@
...
+<odoo>
+        <record id="res_config_settings_view_form" model="ir.ui.view">
+            <field name="name">res.config.settings.view.form.inherit.base.setup</field>
+            <field name="model">res.config.settings</field>
+            <field name="priority" eval="0" />
+            <field name="inherit_id" ref="base_setup.res_config_settings_view_form" />
+            <field name="arch" type="xml">
+                <xpath expr="//div[@name='integration']" position="after">
+
+                    <h2 groups="base.group_no_one">Models to Protect</h2>
```
gurneyalex:
```suggestion
                    <h2 groups="base.group_no_one">Models on which to force "noupdate=True"</h2>
```

base_force_record_noupdate/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                            <div class="o_setting_left_pane">
+                                <field name="set_default_models" force_save="1" />
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label
+                                string="Set default protected models"
+                                for="set_default_models"
+                            />
+                                <div class="text-muted">
+                                    Select this option to set the default models to protect.
```
gurneyalex:
```suggestion
                                    Selecting this option is a shortcut to add the following models to the list of models for which noupdate=True will be forced:
```

base_force_record_noupdate/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                                    The default models are:
+                                    <ul>
+                                        <li>ir.model.access</li>
+                                        <li>ir.rules</li>
+                                        <li>mail.template</li>
+                                        <li>res.groups</li>
+                                        <li>res.lang</li>
+                                        <li>res.users.role</li>
+                                    </ul>
+                                    To update this list, you can add or remove models inheriting from res.config.settings.
```
gurneyalex:
```suggestion
```

base_force_record_noupdate/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                                <field name="set_default_models" force_save="1" />
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label
+                                string="Set default protected models"
+                                for="set_default_models"
+                            />
+                                <div class="text-muted">
+                                    Select this option to set the default models to protect.
+                                    The default models are:
```
gurneyalex:
```suggestion
```

base_force_record_noupdate/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                        <field
+                            name="models_force_noupdate"
+                            nolabel="1"
+                            widget="many2many_tags"
+                            options="{'no_create': True}"
+                            placeholder="Models to force noupdate"
+                        />
+                        <div class="text-muted">
+                            Select the models to force noupdate = True.
+                            When updated, the data with models selected will not be updated.
```
gurneyalex:
```suggestion
                            Select the models on which  to force noupdate = True for the records defined in modules. 
                            When the modules owning the records of these models are updated, manual changes made in the database will not be overwritten.
```

gurneyalex (approved):

base_force_record_noupdate/models/res_config_settings.py
```diff
@@ -0,0 +1,128 @@
...
+                {"noupdate": noupdate}
+            )
+
+    def _get_preset_noupdate_models(self):
+        """
+        Return a list of models that should have noupdate=True by default.
+        """
+        models_force_noupdate = [
+            "res.lang",
+            "ir.rules",
```
gurneyalex, line 106:
```suggestion
            "ir.rule",
```

base_force_record_noupdate/models/res_config_settings.py
```diff
@@ -0,0 +1,128 @@
...
+        models_force_noupdate = (
+            self.env["ir.config_parameter"].sudo().get_param("models_force_noupdate")
+        )
+        # convert the comma-separated list in a list of model IDs
+        if models_force_noupdate:
+
+            models_force_noupdate = self.env["ir.model"].browse(
+                list(map(int, models_force_noupdate.split(",")))
+            )
+            res.update(models_force_noupdate=models_force_noupdate.ids)
```
gurneyalex, line 84: This does not seem to work: on my test database I have 	95,206,81,8,36 in the system parameter, yet when I open the settings, I only see record rules

image

I think you need to send a Command.link: ``
```suggestion
            res.update(models_force_noupdate=[Command.link(id) for id in models_force_noupdate.ids])
```
(warning: untested)

gurneyalex (changes requested):

base_force_record_noupdate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2017 ACSONE SA/NV (<http://acsone.eu>)
```
simahawk, line 1: can you fix the (c) everywhere? :)

base_force_record_noupdate/models/res_config_settings.py
```diff
@@ -0,0 +1,128 @@
...
+
+    _inherit = "res.config.settings"
+
+    set_default_models = fields.Boolean(
+        string="Set default models",
+        help="Set the default models that should have noupdate=True.",
+    )
+    models_force_noupdate = fields.Many2many(
+        "ir.model",
+        string="Models to force noupdate=True",
```
simahawk, line 18: a many to many that becomes a param? Why?
This should probably be a m2m on the company?
All this complexity can be trashed. What is the pro of keeping it?

base_force_record_noupdate/models/ir_model.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        # check models that we need to protect from update
+        noupdate_models = (
+            self.env["ir.config_parameter"].sudo().get_param("models_force_noupdate")
+        )
+        if noupdate_models:
+            models_force_noupdate = list(map(int, noupdate_models.split(",")))
```
simahawk, line 20: I don't understand why we need this complexity. See below.

simahawk (changes requested):

simahawk:
Alex explained me that this has to work cross company. If that's the case I would recommend to add a flag to the ir.model and trash the complexity completely as you'll be able to simply do `self.env["ir.model"]._get(model).force_nupdate` or smth similar.
You can also do a single select for all involved models and cache it.

simahawk:
If you want to keep the param I would simply:
1. trash the settings override (it's a very technical conf, no need for editing settings)
2. hook to set_param to reflect the change on existing records

gurneyalex:
superseeded by https://github.com/OCA/server-tools/pull/2946

PR #2859, opened 2024-02-23
[12.0][DEL] profiler: tour test

thomaspaulb:
I once thought tours were a cool way of testing modules, but I've run into more problems then it has given merit, and also, there's another regular test which covers the use case of the module - coverage won't go down.

Removing the tour test, so I don't have to spend time maintaining it and fixing pipelines.


thomaspaulb:
Merging, since I'm maintainer and this failed test was blocking https://github.com/OCA/server-tools/pull/2857

/ocabot merge patch

PR #2865, opened 2024-02-27
[FIX] Add missing web dependency

Rad0van:
Without explicit web dependency all tests fail like this:

```
File "/home/odoo/src/odoo/odoo/models.py", line 6905, in onchange
    raise NotImplementedError("onchange() is implemented in module 'web'")
NotImplementedError: onchange() is implemented in module 'web'
```
See: https://github.com/odoo/odoo/blob/a9ebd07b8028aa019e142d6809ec3aeba518a965/odoo/models.py#L6904-L6905
And: https://github.com/odoo/odoo/blob/a9ebd07b8028aa019e142d6809ec3aeba518a965/addons/web/models/models.py#L889-L1095

### DIFF
--- a/onchange_helper/__manifest__.py
+++ b/onchange_helper/__manifest__.py
@@ -9,6 +9,6 @@
     "website": "https://github.com/OCA/server-tools",
     "license": "AGPL-3",
     "category": "Generic Modules",
-    "depends": ["base"],
+    "depends": ["base", "web"],
     "installable": True,
 }
--- a/onchange_helper/__manifest__.py
+++ b/onchange_helper/__manifest__.py
@@ -9,6 +9,6 @@
     "website": "https://github.com/OCA/server-tools",
     "license": "AGPL-3",
     "category": "Generic Modules",
-    "depends": ["base", "web"],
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Hi. Thanks for this patch. 
Makes sense. 

However, I don't understand the odoo design here. (why the function is defined in web and not in base module...)

onchange_helper/__manifest__.py
```diff
@@ -9,6 +9,6 @@
     "website": "https://github.com/OCA/server-tools",
     "license": "AGPL-3",
     "category": "Generic Modules",
-    "depends": ["base"],
+    "depends": ["base", "web"],
```
legalsylvain:
```suggestion
    "depends": ["web"],
```

is enough.

legalsylvain:
nice to have : Could you add a simple comment in DEVELOP.rst about that dependency ?

StefanRijnhart (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2873, opened 2024-03-06
[16.0][IMP] database_cleanup: set appropriate labels in each purge view

StefanRijnhart:
instead of 'Purge this model' everywhere.


StefanRijnhart:
I'm trying again for 18.0 https://github.com/OCA/server-tools/pull/3273

pedrobaeza:
I think this is still valid, isn't it?

PR #2874, opened 2024-03-07
Import Auditlog form version 16.0 into 17.0.

heccjj:
1. Import Auditlog code from version 16 into 17.
2. Change version number in __manifest__.py.
3. Because Odoo 17 does no support  attrs and states any more, so need to change those code in views/auditlog_view.xml. Can use https://github.com/pierrelocus/odoo-attrs-replace.git to finished it, but it will change other normal code too, have to change those normal code manully.
4. This is the first time to do a pull request, wish you can accept it.


legalsylvain (commented):
Hi @heccjj. 
First, thanks for your contribution to the OCA.
could you  follow the guidelines to migrate module to preserve history : 
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#technical-method-to-migrate-a-module-from-160-to-170-branch
Thanks !

PR #2875, opened 2024-03-07
[16.0][IMP] attachment_synchronize: add additional error hook

kevinkhao:
this can be useful in case for example if you want to link an attachment.queue that raised an error, to a record. It dodges the restriction with the rollback. As a caveat, it cannot be added to the manual run button and preserve the raise and the hook in a simple way. However, chances are it will already have been run and linked to the record

### DIFF
--- a/attachment_queue/models/attachment_queue.py
+++ b/attachment_queue/models/attachment_queue.py
@@ -127,6 +127,10 @@ class AttachmentQueue(models.Model):
                     self.env.ref(
                         "attachment_queue.attachment_failure_notification"
                     ).send_mail(self.id)
+                self._additional_error_hook(e)
+
+    def _additional_error_hook(self, e):
+        return True
 
     def run(self):
         """


florian-dacosta (approved):
Can you tell more about your use case ?
Anyway, why not LGTM

kevinkhao:
> Can you tell more about your use case ? Anyway, why not LGTM

In conjunction with https://github.com/OCA/server-tools/pull/2875#pullrequestreview-2020069245

I add a hook to link the attachment.queue to whatever record it was trying to process.

For example, in my case 1 attachment queue fills the "done" quantities of 1 picking. If there is an error, for example I try to set the quantities on a product that shouldn't be in that picking, I still link the attachment queue to the picking. Something like 

`

    def _additional_error_hook(self, e):

        if type(e) is WmsImportException:

            e.record.track_model_import(self)

        return super()._additional_error_hook(e)
`

PR #2877, opened 2024-03-07
[17.0][MIG] auditlog: Migration to 17.0

heccjj:
Main Change:
Because Odoo 17 does no support attrs and states any more, so need to change those code in views/auditlog_view.xml. Can use https://github.com/pierrelocus/odoo-attrs-replace.git to finish it, but it will change other normal code too, have to change those normal code manully.


heccjj:
Fixed the error when subscribing a auditlog rule.

PR #2878, opened 2024-03-08
17.0 mig auditlog

heccjj:
1.import auditlog from 16.0
2.move "states" codes in models/rule.py, to control "readonly" by views.
3.change file views/auditlog_view.xml, fix "states" and add "readonly" for some fields in rule form.


nilshamerlinck:
Hello @heccjj you might want to have a look at existing ongoing work in https://github.com/OCA/server-tools/pull/2755 ;-)

PR #2886, opened 2024-03-15
[14.0][ADD] base_view_full_arch

SilvioC2C:
New module to display views' full architecture:
1. for a primary view, field `full_arch` displays the full architecture with all the inheriting views' modifications included
2. for an extension view, field `full_arch` displays the full architecture of the view's inherited primary view

Useful for analysis, debug and development, especially when you're dealing with views or reports that are inherited and customized by dozens of other different views, making it a mess to understand how the combined architecture is supposed to be.


gurneyalex (approved):

.prettierignore
```diff
@@ -0,0 +1,3 @@
+# Ignore .xml files used for tests
```
simahawk, line 1: this is not needed. Put those files into `samples` folder.
See https://github.com/OCA/oca-addons-repo-template/pull/240

PR #2900, opened 2024-04-04
[14.0][ADD] impersonate_login module

Kev-Roche:
This module allows to login as another user,  its purpose is mainly (only?) for testing and the support team.

No tests wrote in this module yet, any hints / help will be appreciated. 
After adding tests, I will migrate the module in v16.


impersonate_login/models/mail_message.py
```diff
@@ -0,0 +1,52 @@
...
+            if request and request.session.impersonate_from_uid:
+                rec.impersonated_author_id = (
+                    self.env["res.users"]
+                    .browse(request.session.impersonate_from_uid)
+                    .partner_id.id
+                )
+            else:
+                rec.impersonated_author_id = False
+
+    @api.depends("author_id", "impersonated_author_id", "body")
```
legalsylvain, line 36: Something is weird in that dependency. Body computation dépends on body...

impersonate_login/models/impersonate_log.py
```diff
@@ -0,0 +1,26 @@
...
+class ImpersonateLog(models.Model):
+    _name = "impersonate.log"
+    _description = "Impersonate Logs"
+
+    user_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="User",
+    )
+    impersonated_user_id = fields.Many2one(
+        comodel_name="res.partner",
```
legalsylvain, line 18:
```suggestion
        comodel_name="res.users",
```

Or réname field...

impersonate_login/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows to login as another user.
+In the chatter, the user who is logged as another user is displayed.
+The mails and messages are sent from the orignal user.
+A table diplays the impersonated logins in technical.
+The user can return to his own user by clicking on the button "Return to my user".
+This module is very useful for the support team.
```
legalsylvain, line 6: Could you add a mention to alternative implémentation :
Auth_admin_passkey ?

legalsylvain (commented):
Thanks for this great module ! Not tested. Quick code review only. Some remarks inline.

Not sure but i think that this module should go in server-auth.

impersonate_login/models/model.py
```diff
@@ -0,0 +1,23 @@
...
+        if request and request.session.impersonate_from_uid:
+            user = self.env["res.users"].browse(request.session.impersonate_from_uid)
+            return super(BaseModel, self.with_user(user))._create(data_list)
+        return super()._create(data_list)
+
+    def write(self, vals):
+        if request and request.session.impersonate_from_uid:
+            user = self.env["res.users"].browse(request.session.impersonate_from_uid)
+            return super(BaseModel, self.with_user(user)).write(vals)
+        return super().write(vals)
```
rvalyi, line 23: May be I am mistaken but **I have big doubts** about this implementation: what if my module B overrides create or write (super common) and does something, (potentially even like writing or creating some other record) on the behalf of the current self.env.user before calling super().create or super().write?

 **Because impersonate_login will not depend on my module B, your _create and write overrides can be arbitrarily called before or after the ones in B**... So I fear I have no guarantee your impersonate_login module will have impersonated the user yet in my overrides in B...


What about renaming the module `base_schizophrenia` or something like that?

Shouldn't this user switch be done during the login phase in res.users instead or in https://github.com/OCA/server-auth/blob/16.0/auth_admin_passkey/models/res_users.py#L57 ?

cc @sebastienbeau @legalsylvain

rvalyi (changes requested):

impersonate_login/models/model.py
rvalyi: also I wonder if the repo should not rather be https://github.com/OCA/server-auth like for auth_admin_passkey for instance...

Kev-Roche:
@legalsylvain @rvalyi thank both of you for your comments, I will take your advices into account and put it in the server-auth repo : https://github.com/OCA/server-auth/pull/637

PR #2901, opened 2024-04-08
[16][FIX] autovacuum_message_attachment : fix inheriting_model feature

florian-dacosta:
I did not  see this mistake during migration.  I am not sure this feature was ever working in the past version...
I fixed it anyway, it should be fine now !

@etobella @thomaspaulb

### DIFF
--- a/autovacuum_message_attachment/models/ir_attachment.py
+++ b/autovacuum_message_attachment/models/ir_attachment.py
@@ -21,8 +21,8 @@ class IrAttachment(models.Model):
         if rule.inheriting_model:
             inheriting_model = self.env[rule.inheriting_model]
             attachment_link = inheriting_model._inherits.get("ir.attachment")
-            att_ids = inheriting_model.search(create_date_domain).mapped(
-                attachment_link + ".ids"
+            att_ids = (
+                inheriting_model.search(create_date_domain).mapped(attachment_link).ids
             )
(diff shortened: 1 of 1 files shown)


autovacuum_message_attachment/models/ir_attachment.py
```diff
@@ -22,7 +22,7 @@ def _get_autovacuum_domain(self, rule):
             inheriting_model = self.env[rule.inheriting_model]
             attachment_link = inheriting_model._inherits.get("ir.attachment")
             att_ids = inheriting_model.search(create_date_domain).mapped(
-                attachment_link + ".ids"
+                attachment_link + ".id"
```
etobella: wouldn't be cleaner to do `inheriting_model.search(create_date_domain).mapped("attachment_link").ids`

thomaspaulb: Agree here, it's non-blocking but not a lot of work I guess?

thomaspaulb (approved):

autovacuum_message_attachment/models/ir_attachment.py
florian-dacosta: I guess, I first did that but I changed to do the minimal change.
I'll update this way, no problem.

etobella (approved):
/ocabot merge patch

PR #2902, opened 2024-04-10
[FIX] auto_backup: TypeError: makedirs() got an unexpected keyword argument 'exist_ok'

eLBati:


valentincastravete:
Hi @eLBati 

That error is related to having a lower Python version that the required by the Odoo 12.0 version:
<a href="https://www.odoo.com/documentation/12.0/administration/install.html#python" target="_blank">Required Python version for Odoo 12.0</a>


As you can see, in the Python2, there is no "exist_ok" in the "make_dirs" function:
<a href="https://docs.python.org/2/library/os.html#os.makedirs" target="_blank">Python2</a>
But in the Python3 version, you can see that it's there:
<a href="https://docs.python.org/3.6/library/os.html#os.makedirs" target="_blank">Python3</a>

_"New in version 3.2: The exist_ok parameter."_

I suggest you manually remove that parameter in the server-tools repo in your project or update the Python version to the required one by the Odoo version 12.0.

I've put the parameter there to not to get an Error when trying to create a directory if it already exists, it just ignores that error:
_If exist_ok is False (the default), an <a href="https://docs.python.org/3.6/library/exceptions.html#OSError" target="_blank">OSError</a> is raised if the target directory already exists._

Regards

eLBati:
Hi @valentincastravete 
my odoo server is running on `Python 3.6.9`

The change does not involve `os.makedirs` but [pysftp.Connection.makedirs()](https://pysftp.readthedocs.io/en/release_0.2.9/cookbook.html#pysftp-connection-makedirs) which does not accept `exist_ok` AFAICS

valentincastravete:
Hi @eLBati 

You are right, I checked on the wrong package, sorry.
Okay, then I suggest you also remove the ", exist_ok=True" from the line 163 on the [test_db_backup.py](https://github.com/eLBati/server-tools/blob/1acd143f4b3419a3433fb0ea9802a8d51b4e24ba/auto_backup/tests/test_db_backup.py#L163) so it can pass the tests.

Could you merge this FIX, please? @thomaspaulb 
Thanks!

eLBati:
@valentincastravete thanks, done

thomaspaulb (approved):

PR #2904, opened 2024-04-11
[15.0][FIX] auto_backup: Backup is work when setting list_db=False in odoo.conf

newtratip:


rrebollo (commented):
CR. While this approach may function, it does not seem like a valid solution, as performing the backup would require violating a serious security measure.

PR #2907, opened 2024-04-16
[16.0][IMP] auditlog: prevent creating logs without log lines

AungKoKoLin1997:
This PR enhances the change detection logic in the write_full and write_fast methods to ensure logs are only generated when significant data changes occur. Previously, compute functions could trigger logging even without actual data modifications, leading to logs that lacked corresponding log lines. This update prevents such redundant logging, optimizing performance and maintaining cleaner log records.

@qrtl QT4483

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -400,7 +400,13 @@ class AuditlogRule(models.Model):
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
-            if self.env.user in users_to_exclude:
+            # Check if any value has actually changed
+            changes = any(
+                old_values[rec_id] != new_values[rec_id]
+                for rec_id in self.ids
+                if rec_id in old_values and rec_id in new_values
+            )
+            if not changes or self.env.user in users_to_exclude:
                 return result
             rule_model.sudo().create_logs(
                 self.env.uid,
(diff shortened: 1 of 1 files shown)


kanda999 (approved):
Functional review: LGTM

AungKoKoLin1997:
@oca-server-tools-maintainers
Can you please take a look this PR?

auditlog/models/rule.py
```diff
@@ -400,7 +400,13 @@ def write_full(self, vals, **kwargs):
...
                 .read(fields_list)
             }
-            if self.env.user in users_to_exclude:
+            # Check if any value has actually changed
+            changes = any(
+                old_values[rec_id] != new_values[rec_id]
+                for rec_id in self.ids
+                if rec_id in old_values and rec_id in new_values
+            )
+            if not changes or self.env.user in users_to_exclude:
```
yostashiro, line 409:
```suggestion
            if new_values == old_values or self.env.user in users_to_exclude:
```

auditlog/models/rule.py
```diff
@@ -423,8 +429,14 @@ def write_fast(self, vals, **kwargs):
...
             new_values = {id_: vals2 for id_ in self.ids}
+            # Check if any value has actually changed
+            changes = any(
+                old_values[rec_id] != new_values[rec_id]
+                for rec_id in self.ids
+                if rec_id in old_values and rec_id in new_values
+            )
             result = write_fast.origin(self, vals, **kwargs)
-            if self.env.user in users_to_exclude:
+            if not changes or self.env.user in users_to_exclude:
```
yostashiro, line 439: Please check if this doesn't skip log creation when a value is changed to false from something else.

yostashiro (changes requested):

AungKoKoLin1997:
This improvement is not valid. I misunderstood compute function behavior and confused. Closing this PR.

Issue #2343, opened 2022-05-17
[15.0] Auditlog: error when creating variants or updating lst_price


BT-jcamacho:
## Module
auditlog

## Describe the bug
When a rule exists for model `product.product`, creating of product.products from the Product Variants List View or changing the lst_price (Sales price) raises an Error.

## To Reproduce
**Affected versions**: v15 at least

1. Create a rule that logs write for model Product (`product.product`)
2. Go to to the Product Variants list view (i.e, from Sales app -> Products -> Products Variants)
3. Create a new variant

OR

1. Open the form view of a variant
2. Click on edit
3. Change the sales price (`lst_price`)

**Expected behavior**
No error should be thrown

**Additional context**
Tested on v15 runboat

I think problems comes from a core onchange in product.product:
https://github.com/odoo/odoo/blob/15.0/addons/product/models/product.py#L277-L285

```python
@api.onchange('lst_price')
def _set_product_lst_price(self):
    for product in self:
        if self._context.get('uom'):
            value = self.env['uom.uom'].browse(self._context['uom'])._compute_price(product.lst_price, product.uom_id)
        else:
            value = product.lst_price
        value -= product.price_extra
        product.write({'list_price': value})
```
where they perform a `write()` that auditlog tries to track. As its an onchange (which is triggered when creating or updating the `lst_price`) the record is a "NewId record", which causes problems when reading the fields to track.

angelvilaplana:
Hi @BT-jcamacho, while testing to migrate from Odoo 14 to 15, I had the same problem. From what I've seen, it's because of what you comment in the issue. It's seen that in version 15 they added this decorator, causing this function to be executed by default when a new product record is created and causing the error to appear.

I've created a PR with a solution to this problem. I've tested it in my environment, and when I create a new product record or edit the sales price, that problem no longer appears.

PR #2912, opened 2024-04-18
[15.0][FIX] auditlog rule: Control the write function in the models when executed by an onchange function

angelvilaplana:
Changes to fix the following issue: https://github.com/OCA/server-tools/issues/2343

The goal of these changes is to fix the issue that when you activate the audit rule for the products, and then you try to create or modify the price, an access error appears.

From what I've been able to see, this is because in version 15 added the onchange decorator in the _set_product_lst_price function. As within this function a write is being executed for the same product. This causes it to go through the logic of the AuditlogRule with an object that has the id as "NewId".  Causing the error to appear when checking the security rules.

I've modified the AuditlogRule class and modify the _make_write function, where inside there, I check if the record id is an object that instance of NewId. If instance of NewId, What it does is ignore the original logic to prevent it from executing the read function. It'll only run the write_full.origin to do the base logic. If this is not the case, and it's a normal id, it'll execute the usual logic.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -379,16 +379,19 @@ class AuditlogRule(models.Model):
             self = self.with_context(auditlog_disabled=True)
             rule_model = self.env["auditlog.rule"]
             fields_list = rule_model.get_auditlog_fields(self)
+            records_write = self.filtered(lambda r: not isinstance(r.id, models.NewId))
+            if not records_write:
+                return write_full.origin(self, vals, **kwargs)
             old_values = {
                 d["id"]: d
-                for d in self.sudo()
+                for d in records_write.sudo()
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
             result = write_full.origin(self, vals, **kwargs)
             new_values = {
                 d["id"]: d
-                for d in self.sudo()
+                for d in records_write.sudo()
                 .with_context(prefetch_fields=False)
                 .read(fields_list)
             }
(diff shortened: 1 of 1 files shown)


cuongnmtm:
The same issue happens on 16.0 and 17.0, blocking tracking write changes on the product.

cuongnmtm (approved):

DucTruongKomit (approved):
Did few tests and code LGTM.

JordiMForgeFlow (approved):

MateuGForgeFlow (approved):

LoisRForgeFlow (approved):
:+1: 

/ocabot merge patch

LoisRForgeFlow:
@angelvilaplana you will need to rebase this PR to be able to merge.

LoisRForgeFlow:
@angelvilaplana looks like branch 15.0 is not very active in the last 2 commits of v15 (single commits in the last 6 months) the pre-commit is already failing so I guess we need a copier update to update the pre-commit configuration and all the dotfiles. Could you open a separated PR with that and ping me?

MateuGForgeFlow:
Hi @angelvilaplana,
I see you have 2 commits in v15 and just one commit in v16.
What do you think about unifying criteria? Maybe squashing commits here?

I am doing the forward port for v17.

Thank you!

angelvilaplana:
Hi @MateuGForgeFlow,

Thank you very much for the message and the suggestion!
I've squashed the commits to unify the criteria as you recommended.

PR #2913, opened 2024-04-19
[IMP] base_exception: Add a hook method / Don't build void exceptions

rousseldenis:
In order to get exceptions without writing recordset, add a hook method.

Improve exceptions dictionaries building using a defaultdict of a browse record.

### DIFF
--- a/base_exception/models/base_exception_method.py
+++ b/base_exception/models/base_exception_method.py
@@ -4,8 +4,8 @@
 # Copyright 2020 Hibou Corp.
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-
 import logging
+from collections import defaultdict
 
 from odoo import _, api, models
 from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


base_exception/models/base_exception_method.py
```diff
@@ -35,35 +35,42 @@ def _rule_domain(self):
...
             to_remove = records_with_rule_in_exceptions - records_with_exception
             to_add = records_with_exception - records_with_rule_in_exceptions
-            # we expect to always work on the same model type
-            if rule_info.id not in rules_to_remove:
-                rules_to_remove[rule_info.id] = main_records.browse()
-            rules_to_remove[rule_info.id] |= to_remove
-            if rule_info.id not in rules_to_add:
-                rules_to_add[rule_info.id] = main_records.browse()
-            rules_to_add[rule_info.id] |= to_add
+            if to_remove:
```
hparfr, line 61: does this `if` improve something ?

hparfr (approved):

hparfr:
by the way, what's your use case for this hook ?

rousseldenis:
> by the way, what's your use case for this hook ?

Because we changed on a model the field exception_ids to a computed stored one for performance reasons and we don't want to do writes() in a compute.

IMHO, some improvements should be done in order to use compute/stored fields instead doing detect_exceptions in write/create. The problem is the trigger.

base_exception/models/base_exception_method.py
rousseldenis: Yes, the dict will be lighter as currently containing all the rules.

lmignon (approved):

PR #2914, opened 2024-04-19
[17.0][MIG] iap_alternative_provider: Migration to 17.0

luka-guru:
Minimal changes for iap_alternative_provider to work on Odoo 17


iap_alternative_provider/models/iap_account.py
```diff
@@ -9,7 +9,6 @@ class IapAccount(models.Model):
     _inherit = "iap.account"
     _name = "iap.account"
```
etobella: This line is not needed

iap_alternative_provider/views/iap_account_view.xml
```diff
@@ -7,14 +7,17 @@
             <xpath expr="//group[@name='account']" position="before">
                 <group string="Info" name="info">
                     <field name="provider" />
-                    <field name="name" />
-                    <field name="company_ids" position="move" />
+                    <field name="name" position="move" />
```
etobella: the move has no sense IMO...

iap_alternative_provider/models/iap_account.py
vvrossem: I don't know the module, but maybe there is a reason it combines both `_inherit` and `_name`, cf. https://www.odoo.com/documentation/17.0/developer/reference/backend/orm.html#classical-inheritance

Also, this line is present since 12.0

iap_alternative_provider/views/iap_account_view.xml
vvrossem: agreed, if `name`  is removed on the model side, it can be removed too on the xml side.

iap_alternative_provider/models/iap_account.py
luka-guru: please clarify if you want it kept or removed so we can move forward with the merge.
@vvrossem @etobella

iap_alternative_provider/views/iap_account_view.xml
luka-guru: If you believe name should be removed from form view I'm happy to do it, just clarify please so we can move on with the merge. 
@etobella @vvrossem

iap_alternative_provider/models/iap_account.py
etobella: It is not needed, the model already exists and you need to define the name if you want to create a new module or have multiple inheritances. It is not the case

iap_alternative_provider/views/iap_account_view.xml
etobella: The move was removing the field from the view. Just remove the line.

etobella:
I would recommend to contact original authors of the module and people that did interesting contributions

@sebastienbeau @FernandoRomera @petrus-v @hendrixcosta Can you provide your point of view, please? :pray:

FrancoMaxime (approved):

gurneyalex (approved):

PR #2916, opened 2024-04-22
[17.0][MIG] tracking_manager

CRogos:


tracking_manager/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "website": "https://github.com/OCA/server-tools",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["Kev-Roche", "sebastienbeau"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "base",
+        "mail",
+    ],
```
GuillemCForgeFlow:
```suggestion
    "depends": ["mail"],
```
no need to set base as a dependency, `mail` already has it included

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,206 @@
...
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    active_custom_tracking = fields.Boolean()
+    tracked_field_count = fields.Integer(compute="_compute_tracked_field_count")
+    automatic_custom_tracking = fields.Boolean(
+        compute="_compute_automatic_custom_tracking",
+        readonly=False,
+        store=True,
+        help=("If tick new field will be automatically tracked if the domain match"),
```
GuillemCForgeFlow:
```suggestion
        help=("If marked, the fields matching the matched by the domain below will be automatically tracked for this model."),
```
perhaps this slightly improves the help description?

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,206 @@
...
+        for rec in self:
+            rec.tracked_field_count = len(rec.field_id.filtered("custom_tracking"))
+
+    def write(self, vals):
+        if "active_custom_tracking" in vals:
+            self.env.registry.clear_cache()
+        return super().write(vals)
+
+
+class IrModelFields(models.Model):
```
GuillemCForgeFlow: could we move this class to another file to have it better organized perhaps?

tracking_manager/views/ir_model.xml
```diff
@@ -0,0 +1,66 @@
...
+                            name="automatic_custom_tracking"
+                            string="Automatic configuration"
+                            invisible="not active_custom_tracking"
+                        />
+                        <field
+                            name="automatic_custom_tracking_domain"
+                            string="Domain"
+                            invisible="not automatic_custom_tracking"
+                            required="automatic_custom_tracking"
+                        />
```
GuillemCForgeFlow:
```suggestion
                        <field
                            name="automatic_custom_tracking_domain"
                            string="Domain"
                            invisible="not automatic_custom_tracking"
                            required="automatic_custom_tracking"
                            widget="domain"
                            options="{'model': 'ir.model.fields', 'in_dialog': True, 'foldable': True}"
                        />
```

tracking_manager/models/ir_model.py
```diff
@@ -0,0 +1,206 @@
...
+            ],
+        }
+
+    @api.depends("automatic_custom_tracking")
+    def _compute_automatic_custom_tracking_domain(self):
+        rules = self._default_automatic_custom_tracking_domain_rules()
+        for record in self:
+            record.automatic_custom_tracking_domain = str(
+                rules.get(record.model) or rules.get("default_automatic_rule")
+            )
```
GuillemCForgeFlow:
```suggestion
        for record in self:
            automatic_custom_tracking_domain = rules.get(record.model) or rules.get("default_automatic_rule", [])
            automatic_custom_tracking_domain.append(('model', '=', record.model))
            record.automatic_custom_tracking_domain = str(automatic_custom_tracking_domain)
```
would it make sense to do like so considering also the widget='domain' changes I have suggested below?
I believe doing it like that we can improve the UI when adding the domain from the Model view. WDYT?

GuillemCForgeFlow (commented):
some suggestions, for the rest I have functionally tested and works fine 👍🏿

tracking_manager/views/ir_model.xml
CRogos: I totally agree, but the record result is a little misleading, because the related table is not filtered. I have the same issue in one of our own customizations. Do you have an idea how we could improve this?
image

CRogos: ok solved it myself.
Thank you for all your input. I always apricate review input.

GuillemCForgeFlow: I think it looks better like this then you can actually visualize the fields you will be tracking if you update before actually updating, it is more user friendly IMHO.

GuillemCForgeFlow (approved):
LGTM! functional and code review 👍🏿 
thank you for accepting the suggestions btw 😄 
you can already mark all discussions as resolved!

MohamedOsman7 (approved):
code & functional review LGTM

CRogos:
@pedrobaeza could you merge this?

pedrobaeza:
Please rebase and fix the new linters:

```
tracking_manager/models/ir_model.py:21:89: E501 Line too long (122 > 88)
tracking_manager/models/ir_model_fields.py:11: [R8180(consider-merging-classes-inherited), IrModelFields] Consider merging classes inherited to "ir.model.fields" from tracking_manager/models/ir_model.py:165:4.
```

CRogos:
@pedrobaeza rebased and fixed.

PR #2903, opened 2024-04-11
[17.0][UPD] copier version 1.20

MohamedOsman7:

PR #2920, opened 2024-04-23
[17.0][REF] dotfiles: copier update

dreispt:
Supersedes #2903


gurneyalex:
I just merged another PR on 17 doing this

PR #2921, opened 2024-04-24
[17.0][MIG] fetchmail notify error to sender: Migration to 17.0

Mikheil-21:
Tested based on instructions from the module.


LoisRForgeFlow:
This module is already merged in 17.0

PR #2923, opened 2024-04-24
[16] attachment synchronize : Improve remote file import management

florian-dacosta:
Current problems are the following : 
1) When import multiple files from a remote server, if an error occurs after the first file, we risk loosing data, because, the first file(s) imported could have been deleted/renamed on remote server.
We need to commit after each imported file to ensure we won't loose any data.

2) If one import task fails, the whole cron stops. It means if we import files from multiple remote server and the first remote  server is unavailable, all import from all remote will be blocked.
We add a warning in the log if there is a failure for a remote but don't block the rest of the process.

@kevinkhao @bealdav @GSLabIt @sebastienbeau


GSLabIt (approved):

kevinkhao (approved):
Code review
LGTM
Thanks

bealdav (approved):
Thanks a lot

PR #2931, opened 2024-05-01
[17.0] dotfiles update needs manual intervention

sbidoul:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,

Handled by @Tecnativa

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.19
+_commit: v1.21.1
 _src_path: gh:oca/oca-addons-repo-template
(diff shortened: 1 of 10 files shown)


legalsylvain:
hi @sbidoul. 

I just tried to fix it. But when I run locally ``pre-commit run -a``, it doesn't fix the errors. Is there a regression on V17 ? 
In v16 and before, running pre-commit was fixing incorrect python and xml files.

Thanks !

sbidoul:
@legalsylvain I don't think there has ever been auto fixes for line too long when it requires to split strings?

pedrobaeza:
FYI I'm handling this.

PR #2905, opened 2024-04-12
[15.0][IMP] database_cleanup: purge fields (forward-port)

astirpe:
This is the forward-port of https://github.com/OCA/server-tools/pull/2491
with the addition of some tests.

PR #2936, opened 2024-05-07
[16.0][IMP] database_cleanup: purge fields (forward-port)

astirpe:
This is the forward-port of #2905 to V16


astirpe:
CC @CasVissers-360ERP

CasVissers-360ERP (approved):
Functional review.

PR #2942, opened 2024-05-15
[14.0] [FIX] base_view_inheritance_extension: parse old value

renda-dev:
There could be some cases where a domain is a string with line breaks ("`\n`") and lots of spaces, that could lead to exceptions.

With this PR we also parse the old domain to clean it up.

### DIFF
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -181,13 +181,16 @@ class IrUiView(models.Model):
                 else:
                     new_value = str(expression.AND([old_domain, new_domain]))
                 new_value = self.str2var_domain_text(new_value)
+                old_value = "".join(old_value.splitlines())
             else:
-                new_value = attribute_node.text
+                # We must ensure that the domain definition has not line breaks because
+                # in update mode the domain cause an invalid syntax error
+                new_value = attribute_node.text.strip()
             if condition:
                 new_value = "{condition} and {new_value} or {old_value}".format(
                     condition=condition,
                     new_value=new_value,
-                    old_value=old_value,
+                    old_value=old_value or [],
                 )
             node.attrib[attribute_name] = new_value
         return source


renda-dev:
Hi @pedrobaeza, could you please give this PR a look whenever you have time? 

This is blocking https://github.com/OCA/sale-workflow/pull/3135, thanks a lot!

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -181,13 +181,16 @@ def inheritance_handler_attributes_domain_add(self, source, specs):
                 else:
                     new_value = str(expression.AND([old_domain, new_domain]))
                 new_value = self.str2var_domain_text(new_value)
+                old_value = self.str2var_domain_text(str(old_domain))
```
pedrobaeza: You are doing this 2 times. Check line 173

renda-dev: Not really: i first need to parse the string that I've got in domain (line 173), to get rid of all extra formatting (spaces, tabs, line breaks), then I parse it again in a clean string (line 184) and use that to create the proper `new_value` (line 193).

Let me know if I've been clear enough, thanks!

renda-dev: To give you more context (https://github.com/OCA/sale-workflow/actions/runs/9097954749/job/25007038324?pr=3135#step:8:736)
immagine

Here after the "and" there is the `new_value` which is the parsed old value with the parsed new value, which gives no problem at all.

But after the "or" there is an empty string which is just the `old_value` (not parsed)

With this PR we also parse the `old_value` to make sure that is properly formatted.

pedrobaeza: Yes, but don't call two times the same method. Optimize the code.

renda-dev: Hi @pedrobaeza, could you check now please?

Only removing line breaks seems to work fine

pedrobaeza (approved):
/ocabot merge patch

PR #2928, opened 2024-04-30
[16.0][ADD] server_action_logging

imlopes:

PR #2945, opened 2024-05-20
[16.0][ADD] server_action_logging

SilvioC2C:
Supersedes #2928


server_action_logging/models/ir_actions_server.py
```diff
@@ -0,0 +1,41 @@
...
+            if force_enable:
+                odoo_sql_db_logger.setLevel(logging.DEBUG)
+            subject = f"Action {act.display_name} ({act})"
+            _logger.info(f"{subject} started")
+            start = fields.Datetime.now()
+            res = act.run()
+            delta = (fields.Datetime.now() - start).total_seconds()
+            _logger.info(f"{subject} completed in {delta:.3f} seconds")
+            if force_enable:
+                odoo_sql_db_logger.setLevel(odoo_sql_db_logger_level)
```
gurneyalex:
```suggestion
        for act in actions:
            force_enable_sql_debug = not odoo_sql_db_logger_has_debug and act.enable_sql_debug
            if force_enable_sql_debug:
                odoo_sql_db_logger.setLevel(logging.DEBUG)
            subject = f"Action {act.display_name} ({act})"
            _logger.info(f"{subject} started")
            start = fields.Datetime.now()
            res = act.run()
            delta = (fields.Datetime.now() - start).total_seconds()
            _logger.info(f"{subject} completed in {delta:.3f} seconds")
            if force_enable_sql_debug:
                odoo_sql_db_logger.setLevel(odoo_sql_db_logger_level)
```

gurneyalex (approved):
looks good. I suggested a small naming improvement.

vvrossem (approved):

server_action_logging/models/ir_actions_server.py
```diff
@@ -0,0 +1,41 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class IrActionsServer(models.Model):
+    _inherit = "ir.actions.server"
+
+    enable_sql_debug = fields.Boolean(
+        string="Enable SQL Debug",
+        help="Enable SQL Debug for this action",
+        default=True,
```
grindtildeath: Do you really want to set `True` as default value? That means it will get enabled for all the existing actions...

If that is the case, I'd suggest to adapt the readme because "if action's field ``Enable SQL Debug`` is set, queries are logged too during the action's execution" makes me think it has to be set manually by the user.
IMO it would be more telling to write something like "SQL queries will be logged during the action's execution, what could be deactivated through the ``Enable SQL Debug`` field".

grindtildeath (approved):
Very nice feature (hopefully they will integrate it in the core of Odoo). Pre approving anyway.

server_action_logging/models/ir_actions_server.py
SilvioC2C: Mh, I inherited the code from #2928 and really didn't think about the field declaration. It would make sense to have the field set to `False` instead, to avoid too much log pollution. I'll update it asap.

PR #2946, opened 2024-05-20
[16.0][ADD] base_force_record_noupdate

SilvioC2C:
ref: BSRD-770


base_force_record_noupdate/models/ir_model_data.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import api, models
+
+
+class IrModelData(models.Model):
+    _inherit = "ir.model.data"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        noupdate = self.env["ir.model"].sudo()._get_noupdate_models().mapped("model")
```
gurneyalex:
```suggestion
        noupdate = set(self.env["ir.model"].sudo()._get_noupdate_models().mapped("model"))
```

gurneyalex (approved):
small improvement suggested.

base_force_record_noupdate/models/ir_model_data.py
SilvioC2C: @gurneyalex for curiosity, why the `set()`?

gurneyalex: to speed up the check on line 14.

gurneyalex: I just checked with timeit, on a list with 6 items, the time spent on the list version is quite small in the worst case (model name not in the list), so I guess we can keep it that way. The version with a set is faster (2/3  of the time with the list, but this is peanuts on small lists)

SilvioC2C: Didn't know sets where faster than lists on containment checks. It's not a big change so I can introduce it.

base_force_record_noupdate/models/ir_model_data.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class IrModelData(models.Model):
+    _inherit = "ir.model.data"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        noupdate_models = self.env["ir.model"].sudo()._get_noupdate_models()
+        if noupdate_models:
+            noupdate_model_names = set(noupdate_models.mapped("model"))
```
simahawk, line 14: why `set`? You shouldn't have any duplicate

base_force_record_noupdate/tests/test_base_force_record_noupdate.py
```diff
@@ -0,0 +1,75 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+from odoo.tests import TransactionCase
+
+MODULE = "base_force_record_noupdate"
+
+
+class TestBaseForceRecordNoupdate(TransactionCase):
+    def test_00_test_model_create_noupdate_propagation(self):
+        # NB: we're only testing models created through UI, the feature is not
```
simahawk, line 11: I don't get this comment. Create is overridden, what's the probl?

base_force_record_noupdate/models/ir_model.py
```diff
@@ -0,0 +1,53 @@
...
+    def _get_noupdate_model_ids(self):
+        return self.search([("force_noupdate", "=", True)]).ids
+
+    @api.model
+    def _propagate_noupdate_to_model_data(self):
+        noupdate = self._get_noupdate_models().mapped("model")
+        imd = self.env["ir.model.data"].sudo()
+        model_data = imd.search([("model", "in", noupdate), ("noupdate", "=", False)])
+        if model_data:
+            model_data.write({"noupdate": True})
```
simahawk, line 45: maybe is good to log something?

simahawk (approved):
LG

base_force_record_noupdate/models/ir_model_data.py
SilvioC2C: `set.__contains__()` is faster than `list.__contains__()` => better performances when checking `vals.get("model") in noupdate_model_names`

base_force_record_noupdate/tests/test_base_force_record_noupdate.py
SilvioC2C: What I mean is that if we create a new model via Python code, we still need to use the UI to set this field to `True`.
The `create` override only applies for models created through the UI: if someone plans to create an all-new model from the Py code, it'll trigger an `ir.model.create()`, but the newly created record will have `force_noupdate = False` (at least until the record is updated via a `write()`).
Thus, the only way to test the `ir.model.create()` override is to mimic the manual creation of a new model.

PR #2965, opened 2024-06-11
[16.0][FIX] auditlog: many2many field fill full colspan

miikanissi:
Here is the view before:
auditlog-before

And here after:
auditlog-after


yostashiro (approved):
LGTM. 👍

flotho (approved):
easy code review
Thansk LGTM

rvalyi (approved):

PR #2972, opened 2024-06-20
[16.0][MIG] sequence_python

ntsirintanis:


sequence_python/models/ir_sequence.py
```diff
@@ -0,0 +1,87 @@
...
+        wrap_uuid = wrap_module(uuid, uuid_elements)
+        wrap_string = wrap_module(string, string.__all__)
+        return {
+            "number": number_next[0] if isinstance(number_next, tuple) else number_next,
+            "number_padded": "%%0%sd" % self.padding % number_next,
+            "sequence": self,
+            "random": wrap_random,
+            "uuid": wrap_uuid,
+            "string": wrap_string,
+            "get_barcode_check_digit": get_barcode_check_digit,
```
ntsirintanis: @thomaspaulb probably we can a less extensive string for key here. Any suggestions?

thomaspaulb: Why though, it's descriptive

sequence_python/models/ir_sequence.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2020 Sunflower IT (<https://sunflowerweb.nl>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+import random
+import string
+import uuid
+from inspect import getmembers, isclass, isfunction
+
+from odoo import fields, models
+from odoo.tools.barcode import get_barcode_check_digit
```
NL66278: This barcode thing seems to be something very specific. Maybe better put in a separate module....

NL66278 (approved):
:+1: One comment that might be taken up, but LGTM.

sequence_python/models/ir_sequence.py
thomaspaulb: Added it to https://github.com/OCA/delivery-carrier/pull/902 instead, where I'm using it.

PR #2977, opened 2024-07-01
[16.0][FIX] base_name_search_improved: don't patch abstract models

StefanRijnhart:
When applying monkey patches, we want to skip abstract models because patching those may mess up the inheritance. An example of this is ir.model which is assigned the studio mixin using inherit = ['studio.mixin', 'ir.model']. If the mixin itself is patched, and the method is overridden once again (in, say, enterprise 15's documents_spreadsheet), the super() method called in that override is the patched version of studio.mixin rather than the override of ir.model in the base module, which is now skipped entirely.


pedrobaeza (approved):

CasVissers-360ERP (approved):
Functional review

rov-adhoc (approved):

PR #2978, opened 2024-07-02
fix: depends on whool instead of setuptools-odoo for V17.0

vrenaville:


sbidoul:
There should actually be no `setup.py` in the 17.0 branches since there is a `pyproject.toml` in each addon. So closing this.

PR #2981, opened 2024-07-03
[16.0][ADD] base_search_multi_value

kevinkhao:
essentially a generic product_search_multi_value


bealdav (commented):
Congratulations for this module and your reactivity here. 
Seems to be a good idea to make it more generic that only search for product.

But don't you think, you might improve this module avoiding to break a such feature ?
May be you may define a convention with a char (by default | for example) in a general config/settings.
If this char is inside the searched string, then search change its behavior to search all split strings.
If this char is not in this search string, then native search is returned

What do you think ?

If you don't have time to improve with the suggested way, you may, add suggestion to roadmap  part

bealdav:
I suppose that server-ux is a better place ?

PR #2982, opened 2024-07-05
[17.0][MIG] fetchmail_notify_error_to_sender, fetchmail_notify_error_to_sender_test: Migration to 17.0

BT-cjimeno:
fetchmail_notify_error_to_sender: Migration from 15.0
fetchmail_notify_error_to_sender_test: Migration from 14.0

Fixes the bug found [here](https://github.com/OCA/server-tools/pull/2574#issuecomment-1547503682)

Modified the tests to include all the raise conditions.


BT-dmoreno (approved):

BT-anieto (approved):
LGTM

fetchmail_notify_error_to_sender/models/fetchmail.py
PieterPaulussen, line 1: Not entirely sure if this is required, but should the name of the file not match the model? `fetchmail_server.py`?

fetchmail_notify_error_to_sender/views/fetchmail_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" ?>
+<odoo>
+    <record model="ir.ui.view" id="view_email_server_form">
```
PieterPaulussen:
```suggestion
    <record id="view_email_server_form" model="ir.ui.view">
```

fetchmail_notify_error_to_sender_test/i18n/ca.po
PieterPaulussen, line 1: Can we remove empty translation files?

fetchmail_notify_error_to_sender/static/description/icon.png
PieterPaulussen, line 1: Could this icon be replaced with the OCA module icon?

PieterPaulussen (approved):
I only have stylistic suggestions (And not even sure they are required). Otherwise seems technically sound. 👍

BT-cjimeno:
Hi @PieterPaulussen, sorry for the late reply and thx for the review, I agree with all your points.

I am not sure if they can all go in the same commit as the migration, or if I need to use a different commit and syntax for the new commit message, since they are improvements not directly related with the migration but with the module.

hbrunn:
I think it's perfectly fine to address @PieterPaulussen's comments in a new commit in the same PR

fetchmail_notify_error_to_sender_test/i18n/ca.po
BT-cjimeno: I got rid of some of them that only had 1 or no terms translated, and also removed a comment at the very bottom of some of them that does not seem needed IMO.

I left some of them that only had a few terms translated just in case...

PieterPaulussen (approved):

PieterPaulussen:
Thanks for the updates @BT-cjimeno

hbrunn (approved):

PR #3000, opened 2024-07-23
[16.0][IMP] fetchmail_attach_from_folder: ir.actions.server

zamberjo:
Added the possibility to indicate a server action to be executed for the matching record.

### DIFF
--- a/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
+++ b/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
@@ -91,6 +91,12 @@ class FetchmailServerFolder(models.Model):
         help="The state messages fetched from this folder should be assigned in Odoo",
     )
     active = fields.Boolean(default=True)
+    action_id = fields.Many2one(
+        comodel_name="ir.actions.server",
+        name="Server action",
+        help="Optional custom server action to trigger for each incoming "
+        "mail, on the record that was created or updated by this mail",
+    )
 
     def button_confirm_folder(self):
         self.write({"state": "draft"})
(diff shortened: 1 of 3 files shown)


fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -227,11 +233,30 @@ def apply_matching(self, connection, msgid):
...
         matched = True if thread_id else False
+        if matched:
+            self.run_server_action(thread_id)
         self.update_msg(connection, msgid, matched=matched)
         if self.archive_path:
             self._archive_msg(connection, msgid)
         return thread_id  # Can be None if no match found.
 
+    def run_server_action(self, matched_object_ids):
+        self.ensure_one()
```
NL66278: This is not needed, the method is perfectly capable of handling a recordset, even if only called with a single record now.

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -227,11 +233,30 @@ def apply_matching(self, connection, msgid):
...
                     self.attach_mail(match, message_dict)
         matched = True if thread_id else False
+        if matched:
+            self.run_server_action(thread_id)
         self.update_msg(connection, msgid, matched=matched)
         if self.archive_path:
             self._archive_msg(connection, msgid)
         return thread_id  # Can be None if no match found.
 
+    def run_server_action(self, matched_object_ids):
```
NL66278, line 243: Please add a test for this method, to increase instead of decrease the test coverage.

NL66278 (changes requested):
Nice addition! A few changes would make it perfect.

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
zamberjo: done

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
zamberjo: tests added

NL66278 (approved):
:+1: very nice. LGTM. Usefull addition.

anmarmo1 (approved):
The functional review is correct

zamberjo:
Hi @OCA/tools-maintainers can someone merge this?

PR #3001, opened 2024-07-23
[16.0][IMP] fetchmail_attach_from_folder: Read e-mails are excluded

zamberjo:
If you have a folder with a large number of emails, `fetch_mail` is executed for all of them.

We add the flag to read only the unread ones. In this way a lot of load is released on the server.

### DIFF
--- a/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
+++ b/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
@@ -97,6 +97,10 @@ class FetchmailServerFolder(models.Model):
         help="Optional custom server action to trigger for each incoming "
         "mail, on the record that was created or updated by this mail",
     )
+    fetch_unseen_only = fields.Boolean(
+        help="By default all undeleted emails are searched. Checking this "
+        "field adds the unread condition.",
+    )
 
     def button_confirm_folder(self):
         self.write({"state": "draft"})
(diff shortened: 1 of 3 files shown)


NL66278:
The customer for which we wrote this originally, and it will be the same with others, had a large problem with users reading the mailbox also in their mail client. Then the mails would no longer be fetched by Odoo. So reading all undeleted mails is intentional. However the performance problem is real. Therefore the possibility to put mails handled by fetchmail into another folder, so they will be read only once. Only reading mails that are unread might be another approach to the problem, if it can be guaranteed that the user thus not read the mails in any other way. However this should then be configurable, and for compatibility reasons default off. I suggest adding a boolean field `fetch_unseen_only`, default False, and add the flag if it has been set.

NL66278 (changes requested):
Reading only unseen messages should be optional.

zamberjo:
> I suggest adding a boolean field `fetch_unseen_only`, default False, and add the flag if it has been set.

I get on with it

zamberjo:
Added `fetch_unseen_only` field.

NL66278 (approved):
:+1: Also a nice thing to refactor the retrieval of search criteria to its own method.

zamberjo:
Hi @OCA/tools-maintainers can someone review/merge this?

zamberjo:
I just saw that there was a conflict. Rebase done

dreispt (approved):

PR #3004, opened 2024-07-25
[MIG][16.0] base_import_odoo

len-foss:


thomaspaulb:
@len-foss Do you feel like finishing your version here or shall I continue in my PR? I fixed pre-commit and added a feature, it's now green.

len-foss:
Oh right pre-commit was foobar at the time of the PR.
I'm not specifically working on it right now so it's better if you do :-)

Closed in favor of https://github.com/OCA/server-tools/pull/3136

PR #3011, opened 2024-08-13
[16][IMP] attachment_synchronize : add hook when pushing file to remote

florian-dacosta:
It may be usefull for instance to be able to pass custom parameters to the fsspec open method

@bealdav @kevinkhao @bguillot 
Easy one if you may review it

### DIFF
--- a/attachment_synchronize/models/attachment_queue.py
+++ b/attachment_synchronize/models/attachment_queue.py
@@ -21,6 +21,12 @@ class AttachmentQueue(models.Model):
         selection_add=[("export", "Export File (External location)")]
     )
 
+    def _write_file_to_remote(self, fs, full_path):
+        self.ensure_one()
+        data = base64.b64decode(self.datas)
+        with fs.open(full_path, "wb") as f:
+            f.write(data)
+
     def _run(self):
         res = super()._run()
         if self.file_type == "export":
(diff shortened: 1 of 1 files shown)


bealdav (approved):

kevinkhao (approved):
LGTM

attachment_synchronize/models/attachment_queue.py
```diff
@@ -21,6 +21,11 @@ class AttachmentQueue(models.Model):
         selection_add=[("export", "Export File (External location)")]
     )
 
+    def _write_file_to_remote(self, fs, full_path, data):
+        self.ensure_one()
```
SirAionTech: `self` is not used in this method, so I don't understand why the constraint `self.ensure_one()`.
I'd suggest to either remove it and use `api.model`
```suggestion
    @api.model
    def _write_file_to_remote(self, fs, full_path, data):
```
or actually use it to retrieve `data`
```suggestion
    def _write_file_to_remote(self, fs, full_path):
        self.ensure_one()
        data = base64.b64decode(self.datas)
```
what do you think?

SirAionTech (approved):
Thanks for the PR!
Only code review: I don't see any issue with this change, just asking little clarification (see comment below).

attachment_synchronize/models/attachment_queue.py
florian-dacosta: I think the `self.ensure_one` is important, because the aims of this hook is to be able to override it so do custom stuff that would depend on the attachment queue object.

But, I agree that the data could be in this method, make sense IMO
I just done it

SirAionTech (changes requested):
Sorry I didn't notice it before, but could you please adapt the commit message to the guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message?
In particular
> please check if the commit message is cut with ellipsis.

florian-dacosta:
Sure, I've changed the commit message, thanks.

SirAionTech (approved):

PR #3016, opened 2024-08-20
[14.0][FIX] utf-8 cached backup error auto_backup

sergiocorato:
Error in backup process:

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xeb in position 14: invalid continuation byte`

How to replicate:
1. create one backup with filestore
2. create another backup without filestore
3. wait cron execution

Second error: the second backup is a copy of the first one, regardless of the type. So the second backup, of the type "dump", is a copy of the "zip" file without the correct extension.

### DIFF
--- a/auto_backup/models/db_backup.py
+++ b/auto_backup/models/db_backup.py
@@ -159,9 +159,9 @@ class DbBackup(models.Model):
                     pass
 
                 with open(os.path.join(rec.folder, filename), "wb") as destiny:
-                    # Copy the cached backup
-                    if backup:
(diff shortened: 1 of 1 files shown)


giarve:
For some reason in version 16.0 the same thing happened to me with 2 filestore (both being the same type) backups but saved in *different locations*!

rrebollo:
@sergiocorato in order to confirm your thesis you should add a test showing the root cause of the error. Your patch has sense but I'm no sure about your diagnosis, look at the comment from @giarve.

PR #3017, opened 2024-08-22
[16.0][FIX] fetch...folder: use message uid's not sequence

NL66278:
When archiving messages, the messages get a new sequence through renumbering. To solve the problems this causes use the message uid's that are garanteed to remain constant for a message within a folder.

### DIFF
--- a/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
+++ b/fetchmail_attach_from_folder/models/fetchmail_server_folder.py
@@ -180,22 +180,26 @@ class FetchmailServerFolder(models.Model):
     def retrieve_imap_folder(self, connection):
         """Retrieve all mails for one IMAP folder."""
         self.ensure_one()
-        msgids = self.get_msgids(connection, self.get_criteria())
-        for msgid in msgids[0].split():
+        message_uids = self.get_message_uids(connection, self.get_criteria())
+        for message_uid in message_uids[0].split():
             # We will accept exceptions for single messages
             try:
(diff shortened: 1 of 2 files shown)


NL66278:
@zamberjo Can you have a look at this?

zamberjo (approved):
Technically it seems correct, I have not been able to test it functionally.

anmarmo1 (approved):
I am not clear about the initial problem that this change solves, but, the initial functionality is still correct, so for me the change is correct.

ikus060:
As highlighted in PR #3222, the reverse processing of emails aims to address a similar issue to the one you're working on here.

But I fail to see the exact modification you are doing in this PR to retreive UID instead of sequences. It's probably lost in all the variable renames.

NL66278:
@ikus060 I introduce the variable name message_uid instead of msgid, because it is a really different thing. msgid is basically the position of the message in the imap folder, and gets updated all the time because of insertions and deletions. message_uid however is constant (as long as the message remains in the same folder). However to work with the uid's the methods to work with messages are replaced by a single method on the connection, uid, with the original methodname as the first parameter: 
connection.uid("search", charset, criteria)
connection.uid("fetch", message_uid, "(RFC822)")
...
Working with the message_uid's is intrinsically more reliable...

ikus060:
ok, I see `connection.uid("search", charset, criteria)` vs `connection.search(None, criteria)`

I've missed that while looking at the changes. Thanks for pointing the changes.

PR #3018, opened 2024-08-23
[16.0][IMP] tracking_manager: Test compatibility

victoralmau:
Previously the `bank_ids` field was used but [account](https://github.com/odoo/odoo/blob/16.0/addons/account/models/res_partner_bank.py#L176) adds messages in partner when a related bank account is created/modified/deleted, causing tests to fail. Related to https://github.com/OCA/server-tools/pull/2789#issuecomment-1870227588

The `user_ids` field of `res.partner` is now used to avoid the side effect in tests and the fake model is removed.

@Tecnativa TT50571


pedrobaeza:
Isn't possible to do the test without creating an extra model? In other place/field?

victoralmau:
I think it would be possible to use the `user_ids` field but that would make a very large diff.

pedrobaeza:
Well, right now you have a large diff changing `bank_ids` by `custom_bank_ids`...

victoralmau:
Changes done: The `user_ids` field of `res.partner` is now used (although everything is **only** in fake models for tests).

pedrobaeza:
I only see `category_ids` used in:

https://github.com/OCA/server-tools/blob/312daa160949916efdb3b828bcf8b97ca82a859c/tracking_manager/tests/test_tracking_manager.py#L192

I think you can use another field and remove the FakeModelLoader.

pedrobaeza (approved):

Kev-Roche (approved):

PR #3019, opened 2024-08-26
[16.0][IMP] tracking_manager: Hide Custom Tracking if the model does not inherit from mail.thread

victoralmau:
Hide Custom Tracking if the model does not inherit from `mail.thread`

Please @pedrobaeza can you review it?

@Tecnativa TT50571

### DIFF
--- a/tracking_manager/views/ir_model.xml
+++ b/tracking_manager/views/ir_model.xml
@@ -9,8 +9,13 @@
         <field name="inherit_id" ref="base.view_model_form" />
         <field name="arch" type="xml">
             <xpath expr="//sheet/group[1]" position="after">
-
-                <group string="Custom Tracking">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

CRogos:
This change would prevent tracking sale.order.line values in sale.order:
image

Nevertheless I agree we also had some trouble creating product.category after activating tracking on this model.

pedrobaeza:
Uhm, how can we detect that situations...?

pedrobaeza:
Or we can simply silently avoid the error:

```
AttributeError: 'model' object has no attribute 'message_post_with_view'
```

CRogos:
How about this suggestion? https://github.com/OCA/server-tools/pull/3021

pedrobaeza:
Let's go with #3021

PR #2991, opened 2024-07-17
[17.0] mig-excel_import_export

Josep-s73:
Standard migration 17.0

PR #3020, opened 2024-08-26
[17.0][IMP] tracking_manager: Test compatibility

victoralmau:
FWP from 16.0: https://github.com/OCA/server-tools/pull/3018

Previously the `bank_ids` field was used but [account](https://github.com/odoo/odoo/blob/16.0/addons/account/models/res_partner_bank.py#L176) adds messages in partner when a related bank account is created/modified/deleted, causing tests to fail. Related to https://github.com/OCA/server-tools/pull/2789#issuecomment-1870227588

The `user_ids` field of `res.partner` is now used to avoid the side effect in tests and the fake model is removed.

@Tecnativa TT50571


pedrobaeza (approved):

pedrobaeza:
Merging for fixing CI like the one in #2991 

/ocabot merge patch

PR #3021, opened 2024-08-27
[16.0][FIX] tracking_manager: prevent error on models without mail.thread 

CRogos:
Alternative of of https://github.com/OCA/server-tools/pull/3019

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -85,6 +85,9 @@ class Base(models.AbstractModel):
 
     def _tm_post_message(self, data):
         for model_name, model_data in data.items():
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
- Don't add empty lines inside a method as a good guideline.
- Put the `getattr` on the first loop where `model_name` is extracted, and do `continue` if not having the method.

pedrobaeza (approved):

CRogos:
@victoralmau could you do the second review?

victoralmau (approved):

pedrobaeza:
/ocabot merge patch

To be fw-ported to 17.0

PR #3025, opened 2024-09-04
[16.0][FIX] tracking_manager: Avoid sending tracking o2m emails if you are subscribed to notes

victoralmau:
Avoid sending tracking o2m emails if you are subscribed to notes

Example use case:
- Enable custom tracking in `project.project`
- Activate the `task_ids` field as custom tracking
- Create a project and assign it to a user (different from ours) and make it a note follower
- Create a task in the project
- The project message of the task tracking task will not be sent to the note followers

**Before**
antes

**After**
despues

@Tecnativa TT50676

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -97,12 +97,13 @@ class Base(models.AbstractModel):
                     }
                     for field_name, messages in messages_by_field.items()
                 ]
-                # use sudo as user may not have access to mail.message
-                record.sudo().message_post_with_view(
+                # We do not use message_post_with_view() because emails would be sent
+                rendered_template = self.env["ir.qweb"]._render(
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Is this the way the regular tracking messages are post? Let's imitate that way.

victoralmau:
The way to do it in “_standard_” is https://github.com/odoo/odoo/blob/16.0/addons/mail/models/mail_thread.py#L627 but we can not use the same because this module does not use `tracking_value_ids` to define the o2m changes, for this it set a specific body according to the changes done; for that reason, the change has been done using the `_message_log()` method and specifying the corresponding body.

pedrobaeza (approved):
Yeah, the important thing is that `_message_log` is the used method.

Kev-Roche (approved):

PR #3026, opened 2024-09-06
[16] attachment synchronize Use cat_file instead of opening and reading

florian-dacosta:
It was like this before but it was changed because not supported by sshfs, but it is now : https://github.com/fsspec/sshfs/pull/45

@bealdav @kevinkhao


bealdav (approved):

flotho (approved):
LGTM
Thanks for the work done

PR #3029, opened 2024-09-09
Draft: [16.0][MIG] #10965 auditlog_security

KKamaa:


auditlog_security/README.rst
```diff
@@ -0,0 +1,98 @@
...
+    :target: https://translation.odoo-community.org/projects/server-tools-16-0/server-tools-16-0-auditlog_security
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/server-tools&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module allows extends auditlog, allowing specific log lines to be viewed only
+by users belonging to specific views, while all other lines are allowed only to
```
hbrunn, line 32:
```suggestion
by users belonging to specific groups, while all other lines are allowed only to 
```

hbrunn:
@KKamaa @gfcapalbo before doing an in-depth review, an architecture question: Why do you do complex stuff with creating and managing ir.rules on the fly, instead of:

- create a computed field `rule_id` on `auditlog.log` and `auditlog.log.line`, where you implement the search function to yield the `auditlog.rule` record for the log's model (by searching for an active rule with that model, there can only be one)
- create a field `allowed_group_ids` on `auditlog.rule`
- add ir.rules with domain `[('rule_id.allowed_group_ids', 'in', user.groups_id.ids)]` for `auditlog.log` and `auditlog.log.line`

This should make the module *much* simpler, and I don't see any reason why this shouldn't work or impact performance.

ntsirintanis:
> * [('rule_id.allowed_group_ids', 'in', user.groups_id.ids)]



> @KKamaa @gfcapalbo before doing an in-depth review, an architecture question: Why do you do complex stuff with creating and managing ir.rules on the fly, instead of:
> 
>     * create a computed field `rule_id` on `auditlog.log` and `auditlog.log.line`, where you implement the search function to yield the `auditlog.rule` record for the log's model (by searching for an active rule with that model, there can only be one)
> 
>     * create a field `allowed_group_ids` on `auditlog.rule`
> 
>     * add ir.rules with domain `[('rule_id.allowed_group_ids', 'in', user.groups_id.ids)]` for `auditlog.log` and `auditlog.log.line`
> 
> 
> This should make the module _much_ simpler, and I don't see any reason why this shouldn't work or impact performance.

Rules need to be dynamic, because of this:

https://github.com/OCA/server-tools/pull/3029/files#diff-d34dd89cc1fa327c78874da4c91035c900201a48c00e87e5c55c30a05a9173a7R92

which is an overkill in my opinion, and also this code part is untested. If we can ignore this condition, then what you suggested is the way to go

hbrunn:
even if you want to keep the fields-per-group functionality, you don't need to write specific rules. In that case keep the auditlog.line.access.rule model, and create a computed field allowed_group_ids on auditlog.log{,.line} whose search function looks at those records.

My reading was that the field-level functionality is covered by fields_to_exclude_ids from the base addon, that's why I considered that part obsolete

auditlog_security/views/auditlog_view.xml
```diff
@@ -0,0 +1,165 @@
...
+                        </p>
+                    </div>
+                </group>
+                <group>
+                    <field name="log_selected_fields_only" />
+                    <!-- in the domain are the blacklisted fields in auditlog they
+                        would be skipped anyway, so we repeat here to avoid incongruencies
+                    -->
+                    <field name="auditlog_line_access_rule_ids">
+                        <tree>
```
ntsirintanis, line 108: So records created here are truly created, but become invisible as soon as we save the form view of the parent record. Here we are creating a one2many line for a model that holds two many2many fields.   Wondering if this is the reason

ntsirintanis: ... and this was because of a record rule. So basically no one can read these lines. Why?

ntsirintanis:
@hbrunn I went ahead and implemented your second suggestion. Let's pretend for the time being that unit tests are not significant at all - their time will come. Some further refinements are needed, though the bulk of the work is there and in tandem with what you said. Please take a look and let me know.

auditlog_security/security/ir_rule.xml
```diff
@@ -15,4 +15,22 @@
         <field name="perm_unlink" eval="False" />
     </record>
```
hbrunn, line 10: I think this rule was supposed to apply to auditlog.log.line, so my answer to https://github.com/OCA/server-tools/pull/3029#discussion_r2066672418 would be remove it, what you do below is sufficient in the new approach, and in the old approach it would be change model_id to auditlog.model_auditlog_line

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,67 @@
...
+        index=True,
+        string="User",
+    )
+    method = fields.Char(compute="_compute_method", store=True, index=True)
+    model_id = fields.Many2one(
+        "ir.model", compute="_compute_model_id", store=True, index=True
+    )
+    res_id = fields.Integer(compute="_compute_res_id", store=True, index=True)
+    allowed_group_ids = fields.Many2many(
+        "res.groups", compute="_compute_allowed_group_ids", store=True
```
hbrunn:
```suggestion
        "res.groups", compute="_compute_allowed_group_ids", search="_search_allowed_group_ids"
```

I expect storing things here will come with quite a performance penalty, writing/creating logs and log lines should be as fast as possible. `_search_allowed_group_ids` would look like

```python
def _search_allowed_group_ids(self, operator, value):
    access_rules = self.env['auditlog.line.access.rule'].search(['|', ('group_ids', operator, value), ('group_ids', '=', False)])
    domains = []
    for access_rule in access_rules:
        domain = [('log_id.model_id', '=', access_rule.auditlog_rule_id.id)]
        if access_rule.field_ids:
            domain.append(('field_id', 'in', access_rule.field_ids.ids))
        domains.append(domain)
    return osv.expression.OR(domains)
```

and an analogue search function for `auditlog.log#allowed_group_ids` - this way you offload all the work to the sql server at view time, and have no impact at creation time.

And given that we don't actually care for the value of this field, the compute function can be `lambda self: self.update({'allowed_group_ids': False})` I think.

In the same vain, I suggest to replace all computed fields above with nonstored related fields, Odoo is smart enough by now to generate proper joins for that.

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    log_selected_fields_only = fields.Boolean(
+        default=True,
+        help="Log only the selected fields, to save space avoid large DB data.",
+    )
+
+    @api.constrains("model_id")
+    def unique_model(self):
+        if self.search_count([("model_id", "=", self.model_id.id)]) > 1:
+            raise ValidationError(_("A rule for this model already exists"))
```
hbrunn: the base module [does](https://github.com/OCA/server-tools/blob/16.0/auditlog/models/rule.py#L151) this already

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,131 @@
...
+        code = "action = env['auditlog.rule']._get_view_log_lines_action()"
+        server_action = (
+            self.env["ir.actions.server"]
+            .sudo()
+            .create(
+                {
+                    "name": "View Log Lines",
+                    "model_id": self.model_id.id,
+                    "state": "code",
+                    "code": code,
```
hbrunn, line 103: would make sense to add groups here if the access rules are restricted to groups I think

auditlog_security/security/ir_rule.xml
```diff
@@ -0,0 +1,36 @@
...
+    </record>
+
+    <record id="auditlog_log_line_rule" model="ir.rule">
+        <field name="name">Access to auditlog.log.line</field>
+        <field name="model_id" ref="auditlog.model_auditlog_log_line" />
+        <field
+            name="domain_force"
+        >[('allowed_group_ids', 'in', user.groups_id.ids)]</field>
+        <field name="perm_read" eval="False" />
+    </record>
```
hbrunn, line 21: I'd have expected to have the above rules apply to base.group_user and for all operations, and two other sets of rules that lift the above restrictions for auditlog.group_auditlog_user and base.group_erp_manager

auditlog_security/views/auditlog_view.xml
```diff
@@ -0,0 +1,165 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+     <record model="ir.ui.view" id="view_auditlog_log_tree">
+        <field name="name">auditlog.log.form</field>
+        <field name="model">auditlog.log</field>
+        <field name="inherit_id" ref="auditlog.view_auditlog_log_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//tree" position="attributes">
+                <attribute name="delete">false</attribute>
```
hbrunn: permissions should take care of that, right?

auditlog_security/security/res_groups.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_can_view_audit_logs" model="res.groups">
+        <field name="name">View Audit Logs</field>
+    </record>
+</odoo>
```
hbrunn: this group is not used anywhere and unnecessary given you'll assign groups in the access rules record I think?

auditlog_security/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_auditlog_log_line_user,auditlog_log_line_user,auditlog.model_auditlog_log_line,base.group_user,1,0,0,0
+access_auditlog_line_access_rule_admin,auditlog_line_access_rule_admin,model_auditlog_line_access_rule,base.group_erp_manager,1,1,1,1
```
hbrunn, line 3: why this line and not stick with the permissions of auditlog?

hbrunn (changes requested):

auditlog_security/models/auditlog_log_line.py
ntsirintanis: amazing, thanks for the wisdom shared here

ntsirintanis: though this _search method should return a domain for searching on res.groups

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,30 @@
...
+    method = fields.Char(related="log_id.method")
+    model_id = fields.Many2one(related="log_id.model_id", store=True)
+    res_id = fields.Integer(related="log_id.res_id")
+    allowed_group_ids = fields.Many2many(
+        "res.groups",
+        compute=lambda self: self.update({"allowed_group_ids": False}),
+        search="_search_allowed_group_ids",
+    )
+
+    def _search_allowed_group_ids(self, operator, value):
```
ntsirintanis, line 39: @hbrunn even though this method works on shell and it returns what it is supposed to return, it does not trigger at all when some user loads the auditlog.log.line tree view. Can you speculate why?

ntsirintanis: Note however, that records are filtered correctly on the tree view, for some user. Results are fine, the method is triggered, it works etc, though whenever I put a breakpoint and try to trigger it in the ui, nothing happens.

hbrunn: you've set the rule not to be applied for read operations, see https://github.com/OCA/server-tools/pull/3029#discussion_r2071136138

auditlog_security/security/ir.model.access.csv
ntsirintanis: well that's for the auditlog_line_access_rule model, introduced in this mr

hbrunn: oh right, misread that

auditlog_security/views/auditlog_view.xml
ntsirintanis: Agreed, though I suspect that that's there because no human should be allowed to delete anything from the ui

auditlog_security/security/ir_rule.xml
ntsirintanis: Sure. Now I am getting rid of anything related to auditlog.log model modifications (besides adding a many2one), so basically 3 rules are needed here:  base.group_user, auditlog.auditlog_user and base.group_erp_manager.  So group_user won't be able to do anything, auditlog_user to just read log lines, and erp_manager to play god with this

hbrunn: I think the domains for auditlog_user and group_erp_manager should both be [(1, '=', 1)], so that after installing this module those groups can do the exact same things as before installing it

ntsirintanis: @hbrunn explain here a little bit more how you envision this please

hbrunn: the way I understand the purpose of this module, it's supposed to allow admins to configure rules such that normal users (base.group_user) can access some log (view) lines.

auditlog itself grants unrestricted read access to group_auditlog_user (because it doesn't set record rules, only access rules), so if you want to keep that, but at the same time allow all backend users access according to the rules configured, you need to lift the restrictions you impose on base.group_user, because all members of group_auditlog_user, group_auditlog_manager, group_erp_manager are also members of group base.group_user.

So I'd expect a rule with the domain you have above and perm_*=True for base.group_user, and one for each of the other groups named here with domain [(1, '=', 1)] to undo this restriction.

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,126 @@
...
+            .search([("model_id.model", "=", model_name)], limit=1)
+        )
+        if rule.auditlog_line_access_rule_ids:
+            return rule.mapped("auditlog_line_access_rule_ids.field_ids.name")
+        return []
+
+    @api.model
+    @tools.ormcache("model_name")
+    def _get_log_selected_fields_only(self, model_name):
+        """Memory-cached translation of model to rule"""
```
ntsirintanis, line 37: @hbrunn I've noticed that as soon as we create an auditlog rule and subscribe, and then trigger a field write, nothing really happens, i.e., no auditlog.log(and line) records were created.

For this to work,  I had to unsubscribe, uncheck the log_selected_fields_only,  save form,  check the log_selected_fields_only,  save form,  and then subscribe again. That's the only way I could actually trigger account.log(and line) records to be created. Any thoughts on that?

ntsirintanis:
@hbrunn resume with your review please

ntsirintanis:
@thomaspaulb you might want to take a look too. The module looks radically different than before now, and things appear to be fine during my personal tests

auditlog_security/security/ir_rule.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="name">Access to auditlog.log.line</field>
+        <field name="model_id" ref="auditlog.model_auditlog_log_line" />
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="perm_read" eval="True" />
+        <field name="perm_write" eval="True" />
+        <field name="perm_create" eval="True" />
+        <field name="perm_unlink" eval="True" />
+    </record>
+
+    <record id="auditlog_log_line_rule_auditlog_user" model="ir.rule">
```
hbrunn: why restrict auditlog_user's permissions? I'd add this group below to the rule lifting the restrictions imposed on base.group_user.

The auditlog module defines this: https://github.com/OCA/server-tools/blob/16.0/auditlog/security/ir.model.access.csv - so I think after installing auditlog_security, those groups should end up with the same permissions.

auditlog_security/security/ir_rule.xml
```diff
@@ -0,0 +1,34 @@
...
+<odoo noupdate="0">
+
+    <record id="auditlog_log_line_rule_base_user" model="ir.rule">
+        <field name="name">Access to auditlog.log.line</field>
+        <field name="model_id" ref="auditlog.model_auditlog_log_line" />
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="perm_read" eval="True" />
+        <field name="perm_write" eval="True" />
+        <field name="perm_create" eval="True" />
+        <field name="perm_unlink" eval="True" />
```
hbrunn:
```suggestion
```



that's the default, I only mentioned those being true because there was perm_read=False before, which was wrong

auditlog_security/views/auditlog_view.xml
hbrunn: but that should be up to administrators fiddling with the permissions, not to some module making this choice i think

auditlog_security/views/auditlog_view.xml
hbrunn: this one is resolved, right?

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,126 @@
...
+        server_action = (
+            self.env["ir.actions.server"]
+            .sudo()
+            .create(
+                {
+                    "name": "View Log Lines",
+                    "model_id": self.model_id.id,
+                    "state": "code",
+                    "code": code,
+                    "groups_id": [(6, 0, allowed_groups.ids)],
```
hbrunn:
```suggestion
                    "groups_id": [(6, 0, allowed_groups.ids)] if not any(not x.group_ids for x in self.auditlog_line_access_rule_ids),
```

if there's an unrestricted access rule and a restricted one, all users should see the action. ir.rules will take care of what they see then

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,126 @@
...
+                }
+            )
+        )
+        self.write({"server_action_id": server_action.id})
+        return server_action
+
+    def subscribe(self):
+        for rule in self:
+            server_action = rule._create_server_action()
```
hbrunn, line 112:
```suggestion
            rule.server_action_id.unlink()
            server_action = rule._create_server_action()
```

or make it write on the action if it exists, create otherwise

auditlog_security/models/auditlog_log.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "auditlog.log"
+
+    rule_id = fields.Many2one("auditlog.rule", compute="_compute_rule_id", store=True)
+
+    @api.depends("model_id")
+    def _compute_rule_id(self):
+        for log in self:
+            log.rule_id = self.env["auditlog.rule"].search(
+                [("model_id", "=", log.model_id.id)]
+            )
```
hbrunn: do we need this whole file after all?

auditlog_security/models/auditlog_rule.py
hbrunn: do a registry reload as in https://github.com/OCA/server-tools/blob/16.0/auditlog/models/rule.py#L249 below in write instead of just clearing the caches

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,126 @@
...
+        ]
+        if any([field in values.keys() for field in cache_invalidating_fields]):
+            # clear cache for all ormcache methods.
+            self.clear_caches()
+        return super(AuditlogRule, self).write(values)
+
+    @api.onchange("model_id")
+    def onchange_model_id(self):
+        # if model changes we must wipe out all field ids
+        self.auditlog_line_access_rule_ids.unlink()
```
hbrunn:
```suggestion
        self.auditlog_line_access_rule_ids = False
```

onchange methods should work on the memory, not the database

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo.osv.expression import OR
+
+
+class AuditlogLogLine(models.Model):
+    _inherit = "auditlog.log.line"
+    _order = "create_date desc"
+
+    user_id = fields.Many2one(related="log_id.user_id")
+    method = fields.Char(related="log_id.method")
+    model_id = fields.Many2one(related="log_id.model_id", store=True)
```
hbrunn, line 15: can we get rid of storing this?

auditlog_security/models/auditlog_log.py
ntsirintanis: Yeah, basically the rule_id computation can also happen in the line model, so I'll see into removing this

auditlog_security/models/auditlog_log_line.py
ntsirintanis: Actually we need it stored for filtering purposes in views etc

auditlog_security/models/auditlog_rule.py
ntsirintanis: Well sure, but when we unsubscribe we get rid of existing action anyway

auditlog_security/models/auditlog_rule.py
ntsirintanis: In that sense then we do not need to assign any groups_id here. We just need to make sure that the user is also an auditlog user

ntsirintanis: ... which is admin work

ntsirintanis:
@hbrunn it appears that we are almost there

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,40 @@
...
+from odoo.osv.expression import OR
+
+
+class AuditlogLogLine(models.Model):
+    _inherit = "auditlog.log.line"
+    _order = "create_date desc"
+
+    user_id = fields.Many2one(related="log_id.user_id")
+    method = fields.Char(related="log_id.method")
+    model_id = fields.Many2one(related="log_id.model_id", store=True, reaonly=True)
```
Kiplangatdan: @ntsirintanis Typo `readonly`

auditlog_security/security/res_groups.xml
Kiplangatdan: @ntsirintanis The `View Audit Logs` group is still being used in the Inuka instances. 
Currently, it has around 11 users. 
It's also being used in approx. 21 record rules. 
This prevents to module from being upgraded.

Kiplangatdan (changes requested):

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,127 @@
...
+            "state",
+            "auditlog_line_access_rule_ids",
+            "log_selected_fields_only",
+        ]
+        if any([field in values.keys() for field in cache_invalidating_fields]):
+            # clear cache for all ormcache methods.
+            self.clear_caches()
+        res = super().write(values)
+        if self._register_hook():
+            modules.registry.Registry(self.env.cr.dbname).signal_changes()
```
hbrunn: super does this already. my suggestion is to replace line 62 with

```python
modules.registry.Registry(self.env.cr.dbname).signal_changes()
```

auditlog_security/models/auditlog_log_line.py
hbrunn:
```suggestion
    model_id = fields.Many2one(related="log_id.model_id")
```

I still strongly think those fields shouldn't be stored, and related fields are readonly by default

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,40 @@
...
+    res_id = fields.Integer(related="log_id.res_id")
+    rule_id = fields.Many2one("auditlog.rule", compute="_compute_rule_id", store=True)
```
hbrunn:
```suggestion
```

see below

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,40 @@
...
+        search="_search_allowed_group_ids",
+    )
+
+    def _search_allowed_group_ids(self, operator, value):
+        access_rules = self.env["auditlog.line.access.rule"].search(
+            ["|", ("group_ids", operator, value), ("group_ids", "=", False)]
+        )
+        domains = []
+        for access_rule in access_rules:
+            domain = [("rule_id", "=", access_rule.auditlog_rule_id.id)]
```
hbrunn:
```suggestion
            domain = [("log_id.model_id", "=", access_rule.auditlog_rule_id.model_id.id)]
```

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,40 @@
...
+                domain.append(("field_id", "in", access_rule.field_ids.ids))
+            domains.append(domain)
+        return OR(domains)
+
+    @api.depends("model_id")
+    def _compute_rule_id(self):
+        for line in self:
+            line.rule_id = self.env["auditlog.rule"].search(
+                [("model_id", "=", line.model_id.id)]
+            )
```
hbrunn:
```suggestion
```

auditlog_security/security/res_groups.xml
hbrunn: the rules from your legacy module should be deleted anyways, as you don't need them any more with this approach. for easier transition, mark them as noupdate in your legacy code, then those records won't be touched on module upgrade

hbrunn (changes requested):
indeed nearly there, and tests will be much simpler too. don't forget to squash commits in the end

auditlog_security/security/res_groups.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_can_view_audit_logs" model="res.groups">
```
ntsirintanis: @thomaspaulb the reason that lines are hidden, is this group. For some reason, users under this group can create, update, delete auditlog.line.access.rule records, but cannot read them.

auditlog_security/models/auditlog_log_line.py
ntsirintanis: OK. Since we use model_id for filtering in some tree views, shall I just create a computed and stored field stored_model_id and use that in the views? Or make this one searchable?

ntsirintanis: basically we do not even need that field anymore, removing

ntsirintanis:
@hbrunn in addition to all remaining changes, I also want to provide a migration script. As the changes between v 11.0 and v. 16.0 of this module are radical, how do you propose we do this?

thomaspaulb:
@ntsirintanis The 11 module was never merged though, so I think we can suffice with a script for internal use only

auditlog_security/models/auditlog_log_line.py
thomaspaulb: Yes, it's nice to be able to group/filter/sort on these fields without trouble.

Actually - can we also add a combined index on (model_id, res_id), for search purposes?

auditlog_security/models/auditlog_log_line.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class AuditlogLogLine(models.Model):
+    _inherit = "auditlog.log.line"
+    _order = "create_date desc"
+
+    user_id = fields.Many2one(related="log_id.user_id")
+    method = fields.Char(related="log_id.method")
+    model_id = fields.Many2one(related="log_id.model_id", store=True)
+    res_id = fields.Integer(related="log_id.res_id")
```
thomaspaulb, line 15: This field should then also become stored.

auditlog_security/models/auditlog_log_line.py
thomaspaulb: ... on the other hand, the storing of these fields is a bit of a downer if you have auditlog installed, a lot of lines in the system, and now you want to install this module but it takes ages for the stored fields to populate.

@ntsirintanis non-stored related fields also support searching and grouping, right? Maybe we can live without the sort option, and then remove the storedness of these fields.

thomaspaulb: Though we do definitely need the combined index, if not on log line, then on log.

thomaspaulb: `create index auditlog_log_model_id_res_id on auditlog_log(model_id, res_id);`

hbrunn: auto_join on log_id should help here

auditlog_security/models/auditlog_rule.py
```diff
@@ -0,0 +1,127 @@
...
+            rule.auditlog_line_access_rule_ids.regenerate_rules()
+        # rule now will have "View Log" Action, make that visible only for admin
+        if res:
+            self.action_id.write({
+                'groups_id': [(6, 0, [self.env.ref('base.group_system').id])]
+            }) 
+        return res
+
+    @api.multi
+    def unsubscribe(self):
```
ntsirintanis, line 121: @hbrunn we somehow need to restore this functionality. User can either subscribe to a rule (so they can see changelogs etc), or unsubscribe from a rule, and stop seeing anything related to that model. Do you have yet another clever way to achieve this easily?

thomaspaulb: @ntsirintanis In the 11.0 version the subscribing / unsubscribing was not a per-user thing, it was just the enabling or disabling of that rule, and maybe not even that, but mainly the creation or removal of the "View Log" server action on the model. My idea behind it was to mimick the "Mass Edit" module in this, where you can say "Create server action".

So maybe it's just that the naming of the button is off and it should just be "Create Server action" and/or "Remove server action". Perhaps you can check how OCA mass edit (now called server_action_mass_edit) calls it, although that implementation changed also and you have to actually manually create a server action yourself in order to use mass editing. But perhaps you can look at an older major-Odoo-version-version of that module.

PR #3031, opened 2024-09-12
Migrate module_analysis to v17

jguerriat:
This is a migration of the V16 module "module_analysis" that gives insight about installed or available module for your odoo instance.
There are only small changes done (in the manifest 16.0.1.0.1 => 17.0.1.0.1, and in a view, i removed  **states="installed"** since this usage is deprecated, and replaced it by **invisible="state != 'installed'"**

The rest of the module remains the same


legalsylvain:
hi @jguerriat . Thanks for porting this module. 
could you change the way you migrated the module, cherry picking commits ? 

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

- it allows to respect history and authorship
- it is more simple to review the "diff" you propose

jguerriat:
@legalsylvain Hello, thanks for your comment, i just cherry picked the commits for the module to maintain module history :).
There are 3 failing checks, i think it all comes from the missing external depency "pygount". I added it in the requirements.txt of the repository, should i do something else so it can pass the tests?

legalsylvain:
in the root folder of the repo, could you run locally : 
```
pre-commit install
pre-commit run -a

```

It should show errores, and automatically fixes some trivial troubles.

you can then commit -a --amend on the last migration commit you did.

legalsylvain (approved):
Diff review. no test.

thanks for porting this module.

Abranes (approved):

PR #3034, opened 2024-09-18
[15.0][FIX] excel_import_export: require attach when res_id false

Saran440:
`attachment_ids` shouldn't required when we import with `import_file`

Step to test:
1. install module excel_import_export_demo
2. import sale order. it will require attachment_ids
Selection_007


Ffferrarii (approved):
Already test

kittiu (approved):

PR #3036, opened 2024-09-26
[16][IMP] attachment_queue automatic retry

florian-dacosta:
For now, when an attachment queue is processsed by a `queue.job`, there is no automatic retry in case of concurrent update because all errors are caught.
And, for the same reason, the RetryableJobError does not work.
This PR aims to restore a consistent behavior about this.
The only bad point is that, in case of RetryableJobError, the failure email will be sent anyway, even if the job will retry the attachment. It would then be kind of a false positive. I don't think to have any other choice, because at the time we send the mail, we can't really know if the job will retry or not (we do not know if the max_retries will be reached).

@kevinkhao @bealdav @SirAionTech

### DIFF
--- a/attachment_queue/models/attachment_queue.py
+++ b/attachment_queue/models/attachment_queue.py
@@ -3,9 +3,11 @@
 import logging
 
 import psycopg2
+from psycopg2 import OperationalError
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
 
 from odoo.addons.queue_job.exception import RetryableJobError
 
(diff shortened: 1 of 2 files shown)


bealdav (approved):
Clean exceptions documented and managed. Thanks

attachment_queue/models/attachment_queue.py
```diff
@@ -115,18 +117,38 @@ def run_as_job(self):
...
+                # a rollback has been made before because of the savepoint.
+                # we need to commit because we re-raise the exception and a rollback
+                # will be performed
+                self.env.cr.commit()  # pylint: disable=E8102
+                raise
+            except Exception as err:
+                self._set_attachment_failure(err)
+
+    def _set_attachment_failure(self, error):
+        _logger.warning(STR_ERROR_DURING_PROCESSING.format(self.id) + str(error))
```
kevinkhao, line 145: I could overload this but to make it clean: what do you think of adding at the end of this function a hook for failure side effects ?

For example your attachment.queue is supposed to do something on a sale order. Then you could use this hook in that way:


    class SaleProcessorException(Exception):
        def __init__(self, message, record=False):
            super().__init__(message)
            self.record = record


on attachment.queue


    def _run_sale_order_processing(self)
        try:
            ...
        except ValueError as e:
            raise SaleProcessorException("Error during processing or sale order xyz", sale_order)

    def _side_effect_error_hook(self, exc):  # I would probably just name it _error_hook
        if type(e) is SaleProcessorException:
            e.record.attachment_queue_id = self
            e.record.attachment_queue_warning = f"An automated processing failed: str(e)"
        ...

The point is to have a clean way to make persistent changes on failure.

florian-dacosta: Well, I guess we could add some kind of `_error_hook`, but I am not sure why it is needed ?
`_set_attachment_failure` is already a usable hook for this no?

With your example, you just would have : 
```
def _set_attachment_failure(self, exc):
    super()._set_attachment_failure()
    if type(e) is SaleProcessorException:
        e.record.attachment_queue_id = self
        e.record.attachment_queue_warning = f"An automated processing failed: str(e)"
```

I don't see any gain of having one more hook, but I may miss something.

kevinkhao: It's not necessary and is a personal preference, trading some conciseness for explicitness

bealdav:
About this PR, you approve or not ? @kevinkhao 

You may define other PR for other features I suppose ?

kevinkhao (approved):

SirAionTech (approved):
Thanks!
It took some time to understand what's happening but the comments helped a lot

PR #3038, opened 2024-09-27
[16][FIX] attachment_queue : fix default channel name

florian-dacosta:
The created channel on the module is a child of "root", so  we have to pass the channel name as `root.attachment_queue`
The consequence passing only "attachment_queue" like it is now, is that the done jobs won't be deleted by the vacuum cron.

I am not sure there is other consequences, from what I test, even without the "root" prefix, the jobs were processed with the right amount of workers configured on the root.attachment_queue channel...

@bealdav @kevinkhao


bealdav (approved):

gbrito (approved):

kevinkhao (approved):
LGTM

PR #3043, opened 2024-10-03
[12.0][FIX] module_analysis: interpret exclude_directories as relative to module's folder

### LINKED ISSUE #3042: module_analysis does not count code lines from odoo core when installed from the debian package
## Module
module_analysis

## To Reproduce
**Affected versions**: all

1. Prerequisites: Odoo is installed using the debian package; module_analysis is installed
2. Go to the account_cancel module page (in Applications app) in debug mode
3. Press the "Refresh Code Analysis" button in the "Technical data" tab

**Expected behavior**
The Python Code Quantity field's value is 0 also it should not, as there are non-empty python files in the module (models/account_bank_statement.py).

**Additional context**
Debugging shows that [this line](https://github.com/OCA/server-tools/blob/12.0/module_analysis/models/ir_module_module.py#L167) is a true condition when odoo is installed by a debian package (root = /usr/lib/python3/dist-packages/odoo/addons/account_cancel), because `exclude_directories` value is `['lib', 'demo', 'test', 'tests', 'doc', 'description']`, so the test returns {'lib'}

The `exclude_directories` comes from ir.config_parameter `module_analysis.exclude_directories` which interpretation is questionable. As a user of the module I expect its value to list folders relative to the module's folder, while in the code if any of the listed folder is found in the module's root path the entire module will be ignored.

I'm working on a PR to fix this.

fcayre:
Fixes #3042

### DIFF
--- a/module_analysis/models/ir_module_module.py
+++ b/module_analysis/models/ir_module_module.py
@@ -164,7 +164,7 @@ class IrModuleModule(models.Model):
         if not path:
             return res
         for root, dirs, files in os.walk(path, followlinks=True):
-            if set(Path(root).parts) & set(exclude_directories):
+            if set(Path(root).relative_to(path).parts) & set(exclude_directories):
                 continue
             for name in files:
                 if name in exclude_files:


fcayre:
Made a 17.0-based PR instead, so closing this one. Sorry for the noise.

PR #3051, opened 2024-10-07
[16.0][FIX] tracking_manager: Avoid error when accessing the value of a field that we do not have access

victoralmau:
Avoid error when accessing the value of a field that we do not have access

Example use case:
- Install `hr_fleet`
- Set the `employee_ids` field of `hr.employee` as a tracking field
- Modify the user to not have permissions in Fleet
- Modify the name of an employee

@Tecnativa TT51160

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -6,6 +6,7 @@
 from collections import defaultdict
 
 from odoo import api, models, tools
+from odoo.exceptions import AccessError
 
 from ..tools import format_m2m
 
@@ -120,7 +121,11 @@ class Base(models.AbstractModel):
             values = initial_values.setdefault(record.id, {})
             if values is not None:
                 for fname in fnames:
-                    values.setdefault(fname, record[fname])
+                    try:
+                        values.setdefault(fname, record[fname])
+                    except AccessError:
+                        # User does not have access to the field (example with groups)
+                        continue
 
     def _tm_finalize_o2m_tracking(self):
         initial_values = self.env.cr.precommit.data.pop(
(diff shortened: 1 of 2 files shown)


tracking_manager/models/models.py
```diff
@@ -120,7 +121,11 @@ def _tm_prepare_o2m_tracking(self):
             values = initial_values.setdefault(record.id, {})
             if values is not None:
                 for fname in fnames:
-                    values.setdefault(fname, record[fname])
+                    try:
+                        values.setdefault(fname, record[fname])
+                    except AccessError:
+                        # User does not have access to the field (example with groups)
+                        break
```
chienandalu: Doesn't this avoid that the tracking is made?

victoralmau: No, this is intended to avoid errors.

All possible values that may be used in the tracking message are set, but if a field has groups, it should not block.

The use case example tries to explain what has been happening so far.

pedrobaeza: I would say this for getting the rest of the fields:

```suggestion
                        continue
```

pedrobaeza:
Can you add a regression test?

victoralmau:
> Can you add a regression test?

I have tried to do it but it is not possible (it is necessary a relationship with `one2many` that can be created/modified in that model and that some of the fields are limited with `groups`). I have taken the opportunity to add a test of the use case of https://github.com/OCA/server-tools/pull/3051

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #3056, opened 2024-10-08
[16.0][FIX] tracking_manager: Test compatibility

victoralmau:
Test compatibility

Related to https://github.com/OCA/server-tools/pull/3051/commits/b41e7ae2c77bf1c157087cc41893e134fab247f9

Do not set the `child_ids` field as tracking in the setup to avoid incorrect data in some cases

Please @pedrobaeza can you review it?

@Tecnativa TT51160

PR #3060, opened 2024-10-09
[17.0][FIX] tracking_manager: Avoid error when accessing the value of a field that we do not have access

victoralmau:
FWP from 16.0: https://github.com/OCA/server-tools/pull/3051 + https://github.com/OCA/server-tools/pull/3056

Avoid error when accessing the value of a field that we do not have access

Example use case:
- Install `hr_fleet`
- Set the `employee_ids` field of hr.employee as a tracking field
- Modify the user to not have permissions in Fleet
- Modify the name of an employee

@Tecnativa TT51160

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -6,6 +6,7 @@
 from collections import defaultdict
 
 from odoo import api, models, tools
+from odoo.exceptions import AccessError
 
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #3071, opened 2024-10-14
[17.0][MIG] fetchmail_attach_from_folder: Migration to 17.0

zamberjo:
Ping @NL66278, if #3017 is accepted I add it.


anmarmo1 (approved):
Functional review is ok

NL66278:
@anmarmo1 Could you also look at #3017 ? Would be a pity if the migration does not contain this.

anmarmo1:
Yes, I will review it

zamberjo:
Hi @NL66278 have you had a chance to take a look at the PR?

fetchmail_attach_from_folder/wizard/attach_mail_manually.py
```diff
@@ -96,12 +96,12 @@ class AttachMailManuallyMail(models.TransientModel):
     _name = "fetchmail.attach.mail.manually.mail"
     _description = __doc__
 
-    wizard_id = fields.Many2one("fetchmail.attach.mail.manually", readonly=True)
-    msgid = fields.Char("Message id", readonly=True)
-    subject = fields.Char(readonly=True)
-    date = fields.Datetime(readonly=True)
-    email_from = fields.Char("From", readonly=True)
-    body = fields.Html(readonly=True)
+    wizard_id = fields.Many2one("fetchmail.attach.mail.manually")
```
NL66278: Why are you removing the `readonly=True` here and elsewhere?

zamberjo: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

> If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.

But I just realized that I didn't put them in the view, when I can I fix it.

NL66278: @zamberjo Will you update the views then?

zamberjo: Hello, as it is a wizard I have left it as it was because I don't see any sense in importing a wizard.

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -26,7 +26,6 @@ class FetchmailServerFolder(models.Model):
     state = fields.Selection(
         [("draft", "Not Confirmed"), ("done", "Confirmed")],
         string="Status",
-        readonly=True,
```
NL66278: Please restore this readonly, as the state should only change through actions, and then I think the PR will be ready to merge.

zamberjo: done

NL66278 (approved):
:+1: LGTM Thank you!

PR #3084, opened 2024-10-17
[18.0][MIG] scheduler_error_mailer

astirpe:


scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,54 @@
...
+            template = my_cron.email_template_id.with_context(**context).sudo()
+            template.send_mail(my_cron.id, force_send=True)
+
+    @api.model
+    def _test_scheduler_failure(self):
+        """This function is used to test and debug this module."""
+        raise UserError(_("Task failure with UID = %d.") % self._uid)
+
+    @api.model
+    def _callback(self, cron_name, server_action_id, job_id):
```
CasVissers-360ERP: @astirpe the original method has changed, can you check?

StefanRijnhart (approved):
Tests are fixed in https://github.com/OCA/server-tools/pull/3190. Please rebase once that one is merged.

astirpe:
@StefanRijnhart I will, thank you!

astirpe:
rebased

scheduler_error_mailer/models/ir_cron.py
StefanRijnhart: Please update your review @CasVissers-360ERP

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -25,28 +25,26 @@ class IrCron(models.Model):
...
     def _test_scheduler_failure(self):
         """This function is used to test and debug this module."""
         raise UserError(_("Task failure with UID = %d.") % self._uid)
+
+    @api.model
+    def _callback(self, cron_name, server_action_id):
+        try:
+            return super()._callback(cron_name, server_action_id)
+        except Exception as e:
+            self._handle_callback_exception(cron_name, server_action_id, e)
```
CRogos, line 50: Is it intended that the exception is not raised again?

maybe we also should move the _handle_callback_exception in this method when odoo also removed it?

CRogos (changes requested):
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

Squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.
Here are some examples:
image

Maybe we could also improve the test and test "def _callback" instead of "def _handle_callback_exception"?

scheduler_error_mailer/tests/test_scheduler_error_mailer.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from unittest.mock import patch
+
+from odoo.tests.common import TransactionCase
+
+from odoo.addons.scheduler_error_mailer.hooks import post_init_hook
+
+
+class TestSchedulerErrorMailer(TransactionCase):
```
CRogos: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0
Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

CRogos:
@astirpe are you still working on this?

astirpe:
@CRogos Thank you for your review! I'm very busy with projects, I could continue with this when I can find some time. Sorry

scheduler_error_mailer/models/ir_cron.py
astirpe: ~I think this is correct as it is. The Odoo base _callback() only logs any exception, not raising.~
And I honestly can't see any reason for moving _handle_callback_exception() inside here.

scheduler_error_mailer/tests/test_scheduler_error_mailer.py
astirpe: Done, thanks!

scheduler_error_mailer/models/ir_cron.py
astirpe: I checked again and indeed the Odoo base _callback() is raising the exception.
I added this commit to re-raise it again:
https://github.com/OCA/server-tools/pull/3084/commits/31459b46d67fbce434394940690ded476b9f07a2
I also squashed the commits you mentioned, thanks!

StefanRijnhart:
@CRogos Is it alright for you now?

StefanRijnhart:
Oh, we're getting errors because the mail template body is no longer valid due to the ir.cron datamodel changes. @astirpe would you consider taking https://github.com/astirpe/server-tools/pull/1?

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -48,3 +48,6 @@ def _callback(self, cron_name, server_action_id):
             return super()._callback(cron_name, server_action_id)
         except Exception as e:
             self._handle_callback_exception(cron_name, server_action_id, e)
+
+            # Re-raise the original job exception
+            raise e
```
CRogos: Can't we write raise without the e to reraise?

astirpe:
Thank you @StefanRijnhart !

scheduler_error_mailer/migrations/18.0.1.0.0/post-migration.py
CRogos, line 1: Are you sure this is the best option?

Couldn't we just call load_data with the ir_cron_email_tpl.xml file?
https://github.com/OCA/openupgradelib/blob/37cc7eeb1383fbe7c3d5cd2666af3b73338a90cd/openupgradelib/openupgrade.py#L292

Something like this:
https://github.com/OCA/account-financial-reporting/commit/8ba049e8

CRogos (changes requested):
Can you squash your latest changes with the Migration commit and also cleanup the commits above according to the migration guide?

scheduler_error_mailer/migrations/18.0.1.0.0/post-migration.py
StefanRijnhart: That would be equivalent to calling `reset_template` on the template, but it would reset other fields as well such as subject and recipients which are customized in our setups. Also, this bit takes care to reset the body in each language (as per code taken from `reset_template`).

StefanRijnhart: For reference: code is based on https://github.com/odoo/odoo/blob/18.0/addons/mail/models/template_reset_mixin.py#L70-L93

StefanRijnhart:
Apologies @astirpe, it turns out that in migrated databases the template_fs of the template might not be set, so here is https://github.com/astirpe/server-tools/pull/2. And please rebase and squash all migration commits as @CRogos suggests.

astirpe:
I will do the rebase and squash, thanks!

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,53 @@
...
+
+            _logger.debug("Sending scheduler error email with context=%s", context)
+
+            template = self.email_template_id.with_context(**context).sudo()
+            template.send_mail(self.id, force_send=True)
+
+    @api.model
+    def _test_scheduler_failure(self):
+        """This function is used to test and debug this module."""
+        raise UserError(_("Task failure with UID = %d.") % self._uid)
```
CRogos: _ should also be replaced

`raise UserError(self.env._("Task failure with UID = %(uid)d.", uid=self._uid))`

scheduler_error_mailer/models/ir_cron.py
astirpe: Yes, done

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -25,28 +25,29 @@ class IrCron(models.Model):
...
-
-        return res
+            template = self.email_template_id.with_context(**context).sudo()
+            template.send_mail(self.id, force_send=True)
 
     @api.model
     def _test_scheduler_failure(self):
         """This function is used to test and debug this module."""
-        raise UserError(_("Task failure with UID = %d.") % self._uid)
+        raise UserError(_("Task failure with UID = %(uid)d."), uid=self._uid)
```
CRogos: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

You can replace the call to translate a string (_) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,53 @@
...
+
+            _logger.debug("Sending scheduler error email with context=%s", context)
+
+            template = self.email_template_id.with_context(**context).sudo()
+            template.send_mail(self.id, force_send=True)
+
+    @api.model
+    def _test_scheduler_failure(self):
+        """This function is used to test and debug this module."""
+        raise UserError(_("Task failure with UID = %(uid)d."), uid=self._uid)
```
CRogos: @astirpe  are you still working on this

```suggestion
        raise UserError(self.env._("Task failure with UID = %(uid)d.", uid=self._uid))
```

scheduler_error_mailer/models/ir_cron.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2012-2013 Akretion Sébastien BEAU,David Beal,Alexis de Lattre
+# Copyright 2016 Sodexis
+# Copyright 2018 bloopark systems (<http://bloopark.de>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import _, api, fields, models
```
CRogos:
```suggestion
from odoo import api, fields, models
```

CRogos (changes requested):

CRogos:
@astirpe are you still working on this?

scheduler_error_mailer/models/ir_cron.py
astirpe: Done, thanks!

CRogos:
There are still some commits which should be squashed: 

https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate
image
(there are more these are only examples)

CRogos:
@astirpe could you do the squashing? I think we can merge afterwards.

astirpe:
@CRogos squashed, thanks!

CRogos (approved):

PR #3086, opened 2024-10-21
[16.0][ADD] upgrade_analysis: add blacklist from v17

hbrunn:
this adds the same blacklist as in [v17](https://github.com/OCA/server-tools/blob/17.0/upgrade_analysis/blacklist.py#L1-L6) because the depreciation has been [backported](https://github.com/OCA/OCB/commit/c9072dfe798d658246ca6fc255f95eb884a6d389), leading to problems with re-running the analysis for 16/17

### DIFF
--- a/upgrade_analysis/blacklist.py
+++ b/upgrade_analysis/blacklist.py
@@ -1,4 +1,9 @@
-BLACKLIST_MODULES = []
+BLACKLIST_MODULES = [
+    "payment_alipay",
+    "payment_ogone",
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Note that I did the same in #2739

hbrunn:
oh thanks, so I'll close this one. Just busy regenerating the analysis files as it seems the ones checked into openupgradev17 are outdated

PR #3088, opened 2024-10-22
[17.0][FIX] auditlog: registry propagation

gurneyalex:
When a new auditlog rule is added / modified / deleted, all workers need to be notified of the change. This is done through registry signaling. The previous code was not using the proper level of signaling resulting in workers not being aware of the changes and not implementing the correct auditlog rules, because they had only invalidated their cache and not reloaded the registry.

We fix this by using the same signaling as implemented in `base_automation` which sets the the registry_invalidated field of env.registry to True to cause a full registry reload.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -737,7 +737,8 @@ class AuditlogRule(models.Model):
         return vals_list
 
     def _update_registry(self):
-        """Update the registry after a modification on automation rules."""
+        """Force a registry reload after rule change"""
+        # this code comes from `base_automation` which has a similar need
         if self.env.registry.ready and not self.env.context.get("import_file"):
             # notify other workers
             self.env.registry.registry_invalidated = True
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -3,7 +3,7 @@
 
 import copy
 
-from odoo import _, api, fields, models, modules
+from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
 FIELDS_BLACKLIST = [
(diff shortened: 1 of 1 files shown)


auditlog/models/rule.py
```diff
@@ -735,3 +735,9 @@ def _update_vals_list(self, vals_list):
                 if isinstance(fieldvalue, models.BaseModel) and not fieldvalue:
                     vals[fieldname] = False
         return vals_list
+
+    def _update_registry(self):
+        """Update the registry after a modification on automation rules."""
```
pedrobaeza: automation?

Apart from that, isn't better to refer that this is mimicking what is shown in `base_automation`?

gurneyalex: copy paste error :sweat:

sebalix (approved):

auditlog/models/rule.py
```diff
@@ -735,3 +735,10 @@ def _update_vals_list(self, vals_list):
                 if isinstance(fieldvalue, models.BaseModel) and not fieldvalue:
                     vals[fieldname] = False
         return vals_list
+
+    def _update_registry(self):
+        """Force a registry reload after rule change"""
+        # this code comes from `base_automation` which has a similar need
+        if self.env.registry.ready and not self.env.context.get("import_file"):
+            # notify other workers
+            self.env.registry.registry_invalidated = True
```
sebalix, line 744: Is it something compatible on Odoo < 17.0 that should be backported?

ivantodorovich (approved):

auditlog/models/rule.py
sebalix: That seems at least compatible with Odoo 16.0, didn't checked previous versions.

max3903 (approved):

PR #3090, opened 2024-10-23
[17.0][FIX] upgrade_analysis: fix constraint logging

hbrunn:
In https://github.com/OCA/OpenUpgrade/pull/4606, I noticed all constraints were put down as deleted, so after some pain and misery I identified https://github.com/OCA/OCB/commit/4c9968397b0714bc90a9c94c4673bd3148db4010 as the cause of the issue, because at the point our patched function is called, the constraint exists already, so `_reflect_constraint`'s result is always falsy, and that's why we never get to log anything there.

As `log_xml_id` already takes care of not inserting duplicates, I see no harm in calling this unconditionally.

### DIFF
--- a/upgrade_analysis/odoo_patch/odoo/addons/base/models/ir_model.py
+++ b/upgrade_analysis/odoo_patch/odoo/addons/base/models/ir_model.py
@@ -1,4 +1,5 @@
 from odoo.addons.base.models import ir_model
+from odoo import models
 
 from ...... import upgrade_log
 from .....odoo_patch import OdooPatch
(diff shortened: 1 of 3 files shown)


hbrunn:
hmm, the PR above still displays wrong definitions it seems

upgrade_analysis/odoo_patch/odoo/addons/base/models/ir_model.py
```diff
@@ -32,11 +32,11 @@ def cons_text(txt):
...
-        if data_list:
-            self.env["ir.model.data"]._update_xmlids(data_list)
             # Begin OpenUpgrade addition
-            for data in data_list:
-                xml_id = data.get("xml_id")
-                module = xml_id.split(".")[0]
-                upgrade_log.log_xml_id(self.env.cr, module, xml_id)
+            xml_id = f"{module}.constraint_{conname}"
+            upgrade_log.log_xml_id(self.env.cr, module, xml_id)
+            self.pool.loaded_xmlids.add(xml_id)
```
hbrunn: this one I found dubious in the beginning but it makes sense: We init the modules twice, after the first loading process `pool.loaded_xmlids` is cleared, and the second one won't see those xmlids because of the change in core linked above. Standard Odoo doesn't encounter this issue, so no problem for them, but we have to take care that the xmlids of constraints are touched because otherwise Odoo deletes all constraints in the second `_process_end`, leading to my intermediate result with the definitions being empty

StefanRijnhart: Good find! If I understand correctly, vanilla Odoo does not mark the xml id of a previously existing constraint as loaded. But why does vanilla Odoo not remove the reflected constraints after a module update like any other xml id that is not registered as loaded? Can you help me understand?

hbrunn: oh I see now it does:

https://github.com/OCA/OCB/blob/17.0/odoo/addons/base/models/ir_model.py#L1882-L1883

that's not part of the diff I linked, and I missed this change when comparing the function to the current code. It's functionally equivalent to the above, but to be in line with the current code I refreshed the function from HEAD in a fixup commit so you see what's going on (let me know when to squash this).

While being on it I excluded the whole odoo patch folder from pre-commit, so that we stay as close as possible to the code we patch in order not to confuse our natural pattern matcher (I didn't manage to get something useful from meld there)

hbrunn: PS: I wonder if it's worthwhile to try and replace this patch, and the ones in

- https://github.com/OCA/server-tools/blob/17.0/upgrade_analysis/odoo_patch/odoo/tools/convert.py
- https://github.com/OCA/server-tools/blob/17.0/upgrade_analysis/odoo_patch/odoo/models.py

with a patch on [_load_xmlid](https://github.com/OCA/OCB/blob/17.0/odoo/addons/base/models/ir_model.py#L2334) and [_update_xmlids](https://github.com/OCA/OCB/blob/17.0/odoo/addons/base/models/ir_model.py#L2260)?

Those we can do without copying the code and avoid this class of problem altogether. Will test this in a couple of hours when my current test run for what I did above is done.

StefanRijnhart: Thanks for checking, and I like your suggestion to patching the xmlid methods themselves.

hbrunn: my first test of this failed and currently I can't dive into this, so let's go with what we have, that generates correct files. squashed the fixup commit

StefanRijnhart (approved):

pedrobaeza (approved):

PR #3095, opened 2024-10-26
[15.0][IMP] excel_import_export: allow selection of all related templates for export action

AungKoKoLin1997:
Before this PR, when two templates exist for the same model, clicking `ADD EXPORT ACTION` for each template
creates separate server actions for each. As a result, if multiple templates are created for the same model,
numerous server actions with the label `Export Excel` will accumulate for that model.

With this PR, if two templates exist for the same model and `ADD EXPORT ACTION` is clicked for each,
they will now share a single server action. Users can then select templates from within the `Export Excel` wizard.
Clicking `REMOVE EXPORT ACTION` on a template will remove only that template from the `Export Excel` selection.
The server action itself will be deleted only when no templates use it for export.

@qrtl QT4815

### DIFF
--- a/excel_import_export/models/xlsx_template.py
+++ b/excel_import_export/models/xlsx_template.py
@@ -429,31 +429,53 @@ self['{}'] = self.env['{}'].search(self.safe_domain(self.domain))
                 inst_dict[itype] = rec.post_import_hook
             rec.instruction = inst_dict
 
-    def add_export_action(self):
-        self.ensure_one()
+    def _get_export_action_domain(self, model):
+        return [
+            ("binding_model_id", "=", model.id),
+            ("res_model", "=", "export.xlsx.wizard"),
+            ("name", "=", "Export Excel"),
+        ]
+
+    def _get_export_action(self, model):
+        export_action_domain = self._get_export_action_domain(model)
+        return self.env["ir.actions.act_window"].search(export_action_domain, limit=1)
+
+    def _create_export_action(self, model):
         vals = {
             "name": "Export Excel",
             "res_model": "export.xlsx.wizard",
-            "binding_model_id": self.env["ir.model"]
-            .search([("model", "=", self.res_model)])
-            .id,
+            "binding_model_id": model.id,
             "binding_type": "action",
             "target": "new",
             "view_mode": "form",
             "context": """
                 {'template_domain': [('res_model', '=', '%s'),
-                                     ('fname', '=', '%s'),
-                                     ('gname', '=', False)]}
+                                    ('export_action_id', '!=', False),
+                                    ('gname', '=', False)]}
             """
-            % (self.res_model, self.fname),
+            % (self.res_model),
         }
-        action = self.env["ir.actions.act_window"].create(vals)
-        self.export_action_id = action
+        return self.env["ir.actions.act_window"].create(vals)
+
+    def add_export_action(self):
+        self.ensure_one()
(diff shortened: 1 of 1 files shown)


excel_import_export/models/xlsx_template.py
```diff
@@ -122,6 +122,9 @@ class XLSXTemplate(models.Model):
         comodel_name="ir.actions.report",
         string="Report Action",
     )
+    is_display = fields.Boolean(
+        help="Technical field to control template visibility in server actions."
+    )
```
yostashiro: I think `is_available` may make a better name.

excel_import_export/models/xlsx_template.py
```diff
@@ -431,29 +434,51 @@ def _compute_output_instruction(self):
 
     def add_export_action(self):
```
yostashiro: I think this method can be restructured in the following manner for better readability:

```
    def add_export_action(self):
        self.ensure_one()
        model = self.env["ir.model"].search([("model", "=", self.res_model)], limit=1)
        export_action = self._get_export_action(model)
        if not export_action:
            export_action = self._create_export_action(model)
        self.export_action_id = export_action
        self.is_available = True
```

excel_import_export/models/xlsx_template.py
```diff
@@ -431,29 +434,51 @@ def _compute_output_instruction(self):
...
+                                    ('gname', '=', False)]}
             """
-            % (self.res_model, self.fname),
+            % (self.res_model),
         }
         action = self.env["ir.actions.act_window"].create(vals)
         self.export_action_id = action
+        self.is_display = True
 
     def remove_export_action(self):
```
yostashiro, line 468: To be a bit more concise:
```
    def remove_export_action(self):
        self.ensure_one()
        self.is_available = False
        if not self.search(
            [("res_model", "=", self.res_model), ("is_available", "=", True)]
        ):
            self.export_action_id.unlink()
        self.export_action_id = False
```

excel_import_export/models/xlsx_template.py
yostashiro: On second thought, we can actually remove this field.  Just adjust `template_domain` of the export action context to use `export_action_id`.

AungKoKoLin1997: I prefer second option to avoid adding new field.

excel_import_export/models/xlsx_template.py
```diff
@@ -429,31 +429,53 @@ def _compute_output_instruction(self):
...
-        if self.export_action_id:
+        if not self.search(
+            [
+                ("res_model", "=", self.res_model),
+                ("export_action_id", "!=", False),
+                ("id", "!=", self.id),
+            ]
+        ):
             self.export_action_id.unlink()
+        self.export_action_id = False
```
yostashiro: The condition `("export_action_id", "!=", False)` is probably not optimal, considering there could be cases where multiple export actions exist for a model.

```suggestion
        export_action = self.export_action_id
        self.export_action_id = False
        if not self.search(
            [("res_model", "=", self.res_model), ("export_action_id", "=", export_action.id)]
        ):
            export_action.unlink()
```

yostashiro (approved):
Code review.

yostashiro:
Thanks, Kitti!

@AungKoKoLin1997 Can you please forward-port this.

PR #3117, opened 2024-11-12
Fix patch message on module_change_auto_install

jeroenev:
The patch message in the logs contains some typo's. This small PR fixes those


legalsylvain (approved):
Thanks.

/ocabot merge patch

PR #3100, opened 2024-10-29
[18.0][MIG] base_view_inheritance_extension

fredericgrall:

PR #3120, opened 2024-11-14
[18.0][MIG] base_view_inheritance_extension: Migration to 18.0

LucasTran380381:
Supersedes https://github.com/OCA/server-tools/pull/3100

- Due to `attrs` attribute is not use anymore in odoo v17 and later. => change to use `context` to make test-case suitable
- Add test-case for `operation="text_add"` and `operation="domain_add"`


hbrunn (changes requested):
migration itself is fine, but the test code you removed tests the valuable case where the dictionary doesn't exist in the base view.

And given there was barely anything to do for this migration, I think it's fair to ask for test cases for `operation="text_add"` and/or `operation="domain_add"`

LucasTran380381:
> migration itself is fine, but the test code you removed tests the valuable case where the dictionary doesn't exist in the base view.
> 
> And given there was barely anything to do for this migration, I think it's fair to ask for test cases for `operation="text_add"` and/or `operation="domain_add"`

Thanks for review. I have add test-case for 2 operation

hbrunn (approved):

sebalix (approved):

PR #3137, opened 2024-12-04
[17.0] [FIX] auditlog: Dismiss logging when not needed.

rven:
When fast logging is enabled on the audit log rules, empty records are created when fields are excluded from the list.
This results in a lot of empty unneeded log records, without any details.

This PR solves the issue so only a log record is created when needed.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -3,7 +3,7 @@
 
 import copy
 
-from odoo import _, api, fields, models
+from odoo import Command, _, api, fields, models
 from odoo.exceptions import UserError
 
 FIELDS_BLACKLIST = [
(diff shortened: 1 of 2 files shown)


StefanRijnhart (changes requested):
Thanks! This is literally a ticket in one of my backlogs :smiley: 

The module name is included in the PR title, but not in the commit message. Can you add it there as well? And strictly speaking, `[MOD]` is not a valid tag listed in https://www.odoo.com/documentation/16.0/contributing/development/git_guidelines.html#tag-and-module-name. I think this counts as a `[FIX]`. Please update the commit message on that point as well.

Do you think you can include a small test so that we can rule out a regression in the future?

StefanRijnhart (approved):
Excellent, thanks!

MiquelRForgeFlow (approved):
LGTM

StefanRijnhart:
@rven Please check https://github.com/OCA/server-tools/pull/3153 that adapts a test to this change

PR #3140, opened 2024-12-10
[16.0][UPD][FIX] tracking_manager: specific O2M fields notify

acsonefho:
In some case, we may have some unconventionals O2M like on `ir.attachment`. The O2M relation is done on `res_id` value. This commit take them in consideration and avoid a traceback.

**How to reproduce:**
Just add a dynamic track on attachment_ids (example on `account.move`).
Then add an attachment on the `account.move`: you should see a new message into the chatter (that works perfectly)
Finally, remove this attachment: without this PR, you should see a traceback into logs. But with this PR, you should see a new message (about the unlink) in the chatter.

image

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -49,13 +49,22 @@ class Base(models.AbstractModel):
         )
         for field_name, owner_field_name in self._tm_get_fields_to_notify():
             owner = self[field_name]
-            data[owner._name][owner.id][owner_field_name].append(
-                {
-                    "mode": mode,
-                    "record": self.display_name,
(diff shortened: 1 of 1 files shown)


sbejaoui (approved):

AnizR (approved):
Great, thanks! 

I had the same 'problem' because one of my users tried to track the field 'message_follower_ids'.

This allows tracking_manager to deal with One2Many/Many2oneReference without throwing an error.

acsonefho:
@rousseldenis Do you have super-power here? :muscle: :sunglasses:

rvalyi (approved):

PR #3142, opened 2024-12-12
[15.0][FIX] upgrade_analysis: new modules cannot be merged/renamed in same version

MiquelRForgeFlow:
**Explanation of first commit:** v14 has OCA's `10n_eu_oss` (https://github.com/OCA/account-fiscal-rule/tree/14.0/l10n_eu_oss), which is renamed in v15 to `l10n_eu_oss_oca` (https://github.com/OCA/account-fiscal-rule/tree/15.0/l10n_eu_oss_oca) due to new module Odoo's `10n_eu_oss` (https://github.com/odoo/odoo/tree/15.0/addons/l10n_eu_oss). Thus, it doesn't makes sense to have in the coverage file the row `|[new] 10n_eu_oss | |Renamed to l10n_eu_oss_oca|`.

Second commit it's obvious, we don't want OU modules appearing in the coverage file.

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -532,7 +532,16 @@ class UpgradeAnalysis(models.Model):
 
         module_domain = [
             ("state", "=", "installed"),
-            ("name", "not in", ["upgrade_analysis", "openupgrade_records"]),
+            (
+                "name",
+                "not in",
+                [
+                    "upgrade_analysis",
+                    "openupgrade_records",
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Thanks for putting the detailed explanation in the PR but can you put in in the commit message as well, and maybe a comment or extra clarification in the docstring?

MiquelRForgeFlow:
@StefanRijnhart done

StefanRijnhart (approved):
Thanks!

legalsylvain (approved):

PR #3146, opened 2024-12-12
[16.0][FIX] auditlog: patch away all write overrides in res.groups for test

hbrunn:
since https://github.com/OCA/OCB/commit/7f52166b8b3e04d7af36f127b9514ba3f557eb5c, we have another override of `res.groups#write` that messes with the cache, so I suggest to patch the class as the ORM knows it, and be sure we call `BaseModel.write`


thomaspaulb (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #3147, opened 2024-12-12
[17.0][FIX] auditlog: patch away all write overrides in res.groups for test (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/server-tools/pull/3146.


MiquelRForgeFlow:
@pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

Issue #2927, opened 2024-04-30
[16.0] Unable to install sentry-sdk dependency: urllib3 conflict


kamzata:
sentry-sdk installation is not possible on Odoo 16 due to urllib3 conflict.

```
venv16 ~ pip install sentry-sdk

Collecting sentry-sdk
  Using cached sentry_sdk-2.0.1-py2.py3-none-any.whl (266 kB)
Requirement already satisfied: certifi in ./venv16/lib/python3.10/site-packages (from sentry-sdk) (2023.11.17)
Collecting urllib3>=1.26.11
  Using cached urllib3-2.2.1-py3-none-any.whl (121 kB)
Installing collected packages: urllib3, sentry-sdk
  Attempting uninstall: urllib3
    Found existing installation: urllib3 1.26.5
    Uninstalling urllib3-1.26.5:
      Successfully uninstalled urllib3-1.26.5
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
requests 2.25.1 requires urllib3<1.27,>=1.21.1, but you have urllib3 2.2.1 which is incompatible.
Successfully installed sentry-sdk-2.0.1 urllib3-2.2.1
```

## Module
Sentry

## Bug Description
sentry-sdk requires a minor urllib3 version

## To Reproduce
**Affected versions**: 

- [ ] 16.0

1. pip install sentry-sdk

**Expected behavior**
sentry-sdk working

**Additional context**
Odoo 16.0 on Ubuntu 22.04 using Python 3.10.12

angelvilaplana:
Hi @kamzata, this error occurs because Odoo requires urllib3 == 1.26.5 and sentry-sdk > 1.9.0 required urllib3 >= 1.26.11 (https://github.com/OCA/server-tools/pull/2649). If you go to requirements.txt (https://github.com/OCA/server-tools/blob/16.0/requirements.txt)  you can see that you must install sentry_sdk equal to or lower than 1.9.0 version.  

Try to install sentry_sdk but indicating the version (pip3 install sentry_sdk==1.9.0). In theory, doing this should work.

kamzata:
Thanks, now it seems to work. Should I install all other dependencies as well ( `pip install -r server-tools/blob/16.0/requirements.txt` ) ?

[...]

PR #3154, opened 2024-12-31
[MIG] sentry: Migration to 18.0

SodexisTeam:


requirements.txt
```diff
@@ -4,3 +4,4 @@ dataclasses
 mako
 odoorpc
 openupgradelib
+sentry_sdk<=1.9.0
```
hparfr, line 7: At the moment 2.20 is the latest version.
https://pypi.org/project/sentry-sdk/

hparfr (changes requested):
Thanks for adressing this migration.

Why pin an old version from 2,5 years ago ?

requirements.txt
dnplkndll: https://github.com/OCA/server-tools/issues/2927
looks like on 18.0 can use 3.12 to bypass
https://github.com/odoo/odoo/blob/18.0/requirements.txt#L80

https://github.com/OCA/server-tools/pull/3174/commits/a1969c2d2cc33221882a07f56b1ecfecc3ae531f
started the testing

StefanRijnhart:
There is another migration attempt that uses a more recent version of sentry-sdk: https://github.com/OCA/server-tools/pull/3174

StefanRijnhart:
Closing as the other version is already merged.

PR #3157, opened 2025-01-06
[16.0] [ADD] compute_field_after_install

paradoxxxzero:
Following #2774 this is `compute_field_after_install` migrated to 16.0 with added fixes:

  - Computation grouping for multiple field with same compute function (i.e. `sale.order._compute_amounts`)
  - Compute last ids first to minimize data inconsistency
  - Global `computed_fields_batch_size` config 
  - Per model `computed_fields_batch_size__<model_name>` config
  - Per field `computed_fields_batch_size__<model_name>__<field_name>` config
  - Skip of precompute fields
  - Improved readmes and unit tests


compute_field_after_install/models/recompute_field.py
```diff
@@ -0,0 +1,137 @@
...
+                )
+                if not records:
+                    tasks.state = "done"
+                    self.env.cr.commit()  # pylint: disable=E8102
+                    break
+                for field in fields:
+                    field_ = records._fields[field]
+                    self.env.add_to_compute(field_, records)
+
+                records.recompute()
```
Kev-Roche: Functionnaly working but with a depreciation warning on recompute():
`compute_field_after_install/models/recompute_field.py:101: DeprecationWarning: Deprecated method recompute(), use flush_model(), flush_recordset() or env.flush_all() instead`
```suggestion
                self.env.flush_recordset(records)
```

legalsylvain (changes requested):
There was a lot of questions in the initial PR regarding the design. (That has never been merged.)
Could you answer here before migrating the module ?
https://github.com/OCA/server-tools/pull/2774
Thanks !

PR #3160, opened 2025-01-09
[16.0][FIX] fetchmail_attach_from_folder: be able to open form view even when connection fails

thomaspaulb:
Background: on Outlook servers and possibly on other providers as well, connect can fail with a UserError. Before this fix this results in the form view not being able to open because the compute cannot complete. This fixes that, so that the connection error may be resolved by the user.


thomaspaulb:
@NL66278

NL66278 (approved):
:+1: LGTM

PR #3059, opened 2024-10-09
[18.0][MIG] base_cron_exclusion: Migration to 18.0

xaviedoanhduy:

PR #3167, opened 2025-01-17
[18.0][FIX] base_cron_exclusion: remove numbercall condition from ir_cron select clause

xaviedoanhduy:
- `numbercall` and `doall` have been removed in commit https://github.com/odoo/odoo/pull/151628/commits/3a9949ad, but in the process of moving to version 18.0 in this PR I missed this change. https://github.com/OCA/server-tools/pull/3059
- also take this opportunity to replace `_` with `self.env._` to improve the old PR above

### DIFF
--- a/base_cron_exclusion/models/ir_cron.py
+++ b/base_cron_exclusion/models/ir_cron.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import _, api, fields, models
+from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (commented):
numbercall part is correct, thank for fixing. For the other part, could you clarify why is it better `self.env._` in this case?

Thanks!

xaviedoanhduy:
> numbercall part is correct, thank for fixing. For the other part, could you clarify why is it better `self.env._` in this case?
> 
> Thanks!

hi @LoisRForgeFlow, there is a quote in the Migration to 18.0 document  (see: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0)

> You can replace the call to translate a string (_) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

LoisRForgeFlow (approved):
Thanks for the pointer :ok_hand:

PR #3170, opened 2025-01-20
[18.0][IMP] upgrade_analysis: don't depend on mako

hbrunn:
given Odoo doesn't depend on mako any more and we have a perfectly fine templating engine built in, I propose to use that instead. The [coverage file](https://github.com/OCA/OpenUpgrade/pull/4709/files#diff-132041309318d21efdc12b1ef17729252c28b0ec8a6a66027f06c3c76d3e0cde) of the v18 migration has been created with this

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 # generated from manifests external_dependencies
 astor
 dataclasses
-mako
 odoorpc
 openupgradelib
--- a/upgrade_analysis/__manifest__.py
+++ b/upgrade_analysis/__manifest__.py
@@ -22,7 +22,7 @@
     "installable": True,
     "depends": ["base"],
     "external_dependencies": {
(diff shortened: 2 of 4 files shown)


legalsylvain (approved):
Nice simplification. Thanks.

I guess you should remove mako here also, dont you think ?

https://github.com/hbrunn/server-tools/blob/18.0-upgrade_analysis-remove-mako/upgrade_analysis/__manifest__.py#L25

hbrunn:
thanks for the headsup! I'm juggling multiple changesets on upgrade_analysis here and missed that file

StefanRijnhart (approved):
Thanks for making me read up on special variables such as `*_value` when using `t-foreach`!

StefanRijnhart:
Merging as both module maintainers agree

/ocabot merge patch

PR #3172, opened 2025-01-20
[18.0][ADD] upgrade_analysis: be more helpful for v18 property conversion

hbrunn:
given the global change of how properties work in v18, it's helpful to see what to do in the analysis file already. The change looks like [that](https://github.com/OCA/OpenUpgrade/pull/4709/commits/5409ce0867c7821a9231c85f798d5d162060ca42) in the analysis file


legalsylvain (approved):
LGTM. Thanks !

StefanRijnhart (approved):

PR #3173, opened 2025-01-21
jsonifier: fix non existing field handling

simahawk:
Non existing fields should be bypassed when in non-strict mode.


simahawk:
Bah, I had a bad remote w/ outdated changes :sob:

PR #3176, opened 2025-01-24
[17.0][IMP] tracking_manager: add domain condition for tracking fields

CRogos:
I would like to extend the functionality of tracking_manager by adding the possibility to enable tracking only on certain condition, by adding a tracking_domain field. The value is only tracked if the filter applies.

image

image

image


fb-ife:
@CRogos ,
can you please provide me with configuration'steps to test the functionality?

Thanks

CRogos:
@fb-ife have you had a look into the screenshots above and the usage documentation file? I've already highlighted the configuration steps. Not sure what else needs to be added?

CRogos:
@sebastienbeau, @Kev-Roche I would really like to add this improvement to this module. Could you have a look at this?
thank you in advance.

MohamedOsman7 (approved):
Tested with the runboat. Functional & code review LGTM

CRogos:
@pedrobaeza what do you think about this change? I would be very pleased if you could do the second review and merge if it is OK.

tracking_manager/static/description/fields.drawio.png
pedrobaeza, line 1: Why changing the file name and putting `drawio` ?

tracking_manager/models/models.py
```diff
@@ -21,6 +22,43 @@ class Base(models.AbstractModel):
     def is_tracked_by_o2m(self):
         return self._name in self.env["ir.model"]._get_model_tracked_by_o2m()
 
+    def _mail_track(self, tracked_fields, initial_values):
+        changes, tracking_value_ids = super()._mail_track(
+            tracked_fields, initial_values
+        )
+
```
pedrobaeza: Don't add empty line inside a method, as it makes difficult the reading to see the continuity and other syntax highlighting like the comments already act as separators.

pedrobaeza (commented):
The problem I see with this change is that affects performance even if you don't define any domain, so I'm not totally liking. You may do this in a extra module?

tracking_manager/static/description/fields.drawio.png
CRogos: This picture is an valid png file, but also an valid draw.io file. 
When you have the draw.io extension installed in vscode, you can edit the rectangles of the picture. That means you always have the unchanged screenshot and can add or remove additional highlighting easily. 

image

image

pedrobaeza: Oh, didn't know. Thanks for telling me.

tracking_manager/models/models.py
CRogos: I have an idea how to improve the performance part. If the improvement is not sufficient, I'll migrate it to a new module, but I am not so a fan of to many micro modules, because of the maintenance work and also from the user perspective finding the right modules.

CRogos: I've removed the empty lines. I also improve the performance impact of this change a little. When there are fields tracked, than an additional search query is executed to check whether a domain is set on the tracked field. If there is not domain set, the rest of the code is skipped.

Do you still think this one query L36-L42 is to much overhead to keep it in this module?

pedrobaeza (commented):
Uhm, I'm not sure that this code fixes the performance thing and even if `_mail_track` is the correct hook. I still prefer to have that as a separate module, but let's see what others think.

CRogos:
@acsonefho @victoralmau what do you think about this change?

acsonefho:
> @acsonefho @victoralmau what do you think about this change?

I have no real opinion on the matter. It could be a specific module or a new feature of this one, both are ok for me.
I also ask to the analyst who already use this module and same opinion as mine.

Thanks for the work and your contribution :muscle:

CRogos:
I did a quick performance check by calling _mail_track 100 times, and there is a significant performance decrease as @pedrobaeza assumed.

image

After inserting a cache to the query loading the tracking_domain value, this performance decrease is improved, but there is still a performance decrease by ~25% compared with before this change.
On the left graph, you can see that the change decreases the performance by ~100% in case the field has a tracking_domain.
image

I am not sure how significant this performance lost is. When we take into account the total save form process, I assume the 25% and also the 100% are not that relevant, but did not traced it with the performance tool.

pedrobaeza:
Thanks for doing the check. I think this enforces the argument of adding this feature in a separate module for those wanting it to assume that decrease.

CRogos:
I think the failing tests is caused/solved by this: https://github.com/OCA/server-tools/pull/3237
I also moved the change into an separate module if this is preferred my the majority: https://github.com/OCA/server-tools/pull/3236

CRogos:
Moved to a separate module: https://github.com/OCA/server-tools/pull/3236

PR #3177, opened 2025-01-26
[16.0][imp] auditlog file only

amon-ra:
I am having a lot of problems when auditing is enabled in several third party modules. This little change skips all writes and save logs to odoo logging.

It is used in production. Hope it helps


StefanRijnhart (changes requested):
* What issues does this solve?
* Might be interesting to have a Logfile/database option but it should be a separate option, not instead of full/fast.
* Please check precommit
* Please add a test

PR #3178, opened 2025-01-28
[FIX] sentry: add compatibility with sentry-sdk from debian bullseye repository

em230418:
Debian Bullseye is used as base docker image for odoo:15.0

PR #3179, opened 2025-01-30
[FW][FIX] sentry: add compatibility with sentry-sdk from debian bullseye repository

em230418:
Debian Bullseye is used as base docker image for odoo:16.0

Forward-port of https://github.com/OCA/server-tools/pull/3178


moylop260:
Thank you!

PR #3181, opened 2025-02-01
[18.0][FIX] sentry: add compatibility with sentry-sdk from debian bullseye …

dnplkndll:
…repository

Debian Bullseye is used as base docker image


moylop260:
It is a FW from https://github.com/OCA/server-tools/pull/3179

So merging fast-tracking

PR #3188, opened 2025-02-04
[18.0][PORT] 3185 from 16.0

sbidoul:
Port #3185 from 16.0 to 18.0.


StefanRijnhart (approved):
Test failure is unrelated and fixed in https://github.com/OCA/server-tools/pull/3190

sbidoul:
Thanks for the fix @StefanRijnhart

PR #3192, opened 2025-02-07
18.0 mig scheduler error mailer

youelEl:


StefanRijnhart:
Duplicate of https://github.com/OCA/server-tools/pull/3084

StefanRijnhart:
Looks like this is equivalent to https://github.com/OCA/server-tools/pull/3084 and we can continue there.

PR #3215, opened 2025-03-04
[16.0][IMP] upgrade_analysis: avoid private methods

MiquelRForgeFlow:
Avoids issue with accessing private methods (due to https://github.com/odoo/odoo/commit/a1adf8355ceb00e6b9b5cbdde00aa87c5501e922).

I am not sure if I did correctly the monkey patch, it's not tested.

### DIFF
--- a/upgrade_analysis/odoo_patch/odoo/__init__.py
+++ b/upgrade_analysis/odoo_patch/odoo/__init__.py
@@ -1,4 +1,5 @@
 from . import addons
 from . import models
 from . import modules
+from . import service
 from . import tools
--- a/upgrade_analysis/odoo_patch/odoo/service/__init__.py
+++ b/upgrade_analysis/odoo_patch/odoo/service/__init__.py
@@ -1,0 +1,1 @@
+from . import model
--- a/upgrade_analysis/odoo_patch/odoo/service/model.py
+++ b/upgrade_analysis/odoo_patch/odoo/service/model.py
@@ -1,0 +1,34 @@
+from odoo.models import BaseModel
+
+
+def _new_get_public_method(model, name):
+    """Get the public unbound method from a model.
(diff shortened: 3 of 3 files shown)


StefanRijnhart (commented):
For my (limited) understanding, do you have a concrete example where a migration triggers the exception?

MiquelRForgeFlow:
This month, when I did a new analysis for v17, it broke when trying to access the `mapped` method, which is decorated with @api.private. This method is called in https://github.com/OCA/server-tools/blob/16.0/upgrade_analysis/models/upgrade_analysis.py#L554.

StefanRijnhart:
Thanks for clarifying. Can't believe they private'd `mapped`! It is literally mentioned as a cool trick in OdooRPC's documentation. Can you add this example to the documentation string of the monkey patch?

StefanRijnhart (changes requested):
This is not how the odoo patch mechanism of this module works. The mechanism works with a derived class: https://github.com/OCA/server-tools/blob/16.0/upgrade_analysis/odoo_patch/odoo/modules/registry.py.

But a patch using the mechanism would not work, because it is only active on the calling side, and the mechanism is not applied when doing remote communication.

Maybe we should ask Odoo to reconsider privating `mapped`? Is it just this one method or are other calls affected as well?

StefanRijnhart:
FYI @sebalix OdooRPC is impacted by this new way of 'privatization' in Odoo.

legalsylvain:
Hi @MiquelRForgeFlow. thanks for your investigation.

At a first sight, I'm not in favor to add core patch, if it's not necessary. Here, If I'm right, https://github.com/OCA/server-tools/pull/3218 can solve the problem you mention.
What do you think ? 

thanks !

MiquelRForgeFlow:
Ok, let's replace this by https://github.com/OCA/server-tools/pull/3218.

PR #3216, opened 2025-03-04
[17.0][FIX] upgrade_analysis: fill upgrade_path

MiquelRForgeFlow:
For some reason, _compute_upgrade_path is not called by default if store=True.

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -88,10 +88,12 @@ class UpgradeAnalysis(models.Model):
         module = self.env["ir.module.module"].search([("name", "=", module_name)])[0]
         if module.is_odoo_module:
             if not self.upgrade_path:
-                return (
-                    "ERROR: no upgrade_path set when writing analysis of %s\n"
-                    % module_name
-                )
+                self._compute_upgrade_path()
+                if not self.upgrade_path:
+                    return (
+                        "ERROR: no upgrade_path set when writing analysis of %s\n"
+                        % module_name
+                    )
             full_path = os.path.join(self.upgrade_path, module_name, version)
         else:
             full_path = os.path.join(


upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -51,7 +51,7 @@ class UpgradeAnalysis(models.Model):
     upgrade_path = fields.Char(
         compute="_compute_upgrade_path",
         readonly=False,
-        store=True,
+        store=False,
```
StefanRijnhart: There is a reason for this: https://github.com/OCA/server-tools/commit/9289acc48d5facb48dbc4a3d919ac64f6e7fcc8b

Maybe you can call the compute method before the value is used if it is empty (it is only in one place)?

legalsylvain:
hi @MiquelRForgeFlow.  
Naive question: is it a new problem ? This code looks to work since V14 without problem. do you know what changed ?

MiquelRForgeFlow:
I don't know @legalsylvain . When I do analysis in v16, I don't have the issue, but when I do it in v17, I needto put store = False to make it work. I don't know why, if it due to some core change in odoo or something else.

legalsylvain (approved):
OK. Then. 
thanks for your investigation.
(this change doesn't hurt anyway).

MiquelRForgeFlow:
@hbrunn did you face this issue in v18 analysis?

StefanRijnhart:
@legalsylvain I introduced this change in 17.0, because I ran into 
`'openupgrade_scripts/scripts/' is not a valid upgrade directory`
If I tried to rely on `--upgrade-path` when there are no analysis files present yet.

StefanRijnhart (changes requested):

upgrade_analysis/models/upgrade_analysis.py
hbrunn: that's what I do in the analysis generating github action: https://github.com/hbrunn/OpenUpgrade/blob/18.0/.github/workflows/generate-analysis.yml#L175

MiquelRForgeFlow: > call the compute method before the value is used if it is empty

Done

hbrunn (approved):

StefanRijnhart (approved):
Thanks for the update!

Issue #3139, opened 2024-12-07
[16.0] auto_backup: TypeError: makedirs() got an unexpected keyword argument 'exist_ok'


santiky:
## Module
auto_backup 16.0

## Describe the bug
This bug is the same than pull request #2902 in 12.0 branch but recent changes 3d39f2b8b47 has reintroduced it in 16.0 branch.

Trying to make a backup in a remote sftp server, throws this error:
>  Odoo Server 16.0:odoo:ERROR:odoo.addons.auto_backup.models.db_backup:Database backup failed: sftp://user@example.com:22
                                                            Traceback (most recent call last):
                                                              File "/opt/odoo/addons-oca/auto_backup/models/db_backup.py", line 215, in backup_log
                                                                yield
                                                              File "/opt/odoo/addons-oca/auto_backup/models/db_backup.py", line 189, in action_backup
                                                                remote.makedirs(rec.folder, exist_ok=True)
                                                            TypeError: Connection.makedirs() got an unexpected keyword argument 'exist_ok'

>
and it does not complete the backup transfer.

## To Reproduce
**Affected versions**:  Only 16.0,

1. Setup a backup in a remote server.
2. Make one backup, it fails.

**Additional context**
Pysftp method makedirs() it does not have any argument, it's different than the os.makedirs method.
https://pysftp.readthedocs.io/en/release_0.2.9/cookbook.html#pysftp-connection-makedirs
Removing "exist_ok" on line 189 solves the error.
https://github.com/OCA/server-tools/blob/6e7e03d85cc58ccadda083c3ea14d04b24092ca2/auto_backup/models/db_backup.py#L189

Regards.

santiky:
Searching in old pull requests there is one with this fix, can it be reapplied? #2939

arielbarreiros96:
Same problem here, #3217 solves the issue

PR #3217, opened 2025-03-04
[16.0][FIX] Database autobackup bugfix: exist_ok parameter does not exists in remote.makedirs() function

ignaciord:
As said on https://github.com/OCA/server-tools/issues/3139 there is no exist_ok parameter in remote.makedirs() function.

So, with this PR the SFTP remote backup will work with no errors.

Thanks,

Ignacio


StefanRijnhart (approved):

arielbarreiros96 (approved):
Ran into the same problem, LGTM

PR #3218, opened 2025-03-04
[16.0][FIX] upgrade_analysis: avoid call privatized 'mapped' function.

legalsylvain:
FIX: bug described in #3215.
(alternative to patch solution)
 
Rational : 'mapped' function has been privatized in https://github.com/odoo/odoo/commit/a1adf8355ceb00e6b9b5cbdde00aa87c5501e922. We so replace 'mapped' by 'search_read'


StefanRijnhart (approved):
Thanks!

MiquelRForgeFlow:
Nice, but we should test first if this approach works.

pedrobaeza (approved):
/ocabot merge patch

rvalyi (approved):

PR #3222, opened 2025-03-06
[16.0] Update fetchmail_server_folder

ikus060:
- Update fields description for clarity
- Support fetch_last_day_only  to fetch only 24 hours of emails
- Process emails in reverse order
- Use BODY.PEEK to avoid marking email as seen
- Support email field 'reply_to' to matching
- Improve logging
- Relayout view to group related fields


fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -33,73 +34,93 @@ class FetchmailServerFolder(models.Model):
...
-    domain = fields.Char(help="Fill in a search filter to narrow down objects to match")
     msg_state = fields.Selection(
         selection=[("sent", "Sent"), ("received", "Received")],
         string="Message state",
         default="received",
         help="The state messages fetched from this folder should be assigned in Odoo",
     )
-    active = fields.Boolean(default=True)
+    active = fields.Boolean(
+        string="Active",
```
NL66278: Odoo will provide this capitalized name automatically, therefore should not be specified according to OCA guidelines.

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -173,15 +194,22 @@ def check_imap_archive_folder(self, connection):
                     _("Could not create archive folder %(folder)s on server %(server)s")
                     % {"folder": self.archive_path, "server": server.name}
                 )
+        
 
     def get_criteria(self):
-        return "UNDELETED" if not self.fetch_unseen_only else "UNSEEN UNDELETED"
+        criteria = "UNDELETED"
```
NL66278: Why replace one clear line, with three in the end doing the same?

If you want to improve it, take away the not operator:
``` return "UNSEEN UNDELETED" if self.fetch_unseen_only else "UNDELETED"```

or, as you might still add the 24 hour crtierium:
``` criteria =  "UNSEEN UNDELETED" if self.fetch_unseen_only else "UNDELETED"```

**Review by NL66278 — CHANGES_REQUESTED** [review]
Thank you for your contribution! To proceed a few things are needed. The first one is to locally use pre-commit before you push your changes. Use a pip virtual environment and pip install pre-commit. After activation of the virtual environment run pre-commit install in your repo directory. This will make your changes pass the pre-commit tests what is really required for a merge.

Another thing is to change strings on fields and help sparingly, as this will necessitate also new translations. Having said that your texts are in many cases a real improvement.

If you could add some unit test for the new functions, that would be nice, preventing a decrease in coverage.

File: fetchmail_attach_from_folder/models/fetchmail_server_folder.py (hunk 2 above)
Comment (ikus060) [reply fetchmail_attach_from_folder/models/fetchmail_server_folder.py]: > one clear line

Let’s say clarity is subjective and differs from one individual to another.

Will make the changes

File: fetchmail_attach_from_folder/models/fetchmail_server_folder.py (hunk 1 above)
Comment (ikus060) [reply fetchmail_attach_from_folder/models/fetchmail_server_folder.py]: Thanks I did not know

**Comment by ikus060** [comment]
I've make the requested changes in addition to running pre-commit hook.

**Comment by NL66278** [comment]
@ikus060 Thank you for that. However pre-commit job is still complaining. Coud you run ```pre-commit run --all-files --show-diff-on-failure --color=always```.

And if possible squash your changes into a single commit.

File: fetchmail_attach_from_folder/views/fetchmail_server.xml [hunk 3]
```diff
@@ -61,50 +61,67 @@
...
+                                        />
+                                        <field
+                                            name="domain"
+                                            placeholder="[('state', '=', 'open')]"
+                                        />
+                                        <field name="match_first" />
+                                        <field
+                                            name="model_order"
+                                            placeholder="name asc"
+                                            attrs="{'readonly': [('match_first','==',False)],                                                     'required': [('match_first','==',True)]}"
```
NL66278: Not sure what happened here, but the string starting with `'required'`could go on the next line, under attrs (or under readonly).

fetchmail_attach_from_folder/views/fetchmail_server.xml
```diff
@@ -61,50 +61,67 @@
...
-                                            'required':
-                                            [('match_first','==',True)]}"
-                                    />
-                                    <field name="flag_nonmatching" />
+
+                                <!-- Matching Options -->
+                                <group>
+                                    <separator string="Matching Options" />
+                                    <group colspan="2">
+                                      <field
```
NL66278: The field element should be indented 4 spaces under group, not sure why pre-commit prettify did not pick this up.

NL66278:
@ikus060 Just some small remarks on the xml formatting and we should be good to go...

NL66278:
@ikus060 You might also have a look at my PR to use message uid instead of number in folder. I think the same problem you solve with handling the messages in reverse order. https://github.com/OCA/server-tools/pull/3017

ikus060:
Here another tentative. Sorry for all the noise. I'm not used to OCA code policies.

See my comments for #3017

NL66278:
@ikus060 No problem. They are here: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

NL66278 (approved):
:+1: LGTM

ikus060:
Is there something blocking the merge ?

NL66278:
@ikus060 Nothing really blocking, but waiting for more reviews. @hbrunn Maybe you want to have a look?

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -175,13 +193,17 @@ def check_imap_archive_folder(self, connection):
                 )
 
     def get_criteria(self):
-        return "UNDELETED" if not self.fetch_unseen_only else "UNSEEN UNDELETED"
+        criteria = "UNSEEN UNDELETED" if self.fetch_unseen_only else "UNDELETED"
+        if self.fetch_last_day_only:
+            yesterday = (datetime.utcnow() - timedelta(days=1)).strftime("%d-%b-%Y")
```
zamberjo: I think it is preferable to use the tools already included in Odoo.
```suggestion
            yesterday = fields.Date.subtract(fields.Date.context_today(self), days=1).strftime("%d-%b-%Y")
```

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -175,13 +193,17 @@ def check_imap_archive_folder(self, connection):
...
+            criteria = f"SINCE {yesterday} " + criteria
```
zamberjo: A little more legible
```suggestion
            criteria = f"SINCE {yesterday} {criteria}"
```

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
```diff
@@ -266,7 +288,7 @@ def run_server_action(self, matched_object_ids):
     def fetch_msg(self, connection, msgid):
         """Select a single message from a folder."""
         self.ensure_one()
-        result, msgdata = connection.fetch(msgid, "(RFC822)")
+        result, msgdata = connection.fetch(msgid, "(BODY.PEEK[])")
```
zamberjo: Do you think this can be configurable?

zamberjo (commented):
Thank you for your pr, I would like to point out a few points I have observed.

fetchmail_attach_from_folder/models/fetchmail_server_folder.py
ikus060: Hum, good point. I would recommand to do something similar to flag_nonmatching.

I could introduce a new fields to mark the email as seen if matching. and another fields to mark the email as seed if not matching ? 

The idea is to let the admin decide which email should be mark as seen.

Maybe you have a different proposal ?

To make it symetric with flag_nonmatching, I could also add a field flag_matching.

NL66278: @ikus060 @zamberjo Would definitely be nice to have this configurable! I do not quite see what the purpose would be of a field flag_matching. The idea of flagging is to make clear to the user that some kind of manual action might be needed, as we do not know what to do with the email in Odoo. As for configuring whether to use BODY.PEEK or RFC822 I would prefer a selection field. (And then preferably have RFC822 as the default, as that is the current way it works and on which organizations might rely). The idea would by default be that users only occasionally look at the mailbox, and then can assume UNSEEN mails have not been fetched by Odoo. On the other hand, if in some organizations somebody must check the mailbox regularly and actually read the emails, BODY.PEEK would be appropriate to make clear to the user which mails this user still needs to deal with.

zamberjo: I agree with @NL66278, imho the best thing to do is to create a selection field leaving RFC822 as default.

ikus060: Since the only difference between BODY.PEEK and RFC822 is that the latter marks the e-mail as read, it would be preferable to offer a more flexible configuration to support other usage scenarios, including :

- Marking all e-mails as read (default): This is the previous behaviour of this addons. Any email consume by the addons are marked as Seen.
- Mark matching e-mails as read : When email is not matchhing anything, then the email is left Unseed.
- Do not mark e-mails as read : This is the behaviour I'm looking for. My use case is to have Odoo consume all email from inbox and attach matching email to the right ressources for tracking and archiving. Users continue to use their mailbox as usual.

If you approve this direction. Do you still recommand a selection field ?

NL66278: @ikus060 Having three options I think actually strengthens the case for a selection field.

ikus060:
Hello All,

I took some time to update this Pull Request with the changes recommended.

NL66278:
@zamberjo I think this is ready for merge now. Can you also have another look?

zamberjo (approved):
LGTM

Maybe in the future someone will be interested in configuring that period of time instead of leaving it fixed at 24 hours, but let's leave it like that for the moment.

NL66278:
For some reason merge commits fails on tests that ran successfully when submitting the PR. On modules that have nothing to do with this PR.

hbrunn:
@NL66278 that's normal behavior, tests are run on the head branch with the PR merged in. And this fails currently because of changes in Odoo, fixed in https://github.com/OCA/server-tools/pull/3271

@ikus060 next time you create a PR please give it a name that's more descriptive than "Update ..."

NL66278:
Conflicts can be resolved by merging this PR: https://github.com/ikus060/server-tools/pull/1

thomaspaulb:
@ikus060 would like to merge but can you rebase and solve conflicts, with help of @NL66278 pr

ikus060:
@thomaspaulb It should be merged.

thomaspaulb:
@ikus060 There are still conflicts: 

image

Also, the commits should be squashed after the conflicts are solved.

ikus060:
@thomaspaulb Thanks for getting back to me on this PR. 

I appreciate that we're moving forward with the merge. But, I want to be transparent about where I'm at: it's been almost a year since I submitted this, and in that time my focus has shifted to other projects. I also no longer have the development environment set up to properly test these changes.

I completely understand that everyone involved is volunteering their time, and I'm grateful for that. At the same time, I think this situation highlights how long review cycles can make it challenging for contributors to stay engaged - especially when we need to context-switch back after many months.

I now resolve the conflict using Github tools. Let hope it's working as expected.

If not, I will need more time to recreate all the setup environment.

thomaspaulb:
@ikus060 I understand where you're coming from. OCA currently has this problem of long review cycles. We're trying at the Board level to see how to fix this, but it's not easy.

On the other hand, you don't need a development environment to fix this. You just check out the branch on Git, run some smart git commands, and re-push. If not, this will go stale and some other developer will have to open a new PR and fix it for you.

-

ikus060:
@thomaspaulb I believe I did everything needed. I don't see any conflict anymore.

thomaspaulb:
I still see it, but maybe it's my Github.

image

ikus060:
@thomaspaulb if I understand it well, this to check verify the unit test code coverage. Am I right ?

It seems I need to write new units tests. Am I really required to write those test ? I would prefer not to, as it's not a priority for me.

thomaspaulb:
@ikus060 Where do you get the idea that you have to write new unit tests in order to solve a Git merge conflict?

The only thing that's needed, basically, is something like

```
git clone git@github.com:OCA/server-tools@patrik-fetchmail-server-folder server-tools
cd server-tools
git fetch origin 16.0
git remote add ikus git@github.com:ikus060/server-tools
git fetch ikus patrik-fetchmail-server-folder
git checkout -b  patrik-fetchmail-server-folder ikus/patrik-fetchmail-server-folder
git rebase -i origin/16.0
# leave the commits the should be there
# solve merge conflicts as they come
git add .
pre-commit run  # sanity check
# then when happy squash commits into one
git push -f
```

It should be, at max, a 15 min job, if not even that, and not worth the discussion IMO

thomaspaulb:
Ah, I think you looked at the wrong part of the screenshot. It is not required to make the coverage green. The important thing to fix is the merge conflicts, which are visible at the bottom of the screenshot as **This branch cannot be rebased due to conflicts**.

NL66278:
@ikus060 @thomaspaulb There will be more conflicts after merging my backports of Odoo 18.0 improvements. I made a branch with the improvements offered by @ikus060 based on the backport branch, so I propose first looking at the backport branch and if OK merging, then merging in the improvement branch, which basically is already approved in the reviews here.

The branch containing both backport and improvements: https://github.com/Therp/server-tools/tree/patrik-fetchmail-server-folder

NL66278:
@ikus060 Would you have a quick look / review of this branch? If I merge it in, then I can proceed to merge your changes.

ikus060:
Closing this PR. Reopen for 18.0 branch #3510

PR #3223, opened 2025-03-10
[16.0] [FIX] sentry: Fix wsgi integration following HTTPocalypse

paradoxxxzero:
The current sentry wsgi integration tries to wrap the `odoo.http.Application` which is the application class and not the wsgi_app object. We should wrap directly the `odoo.http.root` object, with an attribute proxy to forward `session_store` and similar attribute to the original application.

### DIFF
--- a/sentry/hooks.py
+++ b/sentry/hooks.py
@@ -6,7 +6,6 @@ import warnings
 from collections import abc
 
 import odoo.http
-from odoo.service.server import server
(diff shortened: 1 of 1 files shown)


moylop260:
Could you check the sentry logger error, please?

`2025-03-10 11:48:03,039 962 ERROR odoo odoo.addons.sentry.tests.test_client: Test exception`

PR #3231, opened 2025-03-19
[FIX] sentry: temporarily exclude sentry_sdk>=2.23.1

nilshamerlinck:
To avoid following error:
AttributeError: 'InMemoryTransport' object has no attribute 'options'

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ dataclasses
 odoorpc
 openupgradelib
 pysftp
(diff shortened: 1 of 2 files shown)


dnplkndll (approved):
I propose we merge this as is to keep same functionality. then can review the newer version.

kobros-tech (approved):

rvalyi (approved):
LGTM, let's unblock the CI

Issue #2966, opened 2024-06-12
Incompatible chardet dependency between this repo and Odoo


gpothier:
In this repo there is a dependency on `pygount`:
https://github.com/OCA/server-tools/blob/0169df8a9800637d06189e65686ff2b9634a9d0c/requirements.txt#L10
The current version of pygount (1.8.0) requires `chardet = "^5"` (https://github.com/roskakori/pygount/blob/7113dd74687732e4c5d7195050caea8e83b67c9f/pyproject.toml#L68). However, Odoo itself requires `chardet==4.0.0` (https://github.com/odoo/odoo/blob/09ada135847b96fa58d862568603dde61fcf1488/requirements.txt#L4)

The last version of `pygount` to support `chardet = "^4"` is 1.4.0 (https://github.com/roskakori/pygount/blob/49ca668c55d5383c2e512913f3264f1d954bfc7a/pyproject.toml#L68)

If one installs the dependencies from both repositories `requirements.txt` files, Odoo fails to start:
```
pkg_resources.ContextualVersionConflict: (chardet 5.2.0 (/usr/local/lib/python3.9/site-packages), Requirement.parse('chardet<5,>=3.0.2'), {'requests'})
```

This can be solved by specifying the compatible version in this repo:
```
pygount==1.4.0
```

pedrobaeza:
You can add a PR pinning such dependency in the manifest of the affected module.

PR #2967, opened 2024-06-12
[16.0] module_analysis: pin pygount dependency

gpothier:
Fixes https://github.com/OCA/server-tools/issues/2966

PR #3232, opened 2025-03-20
[17.0][FIX] module_analysis: pin pygount

rvalyi:
Same fix as applied to 16.0 yesterday: #2967 as the same bug happens in the 17.0 branch. Here for instance:
https://github.com/OCA/server-tools/actions/runs/13928897997/job/38980680565#step:8:335

Pin the pygount dependency to 1.4.0 to avoid chardet version clash

Fixes https://github.com/OCA/server-tools/issues/2966

cc @legalsylvain @StefanRijnhart @pedrobaeza

### DIFF
--- a/module_analysis/__manifest__.py
+++ b/module_analysis/__manifest__.py
@@ -27,7 +27,7 @@
         "data/ir_cron.xml",
     ],
     "external_dependencies": {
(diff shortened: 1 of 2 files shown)


rvalyi:
replaced with #3237

PR #3240, opened 2025-03-26
[18.0][IMP] database_cleanup: add table purge blacklist with known non-model table

StefanRijnhart:
This is the culprit: https://github.com/OCA/web-api/blob/18.0/endpoint_route_handler/registry.py#L73

### DIFF
--- a/database_cleanup/models/purge_tables.py
+++ b/database_cleanup/models/purge_tables.py
@@ -96,12 +96,16 @@ class CleanupPurgeWizardTable(models.TransientModel):
     _name = "cleanup.purge.wizard.table"
     _description = "Purge tables"
 
+    blacklist = [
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

StefanRijnhart:
@simahawk as the maintainer of the affected module, this may be of interest to you. Without this PR, database_cleanup will happily propose to remove table `endpoint_route`.

StefanRijnhart:
@pedrobaeza @legalsylvain apologies for the mention, but without this change database_cleanup will let Odoo crash immediately if all suggested tables are dropped on databases with `endpoint_route_handler` installed, in a way that is not even recoverable by updating this module.

pedrobaeza (approved):
/ocabot merge patch

PR #3243, opened 2025-03-28
[17.0][FIX] upgrade_analysis: avoid call privatized 'mapped' function. (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/server-tools/pull/3218.

PR #3244, opened 2025-03-28
[18.0][FIX] upgrade_analysis: avoid call privatized 'mapped' function. (forward port)

MiquelRForgeFlow:
Forward port of https://github.com/OCA/server-tools/pull/3243.


legalsylvain (approved):
Thanks.

/ocabot merge patch

PR #3248, opened 2025-04-03
attachment_delete_restrict: using sudo should by pass the restriction

sebastienbeau:
When having other module that call an unlink in a sudo mode the restriction are applied and it raise an error

Fix it by not applying the restriction in case of sudo

### DIFF
--- a/attachment_delete_restrict/models/ir_attachment.py
+++ b/attachment_delete_restrict/models/ir_attachment.py
@@ -64,6 +64,8 @@ class IrAttachment(models.Model):
             return self._raise_delete_attachment_error(allowed_users)
 
     def unlink(self):
+        if self.env.su:
+            return super().unlink()
         res_models = list(set(self.filtered("res_model").mapped("res_model")))
         if res_models:
             models = self.env["ir.model"].search([("model", "in", res_models)])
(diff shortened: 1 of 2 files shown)


attachment_delete_restrict/models/ir_attachment.py
```diff
@@ -64,15 +64,16 @@ def _check_custom_delete_attachment(self, model=None, allow_owner_and_admin=Fals
...
+            if res_models:
+                models = self.env["ir.model"].search([("model", "in", res_models)])
+                name2models = {m.model: m for m in models}
+                for rec in self:
+                    if rec.res_model:
+                        model = name2models[rec.res_model]
+                        if model.restrict_delete_attachment == "default":
+                            rec.sudo()._check_delete_attachment()
+                        else:
+                            rec.sudo()._check_delete_attachment(model)
```
yostashiro: Could we please reduce the diff?
```suggestion
        if self.env.su:
            return super().unlink()
        res_models = list(set(self.filtered("res_model").mapped("res_model")))
        if res_models:
            models = self.env["ir.model"].search([("model", "in", res_models)])
            name2models = {m.model: m for m in models}
            for rec in self:
                if rec.res_model:
                    model = name2models[rec.res_model]
                    if model.restrict_delete_attachment == "default":
                        rec.sudo()._check_delete_attachment()
                    else:
                        rec.sudo()._check_delete_attachment(model)
```

sebastienbeau:
@yostashiro done !

yostashiro (approved):
Thank you! LGTM.

Kev-Roche (approved):
LGTM

FranzPoize:
I'm having issues after this merge

PR #3254, opened 2025-04-04
[17.0][FIX] test.yml: prevent infinite loop of makepot commits

StefanRijnhart:
Only run makepot once. The result will be equivalent in each of those jobs because the pot files will be generated for all the addons at once.

If there are more than one test run with different sets of modules, and in one of those modules a field is added to the base model, the makepot jobs will have different outcomes and will keep on reverting each other's changes upon subsequent test runs.

image
image
image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,6 @@ jobs:
         include:
           - container: ghcr.io/oca/oca-ci/py3.10-odoo17.0:latest
             include: "database_cleanup"
-            makepot: "true"
             name: Rebel test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
             include: "database_cleanup"
@@ -49,7 +48,6 @@ jobs:
           - container: ghcr.io/oca/oca-ci/py3.10-ocb17.0:latest
             exclude: "database_cleanup"
             name: test with OCB
-            makepot: "true"
     services:
       postgres:
         image: postgres:12.0


ivs-cetmix:
Hi @StefanRijnhart , thank you for your contribution! Are you still working on this task?

StefanRijnhart:
@ivs-cetmix Does it need more work?

ivs-cetmix:
> @ivs-cetmix Does it need more work?

Not sure, I though you know this better) If you consider this done, then we can proceed with reviewing it.

StefanRijnhart:
Please do!

.github/workflows/test.yml
```diff
@@ -37,7 +37,6 @@ jobs:
         include:
           - container: ghcr.io/oca/oca-ci/py3.10-odoo17.0:latest
             include: "database_cleanup"
-            makepot: "true"
```
pedrobaeza, line 40: Removing this one, no POT will be generated for database_cleanup AFAIK.

StefanRijnhart: I don't see any filtering in this code chain

https://github.com/OCA/oca-ci/blob/master/bin/oca_export_and_commit_pot#L15
https://github.com/acsone/click-odoo-contrib/blob/master/click_odoo_contrib/makepot.py#L155
https://github.com/acsone/click-odoo-contrib/blob/master/click_odoo_contrib/manifest.py#L39-L41

so it looks to me like a single run will generate POTs for all addons in the directory, regardless of include/exclude directives. Am I missing anything?

StefanRijnhart: Ah sorry, it does say 'for installed addons' in that first link, so you are right. Will amend.

StefanRijnhart:
@pedrobaeza Improved version included in https://github.com/OCA/server-tools/pull/3345.

PR #3255, opened 2025-04-07
[18.0][IMP] upgrade_analysis: improving compare

MiquelRForgeFlow:
This improvements have been taken into consideration in https://github.com/OCA/OpenUpgrade/pull/4915 (+https://github.com/OCA/OpenUpgrade/pull/4918).

See:
Selection_4410
Selection_4411

Now includes https://github.com/OCA/OpenUpgrade/pull/5050.

### DIFF
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -37,8 +37,12 @@ def model_rename_map(model):
     return apriori.renamed_models.get(model, model)
 
 
+def model_merge_map(model):
+    return apriori.merged_models.get(model, model)
+
+
 def model_map(model):
-    return apriori.renamed_models.get(model, apriori.merged_models.get(model, model))
+    return apriori.renamed_models.get(model, model_merge_map(model))
 
 
 def inv_model_map(model):
@@ -489,6 +493,15 @@ def compare_model_sets(old_records, new_records):
                     reprs[module_map(column["module"])].append(text)
                     reprs["general"].append(
                         f"obsolete model {model} "
+                        f"[module {module_map(column['module'])}]"
+                    )
+                elif model_merge_map(model) in new_models:
+                    text = f"obsolete model {model} (merged to {model_map(model)})"
+                    if column["model_type"]:
+                        text += f" [{column['model_type']}]"
+                    reprs[module_map(column["module"])].append(text)
+                    reprs["general"].append(
+                        f"obsolete model {model} (merged to {model_map(model)}) "
                         f"[module {module_map(column['module'])}]"
                     )
                 else:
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
ping: @legalsylvain

MiquelRForgeFlow:
@pedrobaeza

upgrade_analysis/compare.py
```diff
@@ -237,7 +241,7 @@ def match(match_fields, report_fields, warn=False):
         return count
 
     matched_direct = match(
-        ["module", "mode", "model", "field"],
+        ["module", "model", "field", "type"],
```
pedrobaeza: Why removing "mode"?

pedrobaeza (commented):
I don't know enough this module, but it seems more or less OK.

upgrade_analysis/compare.py
MiquelRForgeFlow: Well, because it doesn't matter. The important thing was adding "type", and here "mode" is not necessary. I could have done:

```python
matched_direct = match(["module", "mode", "model", "field", "type"])
matched_direct += match(["module", "model", "field", "type"])
```
but it would be overkill. "mode" only checks if it's first declaration of the field in that model or not.

pedrobaeza: OK, I don't know the historical reasons to be there, and I was thinking in selection fields where mode may be "modify" or similar.

MiquelRForgeFlow: Selection_4471

BTW, I realized why was done that way, the matches. It happens that "type" is filled only when mode is "create", but not when mode is "write". I readjusted the matches once I figured this, and the analysis is fine. I will push it soon.

upgrade_analysis/compare.py
```diff
@@ -121,8 +128,10 @@ def fieldprint(old, new, field, text, reprs):
             text += f" [{old['table']}]"
         if field == "relation":
             text += " [nothing to do]"
-    reprs[module_map(old["module"])].append(f"{fullrepr}: {text}")
+    if field != "module":
+        reprs[module_map(new["module"])].append(f"{fullrepr}: {text}")
     if field == "module":
+        reprs[module_map(old["module"])].append(f"{fullrepr}: {text}")
```
MiquelRForgeFlow, line 134: FYI, I did this change now for one reason:

If you look into https://github.com/OCA/OpenUpgrade/pull/4918, you see that the "now required" is shown in `sale_project`, but I think it's better if it's shown in `product`.

See https://github.com/OCA/OpenUpgrade/pull/5050:
Selection_4482

pedrobaeza (approved):
Merging seeing the result.

/ocabot merge patch

PR #3261, opened 2025-04-10
[17.0][FIX] dbs value has to be kept in case of special proxy config

cvinh:
Same as https://github.com/OCA/server-tools/pull/2828

### DIFF
--- a/dbfilter_from_header/override.py
+++ b/dbfilter_from_header/override.py
@@ -13,11 +13,13 @@ db_filter_org = http.db_filter
 
 
 def db_filter(dbs, host=None):
-    dbs = db_filter_org(dbs, host)
+    dbs_orig = db_filter_org(dbs, host)
     httprequest = http.request.httprequest
     db_filter_hdr = httprequest.environ.get("HTTP_X_ODOO_DBFILTER")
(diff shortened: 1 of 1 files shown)


matmicro:
I propose a change into the OCA module dbfilter_from_header.

The change is to replace the line:
https://github.com/OCA/server-tools/blob/16.0/dbfilter_from_header/override.py:20

`        dbs = [db for db in dbs if re.match(db_filter_hdr, db)]
`
And replace it by:

`        dbs = [db for db in dbs_orig if re.match(db_filter_hdr, db)]
`
This change will allow to combine dbfilter_from_header AND the basic dbfilter.
This is really useful the generate some powerful regex like:

```
dbfilter = ^prod_\w+
dbfilter_from_header = \w+_%d$
```
Or maybe this commit is a regression that was introduced, and do not allow anymore to use both filters at the same time: @cvinh
https://github.com/OCA/server-tools/commit/e39ed297fe29adc487f3802ce65b40794495437c

hbrunn (changes requested):
see https://github.com/OCA/server-tools/pull/3694#pullrequestreview-4875691553

PR #3268, opened 2025-04-22
Update es.po

Lopsanz:
"and" and "or"  shouldn't be translated. Raises ERROR on multilanguage environments


StefanRijnhart:
Please update the translation in Weblate. See https://odoo-community.org/resources/translate

PR #3270, opened 2025-04-25
[16.0][ADD] postgres_vacuum

hbrunn:
This module adds a cronjob that vacuums or analyses postgres tables.

As postgres' autovacuum is not recommended for busy databases, it can be
a better choice to schedule this process manually.

Configuration
=============

1. Go to Settings/Technical/Automation/Scheduled Actions
2. Select actions *Vacuum Postgresql Tables* and *Analyze Postgresql
   Tables*
3. Adapt the schedule to run at a time when your Odoo installation is
   least busy, take care that the analyze and vacuum cronjobs won't run
   simultaneously
4. Adapt the parameters to match your setting, but the defaults should
   be fine for most users
5. Note that the vacuum cronjob is inactive by default, because this is
   a potentially very heavy operation


hbrunn:
tests are fixed in https://github.com/OCA/server-tools/pull/3271

postgres_vacuum/models/ir_cron.py
```diff
@@ -0,0 +1,67 @@
...
+        order_by_clause = "last_analyze"
+        if full_vacuum:
+            order_by_clause = "last_vacuum"
+
+        connection = db_connect(self.env.cr.dbname)
+
+        with connection.cursor() as cr:
+            cr._cnx.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+            # pylint: disable=sql-injection
+            cr.execute(f"SET statement_timeout = '{max_minutes}min'")
```
StefanRijnhart: Looks like this might work. Can you make this more secure?
```
model._postgres_vacuum("10min'; drop table res_users cascade; set statement_timeout = '10", False)
```

postgres_vacuum/readme/ROADMAP.md
```diff
@@ -0,0 +1 @@
+- it would be nice if the cronjob could detect some odoo-specific measure of "busyness" and only run if it is below some threshold
```
StefanRijnhart, line 1: Maybe add some statistics in the model views about number of dead tuples, last vacuum timestamp etc from https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ALL-TABLES-VIEW

postgres_vacuum/models/ir_cron.py
hbrunn: my thinking was that at this point you can just as well say `env.cr('nefarious stuff')`, but you're right of course, defense in depth is always good, especially if it's as simple as here.

postgres_vacuum/readme/ROADMAP.md
hbrunn: added to roadmap

StefanRijnhart (approved):

thomaspaulb (approved):

gurneyalex (approved):

PR #3275, opened 2025-04-30
[18.0][ADD] database_size

StefanRijnhart:
Monitor the size of your Odoo instance.

Configuration
=============

To configure this module, you can review the scheduled action called
'Take model size measurements' and check the time at which you want it
to run. It should only run once a day. If it runs more often, it just
updates the existing set of sizes for the day.

Usage
=====

You can use this module to keep an eye on the development of the size of
your Odoo instance over time. Every day, a snapshot will be taken with
the full size of the database and the attachments. You can query these
daily snapshots, and you can compare the current size with a size at any
date of the past for which there is data.

Enable debug mode, then go to menu Settings -> Technical -> Database
Size.

image

The data that is gathered and that is displayed are:

-  Model - The name of the model to which the data is related
-  Estimated Rows - The number of estimated rows according to the
   Postgresql query planner. For performance reasons, taking the data
   from the planner is preferred over doing an actual count, although
   the results may be imprecise.
-  Bare Table Size - The disk usage of the model table without indexes
   etc.
-  Index Size - The disk usage of the indexes in the model table.
-  Many2many Tables Size - The disk usage of related many2many tables,
   including their indexes. To prevent double counts, many2many tables
   are only correlated with one of their tables (the largest of the
   two).
-  Attachment Size - The disk usage of the attachments linked to the
   model records. Because Odoo will deduplicate attachments by content,
   attachments with the same content may be counted double in the
   attachment size of other models, but will not be counted double then
   linked to records of the same model more than once.
-  Total Table Size - Bare Table Size + Index Size
-  Total Database Size - Total Table Size + Many2many Tables Size
-  Total Model Size - Total Database Size + Attachment Size

If you click on individual records, you can inspect the sizes of each
index and many2many table.

All sizes are in megabytes.

In the 'Compare Size per Model' report view, you can find these data
twice: once for the selected measurement date (default: today), and once
for the selected comparison date (default: one month ago).

image

If you want to compare arbitrary dates, you can start typing the date in
the search box. Be sure to enter the dates in the right format for your
localization.

image


database_size/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2025 Opener B.V. <https://opener.amsterdam>
```
NL66278: There should be no copyright line in init files.

NL66278 (approved):
:+1: One nag but LGTM

database_size/models/ir_model_size.py
```diff
@@ -0,0 +1,261 @@
...
+            "unique(model_id, measurement_date)",
+            "There is already a measurement for this model on the given date",
+        ),
+    ]
+    model_id = fields.Many2one(
+        comodel_name="ir.model",
+        index=True,
+        ondelete="cascade",
+        required=True,
+    )
```
legalsylvain: The current design dosn't allow to uninstall modules.
if you uninstall modules, the previous data will be dropped, which is not the expected behaviour, I suppose. (don't you think ?)

I think we can replace by "model_name" (or add model_name) here. So if module are uninstalled, data are conserved.

database_size/models/__init__.py
StefanRijnhart: Fixed, thanks!

database_size/models/ir_model_size.py
```diff
@@ -0,0 +1,261 @@
...
+        help="Total Model Size in MB. This includes many2many tables",
+        store=True,
+    )
+    total_table_size = fields.Integer(
+        help="Total Table Size in MB. This includes indexes and toast tables",
+    )
+    table_size = fields.Integer(
+        string="Bare Table Size",
+        help="Bare Table Size in MB.",
+    )
```
legalsylvain, line 49:
```suggestion
    table_size = fields.Integer(
        string="Bare Table Size",
        help="Bare Table Size, in Bytes",
    )

    table_size_mb = fields.Float(
        compute=lambda x: x._compute_mb_size("table_size"),
        string="Bare Table Size (MB)",
        help="Bare Table Size in MB.",
    )
    def _compute_mb_size(self, field_name):
        for item in self:
            setattr(item, field_name + "_mb", getattr(item, field_name))
```

With the current design, information under MB are lost. I understand that In most cases, information rounded to the MB is sufficient, but it's a shame to lose a level of information. 
Also, a table can be assumed to be empty if its size is < 1MB.

I therefore suggest keeping the value in Bytes (Integer) , and having a calculated field (float) in MB.

database_size/models/ir_model_size.py
```diff
@@ -0,0 +1,261 @@
...
+                "model_id": model_id,
+                "measurement_date": today,
+                "ir_model_index_size_ids": [],
+                "ir_model_relation_size_ids": [],
+            }
+            for model_id in table2model_id.values()
+        }
+        # Some many2many relation objects are linked explicitely to both models
+        # involved. To prevent counting them double, we will link them to the
+        # largest table. Gather all the related models first.
```
legalsylvain, line 164: Nice catch. However, largest table is not predictive. And this may change over time, which could lead to variations in size. 
What about to take allways the first one ? (first in alphabetical order, or in ID order.)
Also it could simplify algorithm.
not a blocking point, though.

legalsylvain (commented):
hi @StefanRijnhart . Thanks for this great module ! 

some remarks inline. Otherwise, LGTM. (code review & functional test).

database_size/models/ir_model_size.py
StefanRijnhart: The aim of the module is to check for large tables and their growth. Many2many relations are hardly ever symmetric, far from it. Therefore, if the model related table grows, the many2many table grows, and trimming down the large model table will also trim down the many2many table. It would be misleading to attribute the many2many table on any other criteria than size. Indeed, if you start out with 40 partner categories and 10 partners, the linked model will switch but this will only be the case when the table sizes are trivially small. So I would argue against your suggestion.

database_size/models/ir_model_size.py
StefanRijnhart: The aim of the module is to check for large tables and their growth. Many2many relations are hardly ever symmetric, far from it. Therefore, if the model related table grows, the many2many table grows, and trimming down the large model table will also trim down the many2many table. It would be misleading to attribute the many2many table on any other criteria than size. Indeed, if you start out with 40 partner categories and 10 partners, the linked model will switch but this will only be the case when the table sizes are trivially small. So I would argue against your suggestion.

database_size/models/ir_model_size.py
StefanRijnhart: The aim of this module is to monitor the size per model, especially the growth of that size.  Keeping track of very small tables, and of very small differences only distract from that aim. It will create additional ir.model.size records that show up when inspecting the report (Unrounded numbers were my first implementation, but I decided to go for MB when I first saw the huge lists that it produced, covering lots of unused or marginal tables). When comparing two dates, the records with only minimal changes will clutter up the screen when trying to find out which tables changed size in a meaningful way. By only taking megabytes into account, the usability of the module improves and the data it produces remains manageable. I can't think of a benefit to keep track of smaller quantities at the moment. Do you have a concrete goal that can be achieved by this?

database_size/README.rst
```diff
@@ -0,0 +1,142 @@
...
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you can review the scheduled action called
+'Take model size measurements' and check the time at which you want it
+to run. It should only run once a day. If it runs more often, it just
+update the existing set of sizes for the day.
```
hbrunn:
```suggestion
updates the existing set of sizes for the day.
```

database_size/README.rst
```diff
@@ -0,0 +1,142 @@
...
+-  Index Size - The disk usage of the indexes in the model table.
+-  Many2many Tables Size - The disk usage of related many2many tables,
+   including their indexes. To prevent double counts, many2many tables
+   are only correlated with one of their tables (the largest of the
+   two).
+-  Attachment Size - The disk usage of the attachments linked to the
+   model records. Because Odoo will deduplicate attachments by content,
+   attachments with the same content may be counted double in the
+   attachment size of other models, but will not be counted double then
+   linked to records of the same model more than once.
```
hbrunn:
```suggestion
   attachment size of other models, but will not be counted double when
   linked to records of the same model more than once.
```

database_size/models/ir_model_size.py
hbrunn: why is this unexpected? Odoo drops tables of deleted models, so I'd expect also its entry here to vanish.

database_size/report/ir_model_size_report.py
```diff
@@ -0,0 +1,232 @@
...
+    @property
+    def _table_query(self):
+        """Report comparative database size changes between two dates.
+
+        The dates are inserted in the context in this model's `search` method.
+        """
+        measurement_date = self.env.context.get("measurement_date")
+        if measurement_date:
+            if isinstance(measurement_date, str):
+                measurement_date = fields.Date.to_date(measurement_date)
```
hbrunn: I'd call to_date unconditionally, it does the right thing when you pass a date/datetime there. also below

hbrunn (changes requested):
found a few typos

database_size/models/ir_model_size.py
legalsylvain: Because you loose the size of the database in the past.

StefanRijnhart: I agree with @legalsylvain. It would be weird for the database size to shrink retroactively after uninstalling some module. The code is now updated accordingly.

database_size/README.rst
StefanRijnhart: Thanks, taken!

database_size/README.rst
StefanRijnhart: Thanks, taken!

database_size/report/ir_model_size_report.py
StefanRijnhart: Wasn't aware, thanks!

hbrunn (approved):
you might want to add some (inactive) cleanup cronjob, or is the size table supposed to grow indefinitely?

database_size/models/ir_model_size.py
legalsylvain: Hi.

> It will create additional ir.model.size records that show up when inspecting the report (

Well, I used the current implementation, and it looks that there are record for "very small tables", so I'm not sure to understand.

image

> Do you have a concrete goal that can be achieved by this?

No. It was a reflex to think in terms of : “Being precise increases potential uses.”

let's go merge the current implementation. If people needs more precise data, it can be changed later quite easely.

legalsylvain (approved):
Thanks for this great module.

did you planned to backport this module in less recent version ? (I could be interested by v16).

database_size/models/ir_model_size.py
StefanRijnhart: Thanks for accepting my preference! And you are right, it does not make a difference in the number of records.

StefanRijnhart:
@legalsylvain No plans for a backport from my side.

StefanRijnhart:
Oh, before the merge I should probably implement @hbrunn's suggestion of the cleanup cron. Give me a couple of days.

florian-dacosta (approved):
LGTM
You may want to add you as maintainer (not sure if you don't want to or forgot).

About the cleaning, do we really need one more cron for this or would it be enough to add the @api.autovacuum decorator and implement it with a delay that seems reasonable to all? + an easy way to override just in case ?

Something like 
```
    @api.autovacuum
    def _delete_records(self):
        deletion_limit_date = date.today() - timedelta(days=self._removal_interval)
        records = self.search(
            [
                ("create_date", "<", deletion_limit_date),
            ]
        )
        records.unlink()
```
Just a suggestion of course, the separated cron offers more flexibility (allowing to leave it unactive mainly)

StefanRijnhart:
@florian-dacosta Thanks for the suggestion! I'd like to keep this more flexible, as you mention with the option to leave the cron inactive and some parameters to configure the retention period.

hbrunn:
@StefanRijnhart my idea was in no way meant to be blocking, so I'd be happy to merge this as is

StefanRijnhart:
@hbrunn No, I think it's a good idea.

StefanRijnhart:
Feature to purge older data is now pushed. @florian-dacosta of course I just had to use res.config.settings to configure the parameters and then I could use your suggestion after all. Thanks again!

gurneyalex (approved):

legalsylvain:
@vdewulf : I think this module is very interesting for people with huge instances.
maybe some communication in the next newsletter could be great, to promote it.

vdewulf:
> @vdewulf : I think this module is very interesting for people with huge instances. maybe some communication in the next newsletter could be great, to promote it.

Sure!

PR #3249, opened 2025-04-04
[18.0][MIG] base_remote: Migration to 18.0

BhaveshHeliconia:

PR #3279, opened 2025-05-07
[17.0][BKP] base_remote: backport from 18.0

Kimkhoi3010:
Depend on:

- https://github.com/OCA/server-tools/pull/3249


etobella (approved):
/ocabot migration base_remote

I have a doubt if this should be called MIG too, but well, it is just a not relevant question for this.

PR #3280, opened 2025-05-08
[ADD] field_float_nullable: New field type float that supports NULL v…

samirGuesmi:
New field type float that supports NULL values instead of defaulting to 0.0

In standard Odoo, ``fields.Float`` does not support distinguishing
between ``0.0`` and ``NULL`` (i.e., an unset value). By default, if a
float field is left empty in the UI, it is stored as ``0.0`` in the
database. The goal is to allow users and developers to identify whether
a float field has been filled in or not — something that cannot be
reliably determined when ``0.0`` is used by default. The module also
integrates fully with Odoo’s UI (form views, list views, export, custom
filters


field_float_nullable/readme/CONFIGURE.md
```diff
@@ -0,0 +1,10 @@
+[ This file is not always required; it should explain **how to configure the module before using it**; it is aimed at users with administration privileges. 
+
+Please be detailed on the path to configuration (eg: do you need to activate developer mode?), describe step by step configurations and the use of screenshots is strongly recommended.]
+
+
+To configure this module, you need to:
+
+- Go to *App* > Menu > Menu item
+- Activate boolean… > save
+- …
```
sbidoul: remove this file

field_float_nullable/readme/HISTORY.md
```diff
@@ -0,0 +1,22 @@
...
+- [BREAKING] Breaking changes come first.
+  ([#70](https://github.com/OCA/repo/issues/70))
+- [ADD] New feature.
+  ([#74](https://github.com/OCA/repo/issues/74))
+- [FIX] Correct this.
+  ([#71](https://github.com/OCA/repo/issues/71))
+
+## 11.0.x.y.z (YYYY-MM-DD)
+
+- ...
```
sbidoul: remove this file

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+    By default, Odoo uses 0.0 as the default value for fields.Float, which can lead
+    to ambiguity between "unset" and "zero" values.
+    """
+
+    type = "float_nullable"
+    column_type = ("numeric", "numeric")
+    search = True
+
+    def convert_to_column(self, value, record, values=None, validate=True):
+        if value is None or value == "" or (value is False and type(value) is bool):
```
sbidoul:
```suggestion
        if value is None or value == "" or value is False:
```

If it is False, then it is a bool, by definition.

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+            return None
+        try:
+            return float(value) if value is not None else None
+        except (ValueError, TypeError):
+            return None
+
+    def convert_to_cache(self, value, record, validate=True):
+        if value is None or value == "" or (value is False and type(value) is bool):
```
sbidoul:
```suggestion
        if value is None or value == "" or value is False:
```

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+    column_type = ("numeric", "numeric")
+    search = True
+
+    def convert_to_column(self, value, record, values=None, validate=True):
+        if value is None or value == "" or (value is False and type(value) is bool):
+            return None
+        try:
+            return float(value) if value is not None else None
+        except (ValueError, TypeError):
+            return None
```
sbidoul, line 18: Is it not better to call `super().convert_to_column()` instead of re-implementing the conversion logic?

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def convert_to_cache(self, value, record, validate=True):
+        if value is None or value == "" or (value is False and type(value) is bool):
+            return None
+        try:
+            return float(value)
+        except (ValueError, TypeError):
+            return None
```
sbidoul, line 25: Same here, call super?

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+    def convert_to_record(self, value, record):
+        return value
+
+    def convert_to_read(self, value, record, use_name_get=True):
+        return value
+
+    def convert_to_export(self, value, record):
+        if value or value == 0.0:
+            return value
+        return ""
```
sbidoul: This `convert_to_export` method is identical to the one in the parent class, so it can be removed.

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            return float(value)
+        except (ValueError, TypeError):
+            return None
+
+    def convert_to_record(self, value, record):
+        return value
+
+    def convert_to_read(self, value, record, use_name_get=True):
+        return value
```
sbidoul: The default `convert_to_read` does `return False if value is None else value`, is it not better to keep that? I'm not a fan of these False value but I think it's still necessary for xmlrpc.

acsonefho (approved):
Code LGTM :muscle: :candy:

sbidoul (commented):
Thanks for this contrib! A first round of questions on the backend field class.

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+class FloatNullable(fields.Float):
+    """
+    The FloatNullable field is a custom Odoo field type that extends the standard fields.Float
+    to allow NULL values in the database column.
+    By default, Odoo uses 0.0 as the default value for fields.Float, which can lead
+    to ambiguity between "unset" and "zero" values.
+    """
+
+    type = "float_nullable"
+    column_type = ("numeric", "numeric")
```
sbidoul:
```suggestion
```

Use the same property of the base class?

field_float_nullable/fields.py
```diff
@@ -0,0 +1,46 @@
...
+    search = True
```
sbidoul:
```suggestion
```

Any reason to set `search = True` by default on this particular field?

field_float_nullable/fields.py
samirGuesmi: Done

field_float_nullable/fields.py
samirGuesmi: indeed

field_float_nullable/fields.py
samirGuesmi: nop, I forgot to remove it.

samirGuesmi:
Thanks @sbidoul for the review! I've applied the suggested changes.

field_float_nullable/tests/test_float_nullable.py
```diff
@@ -0,0 +1,102 @@
...
+        self.assertFalse(self.record_empty.value_float is None)
+        # check db storage value
+        self.assertIsNone(
+            self._get_db_storge_value(self.record_empty, "value_float_nullable")
+        )
+        with self.assertRaises(AssertionError):
+            self.assertTrue(
+                self._get_db_storge_value(self.record_empty, "value_float_nullable")
+                == 0.0
+            )
```
sbidoul:
```suggestion
          self.assertFalse(
              self._get_db_storge_value(self.record_empty, "value_float_nullable")
              == 0.0
          )
```

field_float_nullable/tests/test_float_nullable.py
```diff
@@ -0,0 +1,102 @@
...
+        with self.assertRaises(AssertionError):
+            self.assertIsNone(
+                self._get_db_storge_value(self.record_empty, "value_float")
+            )
```
sbidoul:
```suggestion
        self.assertTrue(
            self._get_db_storge_value(self.record_empty, "value_float") is not None
        )
```

field_float_nullable/tests/test_float_nullable.py
```diff
@@ -0,0 +1,102 @@
...
+        # pylint: disable=import-outside-toplevel
+        from .test_model import TestFloatNullableModel
+
+        cls.loader.update_registry([TestFloatNullableModel])
+
+        cls.test_model = cls.env[TestFloatNullableModel._name]
+
+        return res
+
+    def _get_db_storge_value(self, record, field_name):
```
sbidoul:
```suggestion
    def _get_db_storage_value(self, record, field_name):
```

field_float_nullable/models/ir_model_fields.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class IrModelFields(models.Model):
+    _inherit = "ir.model.fields"
+
+    ttype = fields.Selection(
+        selection_add=[("float_nullable", "FloatNullable")],
+        ondelete={"float_nullable": "cascade"},
```
AnizR: Did you consider 'set float' instead of 'cascade'?
Deleting every `float_nullable` seems a bit risky.

AnizR (approved):
I just have one little question :smile: 
Code Looks good to me, great contribution !

sbidoul (commented):
A few comments on the style of the tests.

field_float_nullable/static/src/js/float_field.esm.js
```diff
@@ -0,0 +1,70 @@
...
+import {Component} from "@odoo/owl";
+import {_lt} from "@web/core/l10n/translation";
+import {formatFloat} from "@web/views/fields/formatters";
+import {parseFloat} from "@web/views/fields/parsers";
+import {registry} from "@web/core/registry";
+import {standardFieldProps} from "@web/views/fields/standard_field_props";
+import {useInputField} from "@web/views/fields/input_field_hook";
+import {useNumpadDecimal} from "@web/views/fields/numpad_decimal_hook";
+
+export class FloatNullableField extends Component {
```
sbidoul: Is it not better to extend FloatField instead of Component, and call super in parse and get formattedValues to reuse the base logic of the float field. There are examples in Odoo, such as FloatScannableField.

field_float_nullable/models/ir_model_fields.py
samirGuesmi: Thanks @AnizR, nice catch, I've updated the code accordingly.

samirGuesmi:
Thans @sbidoul and @AnizR for the review.
here is the updated version.

field_float_nullable/static/src/js/float_field.esm.js
```diff
@@ -0,0 +1,30 @@
...
+import {formatFloat} from "@web/views/fields/formatters";
+import {parseFloat} from "@web/views/fields/parsers";
+import {registry} from "@web/core/registry";
+
+export class FloatNullableField extends FloatField {
+    parse(value) {
+        if (value === "" || value === null) {
+            return null;
+        }
+        return parseFloat(value);
```
sbidoul: any reason to not call super here?

field_float_nullable/static/src/js/float_field.esm.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    get formattedValue() {
+        if (this.props.value === null || this.props.value === false) {
+            return "";
+        }
+        return formatFloat(this.props.value, {digits: this.props.digits});
```
sbidoul: any reason to not call super here?

field_float_nullable/static/src/js/float_nullable_search.esm.js
```diff
@@ -0,0 +1,19 @@
...
+        this.FIELD_TYPES.float_nullable = "number";
+    },
+
+    validateField(field) {
+        return (
+            this._super(field) ||
+            (field.type === "float_nullable" && field.searchable && !field.deprecated)
+        );
+    },
+});
```
sbidoul, line 19: Why is this patch necessary?

samirGuesmi: On the first run, the field will not appear in the custom search because validateField is executed during setup. This happened before float_nullable is added to FIELD_TYPES.

PR #3285, opened 2025-05-14
[FIX] base_multi_image: Access Error

erlinberg:
I encountered an AccessError when a non-admin user tries to access a `base_multi_image` form or tree view or when any model that inherits from base_multi_image is used. This error happens because the `owner_ref_id` field computation tries to access the `ir.model` table without sufficient permissions.

To fix this, I added .sudo() to the `_selection_owner_ref_id` method to ensure the system has the necessary access rights to read from ir.model, regardless of the user's permissions.

### DIFF
--- a/base_multi_image/models/image.py
+++ b/base_multi_image/models/image.py
@@ -62,7 +62,7 @@ class Image(models.Model):
     @tools.ormcache("self")
     def _selection_owner_ref_id(self):
         """Allow any model; after all, this field is readonly."""
-        return [(r.model, r.name) for r in self.env["ir.model"].search([])]
+        return [(r.model, r.name) for r in self.env["ir.model"].sudo().search([])]
 
     @api.depends("owner_model", "owner_id")
     def _compute_owner_ref_id(self):
--- a/base_multi_image/tests/__init__.py
+++ b/base_multi_image/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_base_multi_image
--- a/base_multi_image/tests/test_base_multi_image.py
+++ b/base_multi_image/tests/test_base_multi_image.py
@@ -1,0 +1,42 @@
+from odoo.tests import new_test_user
+from odoo.tests.common import TransactionCase
+
+
+class TestBaseMultiImage(TransactionCase):
+    def setUp(self):
+        super().setUp()
+        # Create a non-admin user
+        self.base_user = new_test_user(
+            self.env,
+            name="Base User",
+            login="base_user",
+            groups="base.group_user",
+        )
+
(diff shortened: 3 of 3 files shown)


erlinberg:
@legalsylvain @ivs-cetmix @Aldeigja please review

ivs-cetmix:
@erlinberg thank you for you contribution! Did you manage to reproduce this error on the runboat? I checked it on 14.0 with "Product Multi Image" and looks like there are some other access error being raised.

erlinberg:
@ivs-cetmix I reproduced the issue on runboat by adding a simple view action called "Product Images Demo Action" to Sales, which includes a tree and form view for base_multi_image.image. I applied an owner domain and linked it to a menu item called "Product Images Demo", with access rights set for non-admin users.

When opened by the demo user, an AccessError is raised because regular users don’t have access to the ir.model records that base_multi_image.image relies on.

http://oca-server-tools-16-0-57c6d831b62d.runboat.odoo-community.org/

image

ivs-cetmix:
@erlinberg makes sense, thank you! Please also add a more complete description to the commit message. Eg:

```
[FIX] base_multi_image: Access Error

**Before this commit**

Access error is raised if user who doesn't have access to system settings tries to open a record that contains multiple images.

**After this commit**

Access error is not raised any longer because model list is fetched using `sudo`.
```

ivs-cetmix (approved):
Code review LGTM

ivs-cetmix:
@erlinberg please also add tests that cover your change.

base_multi_image/models/image.py
```diff
@@ -62,7 +62,7 @@ class Image(models.Model):
     @tools.ormcache("self")
     def _selection_owner_ref_id(self):
         """Allow any model; after all, this field is readonly."""
-        return [(r.model, r.name) for r in self.env["ir.model"].search([])]
+        return [(r.model, r.name) for r in self.env["ir.model"].sudo().search([])]
```
rousseldenis, line 65: Mmmh, maybe a check_access for `base.group_user` here before ?

erlinberg: I don’t think that adding a check_access provides any real benefit, given that the field is already readonly. Adding it would only introduce unnecessary complexity.

erlinberg:
@ivs-cetmix Updated the commit message and added tests. Thanks!

ivs-cetmix:
> @ivs-cetmix Updated the commit message and added tests. Thanks!

@erlinberg thank you!

base_multi_image/models/image.py
ivs-cetmix: @rousseldenis what do you think ?

PR #3286, opened 2025-05-15
[15.0][REV][FIX] Pin cryptography to keep compatibility with Odoo's pyopenssl==19.0.0

em230418:
Reverted commit disallows to install auto_backup, when using offical Odoo 15.0 docker image, which is based on bullseye [1].
The bullseye has following python packages:

- cryptography 3.3.2 with security patches [2]
- pyopenssl 20.0.1 [3]

[1] https://github.com/odoo/docker/blob/5fb6a842747c296099d9384587cd89640eb7a615/15.0/Dockerfile#L1
[2] https://packages.debian.org/bullseye/python3-cryptography
[3] https://packages.debian.org/bullseye/python3-openssl

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,1 +1,3 @@
+# place an upper bound on cryptography for compatibility with pyopenssl 19 in Odoo's requirements.txt
+cryptography<37
 odoo-test-helper
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -19,7 +19,7 @@ jobs:
           python-version: "3.11"
       - name: Get python version
         run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
-      - uses: actions/cache@v1
+      - uses: actions/cache@v4
         with:
           path: ~/.cache/pre-commit
           key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
--- a/auto_backup/__manifest__.py
+++ b/auto_backup/__manifest__.py
@@ -24,5 +24,5 @@
         "view/db_backup_view.xml",
     ],
     "installable": True,
-    "external_dependencies": {"python": ["pysftp", "cryptography==2.6.1"]},
+    "external_dependencies": {"python": ["pysftp"]},
 }
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 # generated from manifests external_dependencies
 acme<2.0.0
 astor
-cryptography==2.6.1
 dataclasses
 dnspython
 josepy


em230418:
Hello, @sbidoul! Bumping you as maintainer of oca-ci. Could you please install packages from https://github.com/odoo/odoo/blob/15.0/requirements.txt and then install odoo itself?

StefanRijnhart (changes requested):
There still has to be some sort of pin if we don't want to run into https://github.com/odoo/odoo/pull/99829 again. Have a look https://github.com/OCA/server-auth/pull/429

em230418:
> There still has to be some sort of pin if we don't want to run into [odoo/odoo#99829](https://github.com/odoo/odoo/pull/99829) again. Have a look [OCA/server-auth#429](https://github.com/OCA/server-auth/pull/429)

server-auth is other repository and I don't see problems with pining cryptography<37 in that repository.

em230418:
> There still has to be some sort of pin if we don't want to run into https://github.com/odoo/odoo/pull/99829 again.

First `python3 -m pip install requirements.txt` from https://github.com/odoo/odoo/blob/15.0/requirements.txt and then install odoo. That is the thing I suggested to @sbidoul in my message to him in this thread.

StefanRijnhart:
Given that @sbidoul is actually a maintainer of Python's pip I would say he's among those of us in the OCA that is most knowledgeable about the installation of python requirements. Maybe you can reconsider adopting his own approach to this problem (in https://github.com/OCA/server-auth/pull/429).

em230418:
Stéphane Bidoul, sorry for pinging for hasty reasons. In oca-ci packages are already installed like I assumed and asked.

StefanRijnhart (dismissed):
Sorry, I think having '<37' is way better than the current pinning, but I don't agree removing the pin entirely.

StefanRijnhart:
For the record, my suggestion is to follow https://github.com/OCA/server-auth/pull/429 in adding '<37' in requirements.txt, not in test-requirements.txt.

em230418:
> For the record, my suggestion is to follow [OCA/server-auth#429](https://github.com/OCA/server-auth/pull/429) in adding '<37' in requirements.txt, not in test-requirements.txt.

requirements.txt is autogenerated from module. Adding "cryptography<37" to auto_backup manifest is not correct. It is not auto_backups' scope to deal with odoo's subsubdependency.

em230418:
> It is not auto_backups' scope to deal with odoo's subsubdependency.

Pardon. It is not auto_backups' scope to deal with its own subsubdependency. For that we already have requirements.txt from odoo/odoo or OCA/OCB.

sbidoul:
I wrote an explanation of the root cause of a similar problem in https://github.com/OCA/server-auth/issues/424#issuecomment-1268492655

I also agree an exact pin must absolutely be avoided.

> It is not auto_backups' scope to deal with its own subsubdependency

I tend to agree with that. And TBH I don't quite remember right now why I put the upper bound in the manifest in https://github.com/OCA/server-auth/pull/429. I can see a benefit for some users, but also a downside for users who would want to use cryptography>=37 and take care of installing a compatible pyopenssl.

So I'm +1 with putting the upper bound on cryptography in `test-requirements.txt` and see how it goes.

BTW, these topics are super complex. For instance `python3.8 -m pip install -r odoo/requirements.txt pysftp` lead to a valid solution with cryptography 2.6.1 (both with pip and uv pip). These SAT solvers can lead to different valid solutions and which they prefer is not guaranteed in any documentation.

test-requirements.txt
```diff
@@ -1 +1,2 @@
+cryptography<37
```
sbidoul, line 2:
```suggestion
# place an upper bound on cryptography for compatibility with pyopenssl 19 in Odoo's requirements.txt
cryptography<37
```

sbidoul:
> For the record, my suggestion is to follow [OCA/server-auth#429](https://github.com/OCA/server-auth/pull/429) in adding '<37' in requirements.txt, not in test-requirements.txt.

@StefanRijnhart as I don't remember why I preferred that too in 2022 and now putting it in test-requirements.txt sounds somewhat better to me, do you have specific arguments in mind yourself?

StefanRijnhart (commented):
@sbidoul While it not being auto_backup's responsibility, as per OCA's implementation the dependency from the manifest with or without version ends up in requirements.txt, and this will be the basis for many deploys, either directly or indirectly, for the installation of these dependencies. I consider therefore adding the version to auto_backup's manifest as a valid workaround (or helpful guidance) for the indeed 'super complex' issue of preventing the breaking version of cryptography from being installed on any instance.
Then again, I respect your knowledge and authority on the issue so I will retract my blocking review and may abstain from further reviews.

StefanRijnhart (commented):
-- trying to get rid of my own 'Request Changes' review.

sbidoul (approved):

PR #3287, opened 2025-05-15
[FIX] module_change_auto_install : prevent to crash if a module is not present in the addons path.    

legalsylvain:
Rational:

- Set modules_auto_install_enabled=web_no_bubble
- Create a new instance with web_no_bubble installed
- Then restart the server, removing OCA/web repo from the addons path.
For the time being, the server crashes, because 'depends' key is not available.
With that patch, the classic 'module web_no_bubble: not installable, skipped' message is displayed in the log,
but the server doesn't crash.


tarteo (approved):

tarteo:
Makes sense

legalsylvain:
Thanks for your review ! 

/ocabot merge patch

PR #3289, opened 2025-05-20
[18.0][FIX] upgrade_analysis: avoid breaking coverage file

MiquelRForgeFlow:
Sometimes, the template columns are broken due to module length. This fix assures this never happens again.


pedrobaeza (commented):
Isn't better to precompute the width for the largest module name?

MiquelRForgeFlow:
> Isn't better to precompute the width for the largest module name?


No, because the length is already set in the template https://github.com/OCA/server-tools/blob/18.0/upgrade_analysis/templates/module_coverage_template.xml.

pedrobaeza (approved):

StefanRijnhart (approved):

PR #3290, opened 2025-05-20
[18.0][FIX] upgrade_analysis: avoid caring about non stored fields

MiquelRForgeFlow:
Long time ago, I added https://github.com/OCA/OpenUpgrade/commit/b724b4f0d5f4813ac0d09418c66dc491d93e000b in v12. I did it because Odoo started to add computed/related stored fields in v13. But Odoo in last versions has been adding non stored fields that are not compute nor related, like https://github.com/odoo/odoo/blob/18.0/addons/account/models/account_account.py#L142 and https://github.com/odoo/odoo/blob/18.0/addons/account/models/account_move.py#L186. This kind of field adds noise in the analysis file, so let's remove them.

NOTE: Usual `company_dependent` fields and usual `translate` fields are marked as stored, so no problem with these cases.

### DIFF
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -302,15 +302,18 @@ def compare_sets(old_records, new_records):
     for column in old_records:
         if column["field"] == "_order":
             continue
-        # we do not care about removed non stored function fields
+        # we do not care about removed non stored function/related fields
         if not column["stored"] and (column["isfunction"] or column["isrelated"]):
             continue
         if column["mode"] == "create":
             column["mode"] = ""
+        printkeys = printkeys_old.copy()
+        if not column["stored"]:
+            printkeys.extend(["stored"])
         extra_message = ", ".join(
             [
-                k + ": " + str(column[k]) if k != str(column[k]) else k
-                for k in printkeys_old
+                k + ": " + str(column[k] or False) if k != str(column[k]) else k
+                for k in printkeys
                 if column[k]
             ]
         )
@@ -321,7 +324,7 @@ def compare_sets(old_records, new_records):
     for column in new_records:
         if column["field"] == "_order":
             continue
-        # we do not care about newly added non stored function fields
+        # we do not care about newly added non stored function/related fields
         if not column["stored"] and (column["isfunction"] or column["isrelated"]):
             continue
         if column["mode"] == "create":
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Makes sense ! 

Thanks !

StefanRijnhart (commented):
I don't know. For me as a developer, reading about these fields do add to my understanding of the way that Odoo develops. Would it be an option to update the analysis to add `unstored` for NEW or DEL entries of unstored fields?

MiquelRForgeFlow:
> I don't know. For me as a developer, reading about these fields do add to my understanding of the way that Odoo develops. Would it be an option to update the analysis to add unstored for NEW or DEL entries of unstored fields?

Do you mean an option in the form view (a new field) to enable `unstored`? Yeah, but it will be unfilled by default and will not be used for the analysis files in OU. Are you ok with this?

StefanRijnhart:
No, I mean that it says in the analysis something like
```
account      / account.account          / display_mapping_tab (boolean) : NEW hasdefault: default unstored
```
Instead of
```
account      / account.account          / display_mapping_tab (boolean) : NEW hasdefault: default
```
That way the new field still shows up in the analysis, but it is immediately clear to the developer of the scripts that nothing needs to be done.

legalsylvain:
> I don't know. For me as a developer, reading about these fields do add to my understanding of the way that Odoo develops. Would it be an option to update the analysis to add unstored for NEW or DEL entries of unstored fields?

At a first sight, i'm not agree to add such extra change. 
For me, the objective of the analysis file is to enumerate changes that *could* require  openupgrade scripts. If we are sure that a change will never ask a script, i think it should not be added.

I understand your need as a développer, but there are other interesting changes (like changes in functions names / args) that are not listed for the same reason : no action required.

If we décide to add this line, we could add an extra line juste after : ``Nothing to do : non stored fiield``. So we'll save time for openupgraders.

MiquelRForgeFlow:
@pedrobaeza opinions?

StefanRijnhart (approved):
Approving this version, as it implements my suggestion. Thanks!

pedrobaeza:
The code doesn't highlight the results in the file. Can you please show an example?

MiquelRForgeFlow:
Example:

```
account      / account.account          / display_mapping_tab (boolean) : NEW hasdefault: default
```

will be changed to

```
account      / account.account          / display_mapping_tab (boolean) : NEW hasdefault: default, stored: False
```

pedrobaeza (approved):
I suppose this will require an adaptation when migrated to 19.

/ocabot merge patch

PR #3295, opened 2025-05-21
[18.0][ADD] server_action_input_box: Adds a server action wizard input utility

jesanmor:
This module adds a feature that extends server actions by displaying an input dialog before execution.

It allows passing parameters to the action context, such as:
- Text (char)
- Integer
- Float
- Boolean (checkbox)

This can be useful for dynamically filtering or modifying the behavior of the server action based on user input.

The module is named `server_action_input_box` and is compatible with Odoo version 18.0.


server_action_input_box/models/server_action_input_box_line.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ServerActionInputBoxLine(models.Model):
+    _name = "server.action.input.box.line"
+    _description = "Server Action Input Box Line"
+
+    server_action_input_box_id = fields.Many2one(
+        "server.action.input.box", readonly=True, ondelete="cascade"
+    )
+    name = fields.Char("Unique Name", required=True)
```
gdgellatly, line 15: This is called unique name, but has no check to verify that is the case. I am presuming (although early in review) that uniqueness is only within the action itself.

server_action_input_box/models/server_action_input_box_line.py
```diff
@@ -0,0 +1,53 @@
...
+            ("int", "Integer"),
+            ("float", "Floating Point"),
+            ("bool", "Boolean"),
+        ],
+        default="string",
+        required=True,
+    )
+
+    def write(self, vals):
+        if "name" in vals:
```
gdgellatly: Better as constraint on name.

server_action_input_box/models/server_action_input_box_line.py
```diff
@@ -0,0 +1,53 @@
...
+                        vals["name"],
+                    )
+                )
+        record = super().write(vals)
+        return record
+
+    @api.model_create_multi
+    def create(self, vals):
+        record = super().create(vals)
+        for rec in record:
```
gdgellatly: Better as constraint on name (same as above)

server_action_input_box/models/server_action_input_box.py
```diff
@@ -0,0 +1,204 @@
...
+    def parsed_parameters(self, parameters):
+        parsed_parameters_dict = {}
+        for line in self.server_action_input_box_line_ids:
+            try:
+                if line.data_type != "string":
+                    raw_data = parameters[line.name] or 0
+                else:
+                    raw_data = parameters[line.name]
+
+                data_type_conversion = {
```
gdgellatly: This should be outside loop, it is a dispatcher that applies universally. Also the dispatcher itself IMO should be populated from a separate function to allow extension to other field types.

Also, as it stands the dispatcher is not really necessary if `string` selection is renamed `str` because you could simply eval data_type(raw_data) but I'd leave it in.

server_action_input_box/models/server_action_input_box.py
```diff
@@ -0,0 +1,204 @@
...
+        new_context.update(custom_context)
+
+        action_server = self.ir_action_server_id
+        run_server = action_server.with_context(**new_context).sudo().run()
+        return run_server
+
+    def parsed_parameters(self, parameters):
+        parsed_parameters_dict = {}
+        for line in self.server_action_input_box_line_ids:
+            try:
```
gdgellatly, line 166: for me this try/except is way too big and broad. It should only be concerned with ValueError from line 164.

server_action_input_box/models/server_action_input_box.py
```diff
@@ -0,0 +1,204 @@
...
+                if line.data_type != "string":
```
gdgellatly: This section IMO for same reason as before should be pulled out. These kinds of if statements start to couple your business logic to imperative functions. Something like `raw_data = self._sanitize(line.name)` or `raw_data = getattr(self, f'_sanitize_{line.data_type}(line.name)`

server_action_input_box/models/server_action_input_box.py
```diff
@@ -0,0 +1,204 @@
...
+        return res
+
+    @api.model_create_multi
+    def create(self, vals):
+        record = super().create(vals)
+        record.ir_action_server_id = record._create_ir_action_server()
+        return record
+
+    def unlink(self):
+        ir_action_server_id = self.ir_action_server_id
```
gdgellatly, line 205: someone will know better than me, but I feel like an ondelete in a field definition somewhere would negate the need for this.

server_action_input_box/models/ir_action_server.py
```diff
@@ -0,0 +1,48 @@
...
+        return protected_records
+
+    def write(self, vals):
+        server_action_input_box = self.env["server.action.input.box"].search(
+            [("ir_action_server_id", "=", self.id)]
+        )
+        if server_action_input_box:
+            code = server_action_input_box._code_ir_action_server()
+            if server_action_input_box.active_action:
+                model_id = server_action_input_box.model_id.id
```
gdgellatly, line 25: Would rename this `binding_model_id` unless the entry for vals is a typo.

server_action_input_box/models/ir_action_server.py
```diff
@@ -0,0 +1,48 @@
...
+            else:
+                model_id = None
+            vals = {
+                "name": server_action_input_box.name,
+                "model_id": server_action_input_box.model_id.id,
```
gdgellatly, line 30: See above

server_action_input_box/__manifest__.py
```diff
@@ -0,0 +1,32 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Server Action Input Box",
+    "summary": """Shows a parameter input box in a server action
+    under the 'Action' menu of the model.""",
+    "author": "Jesús Sánchez - jesanmor, " "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-tools",
+    "maintainers": ["jesanmor"],
+    "development_status": "Production/Stable",
```
gdgellatly: Probably a way from this. This is a very interesting new module, likely to get a lot of attention. Consider Beta

server_action_input_box/__manifest__.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 jesanmor - Jesús Sánchez <jesanmor.dev@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Server Action Input Box",
+    "summary": """Shows a parameter input box in a server action
+    under the 'Action' menu of the model.""",
+    "author": "Jesús Sánchez - jesanmor, " "Odoo Community Association (OCA)",
```
gdgellatly: String concat

server_action_input_box/.gitignore
```diff
@@ -0,0 +1,17 @@
+# Ignore files generated by the operating system
```
gdgellatly: Remove this file

server_action_input_box/models/server_action_input_box.py
```diff
@@ -0,0 +1,204 @@
...
+        lines_code = self.code.splitlines()
+        indent = "  "
+        indent_lines = "\n".join([indent + line for line in lines_code])
+
+        parameters = ""
+        for line in self.server_action_input_box_line_ids:
+            parameters += indent + line.name + f" = parsed_parameters['{line.name}']\n"
+
+        code = f'''# """ Fixed code block """ #
+context = env.context
```
gdgellatly, line 66: Will comment later when doing functional tests, but initial feel is more complex than needed.

gdgellatly (commented):
Pretty exciting new module tbh. Take these with a grain of salt, I could be wrong on a lot, just a very quick look, initial thoughts and I feel like I need to really test it locally to work out

server_action_input_box/models/server_action_input_box_line.py
jesanmor: Thank you very much for the suggestions and insights!

Regarding the name field: you're right — it's labeled as "Unique Name", but its purpose could be clearer. It's not meant to be globally unique across the model, but rather to serve as a valid Python variable name. This name is used in the server action code to retrieve the input value.

That’s why I validate it using isidentifier() in both create and write, ensuring it conforms to Python syntax rules (e.g., no spaces, special characters, or starting with a digit).

For this reason, a global SQL uniqueness constraint wouldn't be appropriate. The same name can appear in different server actions — uniqueness is only enforced within the same action, and the key point is the validity of the variable name.

Admittedly, "Unique Name" may not be the best label; something like "Variable Name" would be more accurate.

Each server.action.input.box.line record represents an input box shown when executing a server action:

parameter_label: label shown to the user in the input box.

name ("unique name": the name of the variable that collects the value of the input box to then be used in the server action code.

data_type: used to validate user input and convert the value accordingly.

There’s currently some redundancy in the validation logic across create and write, which I’ll refactor. I’ll also improve the field help text and docstrings to clarify all this further.

Lastly, I really appreciate your help and feedback. I’m still quite new to this and have a lot to learn.
I’ll go through the rest of your suggestions carefully and get back to you with my thoughts.

server_action_input_box/models/server_action_input_box.py
jesanmor: Thanks a lot for your observations and suggestions.

I’ve refactored the parsed_parameters method according to your feedback:

Moved the data_type_conversion dispatcher out of the loop and into a separate _get_data_type_conversion method to make it easier to extend.

Separated the raw value handling logic into a _get_raw_data method, to decouple it from the main loop and allow further customization per data type.

The try/except block now only catches ValueError, as you recommended.

Here’s the updated code:
```python
   def _get_data_type_conversion(self):
        return {
            "string": lambda s: s,
            "int": int,
            "float": float,
            "bool": lambda b: b,
        }
    
    def _get_raw_data(self, line, parameters):
        raw_value = parameters.get(line.name)
        return raw_value or 0 if line.data_type != "string" else raw_value

    def parsed_parameters(self, parameters):
        parsed_parameters_dict = {}
        data_type_conversion = self._get_data_type_conversion()

        for line in self.server_action_input_box_line_ids:
            raw_data = self._get_raw_data(line, parameters)
            try:
                convert_data = data_type_conversion[line.data_type]
                data = convert_data(raw_data)
            except ValueError as e:
                raise UserError(
                    _(
                        "The value '%(param_value)s' for '%(param_name)s' "
                        "is not of type '%(data_type)s'."
                    )
                    % {
                        "param_value": parameters.get(line.name),
                        "param_name": line.name,
                        "data_type": line.data_type,
                    }
                ) from e

            parsed_parameters_dict[line.name] = data

        return parsed_parameters_dict

```
Thanks again for your guidance—I'm reviewing more of your suggestions before pushing all the changes.

server_action_input_box/models/server_action_input_box.py
jesanmor: thanks for your comment.

You're absolutely right that in principle this could be handled by using ondelete="cascade" on the ir_action_server_id field. In fact, that was the initial approach. However, I later added protection mechanisms inside the ir.actions.server model itself, overriding both write and unlink methods to prevent external modifications or deletions of protected records linked to a server.action.input.box.

With these safeguards in place, it became necessary to manually control the deletion: first breaking the link and then explicitly removing the action server. At that point, the ondelete="cascade" behavior was no longer appropriate or effective—but I forgot to remove it when refactoring the logic.

I've now corrected that and removed the ondelete, since it's indeed unnecessary with the current deletion flow.

Thanks again for pointing that out—I'll continue reviewing the remaining suggestions before pushing the full update.

server_action_input_box/__manifest__.py
jesanmor: Thanks for the suggestion.

You're right — since this is a new module and could still evolve with community feedback and testing, it makes sense to mark it as `"Beta"` for now. I've updated the manifest accordingly.

server_action_input_box/__manifest__.py
jesanmor: I've fixed the implicit string concatenation and now it's properly written as a single string:

```python
"author": "Jesús Sánchez - jesanmor, Odoo Community Association (OCA)",

File: server_action_input_box/models/server_action_input_box.py (hunk 13 above)
Comment (jesanmor) [reply server_action_input_box/models/server_action_input_box.py:66]: Thanks for your initial feedback! I understand your point about the complexity — I tried to ensure flexibility and robustness, but I’m open to suggestions on simplifying it.
I’ll look forward to your further comments after your functional tests.

Comment (jesanmor) [reply server_action_input_box/models/server_action_input_box.py:66]: Hi gdgellatly,

I've updated the module's documentation, and the README.rst now includes the full content I originally prepared, with visual examples, screenshots, and a detailed explanation of how the module works—including the configuration and purpose of the "unique name" field.

The confusion earlier was due to the fact that the complete documentation was only in the manually written README.rst, and not split into the individual files under the readme/ folder. So when odoo-gen-readme generated the new README.rst, much of that content was missing, making it harder to understand the module’s logic.

Now everything is correctly structured and visible directly in the generated README.rst, making it much easier to understand the module's behavior and configuration at a glance.

You can test the application locally and use the examples included in the readme to verify it works correctly.

Thank you for your time.

File: server_action_input_box/models/server_action_input_box_line.py (hunk 1 above)
Comment (gdgellatly) [reply server_action_input_box/models/server_action_input_box_line.py:15]: Well you can still have a unique sql constraint, it would be something like (name, server_action_input_box_id). Personally I don't use explicit SQL constraints much, especially for something like this where performance is not a concern, I would use @api.constrains and make it a python constraint. As below, it is also better to have those isidentifier checks inside api.constrains rather than overriding create/write so you will finish with something like 

```python
@api.constrains("name, "server_action_input_box_id")
def _check_name(self):
  for action ins self:
     perform validation logic (isidentifier, uniqueness)- raise ValidationError if wrong.
```

jesanmor: Thanks a lot for the suggestion regarding the use of @api.constrains. I've applied it to validate both the uniqueness of name within the same server_action_input_box_id and to ensure it's a valid Python identifier.

This allowed me to remove the custom create and write methods entirely, simplifying the code significantly.

Here is the updated constraint method:
```python
@api.constrains("name", "server_action_input_box_id")
    def _check_name(self):
        for line in self:
            if not line.name.isidentifier():
                raise ValidationError(
                    _(
                        "'%s' is not a valid parameter name. \
                    Remove spaces and special characters or numbers at the beginning.",
                        line.name,
                    )
                )
            domain = [
                ("server_action_input_box_id", "=", line.server_action_input_box_id.id),
                ("name", "=", line.name),
                ("id", "!=", line.id),
            ]
            if self.search_count(domain):
                raise ValidationError(
                    _("the unique name '%s' cannot be duplicated", line.name)
                )
```
Thanks again for your guidance — it's been really helpful to improve the quality and maintainability of the code.

jesanmor:
Hi @gdgellatly,

Thank you for all the suggestions you made — they really helped me improve the quality of the code.
Have you had a chance to review it again? If there’s anything else that needs to be adjusted, I’m happy to work on it.
Or, if everything looks good to you, would you consider approving the pull request?

Thanks again for your time and guidance — it’s been very helpful, especially as this is my first contribution to the OCA. 😊

Best regards,
Jesús (jesanmor)

thomaspaulb:
@jesanmor It's sad to see a nice module not get reviewed, have you considered doing review swaps with other PR's? Review someone else's to get yours reviewed

jesanmor:
> @jesanmor It's sad to see a nice module not get reviewed, have you considered doing review swaps with other PR's? Review someone else's to get yours reviewed

@Thomas Paulb thanks for your kind words about the module.

And apologies for the late reply — I’ve just seen your message. A few months ago I started a new job and things have been quite busy, so I haven’t been very active on GitHub lately.

This contribution was mainly my way of giving something back to the community, since I’ve benefited a lot from OCA projects in the past. To be honest, this was my first contribution and I wasn’t very familiar with how the review process usually works.

I’m also very grateful to @gdgellatly for the time he took to review the code and for his valuable suggestions — they really helped improve the quality of the module.

Your suggestion about reviewing other PRs makes a lot of sense. I’ll definitely keep it in mind, although I’m not entirely sure where to start yet.

Thanks again for taking the time to comment and for the encouragement!

jesanmor:
I accidentally closed the PR while reviewing the conversation.
I've reopened it and will continue addressing the feedback.

PR #3301, opened 2025-05-27
[14.0][FIX]model_read_only: superuser exempt from the restriction

adasatorres:
Good afternoon, the superuser should be exempt from the superuser restriction even if they belong to a group that requires this restriction.


christian-ramos-tecnativa (approved):
LGTM, make sense

rrebollo (approved):
Code Review. LGTM! Although I think the PR description is a little bit confusing: "superuser... from superuser restriction". Don't you mean "...from model readonly restriction" instead?

antoniodavid (approved):
LGTM

edescalona (approved):
LGTM

PR #3303, opened 2025-06-03
[17.0][IMP] base_exception: display description

SilvioC2C:


ivantodorovich (approved):
Hey, thanks for the improvement!

Honestly I would drop the configurable system parameter for the sake of simplicity, and have this be the default behavior.
I don't think we'll break anything with it.

Not blocking though

hparfr (changes requested):
agree with @ivantodorovich do not add an ir.config_parameter for that

hparfr (approved):

PR #3305, opened 2025-06-10
[18.0][FIX] module_change_auto_install: AttributeError: 'NoneType' object has no attribute 'get'

Maarten-Van-Briel:
```
 File "/home/odoo/src/odoo/addons/base_import_module/models/ir_module.py", line 53, in _get_icon_image
    super()._get_icon_image()
  File "/home/odoo/src/odoo/odoo/addons/base/models/ir_module.py", line 268, in _get_icon_image
    countries = self.get_module_info(module.name).get('countries', [])
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'

### DIFF
--- a/module_change_auto_install/patch.py
+++ b/module_change_auto_install/patch.py
@@ -53,7 +53,7 @@ def _overload_load_manifest(module, mod_path=None):
         # Specific case where a previously available module marked as auto installable
         # is NOT available in the addons path.
         # In that case, avoid to crash when trying to get 'depends' key.
-        return
+        return {}
     auto_install = res.get("auto_install", False)
 
     modules_auto_install_enabled_dict = _get_modules_dict_auto_install_config(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: module_change_auto_install/patch.py [hunk 1]
```diff
@@ -53,7 +53,7 @@ def _overload_load_manifest(module, mod_path=None):
         # Specific case where a previously available module marked as auto installable
         # is NOT available in the addons path.
         # In that case, avoid to crash when trying to get 'depends' key.
-        return
+        return {}
```
legalsylvain: Better return super value, dont you think ?
```suggestion
        return res
```

rven:
I included a test scenario to check this see https://github.com/OCA/server-tools/pull/3307

rven:
@legalsylvain This one can be closed

PR #3307, opened 2025-06-11
[18.0] [FIX] module_change_auto_install

rven:
An error is raised when opening the Apps

AttributeError: 'NoneType' object has no attribute 'get'


legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #3312, opened 2025-06-17
[18.0][MIG] attachment_delete_restrict: Migration to 18.0

AungKoKoLin1997:
@qrtl QT5088


attachment_delete_restrict/tests/test_attachment_delete_restrict.py
```diff
@@ -150,7 +89,7 @@ def _allow_group(self):
         raise NotImplementedError
 
 
-class TestAttachmentDeleteGlobal(TestAttachmentDeleteAbstract, AbstractCase):
+class TestAttachmentDeleteGlobal(TestAttachmentDeleteAbstract):
```
yostashiro: What's the reason of changing the structure in this migration?

attachment_delete_restrict/views/res_config_setting_views.xml
```diff
@@ -0,0 +1,51 @@
...
+                    help="Select default level of delete restriction on attachments"
+                    title="Restrict Deletion on attachment"
+                    id="global_restrict_delete_attachment"
+                >
+                    <field
+                        name="global_restrict_delete_attachment"
+                        class="o_light_label"
+                    />
+                </setting>
+
```
yostashiro:
```suggestion
```

attachment_delete_restrict/tests/test_attachment_delete_restrict.py
```diff
@@ -0,0 +1,238 @@
...
+                "login": "test-owner@example.com",
+                "groups_id": [
+                    (
+                        6,
+                        0,
+                        (
+                            cls.env.ref("base.group_user").id,
+                            cls.env.ref("base.group_partner_manager").id,
+                        ),
+                    )
```
yostashiro: Let's switch to this style.
```suggestion
                    Command.set(
                        cls.env.ref("base.group_user").id,
                        cls.env.ref("base.group_partner_manager").id,
                    )
```

attachment_delete_restrict/tests/test_attachment_delete_restrict.py
AungKoKoLin1997: Initially I thought `AbstractCase` is not supported in odoo18. After I deep dive into odoo core code, I found there is an option to handle this by using `allow_inherited_tests_method = True`.
https://github.com/odoo/odoo/blob/389b355e7ec761fe8dc2908ac5aec540b0345c48/odoo/tests/loader.py#L19-L45

attachment_delete_restrict/models/ir_attachment.py
```diff
@@ -0,0 +1,80 @@
...
+        if self.env.su:
+            return super().unlink()
+        res_models = list(set(self.filtered("res_model").mapped("res_model")))
+        if res_models:
+            models = self.env["ir.model"].sudo().search([("model", "in", res_models)])
+            name2models = {m.model: m for m in models}
+            for rec in self:
+                if rec.res_model:
+                    model = name2models[rec.res_model]
+                    if model.restrict_delete_attachment == "default":
```
yostashiro: Can we make it like this in case Odoo fails to set `default` to models for whatever reason? It seems to have happened in runboat and I was wondering why the global setting wasn't working. We may backport this to earlier versions.
```suggestion
                    if model.restrict_delete_attachment == "default" not model.restrict_delete_attachment:
```

yostashiro (changes requested):

yostashiro (approved):
Code review. LGTM.

attachment_delete_restrict/tests/test_attachment_delete_restrict.py
```diff
@@ -0,0 +1,178 @@
...
+        self._set_restrict_mode("owner")
+        with self.assertRaises(ValidationError):
+            self.attachment.with_user(self.user).unlink()
+        self.attachment.with_user(self.user_owner).unlink()
+
+    def test_restrict_owner_admin(self):
+        self._set_restrict_mode("owner")
+        self.attachment.with_user(self.user_admin).unlink()
+
+    def test_restrict_owner_and_custom_user_forbiden(self):
```
kanda999:
```suggestion
    def test_restrict_owner_and_custom_user_forbidden(self):
```

kanda999 (commented):
Minor point

kanda999 (approved):
Functional and code review.

Sayaka-qrtl (commented):
It looks like there’s a difference between the five levels described in README.rst and what’s actually shown on the screen. Should we standardize them to use Owner + Custom as in the UI?

FrancoMaxime (approved):
LGTM: code review

PR #3313, opened 2025-06-20
[18.0][IMP] Add a hook for the format string in base_time_window

TDu:


henrybackman (approved):

base_time_window/models/time_window_mixin.py
```diff
@@ -84,10 +84,13 @@ def check_window_no_overlaps(self):
...
 
+    def _get_display_name_format_string(self):
+        return self.env._("From {start} to {end}")
+
     @api.depends("time_window_start", "time_window_end", "time_window_weekday_ids")
     def _compute_display_name(self):
         for record in self:
-            record.display_name = self.env._("{days}: From {start} to {end}").format(
+            record.display_name = self._get_display_name_format_string().format(
                 days=", ".join(record.time_window_weekday_ids.mapped("display_name")),
```
mmequignon, line 94: days isn't used in the format string.

mmequignon (changes requested):

base_time_window/models/time_window_mixin.py
TDu: Indeed and although the change allows to change the string it does not allow to change parameter of the formatted string.

TDu: Fixed

mmequignon (approved):
LG, even if a tiny witsy test would be super nice

PR #3315, opened 2025-06-23
[18.0][IMP] upgrade_analysis: tell user which selection keys were added/removed

hbrunn:
currently, we just print [all keys](https://github.com/OCA/OpenUpgrade/blob/18.0/openupgrade_scripts/scripts/base/18.0.1.3/upgrade_analysis.txt#L63), which is not helpful for small selections and unmanageable for big ones like the time zone selection further below in the linked file. This formats selections as "added: key1 removed: key2, key3"

### DIFF
--- a/upgrade_analysis/compare.py
+++ b/upgrade_analysis/compare.py
@@ -10,6 +10,7 @@
 
 import collections
 import copy
+from ast import literal_eval
 
 try:
     from odoo.addons.openupgrade_scripts import apriori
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Remember that this is not `[ADD]`, but `[IMP]`.

StefanRijnhart (approved):

upgrade_analysis/compare.py
```diff
@@ -128,6 +129,26 @@ def fieldprint(old, new, field, text, reprs):
             text += f" [{old['table']}]"
         if field == "relation":
             text += " [nothing to do]"
+        if field == "selection_keys":
+            old_selection_keys = old.get("selection_keys") or ""
+            new_selection_keys = new.get("selection_keys") or ""
+            try:
+                old_selection_keys = literal_eval(old_selection_keys)
+                new_selection_keys = literal_eval(new_selection_keys)
+            except Exception:  # pylint: disable=except-pass
```
legalsylvain, line 138: In witch situation it fails here ?

upgrade_analysis/compare.py
```diff
@@ -128,6 +129,26 @@ def fieldprint(old, new, field, text, reprs):
             text += f" [{old['table']}]"
         if field == "relation":
             text += " [nothing to do]"
+        if field == "selection_keys":
+            old_selection_keys = old.get("selection_keys") or ""
```
legalsylvain, line 133:
```suggestion
            old_selection_keys = old.get("selection_keys") or "[]"
```

This could avoid tests of type, dont you think ?

upgrade_analysis/compare.py
hbrunn: when the value is 'function' or whatever else that can't be parsed as python code

upgrade_analysis/compare.py
hbrunn: but we want to do nothing in that case

upgrade_analysis/compare.py
legalsylvain: Nice catch ! 

but in that case, changes can be made in the function (or the constant) accross the version, and analysis will not detect it, right ?
We could maybe add a text to say the Openupgrader "please, verify that the result of the function / constant didn't change".
Not sure, thinking aloud.

legalsylvain (approved):

upgrade_analysis/compare.py
hbrunn: that would add a lot of noise, and is very much out of scope for this PR. So I'd say: no

PR #3311, opened 2025-06-17
[17.0][IMP] tracking_manager: Properties compatibility

victoralmau:
Properties compatibility

ejemplo

@Tecnativa TT56736

PR #3316, opened 2025-06-24
[18.0][IMP] tracking_manager: Properties compatibility

victoralmau:
FWP from 17.0: https://github.com/OCA/server-tools/pull/3311

Properties compatibility

ejemplo

@Tecnativa TT56736

### DIFF
--- a/tracking_manager/README.rst
+++ b/tracking_manager/README.rst
@@ -46,24 +46,24 @@ order_line in a sale order).
 Usage
 =====
 
-- In setting > models: select a model
-- Check "Active" under Custom Tracking.
-- You have two options - 1) manually configure tracked fields one by
-  one, or 2) determine tracked fields based on a specific domain.
-- For 1) manually configure tracked fields one by one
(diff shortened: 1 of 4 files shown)


tracking_manager/README.rst
```diff
@@ -91,9 +91,14 @@ Authors
 Contributors
 ------------
 
+<<<<<<< HEAD
```
pedrobaeza: Solve this merge conflict

pedrobaeza (approved):
Code and tested in runboat. The only weird thing is that the text is "Properties" instead of "Property".

imagen

victoralmau:
> Code and tested in runboat. The only weird thing is that the text is "Properties" instead of "Property".
> 
> imagen

To clarify that IMO that is not an error, i.e., the name of the properties field is indicated (in `crm.lead`, the `lead_properties` field is called _Properties_), and then the name of the property.

carlos-lopez-tecnativa (approved):

PR #3234, opened 2025-03-21
[18.0][FIX]base_view_inheritance_extension: fix deprecation warning

BhaveshHeliconia:
```
DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
```

PR #3317, opened 2025-06-24
[17.0][FIX] base_view_inheritance_extension: multiple DepreciationWarning

vvrossem:
includes https://github.com/OCA/server-tools/pull/3234/commits/d54179e10ae2c2f344dc3378c2ffa1811bddf01c

### DIFF
--- a/base_view_inheritance_extension/models/ir_ui_view.py
+++ b/base_view_inheritance_extension/models/ir_ui_view.py
@@ -4,14 +4,7 @@
 # Copyright 2023 Tecnativa - Carlos Dauden
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 import ast
-import logging
 import re
-
-try:
-    import astor
-except ImportError as err:  # pragma: no cover
-    _logger = logging.getLogger(__name__)
-    _logger.debug(err)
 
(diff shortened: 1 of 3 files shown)


yankinmax (approved):

Camille0907 (approved):
Thank you

base_view_inheritance_extension/models/ir_ui_view.py
```diff
@@ -4,15 +4,8 @@
 # Copyright 2023 Tecnativa - Carlos Dauden
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 import ast
-import logging
 import re
 
-try:
-    import astor
```
StefanRijnhart, line 11: Does this mean that astor can be dropped as a python dependency in the manifest?

vvrossem: I believe so. This should be done in 18.0 too then

vvrossem: 18.0 done in https://github.com/OCA/server-tools/pull/3352

StefanRijnhart (approved):
Thanks!

/ocabot merge minor

PR #3150, opened 2024-12-18
[16.0][MIG] base_model_restrict_update: Migration to 16.0

nobuQuartile:
@qrtl QT4777

PR #3319, opened 2025-06-26
[18.0][MIG] base_model_restrict_update: Migration to 18.0

AungKoKoLin1997:
Migration based on https://github.com/OCA/server-tools/pull/3150

@qrtl QT5088


kanda999:
2025-06-30_12h08_45
When a READONLY user tries to open the product form view, an error occurs:
"You are only allowed to read this record. (product.category - write)"
This issue did not occur in previous versions.
It feels a bit inconvenient that READONLY users can no longer see the product form view.

AungKoKoLin1997:
> This issue did not occur in previous versions.
> It feels a bit inconvenient that READONLY users can no longer see the product form view.

@kanda999 You are right. There is some issue in some cases. I have to fix it.

Sayaka-qrtl (approved):
Functional test. LGTM!

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,70 @@
...
+        return bool(query_res) and (self.env.uid,) not in query_res
+
+    @api.model
+    def check(self, model, mode="read", raise_exception=True):
+        if self.env.su:
+            return True
+        res = super().check(model, mode=mode, raise_exception=raise_exception)
+        if mode != "read":
+            if self._test_readonly(model) or self._test_restrict_update(model):
+                return False
```
dreispt, line 73: Maybe also raise exception when `raise_exception` is True.

AungKoKoLin1997: @dreispt Thanks for the review. I’m seeing an inconsistency with my current logic: for read-only users, invoices still show an active chatter composer and allow posting without any error, whereas on Sales and Purchase documents the chatter is read-only. I haven’t identified the root cause or a fix yet.

Screenshot 2025-08-19 160843

Screenshot 2025-08-19 160920

AungKoKoLin1997: > Maybe also raise exception when `raise_exception` is True.

@dreispt Done!

Regarding my above comment, I found that is bug from odoo standard. So, I will ignore that in this module.

StefanRijnhart (commented):
You based this migration on the PR for 16.0, but you also submitted the migration to 17.0 in https://github.com/OCA/server-tools/pull/3348. Would you mind rebasing this one onto your 17.0 branch?

AungKoKoLin1997:
> You based this migration on the PR for 16.0, but you also submitted the migration to 17.0 in #3348. Would you mind rebasing this one onto your 17.0 branch?

@StefanRijnhart Done!

StefanRijnhart (approved):
Thanks!

base_model_restrict_update/models/ir_model_access.py
```diff
@@ -0,0 +1,72 @@
...
+    def check(self, model, mode="read", raise_exception=True):
+        if self.env.su:
+            return True
+        has_access = super().check(model, mode=mode, raise_exception=raise_exception)
+        if mode != "read":
+            if self._test_readonly(model) or self._test_restrict_update(model):
+                if raise_exception:
+                    raise self._make_access_error(model, mode) from None
+                return False
+        return has_access
```
yostashiro: I think we should stick with the original message instead of using _make_access_error() which doesn't provide the right context and may confuse the user.

```suggestion
        res = super().check(model, mode=mode, raise_exception=raise_exception)
        if mode == "read":
            return res
        if self._test_readonly(model) or self._test_restrict_update(model):
            if not raise_exception:
                return False
            raise AccessError(
                _(
                    "You are only allowed to read this record. (%(model)s - %(mode)s)"
                )
                % {"model": model, "mode": mode}
            )
        return res
```

yostashiro (changes requested):

yostashiro (approved):
Code review.

PR #3320, opened 2025-06-26
[ADD] fix_compute_trans_implied_groups

SirPyTech:
This fix should be in Odoo's `14.0` branch but it's no more maintained, so we have to create a module.
This isn't needed in `16.0` thanks to https://github.com/odoo/odoo/commit/34d6f87d5497182d8ce395a8c90277f5ddde8a95.

I haven't added a test in this module because the only way I know to reproduce this bug is to install the `sale` module and it makes no sense to add such a dependency in this module.

Let me know what you think!


GSLabIt (approved):
LGTM

HekkiMelody (approved):
LGTM.

Ideally this fix should be added to OCB, but I understand that not everyone uses it.

francesco-ooops:
@OCA/tools-maintainers wdyt?

thomaspaulb:
@SirPyTech For the unit test what you could do is to create a separate `_test` module which has the test and also depends on `sale`.

thomaspaulb:
Also, can you explain your reason for not using OCB?

SirPyTech:
Thanks @thomaspaulb for having a look!
I'll try to move this fix to OCB and then maybe I'll be able to add a proper test in `sale` module :rocket:
I didn't use OCB at first because I thought our customer wasn't using it.

SirPyTech:
Moved to https://github.com/OCA/OCB/pull/1306

SirPyTech:
Reopening to allow non-OCB users to enjoy this fix

SirPyTech:
> /ocabot merge nobump

Thanks @thomaspaulb for trying to merge!
The module `letsencrypt` is failing a test only in the OCB execution (see https://github.com/OCA/server-tools/actions/runs/21340400922/job/61418864454#step:8:1115), it is probably temporary, please try again :pray:

PR #3332, opened 2025-07-17
[18.0] upgrade_analysis: fix references to tree views

simahawk:
Fix Uncaught Promise > View types not defined tree found in act_window action undefined.


pedrobaeza (approved):
Yes, fast-tracking this as obvious:

/ocabot merge patch

PR #3335, opened 2025-07-22
[16.0][FIX] tracking_manager: fix specific case when the user want to track special M2M fields (attachment_ids)

acsonefho:
### How to reproduce:
**Install modules:**
- Accounting
- Document (Enterprise)

**Steps:**
- Enable tracking on `account.move` and track changes on `attachment_ids`.
- Go on Document and upload a new document: traceback :boom: 

(Maybe there are others ways to reproduce)

Fix: ensure the field exists on the model.


lmarion-source (approved):

rousseldenis (approved):
@acsonefho Maybe a test for that ?

PR #3339, opened 2025-08-06
[18][MIG] attachment_queue

florian-dacosta:
@Christian-RB @rrebollo @bealdav

PR #3340, opened 2025-08-06
[18][MIG] attachment_synchronize

florian-dacosta:
Depends on https://github.com/OCA/server-tools/pull/3339


StefanRijnhart (changes requested):
Please fix precommit and add a test dependency for the dependency on attachment_queue

florian-dacosta:
@StefanRijnhart The dependecy was merged FYI

attachment_synchronize/views/attachment_queue_views.xml
```diff
@@ -10,18 +10,19 @@
...
+                    required="file_type == 'export'"
+                    readonly="method_type == 'import'"
                 />
                 <field name="fs_storage_id" />
             </xpath>
         </field>
     </record>
 
-    <record id="view_attachment_queue_tree" model="ir.ui.view">
+    <record id="view_attachment_queue_list" model="ir.ui.view">
```
StefanRijnhart, line 21: The recommendation is not to rename view ids in this case. The reason for this is that Odoo does not handle renamed view renames very well during a migration. The migration will halt if the old view is not valid, which is likely the case here because it tries to fetch a `tree` node. Renaming is ok if you provide a pre-migration script renaming the ir.model.data record.

attachment_synchronize/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Attachment Synchronize",
-    "version": "16.0.1.0.1",
+    "version": "18.0.1.0.1",
```
StefanRijnhart: 18.0.1.0.0

attachment_synchronize/tests/common.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2020 Akretion (http://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import base64
+
+from odoo.tests.common import TransactionCase
+
+
+class SyncCommon(TransactionCase):
```
StefanRijnhart, line 10: Please refactor to SavepointCase and move to setUpClass what does not need to be in setUp.

StefanRijnhart (commented):
Can you rename your commit message to follow the `[MIG] $module: Migration to 18.0` convention?

attachment_synchronize/views/attachment_queue_views.xml
florian-dacosta: Hum, I did not know this one, thanks...
Here, I already rename the base view in the dependency `attachment_queue`, where we must have this migration issue.
I think I will go with the migration script then, in both module.

attachment_synchronize/tests/common.py
florian-dacosta: I've moved some stuff to setUpClass.
The class is still TransactionCase, as AFAIK, TransactionCase is the new SavepointCase.

attachment_synchronize/views/attachment_queue_views.xml
florian-dacosta: @StefanRijnhart I did the following test (with attachment_queue, that has been merged already but has the same change on list view xml id).
On a demo database with the module installed, I renamed the `ir.model.data` from view_attachment_queue_list to view_attachment_queue_tree
I update the arch_db (by sql query) to the old one with the `<tree>` tags.
Update the latest_version number to 17.0.1.0.0.
Update my the module attachment_queue, to simulate a migration.
=> The old view is deleted, the new one created and I do not have any crash.

Are you sure the migration script is necessary ? Of course, I may have missed something as I did not run a migration, just tried to simulate it.
If you are quite sure, I'll add  the scripts, in both modules, what do you think ?

StefanRijnhart: It's not worth it to go into individual cases. I have run several times into this issue in Odoo 18 whenever views are renamed. Interestingly, the guidelines mention not to change view ids in order to prevent disruption when migrating depending modules:
```
Replace tree view type by list. This means everywhere in Python, JS or XML where there's a mentioning of this type, but keep XML-IDs that contain tree in its name for avoiding subsequent changes in dependent modules.
```
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

I think there are several ways out of here:
* the 'beta' status of the module warrants a slightly laxer application of the rules.
* you yourself are this module maintainer. Maybe you can agree with one or more other maintainers of this module (there are 4) to rename the ids.
* if you do this in combination with a migration script, I will gladly approve. There are not going to be a lot of modules inheriting from this one.
* If you don't add a migration script on account of your manual testing, and other module maintainers agree, I will refrain from further review but not block it. Other reviewers may be more strict wrt. to the application of the guidelines.

attachment_synchronize/tests/common.py
StefanRijnhart: Ah yes of course! Thanks.

attachment_synchronize/views/attachment_queue_views.xml
florian-dacosta: @StefanRijnhart Thanks for your opinion and for pointing the guidline.
It was a real/innocent question on my side, I am discovering this issue, so I trust you know better than me indeed.
I'll add the migration scripts (seems better than revert the xml_id because of the merged dependency `attachment_queue`)

florian-dacosta: I have added a migration script, it should be fine now.
Thanks again for your help on the matter.

StefanRijnhart (approved):
Thanks for the updates!

bguillot (approved):

PR #3342, opened 2025-08-12
[18.0][FIX] sentry: event dropped if include_context key not present

StefanRijnhart:
In our (Odoo.sh) setup, we were missing almost all events after migrating to Odoo 18. With this fix, the events were back.

Fixes

```
  File "/home/odoo/.local/lib/python3.12/site-packages/sentry_sdk/client.py", line 595, in _prepare_event
    new_event = before_send(event, hint or {})
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/odoo/src/user/oca-server-tools/sentry/hooks.py", line 66, in before_send
    if event.setdefault("tags", {})["include_context"]:
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
KeyError: 'include_context'
2025-07-08 14:45:45,907 14296 INFO quatra-18-0-development-21441249 sentry_sdk.errors: before send dropped event
```

NB. This error is only logged when running Sentry in debug mode, otherwise events are dropped silently.

### DIFF
--- a/database_size/README.rst
+++ b/database_size/README.rst
@@ -65,25 +65,26 @@ Size.
 
 The data that is gathered and that is displayed are:
 
-- Model Name - The name of the model to which the data is related
-- Estimated Rows - The number of estimated rows according to the
-  Postgresql query planner. For performance reasons, taking the data
-  from the planner is preferred over doing an actual count, although the
-  results may be imprecise.
-- Bare Table Size - The disk usage of the model table without indexes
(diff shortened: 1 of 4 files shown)


StefanRijnhart:
Tests are fixed in https://github.com/OCA/server-tools/pull/3343

PR #3346, opened 2025-08-13
[14.0][IMP] base_exception: Placeholders for exception rule description

raumschmiede-joshuaL:
Hello.

I changed the exception rules to use Jinja in exception rule descriptions.
For that I changed a bit more in the module so that for exceptions that were detected on underlying records (like in `sale_exception`) the text is still correctly added to the exceptions summary on the parent records. I adjusted the tests to cover this.

Useful for us as we have a lot of sales exceptions. Some exceptions check whether any child of a sold BOM product contains valid data. For BOMs with lots of children or BOMs in BOMs it is not helpful enough for the user to see that the 1st SO line has the exception "The product has no weight set".
With my change you can use "The product {{obj.default_code}} has no weight set" as description.

My changes do not break modules that depend on `base_exception`. I tested it on `sale_exception`. The only thing that wouldn't work in `sale_exception` is if you use the placeholders that the exceptions summary for the SO line wouldn't be correctly computed as `sale_exception` defines its own fields on the SO line.
**EDIT 2**: With the latest changes `sale_exception` will break,

~I wasn't sure whether it is better to have Jinja or a separate code field as placeholders. I like it more to write code, normal non-dev users would rather use Jinja, like in mail templates. That's why I added both to make everyone happy.
But for the Jinja stuff I added a dependency to "mail" as this provides already a function that parses Jinja. I saw in another OCA module that it was used like that. I did the same to keep it simple and consistent.
If this new dependency is not good, I could move this into a new module.~

**EDIT 2**: I removed the description code placeholders. Can't say whether they are good for the OCA module. I will add this in a separate module in our repo.

I checked which modules depend on `base_exception`. The ones I found I already adjusted to work with the new field to define which fields contain sub-records for which exceptions must be checked.
So far I found 5 modules. I already changed the code in them, but I would create the PRs to change them later after the discussion here.
**EDIT**: I already added the PR for `sales_exception`: https://github.com/OCA/sale-workflow/pull/3861. Can be used as example or for testing if you want. And for that more changes are needed. 

Everything is still more a draft. I would like to hear your opinion on the changes. Also about naming new defs and fields.

Screenshots:

Jinja in exception description:
image
**EDIT 2** Is now `{{...}}` and not `${...}` to avoid new dependency and to have the same as in later Odoo versions.

Exceptions in sales order:
image

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -1,5 +1,6 @@
 # Copyright 2011 Raphaël Valyi, Renato Lima, Guewen Baconnier, Sodexis
 # Copyright 2017 Akretion (http://www.akretion.com)
+# Copyright 2025 Raumschmiede GmbH
 # Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
 # Copyright 2020 Hibou Corp.
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu)
@@ -90,6 +91,22 @@ class BaseExceptionModel(models.AbstractModel):
     @api.model
     def _get_popup_action(self):
         return self.env.ref("base_exception.action_exception_rule_confirm")
+
+    def _add_detected_exceptions_to_self(self):
+        return self != self._get_main_records()
+
+    def _detect_exceptions(self, rule):
+        records = super()._detect_exceptions(rule)
+        # If _get_main_records returns self, it adds the exceptions to itself in
+        # detect_exceptions. If _get_main_records does not return self, it adds the
+        # exceptions here
+        if not self._add_detected_exceptions_to_self():
+            return records
+
+        (self - records).exception_ids = [(3, rule.id)]
+        records.exception_ids = [(4, rule.id)]
+
+        return records
 
     def _check_exception(self):
         """Check exceptions
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -52,28 +52,112 @@ class BaseExceptionModel(models.AbstractModel):
             else:
                 rec.main_exception_id = False
 
-    @api.depends("exception_ids", "ignore_exception")
+    @api.depends(
+        "exception_ids",
+        "exception_ids.name",
+        "exception_ids.description",
+        "exception_ids.description_text_type",
+        "exception_ids.is_blocking",
+        "ignore_exception",
+    )
     def _compute_exceptions_summary(self):
         for rec in self:
             if rec.exception_ids and not rec.ignore_exception:
-                rec.exceptions_summary = "<ul>%s</ul>" % "".join(
-                    [
-                        "<li>%s: <i>%s</i> <b>%s<b></li>"
-                        % tuple(
-                            map(
-                                html.escape,
-                                (
-                                    e.name,
-                                    e.description or "",
-                                    _("(Blocking exception)") if e.is_blocking else "",
-                                ),
-                            )
-                        )
-                        for e in rec.exception_ids
-                    ]
-                )
+                rec.exceptions_summary = rec._get_exceptions_summary()
             else:
                 rec.exceptions_summary = False
+
+    def _get_exceptions_summary(self):
+        self.ensure_one()
+
+        summaries = []
+        for exception in self.exception_ids:
+            summaries += self._get_exceptions_summaries_by_exception(exception)
+
+        return self._get_pretty_exceptions_summary_from_summaries(summaries)
+
+    def _get_exceptions_summaries_by_exception(self, exception):
+        self.ensure_one()
+
+        # True if the description is e.g. empty or flagged as plain description. We
+        # expect that no record is used to render the description, that's why self can
+        # be used to get the description. Even if exception.model is not self._name
+        if not exception.description_needs_rendering():
+            return [self._get_pretty_summary_for_exception(exception)]
+
+        # If the exception needs to be rendered, we have to get the records on which
+        # the exception was detected on to use them for rendering. If the exception
+        # model is self._name or the exception was detected on a sub-record whose model
+        # inherits from base.exception.method, records will the same as self.
+        records = self._get_records_for_description_rendering(exception)
+
+        summaries = []
+        for record in records:
+            # Even with rendering it can happen that multiple records return the same
+            # rendered description, e.g. if the description is plain text but its type
+            # is set to Jinja. The same description will be added multiple times to the
+            # final summary. With that the user knows that multiple sub-records have
+            # the same exception assigned and were rendered
+            summary = record._get_pretty_summary_for_exception(exception)
+            summaries.append(summary)
+
+        return summaries
+
+    def _get_records_for_description_rendering(self, exception):
+        self.ensure_one()
+
+        if exception.model == self._name:
+            # Exception needs the current record to render
+            return self
+
+        sub_record_ids = []
+        # Collect all sub-records that have the exception
+        for field in self._get_sub_exception_field_names():
+            sub_records = self.mapped(field)
+
+            if exception.model != sub_records._name:
+                continue
+
+            for sub_record in sub_records:
+                # For models that inherit from base.exception.method this is always
+                # False. Only for base.exception models it can be True
+                if sub_record._has_exception_rule_assigned(exception):
+                    sub_record_ids.append(sub_record.id)
+
+        if sub_record_ids:
+            return self.env[exception.model].browse(sub_record_ids)
+
+        # If there are no sub-records that have the exception assigned, possible
+        # because their model inherits from base.exception.method, self is returned
+        # to use it to render the description. This means it is not possible to
+        # use the same model in the description syntax as the model on which the
+        # exception was detected on
+        return self
+
+    def _get_pretty_summary_for_exception(self, exception):
+        self.ensure_one()
+
+        args = list(
+            map(
+                html.escape,
+                (
+                    exception.name,
+                    exception.render_description(self),
+                ),
+            )
+        )
+
+        if exception.is_blocking:
+            args.append(_(" <b>(Blocking exception)</b>"))
+        else:
+            args.append("")
+
+        return "%s: <i>%s</i>%s" % tuple(args)
+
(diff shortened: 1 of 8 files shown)


raumschmiede-joshuaL:
@mt-software-de, @jbaudoux, @i-vyshnevska

base_exception/models/base_exception.py
```diff
@@ -21,7 +22,43 @@ class ExceptionRule(models.Model):
...
+        "Jinja: 'Product with SKU ${record.default_code} has no costs set'\n\n"
+        "Variables: 'Product with SKU {var1} has no costs set'.\n"
+        "In code: 'var1 = record.default_code'",
+    )
+    description_template_type = fields.Selection(
+        [
+            ("plain", "Text"),
+            ("jinja", "Jinja"),
+            ("variables", "Variables"),
+        ],
```
mt-software-de: Why do you need to add 3 different ways of rendering?
You could just rely on qweb, this is already usable with the current dependencies of this addon. 
Also I guess it would be fine to always try to render the description, IMO there is no need to declare it in a separate field. Just allow qweb syntax and render the description always.

mt-software-de: Also Jinja2 is not really in use in newer Odoo Versions anymore.

base_exception/models/base_exception.py
```diff
@@ -93,6 +221,13 @@ def _get_main_records(self):
     def _reverse_field(self):
         raise NotImplementedError()
 
+    def _get_sub_exception_field_names(self):
+        """
+        Used in case exceptions need to be checked on underlying records and the
+        detected exceptions need to be added to the parent record.
+        """
+        return []
+
```
mt-software-de: Could you move those changes into an own commit?

base_exception/tests/purchase_test.py
```diff
@@ -12,16 +13,27 @@ class ExceptionRule(models.Model):
...
     )
-    test_purchase_ids = fields.Many2many("base.exception.test.purchase")
 
 
 class PurchaseTest(models.Model):
     _inherit = "base.exception"
-    _name = "base.exception.test.purchase"
-    _description = "Base Exception Test Model"
+    _name = "test.purchase"
+    _description = "Test Model"
```
mt-software-de: Is the change of `_name` necessary? This produces noise in the PR.

base_exception/tests/test_base_exception.py
```diff
@@ -1,61 +1,26 @@
...
+from .common import TestBaseExceptionCommon
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.loader.restore_registry()
-        super().tearDownClass()
 
-    def test_valid(self):
+class TestBaseException(TestBaseExceptionCommon):
+    def test_01_valid(self):
```
mt-software-de: Is changing the names of the test methods necessary? This produces noise in the PR.

mt-software-de (changes requested):
Thx. This sounds like a good extension. 

But there are a few things in need to mention. 

Could you please ensure to create separate commits for each functional change.

Also please try not to refactor the tests in the same commit as your functional changes. This creates a little bit of noise in this PR. Either remove those changes or create a separate commit for it.

base_exception/models/base_exception.py
raumschmiede-joshuaL: > Why do you need to add 3 different ways of rendering?

For normal users Jinja is better. For devs that create the descriptions code is better. And plain just as default.



> always try to render the description

Depends. If we decide at the end to only have one for placeholders, then it is ok.



> Just allow qweb syntax and render the description always

QWeb is more for bigger templates to render. Feels weird to use this. More complex syntax for a small use-case.
But I will test this locally

> Also Jinja2 is not really in use in newer Odoo Versions anymore.

Don't know how many non-dev users would use Jinja. If the others are fine with it, I can remove it. For me ok

base_exception/tests/purchase_test.py
raumschmiede-joshuaL: Not for `base.exception.test.purchase` but for the models below it. `base.exception.test.purchase.line.reverse.parent` is too long, when I started the unit test it said a table name for a relation was too long. That's why I shortened the names

base_exception/tests/test_base_exception.py
raumschmiede-joshuaL: Yes, to determine the order in which the tests are running. When debugging a test which runs as one of the last it is easier to deactivate the tests above that one because you know that they have a smaller number. Without numbers you have to check every test whether it comes in alphabetical order before or after your test you want to debug

base_exception/models/base_exception.py
raumschmiede-joshuaL: @mt-software-de, I implemented QWeb in a new commit for now. Looks like this: 
image

It feels more complicated to write this syntax. Jinja is almost the same, just with less wrapping around the actual code.
And for more complex code a separate code field is better.

Better to remove Jinja and Qweb

base_exception/tests/test_base_exception.py
mt-software-de: There is no reason to do it, you could use test-tags to test only a specific unittest.

mt-software-de:
IMO you should split this PR into 2.

One which is doing your changes for the `_get_sub_exception_field_names`, there you can also Backport https://github.com/OCA/server-tools/pull/2670/commits/b5099bc51ea53592cefd5162b6ce72cc5e6edeee from https://github.com/OCA/server-tools/pull/2670 because there are changes on the `_reverse_field`

And a second PR (this one)
Which just your updates for placeholders in description.
Here i would recommend only to use one method for rendering, since the inline_qweb is not available in v14 stick to `Jinja` without using the dependency to `mail` this would ease the FWP to newer versions.

hparfr (changes requested):
Hi @raumschmiede-joshuaL 

Thanks for your work on this module.

This module is quite sensible from a performance perspective; I fear the jinja rendering will put a penaltly. 

You may add this feature in a separate module or behind a feature flag.

base_exception/models/base_exception.py
raumschmiede-joshuaL: Removed QWeb and description code. Only Jinja and the description is always rendered

base_exception/models/base_exception.py
raumschmiede-joshuaL: Moved into PR https://github.com/OCA/server-tools/pull/3354

base_exception/tests/purchase_test.py
raumschmiede-joshuaL: Because of the backport of commit https://github.com/OCA/server-tools/pull/2670/commits/b5099bc51ea53592cefd5162b6ce72cc5e6edeee this change is not needed anymore. I removed it

base_exception/tests/test_base_exception.py
raumschmiede-joshuaL: Clarified it. Moved this change into PR https://github.com/OCA/server-tools/pull/3354

raumschmiede-joshuaL:
@hparfr, thanks for your reply.

A flag like in the screenshot of the PR description? To choose between plain text description and Jinja. I can add this back

raumschmiede-joshuaL:
@hparfr, I can add it behind feature flag like in the 1st screenshot:

image


Can you check PR https://github.com/OCA/server-tools/pull/3354? Maybe that one can be merged 1st

base_exception/models/base_exception.py
raumschmiede-joshuaL: In the forward port of this PR for Odoo 16.0 I can replace Jinja with inline Qweb. Will check it after this here was merged

mt-software-de (changes requested):
IMO. We should get rid of the two different behaviors. That you can detect exceptions on subrecords and store it there and detect it on subrecords and not store it there. 

By storing the exceptions always where they are really detected, all of your changes could be simplified. 

Also IMO it would make sense to store the summary when the exceptions are detected.
From a performance aspect it should not make a huge difference for plain exceptions.

raumschmiede-joshuaL:
> IMO. We should get rid of the two different behaviors. That you can detect exceptions on subrecords and store it there and detect it on subrecords and not store it there.
> By storing the exceptions always where they are really detected, all of your changes could be simplified.


I think as there are 2 classes, `base.exception.method` and `base.exception`, 2 different behaviours are fine. For the case that you do not want to have `exception_ids` on a model where you will never use it.

 
> Also IMO it would make sense to store the summary when the exceptions are detected. From a performance aspect it should not make a huge difference for plain exceptions.

What is the advantage of saving the summary? The performance drain is probably not much but still a disadvantage for the current state. For my placeholders it is a disadvantage as data can change on a record which is used as placeholder value.
For example if `product_id.default_code` of a sales order line is changed but in the summary the old SKU is still shown until a recompute is triggered

base_exception/models/base_exception.py
```diff
@@ -91,6 +176,28 @@ def _popup_exceptions(self):
...
+    def _detect_exceptions(self, rule):
+        records = super()._detect_exceptions(rule)
+        # If _get_main_records returns self, it adds the exceptions to itself in
+        # detect_exceptions. If _get_main_records does not return self, it adds the
+        # exceptions here
+        if not self._add_detected_exceptions_to_self():
+            return records
+
+        (self - records).exception_ids = [(3, rule.id)]
+        records.exception_ids = [(4, rule.id)]
```
vuwnevska, line 197: IMO this is not obvious behaviour, and probably the deleted comment from detect_exceptions is valid here

raumschmiede-joshuaL: If you have a field `exception_ids` on a model, you would expect that it will be set when exceptions are detected. Otherwise the field would be useless. If you do not want this behaviour, you would probably not want to have `exception_ids` on the record. For that you can inherit from `base.exception.method` instead of `base.exception`.


The way how the exceptions are assigned to the sub-records is the same as in `sale_exception`. From there I copied it. There the comment says that the flush optimizes it, In general I think the lock was maybe only possible because of the reverse field as a lot of writes could happen on the exception field, e.g. `sale_ids`. But this field is gone, so I guess there is no lock anymore

PR #3349, opened 2025-08-18
[14.0][IMP] base_exception: New Exception Type "By python code batch"

raumschmiede-joshuaL:
Hello.

If a code exception is checked on a lot of records, the performance can be slow as the exception's code must be evaled for each record.
For better performance I added `by_py_code_batch` so that every record in `self` can be checked.

Almost the same as the existing type `by_method`, but now with front end code

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -31,12 +31,14 @@ class ExceptionRule(models.Model):
         selection=[
             ("by_domain", "By domain"),
             ("by_py_code", "By python code"),
+            ("by_py_code_batch", "By python code batch"),
             ("by_method", "By method"),
         ],
         string="Exception Type",
         required=True,
         default="by_py_code",
-        help="By python code: allow to define any arbitrary check\n"
+        help="By python code: allow to define any arbitrary check on a single record\n"
+        "By python code batch: allow to define any arbitrary check on all records\n"
         "By domain: limited to a selection by an odoo domain:\n"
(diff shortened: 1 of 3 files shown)


raumschmiede-joshuaL:
@mt-software-de and @i-vyshnevska

base_exception/models/base_exception.py
```diff
@@ -31,12 +31,14 @@ class ExceptionRule(models.Model):
         selection=[
             ("by_domain", "By domain"),
             ("by_py_code", "By python code"),
+            ("by_py_code_batch", "By python code batch"),
```
mt-software-de, line 34: IMO. This PR should be not necessary. 

I don't think there should be a second exception type `by_py_code`.
I think you can try to Backport https://github.com/OCA/server-tools/pull/2670. 
If this is not enough for you, you should maybe think about refactoring your exception py code and maybe move some of your exception code into real code and maybe store some information in db. 

The other possibility would to change the evaluation of the type `by_py_code` and except not a `True` or `False` for a single record. Instead excepting a record set as you tried to to it with your new type. But i don't think that there will be consent (including me). Because then it would be necessary to change all existing rules.

raumschmiede-joshuaL: Another thing would be to add a flag on the exception rule to say whether all or just one record is passed to the code

raumschmiede-joshuaL: @rousseldenis and @lmignon

PR #3350, opened 2025-08-19
[16.0][ADD] export_html_text

AungKoKoLin1997:
This module allows exporting HTML fields as plain text.

@qrtl QT5650


export_html_text/models/html_field_text_config.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HtmlFieldTextConfig(models.Model):
```
yostashiro: In stead of this, I thought it would make a better design if we had an "Export HTML in Text" checkbox in the export dialog. Can you please consider this approach?

image

AungKoKoLin1997: Updated.

Sayaka-qrtl (approved):
functional test. LGTM!

export_html_text/fields.py
```diff
@@ -0,0 +1,21 @@
...
+    if (
+        self.type == "html"
+        and content
+        and record.env.context.get("export_html_as_text")
+    ):
+        return html2plaintext(content)
+    return content
+
+
+Field.convert_to_export = _convert_to_export_patched
```
yostashiro, line 21: Could we extend export_data() and apply a post-process conversion instead of monkey-patching this method? The steps will be a bit more cumbersome but safer.

export_html_text/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* `Quartile <https://www.quartile.co>`_
+
+  * Aung Ko Ko Lin
+
```
yostashiro, line 4:
```suggestion
```

export_html_text/static/src/js/export_html_to_text.esm.js
```diff
@@ -0,0 +1,20 @@
+/** @odoo-module **/
+import {patch} from "@web/core/utils/patch";
+import {ListController} from "@web/views/list/list_controller";
+
+patch(ListController.prototype, "export_html_as_text.controller_flag", {
```
yostashiro, line 5: Can you consider patching ExportDataDialog instead?

yostashiro:
@AungKoKoLin1997 Can we rename this module to `web_export_html_as_text` and create a PR in https://github.com/OCA/web instead?

export_html_text/static/src/js/export_html_to_text.esm.js
AungKoKoLin1997: Not by itself. ExportDataDialog doesn’t build the export RPC context—ListController does—so a dialog-only patch won’t reach Python. We’d still need to pass a flag to the ListController and set the context there. Given that, I’ll keep the ListController patch only; it’s simpler and works.

AungKoKoLin1997:
Proposed in https://github.com/OCA/web/pull/3259 and closed current PR.

PR #3352, opened 2025-08-20
[18.0][IMP] base_view_inheritance_extension: drop unused dependency

vvrossem:


StefanRijnhart (approved):

alexis-via (approved):

pedrobaeza (changes requested):
Please do only one commit, and specify in it the reason to remove it. And the tag shoudn't be `[REM]`, as this is for removing a module (not a part of code): https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

StefanRijnhart:
@pedrobaeza Strictly speaking, none of the [tags](https://www.odoo.com/documentation/16.0/contributing/development/git_guidelines.html#tag-and-module-name) apply. It's not a [FIX] because there is no bug that is fixed, it's not a [REF] because it's not a 'heavy rewrite' etc. What tag do you suggest?

pedrobaeza:
I'm not sure about the goal of the PR - that's why I ask -, but I would guess that it's a dependency that is not used. If so, the tag is IMP, as this "improves" the module, not requiring the dependency, although some would think that is FIX, because it's "fixing" an unneeded dependency. Anyway, definitely is not REM, thought for removing an entire module, the same as ADD for adding an entire module. A lot of people makes a parallelism between this and adding / removing lines of code, but that's the essence of each commit, so no need to reflect it in the commit message. 

And for sure, there shouldn't be two commits here.

StefanRijnhart:
Thanks, I agree that [IMP] is an option here, but note that [REM] is not reserved for whole modules like [ADD] is.
```
[ADD] for adding new modules;
[REM] for removing resources: removing dead code, removing views, removing modules, …;
```

pedrobaeza:
Oh, I see that distinction on the Odoo guidelines, and although arguable (why removing views?), it's like that and we should stick to it for consistency. Not being this thing dead code, we are still saved of that tag.

And the most important thing here is to **not add noise at commit level** (with 2 commits one being the pre-commit result of the other - note that the README artifact is something that I have been discussing with @sbidoul that the bot applies a different template on merging than the one in this repository -), and to **include in the commit message** relevant information: in this case, why we can remove this.

StefanRijnhart:
So squash commits and then a commit message saying `[IMP] base_view_inheritance_extension: drop unused dependency`?

pedrobaeza:
Yes, for example...

StefanRijnhart:
@vvrossem Can you make these changes?

pedrobaeza (approved):
/ocabot merge patch

PR #3353, opened 2025-08-20
[18.0][ADD] module_version_auto_upgrade

ryanc-me:
Make Odoo upgrade modules automatically when their version number changes, like Odoo.sh does.

Useful when adding fields to `res.partner`/`res.users`, which often breaks the UI (and as-of Odoo 18.0, keeping the module page open before restart no longer works)


ryanc-me:
@pedrobaeza just wondering if you have any guidance re testing for this module? the interesting part happens during Odoo (re)start, so I'm not sure if the regular self-testing approach will work

pedrobaeza:
@ryanc-me I don't know how to test that at Odoo level, but I think you shouldn't reinvent the wheel, as this already exists through `click-odoo-update`, and the check is done through a checksum, not module version change (which is not the only indicator that something has changed, i. e., translations).

ryanc-me:
This was a tool we built internally - specifically to solve the `res.users`/`res.partner` issue. I think the main difference is, we can choose when to trigger a module upgrade automatically, by either updating/not the version. It also works as a drop-in (no external tools needed).

I do see your point though - maybe I can leave the PR open for some other feedback?

pedrobaeza:
Yes. You also have https://github.com/OCA/server-tools/tree/18.0/module_auto_update for similar features.

ryanc-me:
After looking at those more, I think they solve a different issue.

Imagine the situation is:
- Odoo hosted on AWS Fargate or s/o (e.g., developer doesn't have shell access, Postgres is locked-down, etc)
- Developer has added a field to `res.users`, so after a restart, the Odoo UI is broken (can't access apps page)

Neither `click-odoo-upgrade` nor `module_auto_update` helps in that case - they require container shell access, or require that the UI is working.

With this module, the dev can bump the module version, then the app is auto-upgraded, no other interaction/access needed

pedrobaeza:
OK, as your prefer. I think the deployment strategy should include the auto-update, not to let Odoo to do that task, but anyway, if you explain the use case in the module, it's OK.

module_version_auto_upgrade/monkeypatch.py
```diff
@@ -0,0 +1,70 @@
...
+    # load_modules is a good spot to hook, because modules __init__.py are called
+    # before this, but the registry is not yet fully loaded (so we avoid some weird
+    # issues with .mapped() and computed fields).
+    modules_to_upgrade = get_upgradeable_modules(registry)
+    if modules_to_upgrade:
+        tools.config["update"].update({k: 1 for k in modules_to_upgrade})
+        update_module = modules_to_upgrade or update_module
+    return _load_modules(
+        registry, force_demo=force_demo, status=status, update_module=update_module
+    )
```
fkantelberg, line 23: I'm currently not sure if your code has to run before the original code or if would be easier if your code runs afterwards because currently you are doing things like graph, check for db initialization twice.

fkantelberg (commented):
I would agree with Pedro that it could be part of `module_auto_upgrade` and use the logic there. Currently you need a shell but this is the reason for your hook and it could integrate quite nicely if you use a config flag to disable/enable the feature.


```python
from odoo import tools


_load_modules = loading.load_modules


def load_modules(registry, force_demo=False, status=None, update_module=False):
    if not tools.config.get("auto_upgrade_modules"):
        # Do nothing
        ...

    # Do the automatic upgrade here
    ...
```

Because it would be part of `module_auto_update` the code to determine the changes could be just:

```python
env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
to_upgrade = env["ir.module.module"]._get_modules_with_changed_checksum()
return to_upgrade.mapped("name")
```

If you put your custom code behind the original `load_modules` it can also be used directly:
```python
env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
env["ir.module.module"].upgrade_changed_checksum()
```

What do you think?

ryanc-me:
@fkantelberg Combining with `module_auto_upgrade` sounds like a good idea. If nothing else, it will be a little less confusing for people ("Does the 'auto_upgrade' module automatically upgrade? Or is it the other one?").

I also like the idea of using the checksum to find which apps update.

A few other considerations:

- At my day job, we primarily need this to deal with `res_{company,users}` field-additions, which AFAIK can only be handled with an upgrade on the command line.
- So, version-based approach would work better for us
- That is also the logic at Odoo.sh, so might be more _comfortable_ for some people

So maybe we can offer both options - something like `auto_upgrade_modules` with values of `checksum` or `version`?

module_version_auto_upgrade/monkeypatch.py
ryanc-me: Agreed, it would be great to make the hook a bit cleaner.

It's been a while, but from memory, I needed to hook this specific spot because here:

1. The env is setup enough that we can get a registry/access models, BUT
2. The `res_company` and `res_users` models haven't been read from yet

fkantelberg:
Sounds like a good common ground :)

PR #3355, opened 2025-08-26
[18.0][FIX] base_name_search_improved: prevent recursion on relational fields

lef-adhoc:
The module's enhanced search functionality can cause a RecursionError when a relational field is used in the smart search configuration. The Odoo ORM triggers this by recursively calling name_search, which results in an infinite loop.

This commit breaks the loop by using a context key ('skip_smart_name_search'). During a recursive call, the module now falls back to the standard search behavior for that specific call, ensuring system stability while keeping the enhanced functionality active for the top-level search.


baf-adhoc (approved):
Functional review ok

PR #3357, opened 2025-08-29
[18.0][FIX] base_name_search_improved: restore v16 behavior in _search_smart_search

lef-adhoc:
In v18 `_search_smart_search` was delegating to `name_search`. This introduces an artificial limit (100 records) before other domain filters are applied, leading to incomplete results.

This PR restores the original behavior from v16:
- Build the domain directly in `_search_smart_search`.

This makes smart search consistent again with standard search and avoids missing records.

### DIFF
--- a/base_name_search_improved/models/ir_model.py
+++ b/base_name_search_improved/models/ir_model.py
@@ -82,36 +82,19 @@ class Base(models.AbstractModel):
     @api.model
     def _search_smart_search(self, operator, value):
         if value and operator in ALLOWED_OPS:
-            matching_records = self.with_context(
-                force_smart_name_search=True
-            ).name_search(name=value, operator=operator)
-            if matching_records:
-                record_ids = [record[0] for record in matching_records]
-                return [("id", "in", record_ids)]
-        return [("id", "=", False)]
-
-    @api.model
-    def _search_display_name(self, operator, value):
-        domain = super()._search_display_name(operator, value)
-        if self.env.context.get(
-            "force_smart_name_search", False
-        ) or _get_use_smart_name_search(self.sudo()):
             all_names = _get_rec_names(self.sudo())
-            additional_domain = _get_name_search_domain(self.sudo())
-
+            domain = _get_name_search_domain(self.sudo())
             for word in value.split():
                 word_domain = []
                 for rec_name in all_names:
                     word_domain = (
                         word_domain and ["|"] + word_domain or word_domain
                     ) + [(rec_name, operator, word)]
-                additional_domain = (
-                    additional_domain and ["&"] + additional_domain or additional_domain
-                ) + word_domain
-
-            return expression.OR([additional_domain, domain])
-
-        return domain
+                domain = (
+                    expression.AND([domain, word_domain]) if domain else word_domain
+                )
+            return domain
+        return []
 
     @api.model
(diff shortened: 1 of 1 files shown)


baf-adhoc (approved):
LGTM. Functional review ok

lef-adhoc:
@dreispt @kittiu @Rad0van Hi, could you check this?

lef-adhoc:
Hi @pedrobaeza, do you happen to know who might see this?

base_name_search_improved/models/ir_model.py
```diff
@@ -82,36 +82,17 @@ def _compute_smart_search(self):
...
                         word_domain and ["|"] + word_domain or word_domain
                     ) + [(rec_name, operator, word)]
-                additional_domain = (
-                    additional_domain and ["&"] + additional_domain or additional_domain
-                ) + word_domain
-
-            return expression.OR([additional_domain, domain])
-
-        return domain
+                domain = (domain and ["&"] + domain or domain) + word_domain
```
dreispt: Would this work?
```suggestion
                domain = expression.AND(domain, word_domain)
```

lef-adhoc: **No, the suggested change wouldn't work as written.**

**Two issues:**
1. `expression.AND()` requires a list: `expression.AND([domain1, domain2])`
2. Missing empty-domain handling: when `domain = []`, it needs special treatment

**Correct version would be:**
```python
domain = expression.AND([domain, word_domain]) if domain else word_domain
```

I tested both approaches in Odoo shell across multiple scenarios. **Results: functionally identical** ✅

The current code works correctly using explicit list concatenation. Your suggested approach would also work, but only with the proper syntax shown above.

**If you prefer, I can update the code to use `expression.AND()` for better readability.** Let me know! @dreispt

lef-adhoc:
@dreispt should we merge this PR?

lef-adhoc:
@dreispt @pedrobaeza Hello, what else is needed to merge this?

base_name_search_improved/models/ir_model.py
dreispt: My concern with the original ir readability: at least I have difficulty understanding it.
Your suggested alternative sounds good.

lef-adhoc: @dreispt done!

lef-adhoc:
@dreispt Will it be possible to merge now?

base_name_search_improved/models/ir_model.py
```diff
@@ -82,36 +82,19 @@ def _compute_smart_search(self):
...
-            matching_records = self.with_context(
-                force_smart_name_search=True
-            ).name_search(name=value, operator=operator)
-            if matching_records:
-                record_ids = [record[0] for record in matching_records]
-                return [("id", "in", record_ids)]
-        return [("id", "=", False)]
-
-    @api.model
-    def _search_display_name(self, operator, value):
```
yostashiro, line 94: Removing this extension seems to make smart search unusable for models that don't rely on `_get_name_search_domain()`. `product.product` is an example.

I don't have a concrete suggestion at this point on how to work around the issue but just reporting what I've noticed.

yostashiro (changes requested):

PR #3358, opened 2025-09-01
[16.0][FIX] auditlog: safeguard name_get for deleted records

ntsirintanis:
Bypasses the call for auditlog.rule.name_get(), if logged record does not exist anymore.

Case that this solves:
- Some customization introduces `unlink` being called on an already unlinked record
- Odoo vanilla would just ignore the second unlink
- Auditlog however tries to name_get it resulting in a breaking error


auditlog/models/rule.py
```diff
@@ -514,8 +514,16 @@ def create_logs(
...
-            name = model_model.browse(res_id).name_get()
-            res_name = name and name[0] and name[0][1]
+            # Safeguard against deleted records
+            rec = model_model.browse(res_id)
+            if rec.exists():
+                try:
+                    pairs = rec.name_get()
+                    res_name = pairs and pairs[0] and pairs[0][1]
+                except Exception:
+                    res_name = f"{model_model._name} id={res_id}"
```
NL66278: There is no need for the model_model._name, as that will be part of the log line anyway. Also we do not see any difference in the log between no longer existing records, and errors in name_get.

Note that both situations cannot happen if consuming code functions properly, but that is why this is a safeguard.

I would propose for the exception case `res_name = "exception in name_get()"` for the not existing case `res_name = "record no longer exists"`. The id of the record is also already given in the log line.

NL66278: Strictly speaking these messages should also be translated, as they become visible for the user.

NL66278 (changes requested):
Nice to have this safeguard, but would like to have a few changes.

NL66278:
@ntsirintanis Please squash the commits and it should be good to go.

thomaspaulb (approved):

NL66278 (approved):
:+1: LGTM And works in practice.

StefanRijnhart (commented):
Is it possible to add a test for this? Would it work to just call unlink twice on the same record in a test?

thomaspaulb:
Most probably yes, @ntsirintanis could you try?

NL66278:
You can unlink the same record twice:
```
In [12]: wegermee = self.env["res.partner"].browse(40)
In [13]: wegermee.unlink()
2025-09-15 12:41:00,386 10106 INFO ? odoo.models.unlink: User #1 deleted mail.message records with IDs: [4195, 4107, 4029] 
2025-09-15 12:41:00,500 10106 INFO ? odoo.models.unlink: User #1 deleted ir.model.data records with IDs: [5516] 
2025-09-15 12:41:00,525 10106 INFO ? odoo.models.unlink: User #1 deleted ir.attachment records with IDs: [187, 91, 123, 155, 59] 
2025-09-15 12:41:00,526 10106 INFO ? odoo.models.unlink: User #1 deleted res.partner records with IDs: [40] 
Out[13]: True
In [14]: wegermee.exists()
Out[14]: res.partner()
In [15]: wegermee.unlink()
2025-09-15 12:41:35,991 10106 INFO ? odoo.models.unlink: User #1 deleted res.partner records with IDs: [40] 
Out[15]: True
In [16]: 
```

thomaspaulb:
I'm pretty sure that when you have an Odoo 16.0 instance with `auditlog` that does not include this fix, but that has a `delete` auditlog rule activated on the record, and you unlink twice, you will run into an error. So such a testcase could be created, the fix commented out, the test run on local, error seen, fix applied, error gone

NL66278:
Actually one of the existing tests with unlink could be adapted to run the unlink a second time and check whether the latest message for the model+res_id refers to no longer existing.

StefanRijnhart (commented):
Thanks! Can you squash commits?

auditlog/tests/test_auditlog.py
```diff
@@ -259,6 +259,28 @@ def test_LogDelete(self):
             1,
         )
 
+    def test_LogName_when_record_unlinked_twice(self):
```
NL66278: Mixed camel casing  and snake style. Please rename to `test_log_name_when_unlinked_twice`.

StefanRijnhart (approved):

PR #3210, opened 2025-02-24
[18.0] [MIG] base_sequence_option

fredericgrall:

PR #3359, opened 2025-09-02
[18.0][MIG] base_sequence_option: Migration to 18.0

TheerayutEncoder:
Standard migration :100:


Saran440:
@TheerayutEncoder I think this module is already migrated https://github.com/OCA/server-tools/pull/3210

PR #3361, opened 2025-09-04
[18.0][FIX] auditlog: Account Invoice Extract Cache Issue (no company set)

BT-sschmid:
AI Extraction throws an error when auditlog rule with Full log is enabled on account.move.line:

```
Error while reloading AI data on account.move 13399: We can't leave this document without any company. Please select a company for this document
```

## Module
audit log
https://github.com/OCA/server-tools/blame/22ed5dc4fb76971dbce0bdfd463cff5f1719c2ec/auditlog/models/rule.py#L400

## Describe the bug
When using the AI Exctraction for invoices, Odoo tries to map a purchase order. 
During this process, the auditlog triggers a cache-invalidation, so that in 
https://github.com/brain-tec/odoo/blob/2718aaa7397e38be0fcd64fe6b5ef5c0be66cfe2/addons/purchase/models/account_invoice.py#L85-L87
the reference to the purchase_id is lost. 
This happens after setting the invoice_line_ids in 
https://github.com/brain-tec/odoo/blob/2718aaa7397e38be0fcd64fe6b5ef5c0be66cfe2/addons/purchase/models/account_invoice.py#L203

## To Reproduce
**Affected versions**: 18

1.  upload a bill with a PO reference existing in Odoo
2.  send to AI
3.  Load the results

**Expected behavior**
No error 

This PR removes the cache invalidation. The tests ares still passing, so I guess it is no more needed in v18

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -395,9 +395,7 @@ class AuditlogRule(models.Model):
             # invalidate_recordset method must be called with existing fields
             if self._name == "res.users":
                 vals = self._remove_reified_groups(vals)
-            # Prevent the cache of modified fields from being poisoned by
-            # x2many items inaccessible to the current user.
-            self.invalidate_recordset(vals.keys())
+
             result = write_full.origin(self, vals, **kwargs)
             new_values = {
                 d["id"]: d


StefanRijnhart (changes requested):
Thank you for your PR and its detailed description. I agree that this line is causing a lot of problems, and that under normal circumstances the green tests validate that it can be safely removed.

The actual issue that it meant to fix involved multi company taxes on products, a situation which could not be reproduced given `auditlog`'s own module dependencies, so we settled for a test on groups and users. This test does not fail after removing the line. However, the issue with products reappears after removing this line.

This week, I was fixing another issue with auditlog and it was caused by the same line. So I'm proposing to replace this line with a more intricate way of cache isolation in https://github.com/OCA/server-tools/pull/3374. It is accompanied by a dedicated test module that also reproduces the issue with products having taxes in multiple companies. Can I ask you to have a look at that PR and see if we can close this one?

BT-sschmid:
@StefanRijnhart thanks for the info. 
Your PR https://github.com/OCA/server-tools/pull/3374 is a quite good solution to prevent cache mixups by the audit_log. 
I'll close this.

Issue #3224, opened 2025-03-12
Auditlog Causes Unit Test Failure


AungKoKoLin1997:
## Describe the Bug
Unit tests fail when there is a multi-company setup and an `auditlog` rule for the model.

# Impacted Versions
- 16.0

Haven't checked in other versions

## To Reproduce
- Enable the `auditlog` module.
- Create an auditlog rule for the` res.partner` model.
- Ensure a parent partner exists with child partners, where at least one child has an assigned company_id.
- Attempt to update the is_company field of the parent partner during a unit test.
- The test will fail with an AccessError.
 OR
- Run the[ account_financial_report](https://github.com/OCA/account-financial-reporting/tree/16.0/account_financial_report) test cases with the `auditlog` module enabled to reproduce the issue.

This is the Traceback when I run test cases of `account_financial_report`.

```
Traceback (most recent call last):
  File "/opt/odoo/odoo/api.py", line 1023, in get
    cache_value = field_cache[record._ids[0]]
KeyError: 22

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1161, in __get__
    value = env.cache.get(record, self)
  File "/opt/odoo/odoo/api.py", line 1030, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: 'res.partner(22,).is_company'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/fields.py", line 1187, in __get__
    recs._fetch_field(self)
  File "/opt/odoo/odoo/models.py", line 3229, in _fetch_field
    self._read(fnames)
  File "/opt/odoo/odoo/models.py", line 3355, in _read
    raise self.env['ir.rule']._make_access_error('read', forbidden)
odoo.exceptions.AccessError: Due to security restrictions, you are not allowed to access 'Contact' (res.partner) records.

Contact your administrator to request access if necessary.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/__w/sfaj-oca/sfaj-oca/account_financial_report/tests/test_open_items.py", line 42, in test_partner_filter
    partner_3.write({"is_company": False})
  File "/__w/sfaj-oca/sfaj-oca/auditlog/models/rule.py", line 396, in write_full
    result = write_full.origin(self, vals, **kwargs)
  File "/opt/odoo/odoo/addons/base/models/res_partner.py", line 734, in write
[...]

PR #3364, opened 2025-09-05
[16.0][FIX] auditlog: Ensure unit tests unpatch methods.

amh-mw:
Though TransactionCase rolls back database changes, it does not call `unlink` on `auditlog.rule` records, meaning that `unsubscribe` is never called to unpatch model methods. This causes subsequent tests in other modules to unexpectedly execute patched methods.

This commit adds the `AuditLogRuleCommon` test class in an attempt to identify and prevent patched methods from contaminating the test environment.

### DIFF
--- a/auditlog/tests/common.py
+++ b/auditlog/tests/common.py
@@ -1,0 +1,33 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class AuditLogRuleCommon(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.models = set()
+
+    @classmethod
+    def create_rule(cls, vals):
+        rule = cls.env["auditlog.rule"].with_context(tracking_disable=True).create(vals)
+        # Keep track of patched models
+        cls.models |= set(rule.model_id.mapped("model"))
+        return rule
+
+    @classmethod
+    def tearDownClass(cls):
+        for rule in cls.env["auditlog.rule"].search([]):
+            try:
+                rule.unsubscribe()
+            except KeyError:  # pragma: no cover
+                continue  # Model not loaded yet
+
+        # Assert no patched methods remain
+        for model in cls.models:
+            for method in ["create", "read", "write", "unlink"]:
+                assert not hasattr(
+                    getattr(cls.env[model], method), "origin"
+                ), f"{model} {method} still patched"
+        super().tearDownClass()
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -2,10 +2,12 @@
 # © 2018 Pieter Paulussen <pieter_paulussen@me.com>
 # © 2021 Stefan Rijnhart <stefan@opener.amsterdam>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo.tests.common import Form, TransactionCase
+from odoo.tests.common import Form
 
 from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
 from odoo.addons.base.models.res_users import name_boolean_group
+
+from .common import AuditLogRuleCommon
 
 
 class AuditlogCommon(object):
@@ -299,11 +301,11 @@ class AuditlogCommon(object):
         self.assertIsInstance(domain[0][2], list)
 
 
-class TestAuditlogFull(TransactionCase, AuditlogCommon):
+class TestAuditlogFull(AuditLogRuleCommon, AuditlogCommon):
     def setUp(self):
         super(TestAuditlogFull, self).setUp()
         self.groups_model_id = self.env.ref("base.model_res_groups").id
-        self.groups_rule = self.env["auditlog.rule"].create(
+        self.groups_rule = self.create_rule(
             {
                 "name": "testrule for groups",
                 "model_id": self.groups_model_id,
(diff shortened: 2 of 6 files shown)


amh-mw:
@AungKoKoLin1997 Would you kindly review this pull request? It fixes #3224.

auditlog/tests/common.py
```diff
@@ -0,0 +1,26 @@
...
+
+class AuditLogRuleCommon(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.models = set()
+
+    @classmethod
+    def create_rule(cls, vals):
+        rule = cls.env["auditlog.rule"].with_context(tracking_disable=True).create(vals)
```
AungKoKoLin1997, line 14:
```suggestion
    def create_rule(cls, name, model_id, log_type=Fast, log_read=True, log_create=True, log_write=True, log_unlink=True, capture_record=True):
        rule = cls.env["auditlog.rule"].with_context(tracking_disable=True).create(
             {
                "name": name,
                "model_id": model_id,
                "log_type": log_type,
                "log_read": log_read,
                "log_create": log_create,
                "log_write": log_write,
                "log_unlink": log_unlink,
                "capture_record": capture_record,
            }
       )
```

auditlog/tests/test_auditlog.py
```diff
@@ -337,18 +335,16 @@ def setUpClass(cls):
...
+            {
+                "name": "test.model",
+                "model_id": cls.test_model.id,
+                "log_type": "fast",
+                "log_read": False,
+                "log_create": True,
+                "log_write": True,
+                "log_unlink": False,
+            }
         )
```
AungKoKoLin1997, line 387:
```suggestion
        cls.auditlog_rule = cls.create_rule("test.model", cls.test_model.id, log_read=False, log_unlink=False, capture_record=False)
```

auditlog/tests/test_auditlog.py
```diff
@@ -459,21 +455,17 @@ def setUpClass(cls):
...
+                "name": "testrule 01",
+                "model_id": cls.contact_model_id,
+                "log_read": True,
+                "log_create": True,
+                "log_write": True,
+                "log_unlink": True,
+                "log_type": "full",
+                "capture_record": True,
+            }
         )
```
AungKoKoLin1997, line 508:
```suggestion
        cls.auditlog_rule = cls.create_rule("testrule 01", cls.contact_model_id, log_type="full")
```

AungKoKoLin1997 (commented):
How about like this?
This can reduce multiple lines.

auditlog/tests/test_auditlog.py
amh-mw: Given that `create_rule` is essentially a stand-in for
```python
    @api.model_create_multi
    def create(self, vals_list):
```
I would either prefer to leave it as is, or adopt a signature even more similar.

amh-mw: Force pushing a small change to support multi-create:
```diff
-        cls.models |= {rule.model_id.model}
+        cls.models |= set(rule.model_id.mapped('model'))
```

AungKoKoLin1997 (approved):
LGTM

auditlog/tests/common.py
```diff
@@ -0,0 +1,26 @@
+from odoo.tests.common import TransactionCase
```
yostashiro, line 3: How about adding a license header with you as the author?

yostashiro (approved):
LGTM. Thanks for the fix!

auditlog/tests/common.py
amh-mw: I added a license header to the file and added myself to the contributors readme.

StefanRijnhart (commented):
This code looks like a welcome improvement to have auditlog tests clean up properly afterwards. But apart from that, is it not the case that the multicompany issue will still persist? Even though it is not triggered in the UI, will it still pop up when records are written in code?

amh-mw:
> is it not the case that the multicompany issue will still persist?

Though #3224 describes itself as a multi-company issue, standard Odoo demo data installed for testing already has multiple companies. I perceive it as a test environment contamination problem, not a multi-company problem. The `docker-compose.test.yaml` I added to that issue is my litmus for success; red before and green after this pull request.

auditlog/tests/common.py
```diff
@@ -0,0 +1,28 @@
...
+    @classmethod
+    def create_rule(cls, vals):
+        rule = cls.env["auditlog.rule"].with_context(tracking_disable=True).create(vals)
+        # Keep track of patched models
+        cls.models |= set(rule.model_id.mapped("model"))
+        return rule
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.env["auditlog.rule"].search([]).unsubscribe()
```
StefanRijnhart:
```
```suggestion
        for rule in cls.env["auditlog.rule"].search([]):
            try:
                rule.unsubscribe()
            except KeyError:
                # Model not loaded yet
                continue
```

StefanRijnhart (commented):
Fair enough, it's true that the issue mentions unit tests explicitely. I'll be fixing the multi-company issue in a separate PR.

StefanRijnhart:
Picked into https://github.com/OCA/server-tools/pull/3371 to fix the tests. There, the problem was indeed not the multi-company aspect fixed in the same PR but the fact that the patched methods mask method overrides loaded after the patching.

auditlog/tests/common.py
amh-mw: Could you give me an example of how to test this behavior? How can a model not be loaded?

StefanRijnhart: Oh, if you run tests on a database on which rules have been manually configured, or on which a module was installed earlier that includes data for rules for models that are not loaded when these tests load.

To reproduce, install account on your database and create a rule on account.move. Then run the tests. It will fail on the KeyError.

amh-mw: I don't think that's a valid test environment.

amh-mw: Since `auditlog` does not depend on `account`, it is impossible for a clean test environment to have an `auditlog.rule` on an `account.move` record on first run. Though it would make development a lot easier if tests could be repeatedly run without dropping and recreating the database every time, it often simply doesn't work. Module installations can override methods or change field permissions in such a way that it becomes impossible for subsequent test run to succeed.

That said, forward-compatibility is a laudable goal. My main hesitation in adding this sort of compatibility code is that it is by definition unable to be tested; permanently red in code coverage.

StefanRijnhart: That seems a small price to pay for the convenience of using the same database for both automated and manual testing.

amh-mw: TIL `pragma: no cover` which sufficiently addresses my concern; added `KeyError` guard to `tearDownClass`.

StefanRijnhart (approved):
Thanks for the update!

/ocabot merge patch

PR #3354, opened 2025-08-25
[14.0][IMP] Backport of 16.0 Improvements & Adjusted Tests

raumschmiede-joshuaL:
Hello.

I split up my other PR https://github.com/OCA/server-tools/pull/3346 to add the new field here.

And I backported a few of your commits that improve the module, @lmignon. 

EDIT: Now it is mostly just a backport. My other commits improve the tests a bit.

@mt-software-de

PR #3365, opened 2025-09-05
[14.0][IMP] base_exception: Better general Handling of Sub-Exceptions

raumschmiede-joshuaL:
Hi,

I moved the new def `_get_sub_exception_field_names` from PR https://github.com/OCA/server-tools/pull/3354 to here.
This here uses the commits of the other PR. I will do a rebase after the other was merged.

I implemented a way to add exceptions to the record on which the exceptions where detected on in case `_get_main_records` returns something else than `self`, like `self.order_id`.
~That commit breaks `sale_exception` because I added `_get_main_records` to `detect_exceptions` so that it is not necessary anymore to do this in an overridden `_detect_exceptions`.~

**EDIT**: I checked `sale_exception` again to update PR https://github.com/OCA/sale-workflow/pull/3861. Seems that it is not breaking the module. The overridden `_detect_exceptions` in `sale.order.line` returns the same as the overridden `_get_main_records`. After that `_get_main_records` is called again on the returned sales orders. So only one useless `_get_main_records` call.

I removed the comment in `detect_exceptions` because it refered to the `_reverse_field` and the Many2many field on the exception rule. As this does not exist anymore, the comment is also not needed anymore

### DIFF
--- a/base_exception/models/base_exception.py
+++ b/base_exception/models/base_exception.py
@@ -1,5 +1,6 @@
 # Copyright 2011 Raphaël Valyi, Renato Lima, Guewen Baconnier, Sodexis
 # Copyright 2017 Akretion (http://www.akretion.com)
+# Copyright 2025 Raumschmiede GmbH
 # Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
 # Copyright 2020 Hibou Corp.
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu)
@@ -90,6 +91,22 @@ class BaseExceptionModel(models.AbstractModel):
     @api.model
     def _get_popup_action(self):
         return self.env.ref("base_exception.action_exception_rule_confirm")
+
+    def _add_detected_exceptions_to_self(self):
+        return self != self._get_main_records()
+
+    def _detect_exceptions(self, rule):
+        records = super()._detect_exceptions(rule)
+        # If _get_main_records returns self, it adds the exceptions to itself in
+        # detect_exceptions. If _get_main_records does not return self, it adds the
+        # exceptions here
+        if not self._add_detected_exceptions_to_self():
+            return records
+
+        (self - records).exception_ids = [(3, rule.id)]
+        records.exception_ids = [(4, rule.id)]
+
+        return records
 
     def _check_exception(self):
         """Check exceptions
--- a/base_exception/models/base_exception_method.py
+++ b/base_exception/models/base_exception_method.py
@@ -3,6 +3,7 @@
 # Mourad EL HADJ MIMOUNE <mourad.elhadj.mimoune@akretion.com>
 # Copyright 2020 Hibou Corp.
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu)
+# Copyright 2025 Raumschmiede GmbH
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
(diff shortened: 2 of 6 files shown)


raumschmiede-joshuaL:
Will add the 2nd commit after the tests for the 1st commit are done to be sure I committed the right things

raumschmiede-joshuaL:
@hparfr and @lmignon, did a rebase. Can you check this here please?

raumschmiede-joshuaL:
@hparfr, if a model inherits from `base.exception.method`, it must implement field `ignore_exception` because that class uses the field in `_get_base_domain` when it tries to find the records on which the rule needs to be applied on. If this field is not implemented, an error is raised.

That's why I want to move it from `base.exception` to `base.exception.method` to have it automatically implemented through inheritance. Is this ok?
If models need a special implementation for this field, it needs to be implemented anyway manually. Like in  `sale.order.line` so that the field is related to `order_id.ignore_exception`.
For other models the field will always be False and never changed. But this would be the case anyway if you have to implement it eventhough you wouldn't use it

base_exception/models/base_exception.py
```diff
@@ -91,6 +92,22 @@ def _popup_exceptions(self):
...
+    def _detect_exceptions(self, rule):
+        records = super()._detect_exceptions(rule)
+        # If _get_main_records returns self, it adds the exceptions to itself in
+        # detect_exceptions. If _get_main_records does not return self, it adds the
+        # exceptions here
+        if not self._add_detected_exceptions_to_self():
+            return records
+
+        (self - records).exception_ids = [(3, rule.id)]
+        records.exception_ids = [(4, rule.id)]
```
mt-software-de, line 107: Here a small minor request. 
```suggestion
        if self._add_detected_exceptions_to_self():
            (self - records).exception_ids = [(3, rule.id)]
            records.exception_ids = [(4, rule.id)]
```

mt-software-de (changes requested):

base_exception/models/base_exception.py
mt-software-de: Is it correct that since you are writing the rules now on the records, the logic in `detect_exceptions` in `base.exception.method` is not used anymore. Or only if the returned main record isn't the same as where it was executed from?

Isn't this quite bad because the checks and writes in `detect_exceptions` are done twice?

Isn't it possible to have the writing in just one place?

raumschmiede-joshuaL: `detect_exceptions` in `base.exception.method` is still needed.
If a model inherits from `base.exception.method`, it has no `exception_ids`, that's why it must return another record in `_get_main_records`. The detected rules in `base.exception.method` are written to that other model. Like it is done right now in `account_move_exception`.

If a model inherits from `base.exception`, there are 2 ways:
1. It overrides `_get_main_records`, like it is done in my PR for `sale_exception`. But as it returns another record, it will not write the detected rules on the sale order lines, only on the sales orders returned by `_get_main_records`. That's why this override here is needed
2. It does not override `_get_main_records`: Then the detected rules are added to self in `base.exception.method` and this here has no effect as it is always False

mt-software-de: Ok. Since you are only adding it if the main record isn't the same. So if the exception is on sale.order and it is detected on sale.order it will add it in `detect_exceptions`.

raumschmiede-joshuaL:
@sebastienbeau and @hparfr, can you check this PR?

PR #3366, opened 2025-09-09
[15.0][FIX] module_analysis: pin pygount

niboo-ave:
Just like https://github.com/OCA/server-tools/pull/2967
PIN pygount.


StefanRijnhart (approved):
Thanks! I would normally ask for a cherry-pick to preserve credits but I guess there is less value in that in this case.

MiquelRForgeFlow (approved):

PR #3367, opened 2025-09-10
[16.0][FIX] base_technical_user: use company from env, not from user + multi company context

lmarion-source:
1) the company should be taken through self.env.company as user.company_id
is now the default company for the user (and not the current one).

2) the actual code does not allow to do recordset.with_company(B).sudo_tech()....
as allowed_company_ids in context can be filled in with current company
(and not the one of the tech user - which is unique)

This was Done by Denis Roussel in 14.0 but somehow forgotten during 16.0 migration (back there in 18.0...)

### DIFF
--- a/base_technical_user/models/models.py
+++ b/base_technical_user/models/models.py
@@ -12,9 +12,21 @@ class Base(models.AbstractModel):
 
     def sudo_tech(self, raise_if_missing=False):
         self_sudoer = self
-        tech_user = self.env.user.company_id.user_tech_id
+        tech_user = self.env.company.user_tech_id
         if tech_user:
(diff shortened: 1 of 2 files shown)


AnizR (approved):
Code LGTM, thanks for your contribution!

acsonefho (approved):

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
> in 14.0 but somehow forgotten during 16.0 migration (back there in 18.0...)

Sad for 15.0 and 17.0. 🤧

StefanRijnhart (approved):
Thanks!

/ocabot merge patch

Issue #2561, opened 2023-02-22
Audit Log: Influences credit note creation


jappi00:
## Describe the bug
When I using audit log to log changes on the account.move.line model. (Image) And creating a payment term which have some delayed payment goals I can't create an credit note.
![Bildschirm­foto 2023-02-22 um 22 38 35](https://user-images.githubusercontent.com/13003175/220765131-791274a4-e9b3-44a9-87d3-3d3b16f52b1a.png)
![Bildschirm­foto 2023-02-22 um 22 38 48](https://user-images.githubusercontent.com/13003175/220765189-bd3c4122-ac3e-4293-ad41-5d413c960e07.png)

## To Reproduce
14e: Configure the auditlog module like above, configure the payment term and create an invoice with this payment term. Confirm it, and try to create an credit note. You get the below error.
![Bildschirm­foto 2023-02-22 um 22 40 47](https://user-images.githubusercontent.com/13003175/220765519-a86e4e0c-f8e4-4d99-90cb-64938116d7c2.png)

**Expected behavior**
No Error

**Additional context**
I have debugged the issue to this point where the update gets called: https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L1036

When I deactivate the rule everything works fine.

I'am very confused. I didn't expect that auditlog have such an impact on the code but I cannot explain why this happens.

jappi00:
I can reproduce it on runbot and community aswell.

This is a massive problem. Maybe somebody more clever has an idea why this is happening?

DucTruongKomit:
Now, I faced kind of same issue in Odoo 16, when click to create a credit note.

amon-ra:
Not a fix but Fast log_type works without problems

StefanRijnhart:
@jappi00 @DucTruongKomit Please check the linked branches. Fix is proposed for 16, 17, and 18.

LoisRForgeFlow:
Fixes merged

Issue #2554, opened 2023-02-16
AuditLog rule on product.template not compatible with multi company rule


NL66278:
Module auditlog

When having product templates with customer taxes in a multi company setting, and there is an audit rule on product.template, the user will no longer be able to open the product form, the user will get a  messsage that he is not allowed to read some tax because of:
This restriction is due to the following rules:
- Tax multi-company

The user set to look at company A will get an error for taxes for company B and vice versa.

Apparently the audit rule influences how the field taxes_id (the same also goes for the Vendor Taxes by the way) is read. Without audit log rule, or the rule unsubscribed, taxes_id only contains the tax the user can see due to the Tax multi-company rule. WIth the rule enabled, taxes_id contains all taxes defined for the product. DESPITE THE RULE ONLY BEING FOR WRITE, CREATE AND UNLINK. So somehow, I was not able to see in the code why, the auditlog rule influences the reading of values for product.template, even without logging reads.

## To Reproduce
I noted this on Odoo 15.0. AuditLog is at 15.0.1.1.0

1. Have multiple companies, and product, account and auditlog installed
2. Add taxes for company A to a product, switch company and add taxes for company B
3. Create an audit rule for product.template, with write, create and unlink logging. Not needed to activate read logging.
4. Try to open the product.template form.

**Expected behavior**
The form should just open and only show the taxes for the active company.

I added a traceback to the generation of the error and a pdb line to product template model to see what was going on. Last part of traceback pasted below. Diff to odoo for traceback and pdb attached.

Traceback:
```
[traceback.txt](https://github.com/OCA/server-tools/files/10755005/traceback.txt)
  File "/home/openeyedev/projecten/callista/custom/src/odoo/odoo/api.py", line 451, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/openeyedev/projecten/callista/custom/src/odoo/odoo/models.py", line 3227, in read
    return self._read_format(fnames=fields, load=load)
  File "/home/openeyedev/projecten/callista/custom/src/odoo/odoo/models.py", line 3247, in _read_format
    vals[name] = convert(record[name], record, use_name_get)
  File "/home/openeyedev/projecten/callista/custom/src/odoo/odoo/models.py", line 5888, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/openeyedev/projecten/callista/custom/src/odoo/odoo/fields.py", line 1108, in __get__
    self.compute_value(record)
[...]

PR #3371, opened 2025-09-16
[16.0][FIX] auditlog: fix caching; [ADD] test_auditlog

### LINKED ISSUE #3308: [17.0] ORM caches many2many field on write without applying company rules
## Module
auditlog

## Describe the bug
When changing any field in a product template, an access error occurs on account.tax records, preventing the record from being saved.

## Actual Behavior
An access rights error occurs on `account.tax` when saving the product template. The issue is caused by the `write_full` method using `sudo()` to read the `taxes_id` field, retrieving tax records from all companies. This value is stored in the ORM cache.
When saving the record, Odoo re-reads the taxes_id field from the cache (containing data across all companies), and attempting to access `display_name` on those ta

## To Reproduce
1. Create two companies.
2. Ceate a product template without assigning a company.
3. Add company-specific taxes to the product template.
4. Enable an auditlog rule to track full write operations on the product template model.
5. Edit any field in the product template form.

## Proposed Solution
Invalidate all read fields after creating the audit log, so that subsequent reads use proper company filtering rather than cached cross-company data.

## Expected behavior
The record should be saved successfully, and changes should be logged without access errors.

StefanRijnhart:
Fix caching and add a dedicated test module to run tests on the accounting models.

Fixes creating credit notes with audit rules on account.move.line, writing taxes on products as a user that does not have access to all companies for which there are taxes on the product, and more.

Fixes #3308, #3224, #2561, #2554


NL66278 (approved):
:+1: LGTM and extensive unit tests

MiquelRForgeFlow (approved):

StefanRijnhart:
Setting this to draft as I've encountered an issue swapping out the cache as is done here.

StefanRijnhart:
Fixed now. Issue is explained in the docstring of the new context manager.

NL66278:
@StefanRijnhart The contextmanager seems to be a quite clean solution. Maybe useful in more places.

StefanRijnhart:
Rebased to fix conflicts

CasVissers-360ERP (approved):
Functional review. Fixed the issue for me.

StefanRijnhart:
Rebased after the merge of https://github.com/OCA/server-tools/pull/3364, which was originally included here.

auditlog/models/rule.py
```diff
@@ -47,6 +47,41 @@ def unchanged(self):
...
+        temporary_cache = api.Cache()
+        for env in self._transaction.envs:
+            env.cache = temporary_cache
+        self._transaction.cache = temporary_cache
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Restore the original cache wherever it was replaced."""
+        for env in self._transaction.envs:
+            env.cache = self._original_cache
```
NL66278, line 81: Revisiting this again, and not having to deep knowledge about how caching works, this question entered my mind: if there is a possibility of one or more of the environments in self._transaction.envs having a different cache then self._transaction.cache before entering the context manager, then after exiting the context manager, they will all have the (original) self._transaction.cache. Could this happen and cause problems?

StefanRijnhart: Well, that this was my original approach, to have one env with an alternate cache, and that approach fell at the first hurdle given how easily Odoo swaps out one env for another (which will have the original cache). So I don't think that approach would work anywhere else. Also, I just checked, and there is no code in Odoo or any 16, 17 or 18 repo that I use in my wide array of setups that patches out the cache like the code in this PR or anything more wild. So no risk I would say.

NL66278: Maybe I was not clear enough, I just wondered whether all the original caches would be correctly restored. Anyway, it is probably completely theoretical.

NL66278:
@thomaspaulb Could you merge this one?

auditlog/models/rule.py
StefanRijnhart: @NL66278 I tried to argue that the situation you describe (one or more envs having a different cache than the cache on the transaction) cannot happen because it cannot work in Odoo (as I happened to have found out during the development phase of this code).

amh-mw (approved):
Code review only, LGTM. Though superior to cache poisoning, I'm curious how much slower this makes `full` audit logging

auditlog/models/rule.py
NL66278: OK, now I understand. Thank you for your answers!

DavidJForgeFlow (approved):

LoisRForgeFlow (approved):
Thanks for this! :+1: 

/ocabot merge patch

StefanRijnhart:
See https://github.com/OCA/server-tools/pull/3409 for an important fix for this change.

PR #3377, opened 2025-09-19
[MIG] base_sequence_option

matteoopenf:


matteoopenf:
https://github.com/OCA/server-tools/pull/3210 is already migrated here

PR #3378, opened 2025-09-19
Remove pygount hardcoded version and improve compatibility

fcayre:
pygount 1.4.0 is now very old and current code does not work on recent versions. It is however easy to improve compatibility accross pygount versions using the code_count property instead of the code one, which was removed in 2.0.0. That way, pygount 1.4.0 stays supported but also modern version up to at least
 3.1.0.

### DIFF
--- a/module_analysis/__manifest__.py
+++ b/module_analysis/__manifest__.py
@@ -26,7 +26,7 @@
         "data/ir_module_type_rule.xml",
     ],
     "external_dependencies": {
-        "python": ["pygount==1.4.0"],
+        "python": ["pygount"],
     },
     "installable": True,
 }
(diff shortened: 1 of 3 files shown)


legalsylvain:
Thanks ! Could you take a look on precommit ?

Otherwise, lgtm !

legalsylvain:
Thanks !

/ocabot merge patch

fcayre:
@legalsylvain sorry, done. Thanks for reviewing!

Note I successfully tested the code with pygount 1.4.0 and 3.1.0 (last published version as of today) with python 3.11.2 (debian 12 bookworm) and python 3.9.2 (debian 11 bullseye). The API used in the module is thus pretty stable (SourceAnalysis.code_count), this is why I propose to drop the version restriction. The CI may also help detect an API incompatibility, so the drawbacks of installing the 1.4.0 appeared greater to me than the advantages.

PR #3386, opened 2025-09-23
[16.0][FIX] auditlog: Fix Missing Log Lines When Setting Empty Field Values

BT-pgaiser:
Fixes the comparison with old values when setting fields to empty values.

The old values for the comparison when using **fast log** mode were initialized with `False`. This causes the problem that when a field is set to an emty value by the user, auditlog does not create a log line because internal checks compare `False` to `False` and do not consider logging the change. Full log mode is not affected by this as there everything is logged by default. However, even in fast log mode, some companies have regulatory requirements which force them to log user actions that set empty values to fields, so that should be logged properly.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -91,6 +91,19 @@ class ThrowAwayCache:
         self._transaction.tocompute = self._original_tocompute
 
 
+class _Sentinel:
+    """A falsy sentinel object as a placeholder for absent values."""
+
+    def __str__(self):
+        return ""
+
+    def __bool__(self):
+        return False
+
+
+_SENTINEL = _Sentinel()
+
+
 class AuditlogRule(models.Model):
     _name = "auditlog.rule"
     _description = "Auditlog - Rule"
@@ -506,7 +519,7 @@ class AuditlogRule(models.Model):
             # afterwards as it could not represent the real state
             # of the data in the database
             vals2 = dict(vals)
-            old_vals2 = dict.fromkeys(list(vals2.keys()), None)
+            old_vals2 = dict.fromkeys(list(vals2.keys()), _SENTINEL)
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -506,7 +506,7 @@ class AuditlogRule(models.Model):
             # afterwards as it could not represent the real state
             # of the data in the database
             vals2 = dict(vals)
-            old_vals2 = dict.fromkeys(list(vals2.keys()), False)
+            old_vals2 = dict.fromkeys(list(vals2.keys()), None)
             old_values = {id_: old_vals2 for id_ in self.ids}
             new_values = {id_: vals2 for id_ in self.ids}
             result = write_fast.origin(self, vals, **kwargs)
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -371,8 +371,7 @@ class TestAuditlogFastWriteEmptyValue(AuditLogRuleCommon):
         )
         cls.auditlog_rule.subscribe()
 
-    def test_when_removing_field_value_a_log_line_is_created(self):
-        self.test_record.write({"phone": False})
+    def assert_has_one_log_line_with_empty_values(self):
         logs = self.env["auditlog.log"].search(
             [
                 ("res_id", "=", self.test_record.id),
@@ -389,6 +388,14 @@ class TestAuditlogFastWriteEmptyValue(AuditLogRuleCommon):
         self.assertFalse(log_lines[0].old_value_text)
         self.assertFalse(log_lines[0].new_value)
         self.assertFalse(log_lines[0].new_value_text)
+
+    def test_when_removing_field_value_a_log_line_is_created(self):
+        self.test_record.write({"phone": False})
+        self.assert_has_one_log_line_with_empty_values()
+
+    def test_when_using_none_field_value_a_log_line_is_created(self):
+        self.test_record.write({"phone": None})
+        self.assert_has_one_log_line_with_empty_values()
 
 
 class TestFieldRemoval(AuditLogRuleCommon):
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -343,6 +343,54 @@ class TestAuditlogFast(AuditLogRuleCommon, AuditlogCommon):
         super(TestAuditlogFast, self).tearDown()
 
 
+class TestAuditlogFastWriteEmptyValue(AuditLogRuleCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.test_model = cls.env["ir.model"].search([("model", "=", "res.partner")])
+        cls.auditlog_rule = cls.create_rule(
+            {
+                "name": "test.model_res_partner",
+                "model_id": cls.test_model.id,
+                "log_type": "fast",
+                "log_read": False,
+                "log_create": False,
+                "log_write": True,
+                "log_unlink": False,
+            }
+        )
+        cls.test_record = (
+            cls.env["res.partner"]
+            .with_context(tracking_disable=True)
+            .create(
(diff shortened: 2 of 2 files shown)


BT-mlopez (approved):
LGTM, unit test look quite complete.

BT-ojossen (approved):
LGTM, but you should add yourself to the contributors

BT-tkarpinski (approved):

BT-pgaiser:
@amh-mw @StefanRijnhart 
Could someone please take a look at this and potentially merge? I don't know who's maintaining the auditlog module.

auditlog/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Audit Log",
-    "version": "16.0.3.0.0",
+    "version": "16.0.3.0.1",
```
StefanRijnhart: Please don´t update the version manually. Ocabot will do this during merge.

auditlog/tests/test_auditlog.py
```diff
@@ -321,6 +321,55 @@ def tearDown(self):
 
 
 class TestAuditlogFast(TransactionCase, AuditlogCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.test_model = (
```
StefanRijnhart: Do you need to setup a whole model? Can't you use a res.partner field like 'website'?

StefanRijnhart (commented):
Thanks! I see the issue. But doing a read is undermining the 'fast' aspect of the 'fast' type rule. Would it work to force any falsy value in the written `vals` to show up in the log lines (conditionally overriding the result of DictDiffer)?

auditlog/__manifest__.py
BT-pgaiser: Got it. Reverted the version bump.

auditlog/tests/test_auditlog.py
BT-pgaiser: I tried that but the other test cases seem to interfere with that, which results in a flaky test. I think having a separate model makes that test case more isolated and stable.

BT-pgaiser:
Thanks for the prompt response. Indeed, if we change the initial old value from `False` to `None` [here](https://github.com/OCA/server-tools/blob/5d3056492f486d7caec01d262bcf0f59c5dfa21a/auditlog/models/rule.py#L466), then the `DictDiffer` object will track the change and a log line is created. However, that log line will still show a change from "empty" to "empty" value, which is incorrect. As I have mentioned, some companies have regulatory requirements that oblige them to log these changes too.
Regarding the performance, yes, doing a round-trip to the DB has an additional cost but in many cases the field values are first read before they are written, which results in a high likelihood that on the side of the DB those values are still in cache so it won't hit the disk. Is there another way to obtain the old field values?

auditlog/tests/test_auditlog.py
StefanRijnhart: Indeed, flakiness of the tests is addressed in https://github.com/OCA/server-tools/pull/3364. I had to picked it in https://github.com/OCA/server-tools/pull/3371 to prevent tests from other modules from failing.

StefanRijnhart:
Isn't it the case that one never gets the old values with 'fast' logging? So having both an empty value for old and new is what is to be expected.

auditlog/tests/test_auditlog.py
StefanRijnhart: Would it help to move the test (based on res.partner) into its own test class?

StefanRijnhart: Note that https://github.com/OCA/server-tools/pull/3364 is now merged, so maybe that helps with your test. Be sure to inherit from `AuditLogRuleCommon` and create your rule with `self.create_rule`.

BT-pgaiser: Thanks. I rebased and it is working now.

BT-pgaiser:
That seems to be the case, yeah. So this is an intentional performance optimisation? How about we add an option to the auditlog rules (applicable only for the fast mode) to let the user decide whether old values should be logged?

StefanRijnhart:
@BT-pgaiser Isn't that what 'full' logging is for?

auditlog/tests/test_auditlog.py
StefanRijnhart: Great!

BT-pgaiser:
If that's the only difference between "fast" and "full" log mode, then yes, simply activating the "full" mode solves the issue. I was thinking about having something in between those two modes, but maybe that does not make a lot of sense.
Can you confirm? If yes, I'd keep the fix for the missing log lines when logging empty values but remove the read of old values.

StefanRijnhart:
@BT-pgaiser Yes please! Keep logging the empty values when specified explicitely in the write vals (in the case of write 'fast'), but don't read old values to compare.

auditlog/tests/test_auditlog.py
```diff
@@ -322,7 +322,34 @@ def tearDown(self):
         super(TestAuditlogFull, self).tearDown()
 
 
-class TestAuditlogFast(AuditLogRuleCommon, AuditlogCommon):
+class TestAuditlogFast(AuditLogRuleCommon):
```
StefanRijnhart: Better add your own test because we want to keep AuditLogCommon on this one, so that it runs the standard test methods from that case.

StefanRijnhart (commented):
Can you squash commits afterwards?

auditlog/tests/test_auditlog.py
BT-pgaiser: Done.

BT-pgaiser:
Commits squashed and test case refactored.

StefanRijnhart (approved):
Thanks, looking good now!

BT-pgaiser:
@StefanRijnhart Can we please merge this? Thanks.

StefanRijnhart:
@BT-pgaiser I would like to see another review from a tools maintainer or previous contributor to this module. The other reviews on this PR so far are clearly 'team service' because I don't see any other reviews in the OCA from them except for one or two other 'BT' PRs.

BT-pgaiser:
@StefanRijnhart Fair enough.
@amh-mw @mymage @tarteo @hbrunn Would one of you be so kind :)

auditlog/tests/test_auditlog.py
```diff
@@ -343,6 +343,54 @@ def tearDown(self):
...
+                {
+                    "name": "testpartner3",
+                    "phone": "123",
+                }
+            )
+        )
+        cls.auditlog_rule.subscribe()
+
+    def test_when_removing_field_value_a_log_line_is_created(self):
+        self.test_record.write({"phone": False})
```
hbrunn: can you do the same test with writing `None` here? Both are valid input, and I suspect you're just replacing when the misbehavior happens.

chances are you need some sentinel object like

```python
class FalsyValue():
    def __bool__(self):
        return False
```

to initialize the dict above with

hbrunn (changes requested):

auditlog/tests/test_auditlog.py
BT-pgaiser: I've added a test for writing `None` values. However, AFAIK, Odoo always represents missing/empty values as `False` and this is consistent with how the web client behaves. Anyhow, I've never seen `None` being used for that in Odoo. Surely, if I write a separate program that directly communicates with the Odoo backend, then I can send it `None` values, but I'd say that's _unspecified behaviour_ so it does not need to be supported. Anyways, the introduced sentinel object will handle both cases now.

hbrunn (approved):

auditlog/models/rule.py
```diff
@@ -463,7 +473,7 @@ def write_fast(self, vals, **kwargs):
             # afterwards as it could not represent the real state
             # of the data in the database
             vals2 = dict(vals)
-            old_vals2 = dict.fromkeys(list(vals2.keys()), None)
+            old_vals2 = dict.fromkeys(list(vals2.keys()), _Sentinel())
```
hbrunn: wait, for better performance, please create an instance of this above and recycle this object, no need to keep creating it

BT-pgaiser: Done.

BT-pgaiser:
@hbrunn Sorry for the delay, I was OoO last week.

hbrunn (approved):
thank you!

/ocabot merge patch

PR #3387, opened 2025-09-24
[17.0][UPD] base_fontawesome: Update FontAwesome to version 6.7.2

zamberjo:


maurochip (approved):

sersanchus (approved):

zamberjo:
@OCA/tools-maintainers, can we have this merged?

ByteMeAsap:
How did this get approved and merged? 

I can only see one icon getting loaded after these changes

image

ByteMeAsap:
This is from before these changes were merged
image

ByteMeAsap:
@pedrobaeza , can you please take a look?

pedrobaeza:
Sorry, I don't use this module nor merged this. Ping the involved people.

ByteMeAsap:
@pedrobaeza , ok, thank you for your reply !!

@thomaspaulb , were you aware of this?

zamberjo:
Hello @ByteMeAsap , can you tell me how to view that view? To check why they are not appearing, thank you.

ByteMeAsap:
Go to any html field (for example: Internal Notes field on Contacts form). Type / , you would get a list of options , from that select Image. Then in the wizard that opens, you would find the Icons tab.

image

zamberjo:
The problem comes from `web_editor/static/src/js/wysiwyg/fonts.js`. The parser used to extract the icons assumes that they come with ::before. We could inherit from this by modifying that regular expression and they would appear. The problem is that I believe **base_fontawesome** should not inherit from **web_editor**.

ByteMeAsap:
@zamberjo , Yes you are right

zamberjo:
I see that **web_editor** has `auto_install = True`, so I think we could add the dependency and fix this regular expression from this same module to keep it updated... What do you think?

ping @thomaspaulb

thomaspaulb:
@zamberjo @ByteMeAsap Agh, sorry for merging this. Though this side-effect of the FA update slipped by multiple people.

I do think because of the `auto_install` flag of `bus` and `web_editor`, practically all Odoo installations that have `base` and `web`, also have `web_editor`. It's only very theoretically that someone might have overridden the autoinstall for whatever reason and have a database without web_editor installed, and then can't use base_fontawesome anymore.

If we'd want to be very sure we could create an auto_install=True glue module `base_fontawesome_web_editor` that modifies the regex to change from `::before` to the new [--fa](https://github.com/OCA/server-tools/blob/17.0/base_fontawesome/static/lib/fontawesome-6.7.2/css/all.css#L2196) that's apparently used.

@zamberjo Would you have time to do that or should I rather revert the merge and we do this slowly?

zamberjo:
I think the best thing to do is to create this new glue module. I'll get started on it.

zamberjo:
You can find it at https://github.com/OCA/server-tools/pull/3423

PR #3388, opened 2025-09-24
[19.0][MIG] base_model_restrict_update: Migration to 19.0

AungKoKoLin1997:
In Odoo 19, has_group() reads a cached `_get_group_ids`() (from `all_group_ids`), so inside a constraint right after changing `group_ids` it can be stale and return the wrong result.
So, I avoided using `has_group()` in a constraint.

Adding a read-only user to System Administrator from the Group form doesn't trigger the constraint method and the error is not raised at the moment.
Same limitation exists in Odoo 19 core.
https://github.com/odoo/odoo/issues/228513

@qrtl QT5088


BhaveshHeliconia (approved):
LGTM!

PR #3374, opened 2025-09-16
 [18.0][FIX] auditlog: fix caching; [ADD] test_auditlog

StefanRijnhart:
Port of https://github.com/OCA/server-tools/pull/3371

Fix caching and add a dedicated test module to run tests on the accounting models.

Fixes creating credit notes with audit rules on account.move.line, writing taxes on products as a user that does not have access to all companies for which there are taxes on the product, and more.

Fixes https://github.com/OCA/server-tools/issues/3308, https://github.com/OCA/server-tools/issues/3224, https://github.com/OCA/server-tools/issues/2561, https://github.com/OCA/server-tools/issues/2554

Includes a port of https://github.com/OCA/server-tools/pull/3364.

PR #3389, opened 2025-09-24
[18.0][FIX] auditlog: Ensure unit tests unpatch methods.

amh-mw:
Though TransactionCase rolls back database changes, it does not call `unlink` on `auditlog.rule` records, meaning that `unsubscribe` is never called to unpatch model methods. This causes subsequent tests in other modules to unexpectedly execute patched methods.

This commit adds the `AuditLogRuleCommon` test class in an attempt to identify and prevent patched methods from contaminating the test environment.

Forward port of 16.0 #3364


auditlog/tests/test_auditlog.py
```diff
@@ -271,11 +272,11 @@ def test_LogDelete(self):
         )
 
 
-class TestAuditlogFull(TransactionCase, AuditlogCommon):
+class TestAuditlogFull(AuditLogRuleCommon, AuditlogCommon):
     def setUp(self):
         super().setUp()
```
StefanRijnhart: If you move this to setUpClass, you can remove the class' now redundant tearDown.

auditlog/tests/test_auditlog.py
```diff
@@ -292,7 +293,7 @@ def tearDown(self):
         super().tearDown()
 
 
-class TestAuditlogExportData(TransactionCase):
+class TestAuditlogExportData(AuditLogRuleCommon):
     def setUp(self):
```
StefanRijnhart: setUpClass here as well, and remove tearDown.

auditlog/tests/test_auditlog.py
```diff
@@ -441,11 +436,11 @@ def test_01_field_and_model_removal(self):
         self.assertFalse(self.auditlog_rule.model_id)
 
 
-class TestAuditlogFullCaptureRecord(TransactionCase, AuditlogCommon):
+class TestAuditlogFullCaptureRecord(AuditLogRuleCommon, AuditlogCommon):
     def setUp(self):
```
StefanRijnhart: setUpClass here as well

StefanRijnhart:
BTW your review on https://github.com/OCA/server-tools/pull/3371 would be much appreciated.

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
Merged with https://github.com/OCA/server-tools/pull/3374.

PR #3398, opened 2025-09-29
[19.0][MIG] attachment_delete_restrict: Migration to 19.0

AungKoKoLin1997:
- Remove the self.env.su condition in unlink, since it is always called with sudo() in Odoo 19.
https://github.com/odoo/odoo/blob/98bae80f57733162c67593539ef47dfd0c6d8797/addons/mail/controllers/attachment.py#L98

@qrtl QT5667


attachment_delete_restrict/models/ir_attachment.py
```diff
@@ -0,0 +1,82 @@
...
+            ]._get_global_delete_attachment_users()
+        if allow_owner_and_admin:
+            users += self.create_uid
+            groups += self.env.ref("base.group_system")
+        allowed_users = groups.user_ids | users
+        if self.env.user not in allowed_users:
+            return self._raise_delete_attachment_error(allowed_users)
+
+    def unlink(self):
+        if self.env.uid == SUPERUSER_ID:
```
yostashiro, line 66: I thought there was a possibility that this change might cause inconveniences in some operational scenarios.

@sebastienbeau Do you have any concerns on this? If you do, we might revert this change and additionally insert a check at the beginning of `mail_attachment_delete()`.

PR #3405, opened 2025-10-01
[18.0][IMP] base_exception: Allow filtering records before exception evalua…

grindtildeath:
…tion

### DIFF
--- a/base_exception/models/exception_rule.py
+++ b/base_exception/models/exception_rule.py
@@ -22,7 +22,12 @@ class ExceptionRule(models.Model):
     name = fields.Char("Exception Name", required=True, translate=True)
     description = fields.Text(translate=True)
     sequence = fields.Integer(help="Gives the sequence order when applying the test")
-    model = fields.Selection(selection=[], string="Apply on", required=True)
+    model = fields.Selection(selection=[], string="Apply on model", required=True)
+    filter_domain = fields.Char(
+        string="Apply on records",
+        default="[]",
+        help="If present, this condition must be satisfied to trigger the exception.",
+    )
 
     exception_type = fields.Selection(
         selection=[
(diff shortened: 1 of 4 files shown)


base_exception/models/exception_rule.py
```diff
@@ -22,7 +22,12 @@ class ExceptionRule(models.Model):
     name = fields.Char("Exception Name", required=True, translate=True)
     description = fields.Text(translate=True)
     sequence = fields.Integer(help="Gives the sequence order when applying the test")
-    model = fields.Selection(selection=[], string="Apply on", required=True)
+    model = fields.Selection(selection=[], string="Apply on model", required=True)
+    filter_domain = fields.Char(
+        string="Apply on records",
+        default="[]",
+        help="If present, this condition must be satisfied to trigger the exception.",
```
simahawk, line 29:
```suggestion
        help="If present, this condition must be satisfied to trigger the exception, no matter the Exception type.",
```

simahawk (approved):
LGTM. Minor remark only.

hparfr (commented):
Hi,

Can you explain why / how to use in the usage.md.

I guess it's a kind of a pre-filter to skip the test of exception on some records. Is it ?

leemannd (approved):

grindtildeath:
@hparfr Yes it's a pre filter to segregate records from a same model where an exception could apply or not.

However, there are no configuration instructions or USAGE in the readme of this module, so I'm not sure it would make much sense to describe this filter when there isn't any other instructions on how to configure exceptions... :shrug:

hparfr:
>However, there are no configuration instructions or USAGE in the readme of this module, so I'm not sure it would make much sense to describe this filter when there isn't any other instructions on how to configure exceptions... 🤷

It's never too late add some doc. At least a small word the source code can be good enough. 

The reason you want to add this filtering feature was not obvious to me because I don't see the benefit in my case (mainly detect by domain exceptions): the filtering is already done with the domain.

If I understand correctly, the prefilter has positive impact on perfs for detecting exception by python code and negative for detecting exceptions by domain.

TDu (approved):

PR #3409, opened 2025-10-06
[16.0][FIX] auditlog: swap out tocompute in ThrowAwayCache as well

StefanRijnhart:
Port of #3407 

# Steps to Reproduce:

Install auditlog and purchase_tier_validation. Create an auditlog rule on the Tier Exception model. Create a new tier exception record for the purchase order model without any fields. In an export, check the value for the stored computed field `model_name`.

# Desired outcome:

The value for `model_name` is 'purchase.order'.

# Actual outcome:

There is no value for the stored computed field `model_name`. When trying to edit the `Fields` model, which depends on a valid value for this field, an error occurs saying 'Compute method failed to assign
tier.validation.exception(1,).valid_model_field_ids'.

# Analysis:

When `read` is executed under the scope of the new `ThrowAwayCache` decorator (#3374), recomputes may be performed and dequeued from `self.env.cache.transaction.tocompute`. To make sure the recomputed values still end up in the non-volatile as well, we need to preserve the tocompute attribute from the original transaction.

### DIFF
--- a/auditlog/models/rule.py
+++ b/auditlog/models/rule.py
@@ -2,9 +2,11 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import copy
+from collections import defaultdict
 
 from odoo import _, api, fields, models, modules
 from odoo.exceptions import UserError
+from odoo.tools.misc import OrderedSet
 
 FIELDS_BLACKLIST = [
     "id",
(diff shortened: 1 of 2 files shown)


etobella (approved):
Tested with Purchase Orders and it works nicely

pedrobaeza (approved):
Let's proceed to solve the critical cache problem. I think the implementation of this is dubious, as it touches very sensitive parts of the ORM, but let's fix it at less in its current state.

/ocabot merge patch

PR #3412, opened 2025-10-08
[14.0][FIX] module_analysis: The 'code' attribute does not exist in SourceAnalysis Class

adasatorres:
Hi, since Pygount doesn’t currently have a pinned version, some newer releases no longer include the 'code' attribute.I propose pinning this add-on to version 1.8.0.


legalsylvain (changes requested):
please don't pin the version. I don't think it's necessary. 
@fcayre implementation is more flexible : https://github.com/OCA/server-tools/pull/3378

legalsylvain:
Thanks ! 

/ocabot merge patch

PR #3413, opened 2025-10-11
[19.0][MIG] session_db 

sbidoul:


FrancoMaxime (approved):
LGTM: code review

session_db/pg_session_store.py
```diff
@@ -0,0 +1,216 @@
...
+    @with_cursor
+    def delete_from_identifiers(self, identifiers: list[str]) -> None:
+        for identifier in identifiers:
+            if not http._session_identifier_re.match(identifier):
+                raise ValueError(
+                    "Identifier format incorrect, "
+                    "did you pass in a string instead of a list?"
+                )
+            self._cr.execute(
+                "DELETE FROM http_sessions WHERE sid LIKE %s||'%%'", (identifier,)
```
sbidoul, line 195: This would not be correct if `identifier` can contain a `%` but I don't think that is possible?

thomaspaulb: Looks like the identifier always has to have [alphanumerical characters](https://github.com/OCA/OCB/blob/736534d7d00a91896c4f31d35eca1cfe994dedf1/odoo/http.py#L941)

thomaspaulb (approved):

PR #3415, opened 2025-10-14
[IMP] session_db: add method delete_from_identifiers

maq-adhoc:
Added the delete_from_identifiers method to PGSessionStore.
This method allows bulk deletion of session records from the http_sessions table based on a list of session identifiers (sid). It is triggered from the backend via the "Revoke" button

### DIFF
--- a/session_db/pg_session_store.py
+++ b/session_db/pg_session_store.py
@@ -149,6 +149,19 @@ class PGSessionStore(sessions.SessionStore):
             (f"{max_lifetime} seconds",),
         )
 
+    @with_lock
+    @with_cursor
+    def delete_from_identifiers(self, identifiers: list[str]) -> None:
+        for identifier in identifiers:
+            if not sessions._sha1_re.match(identifier):
+                raise ValueError(
+                    "Identifier format incorrect, "
+                    "did you pass in a string instead of a list?"
+                )
+            self._cr.execute(
+                "DELETE FROM http_sessions WHERE sid LIKE %s||'%%'", (identifier,)
+            )
+
 
 _original_session_store = http.root.__class__.session_store
 
--- a/base_search_fuzzy/i18n/sl.po
+++ b/base_search_fuzzy/i18n/sl.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-12-01 02:10+0000\n"
-"PO-Revision-Date: 2023-04-11 13:22+0000\n"
+"PO-Revision-Date: 2026-07-25 09:07+0000\n"
 "Last-Translator: Matjaz Mozetic <matjaz@luxim.si>\n"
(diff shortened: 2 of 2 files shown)


sbidoul:
Ah I had not noticed `delete_from_identifiers` was there in 18.0 too.

If I'm not mistaken, the identifiers provided to that method are the first 42 characters of the sid.

It handled that in the 19.0 migration. You may want to review https://github.com/OCA/server-tools/pull/3413 and then extract the relevant part for 18.

maq-adhoc:
@sbidoul Perfect, I backported 19. Regards

sbidoul (approved):

session_db/pg_session_store.py
```diff
@@ -149,6 +149,19 @@ def vacuum(self, max_lifetime=http.SESSION_LIFETIME):
             (f"{max_lifetime} seconds",),
         )
 
+    @with_lock
+    @with_cursor
+    def delete_from_identifiers(self, identifiers: list[str]) -> None:
+        for identifier in identifiers:
+            if not sessions._sha1_re.match(identifier):
```
sbidoul, line 156: _sha1_re is not the same regex as _session_identifier_re. Why do you change this?

maq-adhoc: hi @sbidoul 
The module imports the sessions class from odoo.tools._vendor import sessions.

The generate_key method it calls is this one: https://github.com/odoo/odoo/blob/18.0/odoo/tools/_vendor/sessions.py#L33-L37 which creates a SHA (40 characters) and not a 42-character identifier.

I understand that we could modify the FilesystemSessionStore class to inherit from FilesystemSessionStore (https://github.com/odoo/odoo/blob/18.0/odoo/http.py#L911) and ensure consistent behavior. However, we would have to remove the validations to avoid errors with existing sessions.

Since we are on a stable version with many productive installations, I prefer to be conservative with the changes. I look forward to hearing your opinion.

sbidoul: Ah, I see. And in 19, we don't have that problem because we use the generate_key method from FileSystemSessionStore here: https://github.com/OCA/server-tools/blob/7937302b030748908bfcf01a15ad66e357544024/session_db/pg_session_store.py#L142

Is that right?

So maybe we should do the same here, and support both regexes in delete_from_identifiers?

tarteo: Why should we validate the input anyways? 

Isn't it faster to do an IN operation?
https://github.com/OCA/server-tools/pull/3662/changes#diff-d813321f73ce1687539ce0923c3e35057a0c7c2be7ae3f11ae43a30bcf6cd1bbR158-R166

tarteo:
@maq-adhoc Can you take a look at https://github.com/OCA/server-tools/pull/3662 and review if you have time?
This PR solves multiple issues, and the implementation it consistent with 19.0. 
I resused a snippet (`sessions._sha1_re.match(identifier):`) from this PR, so I will add you to the contributors if you like.
It also includes a few additional tests covering these methods.

PR #3423, opened 2025-10-23
[17.0][ADD] base_fontawesome_web_editor: Ensure FontAwesome >= 6.7.2

zamberjo:


pedrobaeza:
You can remove the glue part, as that's implied when you concatenate both module names.

ByteMeAsap:
Looks good !! But there's one more issue which can be addressed along with this, I think. Right now, if you search for icons, you would find that there are a lot of duplicate icons. For example, you can search for life, you would get multiple icons as can be seen in the following screenshot.
image

We should eliminate the duplicate ones

zamberjo:
I have modified the code to make it more readable. Can you check that it is now correct?

thomaspaulb:
Functionally this seems all good on Runboat now:

image

thomaspaulb (approved):

ByteMeAsap:
@zamberjo , looks good. Can you just change the commit message to [ADD] base_fontawesome_web_editor

zamberjo:
done

ByteMeAsap (approved):
LGTM!!

ByteMeAsap:
@zamberjo , would you like to migrate these changes to v18 or I should pick it up?

zamberjo:
It will take me a while to find the time to do it, so if you can take care of it, that would be perfect.

ByteMeAsap:
@thomaspaulb , can you please merge this one, so that I can start working on its migration to v18

PR #3425, opened 2025-10-23
[16.0][ADD] base_notification

kobros-tech:
if a sales order is under many validation steps sometimes the nitifactions are sent, but if the order is following different routes not perfectly that is coded we will not be notified.

The approach to have a precise notification handling methods based on all probable things ORM like create, update, delete.


kobros-tech:
image

kobros-tech:
image

kobros-tech:
image

kobros-tech:
@sebastienbeau
@mymage
@sbidoul
@wlin-kencove

I need your opinion about this new module?

kobros-tech:
image

dreispt:
Why not add a notification action option to the existing Automated Actions?

kobros-tech:
> Why not add a notification action option to the existing Automated Actions?

Thanks MR. Daniel, I was asking myself what is the difference now after making it.

I will extend the automated action in a new PR to compare and to review.

hussain:
This is nice! We have built a similar functionality (still in progress) that is actually connecting to new unified notification system (email, SMS...etc).

One important point to mention, ensure you have safe net around core ORM calls i.e. extending create/write/unlink using `self.env["base.notification.rule"].sudo()._apply_trigger("on_create", records)` should have try/catch so it does not block ORM calls, catch can use _logger to indicate issue silently. Otherwise any tiny issue would cause ORM to fail simple operation.

kobros-tech:
hello there

@simahawk
@JordiBForgeFlow

PR #3426, opened 2025-10-24
[18.0][UPD] base_fontawesome: Update FontAwesome to version 6.7.2

ByteMeAsap:

PR #3427, opened 2025-10-24
[18.0][MIG] base_fontawesome_web_editor: Migration to 18.0

ByteMeAsap:


ByteMeAsap:
@zamberjo , @thomaspaulb can you review this

zamberjo (approved):
Technical review done, I don't have any projects in Odoo 18.0 to test it along with the other PR.

thomaspaulb:
Quick functional review on Runboat seems to show that it's not working yet:

image

thomaspaulb:
Ah wait, it's because both PR's are needed for a functional check. 

In this case I would prefer to have both of them in one PR, with individual commits.

@pedrobaeza Do you have something against that?

pedrobaeza:
Well, this should follow the logical order: first having the initial module tested and merged, and then this one, so no need to join two PRs together.

thomaspaulb:
@pedrobaeza OK but we have two things here:

- The other PR which introduces a change to base_fontawesome; however on its own it will break web_editor
- This PR which contains a glue module which addresses that breakage

I mean in this case especially if both are merged quickly it doesn't really matter, but I'm just asking about a general case like this - if the glue module is needed for the original module to not break things, wouldn't it make sense to bundle them? Then the whole package can be tested on Runboat together.

pedrobaeza:
I don't think so, as the initial module should put the problem in the roadmap, and note that for most of its users, they don't care, being the problem there for years without noticing. 

In general, roadmap with the problems and indications to install the extra module is the way to go.

thomaspaulb:
@pedrobaeza Not sure if I agree:

- The problem is currently not there in `base_fontawesome`, but once the upgrade is done to 6.7.2 (eg https://github.com/OCA/server-tools/pull/3426 is merged) then this will introduce the problem in web_editor eg icons can't be found anymore. This was the same problem in 17.0 where @ByteMeAsap complained that after the merge the module had broken
- To then fix that would mean to merge the second PR and all is good again

I agree it's a technical discussion because in all likelyhood there is only a few days between these two merges but this temporary situation of a bug being introduced could be avoided by having both modules in one PR

pedrobaeza:
Well, I was talking for the general case. For this one, let's merge both at the same time for not having to make more changes. Are they ready?

thomaspaulb:
Well I wanted to have them in the same PR so that I could save time by just testing on Runboat ^^

But let me install them locally instead and see

thomaspaulb (approved):
Functional review

PR #3430, opened 2025-10-27
[19.0][MIG] field_vector: Migration to 19.0

hailangvn:
This is to migrate field_vector module to 19.0.


nilshamerlinck:
fyi @lmignon

field_vector/readme/CREDITS.md
```diff
@@ -1,4 +0,0 @@
-The development of this module has been financially supported by:
```
lmignon: @hailangvn Why do you remove the credits?

field_vector/fields.py
```diff
@@ -122,31 +122,16 @@ class Vector(fields.Field):
...
-        autopad=fields.Default,
-        string=fields.Default,
-        **kwargs,
-    ):
-        super().__init__(
-            dimensions=dimensions, string=string, autopad=autopad, **kwargs
-        )
-
-    def _setup_attrs(self, model_class, name):
-        res = super()._setup_attrs(model_class, name)
```
lmignon: This code must remain but be adapted...

```python

from odoo.tools.misc import SENTINEL, Sentinel,

....

class Vector(fields.Field):

    def __init__(self, 
        dimensions: int|Sentinel=SENTINEL,
        autopad: bool | Sentinel=SENTINEL,
        string: str | Sentinel = SENTINEL,
        **kwargs
    ):
        super().__init__(
            dimensions=dimensions, string=string, autopad=autopad, **kwargs
        )

File: field_vector/fields.py [hunk 3]
```diff
@@ -122,31 +122,16 @@ class Vector(fields.Field):
...
-            self.dimensions == fields.Default
-            or self.dimensions is None
-            or not isinstance(self.dimensions, int)
-        ):
+    def _setup_attrs__(self, model_class, name):
+        res = super()._setup_attrs__(model_class, name)
+        if self.dimensions is None or not isinstance(self.dimensions, int):
             raise ValueError(
                 "The size of the vector field must be an integer and cannot be None."
             )
```
lmignon:
```python
    if not isinstance(self.dimensions, int) or self.dimensions <= 0:
        raise ValueError(
            "The size of the vector field must be a positive integer."
        )
```

lmignon (changes requested):
Thank you for the migration @hailangvn 
Some comments about your changes...

field_vector/fields.py
hailangvn: Thanks for your suggestion. But the code does nothing. The module can work without it anyway.

lmignon: @hailangvn This code is not useless it's required....
* It documents the method.
* It allows IDEs to suggest the expected parameters.
* Most importantly, it is the basis of the field definition overload mechanism.

Regarding this last point, in Odoo it is possible to redefine a field in module B declared in module A. When redefining, normally only the parameters that change are passed. Therefore, there must be a way to know in the field constructor which values have actually been passed and to dissociate them. This entire mechanism is based on the special value SENTINEL.

lmignon (changes requested):
see my last comment

lmignon (approved):
Thank you @hailangvn for the migration. LGTM (Code review)

field_vector/readme/CONFIGURE.md
```diff
@@ -0,0 +1,10 @@
+[ This file is not always required; it should explain **how to configure the module before using it**; it is aimed at users with administration privileges. 
+
+Please be detailed on the path to configuration (eg: do you need to activate developer mode?), describe step by step configurations and the use of screenshots is strongly recommended.]
+
+
+To configure this module, you need to:
+
+- Go to *App* > Menu > Menu item
+- Activate boolean… > save
```
petrus-v: I suppose this file could be removed or filled, it seems to come from a template (I've seen that come from the original module)

petrus-v (approved):
Code looks good to me !

Thanks

field_vector/readme/CONFIGURE.md
hailangvn: Thanks for your review. The PR has been updated.

petrus-v (approved):

etobella (approved):

PR #3431, opened 2025-10-28
[ADD] base_temporary_action - a module to create temporary actions

baimont:
This addon allows to create temporary actions that will be removed by a cron. Temporary actions are identified by a field 'is_temporary' and removed by the cron every one hour. The deletion cron is configurable.


base_temporary_action/models/ir_actions_actions.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import fields, models
+
+
+class IrActionsActions(models.Model):
+    _inherit = "ir.actions.actions"
+
+    is_temporary = fields.Boolean()
+
+    def cron_remove_temporary_actions(self):
```
sbidoul:
```suggestion
    @api.model
    def cron_remove_temporary_actions(self):
```

sbidoul: But actually, `@api.autovacuum` is simpler than a cron.

sbidoul (commented):
Can you remove the OCA icon? It is obsolete and the bot creates it automatically if not present (see also https://github.com/acsone/bobtemplates.odoo/pull/95).

base_temporary_action/models/ir_actions_actions.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @api.autovacuum
+    def _gc_temporary_actions(self):
+        self.search(
+            [
+                ("is_temporary", "=", True),
+                (
+                    "create_date",
+                    "<",
+                    datetime.now() - relativedelta(hours=1),
```
sbidoul:
```suggestion
                    odoo.fields.Datetime.now() - relativedelta(hours=1),
```

To be explicit that we want the UTC datetime Odoo uses. In practice I think `datetime.datetime.now()` is patched by Odoo to return UTC, but I prefer the explicit version.

sbidoul (approved):

base_temporary_action/readme/CONTEXT.md
```diff
@@ -0,0 +1,5 @@
+We wanted to be able to create a new quotation with default values coming from an external system using a redirect.
```
sbidoul:
```suggestion
We wanted to be able to open a new quotation form (without creating it in the database) with default values coming from an external system using a GET route and redirect.
```

base_temporary_action/readme/CONTEXT.md
```diff
@@ -0,0 +1,5 @@
...
+
+We couldn't find a way to pass our values to a new quotation using Odoo standard.
```
sbidoul:
```suggestion
We couldn't find a way to pass our values to a new quotation form using Odoo standard.
```

base_temporary_action/models/ir_actions_actions.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @api.autovacuum
+    def _gc_temporary_actions(self):
+        self.search(
+            [
+                ("is_temporary", "=", True),
+                (
+                    "create_date",
+                    "<",
+                    fields.Datetime.now() - relativedelta(hours=1),
```
sbidoul: A better default could be the same age as transient model vacuum?

qgroulard (approved):
LGTM (code review)

sbidoul:
@baimont can you change it to LGPL? No reason to use AGPL for such a simple low level module.

baimont:
> @baimont can you change it to LGPL? No reason to use AGPL for such a simple low level module.

with delight

PR #3434, opened 2025-11-03
[FIX] tracking_manager: Fixed wrong usage of create function

baguenth:
Using the wrong parameter in create function leads to, that globally all create-functions called with parameter "vals_list" will break. 

TypeError: Base.create() got an unexpected keyword argument 'vals_list'

### DIFF
--- a/tracking_manager/models/models.py
+++ b/tracking_manager/models/models.py
@@ -165,8 +165,8 @@ class Base(models.AbstractModel):
         return super().write(vals)
 
     @api.model_create_multi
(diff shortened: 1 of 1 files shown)


baguenth:
Hey @rvalyi @sebastienbeau @Kev-Roche @pedrobaeza,

can someone take a look? Should be a straight forward fix.

pedrobaeza (approved):
Yes, it's clear, so fast-tracking it:

/ocabot merge patch

Issue #3435, opened 2025-11-04
[18.0]mail_cleanup: stacktrace when testing locally


leemannd:
## Module
mail_cleanup

## Describe the bug
The below stacktrace is promted when doing a "Fetch now" on an incoming mail server

```
RPC_ERROR

Odoo Server Error

Occured on localhost:8888 on model fetchmail.server on 2025-11-04 10:00:24 GMT

Traceback (most recent call last):
  File "/odoo/src/odoo/http.py", line 2144, in _transactioning
    return service_model.retrying(func, env=self.env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/service/model.py", line 156, in retrying
    result = func()
             ^^^^^^
  File "/odoo/src/odoo/http.py", line 2111, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/http.py", line 2359, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/external-src/odoo-cloud-platform/monitoring_prometheus/models/ir_http.py", line 38, in _dispatch
    res = super()._dispatch(endpoint)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/addons/base/models/ir_http.py", line 333, in _dispatch
    result = endpoint(**request.params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/http.py", line 754, in route_wrapper
    result = endpoint(self, *args, **params_ok)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/addons/web/controllers/dataset.py", line 42, in call_button
    action = call_kw(request.env[model], method, args, kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/odoo/src/odoo/api.py", line 535, in call_kw
    result = getattr(recs, name)(*args, **kwargs)
[...]

PR #3436, opened 2025-11-04
[18.0][FIX] mail_cleanup: command CLOSE illegal in state AUTH, only allowed i…

leemannd:
…n states SELECTED

Fixing: https://github.com/OCA/server-tools/issues/3435

You need to have a `server.select()`. These selects are done into the methods `_cleanup_fetchmail_server` and `_purge_fetchmail_server` if you don't match any of the IF statements. You will never do the `server.select()` and thus endup with a stacktrace. 

https://github.com/OCA/server-tools/blob/18.0/mail_cleanup/models/fetchmail_server.py#L125
https://github.com/OCA/server-tools/blob/18.0/mail_cleanup/models/fetchmail_server.py#L128C30-L128C53

### DIFF
--- a/mail_cleanup/models/fetchmail_server.py
+++ b/mail_cleanup/models/fetchmail_server.py
@@ -51,7 +51,6 @@ class FetchmailServer(models.Model):
         expiration_date = datetime.date.today()
         expiration_date -= relativedelta(days=server.cleanup_days)
         search_text = expiration_date.strftime("(UNSEEN BEFORE %d-%b-%Y)")
-        imap_server.select()
         result, data = imap_server.search(None, search_text)
         for num in data[0].split():
             try:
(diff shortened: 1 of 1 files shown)


cyrilmanuel (approved):

mail_cleanup/models/fetchmail_server.py
```diff
@@ -122,6 +122,7 @@ def fetch_mail(self, raise_exception=True):
             if server.server_type == "imap":
                 try:
                     imap_server = server.connect()
+                    imap_server.select()
```
sebalix, line 123: No need to remove this instruction from `_cleanup_fetchmail_server` and `_purge_fetchmail_server`? No side-effect to keep them?

leemannd: Hello, no side effect to keep them. But I will remove them to be more consistent.

leemannd: Hello @sebalix A review update is welcome

sebalix (approved):

mail_cleanup/models/fetchmail_server.py
sebalix: Hello @leemannd , approved :)

PR #3442, opened 2025-11-07
[ADD] A new module that is dependency map view its a new view type to…

faizanlodhi:
### Summary
This PR adds a new module `dependency_map_view`, which introduces a new view type allowing users to visualize dependencies between records in a graph or map layout.

### Features
- Displays relationships between models/records.
- Helps track record-level dependencies visually.
- Integrates seamlessly with existing Odoo models.

### Usage
Once installed, users can open the Dependency Map view from the corresponding model views to explore related data.

### Task/Issue
N/A (New feature proposal)


legalsylvain:
Hi @faizanlodhi. Thanks for sharing this module.
I made some tests on runboat, and I don't understand the use case of this feature. 
image


could you elaborate in the USAGE.md file, and add some screenshot ? 

thanks ! 

(also, could you squash your commits ?)

dependency_map_view/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.pyc
```
legalsylvain: useless file to remove.

dependency_map_view/hooks.py
```diff
@@ -0,0 +1,49 @@
...
+'dependency_map' view mode to all existing window actions in the system.
+
+The hook ensures that the dependency map view is available across all models
+without requiring manual configuration for each action.
+"""
+
+
+def post_init_hook(env):
+    """
+    Add dependency_map view mode to all existing window actions.
```
legalsylvain: I see it more as an optional action. I don't think all people want to use this new view in all models.

dependency_map_view/models/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+Modules:
+    - ir_ui_view: Extends ir.ui.view to add 'dependency_map' as a new view type
+    - ir_actions: Extends ir.actions.act_window to automatically add dependency
+                  map view to actions for models with relational fields
+
+
+"""
+
+# Import view model extension
```
legalsylvain: all comments can be removeds.
```
```

legalsylvain (commented):
Hi. I Guess you used AI to generate the code ? As a result, there is a lot of totally useless comments. Could you clean that ? 

thanks in advance.

PR #3437, opened 2025-11-05
17.0 [IMP] auditlog: Modernize http_session display_name computation

drkpkg:
- Replace and/or pattern with ternary operator (PEP 308)
- Replace .format() with f-strings (PEP 498)
- Add explicit create_date None check for edge cases
- Remove deprecated fields.Datetime.from_string() call
- Add bulk server actions for subscribe and unsubscribe
- Add base pot file for new translations

The old and/or pattern can fail with falsy but valid values. The ternary operator is safer and more readable.

PR #3432, opened 2025-10-28
17.0 [FIX] auditlog: Unable to create a user session

david-s73:
- When attempting to create a new user session, an error message popped up, so I added that it should take the current date if no date is provided.

- The error message copied from a runboat 

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 1971, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 152, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1999, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 2203, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 221, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 786, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 25, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 21, in _call_kw
    return call_kw(Model, method, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 484, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/odoo/api.py", line 469, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/addons/web/models/models.py", line 1083, in onchange
    snapshot1 = RecordSnapshot(record, fields_spec)
  File "/opt/odoo/addons/web/models/models.py", line 1170, in __init__
    self.fetch(name)
  File "/opt/odoo/addons/web/models/models.py", line 1185, in fetch
    self[field_name] = self.record[field_name]
  File "/opt/odoo/odoo/models.py", line 6695, in __getitem__
    return self._fields[key].__get__(self, self.env.registry[self._name])
  File "/opt/odoo/odoo/fields.py", line 1152, in __get__
    self.recompute(record)
  File "/opt/odoo/odoo/fields.py", line 1379, in recompute
    apply_except_missing(self.compute_value, recs)
  File "/opt/odoo/odoo/fields.py", line 1352, in apply_except_missing
    func(records)
[...]

PR #3443, opened 2025-11-07
[17.0][IMP] Add bulk subscribe actions for auditlog rules

drkpkg:
- Add server actions to trigger subscribe/unsubscribe from list view
- Ensure return type suits server action flow
- Support bulk handling without interfering with form buttons


drkpkg:
@StefanRijnhart after the rebase from #3432 this feature is the only left.
This bulk feature is useful when you have a large number of rules to subscribe or unsubscribe.

Note: #3437 is discarded after the rebase from #3432.

StefanRijnhart (commented):
Thanks for the rebase!
Points of order:
* Next time, maybe keep the original PR, updating the title manually.
* `[FEAT]` does not conform to Odoo commit message prefixes (https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#tag-and-module-name), suggest `[IMP]`.
* Please add the module name in the commit message, like `[IMP] auditlog: add bulk subsccribe actions for rules`.
* Please squash commits into one as they are intimately related.

Code review:
* Returning another action with soft-reload is not necessary. Odoo will already update the rules statuses without it. That means we can avoid adding the new methods as you can just call the existing subscribe/unsubcribe ones.
* The original subscribe method is not reentrant. It will happily create duplicate window actions to set as the rule's `action_id`. Given that the UI now allows to call `subscribe` on subscribed records, can you fix this?

StefanRijnhart (changes requested):
Thanks for the update! Remaining points:

* Please squash commits into one as they are intimately related.
* Don't add additional methods but make original methods reentrant.

PR #3446, opened 2025-11-13
[18.0][IMP] auditlog: Allow developers to by-pass the auditlog.

baptiste-n42:
In high-volume automated processes, such as batch data imports or recalculations (e.g., journal entries, inventory valuation), the performance overhead of audit logging can lead to thread timeouts and job failures.

This feature introduces a way to programmatically disable logging during these specific operations to ensure process stability and completion.


StefanRijnhart (commented):
Developers may think that developers do not need to be audited, but I know managers have a different opinion sometimes!

baptiste-n42:
I’m not really sure in what situations developers themselves would need to use Odoo directly. Anyway, this feature is meant for very specific cases where it’s really needed, and the code using it would still be reviewed and traceable (in most companies).

Things like COGS computation, AVCO, etc. are already transaction heavy jobs on large datasets. When you add audit logging on top, the extra overhead can make the job fail before OdooSh even lets it finish.

@StefanRijnhart, do you feel this use case is too marginal to justify inclusion in the codebase?

StefanRijnhart:
I think it goes against the idea of audit logging! As a developer you can implement a patch locally such as the one you are proposing here, but I would prefer not to include it by default.

baptiste-n42:
Ok, I will close the PR. Thanks for the answer @StefanRijnhart

PR #3460, opened 2025-12-10
[18.0][FIX] fetchmail_attach_from_folder: add missing keyword argument

adrip-s73:
When manually running mail.ir_cron_mail_gateway_action "https://github.com/odoo/odoo/blob/18.0/addons/mail/data/ir_cron_data.xml" , the following error occurs because the base function has the argument set. "https://github.com/odoo/odoo/blob/18.0/addons/mail/models/fetchmail.py"

Error:
image


fetchmail_attach_from_folder/models/fetchmail_server.py
```diff
@@ -66,10 +66,12 @@ def onchange_server_type(self):
         self.state = "draft"
         return result
 
-    def fetch_mail(self):
+    def fetch_mail(self, raise_exception=True):
```
NL66278: instead of `raise_exception=True` put `**kwargs` here.

fetchmail_attach_from_folder/models/fetchmail_server.py
```diff
@@ -66,10 +66,12 @@ def onchange_server_type(self):
...
         result = True
         for this in self:
             if not this.folders_only:
-                result = result and super(FetchmailServer, this).fetch_mail()
+                result = result and super(FetchmailServer, this).fetch_mail(
```
NL66278: and `**kwargs` here. We are not doing anything with any keyword argument here, so using '**kwargs` will pass any keyword argument in the invocation of this method unchanged to the super() method. This will keep working, no matter how many keyword arguments are added or changed in the future.

NL66278 (changes requested):
See comment on using `**kwargs`

NL66278:
@adrip-s73 Could you please take up the requested changes?

fetchmail_attach_from_folder/models/fetchmail_server.py
adrip-s73: done !!

fetchmail_attach_from_folder/models/fetchmail_server.py
adrip-s73: done!!

NL66278 (approved):
👍 LGTM Thanks for your contribution!

PR #3464, opened 2025-12-11
[18.0][MIG] module_analysis: Migration to 18.0

adasatorres:
Standard migration.


legalsylvain (approved):
Diff review. Lgtm.

module_analysis/models/ir_module_type_rule.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2019-Today: GRAP (<http://www.grap.coop/>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class IrModuleType(models.Model):
```
rrebollo:
```suggestion
class IrModuleTypeRule(models.Model):
```

module_analysis/models/ir_module_type_rule.py
```diff
@@ -0,0 +1,29 @@
...
+    _name = "ir.module.type.rule"
+    _description = "Modules Types Rules"
+    _order = "sequence"
+
+    sequence = fields.Integer()
+
+    module_domain = fields.Char(required=True, default="[]")
```
rrebollo, line 16:
```suggestion
    module_domain = fields.Binary(required=True, default=[])
```
I'm not sure if this would require a migration script.

module_analysis/models/ir_module_type_rule.py
```diff
@@ -0,0 +1,29 @@
...
+
+    module_type_id = fields.Many2one(
+        string="Module type", comodel_name="ir.module.type", required=True
+    )
+
+    def _get_module_type_id_from_module(self, module):
+        IrModuleModule = self.env["ir.module.module"]
+        for rule in self:
+            domain = safe_eval(rule.module_domain)
```
rrebollo: If you change it to `Binary` this wouldn't apply

rrebollo (changes requested):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit. Nevertheless I think you should address the duplicate class name for different odoo models.

module_analysis/models/ir_module_type_rule.py
adasatorres: @rrebollo It’s not possible to switch to a binary field because the domain cannot be applied at the view level.

module_analysis/models/ir_module_type_rule.py
adasatorres: @rrebollo I agree.

rrebollo (approved):
Code Review. LGTM!

edescalona (approved):
LGTM

rvalyi (approved):

PR #3479, opened 2026-01-02
[18.0] [IMP] excel_import_export: Code for Compatibility with xlrd 2.0+ and fix date handling

BhaveshHeliconia:
This update fixes a few issues we encountered when running with newer versions of xlrd (2.0 and above), which dropped support for .xlsx files. The existing fallback mechanism wasn't catching this case correctly, causing import failures in updated environments.

I've improved the file detection logic to explicitly check for the file signature (PK header), which allows us to correctly route .xlsx files to openpyxl when needed, while safely keeping the traditional xlrd behavior for legacy .xls files.

I also took the opportunity to clean up how we handle date fields. I noticed that midnight timestamps (like "2025-01-01 00:00:00") were sometimes being passed to Date fields, causing Odoo to reject the import with "unconverted data" errors. I've updated the value extraction logic to enforce the correct format based on the field type—stripping the time component for Date fields while preserving it for Datetime fields.


BhaveshHeliconia:
@pablo-cort-s73 @Reyes4711-S73 @xaviedoanhduy @wlin-kencove Could you please review this? Thanks!

Reyes4711-S73 (approved):
LGTM

PR #3480, opened 2026-01-02
[ADD] base_sparse_field_jsonb

bosd:
References
·	PostgreSQL JSONB documentation: https://www.postgresql.org/docs/current/datatype-json.html
·	GIN index documentation: https://www.postgresql.org/docs/current/gin.html
·	OCA attribute_set: https://github.com/OCA/odoo-pim
·	Related issue: https://github.com/OCA/odoo-pim/issues/153


rvalyi:
@bosd I'm not sure I understand why you depend on attribute_set here. You want to build a supercharged attribute_set module with JSONB, right? That sounds legit but should this be done inside the server-tools base_sparse_fields_jsonb module?  Wouldn't it be better to leave base_sparse_fields_jsonb without attribute_set dependency and create another attribute_set_jsonb module in OCA/odoo-pim or make attribute_set depend on base_sparse_fields_jsonb eventually?

Disclaimer I didn't read your code carefully yet.

bosd:
>  Wouldn't it be better to leave base_sparse_fields_jsonb without attribute_set dependency and create another attribute_set_jsonb module in OCA/odoo-pim or make attribute_set depend on base_sparse_fields_jsonb eventually?

@rvalyi I was indeed contemplating about this.  Placing it in a separeate module in the pim repo would keep this base implementation more basic, clean and flexible. But there are some more modules / improvements to get to a full working stack. Wanted to prevent, that we end up with a lot of glue modules and puzzle to get it up and running. 

I will later separate the latest commits and place it in their own module.

_This module as it sits here, could potentially be merged into base_sparse, but for now i think it is better to keep it separate._

rvalyi:
@bosd this starts to look pretty solid. kudos for the work! I would say what is missing for now are some tests inside this module. You can probably adapt some tests from the original module https://github.com/odoo/odoo/blob/19.0/addons/base_sparse_field/tests/test_sparse_fields.py 
Tell us when it's ready for review.   cc @sebastienbeau

bosd:
@rvalyi , @sebastienbeau , cc: @alexis-via  It is ready for review!

bosd:
@sebastienbeau Can you please review?

PR #3391, opened 2025-09-25
[19.0][MIG] auditlog: Migration to 19.0

lembregtse:

PR #3483, opened 2026-01-02
[19.0][MIG] auditlog: Migration to 19.0

StefanRijnhart:
Continued from https://github.com/OCA/server-tools/pull/3391

* Regular API changes and improved tests taken from #3391
* Adapt ThrowAwayCache to modified caching internals in Odoo. Since https://github.com/odoo/odoo/commit/670897c6dc8e7, `env.transaction.cache` is no longer a data store.
* Redistribute code and views into one file per model. Update view file names to fit the convention.
* Trade slightly misleading 'Subscribe' terminology by 'Confirm'.
* Small improvements to help strings and readme.
* Renewed screenshot images.
* Squashed a number of administrative commits.
* Improved test coverage further.
* Fix issue where inherited test methods are not run since Odoo 18 (backported in https://github.com/OCA/server-tools/pull/3485)


StefanRijnhart:
`test_auditlog` for 19.0 is green at https://github.com/OCA/server-tools/pull/3484

auditlog/models/rule.py
```diff
@@ -59,38 +62,67 @@ class ThrowAwayCache:
...
+        self._transaction.field_data_patches = defaultdict[
+            "Field", defaultdict["IdType", list["IdType"]]
+        ](lambda: defaultdict(list))
+        self._transaction.field_dirty = defaultdict["Field", OrderedSet["IdType"]](
+            OrderedSet
+        )
+        self._transaction.protected = StackMap["Field", OrderedSet["IdType"]]()
+        self._transaction.tocompute = defaultdict["Field", OrderedSet["IdType"]](
+            OrderedSet
+        )
```
hbrunn: I suggest using something like

```python
def copy_empty_defaultdict(_dict):
    return type(_dict)(_dict.default_factory)
```

for all defaultdicts here. This way we don't run into subtle errors if the default factory changes in core, or type hints actually do something and change

hbrunn (changes requested):

auditlog/models/rule.py
StefanRijnhart: Thanks for the suggestion! I've taken a more naive approach using copy + clear: https://github.com/OCA/server-tools/compare/8b2af873bdee80d05eb3d1b10e68cd1dedf65644..8976baea64f9d1528da855809e2e6a3e4b19363f. It seems to work for the StackMap instance as well. Do you agree with this implementation?

StefanRijnhart: (rebased to fix codecov's project coverage relative to the last build of the target branch, which had progressed in the mean time).

hbrunn: hmm, the shallow copy looks innocent enough. do we have any measurements how expensive this is?

hbrunn:
```shell
$ python3 -m timeit -s 'import copy; import collections; test=collections.defaultdict[object, object](lambda x: None, ((object(), object()) for i in range(100)))' 'test2=copy.copy(test); test2.clear()'
200000 loops, best of 5: 1.07 usec per loop

$ python3 -m timeit -s 'import copy; import collections; test=collections.defaultdict[object, object](lambda x: None, ((object(), object()) for i in range(100)))' 'test2=collections.defaultdict(test.default_factory)'
2000000 loops, best of 5: 176 nsec per loop
```

seems negligible if you agree that's a good way to estimate costs

hbrunn (approved):

auditlog/models/rule.py
StefanRijnhart: Thanks for checking!

AEstLo (approved):
LGTM

PR #3487, opened 2026-01-05
[16.0][IMP] module_auto_update: auto_upgrade command

fkantelberg:
As discussed in #2604. This will prevent problems when new fields of res.partner or res.users are added in existing problem when initializing the Odoo shell.

```bash
./odoo-bin --addons-path=ocb/addons,ocb/odoo/addons,custom-addons auto_update -d odoo
```

### DIFF
--- a/module_auto_update/cli/auto_update.py
+++ b/module_auto_update/cli/auto_update.py
@@ -1,0 +1,36 @@
+# © 2025 initOS GmbH
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import odoo
+from odoo import api
+from odoo.cli.shell import Shell
+from odoo.tools import config
+
+
+class ModuleAutoUpgrade(Shell):
+    """Run auto upgrade"""
+
+    name = "auto_update"
+
+    def run(self, args):
+        self.init(args)
+        registry = odoo.registry(config["db_name"])
+        with registry.cursor() as cr:
+            uid = odoo.SUPERUSER_ID
+            # Disabling prefetching of fields here because new fields for
+            # res.partner and res.users already exist in the python structures
+            # but aren't available in the database yet and would cause exception here
+            # This is different from invoking a python shell
+            ctx = (
+                odoo.api.Environment(cr, uid, {})["res.users"]
+                .with_context(prefetch_fields=False)
+                .context_get()
+            )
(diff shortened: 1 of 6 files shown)


module_auto_update/cli/auto_update.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo.cli.shell import Shell
+from odoo.tools import config
+
+
+class ModuleAutoUpgrade(Shell):
+    """Run auto upgrade"""
+
+    name = "auto_update"
+
+    def run(self, args):
```
StefanRijnhart, line 15: Can you add a docstring explaining why prefetch_fields is essential here and what problem it solves? Also add a link to the unmerged Odoo PR.

module_auto_update/cli/auto_update.py
```diff
@@ -0,0 +1,32 @@
...
+        registry = odoo.registry(config["db_name"])
+        with registry.cursor() as cr:
+            uid = odoo.SUPERUSER_ID
+            ctx = (
+                odoo.api.Environment(cr, uid, {})["res.users"]
+                .with_context(prefetch_fields=False)
+                .context_get()
+            )
+
+            minimal_env = api.Environment(cr, uid, ctx)
```
StefanRijnhart: Is the env essentially minimal here as well or is that a copy/paste effect in Holger's suggested code? If not, maybe just call it `env`.

StefanRijnhart (commented):
Thanks for implementing @hbrunn's suggestion! Does this now supersede https://github.com/OCA/server-tools/pull/2604, or do they solve different issues?

module_auto_update/cli/auto_update.py
fkantelberg: I added it a little bit below but a more detailed comment.

module_auto_update/cli/auto_update.py
fkantelberg: Kinda minimal yeah. But doesn't matter and I just renamed it :)

fkantelberg:
For me it would supersede the PR #2604 because it solves the same problem and an additional case if you use the command.

StefanRijnhart (approved):
Thanks!

hbrunn (approved):

PR #3493, opened 2026-01-15
[19.0][IMP] upgrade_analysis: avoid "empty" analysis files

MiquelRForgeFlow:
Before this patch, sometimes analysis files were created for the modules doesnt' contain any "change". Let's avoid that.

Besides other minor fixes, this patch improves the coverage file: No more "No DB layout changes" message on modules that contain the `noupdate_changes.xml` file.

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -26,7 +26,12 @@ except ImportError:
 from .. import compare
 
 _logger = logging.getLogger(__name__)
-_IGNORE_MODULES = ["openupgrade_records", "upgrade_analysis"]
+_IGNORE_MODULES = [
+    "openupgrade_records",
+    "upgrade_analysis",
+    "openupgrade_framework",
+    "openupgrade_scripts",
+]
 
 
 class UpgradeAnalysis(models.Model):
@@ -181,7 +186,7 @@ class UpgradeAnalysis(models.Model):
             remote_model_records, local_model_records
         )
 
-        affected_modules = sorted(
+        affected_modules = list(
             {
                 record["module"]
                 for record in remote_records
@@ -194,7 +199,7 @@ class UpgradeAnalysis(models.Model):
         )
         if "base" in affected_modules:
             try:
-                pass
+                from odoo.addons.openupgrade_scripts import apriori  # noqa: F401
             except ImportError:
                 _logger.error(
                     "You are using upgrade_analysis on core modules without "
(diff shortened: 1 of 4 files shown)


MiquelRForgeFlow:
cc: @hbrunn

pedrobaeza (approved):
Code review

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -65,14 +70,13 @@ def _compute_upgrade_path(self):
...
-        if not upgrade_path:
+        res = config.get("upgrade_path", False)
+        if not res:
             module_path = get_module_path("openupgrade_scripts", display_warning=False)
             if module_path:
                 res = os.path.join(module_path, "scripts")
-        else:
-            res = upgrade_path[0]
+        if isinstance(res, list):
+            res = res[0]
```
hbrunn: what are you fixing here?

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -192,34 +196,29 @@ def analyze(self):
                 + local_model_records
             }
         )
-        if "base" in affected_modules:
-            try:
-                pass
```
hbrunn, line 197: this used to be https://github.com/OCA/server-tools/commit/4a35a88d567f44dfd3e6b5f704f83d8476cc2318, possibly better restore that?

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -266,10 +265,11 @@ def analyze(self):
             )
 
         try:
-            self.generate_noupdate_changes()
+            noupdate_modules = self.generate_noupdate_changes()
         except Exception as e:
             _logger.exception(f"Error generating noupdate changes: {e}")
             general_log += "ERROR: error when generating noupdate changes: {e}\n"
+        no_changes_modules = list(set(no_changes_modules) - set(noupdate_modules))
```
hbrunn, line 276: this breaks when `generate_noupdate_changes` raises, so I suggest to initialize `noupdate_modules` to an empty list before

hbrunn (changes requested):
do we have a branch with the output of this changes for comparison?

BTW please merge https://github.com/OCA/OpenUpgrade/pull/5389 so that we can conveniently generate a baseline

upgrade_analysis/models/upgrade_analysis.py
MiquelRForgeFlow: `res` cannot be a list. Can you guarantee that `os.path.join(module_path, "scripts")` will never be a list?

upgrade_analysis/models/upgrade_analysis.py
MiquelRForgeFlow: Oh, nice find.

MiquelRForgeFlow:
> do we have a branch with the output of this changes for comparison?

No..., but nothing should break. We should expect less modules folders (I mean, less "empty" analysis files) and less "No DB layout changes" in the coverage file.

hbrunn (approved):

PR #3495, opened 2026-01-17
[18][FIX] dbfilter_from_header: db name matching has to be strict

cvinh:
Use case : 
We have 2 databases dbname and dbname-test
In case we setup the dbfilter to dbname, we have 2 databases that match.
So we are donig a strict match now (re.fullmatch since python 3.4)


thomaspaulb (approved):

hbrunn (changes requested):
this is supposed to do exactly what https://github.com/OCA/OCB/blob/18.0/odoo/http.py#L380 does but for headers, so I disagree with this change.

Your case should be handled with appending a $ sign, just as you'd do it if the dbfilter was in the config file. As this seems to be unclear, I suggest to change your PR to an amendment of the readme clarifying this.

PR #3497, opened 2026-01-20
[16.0][IMP] session_db: support binary fields

JonathanOsAlc:
## Summary

This PR fixes a error traceback occurring in the `session_db` module when `bytes` objects are present in the Odoo session (`request.session`). 

When custom modules place binary data in the session (e.g., for temporary image previews before saving a record), the `pg_session_store.py` script crashes during the `json.dumps(dict(session))` call because the standard `json` cannot serialize `bytes` objects natively.

## Traceback Resolved

```python
  File "/home/odoo/instance/odoo/odoo/http.py", line 2072, in __call__
    response = request._serve_db()
  File "/home/odoo/instance/odoo/odoo/http.py", line 1664, in _serve_db
    exc.error_response = self.registry['ir.http']._handle_error(exc)
  File "/home/odoo/instance/odoo/addons/http_routing/models/ir_http.py", line 660, in _handle_error
    cls._post_dispatch(response)
  File "/home/odoo/instance/odoo/addons/utm/models/ir_http.py", line 26, in _post_dispatch
    super()._post_dispatch(response)
  File "/home/odoo/instance/odoo/odoo/addons/base/models/ir_http.py", line 161, in _post_dispatch
    request.dispatcher.post_dispatch(response)
  File "/home/odoo/instance/odoo/odoo/http.py", line 1757, in post_dispatch
    self.request._save_session()
  File "/home/odoo/instance/odoo/odoo/http.py", line 1579, in _save_session
    root.session_store.save(sess)
  File "/home/odoo/instance/extra_addons/server-tools/session_db/pg_session_store.py", line 34, in wrapper
    return func(*args, **kwargs)
  File "/home/odoo/instance/extra_addons/server-tools/session_db/pg_session_store.py", line 49, in wrapper
    return func(self, *args, **kwargs)
  File "/home/odoo/instance/extra_addons/server-tools/session_db/pg_session_store.py", line 111, in save
    payload = json.dumps(dict(session))
  File "/usr/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

## Technical Implementation

- Pre-serialization Hook (session_to_str): Intercepts the session dictionary before JSON serialization. It recursively scans for bytes objects and safely encodes them as base64 strings prefixed with `base64::`.
- Post-deserialization Hook (str_to_session): Intercepts the JSON payload upon retrieval, decoding any string containing this specific prefix back into a standard bytes object.

## Evidence

- Before the fix

Video: https://drive.google.com/file/d/1M3T3rBrKN4jcU2t-zuxDcViyrPwlnn20/view?usp=sharing

- After the fix

Video: https://drive.google.com/file/d/1V6eztsmkGEGkQaJoD725aVlrvE06XLuO/view

### DIFF
--- a/session_db/pg_session_store.py
+++ b/session_db/pg_session_store.py
@@ -2,6 +2,8 @@
 # @author Nicolas Seinlet
 # Copyright (c) ACSONE SA 2022
 # @author Stéphane Bidoul
+import base64
+import binascii
 import json
 import logging
 import os
@@ -66,6 +68,7 @@ class PGSessionStore(sessions.SessionStore):
         self._cr = None
         self._open_connection()
         self._setup_db()
+        self.prefix_binary = "base64::"
 
     def __del__(self):
         self._close_connection()
@@ -108,7 +111,8 @@ class PGSessionStore(sessions.SessionStore):
     @with_lock
     @with_cursor
     def save(self, session):
-        payload = json.dumps(dict(session))
+        json_session = self.session_to_str(dict(session))
+        payload = json.dumps(json_session)
         self._cr.execute(
             """
                 INSERT INTO http_sessions(sid, write_date, payload)
@@ -131,6 +135,7 @@ class PGSessionStore(sessions.SessionStore):
         self._cr.execute("SELECT payload FROM http_sessions WHERE sid=%s", (sid,))
         try:
             data = json.loads(self._cr.fetchone()[0])
+            data = self.str_to_session(data)
         except Exception:
             return self.new()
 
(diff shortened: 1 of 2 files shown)


sbidoul:
Hello and thanks for contributing!

To help me understand your PR, can you elaborate the problem you are addressing? In particular do they also occur with the standard Odoo session or only with session_db? Do they occur with standard Odoo modules or custom code?

Can you also elaborate what the problem with debug=1, as I personally never noticed any problem in debug mode with session_db.

JonathanOsAlc:
@sbidoul 

Hello, and thank you for taking the time to review this PR!

I have made modifications to the code and updated the PR description to be much more detailed. After reviewing the again the issue, I realized that the logic handling int/float conversions and the debug=1 parameter was unnecessary. This PR is now focused on fixing the JSON serialization crash for binary (bytes) data.

To answer your specific questions:

> Does this occur with the standard Odoo session or only with session_db?

This crash specifically occurs when using `session_db`. Standard Odoo handles session storage differently and doesn't face this traceback.

> Does it occur with standard Odoo modules or custom code?

The specific trigger comes from a custom module. We have a "preview" feature that temporarily stores form data (which includes image binaries) in the user's session (request.session) to avoid creating unnecessary records in the database before the user confirms. While triggered by custom code, putting binary data in the session is a valid approach, and installing `session_db` unexpectedly breaks this capability.

> What was the problem with debug=1?

You are absolutely right—there is no inherent problem with debug=1 in session_db. The issue I mentioned in previous commits was actually caused by my own attempt to parse ints/floats (which accidentally converted the "debug": "1" string into an integer). Since I have dropped that int/float logic completely, the debug mode works perfectly without any extra changes.

I have updated the main PR description with the traceback and the technical implementation of the base64 fix. I will also be attaching evidence shortly showing the error and how this patch resolves it.

Let me know if the updated code and scope look good to you!

sbidoul:
> This crash specifically occurs when using session_db. Standard Odoo handles session storage differently and doesn't face this traceback.

Standard Odoo indeed uses `pickle` to serialize sessions while session_db serializes to json.

To be honest my preference would be to handle this (uncommon?) case in the application code. Would it work for you to do the base 64 conversion of the binary field in your module?

If there is a strong reason to do it in session_db, we'll probably want to do this using the stdlib JSONEncoder/JSONDecoder mechanism. But even then that will probably have a non negligible performance impact with the need to check on all string fields if they would need conversion to binary when decoding. 

So yeah, my instinct says this should be done at the application level and not in session_db.

Longer term we can consider using `pickle` in session_db too for better compatibility.

PR #3503, opened 2026-01-25
[FIX][18.0] fetchmail_attach_from_folder tests: mock expunge()

thomaspaulb:
@NL66278 So on 16.0 test runs I saw:

```
2026-01-25 10:24:19,760 890 ERROR odoo odoo.addons.fetchmail_attach_from_folder.models.fetchmail_server_folder: Failed to fetch mail 123 from server Test Fetchmail Server 
  File "/__w/server-tools/server-tools/fetchmail_attach_from_folder/models/fetchmail_server_folder.py", line 188, in retrieve_imap_folder
    self.apply_matching(connection, message_uid)
  File "/__w/server-tools/server-tools/fetchmail_attach_from_folder/models/fetchmail_server_folder.py", line 252, in apply_matching
    self._archive_msg(connection, message_uid)
  File "/__w/server-tools/server-tools/fetchmail_attach_from_folder/models/fetchmail_server_folder.py", line 304, in _archive_msg
    connection.expunge()
AttributeError: 'MockConnection' object has no attribute 'expunge'
```

I fixed that, does it make sense on 18.0 too? It doesn't seem to block merges.

### DIFF
--- a/fetchmail_attach_from_folder/models/fetchmail_server.py
+++ b/fetchmail_attach_from_folder/models/fetchmail_server.py
@@ -4,6 +4,7 @@ import logging
 import re
 
 from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


NL66278:
@thomaspaulb It does make sense to add it to 18.0, but the reason tests do not fail on 18.0 is that some tests are included in 16.0 that where added after the module was already migrated to 17.0 and 18.0. These test where to proof problems where solved in archiving, that also have not been solved in 18.0 So some stuff that is in 16.0 needs to be forward ported to 18.0, to make archiving work properly, and to prevent other problems when archiving active. This because the expunge(), that is part of archiving, resets the msgid's in a mailbox. So we need to use message_id's that are persistent over the life of a mailbox. I will create a forward port PR, or rather just extend this one.

hbrunn (approved):

thomaspaulb:
@NL66278 Once you are OK with this you can approve / merge this yourself!

PR #3505, opened 2026-01-28
[19.0][IMP] base_view_inheritance_extension: wraptext

hbrunn:
this finally lets us deal with text in upstream views that's not accessible to standard inheritance. When you have xml like

```xml
       <some>
           <node>
               plain text 1
               <other_node />
               plain text2
           </node>
       </some>
```

you're pretty much out of luck when you want to change either of the texts, you have to replace the whole node.

This PR allows to say

```xml
      <wraptext expr="//some/node" position="text" element="span" />
      <wraptext expr="//some/node/other_node" position="tail" element="div" />
```

in inherited views, which transforms the above to

```xml
       <some>
           <node>
               <span>plain text 1</span>
               <other_node />
               <div>plain text2</div>
           </node>
       </some>
```

that you can further modify with the inheritance machinery

### DIFF
--- a/base_view_inheritance_extension/README.rst
+++ b/base_view_inheritance_extension/README.rst
@@ -135,7 +135,7 @@ Authors
 Contributors
 ------------
 
--  Holger Brunn <hbrunn@therp.nl>
+-  Holger Brunn <mail@hunki-enterprises.com>
 -  Ronald Portier <rportier@therp.nl>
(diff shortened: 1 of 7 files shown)


StefanRijnhart (approved):
Nice one!

NL66278 (approved):
👍 LGTM and tests are comprehensive.

PR #3511, opened 2026-02-01
[16.0][IMP] Update fetchmail_server_folder

NL66278:
- Update fields description for clarity
- Support fetch_last_day_only  to fetch only 24 hours of emails
- Process emails in reverse order
- Use BODY.PEEK to avoid marking email as seen
- Support email field 'reply_to' to matching
- Improve logging
- Relayout view to group related fields
- Add `seen` selection field

Add `seen` selection field


fetchmail_attach_from_folder/tests/test_match_algorithms.py
```diff
@@ -1,41 +1,53 @@
-# Copyright - 2015-2018 Therp BV <https://acme.com>.
+# Copyright - 2015-2026 Therp BV <https://acme.com>.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
ntsirintanis, line 2: Let's fix these copyright lines (date, acme) here and someplaces else

ntsirintanis (approved):
Read and executed tests, tested locally, only a minor comment

fetchmail_attach_from_folder/tests/test_match_algorithms.py
NL66278: Done!

PR #3515, opened 2026-02-04
[15.0][IMP] onchange_helper: Support one2many fields in onchanges

pedrobaeza:
There are one2many fields that are computed, like for example `sale.order.line~agent_ids` in module `sale_commission`. When another module like `sale_product_set` uses this module to populate the order lines, you get this error:

```
odoo.sql_db: bad query: INSERT INTO "sale_order_line_agent" ("id", "create_date", "create_uid", "object_id", "write_date", "write_uid") VALUES (nextval('sale_order_line_agent_id_seq'), ...) RETURNING id
odoo-1  | ERROR: null value in column "agent_id" of relation "sale_order_line_agent" violates not-null constraint
```

And this is because the commands that are passed consists on just `(0, 0, {})`.

This commit fixes this acting in 2 places:

- Populating correctly the onchange_spec to get those fields.
- Manipulating afterwards o2m fields the same way it's done with the first level to clean many2one values from `(id, display_name)` to just the id.

@Tecnativa TT60729

### DIFF
--- a/onchange_helper/models/base.py
+++ b/onchange_helper/models/base.py
@@ -15,7 +15,16 @@ class Base(models.AbstractModel):
         for fieldname, value in vals.items():
             if fieldname not in record:
                 column = self._fields[fieldname]
-                if value and column.type == "many2one":
+                if column.type == "one2many":
+                    submodel = self.env[column.comodel_name]
+                    for vals in value:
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):
Tested in real case

PR #3489, opened 2026-01-05
[19.0][MIG] sentry: Migration to 19.0

JohanEndare:
Migration of the sentry module from 18.0 to 19.0.

This PR migrates the sentry module for Odoo 19.0 compatibility.

**Changes:**
- Updated module version to 19.0
- Tested and working on Odoo.sh production and staging environments
- Module successfully integrates with Odoo 19.0 error tracking

**Testing:**
- Tested on Odoo.sh
- Sentry integration confirmed working
- Error reporting and tracking functional

cc: @OCA/server-tools-maintainers

PR #3517, opened 2026-02-09
[19.0][MIG] sentry

ntsirintanis:
Preserving commit history for https://github.com/OCA/server-tools/pull/3489


thomaspaulb:
@JohanEndare We continued here, your review is welcome

thomaspaulb:
@ntsirintanis It looks like your last commit are only pre-commit fixes to Johan's commit, so in this case I would just squash them into one. Perhaps also reword his commit message to exclude the draft/untested comments, we're going to assume that once this gets merged it's tested and out of draft, so no need to refer to that.

sentry/hooks.py
```diff
@@ -0,0 +1,154 @@
...
+    if exclude_loggers:
+        for item in exclude_loggers:
+            ignore_logger(item)
+
+    # The server app is already registered so patch it here
+    if server:
+        server.app = SentryWsgiMiddleware(server.app)
+
+    # Patch the wsgi server in case of further registration
+    odoo.http.Application = SentryWsgiMiddleware(odoo.http.Application)
```
mpurnell1:
```suggestion
    if server:
        server.app = SentryWsgiMiddleware(server.app)
    else:
        # post_load runs before the server is created, so wrap root directly.
        # The server will pick up the wrapped root when it starts.
        odoo.http.root = SentryWsgiMiddleware(odoo.http.root)
```
In Odoo 19 (and as far back as 16), `Application` is a [class definition](https://github.com/odoo/odoo/blob/19.0/odoo/http.py#L2644) while `root` is the [actual WSGI app instance](https://github.com/odoo/odoo/blob/19.0/odoo/http.py#L2845) (`root = Application()`). Throughout the Odoo 19 codebase, `odoo.http.root` is the standard way to reference the WSGI app — `odoo.http.Application` is never referenced outside its own definition:
- Server startup — all three server types use root, not Application:
  - [GeventServer(odoo.http.root)](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1550)
  - [PreforkServer(odoo.http.root)](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1555)
  - [ThreadedServer(odoo.http.root)](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1579)
- Test framework — [odoo.http.root.session_store](https://github.com/odoo/odoo/blob/19.0/odoo/tests/common.py#L2314) for session management
- HTTP routing tests — [app=odoo.http.root](https://github.com/odoo/odoo/blob/19.0/addons/http_routing/tests/common.py#L37)

`SentryWsgiMiddleware` expects a WSGI callable (an app instance), not a class constructor. Wrapping the class replaces it with a middleware instance, so `odoo.http.Application` is no longer a class after this line runs. It happens to not break because nothing re-instantiates `Application` afterward, but it's semantically wrong.

NOTE: since post_load runs before server creation, the `if server:` branch is effectively never hit — but it's kept as a safety fallback. The else branch wraps the actual instance that the server will reference when it [initializes](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1550-L1579).

mpurnell1 (commented):
Thank you for migrating this module to Odoo 19! I noticed one small issue while porting this code into our database and thought I would bring it to your attention.

sentry/hooks.py
ntsirintanis: Great catch! Now, with the latest commit, we make sure to wrap odoo.http.root, and the Application wrapping is only kept as a fallback for compatibility

mpurnell1 (approved):

sentry/hooks.py
Kiplangatdan: @ntsirintanis With the latest commit it actually breaks on my local with this error:

`AttributeError: 'SentryWsgiMiddleware' object has no attribute 'session_store'`

Reverting back to the original works:

```
if server:
    server.app = SentryWsgiMiddleware(server.app)

# Patch the wsgi server in case of further registration
odoo.http.Application = SentryWsgiMiddleware(odoo.http.Application)

```

Can you check if it actually works on your end?

Kiplangatdan (approved):
@ntsirintanis Some complains about the unknown option `sentry-xx` in the odoo.conf file. 
Nothing to do with the module but just Odoo 19.0 being strict with conf file.

Image

JohanEndare:
> @ntsirintanis Some complains about the unknown option `sentry-xx` in the odoo.conf file. Nothing to do with the module but just Odoo 19.0 being strict with conf file.
> 
> Image

So it's going to be in the conf file instead of env variables now? Or was it always so?

Kiplangatdan:
> > @ntsirintanis Some complains about the unknown option `sentry-xx` in the odoo.conf file. Nothing to do with the module but just Odoo 19.0 being strict with conf file.
> > Image
> 
> So it's going to be in the conf file instead of env variables now? Or was it always so?

@JohanEndare The latest commit fixes this issue. It's now using a specific `sentry` section instead of adding them under the `options` section. 

The readme is updated. 

Please test and review.

thomaspaulb:
@JohanEndare If you are happy I can merge

thomaspaulb (approved):

thomaspaulb:
@dnplkndll Did your fix still need to be included here? If so, maybe you can submit a PR after it is merged?

dnplkndll:
> @dnplkndll Did your fix still need to be included here? If so, maybe you can submit a PR after it is merged?

https://github.com/OCA/server-tools/pull/3627 was an add on the client side sentry features for front end error and performance monitoring. hoping to get some interest. I had used the feature to have the errors trigger saving the actual click path on the screens. what needs better review now is how the backend and frontend traces connect.

PR #3523, opened 2026-02-18
[18.0][ADD] database_autovacuum_tuning

santostelmo:
Database Autovacuum Tuning helps administrators keep PostgreSQL healthy by
exposing recommended autovacuum settings in Odoo. It provides guidance for sizing thresholds and scale factors so large, busy databases
avoid table bloat and excessive vacuum lag. Use it to standardize autovacuum
configuration across environments and speed up maintenance operations without
manual tuning.


database_autovacuum_tuning/data/config_parameter.xml
```diff
@@ -0,0 +1,23 @@
...
+    <record
+        forcecreate="True"
+        id="config_autovacuum_vacuum_analyze_max_threshold"
+        model="ir.config_parameter"
+    >
+        <field
+            name="key"
+        >database_autovacuum_tuning.autovacuum_vacuum_analyze_max_threshold</field>
+        <field name="value">50000</field>
+    </record>
```
SilvioC2C: Adding these values through `.xml` files may become an issue if they get deleted (so their XMLIDs are deleted too) and then recreated manually (but the XMLIDs are not recreated): the next time Odoo updates the module, it will try to recreate these params, but since their keys may already exist, a `psycopg2.errors.UniqueViolation` is triggered.

I'd be better to use a post-init/post-load script that first checks if these param keys already exist, and - if not - creates them with the default value.

database_autovacuum_tuning/views/menu.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model="ir.ui.menu" id="menu_model_autovacuum_tuning">
+        <field name="name">Database Autovacuum Tuning</field>
+        <field name="sequence" eval="11" />
+        <field name="action" ref="action_model_autovacuum_tuning" />
+        <field name="parent_id" ref="base.next_id_9" />
+    </record>
+</odoo>
```
SilvioC2C: I think this can be moved into `views/database_autovacuum_tuning_views.xml`

database_autovacuum_tuning/data/config_parameter.xml
santostelmo: I tested that and there's no issue when the record is deleted.
But I missed to add `noupdate=True`

database_autovacuum_tuning/views/menu.xml
santostelmo: Moved to `views/database_autovacuum_tuning_views.xml`

santostelmo:
CI failling on auditlog module but don't see any relation with this PR:
Test:  `auditlog.tests.test_auditlog`
`AssertionError: Found unexpected attributes on res.groups: export_data unlink write auditlog_ruled_read auditlog_ruled_create auditlog_ruled_write read auditlog_ruled_unlink auditlog_ruled_export_data create`

wouitmil (approved):
Great!

cyrilmanuel (approved):

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+    analyze_threshold = fields.Integer()
+
+    @api.model
+    def _tune(self):
+        vacuum_threshold = int(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "database_autovacuum_tuning.autovacuum_vacuum_max_threshold",
+                default="100000",
```
gurneyalex:
```suggestion
                default="0",
```

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+            or 100000
```
gurneyalex:
```suggestion
            or 0
```

gurneyalex: When the parameter is unset, we want 0

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        if vacuum_threshold <= 0:
+            return
+        analyze_threshold = int(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "database_autovacuum_tuning.autovacuum_vacuum_analyze_max_threshold",
+                default="50000",
```
gurneyalex:
```suggestion
                default="0",
```

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+    _name = "database.autovacuum.tuning"
+    _description = "Database Autovacuum Tuning"
+
+    name = fields.Char(required=True, help="Table name")
+    vacuum_threshold = fields.Integer()
+    analyze_threshold = fields.Integer()
+
+    @api.model
+    def _tune(self):
+        vacuum_threshold = int(
```
gurneyalex: Handle `ValueError` (for an empty string, something that does not parse as an integer)

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+        if vacuum_threshold <= 0:
+            return
+        analyze_threshold = int(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "database_autovacuum_tuning.autovacuum_vacuum_analyze_max_threshold",
+                default="50000",
+            )
+            or 50000
```
gurneyalex:
```suggestion
            or 0
```

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+            .sudo()
+            .get_param(
+                "database_autovacuum_tuning.autovacuum_vacuum_max_threshold",
+                default="100000",
+            )
+            or 100000
+        )
+        if vacuum_threshold <= 0:
+            return
+        analyze_threshold = int(
```
gurneyalex: Handle ValueError

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        query = """
+            SELECT
+                t.schemaname,
+                t.tablename,
+                st.dead_tuple_count
+            FROM pg_tables AS t
+            JOIN LATERAL pgstattuple(
+                format('%%I.%%I', t.schemaname, t.tablename)::regclass
+            ) AS st ON true
```
gurneyalex: Maybe I was wrong in the spec, I'm sorry about that. This would require the pgstattuple extension to be installed, and it is not going to be available everywhere. 

Better to use a join on `pg_stat_all_tables` which has a `n_dead_tup` column that we can use.

database_autovacuum_tuning/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,12 @@
...
+documentation for sizing thresholds and scale factors so large, busy databases
+avoid table bloat and excessive vacuum lag. Use it to standardize autovacuum
+configuration across environments and speed up maintenance operations without
+manual tuning.
+
+This module is mostly useful for PostgreSQL <= 17. PostgreSQL 18.0 introduces
+the `autovacuum_vacuum_max_threshold` parameter, which already provides the
+capability this module targets.
+
+The `pgstattuple` extension must be installed on the database.
```
gurneyalex, line 12:
```suggestion
```

with the update in the query suggested above applied

database_autovacuum_tuning/readme/USAGE.md
```diff
@@ -0,0 +1,21 @@
+1. Install the module on the database you want to tune.
+2. Ensure the `pgstattuple` extension is installed in PostgreSQL.
```
gurneyalex:
```suggestion
```

database_autovacuum_tuning/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_database_autovacuum_tuning_system,access_database_autovacuum_tuning_system,model_database_autovacuum_tuning,base.group_system,1,1,1,1
```
gurneyalex:
```suggestion
access_database_autovacuum_tuning_system,access_database_autovacuum_tuning_system,model_database_autovacuum_tuning,base.group_system,1,0,1,1
```

database_autovacuum_tuning/tests/test_database_autovacuum_tunning.py
```diff
@@ -0,0 +1,60 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo.tests import TransactionCase
+
+
+class TestAutovacuumTunning(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env.cr.execute("CREATE EXTENSION IF NOT EXISTS pgstattuple")
```
gurneyalex:
```suggestion
```

database_autovacuum_tuning/tests/test_database_autovacuum_tunning.py
```diff
@@ -0,0 +1,60 @@
...
+                "WHERE id = ANY(%s)",
+                (partners.ids,),
+            )
+        self.env.flush_all()
+        self.env.cr.execute("ANALYZE res_partner")
+
+    def _get_dead_tuples(self, schemaname="public", tablename="res_partner"):
+        table = f"{schemaname}.{tablename}"
+        self.env.cr.execute(
+            "SELECT dead_tuple_count FROM pgstattuple(%s::regclass)",
```
gurneyalex: use `pg_stat_all_tables`

database_autovacuum_tuning/tests/test_database_autovacuum_tunning.py
```diff
@@ -0,0 +1,60 @@
...
+
+from odoo.tests import TransactionCase
+
+
+class TestAutovacuumTunning(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env.cr.execute("CREATE EXTENSION IF NOT EXISTS pgstattuple")
+        with cls.env.registry.cursor() as cr:
```
gurneyalex: I believe cursors are disabled in tests to never commit. If you want to test here I would recommend mocking.

gurneyalex (changes requested):

database_autovacuum_tuning/security/ir.model.access.csv
santostelmo: That seems fair to me. The system user does not need to write to the registers.

database_autovacuum_tuning/models/database_autovacuum_tuning.py
santostelmo: When the parameter is unset, we want 0 => Changed that

database_autovacuum_tuning/tests/test_database_autovacuum_tunning.py
santostelmo: Not using pgstattuple extension anymore

database_autovacuum_tuning/tests/test_database_autovacuum_tunning.py
santostelmo: mocking the method now

gurneyalex (approved):

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,85 @@
...
+class DatabaseAutovacuumTuning(models.Model):
+    _name = "database.autovacuum.tuning"
+    _description = "Database Autovacuum Tuning"
+
+    name = fields.Char(required=True, help="Table name")
+    vacuum_threshold = fields.Integer()
+    analyze_threshold = fields.Integer()
+
+    @api.model
+    def _tune(self):
```
simahawk:
```suggestion
    def _db_autovacuum_tune(self):
```

database_autovacuum_tuning/data/ir_cron.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="cron_database_autovacuum_tuning" model="ir.cron">
```
simahawk, line 3: can't we simply use the `@autovacuum` decorator?

database_autovacuum_tuning/models/database_autovacuum_tuning.py
```diff
@@ -0,0 +1,85 @@
...
+                ALTER TABLE {schemaname}.{tablename} SET (
+                    autovacuum_vacuum_scale_factor = 0,
+                    autovacuum_vacuum_threshold = %s,
+                    autovacuum_analyze_scale_factor = 0,
+                    autovacuum_analyze_threshold = %s
+                )
+                """,
+                (vacuum_threshold, analyze_threshold),
+            )
+            self.sudo().create(
```
simahawk, line 34: Is this record created just for reference?

santostelmo: Yes, just for reference

database_autovacuum_tuning/data/ir_cron.xml
santostelmo: We could but we probably will be mixing 2 goals: 
1. @autovacuum removing obsolete or temporary data
2. cron_database_autovacuum_tuning tune vacuum config parameter on tables

database_autovacuum_tuning/models/database_autovacuum_tuning.py
santostelmo: Changed

database_autovacuum_tuning/data/ir_cron.xml
gurneyalex: I agree with @santostelmo this would be confusing the purpose of both are totally different.

database_autovacuum_tuning/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "depends": [
+        "base_setup",
+    ],
+    "data": [
+        "data/config_parameter.xml",
+        "data/ir_cron.xml",
+        "security/ir.model.access.csv",
+        "views/res_config_settings_views.xml",
+        "views/database_autovacuum_tuning_views.xml",
+    ],
```
gurneyalex, line 20: @santostelmo can you set the development status to Alpha ?
```suggestion
    ],
    "development_status": "Alpha",
```

santostelmo: Done

vvrossem (approved):

hbrunn:
an existing approach is https://github.com/OCA/server-tools/tree/16.0/postgres_vacuum that makes vacuum an explicit action. maybe combine both?

gurneyalex:
> an existing approach is https://github.com/OCA/server-tools/tree/16.0/postgres_vacuum that makes vacuum an explicit action. maybe combine both?

Both can coexist, but the goal here is to rely on autovacuum, which is what I understand your module wants to avoid ("autovacuum is not recommended for busy databases, it can be a better choice to schedule this process manually"). In our experience, autovacuum works fine, especially if you manage to get rid of super long "idle in transaction" and "active" connections to the database, which essentially prevent autovacuum from running. The tuning that this module performs is another key factor to get a performant autovacuum as the default settings will not give good results for large tables with many updates. 

It may well be that our use cases do not match yours and the patterns of transaction that lead to the postgres_vacuum addon are indeed not compatible with autovacuum. Both addons serve different purposes, it should be possible to install them together without any negative consequences, so I would be tempted to let them live in parallel.

PR #3525, opened 2026-02-19
[18.0][ADD] base_sequence_template

LauraCForgeFlow:


LoisRForgeFlow (changes requested):
Looks great! Some suggestions to improve:

- add a smart button to see the generated sequences from the template:
Image

- add button in form view:
Image

- add button in tree view (we can save the additional click on "actions")
Image

Also, the name of the module could be just **base_sequence_template**, I don't see a reason to add the multi_company in the name as it is not extending base_multi_company or anything related. You are adding sequence templates that works with one or multiple companies like many standard modules. WDYT?

LauraCForgeFlow:
Thanks for the suggestions @LoisRForgeFlow! I completely agree, I have implemented the changes you proposed.

LoisRForgeFlow (approved):
Awesome! LGTM

PR #3526, opened 2026-02-19
[18.0][REF] auditlog et al: adapt to new Odoo test patch checker

### LINKED ISSUE #3439: Auditlog: Patching Causes Failures in Other Unit Tests
## Describe the Bug
When `auditlog` creates a rule during tests (e.g., for `res.partner`), it monkey-patches methods like `res.partner.write` and keeps a snapshot of the then-current implementation (usually the base method). If another addon (e.g., `partner_contact_address_default`) overrides `write()`, that override is skipped in tests and the call goes to the older base implementation. This causes other addons’ test cases to fail. In normal UI/server runs, everything works as expected; the problem appears only in test runs.

# Impacted Versions
- 16.0

Haven't checked in other versions

## To Reproduce
Run tests together with `auditlog` and other addons (for example, [partner_contact_address_default](https://github.com/OCA/partner-contact/tree/16.0/partner_contact_address_default)).

A similar issue is also occurring with the base_user_role module, as described in https://github.com/OCA/server-backend/issues/321#issue-2632230540.

@qrtl QT5614

StefanRijnhart:
Requires

- [x] https://github.com/OCA/odoo-test-helper/pull/39
- [x] https://github.com/Tecnativa/odoo-test-helper/pull/1

See https://github.com/odoo/odoo/pull/247151

Fixes in `auditlog`:

* Register with the patch checker to prevent exhaustive logging

```
odoo odoo.tests.common: /__w/server-tools/server-tools/auditlog/models/rule.py:232:_patch_method setting res.groups.unlink to <function AuditlogRule._make_unlink.<locals>.unlink_full at 0x7fa51ea90f70>
```

* Align model patching with how it's done in base_automation to fully remove patched methods. Fixes #3439
* Dismantle own patch checker.
* Ensure no patching takes place in setUpClass.

Fixes in other modules:

* Make sure patches or registry updates happen in setUp, not setUpClass and are properly cleaned.

### DIFF
--- a/attachment_queue/tests/test_attachment_queue.py
+++ b/attachment_queue/tests/test_attachment_queue.py
@@ -32,24 +32,19 @@ class TestAttachmentBaseQueue(BaseCommon):
             ).create(vals)
         return self.env["attachment.queue"].create(vals)
 
-    @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
-        cls.loader = FakeModelLoader(cls.env, cls.__module__)
-        cls.loader.backup_registry()
-        from .test_models import AttachmentQueue
-
-        cls.loader.update_registry((AttachmentQueue,))
-
-    @classmethod
-    def tearDownClass(cls):
-        super().tearDownClass()
-        cls.loader.restore_registry()
-        return super().tearDownClass()
-
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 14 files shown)


pedrobaeza (approved):
Let's try to greenify everything involved in the new "pocalypse" (how do we call it? FakeModelLoaderpocalypse?

/ocabot merge patch

PR #3529, opened 2026-02-24
Implement profiler

wouitmil:
The idea of this profiler is to allow profile dynamically some functions.

You can add the function in the settings > profiler menu. Then check the output of the profiling in the same menu.

You can then either check the static flamegraph if you have flameprof lib installed or download the pstats file and explore it with snakeviz downloaded_.pstats 

I'm also intending to be able to download all .pstats files combined fo a function. So you can explore the average with snakeviz


profiler/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "Profiler",
+    "version": "18.0.1.0.0",
+    "category": "Tools",
+    "summary": "yappi profiler decorator with database storage",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/server-tools",
+    "depends": ["base", "edi_exchange_template_oca"],
```
gurneyalex: why the dependency on `edi_exchange_template_oca `?

profiler/models/edi_exchange_template_output.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+from ..tools import profiled
+
+
+class EDIExchangeOutputTemplate(models.Model):
+    _inherit = "edi.exchange.template.output"
```
gurneyalex: looks strange

profiler/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Profiler",
```
gurneyalex: The name of the module is probably too generic, maybe `profiler_yappi`?

profiler/README.rst
```diff
@@ -0,0 +1,96 @@
+========
+Profiler
+========
+
+.. !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+   !! This file is generated by oca-gen-addon-readme !!
+   !! changes will be overwritten.                   !!
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
gurneyalex, line 9: I think you forgot to git add the `readme/` directory

profiler/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        "python": ["yappi", "cairosvg", "flameprof"],
+    },
+    "data": [
+        "security/ir.model.access.csv",
+        "views/profiler_function_views.xml",
+        "views/profiler_result_views.xml",
+        "views/profiler_report_views.xml",
+    ],
+    "installable": True,
+}
```
gurneyalex, line 24: code maturity to set to Alpha?

profiler/models/profiler_result.py
```diff
@@ -0,0 +1,276 @@
...
+
+            # Create temp files
+            fd, pstats_path = tempfile.mkstemp(suffix=".pstats")
+            svg_fd, svg_path = tempfile.mkstemp(suffix=".svg")
+
+            try:
+                # Write pstats data to temp file
+                with os.fdopen(fd, "wb") as f:
+                    f.write(stats_data)
+                os.close(svg_fd)
```
gurneyalex: why close the file descriptor?

profiler/models/profiler_result.py
```diff
@@ -0,0 +1,276 @@
...
+                            <ul>
+                                <li><code>gprof2dot -f pstats profile.pstats
+                                | dot -Tpng -o output.png</code></li>
+                                <li><code>snakeviz profile.pstats</code></li>
+                            </ul>
+                        </div>
+                        """
+
+                    with open(svg_path, "w") as svg_file:
+                        svg_file.write(result.stdout)
```
gurneyalex: now you have a race condition on that file...

profiler/models/profiler_result.py
```diff
@@ -0,0 +1,276 @@
...
+                except Exception as e:
+                    _logger.error(f"Unexpected error in flameprof: {e}")
+                    return f"<p>Unexpected error generating flamegraph: {str(e)}</p>"
+
+                # Convert SVG to PNG using cairosvg to avoid SVG rendering issues
+                try:
+                    import cairosvg
+
+                    png_fd, png_path = tempfile.mkstemp(suffix=".png")
+                    os.close(png_fd)
```
gurneyalex: race condition pending...

profiler/models/profiler_result.py
```diff
@@ -0,0 +1,276 @@
...
+
+                    with open(svg_path) as svg_file:
+                        svg_content = svg_file.read()
```
gurneyalex: the content of the file is already in memory (variable `result` from above)

profiler/models/profiler_result.py
```diff
@@ -0,0 +1,276 @@
...
+                                <li><code>snakeviz profile.pstats</code></li>
+                            </ul>
+                        </div>
+                        """
+
+                    with open(svg_path, "w") as svg_file:
+                        svg_file.write(result.stdout)
+                except FileNotFoundError:
+                    return """
+                    <div style="padding: 20px; background-color: #f8d7da;
```
gurneyalex: these things are really ugly. 

not translatable
hardcoded CSS 

This should be in the qweb template

profiler/__manifest__.py
wouitmil: You're right, this should be removed. It was from my first tests.

profiler/models/edi_exchange_template_output.py
wouitmil: Same as previous one ;-)
It's not needed anymore since I implemented dynamic profiling

profiler/models/profiler_result.py
wouitmil: Sure, that's why the PR is still in draft. I'll clean that next weekend. I needed to get it working first

PR #3535, opened 2026-03-02
[19.0][FIX] upgrade_analysis: don't generate invalid noupdate xml for discarded field values

hbrunn:
in v18, there's currencies set on payment's `payment_method_bank_transfer` record: https://github.com/OCA/OCB/blob/18.0/addons/payment/data/payment_method_data.xml#L593

in v19, no currencies are set: https://github.com/OCA/OCB/blob/19.0/addons/payment/data/payment_method_data.xml#L565

which generates

```xml
<field name="supported_currency_ids"/>
```

in `noupdate_changes.xml`. But this amounts to writing `''` on a many2many field, which fails as that's an invalid value for this field type.

So here I propose to write whatever the field's NULL value is instead, ensuring we generically empty that field in such a case.

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -5,6 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 # flake8: noqa: C901
 
+import ast
 import logging
 import os
 from copy import deepcopy
@@ -351,7 +352,8 @@ class UpgradeAnalysis(models.Model):
                     if record_remote_dict[key].tag == "field":
                         field_name = remote_record.xpath(key)[0].attrib.get("name")
                         if (
-                            field_name
+                            local_record.attrib["model"] not in self.env
+                            or field_name
                             not in self.env[local_record.attrib["model"]]._fields.keys()
                         ):
                             continue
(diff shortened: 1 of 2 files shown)


upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -360,6 +362,15 @@ def _get_xml_diff(
                     for attr in ["eval", "ref"]:
                         if attr in attribs:
                             del attribs[attr]
+                    if record_remote_dict[key].tag == "field" and set(attribs) == {
+                        "name"
```
MiquelRForgeFlow, line 366: why `set(attribs) == {"name"}` comparison? Could you explain a bit? I mean, why not put `"name" in attribs` instead or something else?

MiquelRForgeFlow (commented):
Code is not wrong probably, but could you explain/clarify my doubts?

upgrade_analysis/models/upgrade_analysis.py
hbrunn: this is to test that we're actually in a field node, and that there aren't other attributes set such as `search`. or to put it differently: This ensures we're only intervening when the resulting node would be `<field name="fieldname" />`

MiquelRForgeFlow (approved):

PR #3565, opened 2026-03-19
[18.0][FIX] auditlog: move partner/user dependent tests to post_install

tendil:
This PR updates several auditlog test classes to run as post_install tests instead of at_install.

### Why this change is needed

A subset of auditlog tests creates `res.partner / res.users` records in `setUpClass()` and `setUp()`
In our investigation, these tests were executed during the module loading phase, when the res.partner registry was not yet in its final state.

In that phase:
- `res.partner._fields` did not contain `autopost_bills`
- `default_get(["autopost_bills"])` returned `{}`

As a result, creating a partner failed with:

> psycopg2.errors.NotNullViolation: null value in column "autopost_bills"
> of relation "res_partner" violates not-null constraint

The same tests executed successfully as post-tests, when the registry was fully loaded and:
- `res.partner._fields` did contain `autopost_bills`
- `default_get(["autopost_bills"])` returned `{"autopost_bills": "ask"}`

### Root cause

The problem was not related to auditlog method patching/reverting logic.
The issue was that these tests were running too early (at_install) while depending on a fully initialized res.partner model.

### Fix

The affected test classes are now explicitly marked as:
```python
@tagged("-at_install", "post_install")
```

This makes their execution phase consistent with their actual dependencies.

### Why this is correct

These tests are not validating installation-time behavior.
They validate audit logging behavior on regular ORM operations and require the final registry state for `res.partner` / `res.users`

Running them as `post_install` is therefore the correct lifecycle for these scenarios.

### DIFF
--- a/auditlog/tests/test_auditlog.py
+++ b/auditlog/tests/test_auditlog.py
@@ -2,7 +2,7 @@
 # © 2018 Pieter Paulussen <pieter_paulussen@me.com>
 # © 2021 Stefan Rijnhart <stefan@opener.amsterdam>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
+from odoo.tests import tagged
 from odoo.tools import mute_logger
 
 from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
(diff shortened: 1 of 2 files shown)


ivs-cetmix (approved):
Looks reasonable!

ivs-cetmix:
Hey @hbrunn would appreciate your review!

StefanRijnhart (approved):
Indeed this is a known error when running the tests on a database that has required fields defined in other modules. Thanks!

PR #3566, opened 2026-03-20
[16.0][IMP] database_cleanup: purge orphaned attachments

Aldeigja:
Add wizard to list ir.attachment records whose store_fname points to a missing file on disk when storage is "file", and allow purging those rows.


database_cleanup/tests/test_purge_attachments.py
```diff
@@ -0,0 +1,156 @@
...
+            def patched_unlink(self):
+                if protected_id in self.ids:
+                    # Dynamic message avoids translation lint on test-only UserError.
+                    raise UserError(str(protected_id))
+                return original_unlink(self)
+
+            wizard = env["cleanup.purge.wizard.attachment"].create(
+                {
+                    "purge_line_ids": [
+                        (
```
StefanRijnhart: Can you use the more modern `fields.Command.create`?

database_cleanup/tests/test_purge_attachments.py
```diff
@@ -0,0 +1,156 @@
...
+            os.unlink(IrAttachment._full_path(orphan_protected.store_fname))
+            os.unlink(IrAttachment._full_path(orphan_other.store_fname))
+
+            protected_id = orphan_protected.id
+            other_id = orphan_other.id
+
+            IrModel = env.registry["ir.attachment"]
+            original_unlink = IrModel.unlink
+
+            def patched_unlink(self):
```
StefanRijnhart, line 112: Why does unlink need to be patched? Isn't the `exists` check enough for the validation of the test?

StefanRijnhart (commented):
Thanks! Judging from the name of the PR and the patterns from this module, I would have expected signalling attachments where the model no longer exists or the res_id is not present as a record. This makes sense as well, but to make it explicit what the issue is with each attachment (and to prepare for extension with other possible error conditions), would you mind adding a `reason` field to the wizard line model as is done in cleanup.purge.line, with for now a fixed reason like 'File missing in filestore'?

database_cleanup/tests/test_purge_attachments.py
Aldeigja: Thanks for pointing that out. Fixed.

database_cleanup/tests/test_purge_attachments.py
Aldeigja: The patch is necessary because this test validates the error-resilience path in the purge().  Regardless of whether unlink() raises a UserError for one attachment, the remaining attachments should still be purged. Without the patch, both attachments would be successfully deleted, and we would miss the exception branch entirely.

database_cleanup/models/purge_attachments.py
```diff
@@ -0,0 +1,110 @@
...
+
+class CleanupPurgeLineAttachment(models.TransientModel):
+    _inherit = "cleanup.purge.line"
+    _name = "cleanup.purge.line.attachment"
+    _description = "Cleanup Purge Line Attachment"
+
+    attachment_id = fields.Many2one("ir.attachment")
+    reason = fields.Selection(
+        [
+            (REASON_MISSING_FILE, ("File missing in filestore")),
```
StefanRijnhart:
```suggestion
            (REASON_MISSING_FILE, "File missing in filestore"),
```

database_cleanup/models/purge_attachments.py
```diff
@@ -0,0 +1,110 @@
...
+            attach = line.attachment_id
+            try:
+                attach.unlink()
+                purged_line_ids.append(line.id)
+            except (UserError, ValidationError, AccessError) as exc:
+                self.logger.warning(
+                    "Attachment #%s cannot be deleted: %s",
+                    attach.id,
+                    str(exc),
+                )
```
StefanRijnhart, line 55: Given that you built this resilience, you will want to visualize this as well. Can you store the error message in a new column on the line model, and add it to the list view with optional="hide".

database_cleanup/models/purge_attachments.py
```diff
@@ -0,0 +1,110 @@
...
+                ("type", "=", "binary"),
+            ]
+        )
+        for attach in attachments:
+            full_path = self.env["ir.attachment"]._full_path(attach.store_fname)
+            if not os.path.isfile(full_path):
+                res.append(
+                    (
+                        0,
+                        0,
```
StefanRijnhart: Another one for `Command`

StefanRijnhart (commented):
Thanks for the updates!  Please squash commits.

database_cleanup/models/purge_attachments.py
Aldeigja: Thanks for noticing that.

database_cleanup/models/purge_attachments.py
Aldeigja: Thanks for noticing.

database_cleanup/models/purge_attachments.py
Aldeigja: Thank you, I've added the field and updated the view as well.

StefanRijnhart (commented):
Thanks, the error message is stored correctly.

Image

Usability wise, I think as a user I would want to see some more information about the attachments, the res_model, res_id and res_name, with or without optional visibility. Is this not something that you want for your own purposes?

Aldeigja:
@StefanRijnhart 

> Usability wise, I think as a user I would want to see some more information about the attachments, the res_model, res_id and res_name, with or without optional visibility. 
>Is this not something that you want for your own purposes?

Our objective in creating this PR was to clean approximately 15,000 orphaned attachments simultaneously. We didn’t have to inspect every single attachment.
However, if users need to check details, they can do so by clicking on the line, opening the line pop-up, and then accessing the attachment form.
image

StefanRijnhart (approved):
OK, leaving my last comment as a nice to have. Nice work!

database_cleanup/tests/test_purge_attachments.py
```diff
@@ -0,0 +1,153 @@
...
+from unittest.mock import patch
+
+from odoo.exceptions import UserError
+from odoo.fields import Command
+from odoo.tests.common import tagged
+
+from .common import Common, environment
+
+
+# Use post_install to get all models loaded more info: odoo/odoo#13458
```
tendil: I don't think this comment belongs here

database_cleanup/tests/test_purge_attachments.py
```diff
@@ -0,0 +1,153 @@
...
+@tagged("post_install", "-at_install")
+class TestCleanupPurgeLineAttachment(Common):
+    def setUp(self):
+        """Create two ir.attachment records; delete backing file of one (orphan).
+
+        :var orphan: ir.attachment with backing file removed via os.unlink
+        :var valid: ir.attachment with file intact
+        :return: None
```
tendil: remove redundant lines such as ":return: None" to avoid unnecessary noise

database_cleanup/models/purge_attachments.py
```diff
@@ -0,0 +1,110 @@
...
+        if not purged_line_ids:
+            return True
+        return (
+            self.env["cleanup.purge.line.attachment"]
+            .browse(purged_line_ids)
+            .write({"purged": True})
+        )
+
+
+class CleanupPurgeWizardAttachment(models.TransientModel):
```
tendil, line 67: I think we should move this class to a separate file in the wizard folder

StefanRijnhart: Combining the x.wizard and x.line.wizard models in a single file follows the layout of all other wizards in this module and is allowed as per https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#15file-naming. Shall we keep it this way?

tendil: okay, then let's stick with the current version;
thanks

database_cleanup/tests/test_purge_attachments.py
```diff
@@ -0,0 +1,132 @@
...
+from odoo.exceptions import UserError
+from odoo.fields import Command
+from odoo.tests.common import tagged
+
+from .common import Common, environment
+
+
+@tagged("post_install", "-at_install")
+class TestCleanupPurgeLineAttachment(Common):
+    def setUp(self):
```
tendil: please rewrite it to use `setUpClass`, since that works faster

Aldeigja:
It appears that the CI failure here is unrelated to this change. Check [this](https://github.com/OCA/server-tools/pull/3682#issuecomment-5238802683) comment.

tendil (approved):
Code review: LGTM

database_cleanup/models/purge_attachments.py
```diff
@@ -85,24 +86,36 @@ def find(self):
...
-                    fields.Command.create(
-                        {
-                            "attachment_id": attach.id,
-                            "name": attach.store_fname or attach.name or str(attach.id),
-                            "reason": REASON_MISSING_FILE,
-                        }
+        last_id = 0
+        ir_attachment = self.env["ir.attachment"]
+        while True:
+            rows = ir_attachment.search_read(
```
tendil, line 92: since the purpose of this change is to avoid MemoryError on very large databases, please evict the processed batch from the cache before fetching the next one using `invalidate_recordset(["store_fname", "name"])`

`search_read()` limits the size of a single batch here but does not limit the total memory usage of the orm cache

database_cleanup/tests/test_purge_attachments.py
```diff
@@ -50,6 +50,30 @@ def test_find_orphaned_attachments(self):
             self.assertIn(self.orphan_attach_id, line_attachment_ids)
             self.assertNotIn(self.valid_attach_id, line_attachment_ids)
 
+    def test_find_does_not_prefetch_unrequested_attachment_fields(self):
```
tendil, line 53: please add a regression test that processes multiple batches (the batch size can be patched to a small value) and also verifies that the processed batch is evicted from the cache

otherwise the implementation could start retaining all processed attachments in the Environment cache again while this test would still pass

tendil (approved):
Code Review: LGTM

StefanRijnhart:
@Aldeigja Thanks for the continued effort following up the comments! Can you squash the commits?

StefanRijnhart:
And CI was fixed in the mean time by https://github.com/OCA/server-tools/pull/3707, so ready for merge.

/ocabot merge major

pedrobaeza:
@Aldeigja are you going to fw-port this to upper versions?

ivs-cetmix:
Hi @pedrobaeza  I will announce the module next, let's see if there will be some sponsors or project we can use to forward port it.

pedrobaeza:
Well, this is only an improvement over an existing module, not a new one, and the fw-ports should be part of the initial funding project, as if not, you won't provide a consistent experience across Odoo versions.

ivs-cetmix:
@pedrobaeza I totally agree with you, however we don't have a budget for forward porting it at the moment. Will definitely do this once we have it.

pedrobaeza:
Well, I think this strategy, and more for trivial things like a fw-port of an `[IMP]`, will make you to have to lose more time in the future seeing what is already and what not in upper versions, but it's your decision.

PR #3579, opened 2026-03-27
[18.0][ADD] fetchmail_s3: receive email from S3-compatible buckets

dnplkndll:
Adds an **"S3 Bucket"** incoming-mail server type: polls an S3-compatible bucket for raw `.eml` objects, feeds each through `mail.thread.message_process`, then archives (to a prefix) or deletes it.

**Use case:** AWS SES inbound rules drop messages into S3; this picks them up on the fetchmail cron. Works with any S3-compatible store (AWS S3, MinIO, Hetzner Object Storage, DO Spaces).

**Design:** follows the `google_gmail` / `microsoft_outlook` extension pattern — `selection_add` a `server_type='s3'`, override `connect()` / `button_confirm_login()` / `fetch_mail()`, S3 ops via `boto3` (paginated list → get → copy/delete).

Depends on `boto3`.


dnplkndll:
CI test failures are caused by `oca_list_external_dependencies` not finding `fetchmail` in the addons path, despite it being a core Odoo CE module. The `fetchmail` module is in `odoo/addons/fetchmail/` and should be available.

The same error pattern exists for `test_auditlog` (`setup/test_auditlog` fails with 'Unsupported odoo version 18.0') — this is a pre-existing repo issue.

Pre-commit passes cleanly. The module installs and tests correctly against Odoo 18 CE in our own CI.

PR #3381, opened 2025-09-22
[19.0][MIG] base_name_search_improved: Migration to 19.0

bizzappdev:

PR #3580, opened 2026-03-29
[19.0][MIG] base_name_search_improved: Migration to 19.0

gdgellatly:
## Summary

Continuation of #3381 by @bizzappdev with an additional fix applied:

- **Guard against non-string `name` argument**: In Odoo v19, `name_search` can be called with a list as the `name` parameter in certain m2o widget scenarios (e.g. clicking through from a partner on a sale order with Smart Name Search enabled on fields like `complete_name`, `normalized email`, and `phone`). This causes `AttributeError: 'list' object has no attribute 'replace'`. Added `isinstance(name, str)` guards in both the `Base.name_search` smart search path and the `_register_hook` wrapper to gracefully fall back to standard search when name is not a string.

This addresses the review feedback in:
- https://github.com/OCA/server-tools/pull/3381#discussion_r2669559247 (named arguments - already fixed in bizzappdev's latest)
- https://github.com/OCA/server-tools/pull/3381#discussion_r2902331271 (non-string name guard)


base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,284 @@
...
+                additional_results = [
+                    (res_id, res_name)
+                    for res_id, res_name in smart_results
+                    if res_id not in seen_ids
+                ]
+                return original_results + additional_results
+
+            wrapper._is_smart_patched = True
+            wrapper.origin = original_name_search
+            wrapper._api = getattr(original_name_search, "_api", None)
```
dannyadair: I believe this needs to say
```
-            wrapper._api = getattr(original_name_search, "_api", None)
+            wrapper._api_model = getattr(original_name_search, "_api_model", False)
```

Odoo renamed https://github.com/odoo/odoo/commit/40da85aab905fd4accf8c01887e62b07537ccd2f#diff-8e4705d5ab335a8dfc2c70eeedc502a92695989f97c4f19184b49634c2c4ad55L190

dannyadair (changes requested):
One smart search issue - with the suggested adjustment everything seems to work

dannyadair (approved):

SanamiK (approved):
Functional review. LGTM!

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,284 @@
...
+        help="Use Smart Search for 'name_search', this will affect when "
+        "searching from other records (for eg. from m2o fields",
+    )
+    name_search_ids = fields.Many2many("ir.model.fields", string="Smart Search Fields")
+    name_search_domain = fields.Char(string="Smart Search Domain")
+    smart_search_warning = fields.Html(compute="_compute_smart_search_warning")
+
+    @api.depends("name_search_ids")
+    def _compute_smart_search_warning(self):
+        msgs = []
```
AungKoKoLin1997, line 178: Shouldn't it be inside the loop?

base_name_search_improved/models/ir_model.py
```diff
@@ -0,0 +1,284 @@
...
+            if msgs:
+                rec.smart_search_warning = (
+                    f"<p>In case of performance issues we recommend to review "
+                    f"these suggestions: <ul>"
+                    f"{''.join(f'<li>{x}</li>' for x in msgs)}</ul></p>"
+                )
+            else:
+                rec.smart_search_warning = False
+
+    @api.constrains("name_search_ids", "name_search_domain", "add_smart_search")
```
AungKoKoLin1997, line 200: Shouldn't we also add `use_smart_name_search` to the decorator? If not, toggling it in the UI doesn't clear the `_get_use_smart_name_search` ormcache and has no effect until restart.

AungKoKoLin1997 (commented):
The README seems out of sync with the latest design — it confused me while testing (e.g. CONFIGURE says smart search is enabled automatically on all models, but it's now opt-in). Could you update it to reflect the current behavior while you're at it?

base_name_search_improved/models/ir_model.py
gdgellatly: You're probably right, but this is preexisting and not part of migration. Plenty using prior version in prod, probably better as bugfix PR forward ported, or cherry picked t here.

base_name_search_improved/models/ir_model.py
gdgellatly: Again, not part of migration, so probably better dealt with a bug fix PR cherry picked to here.

gdgellatly:
> The README seems out of sync with the latest design — it confused me while testing (e.g. CONFIGURE says smart search is enabled automatically on all models, but it's now opt-in). Could you update it to reflect the current behavior while you're at it?

We didn't change the behaviour, or the documentation. You wouldn't do that in a migration without making that clear. Literally just a straight migration. Probably better as cherry picked forward port.

paulius-sladkevicius (commented):
LGTM

paulius-sladkevicius (approved):

PR #3581, opened 2026-03-30
[19.0][FIX] database_cleanup: Added missing security rules for two models

wpichler:


pedrobaeza (commented):
Now that we are here, shouldn't these permissions be adjusted for the admin group only?

wpichler:
i've changed it to base.group_system - force push...

StefanRijnhart (commented):
Please check again, I don't see base.group_system in the access rules. Also, please mention the module name in the commit message.

wpichler:
@StefanRijnhart You are right - sorry - made a commit without the changed file... Now it is correct

StefanRijnhart (approved):
Great, thanks!

pedrobaeza (commented):
Can you please split the changes into 2: one for the switch from `group_user` to `group_system`, and the other adding the missing ACLs?

wpichler:
Ready to Merge...

pedrobaeza (approved):
/ocabot merge patch

PR #3583, opened 2026-03-31
[14.0][FIX] model_read_only: address same label warning

HekkiMelody:
Fixes the following warning

```
2026-03-31 12:12:35,210 1 WARNING devel odoo.addons.base.models.ir_model: Two fields (model_name, model_id) of model.readonly.restriction() have the same label: Model. 
```


HekkiMelody:
Pre-commit failure is addressed in https://github.com/OCA/server-tools/pull/3584

PR #3588, opened 2026-04-01
[18.0][ADD] mass_module_updater: wizard to update multiple modules at once

polqubiq:
## Description

  Adds new module `mass_module_updater`.
                                                                                                                                                                                                           This module provides a transient wizard (`mass.module.updater.wizard`) that
  allows updating or installing multiple Odoo modules at once by entering their
  technical names as a comma-separated list.

  Two actions are available:

  - **Update Modules**: upgrades all listed modules (raises `UserError` if any
    module is not installed).                                                                                                                                                                              - **Install Modules**: installs all listed modules (raises `UserError` if any
    module is already installed).

  In both cases, a `UserError` is raised if any module name is not found in                                                                                                                                `ir.module.module`.

  ## Motivation

  Updating or installing modules one by one via the standard Apps interface is
  slow when multiple modules need to be refreshed after a deployment. This wizard
  reduces that to a single operation.

  ## Checklist

  - [x] `README.rst` present and up to date                                                                                                                                                                - [x] `__manifest__.py` follows OCA conventions (license, author, website)
  - [x] Security rules defined in `ir.model.access.csv

### DIFF
--- a/mass_module_updater/README.rst
+++ b/mass_module_updater/README.rst
@@ -1,0 +1,37 @@
+===================
+Mass Module Updater
+===================
+
+This module provides a wizard to update multiple Odoo modules at once by their technical names.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+1. Go to Apps > Mass Module Updater.
+2. Enter the technical names of the modules you want to update, separated by commas.
+3. Click on "Update Modules".
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/server-tools/issues>`_.
+
+Credits
+=======
+
+Authors
+-------
+
+* Pol Reig
+
+Maintainers
+-----------
+
+This module is maintained by QubiQ.
--- a/mass_module_updater/__init__.py
+++ b/mass_module_updater/__init__.py
@@ -1,0 +1,4 @@
+# Copyright 2026 Pol Reig
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import wizards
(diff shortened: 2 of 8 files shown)


polqubiq:
@pedrobaeza I've added the `mass_module_updater` module, could you take a look when you get a chance? Thanks!

pedrobaeza:
Great addition. Right now, I'm not interested in this one, but being here in OCA, other interested people can review it. Formally, CI is green, so the initial pre-requisites are OK.

legalsylvain:
Hi. I don't get the usage.
I mean, I understand the usage to mass install, but not mass update. Could you elaborate when mass update is interesting ?

thanks !

polqubiq:
@pedrobaeza @legalsylvain 
1. It is possible to install / update modules with this module.
2. As developers, we often modify several modules at once, and when deploying, they need to be updated. It is easier to enter the module names and update them all at once than to search for them one by one manually and update them. I think this would help all of us.

legalsylvain:
> As developers, we often modify several modules at once, and when deploying, they need to be updated. It is easier to enter the module names and update them all at once than to search for them one by one manually and update them. I think this would help all of us.

yes. on that topic, I see the way you propose quite unsecure : 

- If you have to select manually modules to update, you'll not be exhautive and can forgive to update ones. For _that need_, some tools like click-odoo-update are totally adapted to your usage.

- update after having launched the server can fail, specially if you modify some base model like partner / company, and you could not have access to the UI to update the module. Update should be done when launching the server (after having pulled new base code)

- And it requires extra work.

I think that a module named "mass_module_installer" could be great indeed. (part of the code you propose). In that context, a more generic way to implement could be to have the possibility to cron the installation. i mean : 
- you select the list of module you want to install (as you propose actually)
- you put a date for the installation (optional)
- when confirming, if the date is put, it will postpone the installation. This feature can be great to make installation during the night. if the date is not set, immediate installation is done.

WDYT ?

polqubiq:
Okey, I'll close the PR then. Thanks for the feedback!

PR #3590, opened 2026-04-01
[18.0][FIX] base_exception: Rollback transaction if we detect exceptions

grindtildeath:
This aims to solve issues when modules not depending on whatever implementation of base_exception, override the same function that triggers the detection of exceptions.

Before this commit, any changes done in such overrides could end up being committed to the database if the MRO did execute such function before the function implementing base_exception that avoids to call super in case an exception is detected. (eg sale.order. action_confirm in sale_exception)

With this commit, in case there is any newly detected exception, or a record with exception that is not ignored, or a blocking exception linked to a record, exception changes will be committed in DB while a specific Exception type will be raised to rollback any changes done in the ongoing transaction. Such an exception will be handled in the UI to refresh the exception_ids field so that the user knows why the action was not completed.

Replaces and fixes:
 - https://github.com/OCA/server-tools/pull/3476
 - https://github.com/OCA/sale-workflow/issues/2667
 - https://github.com/OCA/sale-workflow/pull/3672
 - https://github.com/OCA/sale-workflow/pull/3767
 - https://github.com/OCA/sale-workflow/pull/3885
 - https://github.com/OCA/sale-workflow/pull/3952

Dependent PRs:
 - [ ] https://github.com/OCA/sale-workflow/pull/4255
 - [ ] https://github.com/OCA/purchase-workflow/pull/3012


base_exception/models/base_exception_method.py
```diff
@@ -84,10 +88,33 @@ def detect_exceptions(self):
...
         # table
         # and the "to add" part generates one INSERT (with unnest) per rule.
-        for rule_id, records in rules_to_remove.items():
-            records.write({"exception_ids": [(3, rule_id)]})
-        for rule_id, records in rules_to_add.items():
-            records.write({"exception_ids": [(4, rule_id)]})
+        raise_exception = False
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with Registry(self.env.cr.dbname).cursor() as new_cr:
```
ivantodorovich:
```suggestion
        with self.env.registry.cursor() as new_cr:
```

no need to initialize a new registry

base_exception/models/base_exception_method.py
```diff
@@ -84,10 +88,33 @@ def detect_exceptions(self):
...
+            new_env = Environment(new_cr, self.env.uid, self.env.context)
```
ivantodorovich, line 94:
```suggestion
            records_env = records.with_env(self.env(cr=new_cr))
```

to avoid all the `with_env` everytime you read/write, initialize a records_env variable.

also you can use self.env(cr=) like this

base_exception/models/base_exception_method.py
```diff
@@ -84,10 +88,33 @@ def detect_exceptions(self):
...
+                    rec.ignore_exception
+                    for rec in self.with_env(new_env)
+                    if rec.exception_ids
+                )
+                or any(
+                    rule.is_blocking
+                    for rule in self.with_env(new_env).mapped("exception_ids")
+                )
+            ):
+                raise_exception = True
```
ivantodorovich, line 104: Maybe extract this to a method `records_env._should_raise_exception` ?

ivantodorovich (approved):

hparfr:
ping @florian-dacosta

base_exception/models/base_exception_method.py
grindtildeath: Actually we have many sets of records:
1. self
2. values of rules_to_add
3. values of rules_to_add

For each one we need to use the new cursor, so I guess it makes sense to create the new_env this way. AFAIU `self.env(cr=new_cr)` would just do the same, but we need to reuse it many times anyway

ivantodorovich (approved):

grindtildeath:
ping @jok-adhoc @lav-adhoc @lef-adhoc @maq-adhoc @rousseldenis @florian-dacosta 

After spending hours on this issue, IMO this solves the core issue we have with non dependent module having their changes done in `action_confirm` committed after an exception was detected since the transaction was not fully rolled back.

I then changed the way the exception popup can be displayed and it seems to work smoothly with no side effects and no need of glue modules.

Could you guys please have a look, eventually test this (with their dependent PRs) and let me know what do you think? :pray:

rousseldenis (approved):
LGTM.

Great! @grindtildeath Thanks for this, this seems indeed cleaner to solve the historical issue we have with this.

florian-dacosta:
Hello @grindtildeath Thanks for this PR.
Code review looks fine. There is a pre-commit error, I had this recently and did change the name of the file by .esm.js to solve it. (Not sure it is the right solution).

But I did a test and it is not working as expected. On a demo database with only `sale_management` and `sale_exception` installed, I did add the following line in the method `action_confirm` of sale_management (before the super) : `self.write({"client_order_ref": "TEST"})`.

[sale-workflow.webm](https://github.com/user-attachments/assets/9b39ad00-5c24-47dd-b8af-b1c355520087)

As you can see in the video, after clicking on the confirm button of the sale order, I have the new exception and the client order ref field did not change, all good on this part.
But, I did not have any window with the exceptions + all button are disabled. (including the little arrow to go to partner form for instance, not only the header buttons).

Can you check this issue please ?

grindtildeath:
@florian-dacosta Thanks for your feedback.

I reworked the JS part to make it cleaner and simpler, it seems to solve your issue and works seamlessly on my side. Would be good if you can test again.

ivantodorovich (approved):
CI fails due to pre-commit

florian-dacosta (changes requested):
@grindtildeath 
It seems to work pretty well now, but there is still an issue.
I test with sale_exception, and if I run the cron to test the exceptions on draft order, manually (with button `method_direct_trigger`), I have this error 
`AttributeError: The method 'ir.cron.action_popup_exceptions' does not exist`

This is because the button on _ir.cron_ do trigger the `BaseExceptionError`.

grindtildeath:
@florian-dacosta Tanks for the feedback, must be fixed now

florian-dacosta:
Thanks for the fix @grindtildeath 
It seems to work quite well now. The last thing that bother me a little, is that an exception is raised even when not needed.
Indeed, when you confirm a sale order, you want the exception to revert everything, but when it is the cron that "test exceptions on draft orders", you don"'t need to revert anything the raise is not needed.

It does not seem to be a big problem since the exceptions are commited anyway, but 
It still gives a warning : 2026-04-14 07:44:57,926 31 WARNING db odoo.http: Exceptions detected on model: sale.order
while it is a "false alarm" since it is totally normal to set/remove exeption with the cron.

Also, I guess we may have situation with the inverse issue we try to fix here.
Imagine you override some `test_all_draft_orders` on sale order for instance, and you expect your code to change some stuff after the update of all exceptions, it won't happen because of the raise, while you would not expect a raise since the method `detect_exceptions` succeed in doing what it has been designed for...
overriding `detect_exceptions` directly could also lead to this.
We want to raise only because `detect_exceptions` is called during the action_confirm, but not necessarly for other  cases...

I am not blocking anything as for now, the gain seems better than the potential loss, but I think it may be an issue.
What do you think ?

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +87,30 @@ def detect_exceptions(self):
...
+            if rules_to_add or self_new_env._must_raise_exception_after_detection():
+                raise_exception = True
+        if raise_exception:
+            raise BaseExceptionError(f"Exceptions detected on model: {self._name}")
         return all_exception_ids
 
+    def _must_raise_exception_after_detection(self):
+        return not all(
+            rec.ignore_exception for rec in self if rec.exception_ids
+        ) or any(rule.is_blocking for rule in self.mapped("exception_ids"))
```
simahawk, line 112:
```suggestion
        ) or any(rule.is_blocking for rule in self.exception_ids)
```

base_exception/static/src/js/base_exception.esm.js
```diff
@@ -0,0 +1,40 @@
...
+patch(rpc, {
+    async _rpc(url, params = {}, settings = {}) {
+        try {
+            return await super._rpc(url, params, settings);
+        } catch (error) {
+            if (
+                error.exceptionName ===
+                "odoo.addons.base_exception.exceptions.BaseExceptionError"
+            ) {
+                if (error.data.message.split(":")[1].trim() === params.model) {
```
simahawk, line 30: If I get it right when looking at If you register the exception into any of these registries, you won't need this patch:
```
const errorHandlerRegistry = registry.category("error_handlers");
const errorDialogRegistry = registry.category("error_dialogs");
const errorNotificationRegistry = registry.category("error_notifications");
```
The second one is probably the right one.

Not a blocker but maybe we can add a TODO here 😉

simahawk (approved):
LGTM. As for version bump I would go for `major`.

simahawk:
@florian-dacosta  If I get your explanation right, it sounds like a side effect that is not fully in control.

To allow devs to still store some information before the exception is raised we could add an hook and pass to it the new env. This way you can control it in a very explicit way. 

Not a blocker. maybe we can add a TODO in the roadmap.

florian-dacosta:
> @florian-dacosta If I get your explanation right, it sounds like a side effect that is not fully in control.
> 
> To allow devs to still store some information before the exception is raised we could add an hook and pass to it the new env. This way you can control it in a very explicit way.
> 
> Not a blocker. maybe we can add a TODO in the roadmap.

I was more thinking on a context to get rid of the raise.
So, by default (no context), the error is raise and the rollback is done (required when the detect exceptions is part of the `action_confirm` for instance.
And for a method like `sale.order.test_all_draft_orders` which only goal is to check exceptions, you know the raise is not necessary, you could call it with the special context.

Well anyway, not a blocker like I said, maybe it is anticipating an issue we won't ever have. I have no usecase in mind, just thought about the possible side effects.

florian-dacosta (approved):

base_exception/static/src/js/base_exception.esm.js
grindtildeath: Done in fixup commit, but had to use `errorHandlerRegistry`

hparfr:
@grindtildeath may you please rebase with autosquash

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +89,45 @@ def detect_exceptions(self):
...
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with self.env.registry.cursor() as new_cr:
+            new_env = (
+                Environment(new_cr, self.env.uid, self.env.context)
+                if not test_mode
+                else self.env
+            )
+            for rule_id, records in rules_to_remove.items():
+                records.with_env(new_env).write({"exception_ids": [(3, rule_id)]})
```
vvrossem:
```suggestion
                records.with_env(new_env).write({"exception_ids": [Command.unlink(rule_id)]})
```

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +89,45 @@ def detect_exceptions(self):
...
+            for rule_id, records in rules_to_add.items():
+                records.with_env(new_env).write({"exception_ids": [(4, rule_id)]})
```
vvrossem:
```suggestion
                records.with_env(new_env).write({"exception_ids": [Command.link(rule_id)]})
```

base_exception/models/base_exception_method.py
```diff
@@ -4,14 +4,19 @@
 # Copyright 2020 Hibou Corp.
 # Copyright 2023 ACSONE SA/NV (http://acsone.eu)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import json
 import logging
 from collections import defaultdict
 
 from odoo import _, api, models
```
vvrossem:
```suggestion
from odoo import _, api, models, Command
```

vvrossem (approved):

base_exception/models/base_exception.py
```diff
@@ -66,17 +66,22 @@ def _compute_exceptions_summary(self):
             else:
                 rec.exceptions_summary = False
 
+    def _must_popup_exception(self):
```
divad1196, line 69: `_must_popup_exception`  -> `_should_popup_exception` ? 

`must` express a necessity to enforce, not just the intent to show it.

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +89,47 @@ def detect_exceptions(self):
...
-            records.write({"exception_ids": [(4, rule_id)]})
+        raise_exception = False
+        test_mode = config["test_enable"] and not self.env.context.get(
+            "test_base_exception"
+        )
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with self.env.registry.cursor() as new_cr:
+            new_env = (
+                Environment(new_cr, self.env.uid, self.env.context)
```
divad1196, line 100: `self.env(cr=cr)` instead ?

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +89,47 @@ def detect_exceptions(self):
...
         return all_exception_ids
 
+    def _detect_exception_get_exc_class_values(self):
+        return {
+            "src_model": self._name,
+            "target_model": self._name,
+        }
+
+    def _must_raise_exception_after_detection(self):
+        return not all(
```
divad1196: Can we decompress the expression like this 
```python
main_records = self._get_main_records().filtered("exception_ids")
return not all(
        main_records.mapped("ignore_exception")
        ) or any(rule.is_blocking for rule in records.exception_ids)
```
or
```python
main_records = self._get_main_records().filtered("exception_ids")
ignore_exception = all(
        main_records.mapped("ignore_exception")
)
blocking_exception = any(rule.is_blocking for rule in records.exception_ids)
return not ignore_exception or blocking_exception
```

base_exception/models/base_exception_method.py
```diff
@@ -84,12 +89,47 @@ def detect_exceptions(self):
...
+        test_mode = config["test_enable"] and not self.env.context.get(
+            "test_base_exception"
+        )
+        # Write exceptions in a new transaction to be committed so that we can
+        #  rollback the ongoing one while keeping the exceptions stored
+        with self.env.registry.cursor() as new_cr:
+            new_env = (
+                Environment(new_cr, self.env.uid, self.env.context)
+                if not test_mode
+                else self.env
```
divad1196, line 102: if we use `self.env`, this means we cannot test to rollback?

base_exception/tests/test_base_exception.py
```diff
@@ -146,3 +171,42 @@ def test_blocking_exception(self):
...
+                mock_base_exception_method_env(self, env=new_env),
+                self.assertRaises(BaseExceptionError),
+            ):
+                self.po.button_detect_and_confirm()
+            # Entering assertRaises will create a first savepoint using self.env.cr.
+            # Then, when write is triggered through new_cr in
+            # base.exception.method.detect_exceptions, a second savepoint will be
+            # created using new_cr, and an odoo.sql_db.Savepoint object will be stored
+            # on new_cr._savepoint for this second savepoint.
+            # As the with block of assertRaises is exited a rollback to the first
```
divad1196: Explanation a bit hard to follow IMO. I think that bullet points would help, what do you think?

> 1. Enter `with assertRaises` => Create SavePoint1
> 2. Write using `new_cr` => Create SavePoint2 (after SavePoint1)
> 3. Exit `with assertRaises` => Rollback to SavePoint1 => invalidate Savepoint2
>
> => We need to discard SavePoint2 to avoid `savepoint does not exist` error.
> This happens because new_cr and self.env.cr use the same psycopg2 cursor object behind the scene


What's still not clear to me is when the `write on new_cr` happens

divad1196 (approved):

base_exception/models/base_exception_method.py
grindtildeath: We need to call `Environment` here so that we can patch it in `mock_base_exception_method_env`

base_exception/models/base_exception_method.py
grindtildeath: Not sure to understand your concern here :raised_eyebrow:

base_exception/tests/test_base_exception.py
grindtildeath: > What's still not clear to me is when the write on new_cr happens

It happens in `base.exception.method.detect_exceptions` as explained :grimacing: 

I added some numbers to explicit each action to avoid rewriting everything, hope it's fine.

base_exception/models/base_exception_method.py
SilvioC2C: I think you can use `self.env(cr=new_cr)` by mocking `Environment.__new__()`. Though I think it's fine as it is.

base_exception/tests/test_base_exception.py
```diff
@@ -146,3 +171,43 @@ def test_blocking_exception(self):
...
+            # removed from new_cr._savepoint, but as both self.env.cr and new_cr
+            # use the same psycopg2 cursor object behind the scene, the second
+            # savepoint does not exist anymore in the database.
+            # This situation would actually trigger a "savepoint does not exist"
+            # psycopg2 exception when trying to release or rollback the savepoint
+            # when closing the cursor. Therefore, we can safely remove the reference
+            # to that object to avoid this error when exiting the test.
+            new_cr._savepoint = None
+            self.assertFalse(self.po.exception_ids)
+            self.assertTrue(self.po.with_env(new_env).exception_ids)
```
SilvioC2C, line 214: For the sake of maintainability, would you please write a comment explaining why we expect `self.po.exception_ids` to have different values according to the chosen `Environment`?

SilvioC2C (approved):

base_exception/models/base_exception_method.py
divad1196: `env(cr=cr)` does exactly what you are doing it: It creates a copy. 
https://github.com/odoo/odoo/blob/18.0/odoo/api.py#L650

base_exception/models/base_exception_method.py
divad1196: Sorry, I was indeed not clear.

The code here is not in a test, it's the production.
And we make an edge-case for when the function is in a test `test_mode=True`.

=> `Environment(new_cr, self.env.uid, self.env.context)` is only used in production, not in testing.

In testing, we use `self.env` => We cannot test the double cursor trick to ensure it works

base_exception/models/base_exception_method.py
SilvioC2C: @divad1196 beware: 
- if you use `odoo.api.Environemnt(...)`, it is resolved to `odoo.api.Environment.__new__()`
- if you use `self.env(...)`, it is resolved to `odoo.api.Environment.__call__()` which then calls `odoo.api.Environment.__new__()`
- `odoo.api.Environment.__new__()` will _not_ always create a new copy: if an `odoo.api.Environment` object with the same params (cursor, uid, context, sudo flag, original uid) already exists for the current transaction, such object is returned:
```
# using the shell
env: <odoo.api.Environment object at 0x7f539c402270>
odoo: <module 'odoo' from '/odoo/src/odoo/__init__.py'>
openerp: <module 'odoo' from '/odoo/src/odoo/__init__.py'>
self: res.users(1,)
Python 3.12.13 (main, May 19 2026, 23:48:44) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(Console)
>>> env_1 = env(self.env.cr, self.env.uid, self.env.context, self.env.su)
>>> env_2 = env(self.env.cr, self.env.uid, self.env.context, self.env.su)
>>> env_1 is env_2
True
```

IMHO using `self.env()` or `odoo.api.Environment()` is pretty much the same; `self.env()` just adds one more step.  If we need to mock the usage of both of them, then I think mocking `odoo.api.Environment.__new__()` should be the preferred solution.

base_exception/models/base_exception_method.py
grindtildeath: It is tested through the use of context key `test_base_exception`.

We need this to avoid messing with other modules being tested while this one is installed.

base_exception/models/base_exception_method.py
divad1196: @SilvioC2C I know how they work, but it's irrelevant:
they both ultimately end up calling `__new__`. 

The point of using `self.env(cr=cr)` is that the `__call__` will fetch the other parameters so you don't have to pass them manually like done here. That's the only difference. 
This is why I sent the link to the code.

`Environment` uses what's called a [multiton pattern](https://en.wikipedia.org/wiki/Multiton_pattern) (like a singleton based on paramters).
The code for it is [here](https://github.com/odoo/odoo/blob/18.0/odoo/api.py#L586):
```python
for env in transaction.envs:
    if (env.cr, env.uid, env.su, env.uid_origin, env.context) == (cr, uid, su, uid_origin, context):
        return env
```

 and it won't apply because we passed a new cursor object to `env(cr=cr)`

## TLDR:

There is no reason to not use `env(cr=cr)`

base_exception/models/base_exception_method.py
divad1196: @grindtildeath I don't understand how this impacts other modules. Can you explain?
The current comment only explain why we create a new environment for production, it does not explain why the same method cannot be used in testing.

rousseldenis:
@grindtildeath Could we go forward on this ?

rousseldenis:
@grindtildeath

grindtildeath:
@rousseldenis I would say so, but we need to coordinate the merge between the PRs in the different repositories

simahawk:
> @rousseldenis I would say so, but we need to coordinate the merge between the PRs in the different repositories

Not a problem, we merge this and right after the sale and purchase PRs.

Pls check pre-commit ;)

simahawk:
I'll merge on Monday as I won't have time to followup if we get problems over the weekend ;)

carlos-lopez-tecnativa:
FW-Port to V19 in https://github.com/OCA/server-tools/pull/3729/

PR #3599, opened 2026-04-15
[18.0][ADD] base_trusted_proxies

### LINKED ISSUE #3540: New module to configure the number of trusted proxies in x-forwarded-for
See https://github.com/odoo/documentation/pull/6744 for context.

Nevertheless there are still circumstances where the x-forwarded-for does not and cannot have the client IP as last item. For instance I'm having this situation right now with a customer on Azure Container Apps.

So I propose a server wide module named `base_trusted_proxies` which patches [odoo.http.ProxyFix](https://github.com/odoo/odoo/blob/e665f01e379ca13ee47d6545a9e31ca7bc86db8a/odoo/http.py#L185) to set `x_for` to a number obtained from an `ODOO_TRUSTED_PROXIES` environment variable. The default value would be 1 as it is how Odoo works out of the box.

In the roadmap for this module, we'd add a mention that a special value is reserved to mean that all proxies are trusted and the 1st value in x-forwarded-for is always considered to be the client ip. This must come with the appropriate security caveat.

qgroulard:
Allow configuration of the number of trusted proxies for accurate Client IP detection.

Fixes https://github.com/OCA/server-tools/issues/3540


base_trusted_proxies/post_load.py
```diff
@@ -0,0 +1,22 @@
...
+
+from odoo import http
+
+_logger = logging.getLogger(__name__)
+
+
+def post_load():
+    trusted_proxies_nbr = int(os.environ.get("ODOO_TRUSTED_PROXIES", "1"))
+
+    _logger.info(f"Patch ProxyFix to allow {trusted_proxies_nbr} proxies.")
```
sbidoul:
```suggestion
    _logger.info(f"Patch ProxyFix to trust {trusted_proxies_nbr} proxies in X-Forwarded-For.")
```

base_trusted_proxies/post_load.py
```diff
@@ -0,0 +1,22 @@
...
+import logging
+import os
+
+from odoo import http
+
+_logger = logging.getLogger(__name__)
+
+
+def post_load():
+    trusted_proxies_nbr = int(os.environ.get("ODOO_TRUSTED_PROXIES", "1"))
```
sbidoul: Warn that the module is loaded but not configured if the env var is not set?
Do nothing and warn if it does not parse as int or is < 1?

sbidoul (commented):
LGTM. A couple of comments.

qgroulard:
> LGTM. A couple of comments.

Thanks for the review, I have added all those checks.

PR #3604, opened 2026-04-23
[19.0][MIG] fetchmail_notify_error_to_sender: Migration to 19.0

CRogos:
@NICO-SOLUTIONS could you test?


NICO-SOLUTIONS:
@CRogos 
In current Odoo versions, message_route() no longer reliably raises a ValueError for routing failures, as unmatched routing cases are handled internally via the catchall/fallback mechanism. 
This means the exception-based hook used here is effectively not triggered anymore in most real scenarios, so the notification logic is likely inactive in practice.

If the goal is to handle emails sent to the catchall address without a matching alias, this would require a different approach than relying on ValueError, since it is no longer a reliable signal. 
Instead, one would need to hook into the routing result or extend the catchall processing flow.

Alternatively, this might already be better addressed on Odoo side: the default notification email is a bit misleading, as it suggests the catchall is only for replies, even when emails are sent to unknown aliases. Adjusting that wording could improve clarity without additional custom logic.

fetchmail_notify_error_to_sender/data/email_template_data.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!--Email template -->
+    <record id="email_template_error_notice" model="mail.template">
+        <field name="name">Fetchmail - error notice</field>
+        <field name="email_from">{{ctx.get('sender_message').get('to')}}</field>
```
njeudy, line 6: The migrated template will fail to install on Odoo 19 whenever a fetchmail.server record already exists.

Odoo 19 added mail.template._check_can_be_rendered() (called on create/write), which renders every dynamic field against a real record of the template's model. The template in data/email_template_data.xml uses chained .get():
`
<field name="email_to">{{ctx.get('sender_message').get('from')}}</field>
`
At install time sender_message is not in the context (it is only injected at fetch time by mail.thread.message_route), so ctx.get('sender_message') returns None and None.get('from') raises AttributeError, aborting the installation with:

Oops! We couldn't save your template due to an issue. […] Correct it and try again.

Note _check_can_be_rendered short-circuits when no fetchmail.server exists, which is why installing on a clean database may not surface the problem.

Suggested fix — null-safe access, runtime behaviour unchanged:

```
<field name="email_from">{{ctx.get('sender_message', {}).get('to')}}</field>
<field name="email_to">{{ctx.get('sender_message', {}).get('from')}}</field>

```
>Receiving error with: {{ctx.get('sender_message', {}).get('subject')}}</field>

and the matching t-out expressions in body_html. (ctx.get('route_exception') is a single .get() and is already safe.)

njeudy (changes requested):

fetchmail_notify_error_to_sender/data/email_template_data.xml
CRogos: @njeudy I am not actively working on this module and I think @NICO-SOLUTIONS mentioned that this module might not be needed anymore. Feel free to fork this work and create a new PR.

PR #3610, opened 2026-05-16
[18.0][FIX] fetchmail_attach_from_folder: case-insensitive email matching

### LINKED ISSUE #3402: Email matching fails when addresses contain uppercase letters
**Module**

fetchmail_attach_from_folder

**Bug Description**

Emails containing uppercase letters in the address are not linked to partner records.

**Steps to reproduce the issue**

Configure a folder to attach emails to partners based on the email field.

Save a partner with an email address in uppercase (Name.SURNAME@Domain.com) and/or another in lowercase (name.surname@domain.com).

Drop an email into the monitored folder with the same address but in a different case (e.g., Name.SURNAME@Domain.com).

Notice that no match is found, even though the address exists in Odoo.

**Expected Behavior**

The matching should be case-insensitive, so that:

NAME.SURNAME@DOMAIN.COM
name.surname@domain.com
Name.SURNAME@Domain.com
are all treated as the same address.

**Additional Context**

Odoo version: 18
IMAP/Email server: Gmail

This issue prevents emails with uppercase/mixed-case addresses from being attached to the correct partners, even when the address exists in Odoo.

fsmw:
## Bug

Closes #3402

Emails containing uppercase letters were not linked to partner records because the  match algorithm lowercased the incoming email addresses but then searched with  (case-sensitive exact match). If the partner stored the email as mixed-case (e.g. Name.SURNAME@Domain.com), the match failed.

## Fix

Use Odoo's  operator instead of . This matches Odoo's own approach in core ( line 965) and performs a case-insensitive exact match via PostgreSQL .

## Changes
- : default operator changed from  to 
- : added  covering uppercase email matching

## Compatibility
- Fully backward compatible: existing exact matches still work; now uppercase variants also match.

### DIFF
--- a/fetchmail_attach_from_folder/match_algorithm/email_exact.py
+++ b/fetchmail_attach_from_folder/match_algorithm/email_exact.py
@@ -1,11 +1,11 @@
 # Copyright - 2013-2024 Therp BV <https://therp.nl>.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo.tools.mail import email_split
+from odoo.tools.mail import email_normalize, email_split
 from odoo.tools.safe_eval import safe_eval
 
 
 class EmailExact:
-    """Search for exactly the mailadress as noted in the email"""
+    """Search for exactly the email address as noted in the email."""
 
     def _get_mailaddresses(self, folder, message_dict):
         mailaddresses = []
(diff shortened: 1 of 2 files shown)


NL66278:
@fsmw I think the bug is valid and the problem real. However replacing an equal comparison `=` with `ilike` might run into a lot of other problems. Suppose a mail is fetched for fernando@example.com, and there is a partner with email Fernando@example.com in the database. Your solution will find this, but will also find Alexando-Fernando@example.com. So maybe better to encourage people to search on email_normalized instead of on email (and reflect this in the tests and examples). What do you think?

fsmw:
Hi @NL66278, thanks for the review and the valid concern. Your suggestion has been incorporated:
-  _get_mailaddresses() now uses Odoo's email_normalize() to normalize incoming addresses (strips display names, lowercases, handles edge cases).
-  Since the search values are normalized and contain no % wildcards, =ilike functions exactly as a case-insensitive = — it won't match substrings.
-  Added test coverage for partner stored with uppercase email and for display-name headers ("Name <email@domain>").

Let me know if anything else is needed.

fetchmail_attach_from_folder/match_algorithm/email_exact.py
```diff
@@ -1,23 +1,36 @@
...
         mailaddresses = []
         fields = folder.mail_field.split(",")
         for field in fields:
             if field in message_dict:
                 mailaddresses += email_split(message_dict[field])
-        return [addr.lower() for addr in mailaddresses]
+        # Normalize using email_normalize for consistent matching.
+        # This strips display names, lowercases the address, and handles
+        # edge cases (e.g. "<user@domain.com>" or "User <user@domain.com>").
+        return [email_normalize(addr) or addr.lower() for addr in mailaddresses]
```
NL66278, line 19: Is the `or addr.lower()` not superfluous here?

NL66278: @fsmw Can you answer the question? It is the one thing I would like answered before approve and merge (maybe after removing addr.lower if indeed not needed).

PR #3612, opened 2026-05-17
[18.0][FIX] base_remote: allow RPC authentication without HTTP context

### LINKED ISSUE #3458: base_remote: API authentication broken due to missing HTTP context check
# API/RPC Authentication Fails with "Expected singleton: res.remote()"

## Module
base_remote

## Describe the bug
The `base_remote` module breaks XML-RPC and JSON-RPC authentication, preventing external API clients (OdooRPC Python library, mobile apps using direct RPC, automated scripts, etc.) from authenticating to Odoo. Authentication fails with `ValueError: Expected singleton: res.remote()` while web browser authentication continues to work normally.

## To Reproduce
**Affected versions**: 18.0.1.0.0 (likely affects earlier versions as well)

1. Install the `base_remote` module
2. Attempt to authenticate via XML-RPC or JSON-RPC from an external client:
   ```python
   import xmlrpc.client

   url = 'http://your-odoo-server:8069'
   common = xmlrpc.client.ServerProxy(f'{url}/xmlrpc/2/common')
   uid = common.authenticate('database', 'username', 'password', {})
   ```
3. Authentication fails with error: `ValueError: Expected singleton: res.remote()`

**Expected behavior**
API/RPC authentication should succeed, just as web browser authentication does. External API clients should be able to authenticate and interact with Odoo programmatically.

**Additional context**

**Root cause:**
In `models/base.py`, the `remote` property gracefully handles missing HTTP context (which occurs during RPC calls) by returning an empty recordset:

```python
@property
def remote(self):
    try:
        remote_addr = http.request.httprequest.remote_addr
    except (KeyError, AttributeError, RuntimeError):
        return self.env["res.remote"]  # Returns empty recordset for RPC calls
```

However, `models/res_users.py` unconditionally calls `ensure_one()` on this recordset:

```python
@classmethod
def _auth_check_remote(cls, credential, method):
    with cls.pool.cursor() as cr:
        remote = env["res.users"].remote
        if not config["test_enable"]:
            remote.ensure_one()  # Crashes when remote is empty recordset
```

- Web browser logins: Have HTTP context → `remote` returns 1 record → `ensure_one()` succeeds ✓
- API/RPC logins: No HTTP context → `remote` returns empty recordset → `ensure_one()` fails ❌

**Proposed fix:**
Change line 19 in `models/res_users.py`:

```python
# Current:
if not config["test_enable"]:
    remote.ensure_one()

# Fixed:
if not config["test_enable"] and remote:
    remote.ensure_one()
```

This allows the singleton check to be skipped when no HTTP context exists (RPC calls), while still enforcing device tracking for web browser connections (the module's primary purpose). This follows the same pattern already used for test mode and doesn't compromise the module's functionality.

**Environment:**
- Odoo 18.0
- Python 3.12
- Tested with OdooRPC library and direct XML-RPC calls

fsmw:
Fixes #3458

When authenticating via RPC (XML-RPC, JSON-RPC), there is no active
HTTP request context. The `Base.remote` property returns an empty
`res.remote` recordset in this case.

However, `ResUsers._auth_check_remote` unconditionally called
`ensure_one()` on that recordset, causing:

```
ValueError: Expected singleton: res.remote()
```

This blocked all API/RPC authentication while web login continued
to work normally.

**Fix:** Add `and remote` to the guard before `ensure_one()`,
skipping the singleton check when the remote recordset is empty.
This matches the existing pattern used for `config['test_enable']`
and does not affect web authentication where an HTTP context exists.

**Tests:**
- Added `test_auth_check_remote_no_http_context` which patches
  `Base.remote` to return an empty recordset and verifies that
  `_auth_check_remote` tolerates it without crashing.
- Existing tests for web login remain untouched.

Backward compatible: existing behaviour for browser sessions is
unchanged.

### DIFF
--- a/base_remote/models/res_users.py
+++ b/base_remote/models/res_users.py
@@ -16,7 +16,7 @@ class ResUsers(models.Model):
         with cls.pool.cursor() as cr:
             env = api.Environment(cr, SUPERUSER_ID, {})
             remote = env["res.users"].remote
-            if not config["test_enable"]:
+            if not config["test_enable"] and remote:
                 remote.ensure_one()
         result = method()
         if not result:
--- a/base_remote/tests/test_remote.py
+++ b/base_remote/tests/test_remote.py
@@ -76,7 +76,7 @@ class TestRemote(HttpCase):
         data1["password"] = "Failure!"
         self.assertFalse(
             self.xmlrpc_common.authenticate(
-                self.env.cr.dbname, data1["login"], data1["password"]
+                self.env.cr.dbname, data1["login"], data1["password"], {}
             )
         )
         with self.cursor() as cr:
--- a/base_remote/tests/test_remote.py
+++ b/base_remote/tests/test_remote.py
@@ -76,9 +76,30 @@ class TestRemote(HttpCase):
         data1["password"] = "Failure!"
         self.assertFalse(
             self.xmlrpc_common.authenticate(
-                self.env.cr.dbname, data1["login"], data1["password"], {}
+                self.env.cr.dbname, data1["login"], data1["password"]
             )
         )
         with self.cursor() as cr:
             env = self.env(cr)
             self.assertTrue(env["res.remote"].search([("ip", "=", self.remote_addr)]))
+
+    def test_auth_check_remote_no_http_context(self, *args):
+        """_auth_check_remote must not crash when remote is empty.
+
+        When there is no HTTP context (e.g. some RPC paths), ``Base.remote``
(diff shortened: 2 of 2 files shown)


fsmw:
Closing this PR. The fix causes 3 test failures in the existing base_remote test suite (test_auth_check_remote_no_http_context, test_xmlrpc_login_failure, test_xmlrpc_login_ok) with AssertionError: Found unexpected attributes on base: remote. Alternative approach in #3611 addresses the same issue without breaking tests.

PR #3614, opened 2026-05-20
[19.0][MIG] attachment_queue: Migration to 19.0

dnplkndll:
Port of `attachment_queue` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

Non-mechanical adaptations worth flagging:

- Tests refactored from `odoo_test_helper.FakeModelLoader` → native `odoo.orm.model_classes.add_to_registry`. `odoo-test-helper` is incompatible with 19 (`MetaModel.module_to_models` was renamed/privatized).
- Two test-framework follow-ups skipped with reason inline: lock-contention tests require committed state visible to a second psycopg connection, which Odoo 19's test framework forbids; `test_run_fails_rollback`'s `email_to` assertion was dropped because the mail-template renderer bypasses the registry-augmented fake's `_get_failure_emails` override.

PR #3615, opened 2026-05-20
[19.0][MIG] attachment_synchronize: Migration to 19.0

dnplkndll:
Port of `attachment_synchronize` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

Depends on #3614. The `[DO NOT MERGE]` commit installs `attachment_queue` via `whool` from the open-PR branch until #3614 merges and the 19.0 wheel publishes; the pin is reverted before merge.

Non-mechanical adaptations:

- Test fixtures (`fs.storage` backend + 6 tasks) built inline in `tests/common.py::setUpClass` — OCA CI doesn't load demo data, and `fs_storage` 19.0 also dropped its `fs_storage_demo` record.
- `Registry.enter_test_mode(cr)` → `self.registry_enter_test_mode()` in `test_import.py` (API moved from the Registry to the TestCase in 19).


dnplkndll:
@florian-dacosta @sebastienbeau

  Depends on #3614 

  Two test fixtures rebuilt for the no-demo CI posture: setUpClass  creates the fs.storage backend + import/export tasks inline (was  env.ref to fs_storage.fs_storage_demo and own demo XML, neither loaded in CI); test_import.py uses self.registry_enter_test_mode() in place of the gone Registry.enter_test_mode(cr)

PR #3429, opened 2025-10-24
[IMP] base_search_fuzzy - add support of translatable fields

yvaucher:
depends on:

- https://github.com/odoo/odoo/pull/232993

This allow to create a trigram index on translatable fields.

Searches on translatable fields look like this:

```sql
SELECT * FROM table WHERE COALESCE({column_name}->>'{lang}', {column_name}->>'en_US' % 'text'
```

On a DB with 8M dummy products (duplicated with `odoo populate`),

With a query containing

```sql
WHERE COALESCE("product_template"."name"->>'fr_FR', "product_template"."name"->>'en_US') % 'search text')
```

Without index
Execution Time: 10606.619 ms

With index:
Execution Time: 484.383 ms

:warning: For en_US to work with operator `%` this requires a patch on the core https://github.com/odoo/odoo/pull/232993

PR #3620, opened 2026-05-20
[19.0][MIG] base_search_fuzzy: Migration to 19.0

dnplkndll:
Port of `base_search_fuzzy` from 18.0 to 19.0.
Migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

Includes [#3429](https://github.com/OCA/server-tools/pull/3429) (translatable-fields trigram support) and its follow-up history, replayed as 18.0 commits with original authorship.

- The 18.0 `post_load` hook mutated `odoo.osv.expression.TERM_OPERATORS` / `SQL_OPERATORS` to register the trigram `%` operator. That path is gone in 19.0 (`SQL_OPERATORS` moved to `odoo.orm.utils`; the `Domain` validator keys off `odoo.orm.domains.CONDITION_OPERATORS`; `Field._condition_to_sql` raises `NotImplementedError` for unknown operators). Re-implemented by registering at import time via `@operator_optimization(['%'])` and emitting the predicate through `Domain.custom(to_sql=...)`. `post_load` + `hooks.py` removed.
- The `%` predicate is emitted parenthesized: a translatable field renders as the bare JSON accessor `name->>'en_US'`, and `%` binds tighter than `->>`, so the unparenthesized form fails with `operator does not exist: jsonb ->> boolean`. On 18.0 that wrapping comes from [odoo/odoo#232993](https://github.com/odoo/odoo/pull/232993) (still open); on 19.0 the operator is emitted by this module's own `Domain.custom`, so the parentheses make this PR self-contained with no dependency on that core PR.
- The generated translatable index name is folded to lowercase: `CREATE INDEX` folds the unquoted identifier, but `get_not_used_index` compares it literally against `pg_indexes`, so the mixed-case lang component (`name_en_US_gin_idx`) never matched.

## 19.0 relevance
Odoo 19 core added trigram support, but only for **`like`/`ilike`/`=` acceleration** via a declarative `index='trigram'` field attribute (`fields_textual.py` `condition_to_sql` prefilter, incl. translatable jsonb fields). Core does **not** expose a fuzzy **`%` similarity** domain operator (`%` is absent from `CONDITION_OPERATORS`; no `word_similarity`/`set_limit` in core ORM), nor dynamic UI-driven index management. Those remain this module's purpose, so it stays distinct on 19.0. Core's translatable trigram support covers only its own `like`/`ilike` path; the `%` similarity equivalent for translatable fields is what #3429 (included here) provides.


base_search_fuzzy/models/trgm_index.py
```diff
@@ -0,0 +1,214 @@
...
+                self.env._(
+                    "Cannot create a trigram index on '%(table)s.%(column)s': "
+                    "the column is jsonb (translatable field on Odoo 19.0). "
+                    "PostgreSQL's pg_trgm operator classes only accept text "
+                    "columns. Choose a non-translatable field, or remove the "
+                    "translate flag from the field definition.",
+                    table=table_name,
+                    column=column_name,
+                )
+            )
```
carlos-lopez-tecnativa: I think this commit is out of scope for a module migration: https://github.com/OCA/server-tools/pull/3620/changes/a383be434e9c05a15ccc7da1edfa00724d462549 It would be better to create a new PR for this feature.

Additionally, you restrict this to non-translated fields, but this PR adds support for translated fields, which is another reason not to merge this change into the migration: https://github.com/OCA/server-tools/pull/3429

After reading this code, it seems that Odoo core already supports translated fields with trigram indexes. Could you please check this?
https://github.com/odoo/odoo/blob/1572ce495bcc41df7532c659dc5bc9c7cef50715/odoo/orm/fields_textual.py#L427-L448

carlos-lopez-tecnativa (changes requested):
Please squash the administrative commits, and all commits related to the migration can be squashed into a single commit (keeping the pre-commit auto-fix commit separate).

base_search_fuzzy/models/trgm_index.py
dnplkndll: on further review you are correct on that. this was not necessary. updated

carlos-lopez-tecnativa (changes requested):
In my previous comment, I suggested squashing the administrative commits, but I noticed that you squashed commits that should not have been squashed and merged them into the previous commit, losing the original author. Please review this and rework the Git history correctly.

You can read this documentation if it helps:
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

dnplkndll:
> In my previous comment, I suggested squashing the administrative commits, but I noticed that you squashed commits that should not have been squashed and merged them into the previous commit, losing the original author. Please review this and rework the Git history correctly.
> 
> You can read this documentation if it helps: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

thanks I was trying to write a script and looking for something specific.

carlos-lopez-tecnativa (approved):
LGTM, just a non-blocking comment: I noticed this commit: https://github.com/OCA/server-tools/pull/3620/changes/464431871249074543c4ffc9ac5ac71704ee8266

It has an unusual title, and I would expect it to have a description such as:

`[IMP] base_search_fuzzy: pre-commit auto fixes`

But anyway, this is not a blocker.

@pedrobaeza Could you please review this?

pedrobaeza:
Yes, please follow migration guideline

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

and please include #3429

dnplkndll:
@pedrobaeza I got the pick.

base_search_fuzzy/demo/TrgmIndex_demo.xml
```diff
@@ -0,0 +1,6 @@
+<odoo>
+    <record id="trgm_index_gin_partner_name" model="trgm.index">
+        <field name="field_id" ref="base.field_res_partner__name" />
+        <field name="index_type">gin</field>
```
carlos-lopez-tecnativa, line 4: Please define a language to prevent the field from being empty if it is required at the view level.
Image


```suggestion
        <field name="field_id" ref="base.field_res_partner__name" />
        <field name="index_type">gin</field>
        <field name="lang">en_US</field>
```

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):
LGTM

TT58599

pedrobaeza (approved):
/ocabot merge nobump

PR #3589, opened 2026-04-01
[18.0][MIG] field_vector: Migration to 18.0

IJOL:
## Summary
- Port field_vector from 16.0 to 18.0
- Replace fields.Default with fields.SENTINEL (renamed in v18)
- Fix pre_init_hook signature: (cr) -> (env)
- Fix isinstance calls: (A, B) -> A | B (ruff UP038)
- Fix line length issues
- Version bump to 18.0.1.0.0
- Pre-commit auto fixes applied

## Tests
- Module installs successfully (requires pgvector PostgreSQL extension)
- No dedicated test suite in module

PR #3623, opened 2026-05-22
[18.0][MIG] field_vector: Migration to 18.0

etobella:
Continues https://github.com/OCA/server-tools/pull/3589

Also fixed some bugs:

- On the init, it was declared incorrectly. Fixed it
- Improved dimensions computation
- Added some notes on a problem that might arise that could happen.
- I detected a small license issue
- Fixed some function declaration errors


dnplkndll:
@etobella is there a proper way to document the pg extension requirements for a module? nothing in the manifest I guess.

rvalyi:
> @etobella is there a proper way to document the pg extension requirements for a module? nothing in the manifest I guess. 

Not any declarative one (not in the manifest) I 'm afraid. But we could log something meaningful in the log if the extension is missing I guess...

etobella:
@sbidoul @lmignon this module has a lot of sense with AI and I will propose a config module by using AI after tjis is merged. Makes sense to keep it here or would be better to port it to github.com/OCA/ai?

rvalyi:
> @sbidoul @lmignon this module has a lot of sense with AI and I will propose a config module by using AI after tjis is merged. Makes sense to keep it here or would be better to port it to github.com/OCA/ai?
> 

I would move it to the ai repo. Another reason is server-tools is already quite bloated...

lmignon (approved):
LGTM (Code review)

lmignon:
@etobella Even though vectors are heavily used in AI, they are also widely used outside of it 😏.

In our case, we use vectors to compute similarities between products by representing attribute values as vectors. This is not AI; it's simply mathematics (or vector arithmetic). https://www.odoo.com/fr_FR/event/odoo-experience-2025-6601/track/vector-based-product-similarity-in-odoo-a-new-paradigm-for-recommendations-8120

That said, it's true that the server-tools repository is packed with a large number of addons, and having a smaller, more focused repository could be beneficial. So I'm not opposed to moving it to the AI repository.

lmignon:
> > @etobella is there a proper way to document the pg extension requirements for a module? nothing in the manifest I guess.
> 
> Not any declarative one (not in the manifest) I 'm afraid. But we could log something meaningful in the log if the extension is missing I guess...

@rvalyi  @dnplkndll
 This addon don't use additional pg extension since it's not possible with the odoo ORM.  It register it's own extension and own datatype designed to properly works with the orm and its internal cache.

etobella:
Well, after your answer, I think we can keep it here and leave the config (includes AI) in the AI repository.

/ocabot migration field_vector

luisDIXMIT (approved):
Code review, LGTM!

PR #3630, opened 2026-05-27
[16.0][IMP] session_db: use orjson when available

lmignon:
orjson provides better performance than Python's standard json library. Use it when installed.


lmignon:
ping @sbidoul

sbidoul:
A test is failing.

lmignon:
Closing since the performance gain is not significant for such small json documents... The CPU time spent in json.dumps/loads is not significant — round-trip latency to PostgreSQL dominates.

PR #3632, opened 2026-05-29
[19.0][MIG] database_size: Migration to 19.0

dnplkndll:
Port of `database_size` from 18.0 to 19.0 ([migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0)).

- `read_group` override → `_read_group`: 19.0 deprecated the public `read_group` (would fail checklog) and the web client no longer routes group views through it, so the old override was both noisy and dead. Re-implemented on the non-deprecated `_read_group` with the 19.0 signature (`groupby`/`aggregates`/`having`/`order`); it still just defaults the group order for the size-report views.
- `_sql_constraints` → `models.Constraint` (kept the `uniq_model_measurement_date` name).
- `_compute_model_name`: justified `# pylint: disable=no-search-all` on the `ir.model` load (small registry table; new 19.0 pylint-odoo check). The `information_schema` references are raw table-size SQL, unrelated to translatable-field jsonb assumptions.


StefanRijnhart (commented):
Thank you. Can I ask why the commits don't reflect the history of the module in 18.0? https://github.com/OCA/server-tools/commits/18.0/database_size

StefanRijnhart:
@FREDDYGAMBOABEDOR I don't think @dnplkndll needs you to burn tokens to explain the OCA processes. The original PR comment above links to the source of your walkthrough (and referring to that link is all that your comment should have done in the first place)

dnplkndll:
> Thank you. Can I ask why the commits don't reflect the history of the module in 18.0? https://github.com/OCA/server-tools/commits/18.0/database_size

sorry my bot squash script was not right back then. looks good now the diff just seems to be 19.0 rebase activity.

database_size/report/ir_model_size_report.py
```diff
@@ -0,0 +1,236 @@
...
+                    )
+                new_domain.append((1, "=", 1))
+            else:
+                new_domain.append(clause)
+        if values:
+            self = self.with_context(**values)
+        return self, new_domain
+
+    @api.model
+    def _where_calc(self, domain, active_test=True):
```
StefanRijnhart: This method no longer exists. When opening the comparison menu and selecting a measurement date and/or a historic measurement date (for which there is data available), `_table_query` will still provide the default contents because these dates have not been inserted in the context. The domain will still filter out the selected dates, so if these are not the defaults the result is an empty list.

This also implies a lack of test coverage (mea culpa).

Data setup:

Image

Report selection:

Image

Snippet from the query returned by `_table_query` using the selection above:

```
select '2026-07-01'::date as measurement_date,
    '2026-06-29'::date as historical_measurement_date,
...
from ir_model_size cur
left join ir_model_size hst
    on cur.model = hst.model
       and hst.measurement_date = '2026-06-29'::date
where cur.measurement_date = '2026-07-01'::date
```

StefanRijnhart (commented):
Code looks good. Data gathering seems to work well. I did find a reporting issue:

database_size/report/ir_model_size_report.py
dnplkndll: sorry missed that where like you said with the test passing. it has shown up often, converted to domain.  does test_database_size_report_web_paths work better, turned green on the convert to domain search.

Issue #1664, opened 2019-09-10
12.0 auto_backup access denied list_db


jcarlosmontoya:
Module does not work when list_db parameter is False

![image](https://user-images.githubusercontent.com/7680456/64592736-f3d26b80-d3ac-11e9-9268-95a256ce0a0a.png)

Odoo check this
![image](https://user-images.githubusercontent.com/7680456/64592860-38f69d80-d3ad-11e9-8fdf-fda391dbc8a8.png)

![image](https://user-images.githubusercontent.com/7680456/64592905-4a3faa00-d3ad-11e9-85d0-26eacfdd1298.png)

![image](https://user-images.githubusercontent.com/7680456/64593115-c33f0180-d3ad-11e9-959f-498693bdc908.png)

OmniaGit:
In case of cloud database we always disable the list of it.
so in this case will be very useful having a field that allows to specifies if you want to backup a single database or all the database..

mcerutti:
I'm having the same issue on v14. There's any update ?

mathben:
**This is the reason**
Message :
[MERGE] forward port branch saas-15 up to 43eca3079e4

Author : 
Christophe Simonis

Time :
Commit #b29e053436b69223e25c49955006c910cd59316d

Date :
Mon Aug 06 05:32:12 EDT 2018

**Solution**
We can support to backup multiple database, this is what you want? Or we can disable this feature when no need it?
The reason is to not backup another database then yours when you share the machine.

Issue #1310, opened 2018-07-07
11.0 backup access denied 


abdallamhafeez:
Hi
I have issue when take backup via `auto_backup` and make option `list_db = False` appear error in log `
odoo.exceptions.AccessDenied: Access denied` and backup file created with zero size

jcarlosmontoya:
Hello, this module does not work with list_db = False config parameter

![image](https://user-images.githubusercontent.com/7680456/64592565-93432e80-d3ac-11e9-95e3-45e48ad47783.png)

kos94ok-3D:
Related: #1664

PR #3639, opened 2026-06-06
[18.0][FIX] auto_backup: allow scheduled backup when list_db is disabled

klodr:
## Problem

jobs fail with:

`list_db = False` is a common hardening that only aims at hiding database
management from the **web** interface; a scheduled backup is a trusted
server-side operation and should still run.

## Fix

restores its original value (`try/finally`, even on error). A regression test
(`test_action_backup_local_list_db_disabled`) is added.

## Related

- Long-standing issue, reported in #1310 and #1664.
- #2904 (15.0) proposed the same toggle but without `try/finally` safety or a
  test; #2419 was a different, closed approach.

This is the base PR (18.0); ports to 17.0 / 16.0 / 15.0 / 14.0 follow.

### DIFF
--- a/auto_backup/README.rst
+++ b/auto_backup/README.rst
@@ -113,7 +113,6 @@ Known issues / Roadmap
   changing settings, you need to run the backup from outside of the main
   Odoo instance. How to do this is outlined in `this blog
   post <https://blog.laslabs.com/2016/10/running-python-scripts-within-odoos-environment/>`__.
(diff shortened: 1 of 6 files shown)


klodr:
@OCA/server-tools-maintainers gentle ping — all checks green and mergeable. This is the 18.0 head of the series (#3640 17.0, #3641 16.0, #3642 15.0, #3643 14.0); it may be the easiest entry point for review.

PR #3640, opened 2026-06-06
[17.0][FIX] auto_backup: allow scheduled backup when list_db is disabled

klodr:
## Problem

jobs fail with:

## Fix

restores its original value (`try/finally`, even on error). Adds a regression
test (`test_action_backup_local_list_db_disabled`).

Port of #3639.


klodr:
@OCA/server-tools-maintainers gentle ping — 9/9 checks green, no conflicts, open since June without review. Same fix as #3639 / #3641 / #3642 / #3643, backported per branch.

PR #3642, opened 2026-06-06
[15.0][FIX] auto_backup: allow scheduled backup when list_db is disabled

klodr:
## Problem

jobs fail with:

## Fix

restores its original value (`try/finally`, even on error). Adds a regression
test (`test_action_backup_local_list_db_disabled`).

Port of #3639.

### DIFF
--- a/auto_backup/README.rst
+++ b/auto_backup/README.rst
@@ -107,7 +107,6 @@ Known issues / Roadmap
   you need to run the backup from outside of the main Odoo instance. How to do
   this is outlined in `this blog post
   <https://blog.laslabs.com/2016/10/running-python-scripts-within-odoos-environment/>`_.
-* Backups won't work if list_db=False is configured in the instance.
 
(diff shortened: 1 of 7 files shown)


klodr:
The failing `pre-commit` check is unrelated to this change.

The only red hooks are `setuptools-odoo-make-default` and `setuptools-odoo-get-requirements`, which both crash at import with:

```
ModuleNotFoundError: No module named 'pkg_resources'
```

(recent `setuptools` no longer bundles `pkg_resources`). This breaks the `pre-commit` run for the whole 15.0 branch, independently of this PR — see the [failed job](https://github.com/OCA/server-tools/actions/runs/27057181027/job/79863577023).

The change itself passes `black`, `isort`, `flake8`, `pylint_odoo` and both test jobs (OCB + Odoo). Happy to rebase once the branch's `setuptools-odoo` pin is bumped.

PR #3646, opened 2026-06-09
[19.0][FIX] auditlog: Maximum recursion error when creating new accounts

hi-oerp:
### Module
Auditlog

### Describe the bug
Maximum recursion error when creating new accounts

### To Reproduce
Create a new rule for the account.account module
Check "Log Creates", "Log Writes" and "Full Log" in the new rule

Go to Accounting -> Configuration -> Chart of Account
Create a new account with code
Foreve loading and "RecursionError: maximum recursion depth exceeded"

**Expected behavior**
It should be able to save

**Additional context**
In Odoo, there are 2 methods in the AccountAccount class called _compute_code and _inverse_code.
Auditlog module keeps triggering these 2 methods and causing an infinite loop.

### DIFF
--- a/auditlog/models/auditlog_rule.py
+++ b/auditlog/models/auditlog_rule.py
@@ -365,7 +365,7 @@ class AuditlogRule(models.Model):
 
         @api.model_create_multi
         def create_full(self, vals_list, **kwargs):
-            self = self.with_context(auditlog_disabled=True)
+            self = self.with_context(auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             new_records = create_full.origin(self, vals_list, **kwargs)
             # Take a snapshot of record values from the cache instead of using
@@ -400,7 +400,7 @@ class AuditlogRule(models.Model):
 
         @api.model_create_multi
         def create_fast(self, vals_list, **kwargs):
-            self = self.with_context(auditlog_disabled=True)
+            self = self.with_context(auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             vals_list = rule_model._update_vals_list(vals_list)
             vals_list2 = copy.deepcopy(vals_list)
@@ -443,9 +443,9 @@ class AuditlogRule(models.Model):
             # avoid logs on `read` produced by auditlog during internal
             # processing: read data of relevant records, 'ir.model',
             # 'ir.model.fields'... (no interest in logging such operations)
-            if self.env.context.get("auditlog_disabled"):
+            if self.env.context.get("auditlog_disabled_read"):
                 return result
-            self = self.with_context(auditlog_disabled=True)
+            self = self.with_context(auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             if self.env.user in users_to_exclude:
                 return result
@@ -469,7 +469,12 @@ class AuditlogRule(models.Model):
         users_to_exclude = self.mapped("users_to_exclude_ids")
 
         def write_full(self, vals, **kwargs):
-            self = self.with_context(auditlog_disabled=True)
+            guard = self.env.context.get("auditlog_guard", set())
+            guard_keys = {(self._name, tuple(self.ids))}
+            if bool(guard.intersection(guard_keys)):
+                return write_full.origin(self, vals, **kwargs)
+            guard.update(guard_keys)
+            self = self.with_context(auditlog_guard=guard, auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             fields_list = rule_model.get_auditlog_fields(self)
             records_write = (
@@ -503,7 +508,7 @@ class AuditlogRule(models.Model):
             return result
 
         def write_fast(self, vals, **kwargs):
-            self = self.with_context(auditlog_disabled=True)
+            self = self.with_context(auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             # Log the user input only, no matter if the `vals` is updated
             # afterwards as it could not represent the real state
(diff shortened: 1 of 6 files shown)


hi-oerp:
Good morning.


@sebalix @StefanRijnhart @pedrobaeza @hbrunn  Could you please help me validate this so we can proceed with the approval and merge?

Thanks!

StefanRijnhart (commented):
Thanks! Can you add a reproducing test to https://github.com/OCA/server-tools/tree/19.0/test_auditlog?

StefanRijnhart:
If you check the failing tests, it looks like your changes prevent some other expected logs to be created, so it's not all well.

hi-oerp:
> Thanks! Can you add a reproducing test to https://github.com/OCA/server-tools/tree/19.0/test_auditlog?

Hello @StefanRijnhart I created the tests to verify the infinite recursion issue.

hi-oerp:
> If you check the failing tests, it looks like your changes prevent some other expected logs to be created, so it's not all well.

Hello @StefanRijnhart I have already verified the changes. Could you please regenerate the unit tests for the PR?

Thanks!

auditlog/models/auditlog_rule.py
```diff
@@ -469,38 +471,52 @@ def _make_write(self):
         users_to_exclude = self.mapped("users_to_exclude_ids")
 
         def write_full(self, vals, **kwargs):
-            self = self.with_context(auditlog_disabled=True)
```
StefanRijnhart, line 472: We will want to keep this key, because it is specifically used to disable read logging. Please rename it to auditlog_disabled_read to distinguish it from the new recursion check mechanism.

test_auditlog/tests/test_account_reentrancy.py
```diff
@@ -0,0 +1,57 @@
...
+                "log_create": True,
+                "log_write": True,
+                "log_read": True,
+                "log_unlink": True,
+                "log_type": "full",
+            }
+        )
+        cls.rule.set_to_confirmed()
+
+    def test_write_code_reentrancy(self):
```
StefanRijnhart, line 25: Please add a docstring that specifically mentions the recursive character of account.account's _inverse_code / _compute_code triggering a RecursionError

StefanRijnhart (commented):
Thanks! Nice work on building an actual recursion prevention mechanism. However, I don't think we need to resort to manipulating threads but instead add the auditlog_guard dictionary to the context. Because it is an editable, it can still be updated in nested calls. Would you mind refactoring it?

hi-oerp:
> Thanks! Nice work on building an actual recursion prevention mechanism. However, I don't think we need to resort to manipulating threads but instead add the auditlog_guard dictionary to the context. Because it is an editable, it can still be updated in nested calls. Would you mind refactoring it?

Thank you very much. @StefanRijnhart 

As a first approach, I tried implementing this using the context, but in some cases the context is lost, which causes the recursion issue to reappear. I also attempted to pass the control flag through the transaction, but the recursion continue.

As a last resort, I had to implement the guard using a thread variable.

Note: If this solution is acceptable, I will also need to apply the same implementation to the fast log.

Thanks in advance!

hi-oerp:
Hello @StefanRijnhart I have pushed the latest changes. I also added the context used to skip the read operation.

Please let me know if you have any comments or suggestions. Thank you very much for your feedback!

StefanRijnhart:
Can you be more specific under which the context gets lost? I have a POC here locally with the guard in the context and the test still succeeds.

test_auditlog/tests/test_account_reentrancy.py
```diff
@@ -0,0 +1,63 @@
...
+        """Regression test for recursive audit logging.
+        Ensures that writes triggered by computed/inverse fields (code/code_store)
```
StefanRijnhart: What is `code_store`? Please mention `account.account's _inverse_code / _compute_code` literally.

hi-oerp:
Hello @StefanRijnhart 

Thank you very much for the code review. I've addressed all the requested changes.

Could you please take another look? If everything looks good, I'd appreciate it if you could proceed with the integration. Once it's merged, I'll be able to close the remaining ticket I have for this module.

See you!

hi-oerp:
Hello @StefanRijnhart , Could you please validate whether it is possible to merge this PR?

Thank you in advance!

StefanRijnhart (approved):
Thanks, looking good!

auditlog/models/auditlog_rule.py
```diff
@@ -469,21 +469,31 @@ def _make_write(self):
...
+            guard_keys = {(self._name, tuple(self.ids))}
+
+            if bool(guard.intersection(guard_keys)):
+                return write_full.origin(self, vals, **kwargs)
+
+            guard.update(guard_keys)
+            self = self.with_context(auditlog_guard=guard, auditlog_disabled_read=True)
             rule_model = self.env["auditlog.rule"]
             fields_list = rule_model.get_auditlog_fields(self)
+
```
pedrobaeza: Please don't add empty lines inside the method.

pedrobaeza (commented):
Can you remove `[19.0]` from the commit message?

hi-oerp:
Hello @pedrobaeza , The recommendations were implemented!

Thank you!

auditlog/models/auditlog_rule.py
pedrobaeza: You are still adding a lot of empty lines in your new code.

hi-oerp: Hello. @pedrobaeza
Unnecessary whitespace has been removed. Thank you for the suggestion.

pedrobaeza (approved):

StefanRijnhart:
Making major version bump to signal the modified context key name.

/ocabot merge major

PR #3648, opened 2026-06-10
[18.0][ADD] compute_field_after_install

sebalix:
Migrate module from PR:
- #3157


legalsylvain (changes requested):
There was a lot of questions in the initial PR regarding the design. (That has never been merged.)
Could you answer here before migrating the module ?
https://github.com/OCA/server-tools/pull/2774
Thanks !

sebalix:
@legalsylvain I updated the module description with a warning section.
Regarding the order on `id DESC` it has been done on 16.0. For the use of `pre_init_hook` I'm lacking of context why it hasn't been done that way, probably to reduce the upgrade time to its minimum.

legalsylvain:
Hi . Sorry I didn't express myself clearly. I'd like we finish the discussion regarding the opportunity or not to have this module in the OCA in the first PR, to avoid arguments in many places. If you think this module is relevant, could you expose argument regarding the previous pending remarks ? 
- describe a real use case, not solved in the "classic" way : https://github.com/OCA/server-tools/pull/2774#issuecomment-1839391285 by @legalsylvain 
- raising a lot of side-effects, and the gate to the hell : https://github.com/OCA/server-tools/pull/2774#issuecomment-1839387212 by @pedrobaeza 
- ...

Thanks !

PR #3660, opened 2026-06-22
[18.0][FIX] base_name_search_improved: Call the name_search() method with the appropriate limit

victoralmau:
Call the `name_search()` method with the appropriate limit

It is important to call the name_search() method with the appropriate limit so that the filter is truly effective

Example use case:
- Thousands of products
- We apply filter A + smart_search = B
- By default, name_search() applies limit=100, so we will only get a total of 100 records to build the domain in _search_smart_search()
- When applying the A filter, the smart_search domain will return some incorrect records (4, for example); however, if the total number of possible records were specified as the limit, the correct domain would be returned (20 possible records, for example)

@Tecnativa TT63353


flotho (approved):
easy code review, LGTM

PR #3663, opened 2026-06-24
[IMP] base_fontawesome: add fontawesome_picker field widget

MarianoMartinezGC:
## What

Add a reusable OWL field widget, `fontawesome_picker`, that can be applied to
any `Char` field storing a FontAwesome CSS class.

`<field name="icon" widget="fontawesome_picker"/>`

Instead of typing the class by hand, the user gets a searchable grid of icons
with a live preview and a clear button. Selecting an icon writes its full class
(e.g. fa fa-shopping-cart) into the field.

Why

Fields that store a FontAwesome class are currently filled by typing the class
name from memory. This is error-prone (a typo renders nothing) and slow, since
it forces the user to look the icon up outside Odoo. There is no free OCA widget
for this today.

How

- The icon catalog is read at runtime from the loaded stylesheets (the same
technique as web_editor's icon selector), so it always reflects the
FontAwesome version shipped by this module — no hard-coded list to maintain.
The selector-scan is replicated locally so the module keeps depending on web
only (avoiding web_editor and its globally-cached computeFonts).
- Built as a Component-based field (not a CharField subclass) so the search
input inside the dropdown does not interfere wit
- The dropdown is rendered through the popover service (floats in the viewport,
single scroll); the rendered grid is capped, with a hint to refine the search.
- Registered for form views only.
- Covered by a QUnit suite (catalog read, search filtering, selection, clear,
readonly rendering, capped grid, empty-catalog fallback).

How to test

On a runboat build (database with all modules),
widget="fontawesome_picker" to any Char field of
via Studio or a quick view override). Open the form: the field shows a button
with the current icon; clicking it opens a searchable grid; picking an icon
writes its class and previews it; the clear button empties the value.

FL-556-9206


IVA-BO (approved):

rodrigogarciafl (approved):

MarianoMartinezGC:
@legalsylvain

PR #3665, opened 2026-07-03
[18.0][FIX] autovacuum_message_attachment: Add missing message_type 'auto_comment' and allow to batch deletion

divad1196:
- `auto_comment` message_type was missing
  => Add the option
- Deletion could timeout due to the number of records to delete
  => Added `batch_size` parameter to the rules
- Optimization:
  - `search` performed `orderby` which forces an extra sort on the whole table
    => Use `_search` instead
  -   `unlink` is batched by Odoo natively, and `batch_unlink` was also batching with its own value
    => Make them use the same limit and added a batch size parameter to `batch_unlink`

### DIFF
--- a/autovacuum_message_attachment/models/autovacuum_mixin.py
+++ b/autovacuum_message_attachment/models/autovacuum_mixin.py
@@ -10,18 +10,36 @@ from odoo.tools.safe_eval import datetime, safe_eval
 
 _logger = logging.getLogger(__name__)
 
+# Number of records deleted per unlink() call. Bounds the id list, the SQL
+# DELETE size and the per-commit work, independently of how many records the
+# rule retrieves (rule.batch_size).
+UNLINK_BATCH_SIZE = 5000
+
 
 class AutovacuumMixin(models.AbstractModel):
     _name = "autovacuum.mixin"
     _description = "Mixin used to delete messages or attachments"
 
-    def batch_unlink(self):
+    def batch_unlink(self, batch_size=0):
+        # batch_size == -1 => delete everything in a single unlink (no chunking).
         with Registry(self.env.cr.dbname).cursor() as new_cr:
+            if batch_size == -1:
+                batch_size = len(self)
+            if not batch_size or batch_size < new_cr.IN_MAX:
+                batch_size = new_cr.IN_MAX
+            elif batch_size > new_cr.IN_MAX:
+                # Adapt cursor IN_MAX to batch_size if needed:
+                # unlink silently chunk using IN_MAX (1000-rows)
+                # Assigning on the instance shadows the class attribute
+                # for this cursor only, not the class
+                # (Cursor.IN_MAX stays 1000)
+                new_cr.IN_MAX = batch_size
             new_env = api.Environment(new_cr, self.env.uid, self.env.context)
             try:
-                while self:
-                    batch_delete = self[0:1000]
-                    self -= batch_delete
+                remaining = self
+                while remaining:
+                    batch_delete = remaining[0:batch_size]
+                    remaining -= batch_delete
                     # do not attach new env to self because it may be
                     # huge, and the cache is cleaned after each unlink
                     # so we do not want to much record is the env in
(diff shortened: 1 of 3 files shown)


autovacuum_message_attachment/models/autovacuum_mixin.py
```diff
@@ -5,22 +5,41 @@
...
     _name = "autovacuum.mixin"
     _description = "Mixin used to delete messages or attachments"
 
-    def batch_unlink(self):
+    def batch_unlink(self, batch_size=0):
+        # batch_size <= 0 => delete everything in a single unlink (no chunking).
         with Registry(self.env.cr.dbname).cursor() as new_cr:
+            if batch_size == -1:
+                batch_size = len(self)
+            if not batch_size or batch_size < 0:
```
divad1196: ~~TODO: use `if not batch_size or batch_size < new_cr.IN_MAX:` instead~~

Done

simahawk: TODO Still valid?

simahawk (approved):
LGTM

autovacuum_message_attachment/models/autovacuum_mixin.py
divad1196: yes, change done

autovacuum_message_attachment/models/autovacuum_mixin.py
```diff
@@ -5,22 +5,41 @@
...
     _name = "autovacuum.mixin"
     _description = "Mixin used to delete messages or attachments"
 
-    def batch_unlink(self):
+    def batch_unlink(self, batch_size=0):
+        # batch_size == -1 => delete everything in a single unlink (no chunking).
         with Registry(self.env.cr.dbname).cursor() as new_cr:
+            if batch_size == -1:
+                batch_size = len(self)
+            if not batch_size or batch_size < new_cr.IN_MAX:
```
TDu, line 28:
```suggestion
            if not batch_size:
```
Why not letting the option to the caller to have a batch size smaller than `IN_MAX`

autovacuum_message_attachment/models/autovacuum_mixin.py
```diff
@@ -5,22 +5,41 @@
...
+UNLINK_BATCH_SIZE = 5000
+
 
 class AutovacuumMixin(models.AbstractModel):
     _name = "autovacuum.mixin"
     _description = "Mixin used to delete messages or attachments"
 
-    def batch_unlink(self):
+    def batch_unlink(self, batch_size=0):
+        # batch_size == -1 => delete everything in a single unlink (no chunking).
```
TDu, line 24: Why this option when it open the risk of Out Of Memory errors ?

Also mutating the `IN_MAX` value above a certain value is probably not recommended ! IMO

autovacuum_message_attachment/models/autovacuum_mixin.py
divad1196: @TDu Because a small batch size is almost never a good idea, especially due to IO and database query optimization.

On the other side, Odoo uses `IN_MAX` everywhere: If you had a reason to want smaller batch size, you would likely need it changes for the whole Odoo, not just the batch operations.

autovacuum_message_attachment/models/autovacuum_mixin.py
divad1196: It does not increase the risk of OOM, because little resources are allocated.
The GC will also not run between your batches, because your records are still referenced. 
https://github.com/odoo/odoo/blob/03256802105c6273f62a5305d2395d7771418e0b/odoo/orm/models.py#L4232
https://github.com/odoo/odoo/blob/18.0/odoo/tools/misc.py#L692
On the other side, allowing to increase this limit reduces the risk of timeouts and idle transactions.

The max value acceptable for `IN_MAX` will depend on the context of the database. I was able to raise it to 1M on the customer database without issue, but it was not providing any performance improvment compared to batches of 50k.

divad1196:
@TDu I answered your questions. could you check?

TDu (approved):
LG. although I still thing the batch_size = -1 is not a good idea, because the `IN_MAX` has no upper limit, then.
But it does not look like it is a restriction anymore for Postgresql https://www.postgresql.org/docs/18/limits.html ?

PR #3666, opened 2026-07-06
[IMP] module_auto_update: Auto-update per module version.

amh-mw:
This moves #3353 `module_version_auto_upgrade` logic (original author @ryanc-me) into existing `module_auto_update` module, but disabled by default.


amh-mw:
I have to set this down briefly to deal with some more urgent requests at work, but my intention here is to refactor far enough to have all of original functionality of `module_version_auto_upgrade`, but with auto-update disabled so as not to disrupt existing `module_auto_update` users. I'm still working on the test coverage and documentation, especially with regards to the auto-update toggle.

I don't intend to do auto-update for checksums; it has been added to the roadmap.

PR #3672, opened 2026-07-15
[IMP] sequence_regex: Function to match a string to a sequence

AlienAtSystem:
This is a new module to provide a function that checks whether a string has been generated by a sequence, or rather, if it is formatted such that it fits the scheme in which the sequence generates numbers.

The main purpose is to provide support for dynamic placeholders in sale_quotation_number, which needs to detect if a sale.order name has been generated by the sale.quotation sequence to replace it with a name from the sale.order sequence - the corresponding Pull Request can be found [here](https://github.com/OCA/sale-workflow/pull/4463). However, as it's not dependent on sale and might be useful in other contexts, it is provided as a separate module in server-tools.


rousseldenis (approved):
LGTM

AlienAtSystem:
Is there more steps I need to initiate to get this merged?

PR #3673, opened 2026-07-15
[19.0][FIX] upgrade_analysis: reset discarded fields to their default, not falsy value

MiquelRForgeFlow:
When the previous version set a field explicitly and the new version omits it, _get_xml_diff generated eval="False" using field.falsy_value. This is wrong for fields whose default is truthy (e.g. ir.rule.perm_read/write/create/unlink or active): a fresh install applies the field's default, not its falsy value. The generated noupdate_changes.xml then mismatched a clean install and could even violate model constraints (e.g. ir.rule's _no_access_rights CHECK), breaking the migration.

Use model.default_get() to reset omitted fields to the value a fresh install would give them, falling back to falsy_value when there is no default or it is not a scalar.

Fixes https://github.com/OCA/server-tools/pull/3535.

Bug discovered doing https://github.com/OCA/OpenUpgrade/pull/5820.

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -370,11 +370,23 @@ class UpgradeAnalysis(models.Model):
                         "name"
                     }:
                         # if previous version has set a field but current version
-                        # doesn't, set whatever the NULL value of the field is
-                        # (usually None)
+                        # doesn't, reset it to the value a fresh install would give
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow:
@pedrobaeza @hbrunn

pedrobaeza (approved):
Yes, indeed this is something to fix.

hbrunn (approved):

PR #3675, opened 2026-07-15
[19.0][IMP] database_cleanup: Purge menu has action that does not exist or has been deleted

quoc-pn:
### Problems

Some menus have actions that do not exist or have been deleted. When clicking on the menu "Purge obsolete menu entries" purge lines will be created for obsolete menus and the error below will be raised because of invalid action

image

### Changes

Check whether the action exists by calling exists() before reading its data. If the action does not exist, create a purge line for the corresponding menu.

### DIFF
--- a/database_cleanup/wizards/purge_menus.py
+++ b/database_cleanup/wizards/purge_menus.py
@@ -4,6 +4,7 @@
 # pylint: disable=consider-merging-classes-inherited
 from odoo import api, fields, models
 from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
For not repeating the exact code for adding the entry, please refactor it a bit, and you can use `Command.add` for also shorting the `0, 0` lines.

pedrobaeza (approved):

leemannd (approved):

PR #3676, opened 2026-07-15
[19.0][IMP] database_cleanup: Drop dependent views before dropping tables

quoc-pn:
### Problems

The current implementation only removes constraints before dropping tables. However, PostgreSQL also prevents dropping a table when other database objects, such as views or materialized views, depend on it.

```bash
psycopg2.errors.DependentObjectsStillExist: cannot drop table hr_employee because other objects depend on it
DETAIL:  view hr_employee_public depends on table hr_employee
view hr_holidays_remaining_leaves_user depends on table hr_employee
view hr_leave_report_calendar depends on table hr_employee
```

### Solution

Before dropping a table, detect and remove all views or materialized views that reference the table. Once the dependent objects have been removed, the table can be dropped safely without raising `DependentObjectsStillExist`.

### DIFF
--- a/database_cleanup/wizards/purge_line_tables.py
+++ b/database_cleanup/wizards/purge_line_tables.py
@@ -34,49 +34,11 @@ class CleanupPurgeLineTable(models.TransientModel):
             objs = self.env["cleanup.purge.line.table"].browse(
                 self._context.get("active_ids")
             )
-        tables = objs.mapped("name")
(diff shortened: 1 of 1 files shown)


leemannd (approved):

quoc-pn:
@legalsylvain, It would be great if you had some time to review this PR. Thank you!

PR #3671, opened 2026-07-14
[19.0][IMP] upgrade_analysis: generate changes for renamed noupdate records

hbrunn:
when a noupdate record is moved into another module, content changes were not reflected in `noupdate_changes.xml`.  A case where that's really problematic is https://github.com/OCA/OpenUpgrade/issues/5812

This fix will emit the following extra records on the next run of the analysis cronjob after merge:

openupgrade_scripts/scripts/uom/19.0.1.0/noupdate_changes.xml ([v18](https://github.com/odoo/odoo/blob/18.0/addons/product/data/product_data.xml#L35), [v19](https://github.com/odoo/odoo/blob/19.0/addons/uom/data/uom_data.xml#L5))

```xml
  <record id="decimal_product_uom" model="decimal.precision" forcecreate="True">
    <field name="name">Product Unit</field>
  </record>
```

openupgrade_scripts/scripts/hr_holidays/19.0.1.6/noupdate_changes.xml ([v18](https://github.com/odoo/odoo/blob/18.0/addons/hr_holidays_attendance/data/hr_holidays_attendance_data.xml#L4), [v19](https://github.com/odoo/odoo/blob/19.0/addons/hr_holidays/data/hr_leave_type_data.xml#L81))

```xml
  <record id="holiday_status_extra_hours" model="hr.leave.type">
    <field name="icon_id" ref="hr_holidays.icon_4"/>
    <field name="overtime_deductible" eval="False"/>
    <field name="requires_allocation">False</field>
    <field name="country_id" eval="False"/>
    <field name="hide_on_dashboard">True</field>
  </record>
```

openupgrade_scripts/scripts/base/19.0.1.3/noupdate_changes.xml ([v18](https://github.com/odoo/odoo/blob/18.0/addons/l10n_fr/data/res_country_data.xml#L2), [v19](https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/data/res_country_data.xml#L1701))

```xml
  <record id="dom-tom" model="res.country.group">
    <field name="country_ids" eval="[Command.set([                 ref('yt'),ref('gp'),ref('mq'),ref('gf'),ref('re'),                 ref('pf'),ref('pm'),ref('mf'),ref('bl'),ref('nc'),             ])]"/>
    <field name="code">DOM-TOM</field>
  </record>
```

[...]

PR #3677, opened 2026-07-16
[19.0][FIX] upgrade_analysis: don't reset foreign-module fields on moved noupdate records

MiquelRForgeFlow:
When a noupdate record is moved to another module, its previous definition may set a field declared in a module the new one doesn't depend on (e.g. a module that depends on it). That field is still managed by its own module, so skip it instead of resetting it to its default in the new module's noupdate_changes.xml.

Fixes https://github.com/OCA/server-tools/pull/3671.

Detected in https://github.com/OCA/OpenUpgrade/pull/5828 (in `hr_holidays` module, in `holiday_status_extra_hours` xmlid, that comes from `hr_holidays_attendance`).

### DIFF
--- a/upgrade_analysis/models/upgrade_analysis.py
+++ b/upgrade_analysis/models/upgrade_analysis.py
@@ -314,9 +314,27 @@ class UpgradeAnalysis(models.Model):
             return element.text
         return etree.tostring(element)
 
+    def _get_available_modules(self, module_name):
+        """Return the set of module names a given module may reference in its
+        data files: the module itself plus its (transitive) dependencies.
+        """
+        module = self.env["ir.module.module"].search([("name", "=", module_name)])
+        if not module:
+            return set()
+        upstream = module.upstream_dependencies(
+            exclude_states=("uninstalled", "uninstallable", "to remove")
+        )
+        return {module_name} | set(upstream.mapped("name"))
+
     def _get_xml_diff(
-        self, remote_update, remote_noupdate, local_update, local_noupdate
+        self,
+        remote_update,
+        remote_noupdate,
+        local_update,
+        local_noupdate,
+        local_module,
     ):
+        available_modules = self._get_available_modules(local_module)
         odoo = etree.Element("odoo")
         for xml_id in sorted(local_noupdate.keys()):
             local_record = local_noupdate[xml_id]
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
That is not the correct use of `Fixes` in the commit message if I understand correctly. You can say, 'issue introduced in' or 'regression of', and then preferably refer to the PR https://github.com/OCA/server-tools/pull/3671 rather than the commit.

MiquelRForgeFlow:
Ok, fixed PR message, by referencing the PR instead of the commit.

StefanRijnhart (approved):

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -314,9 +314,29 @@ def _get_node_value(element):
...
+        self,
+        remote_update,
+        remote_noupdate,
+        local_update,
+        local_noupdate,
+        local_module,
     ):
+        available_modules = (
+            self._get_available_modules(local_module) if local_module else None
+        )
```
hbrunn:
```suggestion
        available_modules = self._get_available_modules(local_module)
```

it's fine when this is an empty set, right?

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -355,12 +375,22 @@ def _get_xml_diff(
                     # Does the field still exist?
                     if record_remote_dict[key].tag == "field":
                         field_name = remote_record.xpath(key)[0].attrib.get("name")
+                        model_name = local_record.attrib["model"]
                         if (
-                            local_record.attrib["model"] not in self.env
-                            or field_name
-                            not in self.env[local_record.attrib["model"]]._fields.keys()
+                            model_name not in self.env
+                            or field_name not in self.env[model_name]._fields.keys()
```
hbrunn:
```suggestion
                            or field_name not in self.env[model_name]._fields
```

as you touch this line anyways, you can just as well simplify this

upgrade_analysis/models/upgrade_analysis.py
```diff
@@ -355,12 +375,22 @@ def _get_xml_diff(
...
+                        # When the record was moved to another module, its
+                        # previous definition may set a field that is declared
+                        # in a module the current one does not depend on (e.g.
+                        # the field lives in a module that depends on this one).
+                        # That field is still managed by its own module, so it
+                        # must not be reset from here.
+                        if available_modules is not None:
+                            field = self.env[model_name]._fields[field_name]
+                            if field._module and field._module not in available_modules:
+                                continue
```
hbrunn:
```suggestion
                        field = self.env[model_name]._fields[field_name]
                        if field._module and field._module not in available_modules:
                            continue
```

upgrade_analysis/tests/test_module.py
```diff
@@ -150,6 +150,7 @@ def test_xml_comparison(self):
                     """
                 ),
             },
+            None,
```
hbrunn: why not

```suggestion
            "upgrade_analysis",
```

?

hbrunn (changes requested):
please amend `test_analyze` with some data and probably a patch on the fields to cover most of the code introduced here

hbrunn (approved):

PR #3679, opened 2026-07-18
[18.0][FIX] base_exception: enforce blocking check in wizard

### LINKED ISSUE #3667: [base_exception] wizard ignores is_blocking when "Ignore Exceptions" is checked
## Module
base_exception

## Describe the bug
The Ignore Exceptions wizard, which is shared across all downstream *_exception modules (such as Purchase, Purchase Request, Sale, Account Move, and HR Expense), contains a validation bypass.

Although exceptions marked with is_blocking = True are intended to be non-ignorable, the wizard sets ignore_exception = True directly on the record instead of calling base.exception.action_ignore_exceptions(), which is the method responsible for enforcing the blocking check.

As a result, users can ignore blocking exceptions. Once ignore_exception is set to True, the record is excluded from all future exception checks, making the bypass persistent until the document is reset (for example, by reverting it to the Draft state).

## To Reproduce
**Affected versions**:

1. Create an Exception with is_blocking = True on
2. Create a Record that triggers the rule, click Confirm — the exception wizard opens as expected.
3. Check "Ignore Exceptions", click Close on the wizard.
4. Record Skip Exception

**Expected behavior**
A blocking exception (`is_blocking = True`) should never be ignorable from any entry point, including this wizard — it should raise the same error that `action_ignore_exceptions()` raises, and Confirm should stay blocked.

**Additional context**
Affects every module built on this wizard: `purchase_exception`, `purchase_request_exception` (OCA/purchase-workflow), `sale_exception` (OCA/sale-workflow), `account_move_exception` (OCA/account-invoicing), `hr_expense_exception` (OCA/hr-expense), and any custom `*_exception` module following the same pattern.

collinskipkorir:
Closes #3667

## Problem

The confirmation wizard could allow blocking exceptions to be ignored when a downstream exception module set `ignore_exception = True` directly.

That write can remove the linked exceptions before `action_ignore_exceptions()` validates whether any exception is blocking.

## Solution

Validate the wizard's original exception list before allowing exceptions to be ignored. The wizard then calls `action_ignore_exceptions()` to follow the standard ignore workflow.

## Testing

- Added regression coverage for a downstream wizard that writes `ignore_exception` directly.
- The complete `base_exception` test suite passes.
- Pre-commit, Ruff, Ruff Format, and Pylint checks pass.
- Manually tested using `purchase_exception`.
- Confirmed that a blocking purchase exception raises an error and cannot be ignore

I confirm I have read the DCO (https://github.com/OCA/oca-addons-repo-template/blob/master/CONTRIBUTING.md) and signed it off.

@thomaspaulb help me review this

### DIFF
--- a/base_exception/__manifest__.py
+++ b/base_exception/__manifest__.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "Exception Rule",
-    "version": "18.0.1.1.1",
+    "version": "18.0.1.1.2",
     "development_status": "Mature",
     "category": "Generic Modules",
     "summary": """
--- a/base_exception/tests/test_base_exception.py
+++ b/base_exception/tests/test_base_exception.py
@@ -141,6 +141,13 @@ class TestBaseException(TransactionCase):
         with self.assertRaises(UserError):
             self.po.action_ignore_exceptions()
         self.assertFalse(self.po.ignore_exception)
+        # The confirmation wizard must enforce the same blocking check, even
+        # when a downstream wizard sets ``ignore_exception`` directly.
+        self.exception_rule_confirm.exception_ids = self.po.exception_ids
+        self.exception_rule_confirm.ignore = True
+        with self.assertRaises(UserError):
+            self.exception_rule_confirm.action_confirm()
+        self.assertFalse(self.po.ignore_exception)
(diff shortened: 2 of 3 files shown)


collinskipkorir (commented):
I need a review to this

collinskipkorir:
hi
@sebastienbeau i need a review on this

PR #3686, opened 2026-07-27
[18.0][ADD] global_undo: undo/redo backend ops with Ctrl+Z

polqubiq:
## Why

You save a wrong change, confirm an order, post an invoice, or delete a record — and there is no Ctrl+Z. Cancel/draft flows, manual fixes, or a restore are the only options. Odoo has no general undo for backend work.

## What it does

- Records create / write / unlink and reversible actions (confirm SO/PO, post invoice) as **one step per form save**.
- **Ctrl+Z** / **Ctrl+Shift+Z** and a systray (undo, redo, last 10 steps). Inside inputs, Ctrl+Z stays native.
- History + trash under *Settings → Global Undo*. Unsafe undos are refused, not forced.

## Limits

Skips technical models, ledger rows, mass ops, sudo and non-web-client writes. Does not reverse mail/APIs. Respects accounting integrity.


polqubiq:
Hi @pedrobaeza I've just submitted a PR that I think could be very useful and provide a lot of value to the community.
It's a feature we've really been missing, and we're surprised Odoo still doesn't include it out of the box.

pedrobaeza:
Hi @polqubiq although it seems a very interesting thing, I think it's very risky, as unless you control and program the whole possible actions to undo/redo, you will get unexpected results installing this partially in your instance. It's the same as `mass_editing` (now called `server_action_mass_edit`). It can provoke inconsistent data, and I tend to avoid its usage. It may be also counterproductive for UX experience, as current business flows with explicit "Cancel" or "Reset to draft" buttons make aware the user of what's happening.

Anyway, this can be useful for some, or serve as inspiration for a core feature, so it's always welcome.

PR #3688, opened 2026-07-30
[FIX] odoo_test_xmlrunner: attribute class-level errors to the right file

yajo:
## Bug

When a `setUpClass` or `tearDownClass` fails, unittest creates an `_ErrorHolder`. Because `xmlrunner.result._XMLTestResult.startTest` is never called for these holders, `self.filename` keeps the value from the previous test. The JUnit report then attributes the failure to the wrong source file and leaves `classname` empty.

This was visible in the failing pipeline: https://gitlab.com/moduon/ataraxia/buffet-odoo/-/pipelines/2714813530/test_report?job_name=unittest%3A%20%5B19%5D

## Fix

Parse the `_ErrorHolder` description (`setUpClass/tearDownClass (module.ClassName)`), use `inspect` to locate the real test class, and patch the stored `_TestInfo` in `addError`/`addFailure` so the generated XML points to the correct file and line.

A regression test is included.

Forward-port: OCA/server-tools#3689

Assisted-by: OpenCode + kimi-k2.7-code

### DIFF
--- a/odoo_test_xmlrunner/odoo_tests/loader.py
+++ b/odoo_test_xmlrunner/odoo_tests/loader.py
@@ -1,10 +1,14 @@
+import inspect
 import os
+import re
 
 from odoo.tools import config
 
 if config["test_enable"]:
+    from unittest.suite import _ErrorHolder
+
     from xmlrunner import XMLTestRunner
-    from xmlrunner.result import _XMLTestResult
+    from xmlrunner.result import _XMLTestResult, failfast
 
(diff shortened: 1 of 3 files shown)


yajo:
To help understand better the motivation of the fix, check out this visual representation of a buggy junit xml in Gitlab UI:

image

As you see, a failure in the `setUpClass()` of one module reports wrong error line (that belongs to another, unrelated, module).

PR #3598, opened 2026-04-15
[MIG] odoo_test_xmlrunner: Migration to 19.0

yajo:
Nothing special

@moduon MT-14138

PR #3689, opened 2026-07-30
[FIX] odoo_test_xmlrunner: attribute class-level errors to the right file

yajo:
## Bug

Forward-port of the 18.0 fix for `_ErrorHolder` instances generated by failing `setUpClass`/`tearDownClass`. Because `xmlrunner.result._XMLTestResult.startTest` is never called for these holders, `self.filename` keeps the value from the previous test, so the JUnit report attributes the failure to the wrong source file and leaves `classname` empty.

## Fix

Parse the `_ErrorHolder` description (`setUpClass/tearDownClass (module.ClassName)`), use `inspect` to locate the real test class, and patch the stored `_TestInfo` in `addError`/`addFailure` so the generated XML points to the correct file and line.

A regression test is included.

Related: OCA/server-tools#3688

Assisted-by: OpenCode + kimi-k2.7-code


yajo:
Sorry I thought the module was already in OCA. I'll add the commit to #3598 instead.

PR #3692, opened 2026-08-04
[IMP] bus_alt_connection: redirect cron LISTEN connections

moylop260:
Since the cron workers are woken up through LISTEN cron_trigger, they also keep a long-lived listening connection to the "postgres" database, which is not compatible with PgBouncer in transaction pooling mode, the same limitation this module already solves for the bus dispatcher.

There is no hook to alter only that connection: both WorkerCron.start (prefork) and ThreadedServer.cron_thread (threaded) call sql_db.db_connect("postgres") directly. So flag the cron entry points with a thread-local and redirect their connections to the "postgres" database to the alternate host/port, while the connections actually processing the jobs keep going through the regular db_host/db_port (i.e. PgBouncer).

The new ODOO_CRON_DB_HOST/PORT (or cron_db_host/port) settings fall back to the existing imdispatcher ones, so a single configuration covers both LISTEN/NOTIFY connections.

### DIFF
--- a/bus_alt_connection/README.rst
+++ b/bus_alt_connection/README.rst
@@ -67,9 +67,9 @@ instances.
 It allows you to define how resources should be shared, according to
 your priorities, e.g. :
 
-- key odoo instance on host A can open up to 30 connections
-- while odoo instance on host B, dedicated to reports, can open up to 10
-  connections only
+-  key odoo instance on host A can open up to 30 connections
+-  while odoo instance on host B, dedicated to reports, can open up to
+   10 connections only
 
 And most importantly, it helps you to ensure that ``max_connections``
 will never be reached on pg server side.
@@ -80,8 +80,8 @@ Why is this module needed?
 When configuring PgBouncer, you can choose between 2 transaction pooling
 modes:
 
-- pool_mode = session
-- pool_mode = transaction
+-  pool_mode = session
+-  pool_mode = transaction
 
 If we choose pool_mode = session, then one server connection will be
 tied to a given odoo process until its death, which is exactly what
@@ -108,6 +108,18 @@ That's what this module implements, by overriding the relevant method of
 the
 `Dispatcher <https://github.com/odoo/odoo/blob/12.0/addons/bus/models/bus.py#L105>`__.
 
+What about cron workers?
+------------------------
+
+Since the cron workers were reworked to be woken up through
+``LISTEN cron_trigger``, they also keep a long-lived listening
+connection to the ``postgres`` database, with the same PgBouncer
+limitation as the bus dispatcher.
+
+This module also redirects those cron LISTEN connections to the
+alternate host/port, while the connections actually processing the jobs
+keep going through the regular ``db_host``/``db_port`` (i.e. PgBouncer).
+
 **Table of contents**
 
 .. contents::
@@ -120,8 +132,8 @@ You don't need to install this module in the database(s) to enable it.
 
 But you need to load it server-wide:
 
-- By starting Odoo with ``--load=web,bus_alt_connection``
-- Or by updating its configuration file:
+-  By starting Odoo with ``--load=web,bus_alt_connection``
+-  Or by updating its configuration file:
 
 .. code:: ini
 
@@ -134,12 +146,12 @@ Configuration
 
 You need to define how to connect directly to the database:
 
-- Either by defining environment variables:
+-  Either by defining environment variables:
 
-     - ``IMDISPATCHER_DB_HOST=db-01``
-     - ``IMDISPATCHER_DB_PORT=5432``
+      -  ``IMDISPATCHER_DB_HOST=db-01``
+      -  ``IMDISPATCHER_DB_PORT=5432``
 
-- Or in Odoo's configuration file:
+-  Or in Odoo's configuration file:
 
 .. code:: ini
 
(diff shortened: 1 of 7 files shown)


hugho-ad:
**Tested on:**

- Odoo 19.0, this PR's branch, both server modes: prefork (`--workers N`) and threaded (`--workers 0`)
- PgBouncer 1.25.2, `pool_mode = transaction`
- PostgreSQL 16.14
- Two endpoints reachable by Odoo: one via PgBouncer, one straight to PostgreSQL, so each connection can be attributed from the server side (`pg_stat_activity.client_addr`)
- Maintenance database `postgres`: where both long-lived listeners connect — `LISTEN cron_trigger` (crons) and `listen imbus` (bus)
- Business database: everything else, i.e. registry loading and cron job execution
- Metric: latency between the `NOTIFY` that triggers a cron and the job actually running. Independently of `LISTEN`/`NOTIFY`, cron workers also wake on a fixed 60s timer (`SLEEP_INTERVAL`), so under 5s means the `NOTIFY` reached the listener, and ~60s means only that timer did

**Claim under test**, from the PR description: the cron entry points get their
connections to `postgres` redirected to the alternate host/port, while the
connections that actually process the jobs keep going through
`db_host`/`db_port`.

**Confirmed**, in both server modes. Three things observed separately:

- The listening connection to `postgres` arrives on the direct endpoint.
- Business database connections still arrive via PgBouncer.
- Cron wake-up goes from the 60s timer to immediate; without `cron_db_*` set, same build, it stays at 60s.

---

### The module is optional if you control PgBouncer

In PgBouncer, `pool_mode` can be set per database in the `[databases]` section,
not only globally. Both listeners meet on the `postgres` database, so pinning
only that entry to `pool_mode = session` keeps the server connection assigned
for the life of the client, which is what a `LISTEN` needs — same outcome as
this PR, with no Odoo-side change at all.

`pgbouncer.ini`:

```ini
[databases]
; business traffic: transaction pooling, unchanged
* = host=pg-01 port=5432

; an exact match wins over the wildcard, so this applies to the maintenance
; database only -- the one both listeners connect to
postgres = host=pg-01 port=5432 pool_mode=session

[pgbouncer]
pool_mode = transaction
```

Odoo side: nothing. `db_host` / `db_port` point at PgBouncer as usual, the
module is not installed, not in the addons path, not in `server_wide_modules`,
and no `cron_db_*` / `imdispatcher_db_*` is set. That is how it was verified.

Connection cost of this shape is small: the only permanent connections on
`postgres` are the listeners themselves (one per cron worker plus the bus
dispatcher). Odoo does not keep maintenance-database connections in its own
pool, so the short-lived users of that database — the `NOTIFY` senders, database
listing, health checks — do not accumulate under session pooling.

`DESCRIPTION.md` rejects `pool_mode = session` as a global setting, which is
correct — globally it would tie a server connection to every Odoo process and
defeat the pooling. But it reads as if session pooling were ruled out entirely.
Worth acknowledging the per-database variant.

Two reasons to still prefer the module:

1. It works when PgBouncer is managed by someone else and its config is not yours to change.
2. Resilience — see below.

### Resilience: an undocumented benefit

If the listening connection goes through PgBouncer, restarting PgBouncer (config
reload, upgrade, container restart) takes the whole Odoo instance down. Chain
observed, in prefork:

1. The cron worker opens its listening connection in [`WorkerCron.start`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1471) and registers [`LISTEN cron_trigger`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1477). PgBouncer going down closes it.
2. The worker finds out on its next wake-up: [`WorkerCron.sleep`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1411) polls that same connection, which raises.
3. The exception escapes [`Worker.run`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1344) and the worker exits — `Worker WorkerCron (<pid>) Exception occurred, exiting...`.
4. The master now has an empty cron slot, and only in that case does [`process_spawn`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1032) call [`check_registries`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1018), which opens a cursor on the business database — still via PgBouncer, still down. An idle master never touches the database, which is why the restart is harmless as long as no worker dies.
5. That exception reaches the master loop's handler in [`PreforkServer.run`](https://github.com/odoo/odoo/blob/19.0/odoo/service/server.py#L1210-L1213): `stop(False)` and `return -1`. Instance down.

With the module, the listening connection is direct to PostgreSQL, so step 1
never happens and the chain never starts — the cron worker keeps its PID across
the restart and no respawn is attempted.

Arguably worse than the bug this PR fixes: crons a minute late vs. a full
instance shutdown on routine maintenance. Worth a couple of lines in
`DESCRIPTION.md`.

### Documentation

- `CONFIGURE.md`: `IMDISPATCHER_DB_HOST` / `_PORT` are missing the `ODOO_` prefix the module actually reads — as documented they do nothing. Pre-existing, but the section added here correctly uses `ODOO_CRON_DB_*`, so the file is now inconsistent with itself.
- `DESCRIPTION.md`: "the unique *listening* connection per instance" is no longer true after this PR — there is now one per cron worker (`max_cron_threads`) plus the bus dispatcher. Matters when sizing those direct connections against PostgreSQL's `max_connections`.
- `INSTALL.md`: an instance started with `-d` / `db_name` preloads a registry at boot, which imports the module early enough anyway — so the fix *appears* to work without `server_wide_modules`. It silently does nothing when Odoo boots without preloading any registry (no `-d` / `db_name`, the usual multi-database shape): the cron worker opens its listening connection before the module is ever imported. Worth an explicit warning, since that is the shape PgBouncer usually shows up in.
- `INSTALL.md` on 19.0: `--load=web,bus_alt_connection` drops `rpc` from the default `base,rpc,web` (only `base` and `web` are restored automatically), disabling the `/xmlrpc` and `/jsonrpc` endpoints. Suggest `base,rpc,web,bus_alt_connection`.

Nothing above blocks the merge.

moylop260:
After Hugo comments I think this module is not longer needed since that we can configure the pgbouncer to use different mode based on database

And the database connected from LISTEN/NOTIFY is using `postgres` database


So, only configuring is enough

```cfg
[databases]
; business traffic: transaction pooling, unchanged
* = host=pg-01 port=5432

; an exact match wins over the wildcard, so this applies to the maintenance
; database only -- the one both listeners connect to
postgres = host=pg-01 port=5432 pool_mode=session

[pgbouncer]
pool_mode = transaction
```

Thank you!

PR #3694, opened 2026-08-06
[18.0][FIX] dbs value has to be kept in case of special proxy config

ivilata:
Ported from #2828 using `oca-port`.

### DIFF
--- a/dbfilter_from_header/override.py
+++ b/dbfilter_from_header/override.py
@@ -13,11 +13,13 @@ db_filter_org = http.db_filter
 
 
 def db_filter(dbs, host=None):
-    dbs = db_filter_org(dbs, host)
+    dbs_orig = db_filter_org(dbs, host)
     httprequest = http.request.httprequest
     db_filter_hdr = httprequest.environ.get("HTTP_X_ODOO_DBFILTER")
     if db_filter_hdr:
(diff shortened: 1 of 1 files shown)


ivilata:
@thomaspaulb @hbrunn, you may want to review/approve this direct port from @cvinh's #2828, as you did for 16.0. Thanks!

thomaspaulb (approved):

hbrunn (changes requested):
reexamining this, it's actually quite a radical change that in any case makes

    Please keep in mind that the standard odoo dbfilter configuration is still applied before looking at the regular expression in the header.

from the readme false. So please update that too. And stress that it's super relevant to set the header *everywhere*.

Without this patch, you can set dbfilter = ^db1|db2|db3$ and be sure this Odoo instance will only ever be able to access those three dbs. With the patch, whenever one has access to the unproxied instance or one that doesn't set the header, you can pass whatever other database that's accessible to the database user, which might be an issue for some deployments.

So actually, I'm not sure any more I agree with this patch.

ivilata:
Thanks @hbrunn for the heads up and sorry for the noise, I failed to check if there was an existing PR for this. Since #3475 is already from the original author, I'll retract this PR to avoid dispersing the discussions as well. I'll copy your comment there so that people don't need to jump too much among PRs. :slightly_smiling_face:

PR #3699, opened 2026-08-10
Allow muting the startup message

kvdveer:
Odoo.sh loads the application quite frequently during normal operation (even when not restarting the server). Each time it does, sentry get reinitialized. This sends to a substantial number of messages to Sentry, draining a non-trivial portion of monitoring budget when using SaaS Sentry.

Giving users a way to mute this message will allow them to cut this noise.


StefanRijnhart (approved):
Thanks, can confirm the issue on Odoo.sh 😅

moylop260 (approved):

PR #3701, opened 2026-08-14
[ADD] orm_forward_compatibility: Odoo 19+ ORM helpers backported to 18.0

divad1196:
Minimal `Domain` shim (delegates to odoo.osv.expression) plus the typed `ir.config_parameter` getters get_str/get_int/get_float/get_bool.

Purpose: cut the effort of backporting 19/20 modules to 18.0 by keeping their call sites 1-1 with upstream instead of rewriting them.


legalsylvain:
interesting !

divad1196:
@legalsylvain Don't hesitate to leave a review ! :smile:

divad1196:
Review needed @OCA/server-tools-maintainers Possible review
@simahawk @santostelmo

orm_forward_compatibility/domain.py
```diff
@@ -0,0 +1,78 @@
...
+list-domain, a shim instance can be passed straight to ``search``/``_search``.
+
+Supported surface (extend as new call sites appear):
+- ``Domain([('a', '=', 1), ...])`` and ``Domain('a', '=', 1)`` constructors
+- ``&`` / ``|`` / ``~`` operators
+- ``Domain.AND(iterable)`` / ``Domain.OR(iterable)``
+- ``Domain.TRUE`` / ``Domain.FALSE``
+- ``.optimize_full(model)`` -> validates against the model, returns self
+
+NOT supported (keep such call sites hand-adapted on 18):
```
simahawk, line 21: which kind of impact has such limitation?

orm_forward_compatibility/models/ir_config_parameter.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2026 Camptocamp SA
```
simahawk, line 1: ⚠️ we should keeo Odoo SA copyright w/ the correct year and add ours only if we touched anything

simahawk (changes requested):

orm_forward_compatibility/domain.py
divad1196: I personnally didn't need them.
`any!` isn't popular yet, and `-1d` is usually done with date/datetime and timedelta.

orm_forward_compatibility/models/ir_config_parameter.py
divad1196: This is not copied from Odoo's code. 
Here is how Odoo does it: https://github.com/odoo/odoo/commit/22fa8e67b927c9605f783ea298ed0123338374b0

The code here only fill the gap by mimicking V20 interface.

legalsylvain:
> @legalsylvain Don't hesitate to leave a review ! 😄

too complex for me !

PR #3706, opened 2026-08-18
[16.0][IMP] scheduler_error_mailer - add retries

Kiplangatdan:


ntsirintanis:
@Kiplangatdan can we also add unit tests here, like the other mr?

ntsirintanis (approved):

PR #3707, opened 2026-08-19
[16.0][FIX] upgrade_analysis: prevent dependency on MarkupSafe >= 2.0.0

StefanRijnhart:
This halts test jobs on `could not obtain odoo.addons.__path__ using python` due to the incompatibility of Odoo 16's required version of Jinja2 on Python 3.10

```
++ oca_list_external_dependencies deb
could not obtain odoo.addons.__path__ using python
not found in addons path: base,fs_storage,mail_environment,queue_job,test_performance
```

Avoiding manifestoo obscuring the error with some debug code leads to the following traceback:
```
+ unbuffer /opt/odoo-venv/bin/odoo -d odoo -i account,base,base_import,base_setup,base_sparse_field,bus,crm,fs_storage,iap,mail,mail_environment,purchase,queue_job,sale_management,test_mail,web --http-interface=127.0.0.1 --stop-after-init
Traceback (most recent call last):
  File "/opt/odoo-venv/bin/odoo", line 5, in <module>
    import odoo
  File "/opt/odoo/odoo/__init__.py", line 155, in <module>
    from . import cli
  File "/opt/odoo/odoo/cli/__init__.py", line 11, in <module>
    from . import scaffold
  File "/opt/odoo/odoo/cli/scaffold.py", line 8, in <module>
    import jinja2
  File "/opt/odoo-venv/lib/python3.10/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/opt/odoo-venv/lib/python3.10/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/opt/odoo-venv/lib/python3.10/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/opt/odoo-venv/lib/python3.10/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/opt/odoo-venv/lib/python3.10/site-packages/markupsafe/__init__.py)
```

This points to a compatibility issue with MarkupSafe which has been updated to (currently) 3.0.3 by pip while resolving test dependencies:

<details>
[...]

PR #3708, opened 2026-08-20
[16.0][FIX] upgrade_analysis: Revert #3707 and do it on test-requirements.txt

pedrobaeza:
@Tecnativa

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cryptography
 dataclasses
 dnspython
 josepy
-mako<1.4.0
+mako
 numpy
 odoo_test_helper
 odoorpc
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@StefanRijnhart @sbidoul should I trigger the merge?

StefanRijnhart (commented):
I'm not sure I agree. I think the previous version would also fix the issue in some production deploys (depending on the method of installation) and in other OCA PRs on other repos that (theoretically) would pull in upgrade_analysis as a dependency.

sbidoul:
You are right that the same constraint will need to be applied on other OCA repos that depend on upgrade_analysis.

But assume a security vulnerability appears in mako < 1.4. With this you prevent everyone from upgrading, for a bad reason. If you upgrade jinja2 too everything probably works, since the problem seems to come from the old version of jinja2? 

It's complicated and I agree this solution is not ideal but its better than the direct upper bound, I'm quite sure of that.

There is a similar issue that pops up frequently with cryptography and openssl, and we apply a similar solution, because forcing everyone to use an old version of the cryptography library is obviously not a good thing to do.

StefanRijnhart (approved):
@sbidoul Alright, thanks for confirming.

Issue #3165, opened 2025-01-15
[17.0] Incorrect display of tracked one2many field in chatter (tracking_manager module)


melissalateb:
## Module
tracking_manager

## Describe the bug

- When I enable tracking on a _one2many_ field in a custom model using the tracking_manager module, the tracked changes appear in the chatter. However, instead of showing the display name or a meaningful value, the system displays the raw IDs of the records (e.g., Modified: sale.order.linked, 24) in the chatter.

This issue does not occur when tracking _one2many_ fields on standard models like sale.order.line. It seems specific to custom models. Am i missing something?

- The second thig is when changing a boolean field related (for example here sale.order.linked), it doesn't show both the before and after values; it only displays the new value with the arrow.

## To Reproduce

1. Create a custom model with a One2many field ( in my example : sale.order.linked).
2. Enable tracking on this field.
3. Create records with the _one2many_ field.
4. Modify a boolean related to _one2many_ field save the record.
5. Observe the chatter for tracked changes.

## Image:

**Create**
![Image](https://github.com/user-attachments/assets/74cbb109-5ce3-4119-9411-58633a0440e8)

**Write**
![Image](https://github.com/user-attachments/assets/4bdd4546-792e-4c46-b329-f2c4d6a6b703)

**Expected behavior**
- The same think as Order Lines for tracking value.
- And more details about value changes before and after.

**Additional context**

Odoo 17.0 enterprise

CRogos:
False might be interpreted as `<empty>` and therefore an empty string is displayed. 
Where does the model sale.order.linked comes form? Could it be that this model does not have a display_name attribute? Therefore it displays model+id as fallback?

CRogos:
I would like to add this finding when activating tracking on "child_ids" in res.partner:

- Two message are created. One odoo native and a second by tracking_manager.
- Content of New and Change in the tracking_manager message should be merged.

![Image](https://github.com/user-attachments/assets/e5f63c1c-d320-42f5-af41-73d5ebabbd63)

Maybe we can update the message content of the native message with the more detailed content of the tracking_manager message?

melissalateb:
@CRogos ,

The model `sale.order.linked` is a custom model that I developed, and it doesn't have a display_name defined. Should I explicitly define one to make the tracking work properly?

PR #3712, opened 2026-08-24
[FIX] tracking_manager: display booleans as Yes/No

maciej-wichowski:
A boolean was rendered by the message template as its python value, so a change to `False` showed nothing at all: `Archived : True -> `. Render both values the way the web client renders native tracking values.


maciej-wichowski:
Related to issue #3165

PR #3725, opened 2026-09-14
[18.0][FIX] base_exception: MissingError on records created in the ongoing transaction

sergio-teruel:
detect_exceptions() opens a second, genuinely independent DB connection (self.env.registry.cursor()) to write exception flags so they survive a rollback of the ongoing transaction. It then re-derived main_records through that same second connection (self.with_env(new_env)), which crashes with MissingError whenever _get_main_records() has to read a field on records created earlier in the SAME ongoing transaction and not yet committed (e.g. sale.order.line._get_main_records() reading order_id right after the line was created by write()).

Typical trigger: adding a product to an already confirmed sale order. sale_exception.write() re-runs detect_exceptions() on order_line whenever order_line changes on a confirmed order; the new line only exists in the ongoing transaction, so the second connection can't see it yet.

Resolve main_records once through the original (main) cursor, which does see the just-created records, and only rebind that already-resolved recordset to the new environment for the exception_ids check.

@Tecnativa TT64467

ping @carlosdauden @carlos-lopez-tecnativa


hparfr:
FYI @florian-dacosta @grindtildeath

carlosdauden (approved):

grindtildeath (approved):
Good catch 👍

hparfr (approved):
/ocabot merge minor

PR #3734, opened 2026-09-18
[FIX] base_exception: reload the view after the exception dialog

lef-adhoc:
### The problem

Confirm a sales order that hits an exception rule. The dialog shows up, you tick **Ignore Exceptions** and confirm.

The order **is** confirmed — but the view behind the dialog still shows it as a quotation, with the red exception banner and the Confirm button. You have to reload the page by hand to see the real state, and until you do you can hit Confirm again on an order that is already confirmed.

### How to reproduce

On 18.0 with `sale_exception` installed:

1. Add an `exception.rule` on `sale.order`, `by_py_code`, that matches one customer.
2. Create an order for that customer and confirm it → the exception dialog shows up.
3. Tick **Ignore Exceptions** and confirm.
4. The view still shows *Quotation*, but the order is already `sale` in the database. Reload the page and the real state appears.

### Why

The dialog can change the record: ticking "Ignore Exceptions" calls `action_confirm` on it. But the wizard returns `ir.actions.act_window_close`, which closes the dialog and nothing else, so the view keeps the state it had before.

### The fix

Reload once the dialog is closed, using the `onClose` option of `doAction`:

```js
await env.services.action.doAction(popupAction, {
    onClose: () =>
        env.services.action.doAction({
            type: "ir.actions.client",
            tag: "soft_reload",
        }),
});
```

It goes here rather than in the wizard's return value for two reasons: `action_confirm` on the wizard is inherited by several modules, and this way the fix covers every model using the mechanism, not just sales.


lef-adhoc:
@grindtildeath could you take a look? This follows up on #3590 and OCA/sale-workflow#4255, so you are the one with the context.

The dialog can now confirm the record, and closing it leaves the view showing the previous state. Reloading on close seemed like the smallest fix

UsmanGhias (commented):
Clean patch for OCA/server-tools.

Observations:
- Template and component structure preserves standard inheritance patterns.

Ready for testing on standard environments.

Best,
Usman

grindtildeath (approved):
Makes sense, thank you 👍

